PR #11, opened 2018-09-22
Delete IAM resources if associated instance is not running

sallyom:
@dgoodwin @joelddiaz I noticed if a cluster doesn't complete successfully, there is likely  IAM roles/policies/profiles that are not associated with a running instance (bc they are created in the install process b4 the instance).  The deprovisioner was dependent upon an instance running to delete the IAM resources.  I've added the function of deleting IAM resources for the specific profiles we know to be created with the installer. `${CLUSTER_NAME-master-profile` `${CLUSTER_NAME-worker-profile` and `${CLUSTER_NAME-bootstrap-profile`
have to now pass `--cluster-name mycluster` along with the tags
I've tested different scenerios:  failed launch/successful launch/single/multiple masters  this works well

### DIFF
--- a/contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
+++ b/contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -55,14 +56,15 @@ type awsObjectWithTags struct {
 // deleteFunc type is the interface a function needs to implement to be called as a goroutine.
 // The (bool, error) return type mimics wait.ExponentialBackoff where the bool indicates successful
 // completion, and the error is for unrecoverable errors.
-type deleteFunc func(awsClient *session.Session, filters awsFilter, logger log.FieldLogger) (bool, error)
+type deleteFunc func(awsClient *session.Session, filters awsFilter, clusterName string, logger log.FieldLogger) (bool, error)
 
 // ClusterUninstaller holds the various options for the cluster we want to delete
 type ClusterUninstaller struct {
-	Filters  awsFilter // filter(s) we will be searching for
-	Logger   log.FieldLogger
-	LogLevel string
-	Region   string
+	Filters     awsFilter // filter(s) we will be searching for
+	Logger      log.FieldLogger
+	LogLevel    string
+	Region      string
+	ClusterName string
 }
 
 // NewDeprovisionAWSWithTagsCommand is the entrypoint to create the 'aws-tag-deprovision' subcommand
@@ -89,6 +91,7 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
 	flags := cmd.Flags()
 	flags.StringVar(&opt.LogLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
 	flags.StringVar(&opt.Region, "region", "us-east-1", "AWS region to use")
+	flags.StringVar(&opt.ClusterName, "cluster-name", "", "Name that cluster was installed with")
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
```diff
@@ -697,9 +682,43 @@ func deleteInstanceProfile(instanceProfileID *string, iamClient *iam.IAM, logger
...
+			err = deleteInstanceProfile(profile.InstanceProfile.InstanceProfileId, iamClient, logger)
+			if err != nil {
+				logger.Debugf("error deleting instance profile: %v", err)
+				return false, nil
+			}
+		}
+	}
+	return true, nil
+}
+
```
sallyom: there could be a way so we don't have to depend on the hardcoded installer_type above.. or a better name than `installer_type`..

contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
```diff
@@ -252,25 +258,6 @@ func tagsToMap(tags interface{}) (map[string]string, error) {
...
-		if *ip.Arn == *ipARN {
-			return ip.InstanceProfileId, nil
-		}
-	}
-
-	// if we didn't find it, that means it may already have been deleted
-	return nil, nil
-}
-
 // lbToAWSObjects will create awsObjectWithTags given a list of load balancers
```
sallyom: That function wasn't being used, so I've cleaned it up here..

.gitignore
```diff
@@ -4,8 +4,6 @@
 *.dll
 *.so
 *.dylib
-<<<<<<< HEAD
-=======
```
sallyom: that isn't supposed to be there, right?

contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
```diff
@@ -697,9 +681,88 @@ func deleteInstanceProfile(instanceProfileID *string, iamClient *iam.IAM, logger
...
+	logger.Debugf("deleting instance profile: %v", profileName)
+	_, err = iamClient.DeleteInstanceProfile(&iam.DeleteInstanceProfileInput{
+		InstanceProfileName: &profileName,
+	})
+	if err != nil && err.(awserr.Error).Code() != iam.ErrCodeNoSuchEntityException {
+		logger.Debugf("error deleting instance profile %s: %v", profileName, err)
+		return err
+	}
+	if err == nil {
+		logger.Debugf("deleted profile %s", profileName)
```
joelddiaz: Infof() for non-error

contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
```diff
@@ -697,9 +681,88 @@ func deleteInstanceProfile(instanceProfileID *string, iamClient *iam.IAM, logger
...
+	// now we can delete role
+	logger.Debugf("deleted policies from role %s", roleName)
+	deleteRoleInput := iam.DeleteRoleInput{}
+	deleteRoleInput.RoleName = &roleName
+	if _, err := iamClient.DeleteRole(&deleteRoleInput); err != nil && err.(awserr.Error).Code() != iam.ErrCodeNoSuchEntityException {
+		logger.Debugf("error deleting role %s: %v", roleName, err)
+		return err
+	}
+	if err == nil {
+		logger.Debugf("deleted role %s", roleName)
```
joelddiaz: Infof() for non-error

contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
```diff
@@ -697,9 +681,88 @@ func deleteInstanceProfile(instanceProfileID *string, iamClient *iam.IAM, logger
...
+
+// deleteIAMresources will delete any IAM resources created by the installer that are not associated with a running instance
+// Currently openshift/installer creates 3 roles per cluster, 1 for master|worker|bootstrap and identified by the
+// cluster name used to install the cluster.
+func deleteIAMresources(session *session.Session, filter awsFilter, clusterName string, logger log.FieldLogger) (bool, error) {
+	logger.Debugf("Deleting IAM resources")
+	defer logger.Debugf("Exiting deleting IAM resources")
+	installer_type := []string{"master", "worker", "bootstrap"}
+	for _, t := range installer_type {
+		roleName := fmt.Sprintf("%s-%s-role", clusterName, t)
```
joelddiaz: can we put a comment here explaining the expected naming scheme for the various IAM objects?

contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
sallyom: fixed

contrib/pkg/aws_tag_deprovision/aws_tag_deprovision.go
sallyom: fixed

PR #13, opened 2018-09-24
Add make verify target

twiest:
- [x] Add make verify target
- [x] Add verify-lint target
- [x] Add verify-gofmt target
- [x] Add verify-go-vet target
- [x] Fix make verify-imports target to error when verify-imports fails.
- [x] Fix golint errors in codebase
- [x] Remove fmt and vet targets from build targets (necessary for make verify to work)
- [x] Add .PHONY to Makefile targets that don't produce files of the same name.


twiest:
@dgoodwin hey, I rebased on master and had to update some of the new code to comply with `go vet`'s checks.

Would you mind reviewing this again please?

PR #16, opened 2018-09-25
initial buildah support

joelddiaz:
fedora's docker version doesn't support some of the syntax in the Dockerfile generated by kubebuilder

use builah for building the container


dgoodwin:
Makefile has conflicts after Thomas' PR, but this LGTM otherwise. We can do whatever we want to that Dockerfile but the layered build stuff is now supported in OSBS and the automated release team tooling (or very soon), and I think it's recommended practice and other teams using it as well, so it's probably best to keep that in the Dockerfile.

PR #21, opened 2018-09-27
Modify clusterdeployment copy in controller reconcile

csrwng:
We should be using the copy of the clusterdeployment and not the original one to make modifications.


csrwng:
@dgoodwin ptal

dgoodwin:
/lgtm 

Thanks!

PR #25, opened 2018-09-28
Use new installer to run cluster install job

csrwng:
Separates artifact generation into its own package
Adds necessary environment vars to job to create cluster


dgoodwin:
Awesome!

/lgtm

PR #26, opened 2018-09-28
Fix load balancer deletion hang

ironcladlou:
If the number of load balancers to delete is zero, `lbToAWSObjects` throws an
error which is swallowed by the caller, causing an infinite retry loop in the
exponential backoff wait.

Make the LB deletion function idempotent by correctly handling empty LB lists.


ironcladlou:
The underlying error which is thrown when the LB list is empty:

```
ERRO[2018-09-28T15:21:42-04:00] Error processing load balancers: Error fetching tags for load balancers: InvalidParameter: 1 validation error(s) found.                                 
- missing required field, DescribeTagsInput.LoadBalancerNames. 
```

So, it can't self resolve.

ironcladlou:
Just FYI, the current code still assumes that all errors returned by `DescribeTags` are transient (in the case this PR fixes, a request validation error, the error is actually permanent).

PR #30, opened 2018-10-01
InstallConfig: use secret references instead of strings

csrwng:
Switch to secrets for ssh public key, admin password, and image pull secret

### DIFF
--- a/config/templates/cluster-deployment.yaml
+++ b/config/templates/cluster-deployment.yaml
@@ -45,6 +45,30 @@ objects:
     awsAccessKeyId: ${AWS_ACCESS_KEY_ID}
     awsSecretAccessKey: ${AWS_SECRET_ACCESS_KEY}
 
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: ${CLUSTER_NAME}-pull-secret
+  type: kubernetes.io/dockercfg
+  stringData:
+    ".dockercfg": "${PULL_SECRET}"
+
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: ${CLUSTER_NAME}-ssh-key
+  type: Opaque
+  stringData:
+    ssh-publickey: "${SSH_KEY}"
+
+- apiVersion: v1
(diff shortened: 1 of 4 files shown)


csrwng:
@dgoodwin ptal

config/templates/cluster-deployment.yaml
```diff
@@ -45,6 +45,30 @@ objects:
     awsAccessKeyId: ${AWS_ACCESS_KEY_ID}
     awsSecretAccessKey: ${AWS_SECRET_ACCESS_KEY}
 
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: ${CLUSTER_NAME}-pull-secret
+  type: kubernetes.io/dockercfg
+  stringData:
+    ".dockercfg": "${PULL_SECRET}"
```
dgoodwin: The . here makes me a little nervous for bugs, but could we rename to something more like the others, and non-docker specific. (even pull-secret)

config/templates/cluster-deployment.yaml
```diff
@@ -45,6 +45,30 @@ objects:
...
+  stringData:
+    ssh-publickey: "${SSH_KEY}"
+
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: ${CLUSTER_NAME}-admin-creds
+  type: Opaque
+  stringData:
+    password: letmein
```
dgoodwin: Should be ADMIN_PASSWORD I think.

config/templates/cluster-deployment.yaml
csrwng: Afaik, this is the standard key name for a secret of type "kubernetes.io/dockercfg" (take a look at service account secrets in your cluster). If .dockercfg is not used, I would most likely get a validation error.

config/templates/cluster-deployment.yaml
csrwng: yes :)

csrwng:
Pull updated

PR #32, opened 2018-10-02
Hive Install Manager Sidecar Container

dgoodwin:
Modifies the installer container to just copy the install artifacts to a shared empty volume. Adds another sidecar container with a new hiveutil command that will manage execution of the installer, and upload required artifacts after it completes.

### DIFF
--- a/contrib/pkg/installmanager/installmanager.go
+++ b/contrib/pkg/installmanager/installmanager.go
@@ -1,0 +1,285 @@
+package installmanager
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	// metadataRelativePath is the location of the installers cluster metadata file
+	// relative to our WorkDir.
+	metadataRelativePath        = "metadata.json"
+	adminKubeConfigRelativePath = "auth/kubeconfig"
+)
+
+// InstallManager coordinates executing the openshift-install binary, modifying
+// generated assets, and uploading artifacts to the kube API after completion.
+type InstallManager struct {
+	log         log.FieldLogger
+	LogLevel    string
+	WorkDir     string
+	ClusterName string
+	Namespace   string
+	KubeClient  clientset.Interface
+}
+
+// NewInstallManagerCommand is the entrypoint to create the 'install-manager' subcommand
+func NewInstallManagerCommand() *cobra.Command {
+	opts := &InstallManager{}
+	cmd := &cobra.Command{
+		Use:   "install-manager NAMESPACE CLUSTER_NAME",
+		Short: "Executes and oversees the openshift-installer.",
+		Long:  "The Hive Install Manager runs the phases of the openshift-installer, edits generated assets before completing install, and monitors for artifacts that need to be uploaded back to Hive.",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := opts.Complete(args); err != nil {
+				log.WithError(err).Error("cannot complete command")
+				return
+			}
+
+			if len(args) != 2 {
+				cmd.Help()
+				opts.log.Fatal("invalid command arguments")
+			}
+			// Parse the namespace/name for our cluster deployment:
+			opts.Namespace, opts.ClusterName = args[0], args[1]
+
+			if err := opts.Validate(); err != nil {
+				log.WithError(err).Error("invalid command options")
+				return
+			}
+			if err := opts.Run(); err != nil {
+				log.WithError(err).Error("runtime error")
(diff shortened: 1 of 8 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -140,7 +150,10 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 
 	if !HasFinalizer(cd, hivev1.FinalizerDeprovision) {
 		cdLog.Debugf("adding clusterdeployment finalizer")
-		return reconcile.Result{}, r.addClusterDeploymentFinalizer(cd)
+		err = r.addClusterDeploymentFinalizer(cd)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
```
dgoodwin: Does this change look ok to you @csrwng ? Fixed a problem with the unit tests.

csrwng: So the difference now is that adding the finalizer successfully will result in the sync loop continuing rather than stopping right here. I don't necessarily see a problem with it, but traditionally we've only done one update to the resource we're reconciling per sync loop invocation.

I think we should fix unit tests though because they seem mighty flaky. I am currently running tests on the master branch and they only succeed 3/5 times.

contrib/pkg/installmanager/installmanager.go
```diff
@@ -0,0 +1,242 @@
...
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVar(&opts.LogLevel, "log-level", "info", "log level, one of: debug, info, warn, error, fatal, panic")
+	flags.StringVar(&opts.WorkDir, "work-dir", "/output", "directory to use for all input and output")
+	return cmd
+}
+
+func (m *InstallManager) complete(args []string) error {
```
csrwng: Make public (that would allow executing this command from another command, if needed)

contrib/pkg/installmanager/installmanager.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	m.WorkDir = absPath
+	if _, err := os.Stat(m.WorkDir); os.IsNotExist(err) {
+		m.log.WithField("workdir", m.WorkDir).Fatalf("workdir does not exist")
+	}
+
+	return nil
+}
+
+func (m *InstallManager) validate() error {
```
csrwng: Make public

Makefile
```diff
@@ -41,7 +41,7 @@ install: manifests
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 .PHONY: deploy
 deploy: manifests docker-build
-	kubectl apply -f config/crds
+	kubectl apply -f config/crds --validate=false
```
csrwng: is there a reason you turned off validation?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -244,6 +250,101 @@ func (r *ReconcileClusterDeployment) removeClusterDeploymentFinalizer(cd *hivev1
...
+			{
+				APIGroups: []string{""},
+				Resources: []string{"secrets", "configmaps"},
+				Verbs:     []string{"create", "delete", "get", "list", "update"},
+			},
+		},
+	}
+	currentRole := &rbacv1.Role{}
+	err = r.Client.Get(context.Background(), client.ObjectKey{Name: roleName, Namespace: namespace}, currentRole)
+	cdLog.WithError(err).Warnf("somethings up")
```
csrwng: probably a debugging stmt? :)

pkg/install/generate.go
```diff
@@ -140,24 +139,35 @@ func GenerateInstallerJob(
...
 			ImagePullPolicy: installerImagePullPolicy,
 			Env:             env,
-			Args:            args,
+			Command:         []string{"/bin/sh", "-c"},
+			Args:            []string{"cp -v /bin/openshift-install /output && cp -v /bin/terraform /output && ls -la /output"},
+			VolumeMounts:    volumeMounts,
+		},
+		{
+			Name:            "hive-manager",
+			Image:           "hive-controller",
```
csrwng: We should allow you to configure this image (as well as the installer one)

Makefile
dgoodwin: I'm having problems with this for all our CRDs when I try to deploy them on a 3.11 oc cluster up, it doesn't like the status fields, complains about missing required fields in status. Will remove this for now as I wasn't able to get it working anyhow.

pkg/install/generate.go
dgoodwin: Do we think we'll just have one hive image to run, for multiple possible versions of the installer image?

If so then we could probably still get by with what we last talked about, just specifying the installer image to use directly on your cluster deployment, and presumably the hive image would be configured as part of the deployment template or something like that.

If not, then I think we might be talking ClusterVersion again.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Found the problem, back to original form.

csrwng:
@dgoodwin please update the README and remove the bit about annotating the clusterdeployment with the cluster ID. Other than that, lgtm

dgoodwin:
Done! 

Looks like I'm getting intermittent test fails too now.

PR #33, opened 2018-10-03
Add unit tests for clusterdeployment controller

csrwng:
Moves generated test to integration directory
Adds unit tests with the fake client


pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -17,46 +17,195 @@ limitations under the License.
...
+			existing: []runtime.Object{
+				testClusterDeploymentWithoutFinalizer(),
+			},
+			validate: func(c client.Client) {
+				cd := getCD(c)
+				if cd == nil || !HasFinalizer(cd, hivev1.FinalizerDeprovision) {
+					t.Errorf("did not get expected clusterdeployment finalizer")
+				}
+			},
+		},
```
dgoodwin: This does look a lot better.

PR #35, opened 2018-10-04
Port support for cluster delete-after annotation.

abutcher:


dgoodwin:
Looks good but does appear to be a legit test failure in there.

PR #37, opened 2018-10-10
Add ownerref to uploaded artifacts.

dgoodwin:
The admin kubeconfig secret, and cluster metadata config map, should be cleaned up when a cluster is deleted.

### DIFF
--- a/contrib/pkg/installmanager/installmanager.go
+++ b/contrib/pkg/installmanager/installmanager.go
@@ -2,6 +2,7 @@ package installmanager
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


dgoodwin:
@csrwng could you review this one when time allows.

contrib/pkg/installmanager/installmanager.go
```diff
@@ -61,6 +73,13 @@ func NewInstallManagerCommand() *cobra.Command {
 				log.WithError(err).Error("invalid command options")
 				return
 			}
+
+			var err error
+			opts.KubeClient, opts.DynamicClient, err = getClients()
+			if err != nil {
+				opts.log.WithError(err).Fatal("error creating kube clients")
+			}
+
```
csrwng: Is there a reason you get the client here now instead of in the Run method?

contrib/pkg/installmanager/installmanager.go
```diff
@@ -239,6 +237,19 @@ func (m *InstallManager) uploadClusterMetadata() error {
...
 	}
+
+	cd := &hivev1.ClusterDeployment{}
+	err = m.DynamicClient.Get(context.Background(), types.NamespacedName{Namespace: m.Namespace, Name: m.ClusterName}, cd)
+	if err != nil {
+		m.log.WithError(err).Error("error getting cluster deployment")
+		return err
+	}
+	m.log.Debug("got cluster deployment")
+	if err := controllerutil.SetControllerReference(cd, metadataCfgMap, scheme.Scheme); err != nil {
```
csrwng: shouldn't you also set the controller reference on the kubeconfig secret?

contrib/pkg/installmanager/installmanager.go
dgoodwin: Felt a little like this was the constructor, but I see it's for the command not the install manager. Will move it to run.

contrib/pkg/installmanager/installmanager.go
dgoodwin: Oops!

csrwng:
lgtm ... were you going to squash the commits?

dgoodwin:
Wasn't planning to no as they were discrete, but can if you would prefer.

csrwng:
Nope
/lgtm

PR #43, opened 2018-10-18
Generate Installer's InstallConfig and Mount in Installer Pod

dgoodwin:
Restores the ConfigMap containing a install config, now converts our config to the very similar installer InstallConfig.


csrwng:
Looks like you need to rebase but lgtm

dgoodwin:
Conflicts resolved. I *think* overlapping mounts is ok, I was able to find references to bugs with it being fixed :) and it appears to work now.

dgoodwin:
Going to remove the overlapping mount in next PR anyhow, just realized I need to copy the install config due to how the installer wants to use and delete it, and the fact we're trying to share the output volume. Will put it somewhere else, and hiveutil will copy it in.

PR #47, opened 2018-10-18
contrib/pkg/awstagdeprovision: Allow for OR filters

wking:
Builds on #46; review that first.

AWS docs for filters are a bit sparse, but [the AWS docs][1] have:

> Combining search filters
>
> In general, multiple filters with the same key field (for example, tag:Name, search, Instance State) are automatically joined with OR.  This is intentional, as the vast majority of filters would not be logical if they were joined with AND.  For example, you would get zero results for a search on Instance State=running AND Instance State=stopped.  In many cases, you can granulate the results by using complementary search terms on different key fields, where the AND rule is automatically applied instead.  If you search for tag: Name:=All values and tag:Instance State=running, you get search results that contain both those criteria.  To fine-tune your results, simply remove one filter in the string until the results fit your requirements.

Unfortunately, there seems to be no way to represent OR filters in [ec2's `Filter` structure][2].  With the approach I have here, resources that match multiple filters will be fetched multiple times, and may have parallel, racing delete attempts.  But we should be robust in the face of racing delete attempt, and hopefully one of the deletes will go through before too long ;).

It would be possible to adjust our locally-filtered types (which use `filterObjects`) to avoid this issue for those types.  I may do that in follow-up work, but for now I'm treating all of our types the same way.

[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html#advancedsearch
[2]: https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#Filter

### DIFF
--- a/contrib/pkg/awstagdeprovision/awstagdeprovision.go
+++ b/contrib/pkg/awstagdeprovision/awstagdeprovision.go
@@ -60,7 +60,23 @@ type deleteFunc func(awsClient *session.Session, filters AWSFilter, clusterName 
 
 // ClusterUninstaller holds the various options for the cluster we want to delete
 type ClusterUninstaller struct {
-	Filters     AWSFilter // filter(s) we will be searching for
+
+	// Filters is a slice of filters for matching resources.  A
+	// resources matches the whole slice if it matches any of the
+	// entries.  For example:
+	//
+	//   filter := []map[string]string{
+	//     {
+	//       "a": "b",
+	//       "c": "d:,
+	//     },
+	//     {
+	//       "d": "e",
+	//     },
+	//   }
+	//
+	// will match resources with (a:b and c:d) or d:e.
+	Filters     []AWSFilter // filter(s) we will be searching for
 	Logger      log.FieldLogger
 	LogLevel    string
 	Region      string
@@ -107,15 +123,20 @@ func (o *ClusterUninstaller) Run() error {
 	}
 
 	// launch goroutines
+	goroutines := 0
 	for name, function := range deleteFuncs {
-		go deleteRunner(name, function, awsSession, o.Filters, o.ClusterName, o.Logger, returnChannel)
+		for _, filter := range o.Filters {
+			go deleteRunner(name, function, awsSession, filter, o.ClusterName, o.Logger, returnChannel)
+			goroutines++
+		}
 	}
 
 	// wait for them to finish
-	for i := 0; i < len(deleteFuncs); i++ {
+	for goroutines > 0 {
 		select {
 		case res := <-returnChannel:
-			o.Logger.Debugf("goroutine %v complete", res)
+			goroutines--
+			o.Logger.Debugf("goroutine %v complete (%d left)", res, goroutines)
 		}
 	}
 
@@ -211,8 +232,8 @@ func filterLBsByVPC(lbs []*elb.LoadBalancerDescription, vpc *ec2.Vpc, logger log
 // deleteLBs finds all load balancers under the provided VPC and attempts to delete them
 // returns bool representing whether it has completed its work (ie no LBs left to delete)
 func deleteLBs(vpc *ec2.Vpc, awsSession *session.Session, logger log.FieldLogger) bool {
-	logger.Debugf("Deleting load balancers")
-	defer logger.Debugf("Exiting deleting load balancers")
+	logger.Debugf("Deleting load balancers (%s)", vpc.VpcId)
+	defer logger.Debugf("Exiting deleting load balancers (%s)", vpc.VpcId)
 	elbClient := elb.New(awsSession)
 
 	describeLoadBalancersInput := elb.DescribeLoadBalancersInput{}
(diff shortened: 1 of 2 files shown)


abhinavdahiya:
https://github.com/openshift/hive/pull/47/commits/553c02aeb6744a3e575e66bd840fc643cbb34915 https://github.com/openshift/hive/pull/47/commits/8de157dde0df018c4b24c33e50e91d0b48a02b1e merged

https://github.com/openshift/hive/pull/47/commits/6e7bacc045eda07562b870e18f3809d8694c6c01 doesn't look like necessary and can be merged separate to the PR, so just drop it maybe?

https://github.com/openshift/hive/pull/47/commits/7bdaa6a62e6defcdffa62591cd83dbcf01d23a27 `/lgtm`

wking:
I've pushed 7bdaa6a -> f59dd98, rebasing onto master so GitHub only shows the new-to-this-PR commit in the diff.

> But i'm confused.. why force only one use the first filter? 7bdaa6a#diff-6df0d52da59598889c66f03dd6a7bc67R35 has `aws-tag-deprovision KEY=VALUE ...` `...` means more than one pair might be accepted ?

The `hiveutil` command has accepted multiple filters [since it landed][1], so nothing new there (although this was only documented in #46).  This PR is adding OR support to the library, but not exposing that support in `hiveutil`.  You can still list multiple `KEY=VALUE`, and they'll all get fed into `Filters[0]` (and be AND-ed together by the library).

[1]: https://github.com/openshift/hive/commit/ca98a4fd079b124bf2d14dc262c09a6ba1afb2f8#diff-78131bdbc1d1f117ded34d4efe0ce5eeR97

joelddiaz:
@wking i'm seeing a panic when i tried to run this
```
[jdiaz@goomba hive (wking-or-tag-deletion %)]$ ./bin/hiveutil aws-tag-deprovision --cluster-name fakecluster --loglevel=debug tectonicClusterID=fakeuuid kubernetes.io/cluster/fakecluster=shared kubernetes.io/cluster/fakecluster=owned
panic: assignment to entry in nil map

goroutine 1 [running]:
main.parseFilter(0x0, 0x7ffde7f5a96b, 0x1a, 0x0, 0x16c5b77)
        /home/jdiaz/projects/hive/src/github.com/openshift/hive/contrib/cmd/hiveutil/awstagdeprovision.go:88 +0xa4
```

wking:
>     panic: assignment to entry in nil map

Oops.  Fixed in f59dd98 -> d846d34.

wking:
> We should allow passing of multiple key=value tags that we can OR against.

Currently you can pass multiple to AND.  Did you have a command-line syntax in mind for OR?

joelddiaz:
I was thinking that each command line argument is what we OR against. No fancy combinations of ORs with ANDs...you just get OR. Do we have a need for more complex matching than ORing against a list of key=value pairs at this time?

wking:
> I was thinking that each command line argument is what we OR against.

`hiveutil` has supported multiple `KEY=VALUE` pairs ANDed together since it landed (notes [here][1]).  You're suggesting I change that to use OR instead?

> Do we have a need for more complex matching than ORing against a list of key=value pairs at this time?

The installer doesn't, but the installer doesn't use `hiveutil` ;).  Folks who want to destroy clusters should be using `openshift-install destroy cluster`.  Do you know who uses `hiveutil` and what their use-cases are?

[1]: https://github.com/openshift/hive/pull/46#issue-224106991

joelddiaz:
Yes, I'm suggesting to switch from ANDing the list of key=value pairs to ORing.

@dgoodwin what do you think about changing the default away from ANDing the list to ORing (and taking away the ability to do any kind of ANDing until we find a use-case for it)?

dgoodwin:
I suspect pretty much everyone using it would be specifying single tags, but it's a little hard to know for sure and a tiny bit risky. It's probably fine to do so but we might want to communicate it a little bit.

Would a --or option seem ok?

joelddiaz:
Are you suggesting AND-by-default, and --or will switch to 'OR all of thes key=value pairs passed as parameters'?

dgoodwin:
Yes, but I am up for either, mild risk in changing behavior from AND to OR on folks who might be using it, but I don't think there are a lot.

joelddiaz:
@wking i'm more than willing to make the proposed changes if you don't want to. you want to incorporate the OR-by-default changes?

wking:
I've pushed 971d238 -> 7522689, adding filter qualifiers to some debug logging and converting `hiveutil` to the OR semantics.

contrib/pkg/awstagdeprovision/awstagdeprovision.go
```diff
@@ -108,7 +124,9 @@ func (o *ClusterUninstaller) Run() error {
 
 	// launch goroutines
 	for name, function := range deleteFuncs {
-		go deleteRunner(name, function, awsSession, o.Filters, o.ClusterName, o.Logger, returnChannel)
+		for _, filter := range o.Filters {
+			go deleteRunner(name, function, awsSession, filter, o.ClusterName, o.Logger, returnChannel)
+		}
 	}
 
 	// wait for them to finish
```
joelddiaz: need to increase the number of goroutines we're waiting to complete now that we launch per-key=value.

```suggestion
	// wait for them to finish
       for i := 0; i < len(deleteFuncs) * len(o.Filters); i++ {
```

wking: > need to increase the number of goroutines we're waiting to complete...

Good catch :).  Fixed with 7522689 -> b1cad98.

PR #51, opened 2018-10-24
awstagdeprovision: Fix bucket name in debug output

Miciah:
Dereference the bucket name string pointer when formatting the "error getting tags" debug output.  Neglecting to do so produces error messages like the following:

    DEBUG error getting tags for bucket %!s(*string=0xc420433708): [...]

* `contrib/pkg/awstagdeprovision/awstagdeprovision.go` (`bucketsToAWSObjects`): Dereference `bucket.Name`.


dgoodwin:
Bucket names look pretty mandatory so that dereference should be safe. Thanks @Miciah 

/lgtm

PR #53, opened 2018-10-24
Add support for overriding the Hive/Installer image.

dgoodwin:


dgoodwin:
@twiest done in a rush but I think it's working. May need some touchups.

PR #61, opened 2018-10-31
Installer updates

dgoodwin:
Various changes on getting hive to work with a live install config, and cleanup more reliably.


dgoodwin:
@csrwng how is this looking to you?

Many settings in install-config still won't take effect but we're getting closer.

contrib/pkg/installmanager/installmanager.go
```diff
@@ -302,6 +313,26 @@ func (m *InstallManager) uploadAdminKubeconfig() error {
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
```
csrwng: You're already deferring out.Close() above... why call it again?

csrwng: You may want to call out.Sync() instead to make sure it's flushed to disk

contrib/pkg/installmanager/installmanager_test.go
```diff
@@ -141,6 +148,13 @@ func writeFakeBinary(fileName string, contents string) error {
 	return err
 }
 
+func writeFakeInstallConfig(fileName string) error {
+	// nothing needs to read this so for now just an empty file
+	data := []byte("fakefile")
+	err := ioutil.WriteFile(fileName, data, 0755)
```
csrwng: nit: `return ioutil.WriteFile(fileName, data, 0755)`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -136,6 +136,10 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	cdLog.Info("reconciling cluster deployment")
 	cd = cd.DeepCopy()
 
+	if cd.Status.ClusterUUID == "" {
```
csrwng: I'm wondering if ClusterUUID really belongs in Status. Is there ever a case when the end-user wants to specify a ClusterUUID? (to adopt an existing cluster, for example)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -343,7 +319,49 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
+
+	if cd.Status.ClusterUUID != "" {
+		return fmt.Errorf("cluster UUID already set")
+	}
+
+	cd.Status.ClusterUUID = uuid.New()
+	cdLog.WithField("clusterUUID", cd.Status.ClusterUUID).Info("generated new cluster UUID")
+	err := r.Update(context.TODO(), cd)
+	if err != nil {
+		cdLog.Errorf("error updating cluster deployment: %v", err)
```
csrwng: use `WithError`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -343,7 +319,49 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
+	// Delete the install job in case it's still running:
+	installJob := &batchv1.Job{}
+	err := r.Get(context.Background(),
+		types.NamespacedName{
+			Name:      install.GetInstallJobName(cd),
+			Namespace: cd.Namespace,
+		},
+		installJob)
+	if err != nil && errors.IsNotFound(err) {
+		cdLog.Debugf("install job no longer exists, nothing to cleanup")
```
csrwng: s/Debugf/Debug/

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -111,25 +112,27 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
 				testSecret(adminPasswordSecret, adminCredsSecretPasswordKey, "password"),
 				testSecret(pullSecretSecret, pullSecretKey, "{}"),
 				testSecret(sshKeySecret, adminSSHKeySecretKey, "fakesshkey"),
 			},
-			validate: func(c client.Client) {
+			validate: func(c client.Client, t *testing.T) {
 				cd := getCD(c)
 				if cd == nil || !apiequality.Semantic.DeepEqual(cd, testClusterDeployment()) {
+					t.Logf("%v", cd)
+					t.Logf("%v", testClusterDeployment())
```
csrwng: Are these log stmts intentional?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -197,7 +197,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				testSecret(pullSecretSecret, pullSecretKey, "{}"),
 				testSecret(sshKeySecret, adminSSHKeySecretKey, "fakesshkey"),
 			},
-			validate: func(c client.Client) {
+			validate: func(c client.Client, t *testing.T) {
```
csrwng: In this delete case, validate that the job was deleted as well (add one in existing objects)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: If it were moved to spec, is it ok for us to automatically fill that out if unset (which it normally would be?) I guess that would probably be fine.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
dgoodwin: Good catch. I couldn't add here as this just triggers a re-sync but it seems to fit above in the more straightforward delete test.

PR #62, opened 2018-11-01
awstagdeprovision: add support for deleting V2 LBs and LB Target groups

abhinavdahiya:
To delete these load balancers:

* deleting V2 load balancers
```console
aws elbv2 describe-load-balancers
{
    "LoadBalancers": [
        {
            "DNSName": "crawford-aws-int-a85653d6d18c082c.elb.us-east-1.amazonaws.com",
            "VpcId": "vpc-06de9aca2fabbf1b7",
            "State": {
                "Code": "active"
            },
            "Type": "network",
            "AvailabilityZones": [
                {
                    "ZoneName": "us-east-1c",
                    "SubnetId": "subnet-061048a5f17adb00b",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1d",
                    "SubnetId": "subnet-03506a45cf76bbb78",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1a",
                    "SubnetId": "subnet-0d36ad7ab10e8a721",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1f",
                    "SubnetId": "subnet-00f19af286407f964",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1e",
                    "SubnetId": "subnet-0adbb874efe59de3f",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1b",
                    "SubnetId": "subnet-0122ba9d35b720814",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                }
            ],
            "IpAddressType": "ipv4"
        },
        {
            "DNSName": "crawford-aws-ext-e443ac9111e22219.elb.us-east-1.amazonaws.com",
            "CanonicalHostedZoneId": "Z26RNL4JYFTOTI",
            "CreatedTime": "2018-11-01T15:30:22.577Z",
            "LoadBalancerName": "crawford-aws-ext",
            "Scheme": "internet-facing",
            "VpcId": "vpc-06de9aca2fabbf1b7",
            "State": {
                "Code": "active"
            },
            "Type": "network",
            "AvailabilityZones": [
                {
                    "ZoneName": "us-east-1e",
                    "SubnetId": "subnet-0adbb874efe59de3f",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1a",
                    "SubnetId": "subnet-0d36ad7ab10e8a721",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1b",
                    "SubnetId": "subnet-0122ba9d35b720814",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1f",
                    "SubnetId": "subnet-00f19af286407f964",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1d",
                    "SubnetId": "subnet-03506a45cf76bbb78",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                },
                {
                    "ZoneName": "us-east-1c",
                    "SubnetId": "subnet-061048a5f17adb00b",
                    "LoadBalancerAddresses": [
                        {}
                    ]
                }
            ],
            "IpAddressType": "ipv4"
        }
    ]
}

```
These load balancers are tagged, but the api does not allow filters based on tags, so we do that same as
the V1 Lbs, filter on VpcID

```console
aws elbv2 delete-load-balancer --load-balancer-arn
```

* deleting target groups

```console
aws elbv2 describe-target-groups
{
    "TargetGroups": [
        {
            "TargetGroupName": "crawford-aws-api-external",
            "Protocol": "TCP",
            "Port": 6443,
            "VpcId": "vpc-06de9aca2fabbf1b7",
            "HealthCheckProtocol": "TCP",
            "HealthCheckPort": "6443",
            "HealthCheckIntervalSeconds": 10,
            "HealthCheckTimeoutSeconds": 10,
            "HealthyThresholdCount": 3,
            "UnhealthyThresholdCount": 3,
            "LoadBalancerArns": [],
            "TargetType": "ip"
        },
        {
            "TargetGroupName": "crawford-aws-api-internal",
            "Protocol": "TCP",
            "Port": 6443,
            "VpcId": "vpc-06de9aca2fabbf1b7",
            "HealthCheckProtocol": "TCP",
            "HealthCheckPort": "6443",
            "HealthCheckIntervalSeconds": 10,
            "HealthCheckTimeoutSeconds": 10,
            "HealthyThresholdCount": 3,
            "UnhealthyThresholdCount": 3,
            "LoadBalancerArns": [
                "arn:aws:elasticloadbalancing:us-east-1:816138690521:loadbalancer/net/crawford-aws-int/a85653d6d18c082c"
            ],
            "TargetType": "ip"
        },
        {
            "TargetGroupName": "crawford-aws-services",
            "Protocol": "TCP",
            "Port": 49500,
            "VpcId": "vpc-06de9aca2fabbf1b7",
            "HealthCheckProtocol": "TCP",
            "HealthCheckPort": "49500",
            "HealthCheckIntervalSeconds": 10,
            "HealthCheckTimeoutSeconds": 10,
            "HealthyThresholdCount": 3,
            "UnhealthyThresholdCount": 3,
            "LoadBalancerArns": [
                "arn:aws:elasticloadbalancing:us-east-1:816138690521:loadbalancer/net/crawford-aws-int/a85653d6d18c082c"
            ],
            "TargetType": "ip"
        }
    ]
}
```

These target groups can be tagged, but i couldn't find a way to get those tags, and the describe api doesn't allow for
filtering on tags anyways, so we filter on our VpcId

```console
 aws elbv2 delete-target-group --target-group-arn
```

deleting target groups without deleting all the load balancers that have listeners to this target groups returns
```console
An error occurred (ResourceInUse) when calling the DeleteTargetGroup operation: Target group 'REDACTED' is currently in use by a listener or a rule
```

So we call delete target groups after we have completed delete of all v2 LBs.

/cc @crawford @wking @dgoodwin

### DIFF
--- a/contrib/pkg/awstagdeprovision/awstagdeprovision.go
+++ b/contrib/pkg/awstagdeprovision/awstagdeprovision.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/aws/aws-sdk-go/service/s3"
@@ -266,6 +267,106 @@ func deleteLBs(vpc *ec2.Vpc, awsSession *session.Session, logger log.FieldLogger
 	return false
 }
 
+// filterV2LBsByVPC will find all the load balancers in the provided list that are under the provided VPC
+func filterV2LBsByVPC(lbs []*elbv2.LoadBalancer, vpc *ec2.Vpc, logger log.FieldLogger) []*elbv2.LoadBalancer {
+	filteredLBs := []*elbv2.LoadBalancer{}
+
+	for _, lb := range lbs {
+		if *lb.VpcId == *vpc.VpcId {
+			filteredLBs = append(filteredLBs, lb)
+		}
+	}
+
+	return filteredLBs
+}
+
+func deleteV2LBs(vpc *ec2.Vpc, awsSession *session.Session, logger log.FieldLogger) bool {
+	logger.Debugf("Deleting V2 load balancers (%s)", *vpc.VpcId)
+	defer logger.Debugf("Exiting deleting V2 load balancers (%s)", *vpc.VpcId)
+	elbv2Client := elbv2.New(awsSession)
+
+	total := 0
+	filteredLBs := []*elbv2.LoadBalancer{}
+	if err := elbv2Client.DescribeLoadBalancersPages(&elbv2.DescribeLoadBalancersInput{}, func(results *elbv2.DescribeLoadBalancersOutput, lastPage bool) bool {
+		total += len(results.LoadBalancers)
+		filteredLBs = append(filteredLBs, filterV2LBsByVPC(results.LoadBalancers, vpc, logger)...)
+		return lastPage
+	}); err != nil {
+		logger.Errorf("Error listing V2 load balancers: %v", err)
+		return false
+	}
+	logger.Debugf("from %d total V2 load balancers, %d scheduled for deletion", total, len(filteredLBs))
+
+	if len(filteredLBs) == 0 {
+		// no items left to delete
+		// see if we can delete target groups.
+		return deleteTargetGroups(vpc, awsSession, logger)
+	}
+
+	for _, lb := range filteredLBs {
+		logger.Debugf("Deleting V2 load balancer: %v", *lb.LoadBalancerName)
+		_, err := elbv2Client.DeleteLoadBalancer(&elbv2.DeleteLoadBalancerInput{
+			LoadBalancerArn: lb.LoadBalancerArn,
+		})
+		if err != nil {
+			logger.Debugf("Error deleting V2 load balancer %v: %v", *lb.LoadBalancerName, err)
+		} else {
+			logger.WithField("name", *lb.LoadBalancerName).Info("Deleted load balancer")
+		}
+	}
+	// cleanup target groups
+	return deleteTargetGroups(vpc, awsSession, logger)
+}
+
+// filterTargetGroupsByVPC will find all the target groups in the provided list that are under the provided VPC
+func filterTargetGroupsByVPC(tgs []*elbv2.TargetGroup, vpc *ec2.Vpc, logger log.FieldLogger) []*elbv2.TargetGroup {
+	filteredTGs := []*elbv2.TargetGroup{}
+
+	for _, tg := range tgs {
+		if *tg.VpcId == *vpc.VpcId {
+			filteredTGs = append(filteredTGs, tg)
+		}
+	}
+
+	return filteredTGs
+}
+
+func deleteTargetGroups(vpc *ec2.Vpc, awsSession *session.Session, logger log.FieldLogger) bool {
+	logger.Debugf("Deleting target groups (%s)", *vpc.VpcId)
+	defer logger.Debugf("Exiting deleting target groups (%s)", *vpc.VpcId)
+	elbv2Client := elbv2.New(awsSession)
+
+	total := 0
+	filteredTGs := []*elbv2.TargetGroup{}
+	if err := elbv2Client.DescribeTargetGroupsPages(&elbv2.DescribeTargetGroupsInput{}, func(results *elbv2.DescribeTargetGroupsOutput, lastPage bool) bool {
+		total += len(results.TargetGroups)
+		filteredTGs = append(filteredTGs, filterTargetGroupsByVPC(results.TargetGroups, vpc, logger)...)
+		return lastPage
+	}); err != nil {
+		logger.Errorf("Error listing target groups: %v", err)
+		return false
+	}
+	logger.Debugf("from %d total target groups, %d scheduled for deletion", total, len(filteredTGs))
+
+	if len(filteredTGs) == 0 {
+		// no items left to delete
+		return true
+	}
+
+	for _, tg := range filteredTGs {
+		logger.Debugf("Deleting target groups: %v", *tg.TargetGroupName)
+		_, err := elbv2Client.DeleteTargetGroup(&elbv2.DeleteTargetGroupInput{
+			TargetGroupArn: tg.TargetGroupArn,
+		})
+		if err != nil {
+			logger.Debugf("Error deleting target groups %v: %v", *tg.TargetGroupName, err)
+		} else {
+			logger.WithField("name", *tg.TargetGroupName).Info("Deleted target group")
+		}
+	}
+	return false
+}
+
 // rtHasMainAssociation will check whether a given route table has an association marked 'Main'
 func rtHasMainAssociation(rt *ec2.RouteTable) bool {
 	for _, association := range rt.Associations {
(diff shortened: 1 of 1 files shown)


contrib/pkg/awstagdeprovision/awstagdeprovision.go
```diff
@@ -266,6 +267,106 @@ func deleteLBs(vpc *ec2.Vpc, awsSession *session.Session, logger log.FieldLogger
...
+	}); err != nil {
+		logger.Errorf("Error listing V2 load balancers: %v", err)
+		return false
+	}
+	logger.Debugf("from %d total V2 load balancers, %d scheduled for deletion", total, len(filteredLBs))
+
+	if len(filteredLBs) == 0 {
+		// no items left to delete
+		// see if we can delete target groups.
+		return deleteTargetGroups(vpc, awsSession, logger)
```
abhinavdahiya: should we move this to a separate step in deleteVPC after we call delete V2LBs ? or this makes sense here too?

wking:
> These load balancers are tagged, but the api does not allow filters based on tags, so we do that same as
the V1 Lbs, filter on VpcID

This is doubling down on not allowing clusters to get launched into pre-existing, shared VPCs.  But we can always revisit the logic here if we need to return to allowing shared VPCs, so I'm cautiously ok with this approach ;).

contrib/pkg/awstagdeprovision/awstagdeprovision.go
```diff
@@ -266,6 +267,106 @@ func deleteLBs(vpc *ec2.Vpc, awsSession *session.Session, logger log.FieldLogger
...
+
+	for _, lb := range filteredLBs {
+		logger.Debugf("Deleting V2 load balancer: %v", *lb.LoadBalancerName)
+		_, err := elbv2Client.DeleteLoadBalancer(&elbv2.DeleteLoadBalancerInput{
+			LoadBalancerArn: lb.LoadBalancerArn,
+		})
+		if err != nil {
+			logger.Debugf("Error deleting V2 load balancer %v: %v", *lb.LoadBalancerName, err)
+		} else {
+			logger.WithField("name", *lb.LoadBalancerName).Info("Deleted load balancer")
```
wking: You're probably just following an existing pattern, but it's a bit strange to me that we use `WithField` for name here but `: %v` for the name a few lines up in `Deleting V2 load balancer: %v`.  Should we use `WithField` there too?  If so, do we want to do that in this PR for your additions, or punt it to a whole-file-fix later PR?

abhinavdahiya: punt it to a whole-file-fix later PR is what I want :)

contrib/pkg/awstagdeprovision/awstagdeprovision.go
```diff
@@ -354,6 +455,12 @@ func deleteVPCs(awsSession *session.Session, filters AWSFilter, clusterName stri
 				return false, nil
 			}
 
+			complete = deleteV2LBs(vpc, awsSession, logger)
+			if !complete {
+				logger.Debugf("not finished deleting V2 load balancers, will need to retry")
```
wking: This whole block seems inefficient to me.  Can we try all of the children (`deleteLBs`, `deleteV2LBs`, and `deleteRouteTablesWithVPC`) even if one of them errors or is unfinished?  You could do that with something like:

```go
v1LBComplete := deleteLBs(vpc, awsSession, logger)
v2LBComplete := deleteV2LBs(vpc, awsSession, logger)
err := deleteRouteTablesWithVPC(vpc, ec2Client, logger)
if err != nil {
  logger.Debugf("error deleting route tables: %v", err)
  return false, nil
}
if !v1LBComplete {
  logger.Debugf("not finished deleting load balancers, will need to retry")
  return false, nil
}
if !v2LBComplete {
  logger.Debugf("not finished deleting V2 load balancers, will need to retry")
  return false, nil
}
```

That's still serial, but at least we try them all.  You could use goroutines or some such to make those parallel.  I'm also fine punting either my snippet above or parallel calls to follow-up work.

contrib/pkg/awstagdeprovision/awstagdeprovision.go
wking: > should we move this to a separate step in deleteVPC after we call delete V2LBs ?

Leaving it here works better with [this][1].  But as the `deleteVPC` logic stands, I don't think it matters either way.

[1]: https://github.com/openshift/hive/pull/62#discussion_r230140110

contrib/pkg/awstagdeprovision/awstagdeprovision.go
abhinavdahiya: deleteRouteTablesWithVPC will not succeed unless all the consumers of the route tables are it might be waste of api calls.

contrib/pkg/awstagdeprovision/awstagdeprovision.go
abhinavdahiya: keeping it as is keeps the deleteVPC simple, so keeping it as-is

contrib/pkg/awstagdeprovision/awstagdeprovision.go
wking: > `deleteRouteTablesWithVPC` will not succeed unless all the consumers of the route tables are deleted; So it might have been setup to prevent waste of api calls.

Ok.  And we don't expect any clusters with both v1 and v2 load balancers in the same VPC.  So I'm fine with this as you have it, then :).

wking:
Ok, @abhinavdahiya's addressed all of my questions.  Anyone else want to weigh in before I `/lgtm`?

contrib/pkg/awstagdeprovision/awstagdeprovision.go
dgoodwin: If I spun up a cluster with these new LBs and then used Service type LoadBalancer, would I end up with V1 LBs as well?

abhinavdahiya: >  And we don't expect any clusters with both v1 and v2 load balancers in the same VPC
we do expect that there be both v1 and v2 LBs in the same VPC

@wking would there be any problems in that case?

dgoodwin: It looks like it would work fine to me.

dgoodwin:
Looks fine to me, go for it @wking as soon as the V1+V2 thing is resolved. Suspecting not an issue.

contrib/pkg/awstagdeprovision/awstagdeprovision.go
wking: > we do expect that there be both v1 and v2 LBs in the same VPC.

Oh, in that case I'm back in favor of something like [this][1] to get at least v1 and v2 LB deletion attempts (but not route table deletions?) in the same round of attempted-VPC-deletion.  Otherwise we' could go through a few rounds trying to delete the v1 LBs, and then another few rounds trying to delete the v2 LBs, and end up deeper in the exponential backoff making teardown slower than it needs to be.

[1]: https://github.com/openshift/hive/pull/62#discussion_r230140110

abhinavdahiya: @wking with `https://github.com/openshift/hive/pull/62/commits/b4602eda58f2498f1925ef18d020bc4f98a80782` to `https://github.com/openshift/hive/pull/62/commits/004e9e45c2affde0176c3362ac8ffd6c9f348dd0` trying to delete both v1 and v2 and bailing if any one fails/not complete.

PR #64, opened 2018-11-05
cleanup before install

joelddiaz:
in the event the installer/hive container is restarted due to an error, we want to clean up otherwise terraform can get upset about any remnant files from a previous run.

so clean up the files and attempt an uninstall before trying to install a cluster


contrib/pkg/installmanager/installmanager.go
```diff
@@ -184,7 +196,45 @@ func (m *InstallManager) waitForInstallerBinaries() {
...
+	err := uninstaller.Run()
+
+	return err
+}
+
 func (m *InstallManager) runInstaller() error {
+
+	err := m.cleanupBeforeInstall()
+	if err != nil {
+		m.log.Errorf("error while trying to premtively clean up: %v", err)
```
dgoodwin: Nit: spelling. Could do WithError as well.

contrib/pkg/installmanager/installmanager.go
```diff
@@ -184,7 +196,45 @@ func (m *InstallManager) waitForInstallerBinaries() {
...
+	// first cleanup any remnant cache files from a previous installation attempt
+	deleteFiles := []string{
+		"auth",
+		"metadata.json",
+		"terraform.tfstate",
+		"terraform.tfvars",
+		".openshift_install_state.json",
+	}
+	for _, f := range deleteFiles {
+		os.RemoveAll(path.Join([]string{m.WorkDir, f}...))
```
dgoodwin: We're ignoring an error here, I assume because you want to not care if the file doesn't exist? It might be a little safer to check if it exists and if so, then delete it, just in case we hit some weird error that we should actually know about.

contrib/pkg/installmanager/installmanager.go
joelddiaz: fixed

contrib/pkg/installmanager/installmanager.go
joelddiaz: yes, it just returns nil if the file doesn't exist. i changed it to first check for existence, then do the RemoveAll() and check for any error returned there.

dgoodwin:
Closing and moving to https://github.com/openshift/hive/pull/69

PR #74, opened 2018-11-07
Lookup default AMIs.

dgoodwin:
If no default machine pool AMI is set, we will lookup the latest for
RHCOS and store it there for use on any pools that do not explicitly
store their own AMI. This lookup is done with code from
openshift-install.

In future we expect the RHCOS image to use to be a part of the release
image metadata.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -69,7 +69,7 @@ func Add(mgr manager.Manager) error {
 
 // NewReconciler returns a new reconcile.Reconciler
 func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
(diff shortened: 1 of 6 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -184,6 +185,21 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, r.addClusterDeploymentFinalizer(cd)
 	}
 
+	// Ensure we have an AMI set, if not lookup the latest:
```
csrwng: Just a thought ... when we add support for other cloud providers, it may make sense to remove this code from here and simply call a function specific to the cloud provider. Otherwise, we'll likely end up cluttering this reconcile loop.

dgoodwin: Sounds good to me.

PR #76, opened 2018-11-08
Add code generation requirements to API types

jhernand:
The tools that generate Go clients from Kubernetes API expect the
the existence of the `AddToScheme` and `Resource` functions in the API
package, otherwise compilation of the generated code fails. But the Hive
API doesn't define these functions. This patch adds them.


jhernand:
Note that this isn't a problem for Hive itself, as it doesn't use this generated code, but clients may want to generate it. In particular we do use it currently, and to address this issue we have to patch the `register.go` file prior to building our client.

jhernand:
Thanks @dgoodwin!

Issue #73, opened 2018-11-07
Please add API url and web console URL to the cluster status


elad661:
It would be very useful to know the API URL and the web console URL after the cluster is successfully deployed.

PR #78, opened 2018-11-08
Add kubeconfig secret and API/console URLs to cluster deployment status.

dgoodwin:
Install manager will now explicitly link to the secret where it uploaded
the admin kubeconfig after installation, rather than this being an
assumed name.

Also extracts the server URL from the admin kubeconfig, and stores a
link to the API and web console in status.

Includes a short term hack that will let this roll out to any persistent
environments by populating the admin kubeconfig secret name based on the
current assumption. We can remove this shortly.

CC @jhernand 
Fixes: https://github.com/openshift/hive/issues/73

### DIFF
--- a/config/crds/hive_v1alpha1_clusterdeployment.yaml
+++ b/config/crds/hive_v1alpha1_clusterdeployment.yaml
@@ -11,6 +11,8 @@ spec:
     kind: ClusterDeployment
     plural: clusterdeployments
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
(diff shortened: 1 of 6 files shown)


dgoodwin:
Working on live testing, I might need to adjust role perms first so it can write to status.

/hold

jhernand:
Can't say much about the implementation, but the changes in the API are exactly what we need, thanks.

/lgtm

jhernand:
CC: @elad661

dgoodwin:
Ok found and fixed some problems around use of a kubebuilder CRD status subresource.

Should be good now, can you  take a look @csrwng 

/unhold

Issue #82, opened 2018-11-14
Installer Not Completing Successfully When Run in Hive Pods


dgoodwin:
We're getting failed clusters where all masters are left in NotReady state due to cni not being configured when we run the installer in Hive pods specifically. If we run externally even with the same image via podman, the cluster comes up fine.

Suspect this points to a problem in how Hive's install manager is executing the installer.

End of a failed Hive provision:

```
level=info msg="API v1.11.0+d4cacc0 up"                                                                                                                                                                     [0/3239]
level=debug msg="added kube-controller-manager.1566fa9c9ff3b86e: ip-10-0-10-83_b3efcb53-e801-11e8-8d14-0e50234f79ea became leader"                                                                                 
level=debug msg="added kube-scheduler.1566fa9cca7f44e9: ip-10-0-10-83_b3feb2d0-e801-11e8-98c8-0e50234f79ea became leader"
level=warning msg="RetryWatcher - getting event failed! Re-creating the watcher. Last RV: 64"
level=warning msg="Failed to connect events watcher: Get https://dgoodwin1-api.new-installer.openshift.com:6443/api/v1/namespaces/kube-system/events?resourceVersion=64&watch=true: dial tcp 54.234.142.41:6443: con
nect: connection refused"
level=warning msg="Failed to connect events watcher: Get https://dgoodwin1-api.new-installer.openshift.com:6443/api/v1/namespaces/kube-system/events?resourceVersion=64&watch=true: dial tcp 18.214.173.45:6443: con
nect: connection refused"
level=warning msg="Failed to connect events watcher: Get https://dgoodwin1-api.new-installer.openshift.com:6443/api/v1/namespaces/kube-system/events?resourceVersion=64&watch=true: dial tcp 54.234.142.41:6443: con
nect: connection refused"
level=warning msg="Failed to connect events watcher: Get https://dgoodwin1-api.new-installer.openshift.com:6443/api/v1/namespaces/kube-system/events?resourceVersion=64&watch=true: dial tcp 54.211.8.168:6443: conn
ect: connection refused"
level=warning msg="Failed to connect events watcher: Get https://dgoodwin1-api.new-installer.openshift.com:6443/api/v1/namespaces/kube-system/events?resourceVersion=64&watch=true: dial tcp 54.161.146.11:6443: con
nect: connection refused"
[...]

PR #83, opened 2018-11-16
move from flannel to openshiftsdn

joelddiaz:
openshiftsdn is now the default: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/installconfig.go#L82

revendor installer

this fixes the masters coming up not ready


csrwng:
@joelddiaz @dgoodwin is there a way to get the default from the installer code so we don't have to maintain this ourselves?

pkg/install/convertconfig.go
```diff
@@ -100,8 +100,7 @@ func generateInstallConfig(cd *hivev1.ClusterDeployment, adminPassword, sshKey,
 		},
 		BaseDomain: spec.Config.BaseDomain,
 		Networking: types.Networking{
-			// TODO: installer currently only works with flannel, flagged for TODO on their end here: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/installconfig.go#L82
-			Type: "flannel",
+			Type: "OpenshiftSDN",
```
dgoodwin: I hardcoded as this was all they supported at the time, now I think we can remove the hardcode, and instead use what convertNetworkPlugin returns, see the commented out code at the top of this function. Bring that back to life and then we can just pass through whatever the user specified.

dgoodwin:
Nothing yet, shared validation/defaulting was something I had wondered about though.

joelddiaz:
k, revendored the installer to pull in the new sdn types and made the updates as needed

pkg/install/convertconfig.go
```diff
@@ -34,28 +37,30 @@ import (
...
+	networkType, err := convertNetworkingType(spec.Config.Networking.Type)
+	if err != nil {
+		return nil, err
+	}
+
 	platform := types.Platform{}
 	if spec.Config.Platform.AWS != nil {
 		aws := spec.Config.Platform.AWS
-		platform.AWS = &types.AWSPlatform{
+		////platform.AWS = &types.AWSPlatform{
```
dgoodwin: Just some extraneous comments here and below.

dgoodwin:
Fixes: https://github.com/openshift/hive/issues/82

joelddiaz:
for real this time...test cases updated

PR #84, opened 2018-11-16
Add doc describing how to do a service delivery release.

twiest:


twiest:
@dgoodwin Hey, this is my first stab at a release doc for the SD stuff. I tested most of the individual steps, but I wasn't able to test all of them (although I believe they're correct).

If you run into issues, or if some of the instructions aren't clear, please feel free to ping me on slack (I should get notified on my phone).

Thanks!

Issue #88, opened 2018-11-20
Some `hive-controller-manage` messages contain `ERROR: logging before flag.Parse`


jhernand:
Some of the messages in the log of the `hive-controller-manager` pod contain the `ERROR: logging before flag.Parse`, for example:

```
ERROR: logging before flag.Parse: W1120 11:25:50.863560       1 reflector.go:341] github.com/openshift/hive/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: watch of *v1alpha1.DNSZone ended with: The resourceVersion for the provided watch is too old.
```

This is confusing. In the above example it looks like an error, but it is actually a warning. Can this prefix be removed?

dgoodwin:
@csrwng thoughts? 

It looks like an expected periodic thing from Kube, and I don't think we have much control over it due to glog.

jhernand:
The `W1120 ...` error message is fine, what I am asking is if we can remove the `ERROR: logging before flag.Parse` part. I think there is a way to do it.

PR #93, opened 2018-11-20
Remove extra log message warning

csrwng:
Fixes https://github.com/openshift/hive/issues/88


csrwng:
Applies workaround from https://github.com/kubernetes/kubernetes/issues/17162

dgoodwin:
/lgtm

Thanks!

PR #98, opened 2018-11-22
Fix bugs resulting from use of generate name.

dgoodwin:
We had never tested use of generate name, and were inconsistently using
the cluster deployment name, vs it's spec.config.clusterID.

So far this has surfaced in parsing the cluster name out of kubeconfigs,
and in the arguments we pass to uninstall jobs for the kube cluster ID
tag.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -43,22 +43,23 @@ import (
 )
 
 const (
-	testName              = "foo"
-	installJobName        = "foo-install"
-	uninstallJobName      = "foo-uninstall"
+	testName              = "foo-lqmsh"
+	testClusterID         = "bar"
+	installJobName        = "foo-lqmsh-install"
+	uninstallJobName      = "foo-lqmsh-uninstall"
 	testNamespace         = "default"
-	metadataName          = "foo-metadata"
+	metadataName          = "foo-lqmsh-metadata"
 	adminPasswordSecret   = "admin-password"
 	sshKeySecret          = "ssh-key"
 	pullSecretSecret      = "pull-secret"
 	testUUID              = "fakeUUID"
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -309,6 +310,7 @@ func testClusterDeployment() *hivev1.ClusterDeployment {
 		Spec: hivev1.ClusterDeploymentSpec{
 			ClusterUUID: testUUID,
 			Config: hivev1.InstallConfig{
+				ClusterID: testClusterID,
```
cben: Is a clusterID required?
If not, maybe you also want a test without it.
(we happened to use Hive with generateName and clusterID but what if someone used generateName and didn't specify cluaterID?)

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -43,17 +43,18 @@ import (
 )
 
 const (
-	testName              = "foo"
-	installJobName        = "foo-install"
-	uninstallJobName      = "foo-uninstall"
+	testName              = "foo-lqmsh"
+	testClusterID         = "foo"
```
cben: Just for clarity, I'd use here an value unrelated e.g. "bar" or "myclusterid"

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
dgoodwin: We have no validation framework yet, Thomas is working on it, but when we do ClusterID will be mandatory to get the object into etcd. I will make sure we have this on the Jira card as a required validation.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
dgoodwin: Will do.

PR #100, opened 2018-11-22
Add dnszone crd creation to readme

yaacov:
When running Hive from source I needed to create the DNSZone crd before Hive could run.


dgoodwin:
Thanks!

/lgtm

PR #102, opened 2018-11-22
Add a buildah-build target.

dgoodwin:


csrwng:
@dgoodwin with a newer builder like buildah we should be able to go back to a single multi-stage Dockerfile. We can do that later though.

/lgtm

PR #103, opened 2018-11-22
Fix cdLog typos

yaacov:
When I try to run:
```
make run
```

I get errors:
```
# github.com/openshift/hive/pkg/controller/clusterdeployment
pkg/controller/clusterdeployment/clusterdeployment_controller.go:241: Entry.Errorf call has arguments but no formatting directives
pkg/controller/clusterdeployment/clusterdeployment_controller.go:245: Entry.Errorf call has arguments but no formatting directives
make: *** [Makefile:70: vet] Error 2
```

This PR fixes that.


dgoodwin:
Thanks!

/lgtm

PR #120, opened 2018-11-30
store cluster version into clusterdeployment status

joelddiaz:
when we have a kubeconfig for the remote cluster, fetch the ClusterVersion.Status and store it in ClusterDeployment.Status.ClusterVersionStatus

handle nil values for fields in clusterVersion.Status (so we can pass object validation)

update test cases to work with new need to mock up remote ClusterVersion object
    
refactor the building of a remotecluster kubeclient (and update remotemachineset controller accordingly)

build on top of https://github.com/openshift/hive/pull/127 to get beyond the gomock issues

### DIFF
--- a/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	openshiftapiv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -85,6 +86,9 @@ type ClusterDeploymentStatus struct {
 
 	// AdminKubeconfigSecret references the secret containing the admin kubeconfig for this cluster.
 	AdminKubeconfigSecret corev1.LocalObjectReference `json:"adminKubeconfigSecret"`
+
+	// ClusterVersionStatus will hold a copy of the remote cluster's ClusterVersion.Status
+	ClusterVersionStatus openshiftapiv1.ClusterVersionStatus `json:"clusterVersionStatus"`
 
 	// APIURL is the URL where the cluster's API can be accessed.
 	APIURL string `json:"apiURL"`
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -46,7 +46,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	openshiftapiv1 "github.com/openshift/api/config/v1"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/install"
 )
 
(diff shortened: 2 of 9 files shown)


pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -86,6 +86,9 @@ type ClusterDeploymentStatus struct {
 	// AdminKubeconfigSecret references the secret containing the admin kubeconfig for this cluster.
 	AdminKubeconfigSecret corev1.LocalObjectReference `json:"adminKubeconfigSecret"`
 
+	// ClusterVersion locally stores the remote cluster's clusterversion.status.current.version field
+	ClusterVersion string `json:"clusterVersion"`
+
```
dgoodwin: Realistically I think treating this as just a string will probably get us into a spot where we need to break API compatibility in the near future. We should make this a struct, the Update struct also has a Payload which looks really interesting for our purposes as well and we should.

Should we pull the FULL ClusterVersionStatus? Or just make our own and copy over parts of it? Looking through https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L75 we're not really interested in the Generation or VersionHash but the Conditions and the AvailableUpdates could be extremely useful. I'd be tempted to re-use their type and copy the whole thing.

Thoughts?

CC @csrwng in case he has ideas as well.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -613,3 +658,17 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
 	finalizers.Delete(finalizer)
 	object.SetFinalizers(finalizers.List())
 }
+
+func buildRemoteClusterAPIClient(secretData string) (client.Client, error) {
```
dgoodwin: We should probably get this refactored out to a common utility as this is probably the second usage.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -182,8 +188,28 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 			validate: func(c client.Client, t *testing.T) {
 				cd := getCD(c)
-				assert.Equal(t, cd.Status.APIURL, "https://bar-api.clusters.example.com:6443")
-				assert.Equal(t, cd.Status.WebConsoleURL, "https://bar-api.clusters.example.com:6443/console")
+				assert.Equal(t, "https://bar-api.clusters.example.com:6443", cd.Status.APIURL)
+				assert.Equal(t, "https://bar-api.clusters.example.com:6443/console", cd.Status.WebConsoleURL)
```
dgoodwin: Good catch.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -361,6 +374,38 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
+		remoteClusterVersion := &openshiftapiv1.ClusterVersion{}
+		err = remoteClusterAPIClient.Get(context.Background(),
+			types.NamespacedName{Name: clusterVersionObjectName},
+			remoteClusterVersion)
+		if err != nil {
+			cdLog.WithError(err).Error("error fetching remote clusterversion object")
+			return err
+		}
+
+		cdLog.Debugf("remote desired version for cluster %v: %+v", cd.Name, remoteClusterVersion.Status.Current)
```
csrwng: If you want to report the desired version, it should be remoteCusterVersion.Spec.DesiredUpdate.Version

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
csrwng: I agree, I think there's usefulness in embedding the entire status struct

pkg/controller/clusterdeployment/clusterdeployment_controller.go
joelddiaz: yep. i meant to print the current version out of status, not the desired version in spec

joelddiaz:
Now storing the entirety of the clusterversion.Status as suggested.

Have a temporary hack commit to get around needing to allow empty clusterversion.Status.AvailableVersion[]. Waiting for https://github.com/openshift/api/pull/143 (or something like it) to merge, and then can re-vendor openshift/api.

dgoodwin:
Still a problem, looks like a missing omit empty.

PR #122, opened 2018-12-03
remove VPC endpoints

joelddiaz:
installer wants to start using VPC endpoints (https://github.com/openshift/installer/pull/745), so the uninstaller should find and remove them (they are untaggable objects)

also, remove the code that would remove the individual routes from a route table (it's not needed to remove routes to be able to delete RouteTables)

needed to add these permissions to get the installer/uninstaller working:
ec2:DescribePrefixLists
ec2:DescribeVpcEndpoints
ec2:CreateVpcEndpoint
ec2:DeleteVpcEndpoints

### DIFF
--- a/contrib/pkg/awstagdeprovision/awstagdeprovision.go
+++ b/contrib/pkg/awstagdeprovision/awstagdeprovision.go
@@ -377,6 +377,34 @@ func rtHasMainAssociation(rt *ec2.RouteTable) bool {
 	return false
 }
 
+// deleteVPCEndpoints will find all VPC endpoints associated with the passed in VPC and attempt to delete them
+func deleteVPCEndpoints(vpc *ec2.Vpc, ec2Client *ec2.EC2, logger log.FieldLogger) error {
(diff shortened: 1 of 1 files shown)


dgoodwin:
@dak1n1 looks like some new perms for https://github.com/openshift/installer/pull/444

contrib/pkg/awstagdeprovision/awstagdeprovision.go
```diff
@@ -1017,28 +1047,6 @@ func disassociateRouteTable(rt *ec2.RouteTable, ec2Client *ec2.EC2, logger log.F
...
-		})
-		if err != nil {
-			logger.Debugf("error deleting route from route table: %v", err)
-			return err
-		}
-
-		logger.Infof("Deleted route %v from route table %v", *route.DestinationCidrBlock, *rt.RouteTableId)
-	}
-	return nil
-}
```
dgoodwin: Should we keep this code even if it's been replaced in installer so that hiveutil remains able to uninstall clusters even created prior to this? Maybe just document why and keep it for a few months?

Issue #123, opened 2018-12-05
Hive can't load the admin kubeconfig it created for a new cluster.


yaacov:
**Description**:

After creating a new cluster Hive creates a secret with the new clusters kubeconfig.
Hive then tries to load it and fails.

TL;DR: It looks like  a generateName vs name bug.

**What happen**:

In my Hive logs I get:
```
DEBU[9541] reconcile complete                            clusterDeployment=yaacov-05-jnptv job=yaacov-05-jnptv-install namespace=unified-hybrid-cloud
ERRO[9542] unable to load admin kubeconfig               clusterDeployment=yaacov-05-jnptv controller=remotemachineset error="Secret \"yaacov-05-admin-kubeconfig\" not found" namespace=unified-hybrid-cloud
ERRO[9543] unable to load admin kubeconfig               clusterDeployment=yaacov-05-jnptv controller=remotemachineset error="Secret \"yaacov-05-admin-kubeconfig\" not found" namespace=unified-hybrid-cloud
```

When I do `oc get secrets` I see a kubeconfig secret named `yaacov-05-jnptv-admin-kubeconfig`

dgoodwin:
Ideally more info @yaacov  :)

From slack:

`unable to load admin kubeconfig               clusterDeployment=yaacov-05-jnptv controller=remotemachineset error="Secret \"yaacov-05-admin-kubeconfig\" not found"`
and I have a secret name:
`yaacov-05-jnptv-admin-kubeconfig`

This is just another bug caused by the use of generateName, which we should support. Just need to update the remote machine set controller to use the cluster deployment name, not the cluster name, which is where the admin kubeconfig is actually uploaded.

yaacov:
cc @abutcher

yaacov:
> Ideally more info @yaacov :)

:-) sorry :+1:

dgoodwin:
I'll get this one Andrew and tag you for review.

yaacov:
cc @jhernand

PR #124, opened 2018-12-05
Fix GenerateName Bug in Remote Machine Set Controller, and Uninstall Missing Credentials

dgoodwin:
Fixes: https://github.com/openshift/hive/issues/123

Also fixes an uninstall error resulting from my recent rename of the aws cred secret properties.


yaacov:
Thanks all :+1:

PR #126, opened 2018-12-07
replace tectonicClusterID tag with openshiftClusterID

staebler:
The tectonicClusterID tag is being replaced with openshiftClusterID in the installer. This change will use the new tag name.

Hold on openshift/installer#817.

/hold


dgoodwin:
Thanks @staebler 

/lgtm

dgoodwin:
Just merged the fix in #127 

/retest

PR #128, opened 2018-12-10
Add validation webhooks for hive CRDs.

twiest:
- [x] Add Makefile entries for hiveadmission
- [x] Add hiveadmission to hive image
- [x] Add apiserver for validation webhooks
- [x] Add rbac for validation webhooks
- [x] Add kustomize for validation webhook objects
- [x] Add templates for validation webhook objects that require variable substitution

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ test-integration: generate
 
 # Builds all of hive's binaries (including utils).
 .PHONY: build
-build: manager hiveutil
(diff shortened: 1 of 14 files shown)


twiest:
@dgoodwin Thanks for the feedback, it's definitely made this code better. I believe I've addressed all of the comments.

Would you mind reviewing it again please?

PR #144, opened 2018-12-19
contrib/pkg/awstagdeprovision: Replace with installer/pkg/destroy/aws

wking:
I've moved the package over to the installer with openshift/installer#940 to break the hive <-> installer dependency loop.

While bumping `vendor/`, I also added `non-go` and `unused-packages` to the prune block in `Gopkg.toml`.  This avoids pulling in the installer's newly-vendored Terraform source (openshift/installer#919), but it also removes a whole bunch of previously-vendored stuff.  I don't *think* that will cause any problems, but I guess we'll see in CI.


wking:
Hrm.  Now [`unit` is failing][1] with:

```
go generate ./pkg/... ./cmd/...
gofmt -w -s pkg contrib
go vet ./pkg/... ./cmd/... ./contrib/...
# github.com/openshift/hive/pkg/apis/hive/v1alpha1
pkg/apis/hive/v1alpha1/zz_generated.deepcopy.go:87:8: (*in).DeepCopyInto undefined (type *ipnet.IPNet has no field or method DeepCopyInto)
```

I think I need more background on why we have the install-config types under `v1alpha1`.  @dgoodwin, can you give some more background around 88ebffb7b (#2)?  I see [the pull-request topic post][2] mentions converting `IPNet` to strings.  I can go back though and do that here, but is this approach stale now that we're moving towards `config.openshift.io` types (openshift/installer#680)?  What are these Kube bindings for?

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_hive/144/pull-ci-openshift-hive-master-unit/253/build-log.txt
[2]: https://github.com/openshift/hive/pull/2#issue-216307248

dgoodwin:
Looks like you got this solved but for context if you recall way back in the arch calls a few months ago we had multiple discussions on using InstallConfig types in Kube. (Hive is driven by CRDs, the main one being ClusterDeployment which embeds something very similar to, but different than, InstallConfig. We decided on call the best path would be to fork but keep them as similar as possible because we need different things. One of those differences was IPNet, I didn't want to fight with vendoring and getting kube generation going for that so we just use string.

dgoodwin:
This looks good, thanks @wking  we had a card for this and now we can close it. :) I've tested creating a cluster deployment and everything looks good.

CC @joelddiaz no need to move deprovision code to installer now.

/lgtm

PR #145, opened 2018-12-20
Update CRD for changes to VPC API fields.

dgoodwin:


jhernand:
Looks good. I did the same change manually in my environment and it worked correctly. Thanks @dgoodwin.

PR #146, opened 2018-12-20
Update generated crd for clusterdeployment

csrwng:


dgoodwin:
Dupe of https://github.com/openshift/hive/pull/145 which just merged.

PR #153, opened 2019-01-04
Improve process and docs for deploying in cluster.

dgoodwin:
We no longer use the oc cluster up method and thus the use of local hive
images was no longer useful.

Instead we default to latest published images on registry.ci, and a
simple make deploy to install Hive.


abutcher:
Can we remove the admin email/pass from the examples since we set that up afterwards now?

dgoodwin:
We can indeed, removed.

PR #155, opened 2019-01-08
UUID back to status

dgoodwin:
Re-add ClusterID to status so SD has a reliable place to look for it. The value in spec will be going away soon once installer removes ability to specify it, but this can't be done right now.

Fixes breakage from installer changes last night, re-sync our install config format to match theirs, AWS VPCCIDR moves to top level Network MachineCIDR, and PodCIDR is replaced by ClusterNetworks.

cc @jhernand @oourfali these changes will be landing in a future opshive update, and will require API updates. (there will be others as well)

### DIFF
--- a/config/crds/hive_v1alpha1_clusterdeployment.yaml
+++ b/config/crds/hive_v1alpha1_clusterdeployment.yaml
@@ -117,16 +117,29 @@ spec:
                   type: array
                 networking:
                   properties:
-                    podCIDR:
+                    clusterNetworks:
+                      items:
+                        properties:
+                          cidr:
+                            type: string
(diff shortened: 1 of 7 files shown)


oourfali:
@dgoodwin - thanks! can you specify what are the API changes? That way we'll specify all of them in 
 a card we'll create for doing the adjustments needed.

csrwng:
@dgoodwin this looks good to me, are you expecting to make more changes to it?
/retest

dgoodwin:
I just need to get a tiny test added to check it getting set, realized I missed it after opening, will get it ready momentarily.

dgoodwin:
All set now.

Issue #112, opened 2018-11-27
Use registry.svc.ci.openshift.org/openshift/hive-v4.0:hive as the default hive image


dgoodwin:
Presently we're using local image defaults in the code, this should match installer:

```
defaultInstallerImage           = "registry.svc.ci.openshift.org/openshift/origin-v4.0:installer"                                                                                                           
defaultInstallerImagePullPolicy = corev1.PullAlways                                                                                                                                                         
defaultHiveImage                = "hive-controller:latest"                                                                                                                                                  
defaultHiveImagePullPolicy      = corev1.PullNever       
```

Pull policy should be always as well.

PR #157, opened 2019-01-11
Fix the default hive image in code.

dgoodwin:
Now matches the template.


dgoodwin:
Fixes: https://github.com/openshift/hive/issues/112

PR #158, opened 2019-01-11
PreserveOnDelete functionality

ShivamChamoli:
--> Added PreseveOnDelete to pkg/apis/hive/v1alpha1/clusterdeployment_types.go
--> Logic for deleting the cd if PreserveOnDelete is true and skipping the creation of an uninstall job in pkg/controller/clusterdeployment/clusterdeployment_controller.go (commented code left because of the enhancement that finalizers should be deleted)
--> Unit test added to test the functionality


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -165,6 +165,10 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	}
 
 	if cd.DeletionTimestamp != nil {
+		if cd.Spec.PreserveOnDelete {
+			cdLog.Warn("skipping deprovisioning due to PreserveOnDelete")
+			r.Delete(context.TODO(), cd)
```
dgoodwin: We don't want to delete the cluster deployment, if it has a deletion timestamp it's already been deleted. Kubernetes will clean it up when it can. I don't think we need this block rather we can likely handle it in syncDeletedClusterDeployment.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -459,6 +463,22 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
...
+			return reconcile.Result{}, nil
+		} else {
+			cdLog.Warn(cd.Finalizers)
+			err = r.removeClusterDeploymentFinalizer(cd)
+			if err != nil {
+				cdLog.WithError(err).Error("error removing finalizer")
+			}
+			return reconcile.Result{}, err
+		}*/
+		return reconcile.Result{}, nil
```
dgoodwin: Just do the if HasFinalizer { remote it } here. As far as I can tell that should be all we need.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
ShivamChamoli: @dgoodwin : if I don't delete the unit test fails,  it gets the cd (i.e Assert.Nil(cd) causes it to fail, i think kubernetes hasn't cleaned it up yet)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -165,6 +165,9 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	}
 
 	if cd.DeletionTimestamp != nil {
+		if cd.Spec.PreserveOnDelete {
+			cdLog.Warn("skipping deprovisioning due to PreserveOnDelete")
+		}
```
dgoodwin: We're double warning here, lets remove this block and we'll warn below.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -459,6 +462,19 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 		cdLog.WithField("jobName", installJob.Name).Info("install job deleted")
 	}
 
+	// Skips creation of uninstall job if PreserveOnDelete is true
+	if cd.Spec.PreserveOnDelete {
+		cdLog.Warn("skipping creation of uninstall job")
```
dgoodwin: Add "due to PreserveOnDelete"

pkg/controller/clusterdeployment/clusterdeployment_controller.go
ShivamChamoli: done.

dgoodwin:
I attempted to resolve the merge conflict via Github.

/lgtm

dgoodwin:
Replaced by: https://github.com/openshift/hive/pull/173

Needed to fix the gofmt after the merge. :)

PR #159, opened 2019-01-14
Cleanup errors when admin kubeconfig doesn't yet exist.

dgoodwin:
This was erroring when trying to update status, change this to a warning.


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -379,14 +379,43 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
 			return err
 		}
+	} else {
+		cdLog.Infof("cluster deployment status unchanged")
+	}
+
+	return nil
+}
+
+// parseAdminKubeconfigStatus sets all cluster status fields that depend on the admin kubeconfig.
```
joelddiaz: comment mismatch with actual function name

dgoodwin: Oops, thanks, updated!

PR #164, opened 2019-01-15
Separate ClusterDeployment.Machines into ControlPlane and Compute.

abutcher:


dgoodwin:
/hold

We need to update opshive for some critical fixes, don't want this to merge just yet. Will review after if someone doesn't beat me to it.

pkg/install/convertconfig.go
```diff
@@ -66,7 +66,15 @@ func GenerateInstallConfig(cd *hivev1.ClusterDeployment, sshKey, pullSecret stri
...
-	for _, mp := range spec.Machines {
+
+	// combinedMachinePools contains spec.ControlPlane and spec.Compute MachinePools
+	combinedMachinePools := []hivev1.MachinePool{}
+	combinedMachinePools = append(combinedMachinePools, spec.ControlPlane)
+	for _, mp := range spec.Compute {
+		combinedMachinePools = append(combinedMachinePools, mp)
+	}
+
+	for _, mp := range combinedMachinePools {
```
dgoodwin: Can you make sure the master pool goes in with name "master" for install config here. Technically users could set to whatever they want on our side, but we need to force to masters here.

dgoodwin: Would be a good unit test too.

PR #165, opened 2019-01-16
Federation controller

csrwng:
Updates federation installation script
Adds controller to federate installed clusters


csrwng:
@dgoodwin addressed your comments. Removing the WIP from this PR, given the implementation will hopefully change in the not so distant future.

PR #166, opened 2019-01-16
Fix deprovision only working on us-east-1.

dgoodwin:


dgoodwin:
/hold

Need to do functional testing once I fix provisioning.

PR #167, opened 2019-01-16
Upload Cluster ID before provisioning any cloud resources

dgoodwin:


contrib/pkg/installmanager/installmanager.go
```diff
@@ -307,18 +330,40 @@ func runUninstaller(clusterName, region, clusterID string, logger log.FieldLogge
 	}
 
 	err := uninstaller.Run()
+	if err != nil {
```
csrwng: nit: you could just `return uninstaller.Run()`

PR #170, opened 2019-01-17
Fix RBAC rules for federation controller

csrwng:
Fixes RBAC rules for federation


dgoodwin:
That cluster roles and bindings one is a little scary. Why does federation need those for?

csrwng:
@dgoodwin to create a clusterrole for the federation serviceaccount. But should be able to remove it in the next (aka. real) implementation of the controller

PR #171, opened 2019-01-17
Federation controller - Next version

csrwng:
Next implementation of federation controller.

Calls federation code directly to join a cluster instead of calling CLI in a job.
Vendors in federation (with some manual tweaks)


dgoodwin:
Well that's a nice improvement. Still not idempotent I assume though?

/lgtm

csrwng:
Correct, it’s still not idempotent

PR #175, opened 2019-01-21
Bump github.com/openshift/api

csrwng:
Fixes https://jira.coreos.com/browse/CO-187 (ClusterVersion reported empty by Hive)

Due to the mismatch in the types, not all information from the ClusterVersionStatus in the target cluster was making it to the Hive ClusterDeployment resource.


dgoodwin:
Looks good just a test update needed, good find, thanks!

PR #176, opened 2019-01-21
Make ClusterDeployment.spec mostly immutable

twiest:
- [x] Add validation webhook for ClusterDeployment
- [x] Add aggregated api for ClusterDeployment webhook
- [x] Make ClusterDeployment.spec immutable
- [x] Change ClusterDeployment.spec.compute to be mutable
- [x] Ensure that future fields added to ClusterDeployment.spec are immutable by default
- [x] Add unit test showing that ClusterDeployment.spec.compute can be changed during an update
- [x] Add unit test showing that other fields of ClusterDeployment.spec cannot be changed during and update
- [x] Manually test on cluster that immutability is respected.
- [x] Add toleration for hiveadmission daemonset to be able to land on masters.

### DIFF
--- a/cmd/hiveadmission/main.go
+++ b/cmd/hiveadmission/main.go
@@ -29,5 +29,8 @@ func main() {
 	// TODO: figure out a way to combine logrus and glog logging levels. The team has decided that hardcoding this is ok for now.
 	log.SetLevel(log.InfoLevel)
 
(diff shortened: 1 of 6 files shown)


dgoodwin:
Great stuff @twiest thanks.

/lgtm

PR #180, opened 2019-01-22
Consider multiple machinepools within clusterdeployment for syncing.

abutcher:
Consider multiple machinepools in clusterdeployment when syncing machinesets.

### DIFF
--- a/pkg/controller/remotemachineset/remotemachineset_controller_test.go
+++ b/pkg/controller/remotemachineset/remotemachineset_controller_test.go
@@ -284,6 +284,35 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 				}
 			}(),
 		},
+		{
+			name: "Delete additional machinepool machinesets",
+			localExisting: []runtime.Object{
+				testClusterDeployment([]hivev1.MachinePool{
+					testMachinePool("alpha", 3, testName+"-worker", []string{}),
+				}),
+				testSecret(adminKubeconfigSecret, adminKubeconfigSecretKey, testName),
+				testSecret(adminPasswordSecret, adminPasswordSecretKey, testName),
+				testSecret(sshKeySecret, sshKeySecretKey, testName),
+				testSecret(pullSecretSecret, pullSecretSecretKey, testName),
+			},
+			remoteExisting: []runtime.Object{
+				testMachineSet("foo-alpha-us-east-1a", "alpha", true, 1, 0),
+				testMachineSet("foo-alpha-us-east-1b", "alpha", true, 1, 0),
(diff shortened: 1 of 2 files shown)


pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -227,6 +227,100 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+					t.Errorf("did not get expected remote machine sets: %v", err)
+				}
+				if len(rMSL.Items) != 2 {
+					t.Errorf("did not get expected remote machine sets: found %v", len(rMSL.Items))
+				}
+				for _, ms := range rMSL.Items {
+					if *ms.Spec.Replicas != int32(3) {
+						t.Errorf("machineset has unexpected replicas")
+					}
+				}
```
dgoodwin: Wondering if it would be better to replace test.validate with test.expected and list each machine set name, size, and pool expected. We're getting counts here but not necessarily zones. It would shorten up the test cases a little too.

dgoodwin:
Otherwise looks spot on to me.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -259,17 +316,48 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+				rMSL, err := getRMSL(remoteFakeClient)
+				if assert.NoError(t, err) {
+					for _, eMS := range test.expectedRemoteMachineSets.Items {
+						found := false
+						for _, rMS := range rMSL.Items {
+							if eMS.Name == rMS.Name {
+								found = true
+								assert.Equal(t, *eMS.Spec.Replicas, *rMS.Spec.Replicas)
+								assert.Equal(t, eMS.Generation, rMS.Generation)
+								if !reflect.DeepEqual(eMS.ObjectMeta.Labels, rMS.ObjectMeta.Labels) {
```
abutcher: @dgoodwin I've set this up as a `MachineSetList` named `expectedRemoteMachineSets` and then compare replicas, generation, labels and annotations while also looking to see if there any unexpected/missing sets. The annotation/label comparison seems a little funky to me -- let me know if you'd like to see this simplified in any way.

abutcher: Checking labels/annotations is only going to be important if I continue to make sure we don't stomp extra additional ones. Thinking I'll add that back in.

dgoodwin: SGTM

dgoodwin:
Good stuff, test coverage looks good too.

/lgtm

PR #181, opened 2019-01-22
Federation controller improvements

csrwng:
Vendors version of federation with idempotent JoinCluster function
Adds handling of cluster delete by cleaning up federation artifacts.


csrwng:
It looks like federation will only work when creating resources in the expected namespaces (federation-system and cluster registry). This pull needs an update to compose a unique name for the cluster and place it in the expected namespaces.

pkg/controller/federation/clusterdeployment_federation_controller.go
```diff
@@ -172,14 +185,15 @@ func (r *ReconcileClusterDeploymentFederation) Reconcile(request reconcile.Reque
 
 	err = kubefed2.JoinCluster(hostConfig,
 		targetConfig,
-		federationutil.DefaultFederationSystemNamespace,
-		federationutil.MulticlusterPublicNamespace,
+		cd.Namespace,
+		cd.Namespace,
```
dgoodwin: Slightly confused, it sounded from your comments in main PR like we had to start federating them into the central namespaces and couldn't use the cd.Namespace itself?

pkg/controller/federation/clusterdeployment_federation_controller.go
```diff
@@ -211,6 +225,53 @@ func (r *ReconcileClusterDeploymentFederation) loadSecretData(secretName, namesp
 	return string(retStr), nil
 }
 
+func (r *ReconcileClusterDeploymentFederation) syncDeletedClusterDeployment(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (reconcile.Result, error) {
+
+	// Delete the federated cluster
+	federatedCluster := &fedv1alpha1.FederatedCluster{}
+	err := r.Get(context.TODO(), types.NamespacedName{Name: cd.Name, Namespace: cd.Namespace}, federatedCluster)
+	if err != nil && !errors.IsNotFound(err) {
```
dgoodwin: Should log here, otherwise I think it goes silent.

pkg/controller/federation/clusterdeployment_federation_controller.go
```diff
@@ -211,6 +225,53 @@ func (r *ReconcileClusterDeploymentFederation) loadSecretData(secretName, namesp
...
+		return reconcile.Result{}, err
+	}
+	if err == nil {
+		err = r.Delete(context.TODO(), federatedCluster)
+		if err != nil {
```
dgoodwin: Logging a good idea here as well, and for several of these blocks below.

csrwng:
Added logging statements and tested

PR #186, opened 2019-01-25
Remove AMIs from API.

dgoodwin:
This feature was broken after we merged validation that you cannot
change anything outside of spec.Compute. Rather than adding another
exception, this moves the default AMI pinning to an annotation, as the
entire feature is not going to be needed once we implement the design
for OS image lifecycle, so the sooner it's out of the API the better.

CC @jhernand 

I think this will roll out to opshive smoothly. We will lose the AMI that was pinned when that cluster deployment was created. The controller's will realize this and lookup the latest RHCOS image as soon as they sync the CD again. That will then be stored in the annotaton. It's very possible it will be different than the previous image, and when we sync machine sets it would then be set in-cluster. However that would only take effect for new machines,  i.e. if a node was deleted or scaled up, and in either case it should still work. This should mean we do not need to clear all cluster deployments Monday.

### DIFF
--- a/pkg/controller/clusterdeployment/aws.go
+++ b/pkg/controller/clusterdeployment/aws.go
@@ -25,12 +25,20 @@ import (
 	"github.com/openshift/installer/pkg/rhcos"
 )
 
+const (
+	hiveDefaultAMIAnnotation = "hive.openshift.io/default-AMI"
+)
+
 func isDefaultAMISet(cd *hivev1.ClusterDeployment) bool {
-	defaultAMIKnown := false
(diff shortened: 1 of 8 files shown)


csrwng:
@dgoodwin one lint failure

pkg/controller/clusterdeployment/aws.go
```diff
@@ -25,12 +25,20 @@ import (
...
+	defaultAMI, ok := cd.Annotations[hiveDefaultAMIAnnotation]
+	if ok && defaultAMI != "" {
+		return true
 	}
-	return defaultAMIKnown
+	return false
+}
+
+func setDefaultAMI(cd *hivev1.ClusterDeployment, ami string) {
+	cd.Annotations[hiveDefaultAMIAnnotation] = ami
```
dgoodwin: Outside of unit tests it's not getting persisted. :(

PR #191, opened 2019-01-25
Add e2e test that installs hive and creates a cluster

csrwng:
This script is meant to be invoked by an openshift_installer_src job in CI

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+set -e
+
+component=hive
+TEST_IMAGE=$(eval "echo $IMAGE_FORMAT")
+component=installer
+INSTALLER_IMAGE=$(eval "echo $IMAGE_FORMAT")
+
+ln -s $(which oc) $(pwd)/kubectl
+export PATH=$PATH:$(pwd)
+
+# download kustomize so we can use it for deploying
+opsys=linux
+curl -s https://api.github.com/repos/kubernetes-sigs/kustomize/releases/latest |\
+  grep browser_download |\
+  grep $opsys |\
+  cut -d '"' -f 4 |\
+  xargs curl -O -L
+mv kustomize_*_${opsys}_amd64 kustomize
+chmod u+x kustomize
+
+oc new-project cluster-test
+
+# Install Hive
+make deploy DEPLOY_IMAGE="${TEST_IMAGE}"
+
+CLOUD_CREDS_DIR="/tmp/cluster"
+
+# Create a new cluster deployment
+# TODO: Determine which domain to use to create Hive clusters
+export BASE_DOMAIN="hive-ci.openshift.com"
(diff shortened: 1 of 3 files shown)


hack/e2e-test.sh
```diff
@@ -0,0 +1,47 @@
...
+export BASE_DOMAIN="origin-ci-int-aws.dev.rhcloud.com"
+export CLUSTER_NAME="$(oc get cluster.cluster.k8s.io -n openshift-cluster-api -o jsonpath='{ .items[].metadata.name }')-1"
+export SSH_PUB_KEY="$(cat ${CLUSTER_PROFILE_DIR}/ssh-publickey)"
+export PULL_SECRET="$(cat ${CLUSTER_PROFILE_DIR}/pull-secret)"
+export AWS_ACCESS_KEY_ID=$(cat ${AWS_SHARED_CREDENTIALS_FILE} | awk '/aws_access_key_id/ { print $3; exit; }')
+export AWS_SECRET_ACCESS_KEY=$(cat ${AWS_SHARED_CREDENTIALS_FILE} | awk '/aws_secret_access_key/ { print $3; exit; }')
+
+function teardown() {
+	echo "Deleting ClusterDeployment ${CLUSTER_NAME}"
+	oc delete clusterdeployment ${CLUSTER_NAME}
```
dgoodwin: You probably need a waitforjob here as well.

dgoodwin: This is tricky because deprovision will immediately kill the install job, which means CI probably won't have our pod logs saved when it finishes up. Should we add or re-use an annotation to keep the install job around, disassociated with the cluster delpoyment?

hack/e2e-test.sh
```diff
@@ -0,0 +1,47 @@
...
+   AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" \
+   BASE_DOMAIN="${BASE_DOMAIN}" \
+   INSTALLER_IMAGE="${INSTALLER_IMAGE}" \
+   OPENSHIFT_RELEASE_IMAGE="" \
+   TRY_INSTALL_ONCE="true" \
+   | oc apply -f -
+
+# Wait for the cluster deployment to be installed
+SRC_ROOT=$(git rev-parse --show-toplevel)
+go run "${SRC_ROOT}/contrib/cmd/waitforjob/main.go" "${CLUSTER-NAME}-install"
```
dgoodwin: Maybe not now, but it would be nice if we had a go utililty to run a full integration test suite, so fo all of this, spin up the deployment, wait, make sure it's installed, and then run a suite of our own integration tests, check that we got things out of admin kubeconfig, etc. We don't really have that integration angle covered and it would be fairly easy for us to re-use it locally if it assumed you were logged into a cluster already (i.e. our dev clusters)

csrwng:
@dgoodwin removing the WIP for this PR. It should work when we add the hive-ci public zone to the AWS ci account. For now it won't do anything until invoked by prow.

PR #202, opened 2019-01-31
additional safeguard for PreserveOnDelete

ShivamChamoli:
-- Skips creation of uninstall job if the cluster deployment has PreserveOnDelete set to true

### DIFF
--- a/pkg/install/generate_test.go
+++ b/pkg/install/generate_test.go
@@ -1,0 +1,75 @@
+package install
+
+import (
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/install/generate_test.go
```diff
@@ -0,0 +1,93 @@
...
+	tests := []struct {
+		name      string
+		expectErr bool
+		validate  func(*testing.T)
+	}{
+		{
+			name: "Testing PreserveOnDelete",
+			validate: func(t *testing.T) {
+				cd := testClusterDeployment()
+				cd.Spec.PreserveOnDelete = true
```
dgoodwin: Technically you only need these 5 lines for the test, the struct helps set us up for adding more tests easily in future but doing the assertion inside validate here makes me think we should probably wipe out lines 25-32 and 38-47. validate() would be for "post execution" validation whereas in this case it's actually the whole test, so here I think we can just go even simpler. expectErr was also unused.

Otherwise looks great, thanks!

PR #207, opened 2019-02-04
Add MachinePool labels and taints.

abutcher:


dgoodwin:
LGTM so far, nice progress.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -320,6 +331,16 @@ func (r *ReconcileRemoteMachineSet) generateMachineSetsFromClusterDeployment(cd
 				return nil, err
 			}
 			for _, ms := range icMachineSets {
+				// Apply hive MachinePool labels to MachineSet MachineSpec.
+				ms.Spec.Template.Spec.ObjectMeta.Labels = map[string]string{}
```
abutcher: This is causing update churn since `ObjectMeta.Labels` will remain unset in remote objects when this is empty. Re-adding WIP to fix @dgoodwin

abutcher: Wrapped in a conditional based on the `hivePool.Labels` having been set.

dgoodwin:
/lgtm

Thanks @abutcher

dgoodwin:
/lgtm

Tests won't pass until we get #228 through.

PR #211, opened 2019-02-06
Change hiveadmission to use apiservice annotation for cabundle

twiest:
- [x] Change hiveadmission apiservice cabundle generation from makefile to annotation.
- [x] Change hiveadmission ValidatingWebhookConfiguration client caBundle to default to null.
- [x] Test that hiveadmission deploy works
- [x] Test that hiveadmission correctly validates a ClusterDeployment


twiest:
@dgoodwin hey, I updated this PR to also remove KUBE_CA. I tested it locally with a brand new cluster / hive ClusterDeployment and everything seemed validate or fail to validate correctly.

Would you mind re-reviewing this please?

dgoodwin:
omg nice. I don't even know if we need the operator now.

/lgtm

twiest:
@dgoodwin you're going to kill me, but I fixed more stuff in this PR.

In the demo meeting today, someone demoed something that made me think of the hiveadmission OpenShift template and I realized that nothing in that file needs substitutions anymore, which is why I used an OpenShift template instead of just putting them in Kustomize.

This means that we can put all of those objects into Kustomize and get rid of the hiveadmission OpenShift template.

I did that and tested it out on a new cluster and it all worked as expected.

I swear I'm done messing with this now. :)

Please review one last time.

dgoodwin:
At this point, everything you're moving is already moved in the operator branch I'm going to work on. Each component has to be in it's own yaml file and then they go to go-bindata assets and get applied by the operator, so I'm thinking there's not much point to merging this and fighting the conflicts right now, if it's alright I'd like to propose we close and I'll incorporate any changes in the new files (the annotation is already done) and then get you on the PR to make sure I got everything.

Sound ok?

twiest:
@dgoodwin yep, sounds good to me.  Closing.

PR #212, opened 2019-02-06
Federation controller: handle label sync

csrwng:
Adds field to ClusterDeployment status to keep track of corresponding FederatedCluster resource. Synchronizes labels from ClusterDeployment to FederatedCluster resource.
Adds unit tests to federation controller.

### DIFF
--- a/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
@@ -124,7 +124,7 @@ type ClusterDeploymentStatus struct {
 	Federated bool `json:"federated,omitempty"`
 
 	// FederatedClusterRef is the reference to the federated cluster resource associated with
-	// this ClusterDeployment
+	// this ClusterDeployment.
 	FederatedClusterRef *corev1.ObjectReference `json:"federatedClusterRef,omitempty"`
 
 	// AdminKubeconfigSecret references the secret containing the admin kubeconfig for this cluster.
--- a/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
@@ -123,6 +123,10 @@ type ClusterDeploymentStatus struct {
 	// Federated is true if the cluster deployment has been federated with the host cluster.
 	Federated bool `json:"federated,omitempty"`
 
+	// FederatedClusterRef is the reference to the federated cluster resource associated with
+	// this ClusterDeployment
+	FederatedClusterRef *corev1.ObjectReference `json:"federatedClusterRef,omitempty"`
+
 	// AdminKubeconfigSecret references the secret containing the admin kubeconfig for this cluster.
 	AdminKubeconfigSecret corev1.LocalObjectReference `json:"adminKubeconfigSecret,omitempty"`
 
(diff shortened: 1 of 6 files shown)


pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -123,6 +123,10 @@ type ClusterDeploymentStatus struct {
 	// Federated is true if the cluster deployment has been federated with the host cluster.
 	Federated bool `json:"federated,omitempty"`
 
+	// FederatedClusterRef is the reference to the federated cluster resource associated with
+	// this ClusterDeployment
```
dgoodwin: Nit: ending with a '.' looks consistent with the fields around it.

pkg/controller/federation/clusterdeployment_federation_controller.go
```diff
@@ -56,6 +56,9 @@ const (
 	adminKubeconfigKey = "kubeconfig"
 
 	federatedClustersCRDName = "federatedclusters.core.federation.k8s.io"
+
+	clusterDeploymentNamespaceAnnotation = "hive.openshift.io/cluster-deployment-namespace"
+	clusterDeploymentNameAnnotation      = "hive.openshift.io/cluster-deployment-name"
```
dgoodwin: In other areas i've done namespace/name in an annotation, as it seems a little less cluster on the object but you then have to parse. Do you think splitting into two is a better approach?

dgoodwin: I guess parsing vs checking if two annotations are set.

pkg/controller/federation/clusterdeployment_federation_controller.go
```diff
@@ -159,60 +161,71 @@ func (r *ReconcileClusterDeploymentFederation) Reconcile(request reconcile.Reque
...
+		}
 		return reconcile.Result{}, nil
 	}
 
 	cdLog.Info("reconciling cluster deployment for federation")
 
+	if cd.Status.FederatedClusterRef == nil || len(cd.Status.FederatedClusterRef.Name) == 0 {
+		cdLog.Debugf("setting federated cluster name reference")
+		return reconcile.Result{}, r.setFederatedClusterRef(cd, cdLog)
+	}
```
dgoodwin: Looks like we generate and stamp a cluster ref before we even try to federate, so even if we fail it will re-try with the previous ref? Good thinking.

pkg/controller/federation/clusterdeployment_federation_controller.go
```diff
@@ -159,60 +161,71 @@ func (r *ReconcileClusterDeploymentFederation) Reconcile(request reconcile.Reque
...
+
+	r.joinCluster(cd, cdLog)
```
dgoodwin: is there an err check missing here?

dgoodwin: and potentially in unit tests if so.

csrwng:
Thanks for the review @dgoodwin! I've submitted another commit with fixes

pkg/controller/federation/clusterdeployment_federation_controller.go
csrwng: Makes sense, I switched to one.

pkg/controller/federation/clusterdeployment_federation_controller.go
csrwng: Yup, otherwise retrying join would make no sense.

pkg/controller/federation/clusterdeployment_federation_controller.go
csrwng: Yes, thank you for catching that

PR #213, opened 2019-02-07
Moved config, latest kustomizer requires files to be in same or child directory

jewzaam:


jewzaam:
See https://github.com/openshift/hive/pull/215

PR #217, opened 2019-02-08
Cleanup use of mismatched images and clarify docs.

dgoodwin:
Our template was defaulting to using latest installer/hive, but an older
pinned release image which is not compatible.


dgoodwin:
/hold 

temporarily using the older images for CI, will unhold once that installer bug is fixed.

dgoodwin:
@twiest can you take another look.

csrwng:
Vpc limit exceeded. Hopefully it's temporary
/retest

PR #218, opened 2019-02-08
automatically-updating-install-job

ShivamChamoli:
--> Automatically updates install job when the generation of the cd does not match the cd generation in job's annotations.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strconv"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -57,8 +58,7 @@ const (
 	roleBindingName    = "cluster-installer"
 
 	// deleteAfterAnnotation is the annotation that contains a duration after which the cluster should be cleaned up.
-	deleteAfterAnnotation = "hive.openshift.io/delete-after"
-
+	deleteAfterAnnotation       = "hive.openshift.io/delete-after"
 	adminCredsSecretPasswordKey = "password"
 	pullSecretKey               = ".dockercfg"
 	adminSSHKeySecretKey        = "ssh-publickey"
@@ -70,7 +70,8 @@ const (
 	// HiveImageEnvVar is the optional environment variable that overrides the image to use
 	// for provisioning/deprovisioning. Typically this originates from the HiveConfig and is
 	// set as an EnvVar on the deployment.
-	HiveImageEnvVar = "HIVE_IMAGE"
+	HiveImageEnvVar                       = "HIVE_IMAGE"
+	clusterDeploymentGenerationAnnotation = "hive.openshift.io/cluster-deployment-generation"
 )
 
 // Add creates a new ClusterDeployment Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -320,7 +321,17 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		cdLog.Errorf("error getting job: %v", err)
 		return reconcile.Result{}, err
 	} else {
-		cdLog.Infof("cluster job exists, successful: %v", cd.Status.Installed)
+		cdLog.Infof("cluster job exists, cluster deployment status: %v", cd.Status.Installed)
+		if !cd.Status.Installed {
+			if existingJob.Annotations != nil && cfgMap.Annotations != nil {
+				didGenerationChange, err := r.updateOutdatedConfigurations(cd.Generation, existingJob, cfgMap, cdLog)
+				if err != nil {
+					return reconcile.Result{}, err
+				} else if didGenerationChange {
+					return reconcile.Result{Requeue: true}, nil
+				}
+			}
+		}
 	}
 
 	err = r.updateClusterDeploymentStatus(cd, origCD, existingJob, cdLog)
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -308,7 +310,24 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
 		return reconcile.Result{}, err
 	} else {
+		// If install job exists but the cd generation is different, recreate the job after deleting it
+		// provided the cd is not installed
 		cdLog.Infof("cluster job exists, successful: %v", cd.Status.Installed)
+		cdLog.Infof("cluster job exists, generation: %v", job.Annotations[clusterDeploymentGeneration])
+		if !cd.Status.Installed {
+			cdGeneration := strconv.FormatInt(cd.Generation, 10)
+			cdLog.Infof("cluster generation: %v", cdGeneration)
+			if existingJob.Annotations[clusterDeploymentGeneration] != cdGeneration {
```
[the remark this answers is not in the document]
dgoodwin: Lets get some unit tests added here, and include one that covers this case above where the job exists but has *no* cluster deployment generation set in it's annotations.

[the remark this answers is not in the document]
ShivamChamoli: So if you look at generate.go, where we generate our installer job, I have written code for always adding a generation to the annotation when we generate an installer job. There I also have a check so that if Annotations are null I insert a map of strings (key, value both strings)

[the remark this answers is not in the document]
dgoodwin: That is if you generate a new job, but we're talking about an environment with pre-existing jobs. AFAIK if the job already exists, it won't be regenerated and won't pickup the new annotations. Am I mistaken on this?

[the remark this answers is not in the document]
ShivamChamoli: Ahhh, I get it now. Perfect, makes sense.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -308,7 +310,24 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
+				err = r.Delete(context.TODO(), existingJob, client.PropagationPolicy(metav1.DeletePropagationForeground))
```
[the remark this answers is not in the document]
ShivamChamoli: Do you mean no separate annotation for configmap on the install job? We do need the generation annotation right?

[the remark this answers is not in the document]
dgoodwin: No separate annotation on the configmap, just do what you're doing but assume that if you need to update the job, you need to update the install configmap as well. Effectively treating them as a pair in lockstep.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -65,6 +66,7 @@ const (
 	adminKubeconfigKey          = "kubeconfig"
 	clusterVersionObjectName    = "version"
 	clusterVersionUnknown       = "undef"
+	clusterDeploymentGeneration = "hive.openshift.io/cluster-deployment-generation"
```
csrwng: rename the variable to indicate that it's used as an annotation (ie. `clusterDeploymentGenerationAnnotation`)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -308,7 +310,34 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		cdLog.Errorf("error getting job: %v", err)
 		return reconcile.Result{}, err
 	} else {
+		// If install job exists but the cd generation is different, recreate the job after deleting it
+		// provided the cd is not installed
 		cdLog.Infof("cluster job exists, successful: %v", cd.Status.Installed)
+		if !cd.Status.Installed {
+			cdGeneration := strconv.FormatInt(cd.Generation, 10)
```
csrwng: `strconv.ItoA(cd.Generation)`

csrwng: Also, extract this block into its own function and remove else's

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -308,7 +310,34 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
+			cdLog.Infof("cluster generation: %v", cdGeneration)
+			if existingJob.Annotations != nil {
+				if generation, ok := existingJob.Annotations[clusterDeploymentGeneration]; ok {
+					if generation != cdGeneration {
```
csrwng: instead of doing a string compare, do a number compare and check that the current generation is greater than the one that was used to create the job.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -308,7 +310,34 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
+						err = r.Delete(context.TODO(), existingJob, client.PropagationPolicy(metav1.DeletePropagationForeground))
+						if err != nil {
+							cdLog.WithError(err).Errorf("error deleting outdated install job")
+							return reconcile.Result{}, err
+						} else {
+							cdLog.Infof("successfully deleted outdated install job")
+							err = r.Delete(context.TODO(), cfgMap, client.PropagationPolicy(metav1.DeletePropagationForeground))
```
csrwng: We should handle the configmap separately in the block where we check if it exists. If it already exists, we should update it to contain the values we expect.

dgoodwin: This one was my fault, I suggested that we handle the job and the install config map in lockstep, if we determine we need to delete the job, we know we should probably delete the configmap, and let both be regenerated on the re-sync.

csrwng: it's probably a good idea to always update the configmap if it exists, in case it's changed by other means.

csrwng: hmm, as I finished typing that, I realized that could be bad :) If we already have a job that's based on a current generation, then there's really no need to overwrite the configmap. We could put the generation annotation on the configmap as well though.

ShivamChamoli: Should I go ahead with the annotation on configmap?

dgoodwin:
Needs a rebase on master and to remove the WIP designation.

PR #223, opened 2019-02-14
Add retry logic to e2e-test.sh

twiest:


hack/e2e-test.sh
```diff
@@ -42,19 +95,42 @@ trap 'teardown' EXIT
...
+         CLUSTER_NAME="${CLUSTER_NAME}" \
+         SSH_KEY="${SSH_PUB_KEY}" \
+         PULL_SECRET="${PULL_SECRET}" \
+         AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}" \
+         AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" \
+         BASE_DOMAIN="${BASE_DOMAIN}" \
+         INSTALLER_IMAGE="${INSTALLER_IMAGE}" \
+         OPENSHIFT_RELEASE_IMAGE="" \
+         TRY_INSTALL_ONCE="true" \
+         | oc apply -f - ; then
```
dgoodwin: What can cause this to fail intermittently?

twiest: If the API server is down when the `oc apply` is run. For instance, if the certs are updated and the api server pods are rolled (they're talking about having really short cert periods for CI to test the rolling).

Or if above where `CLUSTER_NAME` is gathered, if that `oc` call fails (although now it'll retry and ultimately error out). 

This is the actual line that my [e2e was failing](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_hive/219/pull-ci-openshift-hive-master-e2e/59) on and then @csrwng pointed out that we should add retries to the other `oc` commands as well.

dgoodwin: Sounds good thanks.

csrwng:
looks like the regular install never finished
/test e2e

PR #225, opened 2019-02-19
Support legacy federated cluster names

csrwng:


csrwng:
An actual successful merge. Yay!

PR #222, opened 2019-02-14
*: Bump to install-config v0.12.0

wking:
Catching up with openshift/installer@dafc79f0 (openshift/installer#1013) and openshift/installer@3b393da8 (openshift/installer#1154).

The uint32 -> int32 cast is slightly dangerous, because it will silently wrap overflowing values (golang/go#19624, golang/go#30209).  But I'll try and get the installer updated to use unsigned types as well, and then we won't have to worry about converting.

PR #226, opened 2019-02-19
WIP: Fix e2e

csrwng:
Combines commits from https://github.com/openshift/hive/pull/222 and https://github.com/openshift/hive/pull/223 to see if together we get passing e2e again


dgoodwin:
e2e is green so I am merging this right f*#&king now.

/lgtm

PR #229, opened 2019-02-21
Match OpenshiftSDN -> OpenShiftSDN network type change.

dgoodwin:
Related to https://github.com/openshift/release/pull/2922 and maybe the cause of the cluster CI failures all day today.

I'm not strictly sure if this required to get through CI or not.

CC @jhernand this will require a change on the UHC side once this rolls out.

### DIFF
--- a/config/templates/cluster-deployment.yaml
+++ b/config/templates/cluster-deployment.yaml
@@ -102,7 +102,7 @@ objects:
     clusterName: ${CLUSTER_NAME}
     baseDomain: ${BASE_DOMAIN}
     networking:
-      type: OpenshiftSDN
+      type: OpenShiftSDN
       serviceCIDR: "172.30.0.0/16"
       machineCIDR: "10.0.0.0/16"
       clusterNetworks:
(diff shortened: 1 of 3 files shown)


jhernand:
@dgoodwin Does this change affect the Hive controller, or only the install/uinstall pods? If it affects the controller then we will need to update UHC and Hive simultaneously.

I don't see any change in the validation configuration of the CRD, so I assume that we don't need to change the existing cluster deployments to use the new value. Or do we? Will the value of existing cluster deployments be used once the installation is complete?

dgoodwin:
The change affects the install pods, UHC will probably need to update as soon as this rolls out. 

I don't *think* it will affect pre-existing clusters who have already had their install jobs run.

dgoodwin:
I think this is required to get CI back, I've confirmed that installconfigs we generate are coming out with the wrong case and thus will not have networking.

dgoodwin:
/override ci/prow/e2e

We need at least two critical fixes before CI will pass, this is one.

PR #228, opened 2019-02-21
Bump to installer master branch

csrwng:
Includes fix for uninstall from installer master repo
Fixes up remotemachineset controller to work with machine-api instead of cluster-api

PR #230, opened 2019-02-22
Short name and custom printer columns for the ClusterDeployment CRD

fabianofranz:
Adds the `cd` short name to the `ClusterDeployment` CRD so that we can do `kubectl get cd`.
Adds `baseDomain` and `status.installed` as custom printer columns so that the printer output looks like

```
$ kubectl get cd
NAME     BASEDOMAIN                                INSTALLED   AGE
ffranz   ffranz-cluster-deployment.openshift.com   false       4m7s
```

@csrwng @dgoodwin makes sense?


csrwng:
@ffranz it won’t pass e2e until #228 merges. I was hoping it would have merged by now. But for whatever reason it just says “in merge pool”

csrwng:
@fabianofranz looks like you need to rebase

PR #234, opened 2019-02-27
Remove limits from pods

fabianofranz:
Removes CPU limits from pods, soon to be checked by the test in https://github.com/openshift/origin/pull/22095.

Re https://jira.coreos.com/browse/CO-262

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -58,9 +58,6 @@ spec:
         imagePullPolicy: Always
         name: manager
         resources:
-          limits:
-            cpu: 100m
-            memory: 256Mi
           requests:
             cpu: 100m
             memory: 75Mi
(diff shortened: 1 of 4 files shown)


dgoodwin:
Is this a good idea for Hive as well? I guess unconstrained CPU is probably fine. What about memory, should we keep those limits?

CC @csrwng

fabianofranz:
@dgoodwin waiting for Derek's comment on the other PR, regarding `memory` and `limits`. The test seems to check every pod with no exception, that's why I initially did it in hive too.

csrwng:
@dgoodwin @fabianofranz the test is for pods running on masters. Afaik we only schedule our webhook pods on masters. It'd be good to understand the reasoning for removing limits on masters to determine if it should apply to other pods as well.

PR #236, opened 2019-02-27
Add syncset types.

abutcher:
SyncSet crd.

### DIFF
--- a/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
@@ -144,6 +144,14 @@ type ClusterDeploymentStatus struct {
 
 	// WebConsoleURL is the URL for the cluster's web console UI.
 	WebConsoleURL string `json:"webConsoleURL,omitempty"`
+
+	// SyncResources is the list of SyncStatus for objects that have been synced.
+	// +optional
+	SyncResources []SyncStatus `json:"syncResources,omitempty"`
+
+	// SyncPatches is the list of SyncStatus for patches that have been applied.
+	// +optional
+	SyncPatches []SyncStatus `json:"syncPatches,omitempty"`
 }
 
 // +genclient
--- a/pkg/apis/hive/v1alpha1/syncset_types.go
+++ b/pkg/apis/hive/v1alpha1/syncset_types.go
@@ -1,0 +1,235 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// SyncSetResourceApplyMode is a string representing the mode with which to
+// apply SyncSet Resources.
+type SyncSetResourceApplyMode string
+
+const (
+	// UpsertResourceApplyMode indicates that objects will be updated
+	// or inserted (created).
+	UpsertResourceApplyMode SyncSetResourceApplyMode = "Upsert"
+
+	// SyncResourceApplyMode inherits the create or update functionality
+	// of Upsert but also indicates that objects will be deleted if created
+	// previously and detected missing from defined Resources in the SyncSet.
+	SyncResourceApplyMode SyncSetResourceApplyMode = "Sync"
+)
+
+// SyncSetPatchApplyMode is a string representing the mode with which to apply
+// SyncSet Patches.
+type SyncSetPatchApplyMode string
+
+const (
+	// ApplyOncePatchApplyMode indicates that the patch should be applied
+	// only once.
+	ApplyOncePatchApplyMode SyncSetPatchApplyMode = "ApplyOnce"
+
+	// AlwaysApplyPatchApplyMode indicates that the patch should be
+	// continuously applied.
+	AlwaysApplyPatchApplyMode SyncSetPatchApplyMode = "AlwaysApply"
+)
+
+// SyncObjectPatch represents a patch to be applied to a specific object
+type SyncObjectPatch struct {
+	// GroupVersionKind is the Group, Version and Kind of the object to be patched.
+	GroupVersionKind schema.GroupVersionKind `json:"groupVersionKind"`
+
+	// Name is the name of the object to be patched.
+	Name string `json:"name"`
+
+	// Namespace is the Namespace in which the object to patch exists.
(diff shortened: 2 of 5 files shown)


pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,132 @@
...
+	// ApplyMode "upsert" indicates update or insert (create).
+	// ApplyMode "sync" allows for object deletion.
+	// +optional
+	ApplyMode SyncSetApplyMode `json:"applyMode,omitempty"`
+
+	// Selector indicates which clusters to maintain objects within
+	// by label. This Selector will apply to clusters in any namespace.
+	// TODO: Expand this selector to include clusters []string which
+	// will apply to clusters by name within the SyncSet's namespace.
+	Selector metav1.LabelSelector `json:"selector"`
```
csrwng: More like:
https://github.com/kubernetes-sigs/federation-v2/blob/087903a9fb1b3077b50112ac98a7c2364e59868b/pkg/controller/util/placement.go#L27-L30

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Namespace is the Namespace in which the object to patch exists.
+	// Defaults to the SyncSet's Namespace.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// Patch is the patch to apply.
+	Patch []byte `json:"patch"`
+
+	// PatchType indicates the PatchType as JSONPatchType (default),
```
csrwng: In the comments you should specify the actual string values and not the name of the go constants.

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,132 @@
...
+	AppliedResources []SyncDescriptor `json:"appliedResources,omitempty"`
+
+	// AppliedPatches is the list of SyncDescriptors for patches that have
+	// been applied.
+	// +optional
+	AppliedPatches []SyncDescriptor `json:"appliedPatches,omitempty"`
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
csrwng: add 
```
// +kubebuilder:subresource:status

```

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,132 @@
...
+	PatchType types.PatchType `json:"patchType,omitempty"`
+}
+
+// SyncDescriptor describes objects that have been created or patches that
+// have been applied using the unique md5 sum of the object or patch.
+type SyncDescriptor struct {
+	GroupVersionKind schema.GroupVersionKind `json:"groupVersionKind"`
+	Name             string                  `json:"name"`
+	Namespace        string                  `json:"namespace"`
+	Hash             []byte                  `json:"hash"`
```
dgoodwin: Will the hash be helpful now? I'd be ok with skipping it if we don't think we need it for anything just yet. I don't think we can use it to determine if something needs to be done or not because we need to check remote state, and the hash there won't match due to many things like creationTimestamp.

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,213 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// Upsert SyncSetResourceApplyMode indicates that objects will be updated
+	// or inserted (created).
+	Upsert SyncSetResourceApplyMode = "Upsert"
```
csrwng: rename the variable to indicate what it is
s/Upsert/UpsertResourceApplyMode/
otherwise, in code that uses this API, it won't be clear what hivev1.Upsert is for.
(Same as the other values)

I would put the set of values for each type in their own const (...) block for improved readability

Use only the name of the variable and not the type in the comment (s| // Upsert SyncSetResourceApplyMode | // UpsertResourceApplyMode| )

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,213 @@
...
+type SyncSetResourceApplyMode string
+
+// SyncSetPatchApplyMode is a string representing the mode with which to apply
+// SyncSet Patches.
+type SyncSetPatchApplyMode string
+
+// SyncObjectPatch represents a patch to be applied to a specific object
+type SyncObjectPatch struct {
+	GroupVersionKind schema.GroupVersionKind `json:"groupVersionKind"`
+	Name             string                  `json:"name"`
```
csrwng: Add comments for Name and GroupVersionKind

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,213 @@
...
+	Reason string `json:"reason,omitempty"`
+	// Message is a human-readable message indicating details about last transition.
+	// +optional
+	Message string `json:"message,omitempty"`
+}
+
+// SyncStatus describes objects that have been created or patches that
+// have been applied using the unique md5 sum of the object or patch.
+type SyncStatus struct {
+	GroupVersionKind schema.GroupVersionKind `json:"groupVersionKind"`
```
csrwng: Add comments to the fields

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,213 @@
...
+	// +optional
+	Patches []SyncObjectPatch `json:"patches,omitempty"`
+}
+
+// SelectorSyncSetSpec defines the SyncSetCommonSpec resources and patches to sync along
+// with a ClusterDeploymentSelector indicating which clusters the SelectorSyncSet applies
+// to in any namespace.
+type SelectorSyncSetSpec struct {
+	SyncSetCommonSpec         `json:",inline"`
+	ClusterDeploymentSelector metav1.LabelSelector `json:"clusterDeploymentSelector,omitempty"`
```
csrwng: If using the ` //+optional ` comment for all optional values, use it consistently
Add field comments

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -0,0 +1,213 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// SyncSet is the Schema for the SyncSet API
+// +k8s:openapi-gen=true
+type SyncSet struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   SyncSetSpec   `json:"spec,omitempty"`
```
csrwng: Add field comments

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -144,6 +144,13 @@ type ClusterDeploymentStatus struct {
 
 	// WebConsoleURL is the URL for the cluster's web console UI.
 	WebConsoleURL string `json:"webConsoleURL,omitempty"`
+
+	// SyncResources is the list of SyncStatus for objects that have been synced.
+	// +optional
+	SyncResources []SyncStatus `json:"syncResources,omitempty"`
+
+	// SyncPatches is the list of SyncStatus for patches that have been applied.
```
csrwng: Use // +optional consistently

csrwng:
@abutcher I would remove the unit test. Other than that, it looks great.

csrwng:
Also, rebase and make sure you run `make crd` to ensure the generated CRD is updated.

PR #238, opened 2019-03-01
Updates for OLM packaging and deployment issues.

dgoodwin:
This is a first draft of the CSV for Hive to be delivered as an OLM package/bundle.

This can successfully be published to a Quay application repo with `operator-courier push config/operator-bundle dgoodwin hive-operator 0.0.5 $TOKEN`

Everything here is currently copied manually. The CRDs will be easy to auto-update, the CSV much more tricky as parts of it are embedded objects, but not the entire Kube Deployment/Role/RoleBinding we already have yaml for... messy.


config/operator-bundle/hive-clusterserviceversion.yaml
```diff
@@ -0,0 +1,186 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: hive-operator-0.0.1
```
dgoodwin: @jmelis any idea if we should use a static name here?

config/operator-bundle/hive-clusterserviceversion.yaml
```diff
@@ -0,0 +1,186 @@
...
+      OpenShift cluster provisioning and management at scale.
+    containerImage: quay.io/dgoodwin/hive:latest
+    createdAt: 2019-02-25T14:29:00Z
+    support: Devan Goodwin
+spec:
+  displayName: Hive
+  description: |
+    OpenShift cluster provisioning and management at scale.
+  keywords: ['kubernetes', 'openshift', 'multi-cluster', 'cluster']
+  version: 0.0.1
```
dgoodwin: @jmelis it looks like it's valid here to do 0.0.1-HASH here. What would be easiest for your pipeline in this regard?

Maybe a make csv target that builds this file up with the version = 0.0.1-HASH, replaces the container images with a hash tag as well? Your pileline can run that, then publish the resulting bundle with operator-courier?

config/operator-bundle/hive-clusterserviceversion.yaml
```diff
@@ -0,0 +1,186 @@
...
+          template:
+            metadata:
+              labels:
+                control-plane: hive-operator
+                controller-tools.k8s.io: "1.0"
+            spec:
+              serviceAccountName: hive-operator
+              containers:
+              # By default we will use the latest CI images published from hive master:
+              - image: registry.svc.ci.openshift.org/openshift/hive-v4.0:hive
```
dgoodwin: This would need replacing as well.

Makefile
```diff
@@ -75,7 +75,7 @@ install: crd rbac
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 .PHONY: deploy
-deploy: generate manifests
+deploy: install generate manifests
```
csrwng: shouldn't you call manifests before generate, since the result of manifests can affect the generated bindata.go?

csrwng:
@dgoodwin apart from the comment above, things look good to me. It'd be nice if we had a make target and some instructions (in the README) on how to generate the csv. But that could definitely be done in a follow-up PR.

hack/generate-operator-bundle.py
```diff
@@ -0,0 +1,100 @@
...
+    if (os.path.isfile(os.path.join('config/crds', file_name))):
+        shutil.copy(full_path, os.path.join(outdir, file_name))
+
+with open('config/templates/hive-csv-template.yaml', 'r') as stream:
+    csv = yaml.load(stream)
+
+csv['spec']['install']['spec']['clusterPermissions'] = []
+
+# Add our operator role to the CSV:
+with open('config/operator/operator_role.yaml', 'r') as stream:
```
csrwng: Just for future improvement ... wouldn't we want to add `config/operator/*_role.yaml` to the roles?

dgoodwin: I don't forsee more roles for the operator, but maybe we could move this one  alongside the others in config/rbac/, and then add config/rbac/*_role.yaml.

dgoodwin:
Updated. I will def add some docs on how to gen the CSV, process might evolve quite a bit but hopefully soon.

dgoodwin:
Sorry @csrwng :( I think this missing was why hiveadmission did not have perms to read it's configmap.

dgoodwin:
I have expanded this PR in scope as I kept discovering problems all related to getting this out to hive-stage, which is needed ASAP.

Added commits to transition to namespace "hive", and fix service account permission problems resulting from my attempt to kubectl apply -f config/rbac, instead of letting them flow through kustomize. 

We probably want to drop kubebuilder annotations for generating rbac soon.

@csrwng sorry, needs another look. :(

dgoodwin:
[test]

config/hiveadmission/hiveadmission_rbac_role.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
csrwng: maybe for validating later, but I don't believe hiveadmission needs any permissions for dnszones. We are able to validate other resources (clusterdeployments, clusterimagesets) without explicit permissions to those.

PR #239, opened 2019-03-01
Override PreserveOnDelete if cd is not installed

ShivamChamoli:
--> Skips the creation of uninstall job only if Spec.PreserveOnDelete and Status.Installed are true. If the cluster deployment is not installed, overrides the PreserveOnDelete to create an uninstall job.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -641,17 +641,22 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 		cdLog.WithField("jobName", installJob.Name).Info("install job deleted")
 	}
 
-	// Skips creation of uninstall job if PreserveOnDelete is true
+	// Skips creation of uninstall job if PreserveOnDelete is true and cluster is installed
 	if cd.Spec.PreserveOnDelete {
-		cdLog.Warn("skipping creation of uninstall job, due to PreserveOnDelete")
-		if controllerutils.HasFinalizer(cd, hivev1.FinalizerDeprovision) {
-			err = r.removeClusterDeploymentFinalizer(cd)
-			if err != nil {
-				cdLog.WithError(err).Error("error removing finalizer")
+		if cd.Status.Installed {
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -473,19 +473,23 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 
 	// Skips creation of uninstall job if PreserveOnDelete is true
 	if cd.Spec.PreserveOnDelete {
-		cdLog.Warn("skipping creation of uninstall job, due to PreserveOnDelete")
-		if controllerutils.HasFinalizer(cd, hivev1.FinalizerDeprovision) {
-			err = r.removeClusterDeploymentFinalizer(cd)
-			if err != nil {
-				cdLog.WithError(err).Error("error removing finalizer")
+		if cd.Status.Installed {
+			cdLog.Warn("skipping creation of uninstall job, due to PreserveOnDelete")
```
abhinavdahiya: This message is not completely correct now.

dgoodwin: It looks about right to me, here we are skipping the job because installed=true and preserve=true.

Maybe just "skipping creation of uninstall job for installed cluster due to PreserveOnDelete=true" for clarity.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -473,19 +473,23 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
...
+				}
+				return reconcile.Result{}, err
 			}
-			return reconcile.Result{}, err
+			return reconcile.Result{}, nil
 		}
-		return reconcile.Result{}, nil
+		// Overriding PreserveOnDelete because we might have deleted the cluster deployment before it finished
+		// installing, which can cause AWS resources to leak
+		cdLog.Infof("Overriding PreserveOnDelete because cluster deployment status is:%v", cd.Status.Installed)
```
dgoodwin: "PreserveOnDelete=true but launching uninstall as cluster was never successfully provisioned"

dgoodwin:
Still needing a rebase.

PR #241, opened 2019-03-04
Cluster image set

csrwng:


config/manifests/01_clusterdeployment_crd.yaml
```diff
@@ -56,6 +56,12 @@ spec:
                 nodes that need to be installed.
               items:
                 properties:
+                  labels:
+                    description: Map of label string keys and values that will be
+                      applied to the created MachineSet's MachineSpec. This list will
+                      overwrite any modifications made to Node labels on an ongoing
+                      basis.
+                    type: object
```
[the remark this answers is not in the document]
csrwng: me not running `make manifests` ... will be fixed in next push

config/samples/hive_v1alpha1_clusterimageset.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: hive.openshift.io/v1alpha1
+kind: ClusterImageSet
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: clusterimageset-sample
+spec:
+  hive:
+    pullSpec: openshift/hive:latest
```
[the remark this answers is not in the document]
csrwng: no, purely made up... will replace with valid ones. This is just the crd sample.

pkg/apis/hive/v1alpha1/clusterimageset_types.go
```diff
@@ -0,0 +1,130 @@
...
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClusterImageSetSpec defines the desired state of ClusterImageSet
+type ClusterImageSetSpec struct {
+	// hiveImage is the Hive image to use when installing or destroying a cluster.
```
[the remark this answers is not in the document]
csrwng: So in recent times, the case of new api type comments has been changed to reflect the names of the external json field names, instead of the Go names, given that those comments are used for external documentation (as in https://github.com/openshift/api/blob/master/operator/v1/types.go#L141-L155). However, that is not very evenly enforced. In our openshift API I see mostly the Go-style casing. I think that whatever we do we should be consistent about it, so I'm okay going to the Go-style case just to keep internal consistency.

pkg/controller/clusterimageset/clusterimageset_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		return reconcile.Result{}, err
+	}
+	logger := log.WithFields(log.Fields{"clusterImageSet": request.NamespacedName.String()})
+
+	logger.Info("reconciling ClusterImageSet")
+
+	if cis.Status.InstallerImage != nil {
+		logger.Debug("clusterImageSet has an InstallerImage set.")
+		err := r.clearStatusErrorResolvingInstallerImage(cis, logger)
+		return reconcile.Result{}, err
```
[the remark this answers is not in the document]
csrwng: Wouldn't we want to create a new ClusterImageSet at that point? I was thinking of these more as immutable (which reminds me I need to add validation)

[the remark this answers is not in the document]
dgoodwin: That is probably ok, we'd do 4.0beta2 or whatever during times like this.

csrwng:
@dgoodwin this should be ready for another look

pkg/imageset/updateinstaller_test.go
```diff
@@ -0,0 +1,177 @@
...
+			existingClusterDeployment: testClusterDeploymentWithErrorCondition(),
+			setupWorkDir:              setupSuccessfulExecutionWorkDir,
+			validateClusterDeployment: validateSuccessfulExecutionAfterFailure,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			client := fake.NewFakeClient(test.existingClusterDeployment)
+			workDir, err := ioutil.TempDir("", "test-update")
```
dgoodwin: Looks like workDir might not be getting cleaned up.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -326,6 +331,74 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+				testSecret(pullSecretSecret, pullSecretKey, "{}"),
+				testSecret(sshKeySecret, adminSSHKeySecretKey, "fakesshkey"),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				job := getImageSetJob(c)
+				if job == nil {
+					t.Errorf("did not find expected imageset job")
+				}
+			},
+		},
```
dgoodwin: Could you add a test for deprovision, when the CD references a ClusterImageSet that has since been deleted. I'm not quite sure what we want there but I think it might be ok, we don't need an installer image anymore, if we do start using the installer image to deprovision in future we still have a reference at least to what was used to install the cluster, although that's probably no good anymore. For now though, we'd just use the latest hive image defaulted in the code, which I think is correct as we want it to remain backwards compat.

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -152,8 +162,49 @@ type ClusterDeploymentStatus struct {
...
+type ClusterDeploymentConditionType string
+
+const (
+	// ClusterImageSetNotFoundCondition is set when the ClusterImageSet referenced by the
+	// ClusterDeployment is not found.
+	ClusterImageSetNotFoundCondition ClusterDeploymentConditionType = "ClusterImageSetNotFound"
+
+	// InstallerImageResolutionFailedCondition is a condition that indicates whether the job
+	// to determine the installer image based on a release image was successful.
+	InstallerImageResolutionFailedCondition ClusterDeploymentConditionType = "InstallerImageResolutionFailed"
```
dgoodwin: Not in this PR, but between this and incoming preflight checks we should start up a document in the repo outlining the possible conditions and what they mean as a contract for us and UHC.

config/samples/hive_v1alpha1_clusterimageset.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: hive.openshift.io/v1alpha1
+kind: ClusterImageSet
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: clusterimageset-sample
+spec:
+  hiveImage: registry.svc.ci.openshift.org/openshift/hive-v4.0:hive
+  releaseImage: quay.io/openshift-release-dev/ocp-release:4.0.0-0.6
```
dgoodwin: This is probably going to be the primary use case to install a cluster going forward, we should probably factor it into the cluster-deployment.yaml template and e2e tests. We could setup make deploy to automatically setup an openshift-latest that points to latest CI images (assuming there is one for release that we can pull)? Or should we drop it into the clusterdeployment template as another object?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -338,17 +343,142 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
+		err := r.Get(context.TODO(), types.NamespacedName{Name: cd.Spec.ImageSet.Name}, imageSet)
+		if err == nil && imageSet.Spec.HiveImage != nil {
+			return *imageSet.Spec.HiveImage, nil
+		}
+		if err != nil && !errors.IsNotFound(err) {
+			cdLog.WithError(err).WithField("clusterimageset", cd.Spec.ImageSet.Name).Error("unexpected error retrieving clusterimageset")
+			return "", err
+		}
+	}
+	return images.GetHiveImage(cdLog), nil
```
dgoodwin: Looks like we fall back to this if IsNotFound, which looks correct. :+1:

pkg/apis/hive/v1alpha1/validating-webhooks/clusterimageset_validating_admission_hook.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	}
+
+	if oldObject != nil {
+		// Add the new data to the contextLogger
+		contextLogger.Data["oldObject.Name"] = oldObject.Name
+	}
+
+	if !reflect.DeepEqual(oldObject.Spec, newObject.Spec) {
+		message := "ClusterImageSet.Spec is immutable"
```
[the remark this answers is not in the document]
csrwng: yes

config/samples/hive_v1alpha1_clusterimageset.yaml
csrwng: We can just add it in as another object in the clusterdeployment template. If ok with you, I'll handle in a follow-up PR.

csrwng:
@dgoodwin rebased and added additional deprovision tests to the clusterdeployment controller. I'll handle the move to using ClusterImageSets by default in a follow-up PR. I'm not sure if there is a "latest" release image, but will definitely look into it.

config/samples/hive_v1alpha1_clusterimageset.yaml
dgoodwin: Sounds good.

csrwng:
sorry @dgoodwin I missed fixing a comment's case. fixed now

PR #243, opened 2019-03-04
Remove HiveConfig image and use operator deployment image instead.

dgoodwin:
This will allow a more seamless CI flow instead of having to bump a
HiveConfig field every time we iterate and build a new operator image.
Assuming Hive controllers and admission should run the same image as the
operator should be safe.


dgoodwin:
@jmelis what can we do about the above bot question?

pkg/operator/hive/hive_controller.go
```diff
@@ -120,6 +122,37 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Regular manager client is not fully initialized here, create out own for some
```
csrwng: s/out/our/

dgoodwin: fixed

dgoodwin:
[test]

dgoodwin:
I have updated this PR to also remove the operator API status/spec we had on HiveConfig, which wasn't yet being used, and causing lots of problems with getting resources into etcd.

dgoodwin:
Also fixed bug where creating the new install config, if it didn't exist, was broken and we were ignoring the error. My bad!

Ready for another look @csrwng

PR #246, opened 2019-03-06
Use ClusterImageSets in clusterdeployment template and in e2e

csrwng:
Ensures that clusterimagesets and corresponding webhook are provisioned by the Hive operator. Makes the use of clusterimagesets the default in the clusterdeployment template and in the e2e test.

### DIFF
--- a/config/templates/cluster-deployment.yaml
+++ b/config/templates/cluster-deployment.yaml
@@ -35,16 +35,16 @@ parameters:
   description: Hive image pull policy
   value: Always
 - name: INSTALLER_IMAGE
-  displayName: OpenShift Installer image URL
-  description: OpenShift Installer image URL
-  value: "registry.svc.ci.openshift.org/openshift/origin-v4.0:installer"
+  displayName: OpenShift Installer image
+  description: OpenShift Installer image. Leave empty to use the default installer image from the release image.
+  value: ""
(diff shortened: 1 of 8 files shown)


config/templates/cluster-deployment.yaml
```diff
@@ -81,6 +89,24 @@ objects:
...
     ssh-publickey: "${SSH_KEY}"
 
+- apiVersion: hive.openshift.io/v1alpha1
+  kind: ClusterImageSet
+  metadata:
+    labels:
+      controller-tools.k8s.io: "1.0"
+    name: origin-latest
+  spec:
+    releaseImage: "registry.svc.ci.openshift.org/openshift/origin-release:v4.0"
```
dgoodwin: Does this work for you? I can't pull it.

csrwng: It did work in my manual test and I can see the imagestream on api.ci. Let me try it from a simple docker client

csrwng: I just spun a vm with just docker on it and I was able to pull it

dgoodwin: weird, i am getting unauthorized with both podman and docker.

csrwng: You may have an old token in your dockercfg for registry.svc.ci.openshift.org

PR #247, opened 2019-03-07
Function to apply arbitrary resource to target cluster

csrwng:
Vendors kubectl code from k8s.io and provides a function that will apply arbitrary bytes to a cluster, given its kubeconfig.


dgoodwin:
/lgtm

[test]

PR #248, opened 2019-03-08
remoteingress controller

joelddiaz:
extend ClusterDeployment to allow defining a list of ingress objects that should be configured on the cluster as IngressController objects for the openshift-ingress-operator to set up.
    
spec.ingress can be empty which will result in default cluster behavior (ie self-signed certs, *.apps.<base domain> default route, etc). otherwise, spec.ingress can provide for additional ingress objects (although you must provide settings for the 'default' ingress object in this case). and once you define ingress objects, you cannot go back to default behavior. all this is enforced by the admission webhooks.
    
when spec.ingress is defined, it will create a syncset object with all the desired ingress definitions so that the appropriate IngressController objects are set up in the cluster.
    
this version of the controller doesn't address having non-self-signed certificates created as part of the setup/sync process. this will come at a later date when the interaction between hive and the certificate request CRDs is finalized.

### DIFF
--- a/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
+++ b/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
@@ -40,7 +40,7 @@ const (
 )
 
 var (
-	mutableFields = []string{"Compute", "PreserveOnDelete"}
+	mutableFields = []string{"Compute", "Ingress", "PreserveOnDelete"}
 )
 
 // ClusterDeploymentValidatingAdmissionHook is a struct that is used to reference what code should be run by the generic-admission-server.
@@ -174,6 +174,17 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec 
 
 	// TODO: Put Create Validation Here (or in openAPIV3Schema validation section of crd)
 
+	if !validateIngressList(&newObject.Spec) {
+		message := "If defining the Ingress list, it must include a default ingress entry"
+		return &admissionv1beta1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
+				Message: message,
+			},
+		}
+	}
+
 	// If we get here, then all checks passed, so the object is valid.
 	contextLogger.Info("Successful validation")
 	return &admissionv1beta1.AdmissionResponse{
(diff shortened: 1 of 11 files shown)


joelddiaz:
@dgoodwin this is worth a review, but we still need to decide which fields from a ClusterIngress that we want to expose https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/apis/ingress/v1alpha1/clusteringress_types.go#L11

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -174,6 +174,17 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec
 
 	// TODO: Put Create Validation Here (or in openAPIV3Schema validation section of crd)
 
+	if !validateIngressList(&newObject.Spec) {
+		message := "If defining the Ingress list, it must include a default ingress entry"
```
vkareh: If I need to include a "default" ingress entry, would the `IngressDomain` be the same as the `BaseDomain`? Or once ingress entries are provided the client would have to ensure that one of them is named "default", regardless of its domain?

We currently limit the number of router shards on the clusters-service to 2, but the labels are arbitrarily entered by the user. Wondering if I need to hard-code one of them as "default" or whether I should silently provide the "default" one using the `BaseDomain`.

/cc @jhernand

vkareh: Whoops, didn't mean to trigger the bot

/uncc @jhernand

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func createClusterIngress(cd *hivev1.ClusterDeployment, ingress hivev1.ClusterIngress) *clusteringress.ClusterIngress {
+	newIngress := clusteringress.ClusterIngress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ingress.Name,
+			Namespace: remoteClusterIngressNamespace,
+		},
+		Spec: clusteringress.ClusterIngressSpec{
+			IngressDomain: stringPtr(ingress.IngressDomain),
```
csrwng: @joelddiaz should we have defaults for settings like replicas and nodePlacement?

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
csrwng: @vkareh looking at the default ingress for a 4.0 cluster, the ingressDomain is set to null in the spec, and then the operator sets one in the status.

joelddiaz: what i had in mind was that you could alter the clusteringress object named 'default', so if you wanted it changed from apps.example.com you could do that. although, the specific case i had in mind was just marking the default ingress as "internal only" (ie backed by a non-internet-facing ELB).

vkareh: if you're creating additional ingress objects, it sort-of makes sense that the default is the BaseDomain marked as internal.

What if I need to create separate internal shards? We limit to 2 currently, but we don't specify whether they need to be internal/external: they can both be internal if they so choose. How do I send the router shard scheme (internal vs internet-facing, for each ingress object) to Hive?

joelddiaz: i'm trying to catch up with the latest changes that have gone into this are of openshift, so forgive my vagueness.

setting internal vs external is dependent on the ingress object (now apparently renamed) exposing that tunable. the object as of this writing seems to only allow setting the domain https://github.com/openshift/api/blob/master/config/v1/types_ingress.go#L24 .

i'm installing a new cluster to investigate these changes and how they'll affect the desired functionality as exposed via hive

@vkareh where are you getting this limit of '2' from? this is the first i hear of it.

vkareh: @joelddiaz - no, we impose the limit ourselves in our cluster service, so nothing upstream - it doesn't relate to Hive (we just call Hive to provision the clusters)

pkg/controller/remoteingress/remoteingress_controller.go
joelddiaz: the new definitions in openshift/api have defaults for replicas (2), and there is a default (but not defined in the comments) for nodeplacement https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L114-L120

PR #249, opened 2019-03-08
Add patch resource helper function

csrwng:
Also adds a command to hiveutil to test either apply or patch commands


csrwng:
@dgoodwin ptal
I haven't been very successful in stubbing things to get a workable unit test. I think a more doable approach is going to be an integration-style test like the ones that kubebuilder generates by default.

In a follow up PR, I'll cleanup the integration tests that we have, add tests for these commands and add them as a make target so we can add them to CI

PR #250, opened 2019-03-08
Generate new OLM bundles and document testing.

dgoodwin:
Updates operator bundle generation script, document how to use it,
generate a bundle and CSV, build an OLM registry image, and then install
everything in a 4.0 cluster.

### DIFF
--- a/build/olm-registry/Dockerfile
+++ b/build/olm-registry/Dockerfile
@@ -1,0 +1,6 @@
+FROM quay.io/openshift/origin-operator-registry:latest
+
+COPY bundle manifests/hive
+RUN initializer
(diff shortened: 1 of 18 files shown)


dgoodwin:
@jmelis @ecordell here's what I ended up with. Using this code, and the bundle I've committed here, I was able to deploy hive and have almost everything come up running properly.

I think the only thing left is a small fix we'll need for the hiveadmission daemonset when run on 3.11, the master node labels are different. Will get this in monday morning.

The bundle is committed here, I will remove it before this goes to master, I just want @jmelis  to be able to see and possibly use it.

PR #251, opened 2019-03-09
Add integration tests for apply/patch

csrwng:
Fixes existing tests so that they pass
Adds tests for the resource apply/patch functions
Fixes error bug in Patch/Apply


jmelis:
[test]

PR #252, opened 2019-03-11
Remove generation of package bundle from OLM script.

dgoodwin:
We will limit this script to only generating a bundle for a specific
CSV. The packages file will be managed by the CD pipeline as it will
maintain a persistent store for the registry image mapping each version
to the one it replaces, for both stage/prod pipelines.


docs/olm.md
```diff
@@ -8,15 +8,27 @@ Ensure the Hive namespace exists:
 oc new-project hive
 ```
 
-Identify the current version of the operator you wish to replace, if applicable. (TODO)
+Identify the current version of the operator you wish to replace, if applicable. If not, you should be able to just specify a fake version.
```
csrwng: In what format? can you include an example?

docs/olm.md
```diff
@@ -8,15 +8,27 @@ Ensure the Hive namespace exists:
...
 
 ```
 GIT_HASH=`git rev-parse --short HEAD`
 GIT_COMMIT_COUNT=`git rev-list 9c56c62c6d0180c27e1cc9cf195f4bbfd7a617dd..HEAD --count`
 ./hack/generate-operator-bundle.py bundle/ 0.1.5-4b53b492 $GIT_COMMIT_COUNT $GIT_HASH quay.io/dgoodwin/hive:latest alpha
```
csrwng: Make it obvious that quay.io/dgoodwin/hive:latest should be replaced with your own image, OR use the default latest image from CI.

docs/olm.md
```diff
@@ -8,15 +8,27 @@ Ensure the Hive namespace exists:
...
+You should now have a ./bundle/[version]/ directory.
+
+You now need to create a package file for the bundle in ./bundle/hive.package.yaml
+
+
+```
+packageName: hive-operator
+channels:
+- name: alpha
+  currentCSV: CURRENTCSVNAME
```
csrwng: It'd be nice if we have add a simple script that will create it for you

docs/olm.md
```diff
@@ -25,9 +37,6 @@ sudo podman push quay.io/dgoodwin/hive-registry:latest
 
 Create a CatlogSource, OperatorGroup, and Subscription for hive:
 
-
-Create a CatalogSource in your cluster where OLM is running. This will spawn a registry pod using your image just published.
-
 ```
 kubectl apply -f hack/olm-registry/
```
csrwng: The catalog source points to your image https://github.com/openshift/hive/blob/master/hack/olm-registry/catalog-source.yaml#L8. Should that be a template instead?

docs/olm.md
dgoodwin: It really doesn't matter, I'll clarify. I believe the version field is pretty freeform from what I've been told.

csrwng:
[test]

csrwng:
[test]

jmelis:
@dgoodwin  `OLM_CHANNEL` is no longer used, I think that should be cleaned

dgoodwin:
Good catch thanks.

docs/olm.md
dgoodwin: Going down this path, I ended up just making the whole doc a script anyone who wants to try can work off.

csrwng:
really nice
/lgtm

PR #254, opened 2019-03-11
SyncSet api updates

abutcher:
Update SyncSet API for gvk and update cd status as lists of SyncSets or SelectorSyncSets with status.

### DIFF
--- a/pkg/apis/hive/v1alpha1/syncset_types.go
+++ b/pkg/apis/hive/v1alpha1/syncset_types.go
@@ -20,7 +20,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 )
 
@@ -55,8 +54,11 @@ const (
 
 // SyncObjectPatch represents a patch to be applied to a specific object
 type SyncObjectPatch struct {
-	// GroupVersionKind is the Group, Version and Kind of the object to be patched.
-	GroupVersionKind schema.GroupVersionKind `json:"groupVersionKind"`
+	// APIVersion is the Group and Version of the object to be patched.
+	APIVersion string `json:"apiVersion"`
+
+	// Kind is the Kind of the object to be patched.
+	Kind string `json:"kind"`
 
 	// Name is the name of the object to be patched.
 	Name string `json:"name"`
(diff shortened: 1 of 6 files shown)


pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -116,12 +118,48 @@ type SyncCondition struct {
...
 }
 
+// SelectorSyncSetObjectStatus describes the status of resources created or patches that
+// have been applied from a SyncSet.
+type SelectorSyncSetObjectStatus struct {
+	// Name is the name of the SelectorSyncSet.
+	Name string `json:"name"`
+
+	// Namespace is the Namespace of the SelectorSyncSet.
+	Namespace string `json:"namespace"`
```
csrwng: we don't need this, since the SelectorSyncSet is cluster scoped

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -116,12 +118,45 @@ type SyncCondition struct {
 	Message string `json:"message,omitempty"`
 }
 
+// SelectorSyncSetObjectStatus describes the status of resources created or patches that
+// have been applied from a SyncSet.
+type SelectorSyncSetObjectStatus struct {
```
csrwng: @abutcher so sorry for the very late comment, but just noticed this since you removed the namespace field. The SelectorSyncSetObjectStatus and SyncSetObjectStatus structs are identical. We could get rid of one. (Just SyncSetObjectStatus).

The other thing I noticed is that we should likely just name the fields like `Resources` and `Patches` instead of `SyncResources` and `SyncPatches`. The sync part is already implied in the type.

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -116,12 +114,33 @@ type SyncCondition struct {
...
+	// SyncResources is the list of SyncStatus for objects that have been synced.
+	// +optional
+	Resources []SyncStatus `json:"resources,omitempty"`
+
+	// SyncPatches is the list of SyncStatus for patches that have been applied.
+	// +optional
+	Patches []SyncStatus `json:"patches,omitempty"`
+
+	// Conditions is the list of conditions indicating apply success or failure as well as
+	Conditions []SyncCondition `json:"conditions"`
```
abutcher: Added `conditions` here as non-optional assuming we always want to indicate overall `ApplySuccess`/failure/`UnknownObject`.

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -79,21 +81,17 @@ type SyncObjectPatch struct {
...
-	// UpdateSuccess indicates whether the resource has been updates or not. If not,
-	// it should include a reason and message for the failure.
-	UpdateSuccess SyncConditionType = "UpdateSuccess"
-
-	// PatchSuccess indicates whether the patch has been applied or not. If not,
-	// it should include a reason and message for the failure.
-	PatchSuccess SyncConditionType = "PatchSuccess"
+	// ApplySuccess indicates whether the resource or patch has been applied or not.
+	// If not, it should include a reason and message for the failure.
+	ApplySuccess SyncConditionType = "ApplySuccess"
```
csrwng: append `SyncCondition` to each condition variable so we can distinguish from other constants in the hivev1 package
so s/ApplySuccess/ApplySuccessSyncCondition/

pkg/apis/hive/v1alpha1/syncset_types.go
csrwng: I'm not sure a general ApplySuccess/Failure condition is useful. We will report those for the individual resources/patches and I think that's enough. This is mainly for the UnknownObject case. I would still leave it as optional.

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -116,12 +114,34 @@ type SyncCondition struct {
 	Message string `json:"message,omitempty"`
 }
 
+// SyncSetObjectStatus describes the status of resources created or patches that have
+// been applied from a SyncSet or SelectorSyncSet.
+type SyncSetObjectStatus struct {
+	// Name is the name of the SyncSet.
+	Name string `json:"name"`
+
+	// SyncResources is the list of SyncStatus for objects that have been synced.
```
csrwng: Fix the comment

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -116,12 +114,34 @@ type SyncCondition struct {
...
+	// +optional
+	Resources []SyncStatus `json:"resources,omitempty"`
+
+	// SyncPatches is the list of SyncStatus for patches that have been applied.
```
csrwng: Fix the comment

PR #255, opened 2019-03-11
resources: add separate Info function

csrwng:
Separates out the Info function so that we can call it first for all resources before applying anything to the cluster.

### DIFF
--- a/contrib/pkg/testresource/command.go
+++ b/contrib/pkg/testresource/command.go
@@ -73,13 +73,19 @@ func newApplyCommand() *cobra.Command {
 			content := mustRead(args[0])
 			kubeconfig := mustRead(kubeconfigPath)
 			helper := resource.NewHelper(kubeconfig, log.WithField("cmd", "apply"))
-			info, err := helper.Apply(content)
+			info, err := helper.Info(content)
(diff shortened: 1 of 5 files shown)


csrwng:
@abutcher ptal

csrwng:
Looks like a timeout from the CI internal registry:
```
I0311 19:09:13.190917      32 log.go:172] Unsolicited response received on idle HTTP channel starting with "HTTP/1.0 408 Request Time-out\r\nCache-Control: no-cache\r\nConnection: close\r\nContent-Type: text/html\r\n\r\n<html><body><h1>408 Request Time-out</h1>\nYour browser didn't send a complete request in time.\n</body></html>\n"; err=<nil>
error: unable to extract layer sha256:fea65f4716612b39b4ff9bf1b3c5c51f816efe03d5db8eaa7e2946faabdd2532 from registry.svc.ci.openshift.org/ci-op-8db7m15n/stable@sha256:6af13f8c75af40f816821a2f30a9561dc652e7c83d9cadd1cc5e6e4922a3ae1f: stream error: stream ID 101; INTERNAL_ERROR
```

/retest

PR #259, opened 2019-03-12
Allow ClusterImageSet.Spec to be mutable

csrwng:


dgoodwin:
/hold

Code looks good I'm just not sure if this is what we want, need to see some more discussion in that email thread. I almost think we want to keep them immutable now, but use a global/HiveConfig value for deprovision instead of the one in the image set.

csrwng:
closing for now
/close

PR #260, opened 2019-03-12
Serving Certificate API changes for ClusterDeployments

csrwng:
API change to optionally request certificates for the target cluster.


csrwng:
@dgoodwin @joelddiaz - this is my proposed api change to trigger creation of cert requests. Please let me know if you think this is enough.

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -80,6 +80,10 @@ type ClusterDeploymentSpec struct {
 
 	// PreserveOnDelete allows the user to disconnect a cluster from Hive without deprovisioning it
 	PreserveOnDelete bool `json:"preserveOnDelete,omitempty"`
+
+	// CreateCertificates, if true, causes the cluster deployment to request TLS serving certificates
+	// for the api and router shards on the target cluster.
+	CreateCertificates bool `json:"generateCertificates,omitempty"`
```
dgoodwin: Name mismatch there, create vs generate in the json.

dgoodwin: What would happen if someone wanted to pass in their own pre-existing certs? I kinda think the API should be able to accomodate that even if it's not needed right now. A system where they give us some, if they don't we generate, would be nice. This flag here would give a third option which is "I didn't pass in certs and I don't want you to generate any" but I'm not sure anyone would want that.

joelddiaz: @dgoodwin so i suppose you'd have a spec.controlPlaneCertBundleSecret and then the ability to bring your own CertBundle for each defined ingress?

csrwng: @dgoodwin would we expect them to give us separate certs for api/apps/[other router shards]? If so, I would assume we want something like:
```
servingCerts:
  api:
    name: blah
  ingress:
  - name: default
    secret:
      name: blah
  - name: other-ingress
    secret:
      name: blahblah
```

joelddiaz: could it work like this:
if api.name is a non-nil secret reference and the secret exists, then that's the 'bring your own certificates' mode
if api.name is a non-nil secret reference and the secret doesn't exist, then we're in 'lets encrypt' mode
if api.name is nil, then it's default installer behavior of self-signed certs

csrwng: Hmm, I would prefer that they explicitly tell us what they want. If you want your certs generated, you shouldn’t have to fill in anything, just say that you want them created and we’ll handle it. We may want to fill out a section in status saying which certs were generated.

joelddiaz: after thinking it though some more, that setup i wrote above would be a pain to reconcile with the possibility that someone could come in and edit the clusterdeployment from the lets-encrypt mode to bring-your-own mode. and the behavior would be a bit too magical for my taste...so nevermind

joelddiaz: i do like organizing the control plane and the ingress list together since they all deal with the external-facing pieces of the cluster. need to check with the ingress team to see if they have plans to own/lifecycle the control plane parts of the cluster as well

csrwng:
@joelddiaz @twiest does this capture what we discussed? Naming suggestions are most welcome :)

config/crds/hive_v1alpha1_clusterdeployment.yaml
```diff
@@ -46,6 +46,32 @@ spec:
               description: BaseDomain is the base domain to which the cluster should
                 belong.
               type: string
+            certificateBundles:
+              description: CertificateBundles is a list of certificate bundles associated
+                with this cluster
+              items:
+                properties:
+                  generate:
+                    description: Generate indicates whether this bundle should be
```
joelddiaz: the way this description currently reads can be interpreted that `generate: true` might mean self-signed certs. maybe update the text to something like:
Generate indicates whether this bundle should have real (ie non-self-signed) certificates generated.

config/crds/hive_v1alpha1_clusterdeployment.yaml
```diff
@@ -46,6 +46,32 @@ spec:
...
+                      generated by the system.
+                    type: boolean
+                  name:
+                    description: Name is an identifier that must be unique within
```
joelddiaz: should the text really read "the bundle *must* be referenced by an ingress or by the control plane serving certs" or should it read "the bundle *can* be referenced by an ingress..."?

csrwng: We’d said that an unreferenced cert bundle would result in a validation error but I’m open to changing that.

csrwng:
Afaik the additional domains/bundles should point to the single load balancer for the masters

config/crds/hive_v1alpha1_clusterdeployment.yaml
csrwng: Fixed

config/crds/hive_v1alpha1_clusterdeployment.yaml
joelddiaz: if that's what we said, i believe you :) it's easier to loosen the restriction going forward if we need to.

twiest: I remember that as well. If we start by requiring it, that allows us to relax it in the future if there's a useful use case.

PR #261, opened 2019-03-12
Remove hiveadmission daemonset node selector.

dgoodwin:
We suspect this is blocking pod creation in hive-stage. Until we sort
out what is blocking this and if we can/should work past it, removing
the node selector moves hiveadmission to run on every compute node.

We may move to something other than a daemonset once this is fully
sorted out if this is desirable long term.

### DIFF
--- a/config/hiveadmission/daemonset.yaml
+++ b/config/hiveadmission/daemonset.yaml
@@ -46,12 +46,6 @@ spec:
             path: /healthz
             port: 9443
             scheme: HTTPS
-      nodeSelector:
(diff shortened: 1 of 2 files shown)


dgoodwin:
Thanks. :) Looks like a flake, or a new regression in 4.0.

csrwng:
I didn’t see any explicit errors which may mean we should increase the time we wait for the job to complete.

dgoodwin:
level=debug msg="Still waiting for the cluster to initialize: Cluster
operator monitoring is still updating"

/retest

PR #262, opened 2019-03-12
Use resource apply in the hive operator.

dgoodwin:
Most changes to deployments/daemonsets or any other objects the operator
manages would not automatically be picked up due to limitations around
comparing the spec. However our new apply library works perfectly here.

Added emptydir volumes for resource apply cache for both
hive-controllers and hive-operator.

Removed some dangerous owner refs from objects to the now-global
HiveConfig. Cleanup will have to be implemented as a future feature.

Removed unnecessary operator permissions to write CRDs. OLM does this
for us now.


dgoodwin:
cc @ShivamChamoli this is how the operator code looks now.

csrwng:
@dgoodwin you have a compile error

dgoodwin:
Rebased off master and fixed the compile error.

pkg/operator/hive/hive.go
```diff
@@ -17,87 +17,87 @@ limitations under the License.
...
 	}
+	h := resource.NewHelper(kubeconfig, hLog)
 
-	var changed bool
-	_, changed, err = resourceapply.ApplyDeployment(r.kubeClient.AppsV1(),
-		recorder, hiveDeployment, expectedDeploymentGen, false)
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme,
+		scheme.Scheme)
+
+	// TODO: it would be nice to be able to log if there were changes or not
```
csrwng: It looks like this is doable using the ToPrinter function passed to the command. I'll look into implementing and return an "action taken" (created, changed, unchanged) along with the error.

dgoodwin:
Found a larger error in here when I went to run it in-cluster from an image, it's running as system:anonymous and can't see anything.

I made two attempts at getting the rest config, first was just mgr.GetConfig(), and the second was:

```
    rules := clientcmd.NewDefaultClientConfigLoadingRules()    
    kc := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, &clientcmd.ConfigOverrides{})    
    restConfig, err := kc.ClientConfig()    
    if err != nil {    
        hLog.WithError(err).Error("error getting restconfig")    
        return err    
    }    
```

Which was what we do in hiveutil to get a client with the permissions from the pod SA. Both resulted in the same error though. Still working on it.

dgoodwin:
I think this is addressed now and hopefully good for e2e.

PR #265, opened 2019-03-13
Add hive-frontend role

dgoodwin:
@jhernand can you give this a look and let me know if it looks sufficient to you?

This may take some trial and error but the more we can get right now the better.

@jmelis you may want a look as well


jhernand:
The permissions look good to me. However I wonder if we will be able/allowed to get those permissions inside an OpenShift dedicated cluster.

dgoodwin:
This will deliver via OLM with hive -operator so we can literally add whatever we want here, and per Jira SRE has no issues with the elevated access at this time.

I am not sure if project requests are required. Should we add them for initial attempt?

dgoodwin:
Ok then this is probably good for a first attempt. I will remove WIP and see if we can get it to roll out.

PR #267, opened 2019-03-14
Add pods/log permission to hive-frontend role.

dgoodwin:
CC @jhernand


jhernand:
Looks good, thanks @dgoodwin.

PR #273, opened 2019-03-15
Increase timeout for CI install job

csrwng:
At least a couple of e2e tests have failed with us timing out waiting for the job to complete, but nothing wrong going on in the install log. This increases the timeout so we don't prematurely fail the job.


csrwng:
@joelddiaz ptal

PR #274, opened 2019-03-18
Add test to verify generated files

csrwng:
Also updates generated out-of-date deepcopy file.


csrwng:
@joelddiaz @twiest this should be ready for review now

hack/verify-generated.sh
```diff
@@ -0,0 +1,32 @@
...
+	echo "Generated deepcopy file is different. Run 'make generate'"
+	exit 1
+fi
+if ! diff -Naup "./pkg/operator/assets/bindata.go" "${tempdir}/bindata.go"; then
+	echo "Generated bindata file is different. Run 'make generate'"
+	exit 1
+fi
+
+mkdir "${tempdir}/crds"
+cp ${SRC_DIR}/config/crds/* "${tempdir}/crds/"
```
twiest: Would it make sense to move this up before any of the generations?

I'm know that today the things in the `crds` directory are only modified by `make manifests`, but what if in the future something in the `crds` directory gets modified by `make generate` ?

In other words, why not copy all of the source compares before making any changes?

csrwng: ack, makes sense

csrwng: thx @twiest, code updated

PR #278, opened 2019-03-19
Deploy a default set of ClusterImageSets via the operator.

dgoodwin:


dgoodwin:
CC @jhernand

hack/e2e-test.sh
```diff
@@ -106,7 +106,7 @@ while [ $i -le ${max_tries} ]; do
          AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" \
          BASE_DOMAIN="${BASE_DOMAIN}" \
          HIVE_IMAGE="${HIVE_IMAGE}" \
-         CUSTOM_RELEASE_IMAGE="${RELEASE_IMAGE}" \
+         RELEASE_IMAGE="${RELEASE_IMAGE}" \
          CLUSTER_IMAGE_SET="custom" \
```
csrwng: @dgoodwin this will cause the e2e to fail because we're pointing to an imageset that doesn't exist. Because we're no longer creating an imageset in the clusterdeployment template, maybe we should add another template for an imageset and have the e2e create one of those for its release image and then point the clusterdeployment to that.

dgoodwin: Is the operator in play for e2e? I thought it would come up, create the default image sets, one of which is the one referenced by default in the cluster deployment template. Then by setting an explicit RELEASE_IMAGE we would be overriding that.

dgoodwin: Ok update with a new template for custom image set, it removes the ability to set per CD overrides. It uses the same vars, HIVE_IMAGE and RELEASE_IMAGE, and includes a custom image set that uses those vars. INSTALLER_IMAGE is removed.

hack/e2e-test.sh
```diff
@@ -136,9 +135,7 @@ while [ $i -le ${max_tries} ]; do
   fi
 
   echo "Applying ClusterDeployment File ${CLUSTER_NAME}. Try #${i}/${max_tries}:"
-  if oc apply -f ${CLUSTER_DEPLOYMENT_FILE} ; then
-    echo "Success"
-    break
+  if oc apply -f ${CLUSTER_DEPLOYMENT_FILE} ; then echo "Success" break
```
csrwng: @dgoodwin you need a `;` between `echo "Success"` and `break`
(the e2e test kept repeating this apply 60 times)

dgoodwin: This looks like a vim keypress screwup, I have no memory of making this change intentionally.

dgoodwin:
Updated. :(

PR #279, opened 2019-03-19
Replacing hiveadmission daemonset to deployment

ShivamChamoli:
--> Replaces the hiveadmission daemonset with a deployment (replication = 2)
--> Removes the daemonset from the clusters that might already have it. Needs to be removed once it is run in opshive

### DIFF
--- a/pkg/operator/hive/hive_controller.go
+++ b/pkg/operator/hive/hive_controller.go
@@ -56,6 +56,9 @@ const (
 	hiveConfigName = "hive"
 
 	hiveOperatorDeploymentName = "hive-operator"
+
+	// legacyDaemonsetName is the old daemonset we will clean up if present.
+	legacyDaemonsetName = "hiveadmission"
 )
 
 // Add creates a new Hive Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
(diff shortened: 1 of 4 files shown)


pkg/operator/hive/hive_controller.go
```diff
@@ -56,6 +56,9 @@ const (
 	hiveConfigName = "hive"
 
 	hiveOperatorDeploymentName = "hive-operator"
+
+	// Daemonset name
+	daemonsetName = "hiveadmission"
```
dgoodwin: legacyDaemonsetName might help someone avoid confusion in future. Also the godoc comment is probably invalid.

pkg/operator/hive/hive_controller.go
```diff
@@ -246,3 +255,25 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
 
 	return reconcile.Result{}, nil
 }
+
+func (r *ReconcileHiveConfig) deleteLegacyComponents(hLog log.FieldLogger) error {
+	// Ensure legacy hiveadmission Daemonset is deleted, we switched to a Deployment:
+	// TODO: this can be removed once rolled out to opshive, our only persistent environment.
```
dgoodwin: Once rolled out to hive-stage and hive-prod. (opshive won't be getting any more updates)

PR #280, opened 2019-03-20
Place `genclient:nonNamespaced` before `genclient`

jhernand:
Without this the code generated by the `informers-gen` tool produces
code that then fails to compile with this error:

```
pkg/hive/informers/externalversions/hive/v1alpha1/interface.go:57:54:
unknown field 'namespace' in struct literal of type clusterImageSetInformer
```


jhernand:
@csrwng please review.

PR #281, opened 2019-03-20
HiveAdmission Fixes

dgoodwin:
Restores the setting of kube/service CA on APIService and webhooks, which was dropped in recent refactoring. This is now performed in the operator, but only if running on a 3.11 cluster.

### DIFF
--- a/config/hiveadmission/clusterimageset-webhook.yaml
+++ b/config/hiveadmission/clusterimageset-webhook.yaml
@@ -2,9 +2,9 @@
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: clusterdeployments.admission.hive.openshift.io
+  name: clusterimagesets.admission.hive.openshift.io
 webhooks:
-- name: clusterdeployments.admission.hive.openshift.io
+- name: clusterimagesets.admission.hive.openshift.io
   clientConfig:
     service:
       # reach the webhook via the registered aggregated API
--- a/pkg/operator/hive/hive_controller.go
+++ b/pkg/operator/hive/hive_controller.go
@@ -41,7 +41,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	//"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -59,6 +58,7 @@ const (
 
 	// legacyDaemonsetName is the old daemonset we will clean up if present.
 	legacyDaemonsetName = "hiveadmission"
+	kubeSystemNamespace = "kube-system"
 )
 
 // Add creates a new Hive Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -76,6 +76,11 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
 	c, err := controller.New("hive-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	r.(*ReconcileHiveConfig).globalClient, err = getClient(mgr.GetScheme())
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


config/hiveadmission/clusterimageset-webhook.yaml
```diff
@@ -2,9 +2,9 @@
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: clusterdeployments.admission.hive.openshift.io
+  name: clusterimagesets.admission.hive.openshift.io
```
csrwng: Thank you 👍

pkg/operator/hive/hiveadmission.go
```diff
@@ -56,33 +68,140 @@ func (r *ReconcileHiveConfig) deployHiveAdmission(hLog log.FieldLogger, h *resou
...
+
 	hLog.Info("hiveadmission components reconciled successfully")
 
 	return nil
 }
+
+func (r *ReconcileHiveConfig) injectCerts(apiService *apiregistrationv1.APIService, webhooks []*admregv1.ValidatingWebhookConfiguration, hLog log.FieldLogger) error {
+
+	// Locate the kube CA by looking up secrets in kube-system namespace, finding one of
+	// type 'kubernetes.io/service-account-token', and reading the CA off it.
```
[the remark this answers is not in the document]
dgoodwin: This would make a lot more sense. :)

pkg/operator/hive/hiveadmission.go
```diff
@@ -56,33 +68,140 @@ func (r *ReconcileHiveConfig) deployHiveAdmission(hLog log.FieldLogger, h *resou
...
+
+	return nil
+}
+
+// is311 returns true if this is a 3.11 OpenShift cluster. We check by looking for a ClusterVersion object,
+// which should only exist on OpenShift 4.x. We do not expect Hive to ever be deployed on pre-3.11.
+func (r *ReconcileHiveConfig) is311(hLog log.FieldLogger) (bool, error) {
+	clusterVersion := &openshiftapiv1.ClusterVersion{}
+	err := r.Client.Get(context.Background(), types.NamespacedName{Name: clusterVersionObjectName}, clusterVersion)
+	// The error package cannot determine a reason for this particular error, so resorting to string checking:
```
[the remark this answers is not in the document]
dgoodwin: Will update.

dgoodwin:
Updated for both suggestions.

pkg/operator/hive/hiveadmission.go
```diff
@@ -56,33 +68,139 @@ func (r *ReconcileHiveConfig) deployHiveAdmission(hLog log.FieldLogger, h *resou
...
+	}
+	kubeCA, ok := firstSATokenSecret.Data["ca.crt"]
+	if !ok {
+		return fmt.Errorf("secret %s did not contain key ca.crt", firstSATokenSecret.Name)
+	}
+	hLog.Debugf("found kube CA: %s", string(kubeCA))
+
+	// Load the service CA:
+	svcCertSignerSecret := &corev1.Secret{}
+	err = r.globalClient.Get(context.Background(), types.NamespacedName{Namespace: "openshift-service-cert-signer", Name: "service-serving-cert-signer-signing-key"}, svcCertSignerSecret)
```
csrwng: @dgoodwin you should be able to get this from the same secret as above, key `service-ca.crt`
If you do, then you can get rid of the globalClient and the need for hive to access additional namespaces.

csrwng: /cc @twiest

dgoodwin: They don't appear to be the same value.

```
(dgoodwin@wrx {metrics} ~/go/src/github.com/openshift/hive) $ k get secret deployer-token-thhp9 -o json | jq '.data."service-ca.crt"' | sha256sum                         (hive/192-168-42-204:8443/developer/hive)
bf4c4bb9a4d807a24a397eb3612ea7e604f6d35f28fe4f77b99c3c6e6984e5df  -
(dgoodwin@wrx {metrics} ~/go/src/github.com/openshift/hive) $ k get secret -n openshift-service-cert-signer service-serving-cert-signer-signing-key -o json | jq '.data."tls.key"' | sha256sum
d316aa93b245ce45b179629f3a85ef63a960996e0ec4f8a83160b880dd862eb4  -
```

csrwng: they aren't, but I don't believe you need the signer CA, the service-ca.crt that gets mounted in the pod should be sufficient.

twiest: This is my understanding as well.

dgoodwin: Updated, separate commit so we can roll back if needed.

pkg/operator/hive/hive_controller.go
```diff
@@ -59,6 +58,7 @@ const (
 
 	// legacyDaemonsetName is the old daemonset we will clean up if present.
 	legacyDaemonsetName = "hiveadmission"
+	kubeSystemNamespace = "kube-system"
```
csrwng: doesn't look like this is needed

dgoodwin: updated

PR #282, opened 2019-03-21
Add initial support for publishing prometheus metrics

dgoodwin:
This is just a starter PR, @ShivamChamoli will be taking over writing most of these. Vendors prometheus, exposes a couple cluster deployment controller metrics, and establishes a metrics calc loop outside of controllers for more global totals calculated every 2 minutes.

This can be tested with "make run" and "curl http://localhost:2112/metrics".

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -70,7 +70,9 @@ func newRootCommand() *cobra.Command {
 			}
 
 			// Create a new Cmd to provide shared dependencies and start components
-			mgr, err := manager.New(cfg, manager.Options{})
+			mgr, err := manager.New(cfg, manager.Options{
+				MetricsBindAddress: ":2112",
+			})
 			if err != nil {
 				log.Fatal(err)
 			}
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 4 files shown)


cmd/manager/main.go
```diff
@@ -139,7 +144,29 @@ func (writer glogWriter) Write(data []byte) (n int, err error) {
...
+
+	metricsClient, err := utils.GetKubeClient(metricsScheme)
+	if err != nil {
+		log.WithError(err).Fatal("error creating metrics kube client")
+	}
+	mc := metrics.Calculator{
+		Client:   metricsClient,
+		Interval: 2 * time.Minute,
+	}
+	go mc.Run()
```
csrwng: I would add this as a runnable to the manager

cmd/manager/main.go
```diff
@@ -139,7 +144,29 @@ func (writer glogWriter) Write(data []byte) (n int, err error) {
...
 	defer glog.Flush()
 	cmd := newRootCommand()
-	err := cmd.Execute()
+
+	log.Debug("starting /metrics http handler")
+	http.Handle("/metrics", promhttp.Handler())
+	go http.ListenAndServe(":2112", nil)
+	log.Info("started /metrics http handler on port 2112")
+
+	metricsScheme := scheme.Scheme
```
csrwng: From the variable, you get the impression that you're using a scheme just for metrics, but it's really an alias for the global scheme.Scheme. It may be better just to use that without the alias so it's understood that you're changing the global one.

cmd/manager/main.go
csrwng: and add a separate add_metrics.go file for it

cmd/manager/main.go
dgoodwin: Moving to the manager start actually let me use manager client and scheme so that cleans this up even more.

dgoodwin:
Updated to use controller-runtime metrics registry, removed my http listener, removed my obsolete metrics in cluster deployment controller, and now launching our metrics goroutine via the manager (and using it's client+scheme)

pkg/controller/metrics/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	metricClusterDeploymentsInstalledTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "hive_cluster_deployments_installed_total",
+		Help: "Total number of cluster deployments that are successfully installed.",
+	})
+)
+
+func init() {
+	metrics.Registry.MustRegister(metricClusterDeploymentsTotal)
+	metrics.Registry.MustRegister(metricClusterDeploymentsInstalledTotal)
+}
```
dgoodwin: @ShivamChamoli just as a note for you, the above is how you can create and register a metric in the cluster deployment controller.

pkg/controller/utils/utils.go
```diff
@@ -92,3 +93,20 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
 	finalizers.Delete(finalizer)
 	object.SetFinalizers(finalizers.List())
 }
+
+// GetKubeClient creates a new Kubernetes dynamic client.
+func GetKubeClient(scheme *runtime.Scheme) (client.Client, error) {
```
dgoodwin: @csrwng if it's ok i'd propose keeping this function, it could be generally useful.

PR #284, opened 2019-03-25
add Type info for the IngressController objects

joelddiaz:
this is needed for the ability to generically apply all kinds of object types (from https://github.com/openshift/hive/pull/247 )


csrwng:
```
msg="\t* aws_route.to_nat_gw.5: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"
```
/retest

PR #286, opened 2019-03-26
Initial syncset controller.

abutcher:


pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+
+	return reconcile.Result{}, nil
+}
+
+// applySyncSet evaluates all resources and patches in a given SyncSet and applies them to the cluster identified by kubeConfig
+func (r *ReconcileSyncSet) applySyncSet(ss hivev1.SyncSet, kubeConfig []byte, syncSetStatus *hivev1.SyncSetObjectStatus, ssLog log.FieldLogger) error {
+	h := r.applierBuilder(kubeConfig, r.logger)
+
+	for _, resource := range ss.Spec.Resources {
+		info, err := h.Info(resource.Raw)
```
csrwng: You need to first call h.Info on all ss.Spec.Resources. If any error occurs, then add the condition and don't try to sync anything.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+			return err
+		}
+
+		resourceHash := md5.Sum(resource.Raw)
+		resourceSyncStatus := hivev1.SyncStatus{
+			APIVersion: info.APIVersion,
+			Kind:       info.Kind,
+			Name:       info.Name,
+			Namespace:  info.Namespace,
+			Hash:       fmt.Sprintf("%x", resourceHash[:]),
```
csrwng: nit: s/resourceHash[:]/resourceHash/

pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+			if rss.Name == resourceSyncStatus.Name && rss.Namespace == resourceSyncStatus.Namespace {
+				found = true
+				if rss.Hash != resourceSyncStatus.Hash {
+					different = true
+				}
+				break
+			}
+		}
+
+		if !found || different {
```
csrwng: The condition here should probably be a little bit more nuanced ... if it's not found, different or a sufficient amount of time has passed since the last condition probe time, then we should go ahead and apply it.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+			ssLog.Debugf("resource %v has not been created or is different", info.Name)
+			err = h.Apply(resource.Raw)
+			if err != nil {
+				ssLog.WithError(err).Error("unable to apply resource")
```
csrwng: I believe we said that within a syncset, if we encounter an error we should not keep applying the rest of the resources because they may not make sense. For example, we fail to apply a CRD and then we try to apply instances of that CRD. So if you run into an error, update the condition and return the error.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+	return syncSetStatus
+}
+
+func (r *ReconcileSyncSet) updateClusterDeploymentStatus(cd *hivev1.ClusterDeployment, origCD *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	// Update cluster deployment status if changed:
+	if !reflect.DeepEqual(cd.Status, origCD.Status) {
+		cdLog.Infof("status has changed, updating cluster deployment")
+		cdLog.Debugf("orig: %v", origCD.Status)
```
csrwng: If we want to see the contents, we should at least output the serialized JSON, otherwise, it's not that useful.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+		cdLog.Debugf("new : %v", cd.Status)
+		err := r.Status().Update(context.TODO(), cd)
+		if err != nil {
+			cdLog.Errorf("error updating cluster deployment: %v", err)
```
csrwng: s/cluster deployment/cluster deployment status/

pkg/controller/syncset/syncset_controller.go
```diff
@@ -0,0 +1,439 @@
...
+}
+
+func (r *ReconcileSyncSet) setUnknownObjectSyncCondition(syncSetConditions []hivev1.SyncCondition, err error) []hivev1.SyncCondition {
+	status := corev1.ConditionFalse
+	reason := unknownObjectNotFoundReason
+	message := fmt.Sprintf("Info available for all SyncSet resources")
+	if err != nil {
+		status = corev1.ConditionTrue
+		reason = unknownObjectFoundReason
+		message = fmt.Sprintf("Unable to gather Info for SyncSet resource: %v", err)
```
csrwng: it'd be good to indicate the index of the resource that caused the error

PR #287, opened 2019-03-26
Prometheus metrics

ShivamChamoli:
Prometheus metrics for install jobs and uninstall jobs

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -29,6 +29,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 
+	"github.com/openshift/hive/pkg/install"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
@@ -41,11 +43,21 @@ var (
 		Name: "hive_cluster_deployments_installed_total",
 		Help: "Total number of cluster deployments that are successfully installed.",
 	})
+	metricInstallJobsRunningTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "hive_install_jobs_running_total",
+		Help: "Total number of install jobs running in Hive.",
+	})
+	metricUninstallJobsRunningTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "hive_uninstall_jobs_running_total",
+		Help: "Total number of uninstall jobs running in Hive.",
+	})
 )
 
 func init() {
 	metrics.Registry.MustRegister(metricClusterDeploymentsTotal)
 	metrics.Registry.MustRegister(metricClusterDeploymentsInstalledTotal)
+	metrics.Registry.MustRegister(metricInstallJobsRunningTotal)
+	metrics.Registry.MustRegister(metricUninstallJobsRunningTotal)
 }
 
 // Add creates a new metrics Calculator and adds it to the Manager.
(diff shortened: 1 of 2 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -97,8 +113,35 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 				installedTotal = installedTotal + 1
 			}
 		}
+		// jobs metrics
+		jobs := &batchv1.JobList{}
+		err = mc.Client.List(context.Background(), &client.ListOptions{}, jobs)
+		if err != nil {
+			log.WithError(err).Error("error listing install jobs")
+		}
+		mcLog.Infof("calculating jobs metrics")
```
dgoodwin: debug for this and it's pair line below when finished.

pkg/controller/metrics/metrics.go
```diff
@@ -97,8 +113,35 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 				installedTotal = installedTotal + 1
 			}
 		}
+		// jobs metrics
+		jobs := &batchv1.JobList{}
+		err = mc.Client.List(context.Background(), &client.ListOptions{}, jobs)
```
dgoodwin: This is listing every job in the cluster, lets use the labels in the query so we *only* fetch install/uninstall jobs and save a lot of memory.

pkg/controller/metrics/metrics.go
```diff
@@ -97,8 +113,35 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+		if err != nil {
+			log.WithError(err).Error("error listing install jobs")
+		}
+		mcLog.Infof("calculating jobs metrics")
+		installJobsTotal := 0
+		uninstallJobsTotal := 0
+		// legacyJobs is the variable that tells us about the number of old jobs that might exist in production
+		legacyJobs := 0
```
dgoodwin: Per the above query change you'll no longer be seeing legacy jobs so lets remove this.

pkg/controller/metrics/metrics.go
```diff
@@ -97,8 +113,35 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+		// calculating number of install jobs, uninstall jobs and legacy jobs using labels
+		for _, job := range jobs.Items {
+			if _, ok := job.Labels[installJobLabel]; ok {
+				installJobsTotal = installJobsTotal + 1
```
dgoodwin: The cards actually specified "running" which will help make this a bit more useful, as we know the number of install jobs will pretty much always be equal to the number of cluster deployments. Tracking the running install/uninstall jobs helps us understand the load a little better.

Lets include "running" in the metric names as well, hive_install_jobs_running_total perhaps.

pkg/controller/metrics/metrics.go
dgoodwin: This should be possible with some params to ListOptions.

pkg/controller/metrics/metrics.go
```diff
@@ -41,11 +42,26 @@ var (
...
+	})
+	metricUninstallJobsRunningTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "hive_uninstall_jobs_running_total",
+		Help: "Total number of uninstall jobs running in Hive.",
+	})
+	// installJobLabel is the label used for counting the number of install jobs in Hive
+	installJobLabel = "hive.openshift.io/install"
+
+	// uninstallJobLabel is the label used for counting the number of uninstall jobs in Hive
+	uninstallJobLabel = "hive.openshift.io/uninstall"
```
dgoodwin: Lets define these labels in one place, remove them from this file and see comment below.

pkg/install/generate.go
```diff
@@ -42,6 +42,11 @@ const (
 	tryInstallOnceAnnotation              = "hive.openshift.io/try-install-once"
 	tryUninstallOnceAnnotation            = "hive.openshift.io/try-uninstall-once"
 	clusterDeploymentGenerationAnnotation = "hive.openshift.io/cluster-deployment-generation"
+	// installJobLabel is the label used for counting the number of install jobs in Hive
+	installJobLabel = "hive.openshift.io/install"
+
+	// uninstallJobLabel is the label used for counting the number of uninstall jobs in Hive
+	uninstallJobLabel = "hive.openshift.io/uninstall"
```
dgoodwin: Lets expose these by making first char upper case, and then reference them in the metrics package, so we have only one definition of them floating around.

PR #288, opened 2019-03-27
Replace incorrect use of dockercfg pull secrets with dockerconfigjson.

dgoodwin:
Pull secrets are returned from try.openshift.com in a format that is invalid if used with a dockercfg secret. The correct secret type is dockercfgjson. 

This change converts our templates and all code to use the new secret type, and key within that secret.

We are not migrating pre-existing pull secrets, rather I have made sure to remove any code that uses those secrets after the cluster is installed, so outdated pull secrets should not be an issue in stage, which is the only environment where they could exist right now. Cluster deployment controller was unnecessarily loading pull secrets and reconstructing the install job even after installation, and the remote machine set controller was loading it just as an input to obtain a installer config, which it definitely does not need the pull secret set in.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -52,6 +52,8 @@ import (
 )
 
 const (
+	controllerName = "clusterDeployment"
+
 	// serviceAccountName will be a service account that can run the installer and then
 	// upload artifacts to the cluster's namespace.
 	serviceAccountName = "cluster-installer"
@@ -61,7 +63,7 @@ const (
 	// deleteAfterAnnotation is the annotation that contains a duration after which the cluster should be cleaned up.
 	deleteAfterAnnotation       = "hive.openshift.io/delete-after"
 	adminCredsSecretPasswordKey = "password"
-	pullSecretKey               = ".dockercfg"
+	pullSecretKey               = ".dockerconfigjson"
 	adminSSHKeySecretKey        = "ssh-publickey"
 	adminKubeconfigKey          = "kubeconfig"
 	clusterVersionObjectName    = "version"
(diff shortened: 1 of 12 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -260,77 +262,83 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
+	if err != nil && errors.IsNotFound(err) {
+		cdLog.Debug("no install job exists")
+		existingJob = nil
+	} else if err != nil {
+		cdLog.WithError(err).Error("error looking for install job")
 		return reconcile.Result{}, err
 	}
 
-	cdLog = cdLog.WithField("job", job.Name)
+	if cd.Status.Installed {
```
csrwng: should this check happen before we try to check for the existence of a job?

pkg/install/generate.go
```diff
@@ -103,7 +104,7 @@ func GenerateInstallerJob(
 			ValueFrom: &corev1.EnvVarSource{
 				SecretKeyRef: &corev1.SecretKeySelector{
 					LocalObjectReference: cd.Spec.PullSecret,
-					Key:                  ".dockercfg",
+					Key:                  dockerConfigSecretKey,
```
csrwng: extra credit: use the constant from corev1

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: existingJob is used regardless when we calculate status down below, so this part was still required, but we could do further refactoring if you would prefer.

pkg/install/generate.go
dgoodwin: On it.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: No, I think it's fine as it is... missed the part below. Thanks.

dgoodwin:
Updated to use the corev1.DockerConfigJsonKey in another spot, kept the existingJob use for now.

Also added the commit to add the 4.0.0-0.8 release as I am trying to get this rolled out today for @jhernand and @kbsingh and it depends on this work.

PR #289, opened 2019-03-27
DNSZone controller enhancements

csrwng:
Allows setting tags on DNS zones
Uses UID as reference when creating zone to avoid duplicates
Sets name servers and Zone ID in DNSZone status


csrwng:
make verify failure fixed :(
Sorry @twiest, if you can take a look again I'd appreciate it.

PR #291, opened 2019-03-29
Fix missing creds, subnets, security groups, and instance profiles on day 2 machine sets.

dgoodwin:
Changes in the installer renamed several objects we expected on name. This adjusts to the new naming. Long term we should query and find these objects so this doesn't happen again.

Added the creds reference on each machineset as the machine controllers are working towards use of minted creds, which will make this required.

Vendors latest installer code which exposes several API changes we have to accommodate somehow.

Day 2 machine sets still broken though, nodes won't join, possibly because of AMI mismatches.


dgoodwin:
/test e2e 

@abutcher can you take a look at this

csrwng:
@dgoodwin the error looks legit
```
level=fatal msg="failed to fetch Kubeconfig Admin Client: failed to load asset \"Install Config\": invalid \"install-config.yaml\" file: networking.clusterNetwork[0].hostPrefix: Invalid value: 9: cluster network host subnetwork prefix must not be larger size than CIDR 10.128.0.0/14"
```

dgoodwin:
This should fix, missed the e2e template.

config/templates/cluster-deployment-customimageset.yaml
```diff
@@ -106,7 +106,7 @@ objects:
       machineCIDR: "10.0.0.0/16"
       clusterNetworks:
         - cidr: "10.128.0.0/14"
-          hostSubnetLength: 9
+          hostSubnetLength: 23
```
dgoodwin: cc @jhernand the installer is now more strict about this value, I incremented to match the installer's defaults, if you are passing 9 from UHC install jobs will start erroring against hive-stage once this merges.

csrwng:
lgtm, will wait for @abutcher signoff

PR #292, opened 2019-03-29
WIP:installer pods restart metric

ShivamChamoli:
--> Basic implementation for computing metrics for installer pod restarts.
NOTE:
--> Initially the metric prints a NaN value as we are dividing by 0 (number of installer pods)
--> Metric name does not follow the naming guideline as 'metricInstallerPodsRestartsAverage'
seems better than 'metricInstallerPodsRestartsAverageTotal'

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -78,8 +78,8 @@ const (
 )
 
 var (
-	metricClusterDeploymentInstallRetriesTotal = prometheus.NewCounterVec(
-		prometheus.CounterOpts{
+	metricClusterDeploymentInstallRetriesTotal = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
 			Name: "hive_cluster_deployment_install_retries_total",
 			Help: "Number of retries for all install jobs, partitioned by cluster.",
 		},
@@ -100,10 +100,17 @@ func Add(mgr manager.Manager) error {
 // NewReconciler returns a new reconcile.Reconciler
 func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 	return &ReconcileClusterDeployment{
+<<<<<<< HEAD
 		Client:                               mgr.GetClient(),
 		scheme:                               mgr.GetScheme(),
 		remoteClusterAPIClientBuilder: controllerutils.BuildClusterAPIClientFromKubeconfig,
 		clusterDeplymentInstallRetriesMetric: metricClusterDeploymentInstallRetriesTotal,
+=======
+		Client:                        mgr.GetClient(),
+		scheme:                        mgr.GetScheme(),
+		amiLookupFunc:                 lookupAMI,
+		remoteClusterAPIClientBuilder: controllerutils.BuildClusterAPIClientFromKubeconfig,
+>>>>>>> final changes for metric
 	}
 }
 
@@ -114,8 +121,6 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
-
-	reconciler := r.(*ReconcileClusterDeployment)
 
 	// Watch for changes to ClusterDeployment
 	err = c.Watch(&source.Kind{Type: &hivev1.ClusterDeployment{}}, &handler.EnqueueRequestForObject{})
@@ -129,13 +134,9 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 		OwnerType:    &hivev1.ClusterDeployment{},
 	})
 
-	// Watch for pods created by a ClusterDeployment:
-	/*err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &hivev1.ClusterDeployment{},
-	})*/
+	// Watch for pods created by an install job:
 	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestsFromMapFunc{
-		ToRequests: handler.ToRequestsFunc(reconciler.selectorPodWatchHandler),
+		ToRequests: handler.ToRequestsFunc(selectorPodWatchHandler),
 	})
 	if err != nil {
 		return err
@@ -162,8 +163,7 @@ type ReconcileClusterDeployment struct {
 
 	// remoteClusterAPIClientBuilder is a function pointer to the function that builds a client for the
 	// remote cluster's cluster-api
-	remoteClusterAPIClientBuilder        func(string) (client.Client, error)
-	clusterDeplymentInstallRetriesMetric *prometheus.CounterVec
+	remoteClusterAPIClientBuilder func(string) (client.Client, error)
 }
 
 // Reconcile reads that state of the cluster for a ClusterDeployment object and makes changes based on the state read
@@ -871,6 +871,7 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
 	return currentSA, nil
 }
 
+<<<<<<< HEAD
 func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (bool, error) {
 	// for now we only support AWS
 	if cd.Spec.AWS == nil || cd.Spec.PlatformSecrets.AWS == nil {
@@ -933,6 +934,9 @@ func dnsZoneName(cdName string) string {
 }
 
 func (r *ReconcileClusterDeployment) selectorPodWatchHandler(a handler.MapObject) []reconcile.Request {
+=======
+func selectorPodWatchHandler(a handler.MapObject) []reconcile.Request {
+>>>>>>> final changes for metric
 	retval := []reconcile.Request{}
 
 	pod := a.Object.(*corev1.Pod)
@@ -943,47 +947,42 @@ func (r *ReconcileClusterDeployment) selectorPodWatchHandler(a handler.MapObject
 	}
 	if pod.Labels != nil {
 		if cdName, ok := pod.Labels[install.ClusterDeploymentNameLabel]; ok {
-			if cdNamespace, ok := pod.Labels[install.ClusterDeploymentNamespaceLabel]; ok {
-				retval = append(retval, reconcile.Request{NamespacedName: types.NamespacedName{
-					Name:      cdName,
-					Namespace: cdNamespace,
-				}})
-			}
+			retval = append(retval, reconcile.Request{NamespacedName: types.NamespacedName{
+				Name:      cdName,
+				Namespace: pod.Namespace,
+			}})
 		}
 	}
 	return retval
 }
 
+// Calculates the restart count for both hive and installer container for a cd and sets it to metricClusterDeploymentInstallRetriesTotal
 func (r *ReconcileClusterDeployment) updateInstallRetriesTotalMetric(clusterDeploymentName string, clusterDeploymentNamespace string, cdLog log.FieldLogger) {
-	labels := map[string]string{"cluster_deployment": clusterDeploymentName, "namespace": "true"}
-	installerPodLabels := map[string]string{install.ClusterDeploymentNameLabel: clusterDeploymentName, install.ClusterDeploymentNamespaceLabel: clusterDeploymentNamespace, install.InstallJobLabel: "true"}
-	currentRetiesMetric, err := metricClusterDeploymentInstallRetriesTotal.GetMetricWith(labels)
+	labels := map[string]string{"cluster_deployment": clusterDeploymentName, "namespace": clusterDeploymentNamespace}
+	installerPodLabels := map[string]string{install.ClusterDeploymentNameLabel: clusterDeploymentName, install.InstallJobLabel: "true"}
+	currentRetriesMetric, err := metricClusterDeploymentInstallRetriesTotal.GetMetricWith(labels)
 	if err != nil {
 		cdLog.Error("Error getting cluster deployment install retries total metrics.")
 	} else {
-		cdLog.Infof("CURRENT VALUE:%v", currentRetiesMetric)
+		cdLog.Infof("currentRetriesMetric:%v", currentRetriesMetric)
 		pods := &corev1.PodList{}
 		err = r.Client.List(context.Background(), client.MatchingLabels(installerPodLabels), pods)
 		if err != nil {
 			log.WithError(err).Error("error listing pods")
 		} else {
 			podRestarts := 0
-			cdLog.Infof("pod length:%v", len(pods.Items))
 			for _, pod := range pods.Items {
 				cdLog.Infof("pod name:%v", pod.Name)
 				for _, cs := range pod.Status.ContainerStatuses {
-					if cs.Name == "installer" {
-						cdLog.Infof("Container Status:%v", cs.Name)
-						cdLog.Infof("POD RESTARTS:%v", cs.RestartCount)
-						podRestarts = int(cs.RestartCount)
-					}
+					cdLog.Infof("Container Name:%v", cs.Name)
+					cdLog.Infof("Pod Restarts:%v", cs.RestartCount)
+					podRestarts = podRestarts + int(cs.RestartCount)
 				}
 			}
-			//TODO - Calculates wrong metrics for now, need to figure out how to get metric value
-			metricClusterDeploymentInstallRetriesTotal.WithLabelValues(clusterDeploymentName, clusterDeploymentNamespace).Add(float64(podRestarts))
-			cdLog.Infof("Pod restarts outside: %v", podRestarts)
+			// Sets the value of the metricClusterDeploymentInstallRetriesTotal
+			metricClusterDeploymentInstallRetriesTotal.WithLabelValues(clusterDeploymentName, clusterDeploymentNamespace).Set(float64(podRestarts))
 		}
-		cdLog.Infof("clusterDeplymentInstallRetriesMetric: %v", r.clusterDeplymentInstallRetriesMetric)
+		cdLog.Infof("Cluster restart metrics: %v", metricClusterDeploymentInstallRetriesTotal)
 	}
 }
 
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/prometheus/client_golang/prometheus"
+	klabels "k8s.io/apimachinery/pkg/labels"
 	"reflect"
 	"strconv"
 	"time"
@@ -78,9 +79,9 @@ const (
 )
 
 var (
-	metricClusterDeploymentInstallRetriesTotal = prometheus.NewGaugeVec(
+	metricClusterDeploymentInstallRetries = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "hive_cluster_deployment_install_retries_total",
+			Name: "hive_cluster_deployment_install_retries",
 			Help: "Number of retries for all install jobs, partitioned by cluster.",
 		},
 		[]string{"cluster_deployment", "namespace"},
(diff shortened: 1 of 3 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -135,6 +141,24 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+			log.WithError(err).Error("error listing pods")
+		} else {
+			podRestarts := 0
+			finalAverage := 0.0
+			for _, pod := range pods.Items {
+				for _, cs := range pod.Status.ContainerStatuses {
+					podRestarts = podRestarts + int(cs.RestartCount)
+				}
+			}
+			finalAverage = float64(podRestarts) / float64(len(pods.Items))
```
dgoodwin: Looking at: https://jira.coreos.com/browse/CO-293 my thought was this should be in the cluster deployment controller, rather than this metrics controller. I'm still learning about prometheus but for a metric like this I  was thinking that our goal is to record each install job once when it completes. As it stands right now I think we might be re-reporting data, if the pod dies and gets restarted we send a fresh influx of all the existing install jobs, even those that have been reported before. It's hard to know for sure without prometheus to test with, which we don't have right now, but I suspect that's what it must do.

I'm also thinking it's not a gauge, rather a histogram or a summary. (I've tried to document starting ideas in the jira card notes for some of these so keep an eye out) https://prometheus.io/docs/practices/histograms/

Prometheus would have our historical data I think, so our goal here should be to report installer restarts when we detect that the job has completed, one time only. I think this could happen here: https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L341 if you were to check if is cd.Status.Installed == false && job.completed = true, we know we've just seen it complete for the first time. However we shouldn't send the metric until we successfully update status below. 

Note that this will also be a good place to record the length of time the job ran for one of these next ones.

pkg/controller/metrics/metrics.go
```diff
@@ -135,6 +141,24 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+			mcLog.WithField("averageNumberOfPodInstallerRestarts", finalAverage).Debug("installer pod restarts")
+			metricUninstallJobsRunningTotal.Set(float64(finalAverage))
```
dgoodwin: Looks like you're updating the wrong metric here as well.

pkg/controller/metrics/metrics.go
dgoodwin: You shouldn't need to calculate the average either per the above metric type, the histogram/summary link above talks about being used for request times, I think that's essentially what we're doing here, it's just the number of install jobs restarts.

dgoodwin:
Ignore my entire previous review. Discussion this morning led to a few discoveries. Jira updated: https://jira.coreos.com/browse/CO-293

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -108,6 +127,15 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	})
+
+	// Watch for pods created by a ClusterDeployment:
+	/*err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &hivev1.ClusterDeployment{},
+	})*/
+	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(reconciler.selectorPodWatchHandler),
+	})
```
dgoodwin: Not sure if you worked this out with Cesar or not, my past experience with this was: https://github.com/openshift/cloud-credential-operator/blob/master/pkg/controller/credentialsrequest/credentialsrequest_controller.go#L119-L176

I think the request func should be removed from the Reconciler object and just be a regular func in this module, or inline like the link. 

I'm not sure if you need a predicate or not, more comments on the predicate in the handler below.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +886,61 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+	if pod.Labels != nil {
+		if cdName, ok := pod.Labels[install.ClusterDeploymentNameLabel]; ok {
+			if cdNamespace, ok := pod.Labels[install.ClusterDeploymentNamespaceLabel]; ok {
+				retval = append(retval, reconcile.Request{NamespacedName: types.NamespacedName{
+					Name:      cdName,
+					Namespace: cdNamespace,
+				}})
+			}
+		}
+	}
```
dgoodwin: So do we need a predicate to avoid calling this func if the pod is not an install job? If we call the func, and see no labels, or see our labels missing, we just return an empty Request array, that should mean no reconciles happen right? In that case maybe no predicate function is needed.

dgoodwin: Overall this looks ok lets just remove it from (r *ReconcileClusterDeployment)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +886,61 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+	return retval
+}
+
+func (r *ReconcileClusterDeployment) updateInstallRetriesTotalMetric(clusterDeploymentName string, clusterDeploymentNamespace string, cdLog log.FieldLogger) {
+	labels := map[string]string{"cluster_deployment": clusterDeploymentName, "namespace": "true"}
+	installerPodLabels := map[string]string{install.ClusterDeploymentNameLabel: clusterDeploymentName, install.ClusterDeploymentNamespaceLabel: clusterDeploymentNamespace, install.InstallJobLabel: "true"}
+	currentRetiesMetric, err := metricClusterDeploymentInstallRetriesTotal.GetMetricWith(labels)
+	if err != nil {
+		cdLog.Error("Error getting cluster deployment install retries total metrics.")
```
dgoodwin: You're ignoring the error here, I think that's good as metrics problems probably should not shut down reconciliation. :+1:

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +886,61 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+			}
+		}
+	}
+	return retval
+}
+
+func (r *ReconcileClusterDeployment) updateInstallRetriesTotalMetric(clusterDeploymentName string, clusterDeploymentNamespace string, cdLog log.FieldLogger) {
+	labels := map[string]string{"cluster_deployment": clusterDeploymentName, "namespace": "true"}
+	installerPodLabels := map[string]string{install.ClusterDeploymentNameLabel: clusterDeploymentName, install.ClusterDeploymentNamespaceLabel: clusterDeploymentNamespace, install.InstallJobLabel: "true"}
+	currentRetiesMetric, err := metricClusterDeploymentInstallRetriesTotal.GetMetricWith(labels)
```
dgoodwin: Reties :)

pkg/install/generate.go
```diff
@@ -47,6 +47,12 @@ const (
 
 	// UninstallJobLabel is the label used for counting the number of uninstall jobs in Hive
 	UninstallJobLabel = "hive.openshift.io/uninstall"
+
+	// ClusterDeploymentNameLabel is the label (along with ClusterDeploymentNamespaceLabel) that is used to identify the installer pod of a particular cluster deployment
+	ClusterDeploymentNameLabel = "hive.openshift.io/cluster-deployment-name"
+
+	// ClusterDeploymentNamespaceLabel is the label (along with ClusterDeploymentNameLabel) that is used to identify the installer pod of a particular cluster deployment
+	ClusterDeploymentNamespaceLabel = "hive.openshift.io/cluster-deployment-namespace"
```
dgoodwin: No need to track the namespace, we know it by looking at the pod itself.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +886,61 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+	if err != nil {
+		cdLog.Error("Error getting cluster deployment install retries total metrics.")
+	} else {
+		cdLog.Infof("CURRENT VALUE:%v", currentRetiesMetric)
+		pods := &corev1.PodList{}
+		err = r.Client.List(context.Background(), client.MatchingLabels(installerPodLabels), pods)
```
dgoodwin: Per below, lets drop the Namepace label, and instead of client.MatchingLabels just build your own client.ListOptions with both a label selector, and a namespace.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +886,61 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+		if err != nil {
+			log.WithError(err).Error("error listing pods")
+		} else {
+			podRestarts := 0
+			cdLog.Infof("pod length:%v", len(pods.Items))
+			for _, pod := range pods.Items {
+				cdLog.Infof("pod name:%v", pod.Name)
+				for _, cs := range pod.Status.ContainerStatuses {
+					if cs.Name == "installer" {
```
dgoodwin: It's actually the "hive" pod that will restart, we just run "installer" long enough to copy the binary to the "hive" pod. Could you make this a constant here: https://github.com/openshift/hive/blob/master/pkg/install/generate.go#L229 and re-use the value.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +886,61 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+						cdLog.Infof("Container Status:%v", cs.Name)
+						cdLog.Infof("POD RESTARTS:%v", cs.RestartCount)
+						podRestarts = int(cs.RestartCount)
+					}
+				}
+			}
+			//TODO - Calculates wrong metrics for now, need to figure out how to get metric value
```
dgoodwin: Uh oh, it looks like it may not be possible to read the current value of a counter, I suspect this is by design, something we shouldn't be doing.

Does this mean we should use a gauge so we can just set it to the value of the restarts on the pod? I know counters are preferred to gauges, and our value should only ever go up, but I can't see another way to work around this.

pkg/install/generate.go
```diff
@@ -284,6 +291,9 @@ func GenerateInstallerJob(
 			ActiveDeadlineSeconds: &deadline,
 			BackoffLimit:          &backoffLimit,
 			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: installerPodLabels,
+				},
```
dgoodwin: Lets apply use the install job label to the pod template as well, so we know it's both a pod for this cluster deployment, *and* it's an install job. Otherwise it will probably get confused when we get to uninstalls, or possibly other things as well.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -82,10 +98,11 @@ func Add(mgr manager.Manager) error {
...
 	return &ReconcileClusterDeployment{
-		Client:                        mgr.GetClient(),
-		scheme:                        mgr.GetScheme(),
-		amiLookupFunc:                 lookupAMI,
-		remoteClusterAPIClientBuilder: controllerutils.BuildClusterAPIClientFromKubeconfig,
+		Client:                               mgr.GetClient(),
+		scheme:                               mgr.GetScheme(),
+		amiLookupFunc:                        lookupAMI,
+		remoteClusterAPIClientBuilder:        controllerutils.BuildClusterAPIClientFromKubeconfig,
+		clusterDeplymentInstallRetriesMetric: metricClusterDeploymentInstallRetriesTotal,
```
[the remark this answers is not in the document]
dgoodwin: cc @twiest

pkg/install/generate.go
ShivamChamoli: not sure If I understand this correctly, Do we want the install job label on installerPodLabels? It has the InstallJobLabel: "true", is that not enough?

pkg/install/generate.go
ShivamChamoli: We need the namespace for watching the pod while reconciling the cd in selectorPodWatchHandler function. I am missing something?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
ShivamChamoli: weirdly I only see the Restart count fo up in the container named installer and not in hive. 
lastState: {}
    name: hive
    ready: true
    restartCount: 0
    state:
      running:
        startedAt: 2019-04-04T15:38:37Z
  - containerID: cri-o://38b3fe16544ce3338b03978b724abfd56ab9490d762aea9de3a70d0c4ed97895
    image: registry.svc.ci.openshift.org/openshift/origin-v4.0-2019-04-04-153251@sha256:a2bfd26fa4e88b089267167f3d53c12c58ecb792c337178a3cb33029635a4783
    imageID: registry.svc.ci.openshift.org/openshift/origin-v4.0-2019-04-04-153251@sha256:a2bfd26fa4e88b089267167f3d53c12c58ecb792c337178a3cb33029635a4783
    lastState:
      terminated:
        containerID: cri-o://3e4d87410b4998fc1211d430b9b9aefa0e9e57769b0b305f5281bb925ec3f20b
        exitCode: 1
        finishedAt: 2019-04-04T15:39:10Z
        reason: Error
        startedAt: 2019-04-04T15:39:05Z
    name: installer
    ready: false
    restartCount: 3
    state:
      terminated:

pkg/controller/clusterdeployment/clusterdeployment_controller.go
ShivamChamoli: To the first comment, yes and yes we do not need the predicate function (confirmed with Cesar). For the second comment, Cesar agrees It should not be part of the reconciler object

pkg/install/generate.go
dgoodwin: My mistake, you're right, it's already there!

Lets drop the namespace label though, we know that from the pod.

pkg/install/generate.go
dgoodwin: You can just use pod.Namespace, it's redudant information in the labels because the pod will always be in the same ns as the ClusterDeployment.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: This sounds like it may have to do with how you're testing a restart cycle, what did you change to make it fail? 

Changing hiveutil command args might be a better option.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -73,6 +75,20 @@ const (
 	clusterImageSetFoundReason            = "ClusterImageSetFound"
 )
 
+var (
+	metricClusterDeploymentInstallRetriesTotal = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "hive_cluster_deployment_install_retries_total",
```
dgoodwin: Correct me if I'm wrong but I believe we should drop the _total suffix now that it's a gauge.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: This commend still needs to be addressed, you need to replace client.MatchingLables with a ListOptions that includes your labels, and the namespace. (as we dropped the namespace from the labels themselves)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +878,56 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+		err = r.Client.List(context.Background(), client.MatchingLabels(installerPodLabels), pods)
+		if err != nil {
+			log.WithError(err).Error("error listing pods")
+		} else {
+			podRestarts := 0
+			for _, pod := range pods.Items {
+				cdLog.Infof("pod name:%v", pod.Name)
+				for _, cs := range pod.Status.ContainerStatuses {
+					cdLog.Infof("Container Name:%v", cs.Name)
+					cdLog.Infof("Pod Restarts:%v", cs.RestartCount)
```
dgoodwin: These should  not be logged at Info, but once you're happy it's working they can probably be removed entirely.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +878,56 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+					podRestarts = podRestarts + int(cs.RestartCount)
+				}
+			}
+			// Sets the value of the metricClusterDeploymentInstallRetriesTotal
+			metricClusterDeploymentInstallRetriesTotal.WithLabelValues(clusterDeploymentName, clusterDeploymentNamespace).Set(float64(podRestarts))
+		}
+		cdLog.Infof("Cluster restart metrics: %v", metricClusterDeploymentInstallRetriesTotal)
```
dgoodwin: Lets replace this log statement with something like if podRestarts > 0: cdLog.WithField("pod", pod.Name).Warn("installer pod containers have restarted X times")

ShivamChamoli: It did not let me use a formatter like %v, so I changed the message a little - cdLog.WithField("Installer Pod Restarts:", podRestarts).Warn("installer pod restarts > 1")

dgoodwin: Lets do

cdLog.WithFields(log.Fields{
    "pod": pod.Name,
    "restarts": podRestarts,
).Warn("installer pod has restarted")

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +879,56 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+		}
+	}
+	return retval
+}
+
+// Calculates the restart count for both hive and installer container for a cd and sets it to metricClusterDeploymentInstallRetries
+func (r *ReconcileClusterDeployment) updateInstallRetriesTotalMetric(clusterDeploymentName string, clusterDeploymentNamespace string, cdLog log.FieldLogger) {
+	labels := map[string]string{"cluster_deployment": clusterDeploymentName, "namespace": clusterDeploymentNamespace}
+	installerPodLabels := map[string]string{install.ClusterDeploymentNameLabel: clusterDeploymentName, install.InstallJobLabel: "true"}
+	_, err := metricClusterDeploymentInstallRetries.GetMetricWith(labels)
```
dgoodwin: Don't think we need this anymore now that we're not going to try to read a counters current value. Can drop the if/else below as well on the err check.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -856,6 +879,56 @@ func (r *ReconcileClusterDeployment) setupClusterInstallServiceAccount(namespace
...
+			log.WithError(err).Error("error listing pods")
+		} else {
+			podRestarts := 0
+			for _, pod := range pods.Items {
+				for _, cs := range pod.Status.ContainerStatuses {
+					podRestarts = podRestarts + int(cs.RestartCount)
+				}
+			}
+			if podRestarts > 0 {
+				cdLog.WithField("Installer Pod Restarts:", podRestarts).Warn("installer pod restarts > 1")
```
dgoodwin: Logger fields should be easy to search for in splunk or kibana, this will show up as "Installer Pod Restarts:=1", it should be more something like "restarts=1", but see the full example I sent as a comment in the PR.

dgoodwin:
I am going to close this PR, fix the verify problem, and submit with a commit of my own as I'd like to build on this to close our a card today. Replacement PR will be coming soon.

dgoodwin:
@ShivamChamoli you need to investigate why your editor doesn't gofmt automatically, it should be preventing this but it happens on most of your PRs.

PR #294, opened 2019-04-01
Deploy external-dns with hive

csrwng:
Initial implementation of DNS management in Hive
Includes the following API changes:
- A boolean in cluster deployment to indicate that we want the cluster's dns to be managed by Hive
- A list of root domains in HiveConfig to configure which domains we're managing DNS for. If you specify that you want your DNS managed, but your baseDomain is not a direct child of these root domains, a validation error will occur.
- External DNS configuration in HiveConfig to include an image and credentials for AWS

Hive operator changes:
- if domains are present in hiveconfig, admission is configured to use a configmap that lists root domains that it should validate
- if external dns is configured in hiveconfig, it will configure and create a deployment as well as a service account for it.

The script that generates the csv now includes the CRD for external-dns as well as its cluster role.


csrwng:
Coming in a separate PR:
- ClusterDeployment controller change to create DNSZone when a ClusterDeployment says to manage its DNS.
- DNSZone controller change to create parent entries when a DNSZone says to hook it up to its parent, along with status updates when the new zone is available.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -37,14 +43,41 @@ const (
 	clusterDeploymentAdmissionGroup    = "admission.hive.openshift.io"
 	clusterDeploymentAdmissionVersion  = "v1alpha1"
 	clusterDeploymentAdmissionResource = "clusterdeployments"
+
+	// ManagedDomainsFileEnvVar if present, points to a simple text
+	// file that includes a valid managed domain per line. Cluster deployments
+	// requesting that their domains be managed must have a base domain
+	// that is a direct child of one of the valid domains.
+	ManagedDomainsFileEnvVar = "MANAGED_DOMAINS_FILE"
```
twiest: Would it make sense to have a general purpose config file for hive admission that could contain "validDomains" as a section?

I'm thinking that in the future, it might be nice to have a place to put other config values and having a structure in place would be really nice.

config/external-dns/deployment.yaml
```diff
@@ -0,0 +1,27 @@
...
+    metadata:
+      labels:
+        name: external-dns
+    spec:
+      strategy:
+        type: Recreate
+      serviceAccountName: external-dns
+      containers:
+      - name: external-dns
+        image: registry.svc.ci.openshift.org/openshift/hive-v4.0:external-dns
```
dgoodwin: Not sure how to handle this image reference. If we deploy this one in stage and prod, it's possible we will be unclear what we're running, deployments won't bounce if a new image is published because nothing has changed, and we might also inadvertently start running new code in prod that we haven't tested in master just by virtue of the pod restarting for some unknown reason.

Feels like we may need to publish our own images for this to pin them for time being. This probably parallels with the card I have in Jira to check in with ART team on what we can do for publishing images outside of the release.

pkg/apis/hive/v1alpha1/dnsendpoint_types.go
```diff
@@ -0,0 +1,91 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DNSEndpointSpec defines the desired state of DNSEndpoint
+type DNSEndpointSpec struct {
+	// Endpoints is the list of DNS records to create/update
+	// +optional
```
dgoodwin: Should this actually be optional?

dgoodwin: Or is this part of that very specific CRD structure external-dns needs.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
csrwng: @twiest would it be ok to add that file when we need it? I don't think this approach limits our choices going forward.

pkg/apis/hive/v1alpha1/dnszone_types.go
```diff
@@ -32,6 +32,12 @@ type DNSZoneSpec struct {
 	// Zone is the DNS zone to host
 	Zone string `json:"zone"`
 
+	// LinkToParentDomain specifies whether DNS records should
+	// be automatically created to link this DNSZone with a
+	// parent domain.
+	// +optional
+	LinkToParentDomain bool `json:"linkToParentDomain,omitempty"`
```
dgoodwin: Sorry for the very basic question but why do we still have a DNSZone CRD in light of the DNSEndpoint, I don't fully understand the differences.

pkg/apis/hive/v1alpha1/hiveconfig_types.go
```diff
@@ -32,6 +44,32 @@ type HiveConfigStatus struct {
 	AggregatorClientCAHash string `json:"aggregatorClientCAHash,omitempty"`
 }
 
+// ExternalDNSConfig contains settings for running external-dns in a Hive
+// environment.
+type ExternalDNSConfig struct {
+
+	// Image is a reference to the image that will run the external-dns controller.
+	// If not specified, a default image will be used.
+	Image string `json:"image,omitempty"`
```
dgoodwin: Just a comment, we can't use this for hive-stage as we don't have perms to edit it, and we'd need it to potentially be automatically rolled out via jenkins, so we'd probably use the deployment.yaml itself right now? 

Image override might be useful for someone external using hive, but then I'm not sure, I'm not totally clear on how all these images need to get pinned and versioned just yet.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
twiest: @csrwng I was just thinking it might be nice to have 1 config file and have the infra already built out for the next one.

It's honestly fine this way.

pkg/apis/hive/v1alpha1/hiveconfig_types.go
csrwng: If we can't edit hiveconfig how do we configure managed dns?

pkg/apis/hive/v1alpha1/dnsendpoint_types.go
csrwng: An empty set of records is valid and useful depending on the sync mode you're using. If you are using sync mode (right now we do upsert), have records you have created previously, and you want them deleted, you'd set the endpoints to empty.

pkg/apis/hive/v1alpha1/dnszone_types.go
csrwng: DNSZone represents a route53 hosted zone. A DNSEndpoint is simply a collection of route53 records. Unfortunately, right now it cannot be named anything other than `DNSEndpoint`. This is something that should be fixed upstream. My preference would be to call this a DNSRecordSet

PR #295, opened 2019-04-02
Apply syncset patches.

abutcher:


pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -68,6 +68,12 @@ type SyncObjectPatch struct {
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
 
+	// PatchApplyMode indicates if the patch apply mode is "AlwaysApply" (default) or "ApplyOnce".
+	// PatchApplyode "AlwaysApply" indicates that the patch should be applied every time reconcilation occurs.
+	// PatchApplyMode "ApplyOnce" indicates that the patch should only be applied once.
+	// +optional
+	PatchApplyMode SyncSetPatchApplyMode `json:"mode,omitempty"`
```
csrwng: The go variable name should match the json name. In this case, it should be ok to just name it Mode or ApplyMode since the context is a patch.

csrwng:
@abutcher one more comment... you'll need to add code to validate that your syncset status for patches matches what you're expecting as we do for resources:
https://github.com/openshift/hive/blob/9be2613f22bf674cb5b89e8906fe03f06d7d7a83/pkg/controller/syncset/syncset_controller_test.go#L649-L662

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -68,11 +68,11 @@ type SyncObjectPatch struct {
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
 
-	// PatchApplyMode indicates if the patch apply mode is "AlwaysApply" (default) or "ApplyOnce".
-	// PatchApplyode "AlwaysApply" indicates that the patch should be applied every time reconcilation occurs.
-	// PatchApplyMode "ApplyOnce" indicates that the patch should only be applied once.
+	// ApplyMode indicates if the patch apply mode is "AlwaysApply" (default) or "ApplyOnce".
+	// Applyode "AlwaysApply" indicates that the patch should be applied every time reconcilation occurs.
```
csrwng: s/Applyode/ApplyMode/

abutcher: 🤦‍♂️

PR #296, opened 2019-04-03
Add beta-3 ImageSet

csrwng:


dgoodwin:
/lgtm
/hold

Cancel when you can see it install, thanks!

PR #298, opened 2019-04-03
Implement SyncSet sync mode

csrwng:


csrwng:
@abutcher this should be ready for review

PR #299, opened 2019-04-06
Generate install-config in hiveutil install wrapper.

dgoodwin:
To help alleviate problems where one version of hive can only generate
one version of the install config, but we need to install many possible
releases which may not support that install config version.

This change now assumes that the hiveutil install wrapper looks up the
cluster deployment, and then generates the install config on the fly. By
pinning one version of hiveutil to one version of the installer (via a
ClusterImageSet), we should remain more compatible over time.

The previous method of generating an install config in the controller
and mounting via config map is still left behind for backward
compatability. Once we can drop beta2 and beta3 ClusterImageSets this
can be removed.

### DIFF
--- a/contrib/pkg/installmanager/installmanager.go
+++ b/contrib/pkg/installmanager/installmanager.go
@@ -14,12 +14,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/hive/pkg/apis"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	"github.com/openshift/hive/pkg/install"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


dgoodwin:
I think this is ready for review now.

contrib/pkg/installmanager/installmanager.go
```diff
@@ -175,12 +178,34 @@ func (m *InstallManager) Run() error {
 
 	m.waitForInstallerBinaries()
 
-	dstInstallConfig := filepath.Join(m.WorkDir, "install-config.yaml")
-	m.log.Debugf("copying %s to %s", m.InstallConfig, dstInstallConfig)
-	err = m.copyFile(m.InstallConfig, dstInstallConfig)
```
csrwng: the copyFile function doesn't seem to be used anywhere anymore, remove it.

contrib/pkg/installmanager/installmanager.go
```diff
@@ -175,12 +178,34 @@ func (m *InstallManager) Run() error {
...
+	// Generate an install-config.yaml:
+	sshKey, err := controllerutils.LoadSecretData(m.DynamicClient, cd.Spec.SSHKey.Name,
```
csrwng: Both ssh key and pull secret should be available as env vars when this runs, should we not use that?

dgoodwin: IIRC we kept those because they seemed to still be needed when we switched from env vars to install config to configure the installer. Looking now I can't see any hits for either OPENSHIFT_INSTALL_SSH_PUB_KEY or OPENSHIFT_INSTALL_PULL_SECRET so it looks to me like those are dead vars and we should probably kill them off.

dgoodwin: Pushed a change to remove those and fix the above, will see if it passes CI.

dgoodwin:
Updated to use env vars for ssh key and pull secret.

PR #302, opened 2019-04-08
Remove and cleanup beta2 ClusterImageSet.

dgoodwin:


pkg/operator/hive/hive.go
```diff
@@ -79,6 +81,30 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		}
 	}
 
+	// Remove legacy ClusterImageSets we do not want installable anymore.
+	removeImageSets := []string{
+		"openshift-v4.0-beta2",
```
csrwng: We should also cleanup "openshift-4.0.0-0.8"

dgoodwin: Removed.

csrwng: Sorry @dgoodwin I gave you the wrong string (I was looking at the filename). It should be `openshift-v4.0.0-0.8` https://github.com/openshift/hive/commit/aa6544fe75ddb5e45966be20782639ddc6228115#diff-b22803057b433f989a4b6cad8b653430R6

dgoodwin: Whoops, I missed it too. Updated.

dgoodwin:
ready to go @csrwng

PR #303, opened 2019-04-08
add support for ingress objects referencing certificates

joelddiaz:
plumb through code to pull the certBundle's secrets that ingress objects refer to so that the syncset created also makes sure a secret is mirrored over to the remote cluster so that the ingressController objects on the remote cluster can come up with custom certificates.

### DIFF
--- a/pkg/controller/remoteingress/remoteingress_controller.go
+++ b/pkg/controller/remoteingress/remoteingress_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 
@@ -28,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -39,15 +41,32 @@ import (
 
 	ingresscontroller "github.com/openshift/api/operator/v1"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	"github.com/openshift/hive/pkg/controller/utils"
+	"github.com/openshift/hive/pkg/resource"
 )
 
 const (
 	controllerName = "remoteingress"
 
-	remoteClusterIngressNamespace = "openshift-ingress-operator"
+	// namespace where the ingressController objects must be created
+	remoteIngressControllerNamespace = "openshift-ingress-operator"
 
-	remoteIngressKind = "IngressController"
+	// while the IngressController objects live in openshift-ingress-operator
+	// the secrets that the ingressControllers refer to must live in openshift-ingress
+	remoteIngressConrollerSecretsNamespace = "openshift-ingress"
+
+	ingressCertificateNotFoundReason = "IngressCertificateNotFound"
+	ingressCertificateFoundReason    = "IngressCertificateFound"
+
+	// requeueAfter2 is just a static 2 minute delay for when to requeue
+	// for the case when a necessary secret is missing
+	requeueAfter2 = time.Minute * 2
 )
+
+// kubeCLIApplier knows how to ApplyRuntimeObject.
+type kubeCLIApplier interface {
+	ApplyRuntimeObject(obj runtime.Object, scheme *runtime.Scheme) error
+}
 
 // Add creates a new RemoteMachineSet Controller and adds it to the Manager with default RBAC. The Manager will set fields on the
 // Controller and Start it when the Manager is Started.
@@ -57,10 +76,16 @@ func Add(mgr manager.Manager) error {
 
 // NewReconciler returns a new reconcile.Reconciler
 func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
+	logger := log.WithField("controller", controllerName)
+	helper, err := resource.NewHelperFromRESTConfig(mgr.GetConfig(), logger)
+	if err != nil {
+		logger.WithError(err).Fatalf("cannot create resource helper for reconciler")
+	}
 	return &ReconcileRemoteClusterIngress{
-		Client: mgr.GetClient(),
-		scheme: mgr.GetScheme(),
-		logger: log.WithField("controller", controllerName),
+		Client:  mgr.GetClient(),
+		scheme:  mgr.GetScheme(),
+		logger:  log.WithField("controller", controllerName),
+		kubeCLI: helper,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


joelddiaz:
@csrwng this should be ready for a real review now. all the pieces are implemented

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -58,12 +81,28 @@ func Add(mgr manager.Manager) error {
...
+		scheme:         mgr.GetScheme(),
+		logger:         log.WithField("controller", controllerName),
+		restConfig:     mgr.GetConfig(),
+		kubeCLIBuilder: kubeCLIBuilderFunc,
 	}
 }
 
+func kubeCLIBuilderFunc(config *rest.Config, logger log.FieldLogger) (kubeCLIApplier, error) {
+	var cliClient kubeCLIApplier
+	clientConfig := operatorutil.GenerateClientConfigFromRESTConfig("anything", config)
```
csrwng: Just fyi ... I'm adding another constructor for resource.Helper that takes a rest.Config in my PR. We should eventually switch to just one.

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -81,22 +120,32 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 	return nil
 }
 
+type reconcileContext struct {
+	clusterDeployment *hivev1.ClusterDeployment
+	certBundleSecrets map[string]map[string][]byte
```
csrwng: this seems like an odd struct :) 
in my controller I just gather a slice of secrets and pass that to the function that generates the syncset
I don't think you need to be able to look up the secrets by name since you already get that from the certificateBundles struct
You also lose some info when only obtaining the secret data and not preserving secret type, annotations, labels

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -169,62 +273,79 @@ func newSyncSetSpec(cd *hivev1.ClusterDeployment, rawExtensions []runtime.RawExt
...
-		}
+	kubeCLI, err := r.kubeCLIBuilder(r.restConfig, rContext.logger)
+	if err != nil {
+		rContext.logger.WithError(err).Error("error building kubeCLI client")
+	}
 
-		return nil
-	} else if err != nil {
-		r.logger.Errorf("error checking for existing syncset: %v", err)
+	serializer := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
```
csrwng: fyi ... in my branch I'm changing the signature of the function to take a scheme and then just build the serializer in the implementation. But hopefully yours merges first and then I'll just update it here as well.

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -233,5 +354,98 @@ func createIngressController(cd *hivev1.ClusterDeployment, ingress hivev1.Cluste
...
+		}
+	}
+
+	return nil
+}
+
+// secretNameForCertificateBundle just stitches together a secret name consisting of
+// the clusterDeployment.Name plus the name of the certificateBundle.Name
+func secretNameForCertificateBundle(certBundleName string, cd *hivev1.ClusterDeployment) string {
+	return fmt.Sprintf("%s-%s", cd.Name, certBundleName)
```
csrwng: I would think we want to create a secret with the same name that was specified in the certificate bundle

joelddiaz: the thinking i had with this naming scheme was to avoid someone naming the certbundle 'default' (or the secret referenced by the certBundle) and then trying to sync down a secret named 'default' on the remote cluster and getting a collision (not that there's a secret named 'default' in the 'openshift-ingress' namespace...but I hope i've explained the idea).

i'm open to just passing through the secret name, but what do you think about the collisions?

csrwng: I was thinking of being able to find the secret on the target cluster if I need to debug the target cluster directly. I know what I named my secret locally, but there's nothing telling me that it's now named something else on the target. I do agree that there's a danger of collisions, so I'm split :) 
If you think this is better, then let's keep as is.

pkg/controller/remoteingress/remoteingress_controller.go
joelddiaz: it took me a while to remember what i was thinking here. i believe what i had in mind was two ingress entries referencing the same certBundle. i was trying to avoid making two secret entries in the syncSet.

i suppose i can still pass the slice with possible duplicate entries, and then when building the syncset, make sure i don't add it twice...

csrwng: This is what I do:
https://github.com/csrwng/hive/blob/master_certs/pkg/controller/controlplanecerts/controlplanecerts_controller.go#L178-L221

csrwng: basically use stringsets to remove dups

joelddiaz: @csrwng i updated it to just save the slice of secrets, and fixed up the tests accordingly

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -313,26 +318,27 @@ func (r *ReconcileRemoteClusterIngress) syncSyncSet(rContext *reconcileContext,
...
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Secret",
+					APIVersion: corev1.SchemeGroupVersion.String(),
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      secretNameForCertificateBundle(certBundle.Name, rContext.clusterDeployment),
+					Namespace: remoteIngressConrollerSecretsNamespace,
+				},
+				Data: secret.Data,
+			}
```
csrwng: @joelddiaz you should copy over the secret type at the very least.
I'll leave it up to you if you think you should also copy annotations and labels

pkg/controller/remoteingress/remoteingress_controller.go
joelddiaz: at least for ingressControllers, you can view the ingresscontroller object itself since it will have the defaultCertificate field filled out

joelddiaz: i ended up going with a scheme where we take the certBundle's secret's name and just prepending the clusterName in front

pkg/controller/remoteingress/remoteingress_controller.go
joelddiaz: done. now i deepcopy the original secret and just update the name and namespace where it needs to land

csrwng: @joelddiaz that sounds fine, though it'd be good to test it. There's a few system-generated in object meta that don't apply to a new object (UID, selfLink, resourceVersion). Which is why I only copy things explicitly --> https://github.com/openshift/hive/pull/310/files#diff-b945d6ca702f17e881579fd8805078cfR337

pkg/controller/remoteingress/remoteingress_controller.go
csrwng: I will follow suit in my pull

csrwng:
@joelddiaz this looks good to me, are you going to remove the WIP in the title?

csrwng:
@joelddiaz this looks good to me, once you squash, I'll tag it

PR #304, opened 2019-04-09
deprovision job duration metrics

ShivamChamoli:
1) Calculated deprovision job duration metrics by assigning a label to the uninstall job which has the unix timestamp when it is created, and found out the difference between the unix timestamp when the uninstall job finished.
2) Used unix timestamp so that there is no date compatibility issue.
3) Metrics were not calculated if the job failed at any step.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -86,10 +86,18 @@ var (
 		},
 		[]string{"cluster_deployment", "namespace"},
 	)
+	deprovisionJobsDurationHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "deprovision_jobs_duration_seconds",
+			Help:    "Deprovision Jobs duration distribution",
+			Buckets: []float64{60, 3600, 7200},
+		},
+	)
 )
 
 func init() {
 	metrics.Registry.MustRegister(metricClusterDeploymentInstallRestarts)
+	metrics.Registry.MustRegister(deprovisionJobsDurationHistogram)
 }
 
 // Add creates a new ClusterDeployment Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -741,6 +749,10 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 		// Uninstall job exists, check it's status and if successful, remove the finalizer:
 		if controllerutils.IsSuccessful(existingJob) {
 			cdLog.Infof("uninstall job successful, removing finalizer")
+			// jobDuration calculates the time elapsed since the uninstall job started for deprovision job
+			jobDuration := existingJob.Status.CompletionTime.Time.Sub(existingJob.Status.StartTime.Time)
+			cdLog.WithField("duration", jobDuration.Seconds()).Debug("uninstall job completed")
+			deprovisionJobsDurationHistogram.Observe(float64(jobDuration.Seconds()))
 			err = r.removeClusterDeploymentFinalizer(cd)
 			if err != nil {
 				cdLog.WithError(err).Error("error removing finalizer")
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -27,8 +27,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 	"strconv"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 )
 
 const (


pkg/install/generate.go
```diff
@@ -381,7 +384,11 @@ func GenerateUninstallerJob(
 	completions := int32(1)
 	deadline := int64((24 * time.Hour).Seconds())
 	backoffLimit := int32(123456) // effectively limitless
-	labels := map[string]string{UninstallJobLabel: "true"}
+	// deprovisionStartTime used for deprovisionJobsDurationHistogram metrics
+	deprovisionStartTime := time.Now()
+	fmt.Printf("Starting time:%v", deprovisionStartTime)
+	fmt.Printf("Unix timestamp:%v", deprovisionStartTime.Unix())
+	labels := map[string]string{UninstallJobLabel: "true", DeprovisionStartTimeLabel: strconv.FormatInt(deprovisionStartTime.Unix(), 10)}
```
dgoodwin: No need for a label, Job.Status has both StartTime and CompletionTime.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -706,6 +724,18 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
...
 			cdLog.Infof("uninstall job successful, removing finalizer")
+			// duration calculates the time elapsed since the uninstall job started for deprovision jobs that have the DeprovisionStartTimeLabel
+			if startTime, ok := existingJob.Labels[install.DeprovisionStartTimeLabel]; ok {
+				convertedTime, err := strconv.ParseInt(startTime, 10, 64)
+				if err != nil {
+					cdLog.Errorf("Could not parse the unix timestamp")
+				} else {
+					currentTime := time.Now().Unix()
+					duration := currentTime - convertedTime
+					cdLog.Infof("uninstall job duration:%v", duration)
```
dgoodwin: You should use the job.Status.StartTime and CompletionTime, currentTime may not be right as the controllers might be down for some time, and we are getting this event later on.

cdLog.WithField("duration", jobDuration).Debug("uninstall job completed")

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -73,6 +75,22 @@ const (
 	clusterImageSetFoundReason            = "ClusterImageSetFound"
 )
 
+var (
+	deprovisionJobsDurationHistogram = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "deprovision_jobs_duration_seconds",
+			Help: "Deprovision Jobs duration distribution",
+			//Buckets: []float64{0.5, 1, 2, 4, 8, 16},
+			Buckets: prometheus.ExponentialBuckets(1, 2, 9),
```
dgoodwin: Can you help me understand these values and how this works?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -73,6 +75,22 @@ const (
...
+		},
+		[]string{"uninstal_job", "namespace"},
```
dgoodwin: Do we want vectors on this one? This metric is used to calculated average uninstall times, but only triggers when it completes successfully. For this one I think we can go without the vector, but let me know if I'm thinking about it wrong.

ShivamChamoli: Actually I thought in the future we may have different namespaces that might have uninstall jobs, if that is not the case then we should not have a vector.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
ShivamChamoli: It creates buckets with the start value 1, step 2 and count 9, so basically buckets that correspond to <1 seconds to <256 seconds

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: I'm really not sure if they make sense here and I'm too new to prometheus, it would be one job per namespace in real world, does that make sense in a histogram? I assume we can average without vectors to get the real data we're after, but I'm not sure if we'd ever want to get an average for these buckets which would always probably be of size one. I think dropping them seems correct here.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Hmm interesting, could you get us three buckets, jobs less than 60 seconds, jobs less than 10 minutes, and anything over that?

ShivamChamoli: yes definitely

dgoodwin:
I am not quite sure what you did on this one. However the commits you want are:

4035fcd3054a9bf8ebddb690dd14b64f7dcc6073 
05b52607e80318943f4e6e3598d62473e4f267e5
11e365fb2ab4933e795214b67c1e5902334f7b75

So lets start over.

git fetch up
git reset --hard up/master (be careful with this one, don't use this in future unless you know what it's doing, which is trashing the current branch and making it equal to up/master)
git cherry-pick [each commit above]
squash into one commit, then git push -f

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -361,6 +370,7 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		} else {
 			cdLog.Debug("provision job exists")
 			r.checkInstallPodRetries(cd, cdLog)
+			cdLog.Infof("cluster job exists, cluster deployment status: %v", cd.Status.Installed)
```
dgoodwin: Drop this line, I cleaned it up recently on line 371.

PR #305, opened 2019-04-10
Deliver a hive-admin role via the operator.

dgoodwin:
This role allows the basic cluster wide permissions we expect to need to
administer a hive cluster.

It grants read access to all the Hive types, as well as jobs, pods, and
pod logs.

It grants write access to the HiveConfig.

It does not grant access to secrets or configmaps.

This is delivered via the operator itself, rather than the CSV because
we need to bind to a group, not a service account.

### DIFF
--- a/config/rbac/hive_admin_role.yaml
+++ b/config/rbac/hive_admin_role.yaml
@@ -1,0 +1,51 @@
+# hive-admin is a role intended for hive administrators who need to be able to debug
+# cluster installations, and modify hive configuration.
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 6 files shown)


dgoodwin:
cc @jmelis @jewzaam @pbergene this is what we're delivering for the hive team to be able to effectively debug, I think it should be a good start. Please make sure you're aware what we're granting ourselves.

We will need the group delivered separately, it should be called "hive-admins"

dgoodwin:
May as well add them, I will update.

dgoodwin:
Added dnsendpoints.

PR #307, opened 2019-04-10
Install Pod Restart Metric + Status

dgoodwin:
Builds on @ShivamChamoli 's PR and adds storing the count on status as well.


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -902,6 +925,61 @@ func dnsZoneName(cdName string) string {
...
+func selectorPodWatchHandler(a handler.MapObject) []reconcile.Request {
+	retval := []reconcile.Request{}
+
+	pod := a.Object.(*corev1.Pod)
+	if pod == nil {
+		// Wasn't a Pod, bail out. This should not happen.
+		log.Errorf("Error converting MapObject.Object to Pod. Value: %+v", a.Object)
+		return retval
+	}
+	if pod.Labels != nil {
```
csrwng: just a style suggestion (feel free to ignore) -- exit early here
if pod.Labels == nil {
     return retval
}

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -902,6 +925,61 @@ func dnsZoneName(cdName string) string {
...
+		if cdName, ok := pod.Labels[install.ClusterDeploymentNameLabel]; ok {
```
csrwng: same as above ... !ok should return early

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -902,6 +925,61 @@ func dnsZoneName(cdName string) string {
...
+
+	if len(pods.Items) > 1 {
+		log.Warnf("found %d install pods for cluster", len(pods.Items))
+	}
+
+	// Calculate restarts across all containers in the pod:
+	podRestarts := 0
+	for _, pod := range pods.Items {
+		for _, cs := range pod.Status.ContainerStatuses {
+			podRestarts = podRestarts + int(cs.RestartCount)
```
csrwng: +=

dgoodwin:
Updated. That integration test is failing a bit too much.

csrwng:
@dgoodwin the e2e failure seems legit... the job is never created by the controller which tells me that the controller might be failing to start. Most likely due to a permissions issue. Did we add pods to the manager role?

dgoodwin:
Nope I did not, thank you, should have noticed that was failing too much, will check if there's an rbac issue.

csrwng:
@dgoodwin looks like you need to run `make generate` again

csrwng:
nm, that passed
/lgtm
/retest

PR #313, opened 2019-04-12
ResourceHelper improvements

csrwng:
- Apply now reports whether the call resulted in a create, modification, or no change
- Makes creating a helper from an existing rest.Config more robust by using the rest.Config directly instead of first serializing to a kubeconfig
- Fixes a bug with serialization that resulted in every apply always calling patch (because of `creationTimestamp: null` being present)

### DIFF
--- a/pkg/resource/apply.go
+++ b/pkg/resource/apply.go
@@ -18,70 +18,83 @@ package resource
 
 import (
 	"bytes"
+	"io"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericclioptions/printers"
 	kcmd "k8s.io/kubernetes/pkg/kubectl/cmd"
 	cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 )
 
-// Apply applies the given resource bytes to the target cluster specified by kubeconfig
-func (r *Helper) Apply(obj []byte) error {
+// ApplyResult indicates what type of change was performed
+// by calling the Apply function
+type ApplyResult string
 
+var (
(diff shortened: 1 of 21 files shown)


pkg/resource/serializer.go
```diff
@@ -0,0 +1,100 @@
...
+// `timeCreated: null` which always causes a mismatch with whatever's already in the server.
+func (r *Helper) Serialize(obj runtime.Object, scheme *runtime.Scheme) ([]byte, error) {
+	printer := printers.NewTypeSetter(scheme).ToPrinter(&jsonPrinter{})
+	buf := &bytes.Buffer{}
+	if err := printer.PrintObj(obj, buf); err != nil {
+		r.logger.WithError(err).Error("cannot serialize runtime object of type %T", obj)
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
```
dgoodwin: Are these two new files our code or copied?

pkg/resource/apply.go
```diff
@@ -18,70 +18,74 @@ package resource
...
-// Apply applies the given resource bytes to the target cluster specified by kubeconfig
-func (r *Helper) Apply(obj []byte) error {
+type ApplyResult string
+
+var (
+	ConfiguredApplyResult ApplyResult = "configured"
+	UnchangedApplyResult  ApplyResult = "unchanged"
+	CreatedApplyResult    ApplyResult = "created"
+	UnknownApplyResult    ApplyResult = "unknown"
+)
```
dgoodwin: This is fantastic, I actually had a Jira card to add this someday.

pkg/resource/serializer.go
csrwng: They're our code, with the exception of the computeDiscoverCacheDir function that I copied over from clientcmd

pkg/resource/apply.go
csrwng: It led me to find out that we were always patching :)

csrwng:
fixed verify error, integration test, hopefully it'll pass now

PR #315, opened 2019-04-12
Add metrics for AWS API client calls.

dgoodwin:


pkg/awsclient/client.go
```diff
@@ -46,6 +48,16 @@ const (
 	awsCredsSecretAccessKey = "aws_secret_access_key"
 )
 
+var (
+	metricAWSAPICalls = prometheus.NewCounterVec(
```
csrwng: @dgoodwin should this be registered somewhere?

dgoodwin: LOL yes thank you. My bad.

dgoodwin: Updated.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -86,7 +86,7 @@ func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 // AddToManager adds a new Controller to mgr with r as the reconcile.Reconciler
 func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
-	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
+	c, err := controller.New("controlplanecerts-controller", mgr, controller.Options{Reconciler: r})
```
csrwng: thx!

PR #319, opened 2019-04-16
save installer output to configmap

joelddiaz:
run the installer without debug loglevel and capture/redirect the regular console output into a file. save it to a configmap (on error f the generateasset phase, and always in the provisioncluster stage).

pipe out the output of the full installer logs $workdir/.openshift_install.log as the binary/pods output.

add some retry logic around updating the clusterdeployment status (lots of controllers updating the clusterdeployment causing stale object update errors).

update tests to simulate outputing of console output to file, and verify it getting saved to the configmap.

### DIFF
--- a/contrib/pkg/installmanager/installmanager.go
+++ b/contrib/pkg/installmanager/installmanager.go
@@ -1,7 +1,6 @@
 package installmanager
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -30,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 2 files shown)


contrib/pkg/installmanager/installmanager.go
```diff
@@ -553,6 +620,36 @@ func uploadAdminPassword(cd *hivev1.ClusterDeployment, m *InstallManager) (*core
 	return s, uploadSecretWithRetries(s, m)
 }
 
+func uploadConfigmapWithRetries(c *corev1.ConfigMap, m *InstallManager) error {
+	uploaded := false
+	var err error
+
+	// try 10 times
+	for i := 1; i < 11; i++ {
```
csrwng: Suggestion: use k8s.io/apimachinery/pkg/util/wait 
```
backoff := retry.DefaultBackoff
backoff.Steps = 10
backoff.Duration =  1 * time.Second
wait.ExponentialBackoff(backoff, func() error { ... })
```
If you want to make the retry time constant, substitute retry.DefaultRetry for retry.DefaultBackoff

contrib/pkg/installmanager/installmanager.go
```diff
@@ -620,29 +620,24 @@ func uploadAdminPassword(cd *hivev1.ClusterDeployment, m *InstallManager) (*core
...
+	backoff.Duration = time.Second
 
-	// try 10 times
-	for i := 1; i < 11; i++ {
-		err = m.DynamicClient.Delete(context.Background(), c)
-		if err != nil && !errors.IsNotFound(err) {
-			m.log.WithError(err).WithField("configmap", c.Name).Warningf("error deleting existing configmap")
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		if err := m.DynamicClient.Delete(context.Background(), c); err != nil && !errors.IsNotFound(err) {
+			m.log.WithError(err).WithField("configmap", c.Name).Warning("error deleting existing configmap")
```
csrwng: @joelddiaz shouldn't you return at this point if the error is something other than not found?

joelddiaz: i suppose that Create() below doesn't stand much of a chance :) updated PR

PR #321, opened 2019-04-16
Provision Job duration metric

ShivamChamoli:
--> Calculates provision job duration metrics

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -93,11 +93,19 @@ var (
 			Buckets: []float64{60, 3600, 7200},
 		},
 	)
+	provisionJobsDurationHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "provision_jobs_duration_seconds",
+			Help:    "Provision Jobs duration distribution",
+			Buckets: []float64{1, 10, 30, 60},
+		},
+	)
 )
 
 func init() {
 	metrics.Registry.MustRegister(metricClusterDeploymentInstallRestarts)
 	metrics.Registry.MustRegister(deprovisionJobsDurationHistogram)
+	metrics.Registry.MustRegister(provisionJobsDurationHistogram)
 }
 
 // Add creates a new ClusterDeployment Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -295,6 +303,8 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		}
 	}
 
+	// reportDurationMetricsFlag is the flag that is used for reporting the provision job duration metric
+	reportDurationMetricsFlag := false
 	// Check if an install job already exists:
 	existingJob := &batchv1.Job{}
 	installJobName := install.GetInstallJobName(cd)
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -630,6 +646,15 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
 
+	cdLog.Infof("REPORT DURATION METRIC FLAG:%v", reportDurationMetricsFlag)
+	// Report the cluster metrics if we need to
+	if reportDurationMetricsFlag {
+		cdLog.Infof("SETTING METRIC")
+		// jobDuration calculates the time elapsed since the install job started
+		jobDuration := job.Status.CompletionTime.Time.Sub(job.Status.StartTime.Time)
+		cdLog.WithField("duration", jobDuration.Seconds()).Debug("install job completed")
+		provisionJobsDurationHistogram.Observe(float64(jobDuration.Seconds()))
+	}
```
dgoodwin: I would move this block back into the original function right before "reconcile complete" just because there's no need to pass that boolean state around to a function that might get re-used elsewhere.

This looks fine however, it should work, I think you need some more testing, if the update fails it should immediately re-sync and try everything again.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -305,6 +315,11 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	} else if err != nil {
 		cdLog.WithError(err).Error("error looking for install job")
 		return reconcile.Result{}, err
+	} else {
```
dgoodwin: No need for the else given the return above.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -384,6 +399,14 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	if err != nil {
 		cdLog.WithError(err).Errorf("error updating cluster deployment status")
 		return reconcile.Result{}, err
+	} else {
```
dgoodwin: Can drop this else as well.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -93,11 +93,19 @@ var (
 			Buckets: []float64{60, 3600, 7200},
 		},
 	)
+	provisionJobsDurationHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "provision_jobs_duration_seconds",
+			Help:    "Provision Jobs duration distribution",
+			Buckets: []float64{60, 3600, 7200},
```
dgoodwin: For install buckets could be a little diff, could you adjust these for 1m, 10m, 30m, 60m

pkg/controller/clusterdeployment/clusterdeployment_controller.go
ShivamChamoli: what if it goes inside if err != nil && errors.IsNotFound(err)  statement on line 312 and sets existingJob to nil? Now if we don't have the else and if we only have the if existingJob.Status.Succeeded > 0 && !cd.Status.Installed {
			reportDurationMetricsFlag = true
} statement, wont that cause an error when it tries to resolve existingJob.Status.Succeeded since existingJob is nil?

dgoodwin: My mistake you are correct.

PR #322, opened 2019-04-17
Hiveutil create-cluster command to create a clusterdeployment

csrwng:


dgoodwin:
Really cool @csrwng thanks!

/lgtm

PR #325, opened 2019-04-18
Add cluster type concept and use it to partition metrics.

dgoodwin:
Adds an optional, freeform label that can be added to clusters, hive.openshift.io/cluster-type. Displayed in short "kubectl get" output if set. Value defaults to "unspecified" if not set. Also copied onto jobs, if set.

Cleanup metrics code to stop flipping from over/under terminology and instead use consistent gt or lt vectors to better match what histograms would do for buckets. Added clusterAccumulator to encapsulate this logic, and more dynamically handle the durations that define the buckets we use for length of time a cluster has been uninstalled.

Added partitioning for metrics based on cluster type, and the age of the cluster, we now report a set of metrics only for clusters created in the last 24h.

Renamed some metrics after reading some best practices, _total should be reserved for counters.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -180,6 +180,7 @@ type ReconcileClusterDeployment struct {
 // +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterimagesets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterimagesets/status,verbs=get;update;patch
 func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+
 	// Fetch the ClusterDeployment instance
 	cd := &hivev1.ClusterDeployment{}
 	err := r.Get(context.TODO(), request.NamespacedName, cd)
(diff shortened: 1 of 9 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -180,6 +180,7 @@ type ReconcileClusterDeployment struct {
 // +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterimagesets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterimagesets/status,verbs=get;update;patch
 func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+
```
twiest: I'm fine with this, but this feels like it was accidental. Do you really want this?

pkg/controller/metrics/metrics_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"testing"
 	"time"
 
-	log "github.com/sirupsen/logrus"
+	//log "github.com/sirupsen/logrus"
```
twiest: Do we still want this line?

dgoodwin: Updated, thanks!

pkg/install/generate.go
```diff
@@ -257,9 +259,15 @@ func GenerateInstallerJob(
...
 	}
 
-	labels := map[string]string{InstallJobLabel: "true"}
-	// installerPodLabels are the labels that are used by metricClusterDeploymentInstallRetries
-	installerPodLabels := map[string]string{ClusterDeploymentNameLabel: cd.Name, InstallJobLabel: "true"}
+	labels := map[string]string{
+		InstallJobLabel:            "true",
+		ClusterDeploymentNameLabel: cd.Name,
+	}
+	typeStr, ok := cd.Labels[hivev1.HiveClusterTypeLabel]
```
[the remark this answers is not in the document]
dgoodwin: I'm not sure, which probably means it should, I will update.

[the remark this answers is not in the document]
dgoodwin: Added a guard as if the cluster doesn't have the label, the job probably shouldn't either.

PR #328, opened 2019-04-22
WIP: force requeue after fixing status fields

joelddiaz:
if we return empty reconcile.Result, nil, then we won't get requeued for the regular object requeue period (15 minutes?)

after fixing the empty fields, we should go ahead and requeue immediately to avoid any delay in processing the clusterdeployment


joelddiaz:
nevermind, looking at the kibana logs more closely indicates that this change is unnecessary

PR #336, opened 2019-04-23
Add logging for when reconciliation loops start and finish.

twiest:


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -183,24 +183,31 @@ type ReconcileClusterDeployment struct {
...
 	start := time.Now()
+	cdLog := log.WithFields(log.Fields{
+		"clusterDeployment": request.Name,
+		"namespace":         request.Namespace,
+		"controller":        controllerName,
+	})
+
+	// For logging, we need to see when the reconciliation loop starts and ends.
+	cdLog.Info("Started reconciling cluster deployment")
+	defer cdLog.Info("Finished reconciling cluster deployment")
```
dgoodwin: This looks redundant with line 214 which includes the elapsed time. Recommend adding that for all of these as it's really handy to watch for long loops.

twiest: I prefer using `defer` as that catches all cases where the reconcile is completed. For instance, the "IsNotFound" and the "Error getting object" cases.

Is there a way we can have the elapsed time in the defer?

dgoodwin: Yeah defer is definitely better here, so if you use an inline func to delay processing the current time it should work, something like (freehand):

defer cdLog.WithField("elapsed", func() time.Duration{
return time.Since(start)
}()).Infof("blah")

dgoodwin: Sorry I think I misclicked resolve conversation somewhere along the way.

twiest: @dgoodwin ok, I updated the code. Does this look acceptable to you?

PR #340, opened 2019-04-24
Attempted fix for long (5-20m) delays creating install jobs.

dgoodwin:
Switches from serially processing 1 reconcile at a time to 100 for most controllers, it might be possible to go to thousands here but this is just a first test. Want to see how this behaves in stage.

Moved creation of a remote client in cluster deployment controller into a guarded block where it's actually used. We were doing this even if we didn't need it for every cluster.

Remove CPU and memory limits per advice from @jeremyeder .

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -29,11 +29,8 @@ spec:
         imagePullPolicy: Always
         name: manager
         resources:
-          limits:
-            cpu: 100m
-            memory: 1Gi
           requests:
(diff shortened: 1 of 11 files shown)


dgoodwin:
I've also filed us a card to add a controller that manages an "unreachable" condition for each cluster. If true we can update all other controllers that use a remote client to skip processing if the condition is true.

jeremyeder:
Thanks for removing the limits.

PR #344, opened 2019-04-25
Remove generation from container build process.

dgoodwin:
This is now failing because master for gomock appears to require go
1.11.

However we don't need to generate code for a Docker build, assume we're
building a container with the latest committed code. A developer using
containers for their work will need to run make generate before building
their container.


dgoodwin:
Force merging, we need to get these Dockerfiles to master so CI will pick them up and we can sort this mess out.

PR #345, opened 2019-04-25
WIP:imageset job metrics

ShivamChamoli:
imageset job metrics

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -90,14 +90,14 @@ var (
 	)
 	deprovisionJobsDurationHistogram = prometheus.NewHistogram(
 		prometheus.HistogramOpts{
-			Name:    "deprovision_jobs_duration_seconds",
+			Name:    "hive_deprovision_jobs_duration_seconds",
 			Help:    "Deprovision Jobs duration distribution",
 			Buckets: []float64{60, 3600, 7200},
 		},
 	)
 	provisionJobsDurationHistogram = prometheus.NewHistogram(
 		prometheus.HistogramOpts{
-			Name:    "provision_jobs_duration_seconds",
+			Name:    "hive_provision_jobs_duration_seconds",
 			Help:    "Provision Jobs duration distribution",
 			Buckets: []float64{1, 10, 30, 60},
 		},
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -109,13 +109,29 @@ var (
 			Buckets: []float64{30, 60, 120, 300, 600, 1200, 1800},
 		},
 	)
+	imagesetJobsDurationHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "imageset_jobs_duration_seconds",
```
dgoodwin: All our metrics should be prefixed with hive_. Could you update this and the ones above that aren't?

ShivamChamoli:
--> Kickstart is giving a weirdly high number (>600) although it launches almost immediately
--> Not sure why the two gauges in metrics.go are always 0
--> As for the duration of the job I need to return a boolean from resolveInstallerImage function which tells me that the job has completed, is it ok to change the function signature to send a boolean along with an error

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -571,6 +587,12 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 		err = r.Create(context.TODO(), job)
 		if err != nil {
 			jobLog.WithError(err).Error("error creating job")
+		} else {
+			// kickstartDuration calculates the delay between creation of cd and start of imageset job
+			kickstartDuration := time.Since(cd.CreationTimestamp.Time)
+			cdLog.Infof("KICKSTART:%v", kickstartDuration.Seconds())
```
dgoodwin: Drop this line please.

pkg/controller/metrics/metrics.go
```diff
@@ -210,6 +222,32 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 			}
 		}
 
+		mcLog.Info("IMAGESET JOBS")
```
dgoodwin: Drop this line please.

pkg/imageset/generate.go
```diff
@@ -47,6 +47,11 @@ else
   echo "0" > /common/success
 fi
 `
+	// ImagesetJobLabel is the label used for counting the number of imageset jobs in Hive
+	ImagesetJobLabel = "hive.openshift.io/imageset"
+
+	// ClusterDeploymentNameLabel is the label (along with ClusterDeploymentNamespaceLabel) that is used to identify the imageset pod of a particular cluster deployment
```
dgoodwin: Can drop the note about ClusterDeploymentNamespaceLabel as we dropped that.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: So this is showing over 600 despite it starting immediately? hmmm. I can't see how that would happen...

dgoodwin:
> --> Kickstart is giving a weirdly high number (>600) although it launches almost immediately
> --> Not sure why the two gauges in metrics.go are always 0

I have a suspicion this is because they get cleaned up once they're successful, so you would only see  them if you listed metrics *while* the job is running. This is probably working as intended.

> --> As for the duration of the job I need to return a boolean from resolveInstallerImage function which tells me that the job has completed, is it ok to change the function signature to send a boolean along with an error

Little confused, can you clarify why you need to return a boolean for if the job has completed or not?

PR #347, opened 2019-04-26
Use `+genclient` for the configuration object

jhernand:
The description of the `HiveConfig` type has the
`+genclient:nonNamespaced` comment, but not `+genclient`. Both need to
be present, otherwise the code generation tools don't generate a client
for the type.


jhernand:
@dgoodwin @nimrodshn please review.

nimrodshn:
@jhernand LGTM :+1:

PR #350, opened 2019-04-29
Using dep for gomock and mockgen

LalatenduMohanty:

## Steps to test this PR

* Delete the `mockgen` binary from from $GOPATH/bin
* Delete the `gomock` package from $GOPATH/src/github.com/golang/mock/
* Run `make`


LalatenduMohanty:
@dgoodwin  Here are the steps to test this PR

* Delete the `mockgen` binary from from $GOPATH/bin
* Delete the `gomock` package from $GOPATH/src/github.com/golang/mock/
* Run `make`

Makefile
```diff
@@ -188,6 +195,9 @@ buildah-push: buildah-build
 install-federation:
 	./hack/install-federation.sh
 
+$(GOPATH)/bin/mockgen:
+	go get -u github.com/golang/mock/mockgen/...
```
dgoodwin: It looks like the vendored code wasn't commited.

What does go get do now in this case? I assume it knows to use the vendored code?

LalatenduMohanty: The vendor directory was already present as we had the change in Gopkg.lock file somehow. Refer https://github.com/openshift/hive/tree/master/vendor/github.com/golang/mock

dgoodwin:
This gets us partway there in that we now fail in the makefile rather than in the dockerfile, which CI reads from master and requires for pushes to fix in a PR.

Lala is looking into vendoring mockgen for a folloup.

/lgtm

PR #351, opened 2019-04-29
avoid needlessly making a non-nil map

joelddiaz:
unconditionally creating the non-nil map was causing the equality check against the remote machinesets to always detect a difference when there are no lables on the machine set. check if we will need to populate the map before allocating he map.

also pull in some unrelated generated mockgen changes


pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -360,7 +360,9 @@ func (r *ReconcileRemoteMachineSet) generateMachineSetsFromClusterDeployment(cd
 			hivePool := findHiveMachinePool(cd, workerPool.Name)
 			for _, ms := range icMachineSets {
 				// Apply hive MachinePool labels to MachineSet MachineSpec.
-				ms.Spec.Template.Spec.ObjectMeta.Labels = map[string]string{}
+				if hivePool.Labels != nil {
+					ms.Spec.Template.Spec.ObjectMeta.Labels = map[string]string{}
+				}
```
dgoodwin: Nice work, thank you for tracking this down.

PR #357, opened 2019-05-01
allow modifying certificateBundles and controlPlaneConfig.servingCertificates

joelddiaz:
when adding a new ingress, it is possible that the new ingress entry will need to reference a new spec.certificiateBundles[] entry. mark certificatesBundles as mutable.

after installing in default mode (ie self-signed certs) cluster, it is reasonable to then migrate to signed certificates which will require making controlPlaneConfig.servingCertificates.* as mutable.

also pull in some unrelated mockgen changes

### DIFF
--- a/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
+++ b/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
@@ -52,7 +52,7 @@ const (
 )
 
 var (
-	mutableFields = []string{"Compute", "Ingress", "PreserveOnDelete"}
+	mutableFields = []string{"CertificateBundles", "Compute", "ControlPlaneConfig", "Ingress", "PreserveOnDelete"}
 )
 
 // ClusterDeploymentValidatingAdmissionHook is a struct that is used to reference what code should be run by the generic-admission-server.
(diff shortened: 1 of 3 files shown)


joelddiaz:
@csrwng what do you think of the possibility of being able to remove/clear out controlPlaneConfig.servingCertificates? is this okay?

i have custom ingress validation code to keep someone from defining an ingress entry, and then trying to remove all of them. do we need something similar to protect opting out of controlPlane certificates?

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -327,6 +327,21 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateUpdate(admissionSpec
 		}
 	}
 
+	// Check to make sure only allowed fields under controlPlaneConfig are being modified
+	hasChangedImmutableControlPlaneConfigFields := hasChangedImmutableControlPlaneConfigFields(&oldObject.Spec, &newObject.Spec)
```
csrwng: seems like an unnecessary variable: `if hasChangedImmutableControlPlaneConfigFields(...) {`
should be enough

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -364,6 +379,33 @@ func hasChangedImmutableField(oldObject, newObject *hivev1.ClusterDeploymentSpec
...
+	newCopy.ServingCertificates = hivev1.ControlPlaneServingCertificateSpec{}
+
+	if !reflect.DeepEqual(origCopy, newCopy) {
+		return true
+	}
+
+	return false
+}
+
+func getOriginalMachinePool(origMachinePools []hivev1.MachinePool, name string) *hivev1.MachinePool {
```
csrwng: Is this used somewhere?

joelddiaz: yeah, in a completely different PR https://github.com/openshift/hive/pull/356 :) will remove it

csrwng:
> @csrwng what do you think of the possibility of being able to remove/clear out controlPlaneConfig.servingCertificates? is this okay?

It should be possible. It should revert to self-signed certs.

csrwng:
/lgtm
@joelddiaz if this is ready, remove the WIP

PR #359, opened 2019-05-03
Adding steps to watch cluster deployment and retrieve password for webconsole

LalatenduMohanty:


README.md
```diff
@@ -117,6 +125,17 @@ oc process -f config/templates/cluster-deployment.yaml \
    | oc apply -f -
 ```
 
+### Watch the ClusterDeployment
+
+* Get the current namespace in which the `oc process` command was executed
+* Get the install pod name
+  ```
+  $ kubectl get pods -n <namespace> | grep install
```
LalatenduMohanty: Will replace this with a `jq` command.

dgoodwin: They have a label that ties them straight to the cluster deployment name as well, so you could use -l to get just that pod and then jq

PR #362, opened 2019-05-06
Fix imageset job hotloop.

dgoodwin:
In the event of a error running the oc adm release command to resolve
the installer image to use, the script does not return an error, nor
does the hiveutil sidecar pod, and the controller sees the job finished,
no installer image set, so it recreates a new job.

This leads to a hotloop where in the event of a long lived failure, the
controllers are deleting and recreating new jobs and pods as quickly as
possible, which leads to excessive reconciliation results in the
controllers.

To fix this change just returns a non-zero exit code when the release
extraction command fails. This causes the job to see an failure and
re-try with backoff. If we run up against the job expiration the
controllers should then see a finished job, no installer image, and
create a new one as they do now.


dgoodwin:
@csrwng does this look ok to you?

csrwng:
/lgtm
Nice fix. Thx!

PR #364, opened 2019-05-06
Add ClusterDeprovisionRequest CRD and controller

csrwng:
Introduces new ClusterDeprovisionRequest CRD. When present, the controller will create an uninstall job for it and update its status when the job is complete.


csrwng:
Needs:
- [ ] unit-tests
- [ ] logging
- [ ] metrics

csrwng:
This should be ready for review

dgoodwin:
cc @jhernand

jhernand:
@vkareh please check review and make sure that the API is usable for us.

@csrwng @dgoodwin For clusters that have a `clusterdeployment` object can we use this as well? Or should we keep deleting the `clusterdeployment`? I don't see any reference from the new CRD to the `clusterdeployment` CRD, so I assume that we will still need to delete it. Please confirm/reject.

jhernand:
@cshinn @dgoodwin @vkareh We will also need permissions to create/list/get/watch this new kind of object in the `hive-frontend` role.

vkareh:
@jhernand - seems like there is a new `GenerateUninstallerJobForClusterDeployment` that allows for uninstalling by referencing the ClusterDeployment, but seems to happen during Reconcile, so most likely we will still delete the CD and Hive will trigger the uninstaller job.

This CRD seems like it's exactly what we need for deprovisioning self-managed clusters. I can test locally and see how this fits with our code.

I agree that we still need frontend permissions for this.

pkg/install/generate.go
```diff
@@ -360,13 +426,15 @@ func GenerateUninstallerJob(
 				"--loglevel",
 				"debug",
 				"--region",
-				cd.Spec.AWS.Region,
-				fmt.Sprintf("kubernetes.io/cluster/%s=owned", cd.Status.InfraID),
-				// Also cleanup anything with the tag for the legacy cluster ID (credentials still using this for example)
-				fmt.Sprintf("openshiftClusterID=%s", cd.Status.ClusterID),
```
dgoodwin: I don't think we can remove openshiftClusterID here, just did a quick check and registry s3 buckets are still being created only with this tag. It probably wouldn't break deprovision, but it would leak buckets without. Shame as it would be nice to be able to drop that cluster ID from the API.

pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	rLog.Debug("uninstall job exists, checking its status")
+
+	// Uninstall job exists, check its status and if successful, set the deprovision request status to complete
+	if controllerutils.IsSuccessful(existingJob) {
+		rLog.Infof("uninstall job successful, setting completed status")
+		// jobDuration calculates the time elapsed since the uninstall job started for deprovision job
+		jobDuration := existingJob.Status.CompletionTime.Time.Sub(existingJob.Status.StartTime.Time)
+		rLog.WithField("duration", jobDuration.Seconds()).Debug("uninstall job completed")
+		hivemetrics.MetricUninstallJobDuration.Observe(float64(jobDuration.Seconds()))
```
dgoodwin: I would report the metric right after you successfully update the request, this would prevent duplicating it in the event we can't save for some reason.

csrwng:
@jhernand @vkareh this PR already adds permission for this new resource to the hive-frontend role:
https://github.com/openshift/hive/pull/364/files#diff-da0e8ba1faea8890ca5371c281d77a24R50

@jhernand you will still need to continue deleting clusterdeployments. This does not replace that. Actually what we will do in phase 2 is that we will implement the clusterdeployment deletion through a clusterdeprovisionrequest. One thing we discussed yesterday was how to make clusterdeployment deletion safer and concluded that it may be better to remove the delete permission from most roles. Instead, the way to delete them would be by setting a "DeleteAfter" field on the clusterdeployment which will tell the controller to delete it by a certain time. (See https://jira.coreos.com/browse/CO-388)

config/rbac/hive_frontend_role.yaml
```diff
@@ -47,6 +47,7 @@ rules:
   - syncidentityproviders
   - selectorsyncsets
   - syncsets
+  - clusterdeprovisionrequests
```
dgoodwin: @jhernand @vkareh new requests API being added to frontend role here.

jhernand: Great, thanks @dgoodwin.

pkg/install/generate.go
csrwng: we're not removing it, but only optionally adding it if not present

csrwng: The thinking is that in some deprovision requests you may not have that available anymore. You should at least be able to remove most resources.

pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
csrwng: ack, makes sense

pkg/install/generate.go
dgoodwin: My mistake I missed it below, this sounds perfect. Should we make ClusterID a pointer to indicate it's optional nature?

csrwng:
@dgoodwin thx for the review, fixes submitted

pkg/install/generate.go
csrwng: We can ... it would be a compatible change for the future, but simply leaving it blank does the trick as well.

PR #365, opened 2019-05-07
Fix PreserveOnDelete clusters rapidly deleted after creation.

dgoodwin:
A bug has surfaced in stage where a UHC test was creating unmanaged
(PreserveOnDelete) clusters and then immediately deleting them, and
their namespace. The controllers still tried to create service account /
role / rolebinding for the pods, even though we have no pod to run for
these clusters, and we just want to clear the finalizer.

I considered trying to shortcut this but the deletion logic is already
pretty tricky, particularly when you consider a PreserveOnDelete
cluster, with an InfraID, which is not yet fully installed, which is an
edge case where we do actually launch an uninstall job for it. However
this does not appear to be the case in stage as they seem to be deleted
before we can get an InfraID generated.

Instead this PR moves the creation of these objects next to the creation
of their jobs. (3 places) This keeps the logic from getting any more
complicated than it already is.

NOTE: if you do create a PreserveOnDelete cluster, it gets far enough
along to have an InfraID, but you delete it before it's installed and
then immediately delete it's namespace, this scenario will still occur
and require manual intervention to clear the finalizer.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -270,12 +270,6 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		}, nil
 	}
 
-	_, err := r.setupClusterInstallServiceAccount(cd.Namespace, cdLog)
-	if err != nil {
-		cdLog.WithError(err).Error("error setting up service account and role")
-		return reconcile.Result{}, err
-	}
-
 	imageSet, err := r.getClusterImageSet(cd, cdLog)
 	if err != nil {
 		return reconcile.Result{}, err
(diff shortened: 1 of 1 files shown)


dgoodwin:
CC @jhernand this should help clear the error rates in stage.

csrwng:
@dgoodwin for the install/uninstall jobs, maybe calling the setupClusterInstallServiceAccount right before the job is created would be better. That way you don't call it every time you check on a running job. It also makes me realize that I'm not calling this from the DeprovisionRequest controller and really should be.

The other thought as I was looking at the code for setupClusterInstallServiceAccount is that we're not checking that the role/rolebinding have the correct spec if we find them. We should likely update them if they're not what we think they should be. When we do call Create, we should ignore an AlreadyExists error, given that we may have gotten incorrect info from the cache.

dgoodwin:
Updated, good catch I didn't see that was still doing it for every reconcile.

Would like to get those error rates down tonight so I'll do a followup to improve tomorrow to create and update, maybe use the resource applier.

PR #368, opened 2019-05-08
Validate SyncSet PatchType

csrwng:
Adds validating webhooks for syncsets and selectorsyncsets. Rejects patches with invalid patch types.


csrwng:
@abutcher @dgoodwin ptal

PR #370, opened 2019-05-08
Migrate patch type for SyncSets with patches

csrwng:
Migrates previous patchType strings to the new short format


csrwng:
@abutcher I've now tested this manually. It should be ready for review

PR #375, opened 2019-05-09
Cluster deprovision request - phase 2

csrwng:
Adds spec.state field to clusterdeployment with possible values 'Absent' and 'Present'
When the state is set to 'Absent', the clusterdeployment is deleted.

Adds validation for the field in the validation webhook and defaulting for it in a new mutating webhook.

Switches the clusterdeployment controller to use ClusterDeprovisionRequests to delete clusterdeployments instead of itself generating a job.

Removes ActiveDeadlineSeconds from install/uninstall jobs, making them run forever.


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -282,6 +287,16 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
 
+	if cd.Spec.State == hivev1.ClusterDeploymentStateAbsent {
+		cdLog.Info("Absent state detected on clusterdeployment. Deleting...")
+		err := r.Delete(context.TODO(), cd)
+		if err != nil {
+			cdLog.WithError(err).Errorf("cannot delete clusterdeployment")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
```
dgoodwin: If I'm reading this right, it looks like if we roll this out while UHC is still just deleting the CD, their code will no longer work and clusters will start piling up.

I think we need an interim hack here where if we see a deletionTimestamp, but state!=absent, then assume to set the state=absent so we get our deprovisioning as before. Then we need to file a card with SDA, wait for the fix to be in stage and prod, then we can kill off the hack and assume everyone is deleting properly.

csrwng: No, that is not the case. You can still delete cluster deployments and they will go away. The only thing that has changed is how we go about deleting them.

dgoodwin: Ah I see so it's state=absent = assumed cd delete, which is what creates the request.

So I guess eventually there would be a phase 3 where you must set state = absent, and we wait until UHC is ready.

csrwng: Kind of... phase 3 is just removing rbac permission to delete clusterdeployments to every role except the hive manager role.

csrwng:
This needs to be rebased... pushing rebased commit shortly

PR #376, opened 2019-05-10
Give up on deprovisioning in deleted namespaces.

dgoodwin:
A fix is coming to UHC next week to prevent namespaces from being deleted before the cluster deployment is gone.

However in meantime, and in case this happens again, we need to watch for the error and if encountered, give up on deprovision, as there's little we can do. This change removes the finalizer and lets the cluster go away.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -219,7 +219,7 @@ type ReconcileClusterDeployment struct {
 //
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=serviceaccounts;secrets;configmaps,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -859,6 +864,13 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 			_, err := controllerutils.SetupClusterInstallServiceAccount(r, cd.Namespace, cdLog)
 			if err != nil {
 				cdLog.WithError(err).Error("error setting up service account and role")
+				if strings.Contains(err.Error(), terminatedErrorString) {
```
csrwng: @dgoodwin should we also confirm that the namespace is getting deleted by doing a get on the namespace, instead of just relying on the error string?

dgoodwin:
Updated to use a namespace deletiontimestamp check instead of string error check, and changed to work with deprovision requests.

csrwng:
@dgoodwin we need to add permissions to the manager role to get/list/watch namespaces

dgoodwin:
Updated for rbac after testing in-pod.

PR #379, opened 2019-05-13
Add counter metric for remote cluster client creation.

dgoodwin:


csrwng:
@dgoodwin I'd like to better understand what we expect to get from this metric. We could create many remote clients without making a single call to them. Don't we want to know what actual calls we make?

dgoodwin:
Perfect world I'd love metrics on the actual calls we use but I didnt see
an easy way to wrap it. Client creation was just an indicator that I
thought might expose problems but it's possibly already covered by
reconcile counters in each controller.

csrwng:
@dgoodwin an idea to accomplish this is to wrap the client config's RoundTripper with our own. Every time a RoundTrip(*Request) function is called on it, simply increment a counter.

dgoodwin:
Closing for reconsideration. Thanks.

PR #380, opened 2019-05-13
Remove spec.state field in clusterdeployment

csrwng:
Removes mutatingwebhook configuration if found.


dgoodwin:
Good thinking on removing the webhook. We should double check stage later today once this rolls out and make sure everything looks healthy.

/lgtm

PR #383, opened 2019-05-14
Add release image source to create-cluster

csrwng:
Adds a flag to the create-cluster command that points to a URL that contains a pull-spec to a release image to use.

The default is https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest
which should return the latest stable release image for OpenShift 4.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -17,8 +17,10 @@ limitations under the License.
 package createcluster
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


contrib/pkg/createcluster/create.go
```diff
@@ -105,15 +108,15 @@ type Options struct {
...
 	UseClusterImageSet bool
 	ManageDNS          bool
 	Output             string
 	IncludeSecrets     bool
 }
 
 const (
-	defaultHiveImage    = "registry.svc.ci.openshift.org/openshift/hive-v4.0:hive"
-	defaultReleaseImage = "registry.svc.ci.openshift.org/openshift/origin-release:v4.0"
+	defaultHiveImage = "registry.svc.ci.openshift.org/openshift/hive-v4.0:hive"
```
dgoodwin: Is our image still updated at this location, just the origin-release that isn't?

csrwng: That’s correct. It occurred to me this morning that there might be a 4.1 image stream now. Will check.

PR #384, opened 2019-05-14
Optional DNS server for DNSZone checks

csrwng:
Allows setting a set of DNS servers to perform SOA record checks when determining if a DNS Zone is ready to be used.

You can specify the DNS servers via an environment variable `ZONE_CHECK_DNS_SERVERS`. To specify multiple servers, separate them with a comma.

Example:
```
oc set env -n hive deployment/hive-operator ZONE_CHECK_DNS_SERVERS=8.8.8.8,8.8.4.4
```

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -18,6 +18,7 @@ package hive
 
 import (
 	"context"
+	"os"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/hive/hive.go
```diff
@@ -36,6 +37,10 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+const (
+	dnsServersEnvVar = "ZONE_CHECK_DNS_SERVERS"
```
csrwng: Note: the env var is redeclared here to avoid circular import references. Ideally this is temporary and if it resolves the DNSZone issue, we'll make it a part of the HiveConfig type.

PR #385, opened 2019-05-16
Deploy CRDs in the hive-operator.

dgoodwin:
Due to an OLM bug where CRDs are only created new, never updated, we
need a temporary workaround until a fix can be delivered. Rather than
having humans kubectl apply all the CRDs we'll do so with our operators.
The CRDs still are referenced in the OLM bundle, but we're re-applying
them here.

Change also adds hive admin and reader permissions to view CRDs, as this
is coming up a lot given the current bugs.

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -61,10 +61,11 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 	}
 	hLog.Info("deployment applied (%s)", result)
 
-	// Deploy the desired ClusterImageSets representing installable releases of OpenShift.
-	// TODO: in future this should be pipelined somehow.
 	applyAssets := []string{
 		"config/manager/service.yaml",
+
+		// Deploy the desired ClusterImageSets representing installable releases of OpenShift.
+		// TODO: in future this should be pipelined somehow.
 		"config/clusterimagesets/openshift-4.0-latest.yaml",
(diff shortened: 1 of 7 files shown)


pkg/operator/hive/hive.go
```diff
@@ -61,17 +61,31 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+		"config/crds/hive_v1alpha1_clusterdeployment.yaml",
+		"config/crds/hive_v1alpha1_clusterdeprovisionrequest.yaml",
+		"config/crds/hive_v1alpha1_clusterimageset.yaml",
+		"config/crds/hive_v1alpha1_dnsendpoint.yaml",
+		"config/crds/hive_v1alpha1_dnszone.yaml",
+		"config/crds/hive_v1alpha1_hiveconfig.yaml",
+		"config/crds/hive_v1alpha1_selectorsyncidentityprovider.yaml",
+		"config/crds/hive_v1alpha1_selectorsyncset.yaml",
+		"config/crds/hive_v1alpha1_syncidentityprovider.yaml",
+		"config/crds/hive_v1alpha1_syncset.yaml",
```
LalatenduMohanty: @dgoodwin Will it cause issue when the issue get fixed in OLM?

dgoodwin: We've spoken with Evan, it doesn't appear so. No guarantees, worst case I think we'd get some kind of thrashing constantly updating the CRD.

PR #387, opened 2019-05-16
Add provisioning/deprovisioning report commands.

dgoodwin:
These hiveutil commands can be used to scan for details about stuck
install and uninstall jobs.


csrwng:
Very nice. I'm thinking that it'd be good to come up with a must-gather command that can optionally collect information specific to an alert. That way, we can kick it off for each alert we receive on Slack.

PR #388, opened 2019-05-20
Control plane certs - use additionalCertificates field

csrwng:
When specifying a default certificate for the control plane, use the AdditionalCertificates field in the APIServer config instead of the DefaultServingCertificate.

We need to do this until the kube-apiserver operator is fixed in 4.x


csrwng:
@twiest ptal

PR #389, opened 2019-05-20
Remote ingress - include toleration to ensure ingresscontroller reload

csrwng:
Adds a toleration to IngressControllers that we sync to the remote cluster with a hash of the current cert secret data as the value. If the cert secret changes, we will update the toleration with a new hash value and cause the IngressController to reload.


csrwng:
@joelddiaz ptal

PR #390, opened 2019-05-20
Change controller-tools from twiest fork to OpenShift Master team's fork.

twiest:
- [x] Change controller-tools vendoring from twiest's fork to OpenShift Master team's fork
- [x] Update kube vendor to 1.13.4
- [x] Fix up Hive to work with Kubernetes 1.13.4.
- [x] Regenerate assets using OpenShift Master Team's controller-tools.
- [x] Remove federation controller since it's not used anymore.
- [x] Convert Hive glog calls to klog.
- [x] Test change


twiest:
@csrwng PTAL

csrwng:
@twiest this looks good to me. In a follow up we can add //+required to any fields that are truly required and also remove the code where we fix up the cluster deployment status to avoid the required validation errors.

csrwng:
@twiest the e2e failure seems legit. For whatever reason the operator doesn't seem to be starting the controller.

twiest:
@csrwng ok, digging deeper. The last e2e failure didn't seem legit. It was a Terraform error creating an IAM role.

twiest:
@csrwng I think this is ready now. Would you mind giving it another look please?

/hold cancel

csrwng:
@twiest this looks good to me. But because this is such a big change, I'd like to be in the office in case there's something we need to immediately fix on stage. I'll tag it on Tuesday.

PR #392, opened 2019-05-21
Delay install retries with sleep.

dgoodwin:
Modifies install manager to write a sleep-seconds.txt file that uses the
current number of retries for exponential backoff when an install fails,
up to a max of 24 hours. This will help prevent constant API calls on
clusters that need attention before they can possibly install
successfully.

Pod install command is now a shell script that optionally looks for this
file, and if found, sleeps the requested number of seconds. This should
allow the feature to be backward compatible with older hive images which
do not write the file.

This approach should limit the memory consumed as the hiveutil
install-manager has exited, and we're just left with a shell sleep.

Additionally this change adds a deprovison immediately after failed installs.

This will help reduce running resources in the event of a long delay to
the next attempt.

We will still attempt to cleanup at the start of a retry if the InfraID
is still set. It will not be if the deprovision after failure
succeessfully cleared it.

### DIFF
--- a/contrib/pkg/installmanager/installmanager.go
+++ b/contrib/pkg/installmanager/installmanager.go
@@ -23,11 +23,13 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -70,6 +72,7 @@ const (
 	adminSSHKeySecretKey                = "ssh-publickey"
 	installerFullLogFile                = ".openshift_install.log"
 	installerConsoleLogFilePath         = "/tmp/openshift-install-console.log"
+	sleepSecondsFilename                = "sleep-seconds.txt"
 )
 
 var (
@@ -217,8 +220,6 @@ func (m *InstallManager) Run() error {
 		m.log.WithError(err).Error("error marshalling install-config.yaml")
 		return err
 	}
-	installConfig := string(d)
-	m.log.Debugf("install config: %s", installConfig)
 	err = ioutil.WriteFile(filepath.Join(m.WorkDir, "install-config.yaml"), d, 0644)
 	if err != nil {
 		m.log.WithError(err).Error("error writing install-config.yaml to disk")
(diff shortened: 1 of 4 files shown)


contrib/pkg/installmanager/installmanager.go
```diff
@@ -249,7 +250,21 @@ func (m *InstallManager) Run() error {
 
 	installErr := m.provisionCluster(cd)
 	if installErr != nil {
-		m.log.WithError(installErr).Error("error running openshift-install")
+		m.log.WithError(installErr).Error("error running openshift-install, running deprovision to clean up")
+		// TODO: should we timebox this deprovision attempt in the event it gets stuck?
+		if err := m.cleanupFailedInstall(cd); err != nil {
+			// Log the error but continue. It is possible we were not able to clear the infraID
+			// here, but we will attempt this again anyhow when the next job reties. The
```
csrwng: s/reties/retries/

docs/developing.md
```diff
@@ -68,3 +68,14 @@ $ dep ensure -v
 * The Dep cache located under *_$GOPATH/pkg/dep_*.
 * If you see any Dep errors during `make vendor`, you can remove local cached directory and try again.
 
+## Developing Hiveutil Install Manager
```
csrwng: nice!

contrib/pkg/installmanager/installmanager.go
```diff
@@ -302,15 +317,43 @@ func (m *InstallManager) waitForFiles(files []string) {
...
+	// this pod restarted before the controller updated the value. Used when we write sleep seconds below.
+	cd, err := m.loadClusterDeployment()
+	currentRetries := cd.Status.InstallRestarts
+	if err != nil {
+		m.log.WithError(err).Error("ignoring error looking up cluster deployment for restart count")
+		return
+	}
+	sleepSeconds := calcSleepSeconds(currentRetries)
+	m.log.WithField("seconds", sleepSeconds).Info("calculated seconds to sleep for")
+	m.log.Infof("string: %s", strconv.Itoa(sleepSeconds))
```
csrwng: should this be an info line?

dgoodwin:
Touchups added.

dgoodwin:
/hold

Nothing is rolling out to stage due to OLM breakage, holding this one so we can use it when it's fixed and we want a redeploy.

PR #396, opened 2019-05-22
Generate kubeconfig secret with additional CAs

csrwng:
Adds field to ClusterDeployment status to hold the raw kubeconfig
Points the original status.AdminKubeconfigSecret to a new generated secret that contains additional CAs
Includes fix to hive operator to force a reload of the hive controllers pod when the AdditionalCertificateAuthorities value changes.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -67,6 +67,7 @@ const (
 	adminCredsSecretPasswordKey = "password"
 	adminSSHKeySecretKey        = "ssh-publickey"
 	adminKubeconfigKey          = "kubeconfig"
+	rawAdminKubeconfigKey       = "raw-kubeconfig"
 	clusterVersionObjectName    = "version"
 	clusterVersionUnknown       = "undef"
 
@@ -760,6 +761,10 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
 				return err
 			}
 		} else {
+			err = r.fixupAdminKubeconfigSecret(adminKubeconfigSecret, cdLog)
+			if err != nil {
+				return err
+			}
 			err = r.setAdminKubeconfigStatus(cd, adminKubeconfigSecret, cdLog)
 			if err != nil {
 				return err
@@ -780,6 +785,36 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
 	} else {
 		cdLog.Infof("cluster deployment status unchanged")
 	}
+	return nil
+}
+
+func (r *ReconcileClusterDeployment) fixupAdminKubeconfigSecret(secret *corev1.Secret, cdLog log.FieldLogger) error {
+	originalSecret := secret.DeepCopy()
+
+	rawData, hasRawData := secret.Data[rawAdminKubeconfigKey]
+	if !hasRawData {
+		secret.Data[rawAdminKubeconfigKey] = secret.Data[adminKubeconfigKey]
+		rawData = secret.Data[adminKubeconfigKey]
+	}
+
+	var err error
+	secret.Data[adminKubeconfigKey], err = controllerutils.FixupKubeconfig(rawData)
+	if err != nil {
+		cdLog.WithError(err).Errorf("cannot fixup kubeconfig to generate new one")
+		return err
+	}
+
+	if reflect.DeepEqual(originalSecret.Data, secret.Data) {
+		cdLog.Debug("secret data has not changed, no need to update")
+		return nil
+	}
+
+	err = r.Update(context.TODO(), secret)
+	if err != nil {
+		cdLog.WithError(err).Error("error updated admin kubeconfig secret")
+		return err
+	}
+
 	return nil
 }
 
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -19,6 +19,7 @@ package hive
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"fmt"
 	"os"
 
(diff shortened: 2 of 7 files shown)


pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -169,6 +169,11 @@ type ClusterDeploymentStatus struct {
 	// this ClusterDeployment.
 	FederatedClusterRef *corev1.ObjectReference `json:"federatedClusterRef,omitempty"`
 
+	// RawAdminKubeconfigSecret references the secret containing the original admin kubeconfig for this cluster.
+	// This is the kubeconfig as generated by the installer. Use the secret referenced by AdminKubeconfigSecret
```
dgoodwin: I would add "This kubeconfig will not be functional by itself if you are using additional certificate authorities, and in general it is safer to use the secret referenced by AdminKubeconfigSecret which will include these additional CAs."

Crazy thought here though, instead of exposing this distinction in the API, it's probably a bad idea to just have one secret with the original in an annotation? Probably bypasses some secret storage benefits I assume.

pkg/operator/hive/hive.go
```diff
@@ -192,9 +193,11 @@ func (r *ReconcileHiveConfig) includeAdditionalCAs(hLog log.FieldLogger, h *reso
 		return err
 	}
 	hLog.Infof("additional cert secret applied (%s)", result)
+	hash := fmt.Sprintf("%x", md5.Sum(additionalCA.Bytes()))
```
dgoodwin: Might be good to comment here that using the hash in the volume name ensures we see a change to the deployment when the CAs change, triggering a pod restart.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -783,6 +787,49 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
+	if err = controllerutil.SetControllerReference(cd, secret, r.scheme); err != nil {
+		cdLog.WithError(err).Error("error setting controller reference on kubeconfig secret")
+		return "", err
+	}
+	err = r.Create(context.TODO(), secret)
+	if err != nil {
+		cdLog.WithError(err).Errorf("cannot create admin kubeconfig secret")
+	}
+	return name, err
+}
```
dgoodwin: Now that this is in place, it looks like the other controllers are still using AdminKubeconfigSecret but still applying the FixupKubeconfig function to it. Do they need to be updated to either use this new secret without Fixup, or use Raw secret with fixup?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -746,20 +746,24 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
 
 	// The install manager sets this secret name, but we don't consider it a critical failure and
 	// will attempt to heal it here, as the value is predictable.
-	if cd.Status.Installed && cd.Status.AdminKubeconfigSecret.Name == "" {
-		cd.Status.AdminKubeconfigSecret = corev1.LocalObjectReference{Name: fmt.Sprintf("%s-admin-kubeconfig", cd.Name)}
+	if cd.Status.Installed && cd.Status.RawAdminKubeconfigSecret.Name == "" {
+		cd.Status.RawAdminKubeconfigSecret = corev1.LocalObjectReference{Name: fmt.Sprintf("%s-admin-kubeconfig", cd.Name)}
 	}
```
dgoodwin: Do you still need the workaround we talked about for old install manager images? Those will set AdminKubeconfigSecret instead of Raw, if we see that set but no Raw we need to copy it to raw and then update AdminKubeconfigSecret.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -783,6 +787,49 @@ func (r *ReconcileClusterDeployment) updateClusterDeploymentStatus(cd *hivev1.Cl
...
 
+func (r *ReconcileClusterDeployment) generateAdminKubeconfigSecret(cd *hivev1.ClusterDeployment, rawSecret *corev1.Secret, cdLog log.FieldLogger) (string, error) {
+	var hashPart string
+	if len(controllerutils.AdditionalCAHash()) == 0 {
+		hashPart = "original"
+	} else {
+		additionalCAHash := controllerutils.AdditionalCAHash()
+		hashPart = additionalCAHash[:20]
+	}
+	name := fmt.Sprintf("%s-%s", rawSecret.Name, hashPart)
```
dgoodwin: so every additional CA change results in a new secret, but we don't cleanup the old ones. That is probably ok but is there a reason we can't just use a constant secret name and avoid leaking them?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: For old install manager images, the RawAdminKubeconfigSecret won't be set, so we'll set it with the code above. After that, the flow is with the new install manager image.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: The reason I use the name with the hash is that I don't have to worry about updating an out of date secret. If the secret exists with the right hash, then I just skip creating it.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: Thanks, forgot about those. I'd prefer to update them to use the raw secret and fix them up, in case there is a delay in the clusterdeployment controller generating the updated secret.

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
csrwng: Hmm, it's not a bad idea. We could even add another key to the secret instead of just an annotation. Let's discuss at scrum.

csrwng:
Reworked to use a single secret with a key for the original kubeconfig (`raw-kubeconfig`) and the regular kubeconfig with additional CAs added.

PR #397, opened 2019-05-23
Change README to use create-cluster instead of template.

dgoodwin:


README.md
```diff
@@ -79,52 +79,20 @@ NOTE: assumes you have previously deployed using one of the above methods.
...
-   PULL_SECRET="${PULL_SECRET}" \
-   AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}" \
-   AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" \
-   HIVE_IMAGE="${HIVE_IMAGE}" \
-   HIVE_IMAGE_PULL_POLICY="${HIVE_IMAGE_PULL_POLICY}" \
-   INSTALLER_IMAGE="${INSTALLER_IMAGE}" \
-   INSTALLER_IMAGE_PULL_POLICY="${INSTALLER_IMAGE_PULL_POLICY}" \
-   RELEASE_IMAGE="${RELEASE_IMAGE}" \
-   | oc apply -f -
+bin/hiveutil create-cluster --base-domain=mycluster.new-installer.openshift.com --pull-secret-file ~/pull-secret --ssh-key-file ~/.ssh/id_rsa.pub mycluster -o yaml
```
LalatenduMohanty: Nit pick, should it be `~/.pull-secret` instead of `~/pull-secret`

README.md
```diff
@@ -79,52 +79,20 @@ NOTE: assumes you have previously deployed using one of the above methods.
...
 ```
 
+To submit to the server you can drop the -o yaml, or pipe to a file and kubectl apply. This command assumes the latest Hive master CI build, and the latest OpenShift stable release. --hive-image can be specified to use a specific Hive image to run the install, and --release-image can be specified to control which OpenShift release image to install in the cluster.
```
LalatenduMohanty: IMO we should give command the command, so that others can copy paste it in their terminal.

LalatenduMohanty: s/This command assumes the latest Hive master CI build, and the latest OpenShift stable release./By default this command uses  the latest Hive master CI build, and the latest OpenShift stable release./

README.md
dgoodwin: Updated.

README.md
dgoodwin: Updated.

dgoodwin: It doesn't seem right to add the same command twice without -o yaml as that just clutters up the readme. How about we just remove the -o yaml and note below to add it if you want to see the output.

README.md
```diff
@@ -79,52 +79,22 @@ NOTE: assumes you have previously deployed using one of the above methods.
...
-   HIVE_IMAGE="${HIVE_IMAGE}" \
-   HIVE_IMAGE_PULL_POLICY="${HIVE_IMAGE_PULL_POLICY}" \
-   INSTALLER_IMAGE="${INSTALLER_IMAGE}" \
-   INSTALLER_IMAGE_PULL_POLICY="${INSTALLER_IMAGE_PULL_POLICY}" \
-   RELEASE_IMAGE="${RELEASE_IMAGE}" \
-   | oc apply -f -
+bin/hiveutil create-cluster --base-domain=mycluster.new-installer.openshift.com --pull-secret-file ~/.pull-secret --ssh-key-file ~/.ssh/id_rsa.pub mycluster
 ```
 
+Add `-o yaml` to the above command if you wish to see the API objects that would be sent to the server, without actually submitting. If you need to make any changes not supported by create-cluster options, the output can be saved, edited, and then submitted with `kubectl apply`.
```
LalatenduMohanty: :+1:

LalatenduMohanty:
/lgtm 

FYI, I have not tested the steps

PR #401, opened 2019-05-24
Workarounds for problems with managed DNSZone cleanup.

dgoodwin:
Workaround an issue we've seen at least on 3.11 where is it possible for
a dnszone to be terminating when it's parent cluster deployment is gone,
despite blockOwnerDeletion on it's owner ref.

Instead we now refuse to remove the cluster deployment finalizer until
managed dnszones are deleted.

Additionally we have seen rare clusters where deleting the parent
cluster deployment did not automatically delete the managed DNSZone.
Also watch for this situation and issue our own delete if encountered.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -819,11 +819,43 @@ func (r *ReconcileClusterDeployment) setAdminKubeconfigStatus(cd *hivev1.Cluster
 	return nil
 }
 
+// ensureManagedDNSZoneDeleted is a safety check to ensure that the child managed DNSZone
+// linked to the parent cluster deployment gets a deletionTimestamp when the parent is deleted.
+// Normally we expect Kube garbage collection to do this for us, but in rare cases we've seen it
+// not working as intended.
+func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	if !cd.Spec.ManageDNS {
+		return nil
+	}
+	dnsZone := &hivev1.DNSZone{}
+	dnsZoneNamespacedName := types.NamespacedName{Namespace: cd.Namespace, Name: dnsZoneName(cd.Name)}
+	err := r.Get(context.TODO(), dnsZoneNamespacedName, dnsZone)
+	if err != nil && !errors.IsNotFound(err) {
+		cdLog.WithError(err).Error("error looking up managed dnszone")
+		return err
+	}
+	if err == nil && dnsZone.DeletionTimestamp == nil {
+		cdLog.Warn("managed dnszone did not get a deletionTimestamp when parent cluster deployment was deleted, deleting manually")
+		err = r.Delete(context.TODO(), dnsZone,
+			client.PropagationPolicy(metav1.DeletePropagationForeground))
+		if err != nil {
+			cdLog.WithError(err).Error("error deleting managed dnszone")
+			return err
+		}
+	}
+	return nil
+}
+
 func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.ClusterDeployment, hiveImage string, cdLog log.FieldLogger) (reconcile.Result, error) {
+
+	err := r.ensureManagedDNSZoneDeleted(cd, cdLog)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 
 	// Delete the install job in case it's still running:
 	installJob := &batchv1.Job{}
-	err := r.Get(context.Background(),
+	err = r.Get(context.Background(),
 		types.NamespacedName{
 			Name:      install.GetInstallJobName(cd),
 			Namespace: cd.Namespace,
@@ -932,6 +964,7 @@ func (r *ReconcileClusterDeployment) addClusterDeploymentFinalizer(cd *hivev1.Cl
 }
 
 func (r *ReconcileClusterDeployment) removeClusterDeploymentFinalizer(cd *hivev1.ClusterDeployment) error {
+
 	cd = cd.DeepCopy()
 	controllerutils.DeleteFinalizer(cd, hivev1.FinalizerDeprovision)
 	err := r.Update(context.TODO(), cd)
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -553,6 +553,23 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 		},
 		{
+			name: "Ensure managed DNSZone is deleted with cluster deployment",
+			existing: []runtime.Object{
+				func() *hivev1.ClusterDeployment {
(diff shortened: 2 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -932,6 +963,20 @@ func (r *ReconcileClusterDeployment) addClusterDeploymentFinalizer(cd *hivev1.Cl
 }
 
 func (r *ReconcileClusterDeployment) removeClusterDeploymentFinalizer(cd *hivev1.ClusterDeployment) error {
+
+	if cd.Spec.ManageDNS {
+		// Block clearing the finalizer if there if this cluster still has a managed dnszone that has not yet
```
csrwng: s/if there//

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -819,11 +819,42 @@ func (r *ReconcileClusterDeployment) setAdminKubeconfigStatus(cd *hivev1.Cluster
...
+		dnsZone := &hivev1.DNSZone{}
+		dnsZoneNSN := types.NamespacedName{Namespace: cd.Namespace, Name: dnsZoneName(cd.Name)}
+		err := r.Get(context.TODO(), dnsZoneNSN, dnsZone)
+		if err == nil {
+			if dnsZone.DeletionTimestamp == nil {
+				cdLog.Warn("managed dnszone did not get a deletionTimestamp when parent cluster deployment was deleted, deleting manually")
+				err = r.Delete(context.TODO(), dnsZone,
+					client.PropagationPolicy(metav1.DeletePropagationForeground))
+				if err != nil {
+					return err
```
csrwng: log the error

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -819,11 +819,42 @@ func (r *ReconcileClusterDeployment) setAdminKubeconfigStatus(cd *hivev1.Cluster
 	return nil
 }
 
+// ensureManagedDNSZoneDeleted is a safety check to ensure that the child managed DNSZone
+// linked to the parent cluster deployment gets a deletionTimestamp when the parent is deleted.
+// Normally we expect Kube garbage collection to do this for us, but in rare cases we've seen it
+// not working as intended.
+func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	if cd.Spec.ManageDNS {
```
csrwng: style: check for the negative condition and return immediately

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -819,11 +819,42 @@ func (r *ReconcileClusterDeployment) setAdminKubeconfigStatus(cd *hivev1.Cluster
...
+		dnsZone := &hivev1.DNSZone{}
+		dnsZoneNSN := types.NamespacedName{Namespace: cd.Namespace, Name: dnsZoneName(cd.Name)}
+		err := r.Get(context.TODO(), dnsZoneNSN, dnsZone)
+		if err == nil {
```
csrwng: style: check for error and return immediately

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -819,11 +819,42 @@ func (r *ReconcileClusterDeployment) setAdminKubeconfigStatus(cd *hivev1.Cluster
...
 }
 
+// ensureManagedDNSZoneDeleted is a safety check to ensure that the child managed DNSZone
+// linked to the parent cluster deployment gets a deletionTimestamp when the parent is deleted.
+// Normally we expect Kube garbage collection to do this for us, but in rare cases we've seen it
+// not working as intended.
+func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	if cd.Spec.ManageDNS {
+		dnsZone := &hivev1.DNSZone{}
+		dnsZoneNSN := types.NamespacedName{Namespace: cd.Namespace, Name: dnsZoneName(cd.Name)}
```
csrwng: style: use consistent name for the variable `dnsZoneNamespacedName` is used below

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -932,6 +963,20 @@ func (r *ReconcileClusterDeployment) addClusterDeploymentFinalizer(cd *hivev1.Cl
 }
 
 func (r *ReconcileClusterDeployment) removeClusterDeploymentFinalizer(cd *hivev1.ClusterDeployment) error {
+
+	if cd.Spec.ManageDNS {
```
csrwng: Not sure that this is needed. All calls to removeClusterDeploymentFinalizer occur in the same function that calls `ensureManagedDNSZoneDeleted` above. And because we're using foreground propagation, the DNSZone should be gone by this point.

dgoodwin: The if or the the whole block? 

This one relates to the namespace deletion, which UHC assumes it can do as soon as the cluster deployment is gone. We saw this in stage where the cluster deployment is gone, the namespace is deleted, the dnszone has a deletionTimestamp after the ns was deleted, but can't act because the AWS creds secret also was deleted. This part of the fix makes sure the cluster doesn't go away until the DNSZone does (which Kube should be doing for us anyhow, but doesn't seem to 100% reliably), which in turn keeps UHC from deleting the namespace until we're ready. (alternative would be asking them to wait for both cluster deployment and dnszone to be gone which they're willing to do, but seemed best to handle this internally)

dgoodwin: Or a rephrase, the other part of this fix ensures the DNSZone gets a delete request, this part ensures we wait for it to complete before we let the CD go away which would cause UHC to delete the namespace.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Updates for all of these pushed.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: The whole block :) 
I didn't see another path to this block except after calling the `ensureManagedDNSZoneDeleted` function you introduced above. And if you've already called that function, then this seems unnecessary.

https://github.com/openshift/hive/blob/9fb4db000c0108175785f3a09230960ca990a49a/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L856-L957

csrwng: my understanding of DeletePropagationForeground is that the delete will wait for the resource to actually go away

dgoodwin: I see, that makes sense, will remove.

dgoodwin: done

PR #403, opened 2019-05-24
wildcard ingress migration and webhook validation

joelddiaz:
- [x] don't allow the domain field on an ingress entry to lead with `*`. this gets copied over to the remote cluster, and the ingress controller will reject the leading `*` (the cluster's ingress controller assumes the domain is a wildcard).
- [x] ingress controller doesn't allow ingress entries where the ingress domain specified is different than the domain the cluster is under (eg cluster.example.com can't have an ingress like extra.notexample.com, it must be under cluster.example.com).
- [x] migrate old clusterdeployments with wildcard ingress entries to new format.

### DIFF
--- a/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
+++ b/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
@@ -443,7 +443,8 @@ func hasClearedOutPreviouslyDefinedIngressList(oldObject, newObject *hivev1.Clus
 
 func validateIngressDomainsShareClusterDomain(newObject *hivev1.ClusterDeploymentSpec) bool {
 	// ingress entries must share the same domain as the cluster
-	regexString := fmt.Sprintf("(?i).*%s.%s$", newObject.ClusterName, newObject.BaseDomain)
+	// so watch for an ingress domain ending in: .<clusterName>.<baseDomain>
+	regexString := fmt.Sprintf(`(?i).*\.%s.%s$`, newObject.ClusterName, newObject.BaseDomain)
 	sharedSubdomain := regexp.MustCompile(regexString)
 
 	for _, ingress := range newObject.Ingress {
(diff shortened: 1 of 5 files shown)


pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -455,6 +441,31 @@ func hasClearedOutPreviouslyDefinedIngressList(oldObject, newObject *hivev1.Clus
 	return false
 }
 
+func validateIngressDomainsShareClusterDomain(newObject *hivev1.ClusterDeploymentSpec) bool {
+	// ingress entries must share the same domain as the cluster
+	regexString := fmt.Sprintf("(?i).*%s.%s$", newObject.ClusterName, newObject.BaseDomain)
```
csrwng: what is the purpose of the ?i (for my own education)... also, shouldn't the regexp include a period before the cluster name?

joelddiaz: case insenstive.
yes, adding the period will improve the pattern matching. will add that.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -868,3 +893,68 @@ func testAvailableDNSZone() *hivev1.DNSZone {
...
+	return cd
+}
+
+func TestClusterDeploymentWildcardDomainMigration(t *testing.T) {
+	apis.AddToScheme(scheme.Scheme)
+
+	tests := []struct {
+		name              string
+		existing          *hivev1.ClusterDeployment
+		migrationExpected bool
```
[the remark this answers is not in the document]
joelddiaz: ok. added struct to allow listing any expected domains and what that string should look like for each defined ingress entry.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
joelddiaz: added the leading `.` as part of the shared domain validation

PR #407, opened 2019-05-28
Handle long clusterdeployment names

csrwng:
Introduces utility from origin to shorten long names, switches to using the utility wherever we generate names of API resources from a clusterdeployment name.

Fixes https://jira.coreos.com/browse/CO-435

### DIFF
--- a/pkg/imageset/generate.go
+++ b/pkg/imageset/generate.go
@@ -17,7 +17,6 @@ limitations under the License.
 package imageset
 
 import (
-	"fmt"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


pkg/imageset/generate.go
```diff
@@ -185,7 +185,7 @@ func GenerateImageSetJob(cd *hivev1.ClusterDeployment, imageSet *hivev1.ClusterI
 
 // GetImageSetJobName returns the expected name of the imageset job for a ClusterImageSet.
 func GetImageSetJobName(cdName string) string {
-	return fmt.Sprintf("%s-imageset", cdName)
+	return apihelpers.GetResourceName(cdName, "imageset")
```
dgoodwin: Not for this PR but I think we should establish a convention that any dependent object we create, but don't want to explicitly add as a link in the API, must have a hive.openshift.io/cluster-deployment-name label.

PR #410, opened 2019-05-30
scaffolding for saving log tarball on failures

joelddiaz:
allow storing the ssh private key in the ssh secret

extend installmanager to set up ssh agent (if necessary), read in terraform file, generate script to fetch logs, and run the script to fetch the logs

TODO: store the saved log somewhere (S3, the in-cluster registry)

### DIFF
--- a/config/templates/cluster-deployment.yaml
+++ b/config/templates/cluster-deployment.yaml
@@ -12,6 +12,10 @@ parameters:
   displayName: SSH Key
   description: Your public SSH key to reach instances.
   required: true
+- name: SSH_PRIV_KEY
+  displayName: SSH Private Key
+  description: The private key counterpart to the public ssh key to allow pulling logs.
+  required: false
 - name: BASE_DOMAIN
   displayName: Base DNS Domain
   description: Base DNS domain for your cluster. Will be combined with cluster name when creating entries.
@@ -84,6 +88,7 @@ objects:
   type: Opaque
   stringData:
     ssh-publickey: "${SSH_KEY}"
+    ssh-privatekey: "${SSH_PRIV_KEY}"
 
 - apiVersion: hive.openshift.io/v1alpha1
   kind: ClusterDeployment
--- a/contrib/pkg/installmanager/installmanager_test.go
+++ b/contrib/pkg/installmanager/installmanager_test.go
@@ -61,6 +61,29 @@ echo "fakekubeconfig" > $WORKDIR/auth/kubeconfig
 echo "fakepassword" > $WORKDIR/auth/kubeadmin-password
 echo "some fake installer log output" >  /tmp/openshift-install-console.log
 `
+
+	fakeSSHAddBinary = `#!/bin/bash
+KEY_FILE_PATH=${1}
+
+if [[ ${KEY_FILE_PATH} != "%s" ]]; then
+		echo "Parameter not what expected"
+		exit 1
+fi
+
+exit 0
+`
+	fakeSSHAgentSockPath = "/path/to/agent/sockfile"
+	fakeSSHAgentPID      = "12345"
+
+	alwaysErrorBinary = `#!/bin/sh
+exit 1`
+)
+
+var (
+	fakeSSHAgentBinary = `#!/bin/sh
+echo "SSH_AUTH_SOCK=%s; export SSH_AUTH_SOCK;"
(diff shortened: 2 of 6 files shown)


contrib/pkg/installmanager/installmanager_test.go
```diff
@@ -454,3 +455,253 @@ last line with password in text`,
...
+						}
+					}
+				}
+			}
+		}
+	]
+}`, instanceName, ip)
+}
+
+func TestInstallManagerSSH(t *testing.T) {
```
csrwng: @joelddiaz I think we could simplify this test to limit it to testing your code (which basically parses ssh-agent output, sets env vars and calls ssh-add). You could supply alternate binaries for `ssh-agent` and `ssh-add` just to ensure that they are properly being called (and that you're properly parsing the output)

csrwng: we kind of already do that with the installer binary: https://github.com/openshift/hive/blob/34814cafa464c8040918a15767d758ce1a957c9a/contrib/pkg/installmanager/installmanager_test.go#L55-L64

config/templates/cluster-deployment.yaml
```diff
@@ -12,6 +12,10 @@ parameters:
   displayName: SSH Key
   description: Your public SSH key to reach instances.
   required: true
+- name: SSH_PRIV_KEY
```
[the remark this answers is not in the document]
joelddiaz: k. also updated create-cluster with updated PR

contrib/pkg/installmanager/installmanager_test.go
joelddiaz: k. have some fake ssh-agent and ssh-add instead of relying on actual binaries

csrwng:
@joelddiaz you have a mismatch between the arg name declared in installmanager and the one passed in the generated install job. 
/lgtm cancel

joelddiaz:
for posterity, if we decide to try to upload logs to a registry, something like this could be a starting point

```
package main

import (
        "context"
        "fmt"

        "github.com/containers/image/copy"
        "github.com/containers/image/signature"
        "github.com/containers/image/transports/alltransports"
)

func main() {
        var ctx context.Context
        ctx = context.TODO()

        var policy *signature.Policy
        policy, err := signature.DefaultPolicy(nil)
        if err != nil {
                panic("error creating default policy")
        }
        policyCtx, err := signature.NewPolicyContext(policy)
        if err != nil {
                panic("error creating policy context")
        }
        srcRef, err := alltransports.ParseImageName("tarball://home/jdiaz/log-bundle.tar.gz")
        if err != nil {
                panic("error setting source ref")
        }
        destRef, err := alltransports.ParseImageName("docker://quay.io/jdiaz/testing123")
        if err != nil {
                panic("error setting destination ref")
        }
        result, err := copy.Image(ctx, policyCtx, destRef, srcRef, &copy.Options{})
        if err != nil {
                fmt.Printf("ERR: %+v\n", err)
                panic("error copying image")
        }
        fmt.Printf("RES: %+v\n", string(result))
}

```

pkg/install/generate.go
```diff
@@ -223,7 +252,8 @@ func GenerateInstallerJob(
 				// a sleep-seconds.txt file. If one is written, we will sleep that number of seconds. This allows exponential backoff
 				// for failing installs.
 				fmt.Sprintf(
-					"/usr/bin/hiveutil install-manager --work-dir /output --log-level debug --install-config /installconfig/install-config.yaml --region %s %s %s; installer_result=$?; if [ -f /output/sleep-seconds.txt ]; then sleep_seconds=$(cat /output/sleep-seconds.txt); echo \"sleeping for $sleep_seconds seconds until next retry\"; sleep $sleep_seconds; fi; exit $installer_result",
+					"/usr/bin/hiveutil install-manager --work-dir /output --log-level debug --install-config /installconfig/install-config.yaml --ssh-priv-key-path %s --region %s %s %s; installer_result=$?; if [ -f /output/sleep-seconds.txt ]; then sleep_seconds=$(cat /output/sleep-seconds.txt); echo \"sleeping for $sleep_seconds seconds until next retry\"; sleep $sleep_seconds; fi; exit $installer_result",
```
dgoodwin: Old versions of the install image (i.e. all current cluster image sets) do not have this new flag, and thus all installs are now broken in stage. I don't think this has made it to prod.

PR #411, opened 2019-05-31
Use oc instead of kubectl in Makefile apply commands.

dgoodwin:
Bypasses a ValidationError regarding unknown field "nullable" when
applying our CRDs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ run-operator: generate fmt vet
 # Install CRDs into a cluster
 .PHONY: install
 install: crd
-	kubectl apply -f config/crds
+	oc apply -f config/crds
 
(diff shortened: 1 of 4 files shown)


csrwng:
@dgoodwin for consistency should we also change the use of kubectl in:
- README.md
- hack/olm-registry-deploy.sh
- docs/developing.md

and, also we wouldn't need to link it in the e2e script:
https://github.com/openshift/hive/blob/1d85899ceceff61004efab367e6142edcc6cfabb/hack/e2e-test.sh#L11

LalatenduMohanty:
We list `kubectl` as prerequisite in https://github.com/openshift/hive#prerequisites ,also we have many `kubectl` commands  in the `readme` which needs to be fixed for consistency.

PR #424, opened 2019-06-06
Don't update cd when unreachable cond is absent in reachable cluster

LalatenduMohanty:


dgoodwin:
Much simpler, thanks!

/lgtm

PR #427, opened 2019-06-06
Update clusterdeployment controller to handle deleted resources

csrwng:
Removes unnecessary code to fixup status fields in clusterdeployment controller (the fields are no longer required in latest crds)

Implements the following pattern when deleting a related resource:
- If deleting the resource, run the delete and return immediately (don't assume we can continue). No requeue is needed because the delete will cause a requeue.
- When retrieving the resource, if the resource is being deleted (DeletionTimestamp is non-nil), return and requeue for an interval of time. (Otherwise we are not guaranteed that the resource will be requeued)

Other than that, a few minor fixes in the code.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -81,7 +81,7 @@ const (
 
 	dnsZoneCheckInterval = 30 * time.Second
 
-	defaultRequeueTime = time.Second
+	defaultRequeueTime = 10 * time.Second
 
 	jobHashAnnotation = "hive.openshift.io/jobhash"
 )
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -272,23 +272,6 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
-		if err != nil {
-			cdLog.WithError(err).Error("error updating cluster deployment")
-			return reconcile.Result{}, err
-		}
-		return reconcile.Result{
-			Requeue:      true,
-			RequeueAfter: defaultRequeueTime,
-		}, nil
-	}
-
```
LalatenduMohanty: @csrwng Why is this not required any more?

csrwng: because this was temporary migration code that was used to fix outdated syncsets on stage/production. It's been a while since that code was in place and there are no more outdated syncsets that need to be migrated.

LalatenduMohanty: Got it, thanks.

PR #428, opened 2019-06-07
Add hiveutil create-cluster --delete-after.

dgoodwin:


contrib/pkg/createcluster/create.go
```diff
@@ -154,6 +158,7 @@ func NewCreateClusterCommand() *cobra.Command {
 	flags.StringVar(&opt.SSHPublicKey, "ssh-public-key", "", "SSH public key for cluster")
 	flags.StringVar(&opt.BaseDomain, "base-domain", "new-installer.openshift.com", "Base domain for the cluster")
 	flags.StringVar(&opt.PullSecret, "pull-secret", "", "Pull secret for cluster. Takes precedence over pull-secret-file.")
+	flags.StringVar(&opt.DeleteAfter, "delete-after", "", "Delete this cluster after the given duration. (i.e. 8h)")
```
csrwng: Should we set a default value of 8h ?

dgoodwin: This tool seems more useful than the old template which was much more dev specific, I was worried in this case that a default delete after might bite someone. Thinking we should play it safer now.

csrwng: ack

PR #430, opened 2019-06-07
Client Metrics

dgoodwin:
Adds a transport wrapper which parses a little info about the request to determine what resource we're dealing with, and then increments a prometheus counter with labels for method, resource, and what controller the client is from. This allows us to identify who is causing hotloops with excessive writes.

Only tricky part of this was the rest config copying, last week it did not function properly (no metrics), today it does, so we'll see how this goes in stage.

### DIFF
--- a/pkg/controller/clusterversion/clusterversion_controller.go
+++ b/pkg/controller/clusterversion/clusterversion_controller.go
@@ -37,12 +37,14 @@ import (
 
 	openshiftapiv1 "github.com/openshift/api/config/v1"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 )
 
 const (
 	clusterVersionObjectName = "version"
 	clusterVersionUnknown    = "undef"
+	controllerName           = "clusterversion"
 )
 
 // Add creates a new ClusterDeployment Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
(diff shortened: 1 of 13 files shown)


pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -37,12 +37,14 @@ import (
...
 	openshiftapiv1 "github.com/openshift/api/config/v1"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 )
 
 const (
 	clusterVersionObjectName = "version"
 	clusterVersionUnknown    = "undef"
+	controllerName           = "clusterversion"
```
LalatenduMohanty: :+1:

pkg/controller/metrics/clientwrapper.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	options := client.Options{
+		Scheme: mgr.GetScheme(),
+		Mapper: mgr.GetRESTMapper(),
+	}
+	c, err := client.New(cfg, options)
+	if err != nil {
+		log.WithError(err).Fatal("unable to initialize metrics wrapped client")
```
LalatenduMohanty: We are not returning from here in case of error. Is this as designed?

pkg/controller/metrics/clientwrapper.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func init() {
+	metrics.Registry.MustRegister(metricKubeClientRequests)
+}
+
+// NewClientWithMetricsOrDie creates a new controller-runtime client with a wrapper which increments
+// metrics for requests by controller name, HTTP method, and URL path. The client will re-use the
+// managers cache. This should be used in all Hive controllers.
+func NewClientWithMetricsOrDie(mgr manager.Manager, ctrlrName string) client.Client {
```
LalatenduMohanty: We do not returning any error from this function, so why not have the function name as `NewClientWithMetrics`

pkg/controller/metrics/clientwrapper.go
dgoodwin: Yeah log.Fatal includes a sys.Exit, so we're exiting the program if something goes wrong here as it's controller initialization. This is why I went with the "OrDie" in the function name.

LalatenduMohanty: Got it. Thanks.

PR #434, opened 2019-06-10
Take control of hive-frontend role and service account.

dgoodwin:
We suspect bugs in OLM upgrades that are killing the service account.

### DIFF
--- a/config/operator/operator_role.yaml
+++ b/config/operator/operator_role.yaml
@@ -15,6 +15,7 @@ rules:
   - services/finalizers
   - events
   - configmaps
+  - namespaces
   verbs:
   - get
   - list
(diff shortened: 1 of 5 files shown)


dgoodwin:
This is a tricky one to roll out but @jhernand has confirmed that it's ok if we rotate the SA again. I will put a hold to not merge this until everyone is in tomorrow.

/hold

dgoodwin:
I think what will happen here:

- OLM will apply the new CSV, the SA has an owner ref pointing to the old and will be GC'd. 
- Hive's operator will create a new SA in it's place with the same name.

dgoodwin:
updated to grant the operator permission to create/update namespaces so it can create the frontend role. Found this error in the hive operator pod artifacts.

PR #437, opened 2019-06-10
SyncSetInstance API and controller

csrwng:


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -0,0 +1,714 @@
...
+	}
+
+	cd := &hivev1.ClusterDeployment{}
+	cdName := types.NamespacedName{Namespace: ssi.Namespace, Name: ssi.Spec.ClusterDeployment.Name}
+	ssiLog = ssiLog.WithField("clusterdeployment", cdName)
+	err = r.Get(context.TODO(), cdName, cd)
+	if err != nil {
+		ssiLog.WithError(err).Error("error looking up clusterdeployment")
+		return reconcile.Result{}, nil
+	}
```
dgoodwin: Add the unreachable check and skip from: https://github.com/openshift/hive/pull/438

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -0,0 +1,714 @@
...
+	ssiLog.Info("done reconciling syncsetinstance")
+	return reconcile.Result{}, applyErr
+}
+
+func (r *ReconcileSyncSetInstance) addSyncSetInstanceFinalizer(ssi *hivev1.SyncSetInstance) error {
+	controllerutils.AddFinalizer(ssi, hivev1.FinalizerSyncSetInstance)
+	return r.Update(context.TODO(), ssi)
+}
+
+func (r *ReconcileSyncSetInstance) syncDeletedSyncSetInstance(ssi *hivev1.SyncSetInstance, ssiLog log.FieldLogger) (reconcile.Result, error) {
```
dgoodwin: Is the finalizer needed at all?

pkg/apis/hive/v1alpha1/syncsetinstance_types.go
```diff
@@ -0,0 +1,107 @@
...
+	ResourceApplyMode SyncSetResourceApplyMode `json:"resourceApplyMode,omitempty"`
+
+	// SyncSetHash is a hash of the contents of the syncset or selectorsyncset spec.
+	// Its purpose is to cause a syncset instance update whenever there's a change in its
+	// source.
+	SyncSetHash string `json:"syncSetHash,omitempty"`
+}
+
+// SelectorSyncSetReference is a reference to a SelectorSyncSet
+type SelectorSyncSetReference struct {
```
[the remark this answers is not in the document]
csrwng: It is a different type of reference since it's neither LocalObjectReference or ObjectReference so I would say yes

pkg/controller/syncsetinstance/syncsetinstance_controller.go
csrwng: hmm... so I could rework the way I handle not finding the corresponding selectorsyncset/syncset (right now I do a delete, then handle the deleted syncsetinstance). I think it could work. Will update.

csrwng:
This is still pending some manual testing. After that, I'll remove the WIP tag

csrwng:
Tested this manually and so far it's working as expected. Removed the WIP tag.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -0,0 +1,826 @@
...
+		return err
+	}
+
+	// Watch for changes to SyncSetInstance
+	err = c.Watch(&source.Kind{Type: &hivev1.SyncSetInstance{}}, &handler.EnqueueRequestForObject{})
+
+	// Watch for changes to ClusterDeployments
+	err = c.Watch(&source.Kind{Type: &hivev1.ClusterDeployment{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(r.(*ReconcileSyncSetInstance).handleClusterDeployment),
+	})
```
dgoodwin: Do we actually need to watch cluster deployments in this controller? I would kind of expect the SyncSet controller to do this, and update the SyncSetInstance if necessary.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
dgoodwin: I assume finalizer proved useful in the end?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
csrwng: @dgoodwin the reason that I watch cluster deployments is that whether you can apply a syncsetinstance or not depends on the clusterdeployment status. If the installed status is not true, we need to wait for it to be true (and be notified of that change).

csrwng:
> I assume finalizer proved useful in the end?

Yes, after considering it a bit more it made sense that if for whatever reason you delete a syncsetinstance, we don't let it go away without first cleaning up the remote cluster (because it contains status about what was applied).

dgoodwin:
Ok let's merge Monday morning and see how she goes

PR #438, opened 2019-06-11
Skip reconciles for unreachable clusters.

dgoodwin:
This code mistakenly assumed that if you tried to hit a cluster that
already has unreachable condition set, we should return that as an
error. This however causes the controller-runtime framework to keep
retrying, with backoff, but still a lot of reconciling and errors
logged.

Instead we should check for the condition and skip reconcile, without
considering this an error.

In time, if the cluster comes back, the unreachable controller will
clear the condition at which point all these controllers will resume
normal operation.

### DIFF
--- a/pkg/controller/remotemachineset/remotemachineset_controller.go
+++ b/pkg/controller/remotemachineset/remotemachineset_controller.go
@@ -81,7 +81,7 @@ func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 		Client:                        hivemetrics.NewClientWithMetricsOrDie(mgr, controllerName),
 		scheme:                        mgr.GetScheme(),
 		logger:                        log.WithField("controller", controllerName),
-		remoteClusterAPIClientBuilder: controllerutils.GetClusterAPIClient,
+		remoteClusterAPIClientBuilder: controllerutils.BuildClusterAPIClientFromKubeconfig,
 		awsClientBuilder:              awsclient.NewClient,
 	}
 }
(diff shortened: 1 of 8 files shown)


pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -141,6 +141,11 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
 		"clusterDeployment": cd.Name,
 		"namespace":         cd.Namespace,
 	})
+	// If the cluster is unreachable, do not reconcile.
+	if controllerutils.HasUnreachableCondition(cd) {
+		cdLog.Debug("skipping cluster with unreachable condition")
+		return reconcile.Result{}, nil
+	}
```
LalatenduMohanty: :+1:

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -176,6 +177,26 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+				}(),
+				testSecret(adminKubeconfigSecret, adminKubeconfigSecretKey, testName),
+				testSecret(adminPasswordSecret, adminPasswordSecretKey, testName),
+				testSecret(sshKeySecret, sshKeySecretKey, testName),
+			},
+			remoteExisting: []runtime.Object{},
+			expectedRemoteMachineSets: func() *machineapi.MachineSetList {
+				return &machineapi.MachineSetList{}
+			}(),
+		},
```
LalatenduMohanty: :+1:

PR #440, opened 2019-06-11
Fixes make deploy for Mac

danehans:
Previously, `make deploy` fails to run on Mac OS:
```
$ make deploy
<SNIP>
# Deploy the operator manifests:
mkdir -p overlays/deploy
cp overlays/template/* overlays/deploy
if [[ "`uname`" == "Darwin" ]]; then \
	    sed -i "" -e "s|IMAGE_REF|registry.svc.ci.openshift.org/openshift/hive-v4.0:hive|" overlays/deploy/image_patch.yaml; \
	else \
	    sed -i -e "s|IMAGE_REF|registry.svc.ci.openshift.org/openshift/hive-v4.0:hive|" overlays/deploy/image_patch.yaml; \
	fi
sed: can't read : No such file or directory
make: *** [deploy] Error 2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,11 +90,7 @@ deploy: manifests install generate
 	# Deploy the operator manifests:
 	mkdir -p overlays/deploy
 	cp overlays/template/* overlays/deploy
-	if [[ "`uname`" == "Darwin" ]]; then \
-	    sed -i "" -e "s|IMAGE_REF|$(DEPLOY_IMAGE)|" overlays/deploy/image_patch.yaml; \
-	else \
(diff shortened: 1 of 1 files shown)


csrwng:
@danehans are you by any chance using gnu sed on your mac? I don't see this issue with the default sed (and in fact, we have the extra code so that it can work with the bsd sed on macos). Maybe a better fix would be to do a test to determine which form of the -i argument is needed.

csrwng:
addressed by https://github.com/openshift/hive/pull/458
/close

PR #442, opened 2019-06-12
CO-455 Send install job duration metric only for the first time install

LalatenduMohanty:

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -70,7 +70,8 @@ const (
 
 	defaultRequeueTime = 10 * time.Second
 
-	jobHashAnnotation = "hive.openshift.io/jobhash"
+	jobHashAnnotation          = "hive.openshift.io/jobhash"
+	firstTimeInstallAnnotation = "hive.openshift.io/first-time-install"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -512,9 +513,18 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 				cdLog.Errorf("error creating job: %v", err)
 				return reconcile.Result{}, err
 			}
-			kickstartDuration := time.Since(cd.CreationTimestamp.Time)
-			cdLog.WithField("elapsed", kickstartDuration.Seconds()).Info("calculated time to install job seconds")
-			metricInstallDelaySeconds.Observe(float64(kickstartDuration.Seconds()))
+			if value := cd.Annotations[firstTimeInstallAnnotation]; value == "" {
```
csrwng: a slightly better check would be:
`if _, ok := cd.Annotations[firstTimeInstallAnnotation]; !ok {`

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -192,6 +192,21 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+		{
+			name: "Check firstTimeInstallAnnotation in install job",
+			existing: []runtime.Object{
+				testClusterDeployment(),
+				testSecret(corev1.SecretTypeDockerConfigJson, pullSecretSecret, corev1.DockerConfigJsonKey, "{}"),
+				testSecret(corev1.SecretTypeOpaque, sshKeySecret, adminSSHKeySecretKey, "fakesshkey"),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				if value := cd.Annotations[firstTimeInstallAnnotation]; value != "true" {
```
csrwng: I would just add this check to the test for creating an install job 
https://github.com/openshift/hive/pull/442/files#diff-ae597ba7bf52a3a731829cf7f676741bR150

LalatenduMohanty: Added the check to https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go#L124-L136 as it should be part of default install job. Also https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go#L138 does not have the annotation as it has https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go#L141

PR #444, opened 2019-06-12
Remove duplicate syncset status entries

csrwng:
Fixes existing status entries with duplicate entries


dgoodwin:
/lgtm

What can I check tomorrow morning to know we're ok?

PR #445, opened 2019-06-13
Remote client metrics

dgoodwin:


pkg/resource/kubeconfig_factory.go
```diff
@@ -23,20 +26,42 @@ func (r *Helper) getKubeconfigFactory(namespace string) (cmdutil.Factory, error)
 	}
 	r.logger.Debug("creating client config from kubeconfig")
 	clientConfig := clientcmd.NewNonInteractiveClientConfig(*config, "", overrides, nil)
+	if r.metricsEnabled {
+		restConfig, err := clientConfig.ClientConfig()
```
csrwng: @dgoodwin the issue is that every time you call ClientConfig you get a new copy of the config. Modifying the local copy won't affect the result of future calls to ClientConfig (as happens in line 62). If you want the same config to be returned every time, simply add a field to kubeconfigClientGetter to set a modified config, and return *that* config in the ToRESTConfig method. The other consideration is that most likely the client config you get will already have a WrapTransport function set. So you will need to wrap *that* function with your function.

dgoodwin:
Ready for review now I think.

dgoodwin:
Sample metrics:

```
hive_kube_client_requests_total{controller="clusterDeployment",method="DELETE",remote="false",resource="batch/v1/jobs",status="200 OK"} 1
hive_kube_client_requests_total{controller="clusterDeployment",method="POST",remote="false",resource="batch/v1/jobs",status="201 Created"} 2
hive_kube_client_requests_total{controller="clusterDeployment",method="POST",remote="false",resource="core/v1/configmaps",status="201 Created"} 1
hive_kube_client_requests_total{controller="clusterDeployment",method="POST",remote="false",resource="hive.openshift.io/v1alpha1/clusterdeprovisionrequests",status="201 Created"} 1
hive_kube_client_requests_total{controller="clusterDeployment",method="PUT",remote="false",resource="hive.openshift.io/v1alpha1/clusterdeployments",status="200 OK"} 4
hive_kube_client_requests_total{controller="clusterDeprovisionRequest",method="POST",remote="false",resource="batch/v1/jobs",status="201 Created"} 1
hive_kube_client_requests_total{controller="clusterDeprovisionRequest",method="PUT",remote="false",resource="hive.openshift.io/v1alpha1/clusterdeprovisionrequests",status="200 OK"} 2
```

pkg/controller/utils/clientwrapper.go
```diff
@@ -29,15 +29,17 @@ func init() {
 }
 
 // NewClientWithMetricsOrDie creates a new controller-runtime client with a wrapper which increments
-// metrics for requests by controller name, HTTP method, and URL path. The client will re-use the
-// managers cache. This should be used in all Hive controllers.
+// metrics for requests by controller name, HTTP method, URL path, and whether or not the request was
+// to a remote cluster.. The client will re-use the managers cache. This should be used in
+// all Hive controllers.
```
LalatenduMohanty: Moving to `pkg/controller/utils` is fine as `The client method should be used in all Hive controllers.`

pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -115,6 +115,12 @@ func (r *ReconcileClusterVersion) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, nil
 	}
 
+	// If the cluster is not installed, do not reconcile.
```
csrwng: 👍

pkg/resource/restconfig_factory.go
```diff
@@ -10,6 +13,19 @@ import (
 )
 
 func (r *Helper) getRESTConfigFactory(namespace string) (cmdutil.Factory, error) {
+	if r.metricsEnabled {
+		// Copy the possibly shared restConfig reference and add a metrics wrapper.
+		cfg := rest.CopyConfig(r.restConfig)
+		cfg.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
```
csrwng: @dgoodwin I think you should take into consideration the case where WrapTransport is already set to some value. I don't think that happens in any of our use cases currently, but it does happen. I saw it when we were building a client from a kubeconfig pointing to service account credentials.

pkg/resource/kubeconfig_factory.go
```diff
@@ -20,19 +23,42 @@ func (r *Helper) getKubeconfigFactory(namespace string) (cmdutil.Factory, error)
 		overrides.Context.Namespace = namespace
 	}
 	clientConfig := clientcmd.NewNonInteractiveClientConfig(*config, "", overrides, nil)
+	restConfig, err := clientConfig.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+	if r.metricsEnabled {
+		restConfig.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
```
csrwng: See comment for restconfig_factory.go
This is not an issue as long as we're using certificate-based auth as returned by the installer. But if at some point we switch to a service account + token, then you will likely have a WrapTransport already set.

pkg/resource/restconfig_factory.go
dgoodwin: Yeah I tested and it was unset but I can see if I can get it to wrap the existing.

dgoodwin:
Should be ready now.

PR #455, opened 2019-06-20
Adding golangci-lint to Makefile

LalatenduMohanty:


LalatenduMohanty:
https://github.com/golangci/golangci-lint is getting popular as it is a better tool for go code linting. IMO we should include this as our default lint tool in future. However we can not include this as default at this point as it found lot of issues which needs to be fixed first.

twiest:
This looks fine to me, but I'll let @dgoodwin weigh in on if this is the direction he'd like us to go in.

PR #457, opened 2019-06-20
Use hiveutil create-cluster in e2e test

csrwng:
Switches to using `hiveutil create-cluster` in e2e test
Adds --install-once and --uninstall-once flags to create-cluster command


csrwng:
Moved deprovision to its own package so `go run` can work

csrwng:
@dgoodwin this should be ready now

PR #458, opened 2019-06-20
Use kustomize edit to set image in make deploy

csrwng:
Removes need for use of sed, cleaner code


csrwng:
Addresses https://github.com/openshift/hive/pull/440

PR #460, opened 2019-06-21
Fixing ineffectual assignment to errors

LalatenduMohanty:

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -156,12 +156,14 @@ func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 	c, err := controller.New("clusterdeployment-controller", mgr, controller.Options{Reconciler: r, MaxConcurrentReconciles: controllerutils.GetConcurrentReconciles()})
 	if err != nil {
+		log.WithField("controller", controllerName).WithError(err).Error("Error getting new cluster deployment")
 		return err
 	}
 
 	// Watch for changes to ClusterDeployment
 	err = c.Watch(&source.Kind{Type: &hivev1.ClusterDeployment{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -170,6 +170,9 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 		IsController: true,
 		OwnerType:    &hivev1.ClusterDeployment{},
 	})
+	if err != nil {
+		return err
```
dgoodwin: I suspect all of these should have a log.WithError().Error(), otherwise we're probably not getting any info in the logs.

csrwng: In this case, the controller wouldn't start so we'd know something's up but yeah, a log statement wouldn't hurt

LalatenduMohanty: @csrwng @dgoodwin I search through the whole Hive code base and no where we have used `log.WithError().Error()` in the `func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error`.  I also could not verify about what error `Watch` returns. So if you guys think we should add logs then we need to add it for 
```
pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go:
pkg/controller/clusterversion/clusterversion_controller.go:
pkg/controller/controlplanecerts/controlplanecerts_controller.go:
pkg/controller/dnszone/dnszone_controller.go:
pkg/controller/installlogmonitor/installlogmonitor_controller.go:
pkg/controller/remoteingress/remoteingress_controller.go:
pkg/controller/remotemachineset/remotemachineset_controller.go:
pkg/controller/syncidentityprovider/syncidentityprovider_controller.go:
pkg/controller/syncset/syncset_controller.go:
pkg/controller/syncsetinstance/syncsetinstance_controller.go:
pkg/controller/unreachable/unreachable_controller.go:
pkg/operator/hive/hive_controller.go:
```

LalatenduMohanty: @dgoodwin @csrwng Let me know your thoughts on this.

PR #464, opened 2019-06-21
Vendor Heptio Velero

twiest:


LalatenduMohanty:
@twiest IMHO we should vendor the code when we have code in Hive which uses Velero.

twiest:
@LalatenduMohanty Yeah, we've discussed this previously on the team. The idea of having a separate vendor PR is to make it easier to develop code against the vendor.

If we do one PR with separate "vendor" and "new code" commits, then it becomes a huge pain to constantly be rebasing the "vendor" commit off of master while we're writing the "new code" commit.

Basically this way is just simpler and we can always remove the vendor if we end up not using it.

If you still object to this PR, then we can discuss this pattern again as a team at the next stand up that @dgoodwin attends.

LalatenduMohanty:
> If we do one PR with separate "vendor" and "new code" commits, then it becomes a huge pain to constantly be rebasing the "vendor" commit off of master while we're writing the "new code" commit.
> 

I have not seen issues with rebasing because none of the other code in Hive will be using the velero except your branch. 

> Basically this way is just simpler and we can always remove the vendor if we end up not using it.

How do you guys track the removal of something from vendor directory?

If this is something already discussed in the team meeting then we should go ahead with the PR. However I am not going to `lgtm` it as I am not used to this pattern. So hopefully someone else can approve it and it will be merged.

twiest:
> I have not seen issues with rebasing because none of the other code in Hive will be using the velero except your branch.

There are things outside of the vendor directory that can cause conflicts. For instance, the `Gopkg.toml` and `Gopkg.lock` files. We've had situations where we had to fully re-vendor because the Gopkg.lock file had so many conflicts that it was just easier to re-vendor. We then had to rebase the "new code" commit off of that re-vendor commit. It was a mess.


> How do you guys track the removal of something from vendor directory?

Gopkg.toml file is the source of truth. If we see something in there that isn't used, we can remove it and re-run dep to clean it up.

> If this is something already discussed in the team meeting then we should go ahead with the PR. However I am not going to lgtm it as I am not used to this pattern. So hopefully someone else can approve it and it will be merged.

No prob. @joelddiaz would you mind reviewing this?

PR #465, opened 2019-06-25
Regenerate deprovision jobs when spec changes.

dgoodwin:
Similar to what we do for install jobs, track a hash of the spec we
generate and store in an annotation. If this is missing or changed on
the current job, delete it and allow another to be created.

### DIFF
--- a/pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
+++ b/pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
@@ -9,6 +9,7 @@ import (
 
 	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
@@ -29,7 +30,8 @@ import (
 )
 
 const (
-	controllerName = "clusterDeprovisionRequest"
+	controllerName    = "clusterDeprovisionRequest"
+	jobHashAnnotation = "hive.openshift.io/jobhash"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


dgoodwin:
Needs review.

pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
```diff
@@ -183,6 +195,28 @@ func (r *ReconcileClusterDeprovisionRequest) Reconcile(request reconcile.Request
 		metricUninstallJobDuration.Observe(float64(jobDuration.Seconds()))
 		return reconcile.Result{}, nil
 	}
+
+	// Check if the job should be regenerated:
+	newJobNeeded := false
+	if _, ok := existingJob.Annotations[jobHashAnnotation]; !ok {
```
csrwng: Do we need a check for existingJob.Annotations == nil ?

pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
```diff
@@ -183,6 +195,28 @@ func (r *ReconcileClusterDeprovisionRequest) Reconcile(request reconcile.Request
...
+		// this job predates tracking the job hash, so assume we need a new job
+		newJobNeeded = true
+	}
+
+	if existingJob.Annotations[jobHashAnnotation] != jobHash {
+		// delete the job so we get a fresh one with the new job spec
+		newJobNeeded = true
+	}
+
+	if newJobNeeded {
```
csrwng: If the job.DeletionTimestamp is not nil we should not need to call delete again

dgoodwin:
Addressed, thanks!

pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
```diff
@@ -183,6 +195,32 @@ func (r *ReconcileClusterDeprovisionRequest) Reconcile(request reconcile.Request
...
+	// Check if the job should be regenerated:
+	newJobNeeded := false
+	if existingJob.Annotations == nil {
+		newJobNeeded = true
+	} else if _, ok := existingJob.Annotations[jobHashAnnotation]; !ok {
+		// this job predates tracking the job hash, so assume we need a new job
+		newJobNeeded = true
+	}
+
+	if existingJob.Annotations[jobHashAnnotation] != jobHash {
```
csrwng: should be `else if`'d with the statement above so it benefits from the nil check

pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
```diff
@@ -183,6 +195,32 @@ func (r *ReconcileClusterDeprovisionRequest) Reconcile(request reconcile.Request
...
+		// delete the job so we get a fresh one with the new job spec
+		newJobNeeded = true
+	}
+
+	if newJobNeeded {
```
csrwng: Just style nit... less indentation with
`if newJobNeeded && existingJob.DeletionTimestamp == nil {`

dgoodwin: This was to avoid the Infof line at the end of the function which is inaccurate in this case.

csrwng: ack

PR #471, opened 2019-07-01
Switch to golang 1.12

csrwng:


twiest:
@csrwng is this still a wip?  If not, I'll +1 it.

csrwng:
@twiest thanks, I wanted it to let it build the images and see if there were any issues first before removing the WIP

csrwng:
@twiest looks like things are building fine... removing the WIP tag

PR #473, opened 2019-07-03
SyncSet: Add SecretReference to (Selector)SyncSet crd.

abutcher:
SyncSet CRD changes to allow specifying a list of SecretReferences to existing secrets on the hive cluster to sync to target clusters/secrets.

https://jira.coreos.com/browse/CO-331


pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -175,6 +219,16 @@ type SyncSetCommonSpec struct {
...
+
+	// SyncResources is the list of objects to sync from existing resources.
+	// +optional
+	SyncResources []SyncResource `json:"syncResources,omitempty"`
+
+	// SyncResourceApplyMode indicates if the SyncResource apply mode is "upsert" (default) or "sync".
+	// ApplyMode "upsert" indicates create and update.
+	// ApplyMode "sync" indicates create, update and delete.
+	// +optional
+	SyncResourceApplyMode SyncSetResourceApplyMode `json:"syncResourceApplyMode,omitempty"`
```
staebler: Why does the apply mode need to be different for the ones synced from existing resources? Is there a use case for applying one way from the raw data and another way from the existing resource within the same syncset?

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -65,6 +65,40 @@ type SyncObjectPatch struct {
...
 }
 
+// SyncResource represents an existing object to be synced
+type SyncResource struct {
+	Source SyncResourceSource `json:"source"`
+	Target SyncResourceTarget `json:"target"`
+}
+
+// SyncResourceSource describes a source object to sync to clusters SyncSet applies to
+type SyncResourceSource struct {
```
staebler: Why not use `ObjectReference` instead of defining new types? That is the general guidance from the API conventions. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#references-to-related-objects

dgoodwin: +1 for ObjectReference, although I actually envisioned that this would be specifically for Secrets. We already built a mechanism for generic objects, it's just a matter of Secret data. Do we think there's a strong driver to also allow the copy of generic resources? We'd then have two ways to do that.

pkg/apis/hive/v1alpha1/syncset_types.go
abutcher: I don't think we need a separate apply mode. I'll remove this one.

pkg/apis/hive/v1alpha1/syncset_types.go
abutcher: I've changed the `source` to be an `ObjectReference` but we still want to keep the idea of a target in the event that namespace/name will not be the same as they are currently, correct?

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -65,6 +65,23 @@ type SyncObjectPatch struct {
 	PatchType string `json:"patchType,omitempty"`
 }
 
+// SecretResource represents an existing secret object to be synced
+type SecretResource struct {
+	Source corev1.ObjectReference `json:"source"`
```
abutcher: Enforcing that this is a secret will be the job of the controller and result in an error sent to the associated `SyncSetInstance` status.

pkg/apis/hive/v1alpha1/syncset_types.go
staebler: We want to keep the `target` field, but it should be an `ObjectReference` too.

pkg/apis/hive/v1alpha1/syncset_types.go
staebler: Is the webhook validation not sophisticated enough to handle the enforcement? Or are you arguing that it is more appropriate for the controller to handle it?

abutcher: I'm not sure if the webhook validation is or isn't but I can look into it.

pkg/apis/hive/v1alpha1/syncset_types.go
abutcher: Okay 👍

pkg/apis/hive/v1alpha1/syncset_types.go
dgoodwin: The webhook validation is "hiveadmission" that Thomas put together. Benefit of using it would be that the caller immediately gets their API request rejected, rather than the API server accepting it and the error only showing up in condition if you went to look for it. Admission would probably be the best spot for this one, it cannot do any etcd loading/querying, but in this case we can check the type easily and reject if not a secret.

pkg/apis/hive/v1alpha1/validating-webhooks/selector_syncset_validating_admission_hook.go
```diff
@@ -161,6 +162,18 @@ func (a *SelectorSyncSetValidatingAdmissionHook) validateCreate(admissionSpec *a
 		}
 	}
 
+	if invalid, ok := checkSecretResources(newObject.Spec.SecretResources); !ok {
+		message := fmt.Sprintf("Failed validation: Invalid secret resource kind detected: %s. SecretResources may only contain references to kind: secret.", invalid)
+		contextLogger.Infof(message)
+		return &admissionv1beta1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
```
staebler: I would expect this to be `metav1.StatusReasonInvalid` and for the details to indicate which field was invalid. See https://github.com/kubernetes/apimachinery/blob/c0632ccbde11a13044c0f41b7e3c588f16ecda8b/pkg/api/errors/errors.go#L222-L245.

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -250,11 +246,15 @@ func checkValidPatchTypes(patches []hivev1.SyncObjectPatch) (string, bool) {
...
-func checkSecretReferences(secrets []hivev1.SecretReference) (string, bool) {
+func checkSecretReferences(secrets []hivev1.SecretReference) field.ErrorList {
+	errs := field.ErrorList{}
 	for _, secret := range secrets {
+		if len(secret.Source.Kind) == 0 {
+			errs = append(errs, field.Required(field.NewPath("kind"), "SecretReference.Source.Kind is required"))
+		}
 		if secret.Source.Kind != "secret" {
-			return secret.Source.Kind, false
+			errs = append(errs, field.Invalid(field.NewPath("kind"), secret.Source.Kind, fmt.Sprintf("Invalid SecretReference.Source kind detected: %s. SecretReferences may only contain references to kind: secret", secret.Source.Kind)))
```
staebler: This should be NotSupported instead of Invalid.

`field.NotSupported(fldPath.Index(i).Child("source").Child("kind"), secret.Source.Kind, []string{"secret"})`

Although, really, you need to check the entire GVK not just the kind. There could be another API group that has a kind of secret.

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -224,3 +245,43 @@ func checkValidPatchTypes(patches []hivev1.SyncObjectPatch) (string, bool) {
 	}
 	return "", true
 }
+
+func checkSecretReferences(secrets []hivev1.SecretReference) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for i, secret := range secrets {
+		sourcePath := field.NewPath("spec").Child("secretReferences").Index(i).Child("source")
```
staebler: Create a function to do the checks for the `ObjectReference` rather than duplicating it for source and target.

```
for i, secret := range secrets {
  validateSecretRef(secret.Source, fldPath.Index(i).Child("source")
  validateSecretRef(secret.Target, fldPath.Index(i).Child("target")
}
```

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -224,3 +245,43 @@ func checkValidPatchTypes(patches []hivev1.SyncObjectPatch) (string, bool) {
 	}
 	return "", true
 }
+
+func checkSecretReferences(secrets []hivev1.SecretReference) field.ErrorList {
```
staebler: Pass in the `field.Path` that points to the secrets being validated.

staebler: Prefer `validate` term over `check`.

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook_test.go
```diff
@@ -83,6 +85,88 @@ func TestSyncSetValidate(t *testing.T) {
...
+			operation:       admissionv1beta1.Update,
+			syncSet:         testSecretReferenceSyncSet(),
+			expectedAllowed: true,
+		},
+		{
+			name:      "Test invalid SecretReference kind create",
+			operation: admissionv1beta1.Create,
+			syncSet: func() *hivev1.SyncSet {
+				ss := testSecretReferenceSyncSet()
+				ss.Spec.SecretReferences[0].Source.Kind = "wrongkind"
```
staebler: Separate this into two distinct sub tests so that we verify that we are checking both `Source` and `Target` and not just one of the two.

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook_test.go
```diff
@@ -83,6 +85,88 @@ func TestSyncSetValidate(t *testing.T) {
...
+				return ss
+			}(),
+			expectedAllowed: false,
+		},
+		{
+			name:      "Test invalid SecretReference kind update",
+			operation: admissionv1beta1.Update,
+			syncSet: func() *hivev1.SyncSet {
+				ss := testSecretReferenceSyncSet()
+				ss.Spec.SecretReferences[0].Source.Kind = "wrongkind"
```
staebler: Separate into two subtests

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook_test.go
```diff
@@ -83,6 +85,88 @@ func TestSyncSetValidate(t *testing.T) {
...
+				return ss
+			}(),
+			expectedAllowed: false,
+		},
+		{
+			name:      "Test invalid SecretReference no name create",
+			operation: admissionv1beta1.Create,
+			syncSet: func() *hivev1.SyncSet {
+				ss := testSecretReferenceSyncSet()
+				ss.Spec.SecretReferences[0].Source.Name = ""
```
staebler: Two subtests

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook_test.go
```diff
@@ -83,6 +85,88 @@ func TestSyncSetValidate(t *testing.T) {
...
+				return ss
+			}(),
+			expectedAllowed: false,
+		},
+		{
+			name:      "Test invalid SecretReference no name update",
+			operation: admissionv1beta1.Update,
+			syncSet: func() *hivev1.SyncSet {
+				ss := testSecretReferenceSyncSet()
+				ss.Spec.SecretReferences[0].Source.Name = ""
```
staebler: Two subtests

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook_test.go
```diff
@@ -83,6 +85,88 @@ func TestSyncSetValidate(t *testing.T) {
...
+				return ss
+			}(),
+			expectedAllowed: false,
+		},
+		{
+			name:      "Test invalid SecretReference invalid fields set create",
+			operation: admissionv1beta1.Create,
+			syncSet: func() *hivev1.SyncSet {
+				ss := testSecretReferenceSyncSet()
+				ss.Spec.SecretReferences[0].Source.FieldPath = "dontset"
```
staebler: 4 subtests

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook_test.go
```diff
@@ -83,6 +85,88 @@ func TestSyncSetValidate(t *testing.T) {
...
+				return ss
+			}(),
+			expectedAllowed: false,
+		},
+		{
+			name:      "Test invalid SecretReference invalid fields set update",
+			operation: admissionv1beta1.Update,
+			syncSet: func() *hivev1.SyncSet {
+				ss := testSecretReferenceSyncSet()
+				ss.Spec.SecretReferences[0].Source.FieldPath = "dontset"
```
staebler: 4 subtests

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -224,3 +245,43 @@ func checkValidPatchTypes(patches []hivev1.SyncObjectPatch) (string, bool) {
...
+	allErrs := field.ErrorList{}
+	for i, secret := range secrets {
+		sourcePath := field.NewPath("spec").Child("secretReferences").Index(i).Child("source")
+		if len(secret.Source.Kind) == 0 {
+			allErrs = append(allErrs, field.Required(sourcePath.Child("kind"), "Kind is required"))
+		}
+		if secret.Source.Kind != "secret" {
```
staebler: This function should validate the Group as well, which must be empty.

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -224,3 +245,43 @@ func checkValidPatchTypes(patches []hivev1.SyncObjectPatch) (string, bool) {
...
+		}
+		if len(secret.Source.FieldPath) != 0 {
+			allErrs = append(allErrs, field.Invalid(sourcePath.Child("fieldPath"), secret.Source.FieldPath, "FieldPath should not be set"))
+		}
+		if len(secret.Source.UID) != 0 {
+			allErrs = append(allErrs, field.Invalid(sourcePath.Child("UID"), secret.Source.UID, "UID should not be set"))
+		}
+
+		targetPath := field.NewPath("spec").Child("secretReferences").Index(i).Child("target")
+		if len(secret.Target.Kind) == 0 {
```
staebler: How do we handle the API version of the target? Is there any concern that the target may not be able to accept the version sent from the controller?

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
abutcher: Which group should be validated? The `ObjectReference` doesn't have group information, right?

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
abutcher: Based on Friday convo, I believe we're deferring worrying about the `APIVersion` of the source not matching the target.

pkg/apis/hive/v1alpha1/validating-webhooks/syncset_validating_admission_hook.go
staebler: The group must be empty, which is the group for the core API.

pkg/apis/hive/v1alpha1/syncset_types.go
```diff
@@ -175,6 +185,10 @@ type SyncSetCommonSpec struct {
 	// Patches is the list of patches to apply.
 	// +optional
 	Patches []SyncObjectPatch `json:"patches,omitempty"`
+
+	// SecretReferences is the list of secrets to sync from existing resources.
+	// +optional
+	SecretReferences []SecretReference `json:"secretReferencees,omitempty"`
```
dgoodwin: "secretReferencees" typo in your json key here will cause problems in real world.

abutcher: Updated! Huge catch

PR #474, opened 2019-07-04
Add leader election for both operator and controllers.

dgoodwin:
Primarily to prevent any problems during upgrades when both new and old
versions of the code may be running.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -27,7 +27,9 @@ import (
 )
 
 const (
-	defaultLogLevel = "info"
+	defaultLogLevel         = "info"
+	hiveNamespace           = "hive"
+	leaderElectionConfigMap = "hive-controllers-leader"
 )
 
 type controllerManagerOptions struct {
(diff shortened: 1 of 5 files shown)


cmd/manager/main.go
```diff
@@ -56,7 +58,10 @@ func newRootCommand() *cobra.Command {
 
 			// Create a new Cmd to provide shared dependencies and start components
 			mgr, err := manager.New(cfg, manager.Options{
-				MetricsBindAddress: ":2112",
+				MetricsBindAddress:      ":2112",
+				LeaderElection:          true,
+				LeaderElectionNamespace: hiveNamespace,
+				LeaderElectionID:        leaderElectionConfigMap,
```
LalatenduMohanty: :+1:

cmd/manager/main.go
```diff
@@ -27,7 +27,9 @@ import (
 )
 
 const (
-	defaultLogLevel = "info"
+	defaultLogLevel         = "info"
+	hiveNamespace           = "hive"
+	leaderElectionConfigMap = "hive-controllers-leader"
```
LalatenduMohanty: WDYT about variable name  `leaderElectionConfigMapName` ?
```
	// LeaderElectionID determines the name of the configmap that leader election
	// will use for holding the leader lock.
	LeaderElectionID string
```

PR #479, opened 2019-07-12
Gather Cluster Logs on Install Failure

dgoodwin:
This change adds a new persistent volume for every install job, mounted to /logs. If an install fails, we will run the gather logs script and copy that tarball to the logs volume with a timestamp. Each attempt will result in a new tarball being added to the volume.

Failure to gather logs is not a terminal error and we just carry on with cleanup regardless.

The /logs volume is only used for the logs, we still use an emptyDir volume for the installer runtime files, I did not want to move these over to do concerns about cross contaminating between install attempts, and difficulties rolling out a change to use sub-dirs with past ClusterImageSets.

Once install completes, if successful and without restarts we will delete the PVC immediately. If there were restarts, we will delete the pvc after 7 days.

WARNING: extremely rough, just wanted to get something up for review before I'm out next week so I can start on feedback first thing Monday in a week. The code should work for gathering logs, cleanup is not yet tested.


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1234,6 +1251,49 @@ func (r *ReconcileClusterDeployment) deleteJobOnHashChange(existingJob, generate
 	return newJobNeeded, nil
 }
 
+// cleanupInstallLogPVC will immediately delete the PVC (should it exist) if the cluster was installed successfully, without retries.
+// If there were retries, it will delete the PVC if it has been more than 7 days since the job was completed.
+func (r *ReconcileClusterDeployment) cleanupInstallLogPVC(cd *hivev1.ClusterDeployment, existingJob *batchv1.Job, cdLog log.FieldLogger) error {
```
staebler: Don't forget to add reconcile tests that verify that the pvc is cleaned up in the scenarios where you expect it to be.

pkg/installmanager/installmanager.go
```diff
@@ -619,59 +664,54 @@ func (m *InstallManager) loadClusterDeployment() (*hivev1.ClusterDeployment, err
...
-		return err
+		return "", err
 	}
 
-	logTarball, err := runGatherScript(bootstrapIP, fetchLogsScriptTemplate, m)
+	logTarball, err := m.runGatherScript(bootstrapIP, fetchLogsScriptTemplate, m.WorkDir)
 	if err != nil {
 		m.log.Error("failed to run script for gathering logs")
-		return err
+		m.log.Warn("sleeping for a long long time")
```
staebler: Did you intend to leave this in here? Or is this a remnant from your testing?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1234,6 +1251,49 @@ func (r *ReconcileClusterDeployment) deleteJobOnHashChange(existingJob, generate
...
+			cdLog.WithError(err).Error("error deleting install logs PVC")
+			return err
+		}
+		return nil
+	}
+
+	// Otherwise, delete if more than 7 days have passed.
+	if time.Now().Sub(existingJob.Status.CompletionTime.Time) > (7 * 24 * time.Hour) {
+		cdLog.WithField("pvc", existingJob.Name).Info("deleting logs PersistentVolumeClaim for cluster that was installed after restarts more than 7 days ago")
+		err := r.Delete(context.TODO(), cd)
```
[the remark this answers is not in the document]
dgoodwin: Haha what the f. Lets just chalk this up to the 'untested' part of my PR description. :)

pkg/installmanager/installmanager.go
dgoodwin: Remnant of testing.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1234,6 +1251,49 @@ func (r *ReconcileClusterDeployment) deleteJobOnHashChange(existingJob, generate
...
+			return nil
+		}
+		cdLog.WithError(err).Error("error looking up install logs PVC")
+		return err
+	}
+
+	if cd.Status.InstallRestarts == 0 {
+		cdLog.WithField("pvc", existingJob.Name).Info("deleting logs PersistentVolumeClaim for installed cluster with no restarts")
+		err := r.Delete(context.TODO(), cd)
+		if err != nil {
```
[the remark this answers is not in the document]
dgoodwin: Don't really have a preference, will adjust.

dgoodwin:
>     2. The logs are only captured if bootstrapping fails. We also want to capture logs when bootstrapping is successful but the cluster never makes it to full installation.

I'm not following this one @staebler, the code will attempt to gather logs if there was an install failure of any kind, i.e. any non-zero response from openshift-install create cluster. Does this cover both scenarios? (I thought it did and the failure I'm simulating in this PR I think is in the post-bootstrap phase)

staebler:
> > ```
> > 2. The logs are only captured if bootstrapping fails. We also want to capture logs when bootstrapping is successful but the cluster never makes it to full installation.
> > ```
> 
> I'm not following this one @staebler, the code will attempt to gather logs if there was an install failure of any kind, i.e. any non-zero response from openshift-install create cluster. Does this cover both scenarios? (I thought it did and the failure I'm simulating in this PR I think is in the post-bootstrap phase)

On failure, the installmanager attempts to ssh into the bootstrap machine to run `install-gather.sh`. However, if the bootstrapping completed successfully, then there will not be a bootstrap machine any more. It will have been torn down by the installer.

dgoodwin:
Yikes, ok that sucks. I need to figure out a way to simulate that. I wonder if the installer-gather.sh script exists on the masters. Also curious where the "openshift-install gather" command fits in, I just noticed that.

staebler:
> Yikes, ok that sucks. I need to figure out a way to simulate that. I wonder if the installer-gather.sh script exists on the masters. Also curious where the "openshift-install gather" command fits in, I just noticed that.

Once the masters are up, we should be using `oc adm must-gather` to gather the logs.

dgoodwin:
/test e2e

This should be ready for review now, removing WIP.

LalatenduMohanty:
> One question about day-to-day using this. How will we be able to view the PV contents after the logs are collected/saved?

@dgoodwin I also have the same question.

Makefile
```diff
@@ -190,6 +190,12 @@ docker-push:
 buildah-build:
 	$(SUDO_CMD) buildah bud --tag ${IMG} .
 
+# Build the code locally and build+push an image with the local binaries rather than doing a full in-container compile.
+.PHONY: buildah-dev-push
+buildah-dev-push: build
+	$(SUDO_CMD) buildah bud -f Dockerfile.dev --tag ${IMG} .
+	$(SUDO_CMD) buildah push ${IMG}
+
```
LalatenduMohanty: IMO we should add this information (how a developer can use it) in the developing documentation.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -508,6 +510,28 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+				if err != nil {
+					cdLog.WithError(err).Error("error creating pvc")
+					return reconcile.Result{}, err
+				}
+			} else if err != nil {
+				cdLog.WithError(err).Error("error getting persistent volume claim")
+				return reconcile.Result{}, err
+			}
+		}
+
```
LalatenduMohanty: Can we move this to a function? At this point clusterdeployment_controller.go reconcile function has more code than what should be in a single function. May be moving it to a function would enable us to add tests for it.

Makefile
dgoodwin: I switched the developing Hive docs ref to this fully, I can't really think of a reason not to unless someone didn't want to setup a Go env and instead wanted to do all their compiling in a Docker container, but that would be a pretty terrible idea.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Will do.

dgoodwin: > > One question about day-to-day using this. How will we be able to view the PV contents after the logs are collected/saved?
> 
> @dgoodwin I also have the same question.

I've been just rsh'ing into the current install pod. If however a later install completed I think we'll want to do up a script like: https://blog.openshift.com/transferring-files-in-and-out-of-containers-in-openshift-part-3/

dgoodwin:
Updated.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: That article is for copying files *into* the PV, not for copying files out of it. To copy files out, the easiest thing is to do `oc cp ${pod-name}:${path-to-file}`. The pod used could be the current install pod. Or it may be helpful to create a script that starts a pod specifically for mounting the PV to get the desired logs.

pkg/apis/hive/v1alpha1/hiveconfig_types.go
```diff
@@ -31,6 +31,9 @@ type HiveConfigSpec struct {
 	// a cluster deployment(if specified), with precedence given to the contents of the pull secret for the cluster deployment.
 	// +optional
 	GlobalPullSecret *corev1.LocalObjectReference `json:"globalPullSecret,omitempty"`
+
+	// FailedProvisionConfig is used to configure settings related to handling provision failures.
+	FailedProvisionConfig *FailedProvisionConfig `json:"failedProvisionConfig,omitempty"`
```
staebler: I recommend that we not make this field a pointer and `omitempty`. As it is, the default behavior is to *not* gether logs. I think we want the default behavior to be that logs will be gathered. Along those lines, I would like to see the `failedProvisionConfig.gatherLogs` field changed to `failedProvisionConfig.noLogsGather`.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -63,6 +63,7 @@ const (
 	jobHashAnnotation                     = "hive.openshift.io/jobhash"
 	firstTimeInstallAnnotation            = "hive.openshift.io/first-time-install"
 	deleteAfterAnnotation                 = "hive.openshift.io/delete-after" // contains a duration after which the cluster should be cleaned up.
+
```
staebler: Remove extra blank line.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -600,6 +609,29 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileClusterDeployment) reconcileLogsPVC(cd *hivev1.ClusterDeployment, pvc *corev1.PersistentVolumeClaim, cdLog log.FieldLogger) error {
```
staebler: This is not a good name for this function. It is not doing any reconciling. It is just creating the PVC if it does not exist.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -600,6 +609,29 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+	if err := controllerutil.SetControllerReference(cd, pvc, r.scheme); err != nil {
```
staebler: Move this down to the section where the PVC is created. There is no need to do this except in that case.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -480,7 +482,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			hivemetrics.GetClusterDeploymentType(cd)).Set(
 			time.Since(cd.CreationTimestamp.Time).Seconds())
 
-		job, err := install.GenerateInstallerJob(
+		job, pvc, err := install.GenerateInstallerJob(
```
staebler: I would prefer to see the PVC created separately from the installer job and the PVC passed in as a parameter to `GenerateInstallerJob`. The creation of the PVC is really a separate action from creating the job. It has a direct relationship with the clusterdeployment and only a passing relationship with the job--in much the same way that the serviceaccount is creating separately from the job.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -600,6 +609,29 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
 }
 
+func (r *ReconcileClusterDeployment) reconcileLogsPVC(cd *hivev1.ClusterDeployment, pvc *corev1.PersistentVolumeClaim, cdLog log.FieldLogger) error {
+	if err := controllerutil.SetControllerReference(cd, pvc, r.scheme); err != nil {
+		cdLog.WithError(err).Error("error setting controller reference on pvc")
+		return err
+	}
+
+	existingPVC := &corev1.PersistentVolumeClaim{}
+	err := r.Get(context.TODO(), types.NamespacedName{Name: pvc.Name, Namespace: pvc.Namespace}, existingPVC)
```
staebler:
```
switch err := r.Get(...); {
case err == nil:
   // PVC already exists
   return nil
case !apierrors.IsNotFound(err):
   // Error fetching PVC
   return err
}
// Continue with creating the PVC
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1246,6 +1278,52 @@ func (r *ReconcileClusterDeployment) deleteJobOnHashChange(existingJob, generate
...
+		cdLog.WithField("pvc", pvc.Name).Info("deleting logs PersistentVolumeClaim for installed cluster with no restarts")
+		if err := r.Delete(context.TODO(), pvc); err != nil {
+			cdLog.WithError(err).Error("error deleting install logs PVC")
+			return err
+		}
+		return nil
+	}
+
+	// Otherwise, delete if more than 7 days have passed.
+	if time.Since(existingJob.Status.CompletionTime.Time) > (7 * 24 * time.Hour) {
```
staebler: I would like to see this as a test of the status condition on the clusterdeployment rather than the job completion time. It is possible that the job may have been deleted by the user in the 7 days since it completed. The controller needs to be resilient to this possibility. The other option is to immediately delete the PVC when the job no longer exists. In either case, the controller cannot rely on getting the name of the PVC from the name of the job.

This is another item in the column in favor of separating the creation of the PVC from the creation of the job.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1246,6 +1278,52 @@ func (r *ReconcileClusterDeployment) deleteJobOnHashChange(existingJob, generate
...
+		cdLog.WithField("pvc", pvc.Name).Info("deleting logs PersistentVolumeClaim for cluster that was installed after restarts more than 7 days ago")
+		if err := r.Delete(context.TODO(), pvc); err != nil {
+			cdLog.WithError(err).Error("error deleting install logs PVC")
+			return err
+		}
+		return nil
+	}
+
+	cdLog.WithField("pvc", pvc.Name).Debug("preserving logs PersistentVolumeClaim for cluster with install restarts for 7 days")
```
staebler: Are we relying on periodic resyncs of the clusterdeployment to pick up when the PVC has expired? Does kubebuilder do anything to swallow syncs when the revision of the clusterdeployment has not changed?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Yeah this is the approach I wanted from that article, spinning up a dummy pod so we have something we can reliably use to cp or rsh, as the install may have completed successfully after a few failures.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: We don't seem to have any such condition, I think I will add a Status.InstalledTimestamp, have been thinking about recording some other timestamps such as DNS resolution start/stop times for better metrics.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: We are relying on it yes, if kubebuilder is doing so then we need to rethink some things, I would be quite surprised if that were the case though. Reconciliation should be consistent and periodic even if nothing has changed in the desired state.

dgoodwin:
Updated, tests fixed, added functionality for oc adm must-gather attempt if bootstrap node attempt fails.

pkg/installmanager/installmanager.go
```diff
@@ -619,14 +667,56 @@ func (m *InstallManager) loadClusterDeployment() (*hivev1.ClusterDeployment, err
...
+// gatherLogs will attempt to gather logs after a failed install. First we attempt
+// to gather logs from the bootstrap node. If this fails, we may have made it far enough
+// to teardown the bootstrap node, in which case we then attempt to gather with
+// 'oc adm must-gather', which would gather logs from the cluster's API itself.
+// If neither succeeds we do not consider this a fatal error,
+// we're just gathering as much information as we can and then proceeding with cleanup
+// so we can re-try.
+func (m *InstallManager) gatherLogs(cd *hivev1.ClusterDeployment) {
+
+	err := m.gatherBootstrapNodeLogs(cd)
```
staebler: I would like the two sections of this function to have similar structure. Also, we do not need a warning log unless both methods fail. We can log the error associated with each method as an info log to cover scenarios where bootstrap logging may be failing when it should be succeeding.
```
if err := m.gatherBootstrapNodeLogs(cd); err != nil {
  m.log.Info(...)
  return
} else {
  m.log.WithError(err).Warn(...)
}
if err := m.gatherClusterLogs(cd); err != nil {
  m.log.Info(...)
  return
} else {
  m.log.WithError(err).Warn(...)
}
m.log.Warn("could not fetch logs") // or some such message
```

pkg/operator/hive/hive.go
```diff
@@ -54,6 +53,17 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		hiveDeployment.Spec.Template.Spec.Containers[0].Env = append(hiveDeployment.Spec.Template.Spec.Containers[0].Env, hiveImageEnvVar)
 	}
 
+	// By default we will try to gather logs on failed installs:
+	skipGatherLogs := "false"
```
staebler:
```
skipGatherLogs := strconv.FormatBool(instance.Spec.FailedProvisionConfig.SkipGatherLogs)
```
But, we can probably do away with the `skipGatherLogs` variable.

pkg/install/generate.go
```diff
@@ -248,7 +283,7 @@ func GenerateInstallerJob(
 
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        GetInstallJobName(cd),
+			Name:        installJobName,
```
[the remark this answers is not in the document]
dgoodwin: Yeah just a remnant from when we were re-using the job name on the pvc. Removed.

pkg/installmanager/installmanager.go
```diff
@@ -437,8 +464,28 @@ func runUninstaller(clusterName, region, infraID string, logger log.FieldLogger)
...
+	m.log.Info("running openshift-install create manifests")
+	err := m.runOpenShiftInstallCommand([]string{"create", "manifests", "--dir", m.WorkDir})
+	if err != nil {
+		m.log.WithError(err).Error("error generating installer assets")
+		return err
+	}
+
+	// If the failure test annotation is set, write a bogus resource into the manifests which will
+	// cause a late failure in the install process, enough that we can gather logs from the cluster.
+	if _, ok := cd.Annotations[hiveInstallFailureTestAnnotation]; ok {
```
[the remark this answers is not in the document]
dgoodwin: Went with the latter but with err == nil.

dgoodwin:
Updated.

pkg/installmanager/installmanager.go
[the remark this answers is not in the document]
staebler: Yes, `err == nil` is much better. :)

pkg/installmanager/installmanager.go
```diff
@@ -682,19 +680,20 @@ func isGatherLogsEnabled() bool {
...
 		return
 	}
 
-	m.log.WithError(err).Warn("error fetching logs from bootstrap node")
-
 	if err := m.gatherClusterLogs(cd); err != nil {
-		m.log.WithError(err).Warn("error fetching logs with oc adm must-gather")
+		m.log.WithError(err).Info("error fetching logs with oc adm must-gather")
 	} else {
 		m.log.Info("successfully ran oc adm must-gather")
```
staebler: Need a `return` here.

dgoodwin: Whoops, fixed!

PR #480, opened 2019-07-12
Add Velero Backup When ClusterDeployment Changes

twiest:
- [x] Add watch for changes to clusterdeployment objects
- [x] Add reconcile that checks all clusterdeployment objects in a given
      namespace
- [x] Create Velero backup object on ClusterDeployment change so that a
      backup is taken immediately.
- [x] Add unit tests around core functionality
- [x] Test that backup object is created correctly.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -24,6 +24,8 @@ import (
 	openshiftapiv1 "github.com/openshift/api/config/v1"
 	_ "github.com/openshift/generic-admission-server/pkg/cmd"
 	awsprovider "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsproviderconfig/v1beta1"
+
+	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
 )
 
 const (
@@ -87,6 +89,10 @@ func newRootCommand() *cobra.Command {
 			}
 
 			if err := crv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+				log.Fatal(err)
+			}
+
+			if err := velerov1.AddToScheme(mgr.GetScheme()); err != nil {
 				log.Fatal(err)
 			}
 
--- a/config/rbac/manager_role.yaml
+++ b/config/rbac/manager_role.yaml
@@ -226,3 +226,17 @@ rules:
   - patch
   - list
   - watch
+- apiGroups:
+  - hive.openshift.io
+  resources:
+  - clusterdeployments
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - velero.io
+  resources:
+  - backups
+  verbs:
+  - create
--- a/pkg/controller/add_velerobackup.go
+++ b/pkg/controller/add_velerobackup.go
@@ -1,0 +1,7 @@
+package controller
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	// TODO: Uncomment this line once card CO-488 is completed (adding HiveConfig option for this).
+	//AddToManagerFuncs = append(AddToManagerFuncs, velerobackup.Add)
+}
--- a/pkg/controller/syncset/syncset_controller.go
+++ b/pkg/controller/syncset/syncset_controller.go
@@ -2,8 +2,6 @@ package syncset
 
 import (
 	"context"
-	"crypto/md5"
-	"encoding/json"
 	"fmt"
 
 	log "github.com/sirupsen/logrus"
@@ -44,7 +42,7 @@ func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 		Client:      controllerutils.NewClientWithMetricsOrDie(mgr, controllerName),
 		scheme:      mgr.GetScheme(),
 		logger:      log.WithField("controller", controllerName),
-		computeHash: getHash,
+		computeHash: controllerutils.GetChecksumOfObject,
 	}
 	return r
 }
@@ -464,12 +462,4 @@ func syncSetInstanceNameForSyncSet(cd *hivev1.ClusterDeployment, syncSet *hivev1
 func syncSetInstanceNameForSelectorSyncSet(cd *hivev1.ClusterDeployment, selectorSyncSet *hivev1.SelectorSyncSet) string {
 	syncSetPart := helpers.GetName(selectorSyncSet.Name, "selector-syncset", validation.DNS1123SubdomainMaxLength-validation.DNS1123LabelMaxLength)
 	return fmt.Sprintf("%s-%s", cd.Name, syncSetPart)
-}
-
-func getHash(data interface{}) (string, error) {
-	b, err := json.Marshal(data)
-	if err != nil {
-		return "", err
-	}
-	return fmt.Sprintf("%x", md5.Sum(b)), nil
 }
--- a/pkg/controller/test/clusterdeployment/clusterdeployment.go
+++ b/pkg/controller/test/clusterdeployment/clusterdeployment.go
@@ -1,0 +1,94 @@
+package clusterdeployment
+
+import (
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+)
+
+// Option defines a function signature for any function that wants to be passed into BuildClusterDeployment
+type Option func(*hivev1.ClusterDeployment)
+
+// Checksum defines a function signature for checksumming objects.
+type Checksum func(*hivev1.ClusterDeployment) (string, error)
+
+// BuildClusterDeployment runs each of the functions passed in to generate a cluster deployment.
+func BuildClusterDeployment(opts ...Option) *hivev1.ClusterDeployment {
+	retval := &hivev1.ClusterDeployment{}
+	for _, o := range opts {
+		o(retval)
+	}
+
+	return retval
+}
+
+// WithName sets the ClusterDeployment.Name field when building a clusterdeployment with BuildClusterDeployment.
+func WithName(name string) Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		clusterDeployment.Name = name
+	}
+}
+
+// WithNamePostfix appends the string passed in to the ClusterDeployment.Name field when building a clusterdeployment with BuildClusterDeployment.
+func WithNamePostfix(postfix string) Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		clusterDeployment.Name = clusterDeployment.Name + "-" + postfix
+	}
+}
+
+// WithNamespace sets the ClusterDeployment.Namespace field when building a clusterdeployment with BuildClusterDeployment.
+func WithNamespace(namespace string) Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		clusterDeployment.Namespace = namespace
+	}
+}
+
+// WithAnnotationsSet ensures that ClusterDeployment.Annotations is not nil.
+func WithAnnotationsSet() Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		// Only set if Nil (don't wipe out existing)
+		if clusterDeployment.Annotations == nil {
+			clusterDeployment.Annotations = map[string]string{}
+		}
+	}
+}
+
+// WithEmptyAnnotations ensures that annotations is set and that map is empty.
+func WithEmptyAnnotations() Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		// Wipe out existing so that it's actually empty.
+		clusterDeployment.Annotations = map[string]string{}
+	}
+}
+
+// WithEmptyBackupChecksum ensures that the checksum annotation is empty.
+func WithEmptyBackupChecksum() Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		// Case where annotations don't exist, so checksum doesn't exist.
+		if clusterDeployment.ObjectMeta.Annotations == nil {
+			return
+		}
+
+		// Case where Annotations exist, -and- checksum annotation exists.
+		delete(clusterDeployment.Annotations, controllerutils.LastBackupAnnotation)
+
+		// If that was the only entry, then we should nil out annotations (this makes comparisons with FakeClient work).
+		if len(clusterDeployment.Annotations) == 0 {
+			clusterDeployment.Annotations = nil
+		}
+	}
+}
+
+// WithBackupChecksum sets the object's checksum attribute to the correct value.
+func WithBackupChecksum(checksum Checksum) Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		// In order for the checksum to be correct, the checksum needs to be calculated AFTER all other changes have been made
+		checksum, err := checksum(clusterDeployment)
+		if err != nil {
+			// On error, panic as this should never happen.
+			panic(err)
+		}
+
+		WithAnnotationsSet()(clusterDeployment)
+		clusterDeployment.Annotations[controllerutils.LastBackupAnnotation] = checksum
+	}
+}
--- a/pkg/controller/utils/constants.go
+++ b/pkg/controller/utils/constants.go
@@ -1,0 +1,6 @@
+package utils
+
+const (
+	// LastBackupAnnotation is the Kubernetes Annotation key that holds the object checksum from the last backup.
+	LastBackupAnnotation = "hive.openshift.io/last-backup-checksum"
+)
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -1,7 +1,9 @@
 package utils
 
 import (
+	"crypto/md5"
 	"encoding/json"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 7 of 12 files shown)


pkg/controller/backup/clusterdeployment_test.go
```diff
@@ -0,0 +1,415 @@
...
+			expectedError: nil,
+		},
+		{
+			name:                 "Checksum error case",
+			getHashOfObjectsFunc: getHashOfObjectsFuncReturnsError,
+			clusterDeployments: []*hivev1.ClusterDeployment{
+				changedClusterDeployment().ClusterDeployment,
+			},
+			existingObjects: []runtime.Object{},
+			//expectedError:              utilerrors.NewAggregate([]error{utilerrors.NewAggregate([]error{errDefault})}),
```
joelddiaz: old dev code?

pkg/controller/utils/clusterdeploymentbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// WithNamespace sets the namespace of the cluster deployment
+func (n *ClusterDeploymentBuilder) WithNamespace(namespace string) *ClusterDeploymentBuilder {
+	n.ClusterDeployment.Namespace = namespace
+	return n
+}
+
+// WithBackupChecksum sets the backup checksum for the clusterdeployment using the preveiously set checksum function.
+func (n *ClusterDeploymentBuilder) WithBackupChecksum() *ClusterDeploymentBuilder {
```
joelddiaz: is there ever a case where you would normally do additional modifications to the object after calling WithBackupChecksum() (besides test cases)?

pkg/controller/backup/backup_controller.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// For logging, we need to see when the reconciliation loop starts and ends.
+	contextLogger.Info("reconciling backups and Hive object changes")
+	defer func() {
+		dur := time.Since(start)
+		hivemetrics.MetricControllerReconcileTime.WithLabelValues(controllerName).Observe(dur.Seconds())
+		contextLogger.WithField("elapsed", dur).Info("reconcile complete")
+	}()
+
+	clusterDeploymentBackup := newClusterDeploymentBackup(r)
```
LalatenduMohanty: @twiest Is there any specific reason function `newClusterDeploymentBackup` is not part of `ReconcileBackup` struct i.e. `r. newClusterDeploymentBackup()`

pkg/controller/utils/clusterdeploymentbuilder.go
twiest: Changes can definitely be made, but the checksum will then be incorrect.

This means that the object will be considered "changed since last backup".

Did that answer your question?  Are you asking for a change here?

pkg/controller/utils/velero.go
```diff
@@ -0,0 +1,42 @@
...
+		"pods",
+		"jobs.batch",
+	}
+)
+
+// CreateVeleroBackupObjectForNamespace creates a Velero Backup object for the namespace specified.
+// The Backup options are set specifically for Hive object backups.
+// DO NOT use this function call for any other type objects as it may not back them up correctly.
+func CreateVeleroBackupObjectForNamespace(client client.Client, namespace string) error {
+	backupName := fmt.Sprintf("backup-%v-%v", namespace, uuid.NewUUID())
```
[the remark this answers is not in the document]
twiest: We can definitely do a datestamp instead of a UUID. Worst case scenario would be when we go create the velero backup object we'd get an error. We can then return a reconcile result with a delay / requeue.

Is that what we should do?

pkg/controller/utils/clusterdeploymentbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type clusterDeploymentChecksumFunc func(clusterDeployment *hivev1.ClusterDeployment) (string, error)
+
+// ClusterDeploymentBuilder is a utility to build ClusterDeployments given specific requirements.
+// It's especialy useful in cases where simply defining a ClusterDeployment struct isn't sufficient
+// (e.g. when declaring a variable that needs to checksum attribute set.)
+type ClusterDeploymentBuilder struct {
+	ClusterDeployment *hivev1.ClusterDeployment
+	ChecksumFunc      clusterDeploymentChecksumFunc
```
staebler: Why not make this a parameter to `WithBackupChecksum` rather than including it in the builder type?

pkg/controller/utils/clusterdeploymentbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+	// * these two lines MUST be next to eachother
+	// * the checksum needs to be calculated AFTER all other changes have been made
+	checksum, err := n.ChecksumFunc(n.ClusterDeployment)
+
+	if err != nil {
+		n.Errors = append(n.Errors, err)
+		return n
+	}
+
+	n.WithAnnotationsSet().ClusterDeployment.ObjectMeta.Annotations[LastBackupAnnotation] = checksum
```
staebler: `ObjectMeta` can be omitted.

pkg/controller/utils/clusterdeploymentbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+// WithEmptyBackupChecksum ensures that the cluster deployment's backup checksum is not set.
+// It attempts to emulate kube annotation behavior.
+func (n *ClusterDeploymentBuilder) WithEmptyBackupChecksum() *ClusterDeploymentBuilder {
+	// Case where annotations don't exist, so checksum doesn't exist.
+	if n.ClusterDeployment.ObjectMeta.Annotations == nil {
+		return n
+	}
+
+	// Case where Annotations exist, but checksum annotation doesn't exist.
+	if _, ok := n.ClusterDeployment.ObjectMeta.Annotations[LastBackupAnnotation]; !ok {
```
staebler: This check is not needed. The subsequent delete will be a no-op if the annotation is not in the map.

pkg/controller/backup/clusterdeployment.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return r.checksumFunc(metaCopy, clusterDeployment.Spec)
+}
+
+func (r *ClusterDeploymentBackup) updateLastBackupChecksum(clusterDeployments []*hivev1.ClusterDeployment) error {
+	errors := []error{}
+
+	for _, clusterDeployment := range clusterDeployments {
+		clusterDeployment, err := controllerutils.
+			NewClusterDeploymentBuilderWithClusterDeployment(r.checksum, clusterDeployment).
```
staebler: I am still not 100% sold on this approach here. The language makes it sounds like a new clusterdeployment is being built rather than the existing clusterdeployment being modified. And it is bit wordy compared to something like the following--which also has the benefit of being pretty clear about what it is doing.

```
controllerutils.SetBackupChecksum(clusterDeployment, r.checksum)
```

pkg/controller/utils/clusterdeploymentbuilder.go
staebler: One thing you could do here to ensure that `WithBackupChecksum` is the last thing called in the chain is to have it return a different object.

```
func (n *ClusterDeploymentBuilder) WithBackupChecksum() *ClusterDeploymentChecksummedBuilder {
   ...
}

type ClusterDeploymentChecksummedBuilder struct {
  builder *ClusterDeployemntBuilder
}

func (n *ClusterDeploymentChecksummedBuilder) GenerateClusterDeployment (...) {
  return n.builder.GenerateClusterDeployment()
}
```

joelddiaz: no change requested. just making sure i understand what's going on.

pkg/controller/utils/velero.go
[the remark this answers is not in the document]
joelddiaz: don't know if the retry code is worth it just to be able to have timestamped names. maybe just going with the UUID is enough, and some queries with piping to jq can sort out each backup objects creation date fields.

@staebler WDYT?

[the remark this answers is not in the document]
staebler: I like timestamp better. I think it makes it easier for an end user to deal with if they are sortable by name.

I also think it is worthwhile to place limiting on how rapidly backups are created. I don't know if it is helpful to create backups so rapidly that two backups may have the same timestamp. For what it's worth, having a backup resource would help with that.

pkg/controller/backup/clusterdeployment.go
```diff
@@ -0,0 +1,169 @@
...
+	cd, ok := a.Object.(*hivev1.ClusterDeployment)
+	if !ok {
+		// Wasn't a ClusterDeployment, bail out. This should not happen.
+		r.reconciler.logger.Errorf("Error converting MapObject.Object to ClusterDeployment. Value: %+v", a.Object)
+		return []reconcile.Request{}
+	}
+
+	contextLogger := addClusterDeploymentLoggerFields(r.reconciler.logger, cd)
+
+	hasChanged, err := r.hasClusterDeploymentChanged(cd)
```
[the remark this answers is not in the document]
twiest: You are correct. This was intended specifically to not do a reconcile unless the watched object has actually changed.

It is unlikely to have multiple cluster deployments in the same namespace. 

Note, however, that the intention is that we'll eventually reconcile all changes for all hive objects in the given namespace.

I'm happy to remove this as it's a pre-optimization and we can always add it back in if we determine that the reconcile is too heavy.

pkg/controller/utils/clusterdeploymentbuilder.go
twiest: It was mostly so that the 90% correct objects could specify it once and then all of the subsequent calls wouldn't need it specified.

I'm fine changing it if you feel strongly it should be the other way around.

pkg/controller/backup/clusterdeployment.go
```diff
@@ -0,0 +1,169 @@
...
+	contextLogger := addClusterDeploymentLoggerFields(r.reconciler.logger, clusterDeployment)
+
+	newChecksum, err := r.checksum(clusterDeployment)
+	if err != nil {
+		contextLogger.WithError(err).Info("error calculating checksum of cluster deployment")
+		return false, err
+	}
+
+	// Make sure newChecksum is different than the old checksum.
+	oldChecksum := clusterDeployment.Annotations[controllerutils.LastBackupAnnotation]
```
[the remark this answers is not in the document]
twiest: That's an interesting idea. We went through several designs, but I don't believe we considered anything like that.

I'm fine changing the design, but I'd prefer to run it by @dgoodwin and @csrwng first.

I'd also like to have some time to think it through to see if I can poke holes in the new design.

If we decide to go that route, would you mind if I implement that as a separate PR or would you prefer to block this PR until that decision is made?

pkg/controller/backup/clusterdeployment.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, clusterDeployment := range clusterDeployments {
+		clusterDeployment, err := controllerutils.
+			NewClusterDeploymentBuilderWithClusterDeployment(r.checksum, clusterDeployment).
+			WithBackupChecksum().
+			GenerateClusterDeployment()
+
+		if err != nil {
+			aggregate, ok := err.(utilerrors.Aggregate)
+			if ok {
+				// Unpack the aggregate error so that we don't return an aggregate of an aggregate.
```
staebler: You can add the aggregate here and then `Flatten` before returning from the function.

pkg/controller/add_backup.go
```diff
@@ -0,0 +1,7 @@
+package controller
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	// TODO: Uncomment this line once card CO-488 is completed (adding HiveConfig option for this).
```
LalatenduMohanty: Are you planning to complete `CO-488` in this PR or in a subsequent PR?

pkg/controller/backup/backup_controller.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Get all ClusterDeployments that have a different hash than the last backup hash.
+	changedCDs, err := clusterDeploymentBackup.getChangedInNamespace(request.Namespace)
```
LalatenduMohanty: I find the function name `getChangedInNamespace` difficult with respect to what it is doing. Suggestion `getModifiedClusterDeployments`

pkg/controller/backup/backup_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	logger log.FieldLogger
+}
+
+// Reconcile ensures that all Hive object changes have corresponding Velero backup objects.
+// ClusterDeployment.Spec.Config.Machines
+// +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterdeployments,verbs=get;list;watch
+// +kubebuilder:rbac:groups=velero.io,resources=backups,verbs=create
+func (r *ReconcileBackup) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	start := time.Now()
+	contextLogger := addReconcileRequestLoggerFields(r.logger, request)
```
LalatenduMohanty: IMO we can just call the `contextLogger` as `logger` as it will be more easy for the eyes to read. If someone is interested to know how logger is working they can find how it is implemented.

pkg/controller/utils/velero.go
```diff
@@ -0,0 +1,42 @@
...
+		// pods to BE BACKED UP (not excluded)!!!
+		"pods",
+		"jobs.batch",
+	}
+)
+
+// CreateVeleroBackupObjectForNamespace creates a Velero Backup object for the namespace specified.
+// The Backup options are set specifically for Hive object backups.
+// DO NOT use this function call for any other type objects as it may not back them up correctly.
+func CreateVeleroBackupObjectForNamespace(client client.Client, namespace string) error {
```
LalatenduMohanty: IMO this function should not be in ` pkg/controller/utils/`. This is perfect for keeping in `pkg/controller/backup`. Let me know if I am missing anything.

pkg/controller/utils/velero.go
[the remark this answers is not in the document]
LalatenduMohanty: +1 for timestamp in the name.

pkg/controller/backup/backup_controller.go
```diff
@@ -0,0 +1,119 @@
...
+		return reconcile.Result{}, err
+	}
+
+	if len(changedCDs) == 0 {
+		// Nothing changed, so nothing to back up (i.e. don't create a Velero backup object).
+		return reconcile.Result{}, nil
+	}
+
+	// There are changes that need to be backed up.
+	err = controllerutils.CreateVeleroBackupObjectForNamespace(r.Client, request.Namespace)
```
LalatenduMohanty: Just `CreateVeleroBackupObject` as function name sounds better. As the code passes the namespace as argument it is easy to understand it is per namespace.

pkg/controller/backup/clusterdeployment.go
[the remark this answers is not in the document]
staebler: In light of the expectation that we the controller will be monitoring multiple types of objects, I am fine with leaving it as is.

pkg/controller/backup/clusterdeployment.go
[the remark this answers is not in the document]
staebler: We can tackle that idea as a separate PR. It should be easy to switch over from using annotations to using a separate resource if we decide to go that route.

pkg/controller/utils/clusterdeploymentbuilder.go
staebler: I'll reserve judgement on this until I see how it plays out with the switch to functional options.

pkg/controller/backup/backup_controller.go
staebler: Is the intention here that this function will eventually return all of the hive objects in the namespace that have been modified? I do agree that `getChangedInNamespace` is a bit jolting. I too prefer something like `getModifiedClusterDeployments`--with an eventual name of `getModifiedHiveObjects`.

pkg/controller/backup/backup_controller.go
staebler: +1

pkg/controller/backup/backup_controller.go
twiest: So the name fully is "clusterDeploymentBackup.getChangedInNamespace". I didn't want to repeat ClusterDeployment in the function name because it was in the struct name.

That being said, clusterDeploymentBackup as an object type is likely going away, which would necessitate renaming the function to be more specific.

The intention is for there to be a function that gets all modified hive objects in a namespace. But that function will likely call the ClusterDeployment specific call, so this function will still exist in some form or another.

pkg/controller/utils/velero.go
staebler: I agree that `pkg/controller/utils` is not a good place for this. I suggest either keep it in `pkg/controller/backup` or, if it is something not tied to controller behavior, then put it in `pkg/backup`. As it is, this file and its own function are small enough to make it hard to justify its own distinct package.

pkg/controller/backup/backup_controller.go
staebler: +1

pkg/controller/add_backup.go
twiest: Subsequent PR.

pkg/controller/backup/backup_controller.go
twiest: I believe I got the pattern of calling the newer logger `contextLogger` from @dgoodwin. 

I believe the idea is to distinguish it from the standard logger.

I'm fine changing this, but want @dgoodwin to weigh in first.

dgoodwin: I'm big on contextual loggers but can't recall using it as a var name, though it's entirely possible. Typically we try to come up with a short name that also reflects what the context the logger is carrying, ex cdLog, or in this case perhaps nsLog. "logger" would be fine as well but slight pref to indicating what's in there.

twiest:
@dgoodwin @LalatenduMohanty I rebased and I believe I've addressed all of the feedback. 

Please review.

PR #484, opened 2019-07-29
Add velero backup option to HiveConfig

twiest:
- [x] Add an option to HiveConfig that conditionally enables the
      velero backup controller
- [x] Test that the HiveConfig option flows down to the Hive manager


pkg/apis/hive/v1alpha1/hiveconfig_types.go
```diff
@@ -41,6 +46,21 @@ type HiveConfigStatus struct {
 	AggregatorClientCAHash string `json:"aggregatorClientCAHash,omitempty"`
 }
 
+// BackupConfig contains settings for the Velero backup integration.
+type BackupConfig struct {
+	// VeleroBackup specifies configuration for Velero backup integration.
+	// +optional
+	Velero VeleroBackupConfig `json:"velero,omitempty"`
```
dgoodwin: Tiny mismatch with your var name and the godoc comment here.

pkg/operator/hive/hive.go
```diff
@@ -61,6 +61,11 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		hiveDeployment.Spec.Template.Spec.Containers[0].Env = append(hiveDeployment.Spec.Template.Spec.Containers[0].Env, dnsServersEnvVar)
 	}
 
+	if instance.Spec.Backup.Velero.Enabled {
+		hLog.Infof("Velero Backup Enabled.")
+		hiveDeployment.Spec.Template.Spec.Containers[0].Command = append(hiveDeployment.Spec.Template.Spec.Containers[0].Command, "--velero-backup")
+	}
```
dgoodwin: This is slightly different from how we're passing most things from hiveconfig to hive via env vars. Any thoughts on which approach is better and when/where we should draw the line between the two? I kinda like the CLI as opposed to env vars as I can run the binary with --help to see options, rather than having env vars embedded all over the code.

Right now the env vars are: global pull secret, zone check dns servers.

Curious what the team thinks. cc @staebler @abutcher @LalatenduMohanty @joelddiaz

pkg/apis/hive/v1alpha1/hiveconfig_types.go
twiest: Whoops!  I wonder why the linter didn't catch that.

pkg/operator/hive/hive.go
twiest: I'm happy to do it either way. I wasn't trying to "blaze a new trail" here, I just didn't see the other way.

I'll do it the env var way for now and then if we decide to do CLI options, I'll add this back.

dgoodwin: Ok FWIW I kinda like it. :)

cmd/manager/main.go
```diff
@@ -29,9 +29,10 @@ import (
 )
 
 const (
-	defaultLogLevel         = "info"
-	hiveNamespace           = "hive"
-	leaderElectionConfigMap = "hive-controllers-leader"
+	defaultLogLevel           = "info"
+	defaultEnableVeleroBackup = false
```
LalatenduMohanty: IMO the variable name should be `enableVeleroBackup `. both `default` and `enable` sounds odd.

pkg/controller/utils/constants.go
```diff
@@ -3,4 +3,7 @@ package utils
 const (
 	// LastBackupAnnotation is the Kubernetes Annotation key that holds the object checksum from the last backup.
 	LastBackupAnnotation = "hive.openshift.io/last-backup-checksum"
+
+	// VeleroBackupEnvVar is the name of the environment variable used to tell the controller manager to enable velero backup integration.
+	VeleroBackupEnvVar = "HIVE_VELERO_BACKUP"
```
LalatenduMohanty: It is used outside of controllers i.e. in  pkg/operator/hive/hive.go . IMO it should go to https://github.com/openshift/hive/blob/master/pkg/constants/constants.go

cmd/manager/main.go
twiest: Actually, this constant is never used. Thanks for pointing it out, I'll remove it.

pkg/controller/utils/constants.go
twiest: You're right, thanks for the comment. Will move.

twiest:
@dgoodwin @LalatenduMohanty I believe I've addressed all of the feedback. PTAL.

pkg/operator/hive/hive.go
LalatenduMohanty: I like the env var way as does not tie us with the CLI strongly. However I am not very sure which will be maintainable long term.

PR #486, opened 2019-07-30
Support empty global pull secret instead of absent.

dgoodwin:
The nil check did not work if globalPullSecret was specified but empty,
which has turned up as a helpful thing to support for AppSRE templates
across environments.

Also check for an empty name and skip global pull secret if this is the
case.


jmelis:
This seems to be what we need. LGTM

PR #487, opened 2019-08-01
Changing retry log to warning for ingress certficate bundle

LalatenduMohanty:


pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -163,7 +163,7 @@ func (r *ReconcileRemoteClusterIngress) Reconcile(request reconcile.Request) (re
 	// can't proceed if the secret(s) referred to doesn't exist
 	certBundleSecrets, err := r.getIngressSecrets(rContext)
 	if err != nil {
-		rContext.logger.WithError(err).Error("will need to retry until able to find all certBundle secrets")
+		rContext.logger.Warningf("will need to retry until able to find all certBundle secrets : %v", err)
```
LalatenduMohanty: This log should not be an error as it communicates that we need to retry till we get the cert information for remote ingress controller.

PR #489, opened 2019-08-02
Create Velero backup object on syncset change

twiest:
- [x] Convert the velerobackup controller to use a generic object.
      This will allow for a massive amount of code reduction per hive type.
- [x] Convert unit tests to new generic way.
- [x] Create a Velero backup object on syncset change
- [x] Update unit tests to test existing functionality with syncsets
- [x] Test the code functionality against kubernetes


pkg/controller/test/syncset/syncset.go
```diff
@@ -0,0 +1,94 @@
+package syncset
```
LalatenduMohanty: IMO the test directory should be pkg/test. Keeping it under `controller` suggests that it should be used within `controllers`.

config/rbac/manager_role.yaml
```diff
@@ -234,6 +234,14 @@ rules:
...
   - list
   - watch
+- apiGroups:
+  - hive.openshift.io
+  resources:
+  - syncsets
+  verbs:
+  - get
+  - list
+  - watch
```
LalatenduMohanty: I can see we have three instances of 
```
- apiGroups:
  - hive.openshift.io
  resources:
  - syncsets
```
IMO we should have a single instance as it will be easier for humans.

pkg/controller/test/syncset/syncset.go
twiest: I honestly don't care where it lands, but didn't we agree in the conversation to put it under `controllers`?

twiest: We talked and apparently I misunderstood the original consensus of putting this under `pkg/test`. I'll move this.

config/rbac/manager_role.yaml
twiest: We decided that a Jira card would be created to decide how to handle roles in the future (manually created / maintained vs autogenerated by kubebuilder).

In the mean time, we will continue with our current pattern of annotating each controller with the specific permissions that it needs, which will generate multiple entries for the same resource in the role.

twiest:
@dgoodwin @joelddiaz @LalatenduMohanty  I believe I've addressed all feedback, PTAL.

pkg/controller/velerobackup/syncset_test.go
```diff
@@ -0,0 +1,280 @@
...
+				Object: &hivev1.ClusterDeployment{},
+			},
+			expectedRequests: []reconcile.Request{},
+		},
+		{
+			name: "SyncSet unchanged since last backup case",
+			mapObject: handler.MapObject{
+				Object: testsyncset.BuildSyncSet(unchangedSyncSetBase()),
+			},
+			expectedRequests: changedSyncSetReconcileRequests(),
```
joelddiaz: the function name changedSyncSetReconcileRequest() can probably be improved since it's the same for the changed an unchanged case

pkg/controller/velerobackup/syncset_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// Arrange
+			r := newReconcileBackupUsingFakeClient(test.existingObjects, test.getHashOfObjectsFunc)
+
+			// Act
+			actualError := r.updateSyncSetLastBackupChecksum(test.syncSets)
+			actualSyncSetList := &hivev1.SyncSetList{}
+			r.List(context.TODO(), actualSyncSetList, client.InNamespace(namespace))
```
joelddiaz: no error checking. are you relying on the assert a few lines below to catch unexpected values in actualSyncSetList?

pkg/controller/velerobackup/test_helpers.go
```diff
@@ -0,0 +1,113 @@
...
+		return false
+	}
+
+	// Test for aggregate errors (and make sure they contain equal errors)
+	aggregateErrorType := reflect.TypeOf(utilerrors.NewAggregate([]error{errDefault}))
+	if lhsType == aggregateErrorType {
+		return areAggregateErrorsEqualType(lhs.(utilerrors.Aggregate), rhs.(utilerrors.Aggregate))
+	}
+
+	return false
```
joelddiaz: return true here?

pkg/controller/velerobackup/test_helpers.go
```diff
@@ -0,0 +1,113 @@
+package velerobackup
```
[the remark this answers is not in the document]
twiest: `helpers_test.go` it is.

pkg/controller/velerobackup/syncset_test.go
twiest: Yes, since the test and code are fully controlling what's in fakeclient, so the assert using the expectedSyncSets should always be accurate. If it errors, it will not be accurate and the test will fail.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -77,9 +86,15 @@ type ReconcileBackup struct {
...
+type modifiedObjects struct {
+	clusterDeployments []*hivev1.ClusterDeployment
+	syncSets           []*hivev1.SyncSet
+	hasModified        bool
+}
+
 // Reconcile ensures that all Hive object changes have corresponding Velero backup objects.
-// ClusterDeployment.Spec.Config.Machines
 // +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterdeployments,verbs=get;list;watch
+// +kubebuilder:rbac:groups=hive.openshift.io,resources=syncsets,verbs=get;list;watch
```
joelddiaz: for updating the checksum, doesn't that need more than get/list/watch?

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -112,16 +127,58 @@ func (r *ReconcileBackup) Reconcile(request reconcile.Request) (reconcile.Result
...
+	}
+
+	if len(modified.clusterDeployments) > 0 {
+		modified.hasModified = true
+	}
+
+	// Get all SyncSets that have a different hash than the last backup hash.
+	modified.syncSets, err = r.getModifiedSyncSetsInNamespace(namespace)
+	if err != nil {
+		rrLogger.WithError(err).Error("error finding changed cluster deployments in namespace")
```
joelddiaz: s/cluster deployments/syncsets/

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -112,16 +127,58 @@ func (r *ReconcileBackup) Reconcile(request reconcile.Request) (reconcile.Result
...
+
+	if len(modified.syncSets) > 0 {
+		modified.hasModified = true
+	}
+
+	return modified, nil
+}
+
+func (r *ReconcileBackup) updateAllLastBackupChecksum(modified *modifiedObjects, rrLogger *log.Entry) {
 	// If the above is successful, save this objects new checksum to the "lastBackupChecksum" annotation.
```
joelddiaz: left-over comment from refactoring? what is "above" the first line in the func?

pkg/test/syncset/syncset.go
```diff
@@ -0,0 +1,94 @@
...
+		if len(syncSet.Annotations) == 0 {
+			syncSet.Annotations = nil
+		}
+	}
+}
+
+// WithBackupChecksum sets the object's checksum attribute to the correct value.
+func WithBackupChecksum(checksum Checksum) Option {
+	return func(syncSet *hivev1.SyncSet) {
+		// In order for the checksum to be correct, the checksum needs to be calculated AFTER all other changes have been made
```
joelddiaz: minor nit. i'd prefer if this comment was up with the function doc three lines up (so that i can see this in VSCode when calling the func :) )

twiest:
@joelddiaz ok, I've addressed your feedback. PTAL.

joelddiaz:
looks fine to me with the changes, but i understand @staebler might have some feedback, so we'll wait on that

twiest:
@staebler @joelddiaz I believe I've addressed all of your feedback. PTAL.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -130,6 +136,36 @@ func addReconcileRequestLoggerFields(logger log.FieldLogger, request reconcile.R
 	})
 }
 
+func (r *ReconcileBackup) registerHiveObjectWatches(c controller.Controller) error {
+	// Watch for changes to ClusterDeployment
```
staebler: At the top of the file,
```
var (
	typesToWatch = []runtime.Object{
		&hivev1.ClusterDeployment{},
		&hivev1.SyncSet{},
	}
)
```

Here,
```
func (r *ReconcileBackup) registerHiveObjectWatches(c controller.Controller) error {
	for _, t := range typesToWatch {
		if err := c.Watch(&source.Kind{Type: t.DeepCopyObject()}, &handler.EnqueueRequestsFromMapFunc{
			// Queue up the NS for this Hive Object
			ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
				return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: a.Meta.GetNamespace()}}}
			}),
		}); err != nil {
			return err
		}
	}
	return nil
}

```

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -130,6 +136,36 @@ func addReconcileRequestLoggerFields(logger log.FieldLogger, request reconcile.R
...
+		return err
+	}
+
+	// If we get here, then no error occurred.
+	return nil
+}
+
+func (r *ReconcileBackup) hiveObjectWatchHandler(a handler.MapObject) []reconcile.Request {
+	// Queue up the NS for this Hive Object
+	retval := []reconcile.Request{
```
staebler: Since there is so much nesting here, it may be a little easier to read like the following.
```
reqs := make([]reconcile.Request, 1)
reqs[0].Namespace = a.Meta.GetNamespace()
return reqs
```

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -153,3 +189,103 @@ func createVeleroBackupObject(client client.Client, namespace string) error {
 
 	return client.Create(context.TODO(), backup)
 }
+
+func (r *ReconcileBackup) getAllHiveObjectsInNamespace(namespace string) ([]*hiveObject, error) {
+	nsLogger := addNamespaceLoggerFields(r.logger, namespace)
+
+	allNSObjs := []runtime.Object{}
+
+	allClusterDeployments := &hivev1.ClusterDeploymentList{}
```
staebler: As with the watch code, make this generic by doing something like the following.
```
	for _, t := range typesToList {
		listObj := t.DeepCopyObject()
		if err := r.List(context.TODO(), listObj, client.InNamespace(namespace)); err != nil {
			nsLogger.WithError(err).Errorf("Couldn't get list of %s", t.GetObjectKind().GroupVersionKind().Kind)
			return nil, err
		}
		list, err := meta.ExtractList(listObj)
		if err != nil {
			// log something
			return nil, err
		}
		allNSObjs = append(allNSObjs, list...)
	}
```

Put a `typesToList` at the top of the file.
```
var (
	typesToList = []runtime.Object{
		&hivev1.ClusterDeploymentList{},
		&hivev1.SyncSetList{},
	}
)
```

staebler: The include is 	"k8s.io/apimachinery/pkg/api/meta".

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -153,3 +189,103 @@ func createVeleroBackupObject(client client.Client, namespace string) error {
...
+		}
+
+		allHiveObjects = append(allHiveObjects, hobj)
+	}
+
+	return allHiveObjects, nil
+}
+
+func (r *ReconcileBackup) getModifiedHiveObjectsInNamespace(namespace string) ([]*hiveObject, error) {
+	allHiveObjects, err := r.getAllHiveObjectsInNamespace(namespace)
```
staebler: I don't see a compelling need to build up the slice of all hive objects. Let's process each hive object as we get it instead.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -153,3 +189,103 @@ func createVeleroBackupObject(client client.Client, namespace string) error {
...
+	if err != nil {
+		nsLogger.WithError(err).Error("Couldn't get list of SyncSets")
+		return nil, err
+	}
+
+	for i := range allSyncSets.Items {
+		allNSObjs = append(allNSObjs, &allSyncSets.Items[i])
+	}
+
+	allHiveObjects := []*hiveObject{}
```
staebler: When you know the size of the slice that you need, it is better to do this like the following.
```
allHiveObjects := make([]*hiveObject{}, len(allNSObjs))
for i, nsObj := range allNSObjs{
  hobj, err := newHiveObject(...)
  allHiveObjects[i] = hobj
}
```

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,100 @@
...
+		}, nil
+	}
+
+	return nil, fmt.Errorf("Unknown Type")
+}
+
+func (h *hiveObject) hasChanged() (bool, error) {
+	hobjLogger := addHiveObjectLoggerFields(h.logger, h)
+
+	newChecksum, err := h.calculateChecksum()
```
staebler: Calculating the checksum should not raise an error. If it does, then this could cause a block where the Valero backup is never created because of one resource for which the checksum could not be calculated. Instead, if there is an error calculating the checksum, then a default checksum should be used instead.

pkg/test/syncset/syncset.go
```diff
@@ -0,0 +1,95 @@
...
+	retval := &hivev1.SyncSet{}
+	for _, o := range opts {
+		o(retval)
+	}
+
+	return retval
+}
+
+// WithName sets the SyncSet.Name field when building a syncset with BuildSyncSet.
+func WithName(name string) Option {
```
staebler: We should consider making some of these common options available in a generic way instead of needing to copy these for each specific type.

In the pkg/test/generic package,
```
type Option func(obj metav1.Object)

func WithName(name string) Option {
  return func(obj metav1.Object) {
    obj.SetName(name)
  }
}
```
In, for example, pkg/test/syncset
```
func Generic(opt generic.Option) Option {
  return func(syncSet *hivev1.SyncSet) {
    opt(syncSet)
  }
}
```
To use,
```
syncset.Build(syncset.Generic(generic.WithName("test-name")))
```

It makes the caller code a bit more dense at the benefit of not having to copy blocks of code in the callee side.

pkg/controller/velerobackup/helpers_test.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+var (
+	errDefault = fmt.Errorf("Not a real error")
+	errStatus  = errors.StatusError{
+		ErrStatus: metav1.Status{},
+	}
+)
+
+func assertHiveObjectsEqual(t *testing.T, expected, actual []*hiveObject) {
```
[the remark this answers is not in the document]
twiest: It does. With `assert.Equal(t, test.expectedHiveObjects, actualHiveObjects)`, I get:

```
Error:      	Not equal: 
            	expected: []*velerobackup.hiveObject{(*velerobackup.hiveObject)(0xc00003ca00)}
            	actual  : []*velerobackup.hiveObject{(*velerobackup.hiveObject)(0xc0002a1840)}
           	            	
          	Diff:
Test:       	TestReconcile/1_unchanged_ClusterDeployment_case
```

pkg/controller/velerobackup/helpers_test.go
```diff
@@ -0,0 +1,179 @@
...
+	obj, err := newHiveObject(ro, r.checksumFunc, r.logger)
+
+	if err != nil {
+		panic("This should never happen since we control the test data.")
+	}
+
+	return obj
+}
+
+func newReconcileBackupUsingFakeClient(existingObjects []runtime.Object, getHashOfObjectsFunc controllerutils.ChecksumOfObjectsFunc) *ReconcileBackup {
```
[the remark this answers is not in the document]
twiest: Yes. Will fix. :)

pkg/controller/velerobackup/hiveobject_test.go
```diff
@@ -0,0 +1,185 @@
...
+			name:             "Valid ClusterDeployment Checksum",
+			checksumObject:   testclusterdeployment.Build(unchangedClusterDeploymentBase()),
+			expectedChecksum: testclusterdeployment.Build(unchangedClusterDeploymentBase()).Annotations[controllerutils.LastBackupAnnotation],
+			checksumFunc:     controllerutils.GetChecksumOfObjects,
+		},
+		{
+			name:             "Error from Checksum func",
+			checksumObject:   testclusterdeployment.Build(unchangedClusterDeploymentBase()),
+			expectedChecksum: errChecksum,
+			checksumFunc:     getHashOfObjectsFuncReturnsError,
```
staebler: Rather than having to mock out the checksum function, rework this test so that the test cases specifies the meta and spec as inputs and use a spec value that cannot be marshalled. Then you can get rid of the `checksumFunc` parameter entirely.

For example,
```
tests := []struct{
  name string
  meta *metav1.Object
  spec interface{}
  expectedChecksum string
}{
  {
    name: "Error from Checksum func",
    meta: &metav1.ObjectMeta{},
    spec: math.Inf(1),
    expectedChceksum: errChecksum,
  },
}
```

pkg/controller/velerobackup/helpers_test.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func changedClusterDeploymentBase() testclusterdeployment.Option {
+	return func(clusterDeployment *hivev1.ClusterDeployment) {
+		clusterDeployment.Name = "someclusterdeployment-changed"
+		clusterDeployment.Namespace = namespace
+	}
+}
+
+func emptyReconcileBackup() *ReconcileBackup {
```
staebler: The only use for this function is to get a logger to use. So, instead, have a func that returns the logger. This func can then be used in `fakeClientReconcileBackup`, too.

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,113 @@
...
+	case nil:
+		return nil, fmt.Errorf("Can't convert nil to a HiveObject")
+	}
+
+	return nil, fmt.Errorf("Unknown Type: %T", object)
+}
+
+func (h *hiveObject) hasChanged() bool {
+	if h.checksum == errChecksum {
+		// The checksum function errored. In this case we assume the object has NOT changed.
```
staebler: I think we want to treat `errChecksum` identically to all other checksums. If the calculated checksum and the annotation checksum are different, then the object has changed, regardless of whether on of those checksums is `errChecksum`.

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return checksum
+}
+
+func (h *hiveObject) getChecksumAnnotation() string {
+	tmpmeta, err := meta.Accessor(h.object)
+	if err != nil {
+		// This should never happen because we should always be an object.
+		panic(err)
```
staebler: Understanding that this is a logic error and not a runtime error, we still should not panic. Instead, log it as an error and return an empty string, treating it as though the annotation is not set.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -130,6 +136,36 @@ func addReconcileRequestLoggerFields(logger log.FieldLogger, request reconcile.R
...
+func (r *ReconcileBackup) hiveObjectWatchHandler(a handler.MapObject) []reconcile.Request {
+	// Queue up the NS for this Hive Object
+	retval := []reconcile.Request{
+		{NamespacedName: types.NamespacedName{
+			Namespace: a.Meta.GetNamespace(),
+		}}}
+	return retval
+}
+
 // createVeleroBackupObjectForNamespace creates a Velero Backup object for the namespace specified.
```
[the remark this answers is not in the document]
twiest: Fixed

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	annotations := tmpmeta.GetAnnotations()
+	return annotations[controllerutils.LastBackupAnnotation]
+}
+
+func (h *hiveObject) setChecksumAnnotation(checksum string) {
+	// Not checking err because we are guaranteed to be an object.
```
staebler: Always check the error even though it is a logic error for the error to not be nil. If the error is not nil, log it as an error and return from the function.

Alternatively, store a pointer to the meta for the object in the hiveObject. We are already grabbing it during `newHiveObject`, so we may as well store it.

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	// Actually set the checksum
+	annotations[controllerutils.LastBackupAnnotation] = checksum
+	tmpmeta.SetAnnotations(annotations)
+}
+
+func addHiveObjectLoggerFields(logger log.FieldLogger, hobj *hiveObject) *log.Entry {
+	return logger.WithFields(log.Fields{
+		// "name":      hobj.meta.Name,
```
staebler: If you store a pointer to the meta in the hiveObject, then you can get the name and namespace here.

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,113 @@
...
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+
+	// Actually set the checksum
+	annotations[controllerutils.LastBackupAnnotation] = checksum
+	tmpmeta.SetAnnotations(annotations)
+}
+
+func addHiveObjectLoggerFields(logger log.FieldLogger, hobj *hiveObject) *log.Entry {
```
staebler: Why is this function not a member of hiveObject? If you do make it a member, then you can shorten the name to `addLoggerFields`.

staebler: This function should return a `log.FieldLogger` rather than ` *log.Entry`.

pkg/controller/velerobackup/hiveobject_test.go
```diff
@@ -0,0 +1,185 @@
...
+func TestSetChecksum(t *testing.T) {
+	tests := []struct {
+		name             string
+		hiveObject       *hiveObject
+		expectedChecksum string
+	}{
+		{
+			name:             "Valid ClusterDeployment Checksum",
+			hiveObject:       ro2ho(testclusterdeployment.Build(changedClusterDeploymentBase())),
+			expectedChecksum: testclusterdeployment.Build(changedClusterDeploymentBase(), testclusterdeployment.WithBackupChecksum(defaultChecksumFunc)).Annotations[controllerutils.LastBackupAnnotation],
```
staebler: This is too complicated for what you are trying to test. Instead, do something like this. Note that you do not need to test each different type individually as setting the checksum does not depend upon the type.
```
func TestSetChecksum(t *testing.T) {
  ro := &hivev1.ClusterDeployment{
      Annotations: map[string]string{controllerutils.LastBackupAnnotation: "old-checksum"},
    },
  }
  hiveObject := &hiveObject{
    object: ro,
    checksum: "new-checksum"
  }
  hiveObject.setChecksumAnnotation()
  actualChecksum := ro.Annotations[controllerutils.LastBackupAnnotation]
  assert.Equals(t, "new-checksum", actualChecksum, "unexpected checksum")
}
```

pkg/controller/velerobackup/hiveobject.go
```diff
@@ -0,0 +1,113 @@
...
+	tmpmeta, err := meta.Accessor(h.object)
+	if err != nil {
+		// This should never happen because we should always be an object.
+		panic(err)
+	}
+	annotations := tmpmeta.GetAnnotations()
+	return annotations[controllerutils.LastBackupAnnotation]
+}
+
+func (h *hiveObject) setChecksumAnnotation(checksum string) {
```
staebler: This function should not take a parameter. The hiveObject already knows the checksum to set.

pkg/controller/velerobackup/hiveobject.go
twiest: In the last review, didn't we get rid of the pointer to the meta object in hiveObject?

I'm fine adding it back, just want to know for the future when I should store things and when I should compute things.

staebler: Yeah, you are right. I did recommend getting rid of it. I think you'll find that I am not always consistent. Let's leave it using `meta.Accessor`. That is the better approach. Or play around with storing the runtime object as a meta.Object instead of as a runtime.Object. That would obviate the need for meta.Accessor.

twiest: If we were to only store `meta.Object`, I don't think I can hand that to `client.Update` ?

Or is there a way to get the `runtime.object` back out of `meta.Object` (this doesn't seem likely to me as it'd be a circular reference)?

twiest:
@staebler I believe I've addressed your latest feedback. PTAL.

I need to test the latest changes against a real server. I'll remove the WIP once that is done.

Feel free to suggest changes even with the WIP in place as the testing is the only thing left.

pkg/controller/velerobackup/hiveobject.go
staebler: You can cast from the `meta.Object` to the `runtime.Object`.
```
func TestObjectMeta(t *testing.T) {
	obj := &hivev1.ClusterDeployment{}
	objAsMeta, err := meta.Accessor(obj)
	if !assert.NoError(t, err, "could not convert to meta.Object") {
		return
	}
	_, ok := objAsMeta.(runtime.Object)
	assert.True(t, ok, "could not case to runtime.Object")
}
```

staebler: Also, for what it is worth, we can probably forego using `meta.Accessor` and just cast from `runtime.Object` to `meta.Object`.

twiest: That is super interesting. Got rid of `meta.Accessor`. Let me know if I implemented it correctly.

pkg/controller/velerobackup/velerobackup_controller_test.go
```diff
@@ -21,21 +21,21 @@ func TestReconcile(t *testing.T) {
...
-		expectedClusterDeployments []*hivev1.ClusterDeployment
+		name                string
+		request             reconcile.Request
+		existingObjects     []runtime.Object
+		expectedResult      reconcile.Result
+		expectedError       error
+		expectedHiveObjects []*hiveObject
 	}{
 		{
 			name: "Empty Namespace case",
```
staebler: Appending "case" at the end of the test case name is redundant. It unnecessarily makes the name of the subtest longer.

pkg/controller/velerobackup/hiveobject_test.go
```diff
@@ -0,0 +1,176 @@
...
+			checksumFunc:    controllerutils.GetChecksumOfObjects,
+		},
+		{
+			name:            "Changed SyncSet",
+			hiveObject:      ro2ho(testsyncset.Build(changedSyncSetBase())),
+			expectedChanged: true,
+			checksumFunc:    controllerutils.GetChecksumOfObjects,
+		},
+		{
+			name:            "Error from Checksum func",
```
staebler: This subtest is not testing what the name implies it should be testing. This test is redundant with the "Unchanged SyncSet" subtest.

staebler:
I am not seeing how we are handling the case where an object is deleted from the namespace. The expected behavior, I would think, is to have that trigger a new backup. However, I think that that change would be ignored since we do not keep a list of the objects included in the backup.

twiest:
@staebler that's a good point. Let's discuss in Monday's standup. Are you ok if this PR goes in without that fix and we'll do a different one with whatever solution is decided?

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -151,5 +182,75 @@ func createVeleroBackupObject(client client.Client, namespace string) error {
...
+		hobjLogger := hobj.addLoggerFields()
+
+		hobj.setChecksumAnnotation()
+		obj := hobj.object
+
+		err := r.Update(context.TODO(), obj)
+		if err != nil {
+			hobjLogger.WithError(err).Error("Failed update.")
+			errors = append(errors, err)
+			continue // It errored, so jump to next loop.
```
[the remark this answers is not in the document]
twiest: @staebler as it sits now, we're not re-queuing the namespace immediately (we just log the error).

Do you still think we should return on first update failure?

staebler:
> @staebler that's a good point. Let's discuss in Monday's standup. Are you ok if this PR goes in without that fix and we'll do a different one with whatever solution is decided?

That's fine. This may be good motivation to use the separate CRD to track the state of the backup.

pkg/controller/velerobackup/velerobackup_controller.go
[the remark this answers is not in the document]
staebler: The controller does not explicitly re-queue the namespace. But the resources that the controller is watching will have changed, which will cause the namespace to be re-queued.

That assumes that there was success in setting the annotation for at least one resource. In the scenario where the update of the first resource fails, then the namespace will not be re-queued. That leaves open a small possibility of error in the scenario where a resource is updated, is backed up, fails to get its annotation set, and is then reverted back to its original state. Then the backup controller will think that the backup contains the original state of the resource instead of the intermediate state so will not trigger a new backup.

pkg/controller/velerobackup/velerobackup_controller_test.go
```diff
@@ -139,14 +161,168 @@ func TestReconcile(t *testing.T) {
...
+			// Arrange
+			r := fakeClientReconcileBackup(test.existingObjects)
+
+			// Act
+			actualError := r.updateHiveObjectsLastBackupChecksum(test.hiveObjects)
+			actualHiveObjects, err := r.getHiveObjectsInNamespace(namespace, filterAllObjects)
+
+			// Assert
+			assert.NoError(t, err)
+			assertHiveObjectsEqual(t, test.expectedHiveObjects, actualHiveObjects)
```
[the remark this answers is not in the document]
twiest: @staebler I'm fuzzy on what you're asking here.

Are you asking me to recreate the logic of `getHiveObjectsInNamespace` in the test code and return `runtime.Objects` instead of `hiveObjects`?

What's the problem with using non-test code to simplify / minimize test code?

I understand it's an implementation detail, but it's also a utility. I'm not understanding when it's ok to use a non-test code utility vs recreating the logic in test code.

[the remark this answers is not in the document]
staebler: My issue is that the component being tested is `ReconcileBackup`. The `getHiveObjectsInNamespace` function is a big portion of what that component does. If you rely on that function to determine what the expected output is, then you are coupling the expected output to the behavior of the unit being tested. If there are errors in that function, it may result in tests that report success when they should fail.

Give me a few minutes, and I will post up a sample of what I am thinking would be a cleaner test.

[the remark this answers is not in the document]
staebler:
```
func TestUpdateLastBackupChecksum(t *testing.T) {

	tests := []struct {
		name              string
		hiveObjects       []*hiveObject
		existingObjects   []runtime.Object
		expectedChecksums []string
		expectedError     error
	}{
		{
			name:              "No HiveObjects case",
			expectedChecksums: []string{},
		},
		{
			name: "1 changed ClusterDeployment case",
			hiveObjects: []*hiveObject{
				{
					object:   testclusterdeployment.Build(changedClusterDeploymentBase()),
					checksum: "new-checksum",
				},
			},
			existingObjects: []runtime.Object{
				testclusterdeployment.Build(changedClusterDeploymentBase()),
			},
			expectedChecksums: []string{
				"new-checksum",
			},
		},
		{
			name: "Multiple changed HiveObjects case",
			hiveObjects: []*hiveObject{
				{
					object:   testclusterdeployment.Build(changedClusterDeploymentBase()),
					checksum: "new-cd-checksum",
				},
				{
					object:   testsyncset.Build(changedSyncSetBase()),
					checksum: "new-syncset-checksum",
				},
			},
			existingObjects: []runtime.Object{
				testclusterdeployment.Build(changedClusterDeploymentBase()),
				testsyncset.Build(changedSyncSetBase()),
			},
			expectedChecksums: []string{
				"new-cd-checksum",
				"new-syncset-checksum",
			},
		},
		{
			name: "List error case",
			hiveObjects: []*hiveObject{
				{
					object:   testclusterdeployment.Build(changedClusterDeploymentBase()),
					checksum: "new-checksum",
				},
			},
			expectedChecksums: []string{},
			expectedError:     utilerrors.NewAggregate([]error{&errStatus}),
		},
	}

	for _, test := range tests {
			// Arrange
			r := fakeClientReconcileBackup(test.existingObjects)
			for _, ho := range test.hiveObjects {
				ho.logger = r.logger
			}

			// Act
			actualError := r.updateHiveObjectsLastBackupChecksum(test.hiveObjects)

			// Assert
			actualChecksums := make([]string, len(test.existingObjects))
			for i := range test.existingObjects {
				obj := test.existingObjects[i].DeepCopyObject()
				meta := obj.(metav1.Object)
				err := r.Get(context.TODO(), types.NamespacedName{Name: meta.GetName(), Namespace: meta.GetNamespace()}, obj)
				assert.NoError(t, err, "unexpected error fetching object")
				actualChecksums[i] = meta.GetAnnotations()[controllerutils.LastBackupAnnotation]
			}
			assert.Equal(t, test.expectedChecksums, actualChecksums, "unexpected checksums")
			assertErrorsAreEqualType(t, test.expectedError, actualError)
		})
	}
}
```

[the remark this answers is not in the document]
staebler: The preceding test validates that the `updateHiveObjectsLastBackupChecksum` function updates the checksum of each of the runtime objects composed in the hive objects passed to the function.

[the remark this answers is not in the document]
staebler: I would also accept the following.
1) Create a utility function separate from `ReconcileBackup`. This function should takes a client, a slice of type lists, and a predicate function to filter desired runtime objects and return a slice of runtime objects.
2) Create unit tests for the new utility function.
3) Have `getHiveObjectsInNamespace` call the new utility function.
4) Use the new utility function in the unit tests of `ReconcileBackup`.
However, even with that, the `ReconcileBackup` would be comparing runtime objects and not hive objects.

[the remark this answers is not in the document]
twiest: Ok, yeah, I decided to do that. That feels cleaner to me. Please let me know if I implemented it how you were thinking.

twiest:
@staebler ok, latest feedback has been addressed. PTAL.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -151,5 +175,84 @@ func createVeleroBackupObject(client client.Client, namespace string) error {
...
+	return nil
+}
+
+func addNamespaceLoggerFields(logger log.FieldLogger, namespace string) log.FieldLogger {
+	return logger.WithField("namespace", namespace)
+}
+
+type predicateType func(obj runtime.Object) bool
+
+func getRuntimeObjects(client client.Client, typesToList []runtime.Object, predicate predicateType) ([]runtime.Object, error) {
```
staebler: Sorry, I did not think through my suggestion for this function thoroughly enough. This function should take a namespace as a parameter so that the call to `client.List` can do the namespace filtering. I was expecting the predicate to mimic what you had before where where there is one predicate that gets all of the changed object and one that gets all of the objects. But now that I think about it, that is not practical as it requires converting to a `hiveObject` prior to check whether the checksum has changed. So, the predicate is not necessary.

pkg/controller/velerobackup/velerobackup_controller_test.go
```diff
@@ -139,14 +161,219 @@ func TestReconcile(t *testing.T) {
...
+		{
+			name: "1 changed ClusterDeployment",
+			hiveObjects: []*hiveObject{
+				ro2ho(testclusterdeployment.Build(changedClusterDeploymentBase())),
+			},
+			existingObjects: []runtime.Object{
+				testclusterdeployment.Build(changedClusterDeploymentBase()),
+			},
+			expectedObjects: []runtime.Object{
+				testclusterdeployment.Build(changedClusterDeploymentBase(), testclusterdeployment.Generic(generic.WithBackupChecksum(defaultChecksumFunc))),
```
staebler: I like this much better than what you had before. I would still prefer to see tests that were more explicit about what they were testing and less dependent upon the behavior of other components (namely, be explicit that I expect this particular checksum to be set, and do not rely on things like `newHiveObject` and `WithBackupChecksum` to determine the checksum behind the scenes). I will not hold up this PR for that wish, though.

twiest: I get your point, I lean more on the "make utils that make creating tests easy" side. Maybe this is a philosophical discussion that we should have after standup sometime.

pkg/controller/velerobackup/velerobackup_controller.go
twiest: Ah, that makes sense. Fixing.

twiest:
@staebler ok, latest feedback addressed. PTAL.

PR #490, opened 2019-08-06
Use the updated global pull secret from hive config

LalatenduMohanty:

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1486,7 +1486,14 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
 	}
 
 	// Check if global pull secret from env as it comes from hive config
-	globalPullSecret := os.Getenv("GLOBAL_PULL_SECRET")
+	globalPullSecretName := os.Getenv(constants.GlobalPullSecret)
+	var globalPullSecret string
+	if len(globalPullSecretName) != 0 {
+		globalPullSecret, err = controllerutils.LoadSecretData(r.Client, globalPullSecretName, constants.HiveNamespace, corev1.DockerConfigJsonKey)
+		if err != nil {
+			return "", errors.Wrap(err, "global pull secret could not be retrived")
+		}
+	}
 
 	switch {
 	case globalPullSecret != "" && localPullSecret != "":
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -45,6 +45,7 @@ const (
 	metadataName          = "foo-lqmsh-metadata"
 	sshKeySecret          = "ssh-key"
 	pullSecretSecret      = "pull-secret"
+	globalPullSecret      = "global-pull-secret"
 	adminKubeconfigSecret = "foo-lqmsh-admin-kubeconfig"
 	adminKubeconfig       = `clusters:
 - cluster:
@@ -1320,6 +1321,7 @@ func getInstallJob(c client.Client) *batchv1.Job {
 }
 
 func TestUpdatePullSecretInfo(t *testing.T) {
+	apis.AddToScheme(scheme.Scheme)
 	testPullSecret1 := `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`
 
 	tests := []struct {
@@ -1442,22 +1444,36 @@ func getCDFromClient(c client.Client) *hivev1.ClusterDeployment {
 	return nil
 }
 
+func createGlobalPullSecretObj(secretType corev1.SecretType, name, key, value string) *corev1.Secret {
+	secret := &corev1.Secret{
+		Type: secretType,
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: constants.HiveNamespace,
+		},
+		Data: map[string][]byte{
+			key: []byte(value),
+		},
+	}
+	return secret
+}
+
 func TestMergePullSecrets(t *testing.T) {
+	apis.AddToScheme(scheme.Scheme)
 
 	tests := []struct {
 		name             string
 		localPullSecret  string
 		globalPullSecret string
 		mergedPullSecret string
-		existingCD       []runtime.Object
+		existingObjs     []runtime.Object
 		expectedErr      bool
-		expectedMismatch bool
 	}{
 		{
 			name:             "merged pull secret should be be equal to local secret",
 			localPullSecret:  `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
 			mergedPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
-			existingCD: []runtime.Object{
+			existingObjs: []runtime.Object{
 				func() *hivev1.ClusterDeployment {
 					cd := getCDWithoutPullSecret()
 					cd.Spec.PullSecret = &corev1.LocalObjectReference{
@@ -1466,30 +1482,13 @@ func TestMergePullSecrets(t *testing.T) {
 					return cd
 				}(),
 			},
-		},
-		{
-			name:             "Test should fail as merged pull secret is not equal to local secret",
-			localPullSecret:  `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
-			mergedPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "aaaaaaaaaaaaaaa"}}}`,
-			existingCD: []runtime.Object{
-				func() *hivev1.ClusterDeployment {
-					cd := getCDWithoutPullSecret()
-					cd.Spec.PullSecret = &corev1.LocalObjectReference{
-						Name: pullSecretSecret,
-					}
-					return cd
-				}(),
-			},
-			expectedMismatch: true,
 		},
 		{
 			name:             "merged pull secret should be be equal to global pull secret",
 			globalPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
 			mergedPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
-			existingCD: []runtime.Object{
-				func() *hivev1.ClusterDeployment {
-					return getCDWithoutPullSecret()
-				}(),
+			existingObjs: []runtime.Object{
+				getCDWithoutPullSecret(),
 			},
 		},
 		{
@@ -1497,7 +1496,7 @@ func TestMergePullSecrets(t *testing.T) {
 			localPullSecret:  `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
 			globalPullSecret: `{"auths":{"cloud.okd.com":{"auth":"b34xVjWERckjfUyV1pMQTc=","email":"abc@xyz.com"}}}`,
 			mergedPullSecret: `{"auths":{"cloud.okd.com":{"auth":"b34xVjWERckjfUyV1pMQTc=","email":"abc@xyz.com"},"registry.svc.ci.okd.org":{"auth":"dXNljlfjldsfSDD"}}}`,
-			existingCD: []runtime.Object{
+			existingObjs: []runtime.Object{
 				func() *hivev1.ClusterDeployment {
 					cd := getCDWithoutPullSecret()
 					cd.Spec.PullSecret = &corev1.LocalObjectReference{
@@ -1509,10 +1508,8 @@ func TestMergePullSecrets(t *testing.T) {
 		},
 		{
 			name: "Test should fail as local an global pull secret is not available",
-			existingCD: []runtime.Object{
-				func() *hivev1.ClusterDeployment {
-					return getCDWithoutPullSecret()
-				}(),
+			existingObjs: []runtime.Object{
+				getCDWithoutPullSecret(),
 			},
 			expectedErr: true,
 		},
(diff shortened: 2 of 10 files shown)


pkg/operator/hive/hive.go
```diff
@@ -230,19 +232,16 @@ func (r *ReconcileHiveConfig) includeAdditionalCAs(hLog log.FieldLogger, h *reso
...
-		ValueFrom: &corev1.EnvVarSource{
-			SecretKeyRef: &corev1.SecretKeySelector{
-				LocalObjectReference: *instance.Spec.GlobalPullSecret,
-				Key:                  corev1.DockerConfigJsonKey,
-			},
-		},
+		return nil
+	}
+
+	GlobalPullSecretEnvVar := corev1.EnvVar{
```
staebler: Variable should start with a lowercase letter.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1486,7 +1486,14 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
 	}
 
 	// Check if global pull secret from env as it comes from hive config
-	globalPullSecret := os.Getenv("GLOBAL_PULL_SECRET")
+	globalPullSecretName := os.Getenv(constants.GlobalPullSecret)
+	var globalPullSecret string
+	if len(globalPullSecretName) != 0 {
+		globalPullSecret, err = controllerutils.LoadSecretData(r.Client, globalPullSecretName, cd.Namespace, corev1.DockerConfigJsonKey)
```
staebler: Namespace needs to be "hive".

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1486,7 +1486,14 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
...
+		if err != nil {
+			return "", err
```
staebler: Wrap a meaningful message around the error that indicates that the error was in getting the global pull secret.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1482,31 +1482,6 @@ func TestMergePullSecrets(t *testing.T) {
 			},
 			expectedMismatch: true,
 		},
-		{
```
staebler: Why are we dropping these tests?

pkg/operator/hive/hive.go
```diff
@@ -230,19 +232,16 @@ func (r *ReconcileHiveConfig) includeAdditionalCAs(hLog log.FieldLogger, h *reso
 	return nil
 }
 
-func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, h *resource.Helper, instance *hivev1.HiveConfig, hiveDeployment *appsv1.Deployment) {
+func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, h *resource.Helper, instance *hivev1.HiveConfig, hiveDeployment *appsv1.Deployment) error {
```
staebler: This function never returns are non-nil error. Is there another reason why this function should `error`?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1486,7 +1486,14 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
 	}
 
 	// Check if global pull secret from env as it comes from hive config
-	globalPullSecret := os.Getenv("GLOBAL_PULL_SECRET")
+	globalPullSecretName := os.Getenv(constants.GlobalPullSecret)
```
staebler: I would **much** rather the global pull secret name came into the controller via dependency injection rather than having the controller read an environment variable.

pkg/operator/hive/hive.go
LalatenduMohanty: No, we can remove the error as the return type.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: Sorry, did not get your point. Can you please elaborate?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: I could not find a way to mimic global pull secret in hive deployment for the tests. As now we load the global pull secret from the hive namespace. If you know of a way , let me know.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: An environment variable is logically similar to a command-line argument. It is something that should be read during the startup of the application and then passed around as a parameter where needed. This allows for validation of user inputs in one early place. It also makes it easier to unit test components by reducing the coupling to external resources.

So, rather than having the clusterDeployment controller reading the name of the global pull secret directly from the OS environment, the controller should be created with the name of the global pull secret.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
staebler: What is preventing the unit tests from adding a secret to the hive namespace?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1520,25 +1561,27 @@ func TestMergePullSecrets(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			fakeClient := fake.NewFakeClient(test.existingCD...)
+			if test.globalPullSecret != "" {
+				globalPullSecret := createGlobalPullSecretObj(corev1.SecretTypeDockerConfigJson, globalPullSecret, corev1.DockerConfigJsonKey, test.globalPullSecret)
+				test.existingObjs = append(test.existingObjs, globalPullSecret)
+			}
+			fakeClient := fake.NewFakeClient(test.existingObjs...)
```
staebler: This test needs to register the hive api with the scheme. There needs to be a call to the following somewhere before the fake client is created. In other tests, this is usually done at the top of the function.
```
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1486,30 +1526,31 @@ func TestMergePullSecrets(t *testing.T) {
 			name:             "merged pull secret should be be equal to global pull secret",
 			globalPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
 			mergedPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
-			existingCD: []runtime.Object{
-				func() *hivev1.ClusterDeployment {
-					return getCDWithoutPullSecret()
-				}(),
+			existingObjs: []runtime.Object{
+				getCDWithoutPullSecret(),
+				getHiveDeploymentObj(),
```
staebler: The clusterDeployment controller has no dependency on the hive deployment resource. It is not necessary to add it.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1520,25 +1561,27 @@ func TestMergePullSecrets(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			fakeClient := fake.NewFakeClient(test.existingCD...)
+			if test.globalPullSecret != "" {
+				globalPullSecret := createGlobalPullSecretObj(corev1.SecretTypeDockerConfigJson, globalPullSecret, corev1.DockerConfigJsonKey, test.globalPullSecret)
+				test.existingObjs = append(test.existingObjs, globalPullSecret)
```
staebler: To test the existence of the global pull secret, the GLOBAL_PULL_SECRET environment variable needs to be set.

LalatenduMohanty: Right, this what I am missing. Thanks

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: This test does not do a reconcile and does need to know all resources as only a single function is getting tested. Let me know if you still think we need to add `apis.AddToScheme(scheme.Scheme)`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: Currently the deployment for controllers is getting created at https://github.com/openshift/hive/blob/master/pkg/operator/hive/hive.go#L86 , also check https://github.com/openshift/hive/blob/master/pkg/operator/hive/hive.go#L43. We need to do changes to put the framework which would allow the model you are suggesting.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
staebler: The test adds resources to the fake client and does gets of resources. Both of those require the types added to the scheme. What do you get when you run the following without adding the api to the scheme?
```
go test ./pkg/controller/clusterdeployment/ -run TestMergePullSecrets
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1442,22 +1443,36 @@ func getCDFromClient(c client.Client) *hivev1.ClusterDeployment {
...
 	}{
 		{
 			name:             "merged pull secret should be be equal to local secret",
 			localPullSecret:  `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
 			mergedPullSecret: `{"auths": {"registry.svc.ci.okd.org": {"auth": "dXNljlfjldsfSDD"}}}`,
-			existingCD: []runtime.Object{
+			existingObjs: []runtime.Object{
 				func() *hivev1.ClusterDeployment {
 					cd := getCDWithoutPullSecret()
 					cd.Spec.PullSecret = &corev1.LocalObjectReference{
```
staebler: Why are the tests explicitly getting a clusterdeployment without a pull secret and then adding the pull secret rather than using `testClusterDeployment`, which already has the pull secret added?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1509,7 +1522,7 @@ func TestMergePullSecrets(t *testing.T) {
 		},
 		{
 			name: "Test should fail as local an global pull secret is not available",
-			existingCD: []runtime.Object{
+			existingObjs: []runtime.Object{
 				func() *hivev1.ClusterDeployment {
```
staebler: Remove this func. Call `getCDWithoutPullSecret` directly.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1471,7 +1486,7 @@ func TestMergePullSecrets(t *testing.T) {
 			name:             "Test should fail as merged pull secret is not equal to local secret",
```
staebler: I do not understand the point of this subtest. I read the `mergedPullSecret` as the expected value of the merged pull secret. Why would we have a subtest where we expected that the expected value differs from the actual value? How does this add to the previous subtest where we used the same inputs and validated that the actual merged pull secret equals the expected merged pull secret?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1520,25 +1533,32 @@ func TestMergePullSecrets(t *testing.T) {
...
 			}
 
-			localSecretObject := testSecret(corev1.SecretTypeDockercfg, pullSecretSecret, corev1.DockerConfigJsonKey, test.localPullSecret)
-			err := rcd.Create(context.TODO(), localSecretObject)
-			if err != nil {
-				t.Errorf("unexpected error: %v", err)
-				return
+			if test.localPullSecret != "" {
+				localSecretObject := testSecret(corev1.SecretTypeDockercfg, pullSecretSecret, corev1.DockerConfigJsonKey, test.localPullSecret)
+				err := rcd.Create(context.TODO(), localSecretObject)
```
staebler: It would be simpler to add the local pull secret to the existing objects passed to the fake client when it is built, like is being done for the global pull secret.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: We are doing some testing without local pull secret too, hence we need `getCDWithoutPullSecret`

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: Agree

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: Got it. Thanks for the explanation.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: I think we can remove it.

LalatenduMohanty:
@staebler Addressed your comments. I think it is ready to get merged.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
staebler: I understand the use for the `getCDWithoutPullSecret` function. However, here, you ultimately build a clusterdeployment **with** a pull secret. There is no need to create a clusterdeployment explicitly without a pull secret and then add the pull secret rather rather than using the function that already exists that creates a clusterdeployment with the pull secret.
```
existingObjs: []runtime.Object{
  testClusterDeployment(),
},
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1509,41 +1508,40 @@ func TestMergePullSecrets(t *testing.T) {
...
 			cd := getCDFromClient(rcd.Client)
 			if test.globalPullSecret != "" {
-				os.Setenv("GLOBAL_PULL_SECRET", test.globalPullSecret)
+				os.Setenv(constants.GlobalPullSecret, globalPullSecret)
 			}
-			defer os.Unsetenv("GLOBAL_PULL_SECRET")
+			defer os.Unsetenv(constants.GlobalPullSecret)
+
 			expetedPullSecret, err := rcd.mergePullSecrets(cd, rcd.logger)
 			if test.expectedErr {
```
staebler: If the test does not expect an error, then the test should assert that there was not an error.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1486,7 +1486,14 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
...
+	var globalPullSecret string
+	if len(globalPullSecretName) != 0 {
+		globalPullSecret, err = controllerutils.LoadSecretData(r.Client, globalPullSecretName, constants.HiveNamespace, corev1.DockerConfigJsonKey)
+		if err != nil {
+			return "", errors.Wrap(err, "global pull secret could not be retrived")
```
staebler: It would be nice to add a subtest to `TestMergePullSecrets` that validates the behavior when the global pull secret environmet variable is set but the referenced secret does not exist.

PR #491, opened 2019-08-06
SyncSetInstance SecretReference sync

abutcher:


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -623,8 +691,132 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+		}
+
+		var secretReferenceSyncConditions []hivev1.SyncCondition
+
+		// determine if secret is found, different or should be reapplied based on last probe time
+		found := false
+		different := false
+		shouldReApply := false
+		for _, rss := range ssi.Status.SecretReferences {
+			if rss.Name == secretReferenceSyncStatus.Name &&
```
staebler: Can we make a `

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -623,8 +691,132 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+			Hash:       hash,
+		}
+
+		var secretReferenceSyncConditions []hivev1.SyncCondition
+
+		// determine if secret is found, different or should be reapplied based on last probe time
+		found := false
+		different := false
+		shouldReApply := false
+		for _, rss := range ssi.Status.SecretReferences {
```
staebler: Can we make the following function instead of having this for loop and subsequent if statement here and in `reconcileDeletedSyncSetResources`?
```
func findSecretRefStatus(hivev1.SyncStatus, []hivev1.SyncStatus) *hivev1.SyncStatus
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -623,8 +691,132 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+			Namespace:  secretReference.Target.Namespace,
+			Resource:   secret.GetObjectMeta().GetResourceVersion(),
+			Hash:       hash,
+		}
+
+		var secretReferenceSyncConditions []hivev1.SyncCondition
+
+		// determine if secret is found, different or should be reapplied based on last probe time
+		found := false
+		different := false
```
staebler: Do we care about the distinction between `different` and `shouldReApply` enough to warrant separate variables?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -623,8 +691,132 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+
+	// ssi.Status.SecretReferences = syncStatusList
+
+	// We've saved apply and deletion errors separately, if either are present return an error for the controller to trigger retries
+	// and go into exponential backoff if the problem does not resolve itself.
+	if applyErr != nil {
+		return applyErr
+	}
+
+	if delErr != nil {
```
staebler: `delErr` can never be anything other than `nil` here.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -623,8 +691,132 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+		} else {
+			ssiLog.Debugf("resource %s/%s (%s) has not changed, will not apply", secret.Namespace, secret.Name, secret.Kind)
+			secretReferenceSyncStatus.Conditions = secretReferenceSyncConditions
+		}
+
+		syncStatusList = append(syncStatusList, secretReferenceSyncStatus)
+
+		// If an error applying occurred, stop processing right here
+		if applyErr != nil {
+			break
```
staebler: Why would we stop processing the rest of the secrets if there was an error with one of them? We should be aggregating errors together instead of bailing out at the first error.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -548,6 +557,65 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
+			ssiLog.WithField("resource", fmt.Sprintf("%s/%s", existingStatus.Namespace, existingStatus.Name)).
+				WithField("apiversion", existingStatus.APIVersion).
+				WithField("kind", existingStatus.Kind).Debug("resource not found in updated status, will queue up for deletion")
+			deletedStatusList = append(deletedStatusList, existingStatus)
+			deletedStatusIndices = append(deletedStatusIndices, i)
+		}
+	}
+
+	// If an error occurred applying resources, do not delete yet
+	if err != nil {
```
staebler: Along with continuing to apply other secrets after a failure, we should attempt to delete secrets even if there was an error during an apply.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -623,8 +691,132 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+			Kind:       secretReference.Target.Kind,
+			Name:       secretReference.Target.Name,
+			Namespace:  secretReference.Target.Namespace,
+			Resource:   secret.GetObjectMeta().GetResourceVersion(),
+			Hash:       hash,
+		}
+
+		var secretReferenceSyncConditions []hivev1.SyncCondition
+
+		// determine if secret is found, different or should be reapplied based on last probe time
```
staebler: I would find this function easier to follow if it looked like the following.
```
if rss := findSecretRefStatus(secretReferenceSyncStatus, ssi.Status.SecretReferences);
    rss == nil || needToReApplySecretRef(secretReferenceSyncStatus, rss) {
    // Apply secret ref
} else {
    // Do no apply secret ref
}
```
In other words, I would like to see a lot of the logic code moved to helper functions to (1) reduce control block indentation and (2) separate application code from logic code.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -548,6 +557,65 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
+			return nil, err
+		}
+		gvr := gv.WithResource(deletedStatus.Resource)
+		itemLog.Debug("deleting resource")
+		err = dynamicClient.Resource(gvr).Namespace(deletedStatus.Namespace).Delete(deletedStatus.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				itemLog.WithError(err).Error("error deleting resource")
+				index := deletedStatusIndices[i]
+				existingStatusList[index].Conditions = r.setDeletionFailedSyncCondition(existingStatusList[index].Conditions, err)
```
staebler: If there is an error that is not NotFound, then the status for that secret needs to be appended to `newStatusList` rather than updating the condition in the status in `existingStatusList`.

staebler:
dfb39eff..3537335f
* Save installer log and metadata directly in clusterprovision instead of in configmaps.
* Save kubeconfig and admin password at the same time that the metadata is saved. That way we avoid building out the cluster if there is a problem with them.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
abutcher: Asked about this in standup today. Consensus was that we were failing at first error by design to mimic other config management systems and that if users would like resources to be created separately then those resources should be in separate syncsets.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
abutcher: I've updated this to add the `deletedStatus` back into the `newStatusList` but I'd like to also add a test for this.

dgoodwin:
/test e2e

@staebler can you give this another look so we can merge, should allow us to close sprint 170.

staebler:
> /test e2e
> 
> @staebler can you give this another look so we can merge, should allow us to close sprint 170.

Sure. I was not aware that this was ready for another review.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -574,119 +638,130 @@ func (r *ReconcileSyncSetInstance) applySyncSetPatches(ssi *hivev1.SyncSetInstan
...
+		syncStatusList = append(syncStatusList, secretReferenceSyncStatus)
+
+		// If an error applying occurred, stop processing right here
+		if applyErr != nil {
+			break
 		}
 	}
 
-	return newStatusList, nil
+	ssi.Status.SecretReferences = r.reconcileDeleted("secret", ssi.Spec.ResourceApplyMode, dynamicClient, ssi.Status.SecretReferences, syncStatusList, applyErr, ssiLog)
```
staebler: The `err` parameter passed to `reconcileDeleted` should be non-nil if any of `readErr`, `hashErr`, or `applyErr` are non-nil. I don't know if there is any reason to keep the three classes of errors separate. Ultimately, each one is just an error applying the secret to the remote cluster, whether that is an error in getting the data to apply or in sending the data to apply.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -713,22 +788,25 @@ func (r *ReconcileSyncSetInstance) updateSyncSetInstanceStatus(ssi *hivev1.SyncS
...
 		hivev1.UnknownObjectSyncCondition,
-		status,
-		reason,
-		message,
-		controllerutils.UpdateConditionNever)
-	return syncSetConditions
+		corev1.ConditionTrue,
+		unknownObjectFoundReason,
+		fmt.Sprintf("Unable to gather Info for SyncSet resource at index %v in resources: %v", index, err),
+		controllerutils.UpdateConditionNever,
```
staebler: Why is the `UpdateConditionNever`? I would think that you would want it to be updated if the resource with the problem changed. For example, if there was a problem with index 1 and 3 and the user resolved the issue with index 1, you would want the condition to then show that there was a problem with index 3 instead of continuing to show that there was a problem with index 1.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -325,23 +321,208 @@ func TestSyncSetReconcile(t *testing.T) {
...
 	}
 
 	for _, test := range tests {
 		apis.AddToScheme(scheme.Scheme)
 		t.Run(test.name, func(t *testing.T) {
 			var ssi *hivev1.SyncSetInstance
 			cd := testClusterDeployment()
+
 			runtimeObjs := []runtime.Object{cd, kubeconfigSecret()}
+			for _, o := range test.existingObjs {
```
staebler:
```
runtimeObjs = append(runtimeObjs, test.existingObjs...)
```

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -524,6 +705,26 @@ func testSyncSetWithPatches(name string, patches ...hivev1.SyncObjectPatch) *hiv
...
+		},
+		Spec: hivev1.SyncSetSpec{
+			ClusterDeploymentRefs: []corev1.LocalObjectReference{
+				{
+					Name: testName,
+				},
+			},
+		},
+	}
+	for _, sr := range refs {
```
staebler:
```
ss.Spec.SecretReferences = append(ss.Spec.SecretReferences, refs...)
```

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -556,6 +757,27 @@ func testSelectorSyncSetWithResources(name string, resources ...runtime.Object)
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: hivev1.SelectorSyncSetSpec{
+			ClusterDeploymentSelector: metav1.LabelSelector{
+				MatchLabels: matchLabels,
+			},
+		},
+	}
+	for _, sr := range refs {
```
staebler:
```
ss.Spec.SecretReferences = append(ss.Spec.SecretReferences, refs...)
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
abutcher: I've changed this to `UpdateConditionIfReasonOrMessageChange` so that we get an updated condition message in that case.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
abutcher: Updated to only use `applyErr` for each case.

staebler:
@abutcher I gave this a lgtm, but it still needs the WIP removed.

PR #492, opened 2019-08-06
Temporary syncsetinstance hot loop fix.

abutcher:


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -368,6 +368,14 @@ func (r *ReconcileSyncSetInstance) applySyncSet(ssi *hivev1.SyncSetInstance, spe
 		ssiLog.WithError(err).Error("an error occurred applying syncset patches")
 		return err
 	}
+
+	// Temporary fix for status hot loop: do not update ssi.Status.{Patches,Resources} with empty slice.
```
staebler: We need this in a `defer` since the controller will try to update the `syncsetinstance` even when `applySyncSet` returns an error.

PR #493, opened 2019-08-06
Use latest hive image for processing image resolution job.

dgoodwin:
The log gathering PR merged earlier today missed the fact that image
resolution jobs use the new script, but an old hive image to process the
results of the script (if the ClusterImageSet predates the change). This
causes the CLI image to be resolved but never set on the
ClusterDeployment status. As a result the new code will not create an
install job until the CLI image is resolved and saved in status.

Temporary fix that will probably need to be cleaned up in this commit.
Uses the latest Hive image (i.e. what we're running the controllers
from, as set by the operator), falling back to the hive image in the
ClusterImageSet. This change is likely permanent considering the work
we're doing this week, only the fallback will likely be removed.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -422,7 +422,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	if cd.Status.InstallerImage == nil {
+	if !cd.Status.Installed && (cd.Status.InstallerImage == nil || cd.Status.CLIImage == nil) {
 		return r.resolveInstallerImage(cd, imageSet, releaseImage, hiveImage, cdLog)
 	}
 
@@ -766,7 +766,7 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 	// If job exists and is finished, delete so we can recreate it
 	case err == nil && controllerutils.IsFinished(existingJob):
 		jobLog.WithField("successful", controllerutils.IsSuccessful(existingJob)).
-			Warning("Finished job found, but installer image is not yet resolved. Deleting.")
+			Warning("Finished job found, but all images not yet resolved. Deleting.")
 		err := r.Delete(context.Background(), existingJob,
 			client.PropagationPolicy(metav1.DeletePropagationForeground))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -422,7 +422,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	if cd.Status.InstallerImage == nil {
+	if cd.Status.Installed && (cd.Status.InstallerImage == nil || cd.Status.CLIImage == nil) {
```
staebler: I don't follow here. Why do we only want to resolve the installer image when the cluster has already been installer?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -422,7 +422,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	if cd.Status.InstallerImage == nil {
+	if !cd.Status.Installed && (cd.Status.InstallerImage == nil || cd.Status.CLIImage == nil) {
```
dgoodwin: I'm nervous about this line, I don't want to go regenerate a bunch of imageset jobs for clusters that are old, already installed, but don't have a CLIImage set (which they wouldn't because they're old)

Adding the Installed guard is meant to address this.

Otherwise this *should* recreate the imageset job due to the code in this function that deletes the job if it's finished. (but we don't have our images set due to this clause)

pkg/imageset/generate.go
```diff
@@ -98,6 +100,13 @@ func GenerateImageSetJob(cd *hivev1.ClusterDeployment, releaseImage, serviceAcco
 		},
 	}
 
+	hiveUtilImage := os.Getenv(images.HiveImageEnvVar)
+	if hiveUtilImage == "" {
```
dgoodwin: This if can probably be deleted soon.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Yeah it was quite wrong when I first pushed, should be !Installed.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: OK. This looks an adequate approach to me. The controller does not need the installer image or the CLI image after the cluster has been installed, so there is no need to resolve the images at that point.

PR #494, opened 2019-08-07
Making pull secret file and environment variable optional for hiveutil

LalatenduMohanty:

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -118,12 +118,19 @@ const (
 // NewCreateClusterCommand creates a command that generates and applies cluster deployment artifacts.
 func NewCreateClusterCommand() *cobra.Command {
 	var homeDir = "."
+
 	if u, err := user.Current(); err == nil {
 		homeDir = u.HomeDir
 	}
 	defaultSSHPublicKeyFile := filepath.Join(homeDir, ".ssh", "id_rsa.pub")
-	defaultPullSecretFile := filepath.Join(homeDir, ".pull-secret")
 	defaultAWSCredsFile := filepath.Join(homeDir, ".aws", "credentials")
+
+	defaultPullSecretFile := filepath.Join(homeDir, ".pull-secret")
+	if _, err := os.Stat(defaultPullSecretFile); os.IsNotExist(err) {
+		defaultPullSecretFile = ""
+	} else if err != nil {
+		log.WithError(err).Errorf("%v can not be used", defaultPullSecretFile)
+	}
 
 	opt := &Options{}
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


contrib/pkg/createcluster/create.go
```diff
@@ -263,21 +281,21 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
 	if imageSet != nil {
 		result = append(result, imageSet)
 	}
+
+	// Adding pull secret name to cluster deployment if pull secret exists
+	if pullSecret != nil {
+		cd.Spec.PullSecret = &corev1.LocalObjectReference{
+			Name: fmt.Sprintf("%s-pull-secret", cd.Name),
+		}
+	}
```
LalatenduMohanty: I am not sure if this is an acceptable, however I am open to suggestion.

joelddiaz: seems fine to me. if you want to make it a little less magical, i would store the "pull secret name" template as some const pullSecretNameTemplate = "%s-pull-secret", and just use it in your Sprintf() (not sure how much practical difference it makes...)

contrib/pkg/createcluster/create.go
```diff
@@ -118,13 +118,18 @@ const (
...
+	var defaultPullSecretFile string
+
 	if u, err := user.Current(); err == nil {
 		homeDir = u.HomeDir
 	}
 	defaultSSHPublicKeyFile := filepath.Join(homeDir, ".ssh", "id_rsa.pub")
-	defaultPullSecretFile := filepath.Join(homeDir, ".pull-secret")
 	defaultAWSCredsFile := filepath.Join(homeDir, ".aws", "credentials")
 
+	if _, err := os.Stat(filepath.Join(homeDir, ".pull-secret")); err == nil {
```
staebler: My preference would be to do it like the following so that we are not duplicating the code of building out the path.
```
defaultPullSecretFile := filepath.Join(homeDir, ".pull-secret")
if _, err := os.Stat(defaultPullSecretFile); err != nil {
  defaultPullSecretFile = ""
}
```

contrib/pkg/createcluster/create.go
```diff
@@ -263,21 +281,21 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
...
-	pullSecret, err := o.generatePullSecret()
-	if err != nil {
-		return nil, err
-	}
-	sshSecret, err := o.generateSSHSecret()
-	if err != nil {
-		return nil, err
+
+	// With introduction of global pull secret, pull secret is optional now
+	if pullSecret == nil {
```
staebler: Is the ordering of these objects important in some way? If not, then it would be better to do the following.
```
result = append(result, awsCreds, sshSecret)
if pullSecret != nil {
  result = append(result, pullSecret)
}
return result, err
```

contrib/pkg/createcluster/create.go
staebler: if `!os.IsNotExist(err)`, then we should warn the user that the pull secret file will not be used even though it exists.

contrib/pkg/createcluster/create.go
LalatenduMohanty: Yeah, it will be more and less the same thing.

PR #499, opened 2019-08-09
Add script to extract cluster logs from PVC after failures.

dgoodwin:
If the install pod is no longer running, we need a way to get access to
the logs PVC. This script creates a dummy deployment we can use to copy
from, finds the pod name, waits for it to come online, copies the logs
out, and then deletes the deployment.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,5 +1,15 @@
 # Troubleshooting
 
+## Cluster Install Failure Logs
+
+In the event a cluster is brought up but overall installation fails, either during bootstrap or cluster initialization, Hive will attempt to gather logs from the cluster itself. These logs are stored in a persistent volume created for every install job. If the install succeeds on the first attempt, the persistent volume claim is immediately deleted. If the install has had any errors along the way, it will be preserved for 7 days and then removed.
+
+You can access these logs gathered from the cluster with the following script found in the hive.git repository:
+
+```bash
+$ hack/logextractor.sh mycluster ./extracted-logs/
+```
+
 ## Deprovision
 
 After deleting your cluster deployment you will see an uninstall job created. If for any reason this job gets stuck you can:
--- a/hack/logextractor.sh
+++ b/hack/logextractor.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+#
+# This script can be used to extract the failure cluster logs tarball from the PVC created for an install pod.
+#
+# Usage: ./logextractor.sh [CLUSTER_DEPLOYMENT_NAME] [DEST_DIR]
+#
+
+POD_NAME="hive-logextractor"
+PVC_NAME=$(oc get pvc -l hive.openshift.io/cluster-deployment-name=$1 -o jsonpath="{.items[0].metadata.name}" 2> /dev/null)
+
+if [ -z "$PVC_NAME" ]; then echo "Could not find PVC for ClusterDeployment $1"; exit 1; fi
+
+echo "Found PVC for cluster deployment: $PVC_NAME"
+
+echo "apiVersion: v1
+kind: Pod
+metadata:
+  name: $POD_NAME
+spec:
+  volumes:
+  - name: logsvol
+    persistentVolumeClaim:
+      claimName: $PVC_NAME
+  containers:
+  - image: centos:7
+    command: ["/bin/sh"]
+    args: ["-c", "while true\; do sleep 100000000\; done"]
+    imagePullPolicy: Always
+    name: logextractor
+    resources: {}
+    volumeMounts:
+    - mountPath: "/logs"
+      name: logsvol
+    terminationMessagePath: /dev/termination-log
+    terminationMessagePolicy: File" | oc apply -f -
+
+
+echo "waiting for $POD_NAME pod to be ready"
+
+oc wait --for=condition=Ready --timeout=120s pod/$POD_NAME
+
+echo "copying logs to $2"
+oc cp $POD_NAME:/logs $2
+
+echo "deleting logextractor pod"
+oc delete pod $POD_NAME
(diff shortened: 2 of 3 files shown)


hack/logextractor.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
+#
+# This script can be used to extract the failure cluster logs tarball from the PVC created for an install pod.
+#
+# Usage: ./logextractor.sh [LOGS_PVC_NAME] [DEST_DIR]
+#
+
+echo "apiVersion: extensions/v1beta1
+kind: Deployment
```
staebler: Is there a reason why we cannot use a Pod directly instead of a Deployment?

hack/logextractor.sh
```diff
@@ -0,0 +1,49 @@
...
+  labels:
+    run: logextractor
+  name: logextractor
+spec:
+  selector:
+    matchLabels:
+      run: logextractor
+  template:
+    metadata:
+      creationTimestamp: null
```
staebler: Remove this.

hack/logextractor.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
+#
+# This script can be used to extract the failure cluster logs tarball from the PVC created for an install pod.
+#
+# Usage: ./logextractor.sh [LOGS_PVC_NAME] [DEST_DIR]
```
staebler: It would be nice if the argument was the name of the clusterdeployment instead of the PVC name. We should be able to find the PVC using labels, assuming the PVC is labeled with the clusterdeployment name.

hack/logextractor.sh
```diff
@@ -0,0 +1,49 @@
...
+      labels:
+        run: logextractor
+    spec:
+      volumes:
+      - name: logsvol
+        persistentVolumeClaim:
+          claimName: $1
+      containers:
+      - image: centos/httpd-24-centos7
+        imagePullPolicy: Always
```
staebler: Do we need all the extras for running an http server? Can we use something slimmer like centos:7?

hack/logextractor.sh
dgoodwin: Well originally I did this because the instructions I was following added the volume via CLI after creation, which you can't do with a pod. But then this later became inline yaml so at this point no, I don't think there's a reason it can't be a pod, will switch.

hack/logextractor.sh
```diff
@@ -0,0 +1,49 @@
...
+POD_NAME=$(kubectl get pod -l run=logextractor -o jsonpath="{.items[0].metadata.name}")
+echo "waiting for pod to be ready: $POD_NAME"
+
+kubectl wait --for=condition=Ready --timeout=120s pod/$POD_NAME
+
+echo "copying logs to ./logs"
+kubectl cp $POD_NAME:/logs $2
+
+echo "deleting logextractor deployment"
+kubectl delete deployment logextractor
```
LalatenduMohanty: IMO we should use `oc` binary to keep things consistent. We suggest `oc` instead of `kubectl` throughout our documentation.

dgoodwin:
Feedback addressed and added a note in the troubleshooting doc.

docs/troubleshooting.md
```diff
@@ -1,5 +1,15 @@
 # Troubleshooting
 
+## Cluster Install Failure Logs
+
+In the event a cluster is brought up but overall installation fails, either during bootstrap or cluster initialization, Hive will attempt to gather logs from the cluster itself. These logs are stored in a persistent volume created for every install job. If the install is successful, the persistent volume claim is immediately deleted. If the install has had any errors along the way, it will be preserved for 7 days and then removed.
```
staebler: s/If the install is successful/If the install succeeds on the first attempt/

docs/troubleshooting.md
```diff
@@ -1,5 +1,15 @@
...
+
+You can access these logs gathered from the cluster with the following script found in the hive.git repository:
+
+```bash
+$ sh hack/logextractor.sh mycluster ./extracted-logs/
```
staebler: Is there any downside to simply making `logextractor.sh` executable so that the preceding `sh` is not necessary?

hack/logextractor.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+# This script can be used to extract the failure cluster logs tarball from the PVC created for an install pod.
+#
+# Usage: ./logextractor.sh [CLUSTER_DEPLOYMENT_NAME] [DEST_DIR]
+#
+
+set -e
+
+POD_NAME="hive-logextractor"
+PVC_NAME=$(oc get pvc -l hive.openshift.io/cluster-deployment-name=$1 -o jsonpath="{.items[0].metadata.name}")
```
staebler: For cleaner output in the case where the PVC does not exist, could I suggest the following. I think you need to get rid of the `set -e` for this to work, though you may want to do that anyway.

```
PVC_NAME=$(oc get pvc -l hive.openshift.io/cluster-deployment-name=$1 -o jsonpath="{.items[0].metadata.name}" 2> /dev/null)
if [ -z "$PVC_NAME" ]; then { echo "Could not find PVC for ClusterDeployment $1"; exit 1 }; fi
```
Otherwise, the script will put out esoteric output which may make it difficult to understand what went wrong.
```
error: error executing jsonpath "{.items[0].metadata.name}": Error executing template: array index out of bounds: index 0, length 0. Printing more information for debugging the template:
	template was:
		{.items[0].metadata.name}
	object given to jsonpath engine was:
		map[string]interface {}{"kind":"List", "apiVersion":"v1", "metadata":map[string]interface {}{"selfLink":"", "resourceVersion":""}, "items":[]interface {}{}}
```

hack/logextractor.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "waiting for $POD_NAME pod to be ready"
+
+oc wait --for=condition=Ready --timeout=120s pod/$POD_NAME
+
+echo "copying logs to $2"
+oc cp $POD_NAME:/logs $2
+
+echo "deleting logextractor pod"
+oc delete pod $POD_NAME
```
staebler: The script should delete the pod even if there was a failure waiting for the pod to be ready or copying the logs.

dgoodwin:
Updated again, removed the set -e to let the cleanup happen regardless of errors.

PR #500, opened 2019-08-12
Run generate  bydefult before verify

LalatenduMohanty:


csrwng:
@LalatenduMohanty did you intend to include the 2 commits in this PR?

LalatenduMohanty:
> @LalatenduMohanty did you intend to include the 2 commits in this PR?

Nope, it was a mistake. Fixed it.

csrwng:
So we already invoke generate as part of the all target because build (manager, operator, etc) depends on it. Is the purpose of this to run generate before the verify test?

LalatenduMohanty:
> Is the purpose of this to run generate before the verify test?

Right. Else it fails in verify if changes supposed to be done in `generate`. So why not make it generate before verify?

csrwng:
> Right. Else it fails in verify if changes supposed to be done in `generate`. So why not make it generate before verify?

That's fine, just wanted to understand why you're adding it.

/lgtm

PR #501, opened 2019-08-13
Certificate utility

csrwng:
Adds support for serving certs to the create-cluster command
Adds a new command to hiveutil to create a serving certificate through letsencrypt

### DIFF
--- a/contrib/pkg/certificate/create.go
+++ b/contrib/pkg/certificate/create.go
@@ -1,0 +1,179 @@
+package certificate
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"os/user"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/route53"
+
+	awsclient "github.com/openshift/hive/pkg/awsclient"
+)
+
+// Options is the set of options to create a new server certificate
+type Options struct {
+	Name       string
+	BaseDomain string
+	Email      string
+	Region     string
+	OutputDir  string
+	WaitTime   time.Duration
+}
+
+// NewCreateCertifcateCommand returns a command that will create a letsencrypt serving cert
+// for OpenShift clusters on AWS.
+func NewCreateCertifcateCommand() *cobra.Command {
+	username := "user"
+	homeDir := "/tmp"
+	user, err := user.Current()
+	if err == nil {
+		username = user.Username
+		homeDir = user.HomeDir
+	}
+
+	opt := &Options{}
+	cmd := &cobra.Command{
+		Use:   "create CLUSTER_NAME",
+		Short: "Creates a new serving certificate for the given cluster",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := opt.Complete(cmd, args); err != nil {
+				return
+			}
+			log.SetLevel(log.InfoLevel)
+			err := opt.Run()
+			if err != nil {
+				log.WithError(err).Error("Error")
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVar(&opt.BaseDomain, "base-domain", "new-installer.openshift.com", "Base domain for the cluster")
+	flags.StringVar(&opt.Email, "email", fmt.Sprintf("%s@redhat.com", username), "email address to use for certificate")
+	flags.StringVar(&opt.Region, "region", "us-east-1", "AWS region for certificate hosted zone")
+	flags.StringVar(&opt.OutputDir, "output-dir", homeDir, "Output directory for certs. Defaults to home directory")
+	flags.DurationVar(&opt.WaitTime, "wait-time", 30*time.Second, "amount of time to wait after creating dns entries so that they can be authenticated")
+	return cmd
+}
+
+// Complete finalizes command options
+func (o *Options) Complete(cmd *cobra.Command, args []string) error {
+	o.Name = args[0]
+	return nil
+}
+
+// Run executes the command
+func (o *Options) Run() error {
+	certbotPath, err := exec.LookPath("certbot")
+	if err != nil {
+		return errors.New("certbot is required to create a certificate, install it by following instructions here: https://certbot.eff.org")
+	}
+	baseDomainID, err := o.getBaseDomainID()
+	if err != nil {
+		return err
+	}
+	certbotDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return errors.New("cannot create temporary directory for certbot")
+	}
+	log.Infof("Executing certbot command in directory %s", certbotDir)
+	certbotCommand := exec.Command(certbotPath,
+		"certonly",
+		"--non-interactive",
+		"--manual",
+		"--manual-auth-hook", fmt.Sprintf("%s certificate auth-hook --wait-time %v %s %s", os.Args[0], o.WaitTime, o.Region, baseDomainID),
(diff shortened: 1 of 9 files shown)


contrib/pkg/certificate/create.go
```diff
@@ -0,0 +1,174 @@
...
+	flags.StringVar(&opt.Email, "email", fmt.Sprintf("%s@redhat.com", username), "email address to use for certificate")
+	flags.StringVar(&opt.Region, "region", "us-east-1", "AWS region for certificate hosted zone")
+	flags.StringVar(&opt.OutputDir, "output-dir", homeDir, "Output directory for certs. Defaults to home directory")
+	flags.DurationVar(&opt.WaitTime, "wait-time", 30*time.Second, "amount of time to wait after creating dns entries so that they can be authenticated")
+	return cmd
+}
+
+// Complete finalizes command options
+func (o *Options) Complete(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
```
staebler: Set the `Args` field in the `cobra.Command` object to `ExactArgs(1)`. That will take care of this check for you.

contrib/pkg/certificate/create.go
```diff
@@ -0,0 +1,174 @@
...
+	if hostedZoneID == "" {
+		return "", fmt.Errorf("could not find public hosted zone with domain %s", o.BaseDomain)
+	}
+	return hostedZoneID, nil
+}
+
+func copyFile(src, dst string) error {
+	b, err := ioutil.ReadFile(src)
+	if err != nil {
+		return fmt.Errorf("cannot read %s: %v", src, err)
```
staebler: Since we are already vendoring `github.com/pkg/errors` courtesy of the installer, I would prefer to see us use that package here as follows.
```
return errors.Wrapf(err, "cannot read %s", src)
```

contrib/pkg/certificate/hook.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	o.Region = args[0]
+	o.HostedZoneID = args[1]
+
+	o.Domain = os.Getenv("CERTBOT_DOMAIN")
+	o.Value = os.Getenv("CERTBOT_VALIDATION")
+
+	if o.Domain == "" || o.Value == "" {
+		log.Infof("Required environment variables CERTBOT_DOMAIN and CERTBOT_VALIDATION must be present")
+		return fmt.Errorf("invalid")
```
staebler: There is formatting being done, so we can use the following instead.
```
return errors.New("invalid")
```

contrib/pkg/createcluster/create.go
```diff
@@ -196,6 +202,11 @@ func (o *Options) Validate(cmd *cobra.Command) error {
 	if !o.UseClusterImageSet && len(o.ClusterImageSet) > 0 {
 		cmd.Usage()
 		log.Info("If not using cluster image sets, do not specify the name of one")
+		return fmt.Errorf("invalid option")
+	}
+	if len(o.ServingCert) > 0 && len(o.ServingCertKey) == 0 {
```
staebler: Should this be the following instead to also validate that the key was not provided with the cert?
```
if (len(o.ServingCert) == 0) != (len(o.ServingCertKey) == 0) {
```

docs/developing.md
```diff
@@ -135,6 +135,31 @@ Alternatively, you can run the Dep command directly.
...
+
+Prerequisites:
+* The `certbot` command must be available and in the path of your machine. You can obtain by following instructions at:
+  [https://certbot.eff.org/docs/install.html](https://certbot.eff.org/docs/install.html)
+* You must have credentials for AWS available in your command line, either by a configured ~/.aws/credentials or environment variables.
+
+### Generating a Certificate
+
+1. Ensure that the hiveutil binary is available (`make hiveutil`)
+2. Run: `hiveutil certificate create ${CLUSTER_NAME} --domain ${BASE_DOMAIN}`
```
staebler: Should we also provide a util for the user to delete the certificate that they created?

contrib/pkg/createcluster/create.go
csrwng: What triggers adding a cert section is the cert param, if you only specify a key, it's just ignored. (Maybe shouldn't be?)

docs/developing.md
csrwng: There's really no corresponding delete. All the create does is that it places a new cert key in the output directory. If you don't want it anymore, just delete the file :)

csrwng:
Thx @staebler, comments addressed

docs/developing.md
staebler: That makes sense. I wasn't sure if it created an artifact on AWS as well or not.

csrwng: It does create a DNS record on AWS but only temporarily to validate that you own the domain you’re getting a cert for.

docs/developing.md
```diff
@@ -135,6 +135,31 @@ Alternatively, you can run the Dep command directly.
...
 ```
 
+## Using Serving Certificates
+
+The hiveutil command includes a utility to generate Letsencrypt certificates for use with clusters you create in Hive.
+
+Prerequisites:
+* The `certbot` command must be available and in the path of your machine. You can obtain by following instructions at:
+  [https://certbot.eff.org/docs/install.html](https://certbot.eff.org/docs/install.html)
+* You must have credentials for AWS available in your command line, either by a configured ~/.aws/credentials or environment variables.
```
LalatenduMohanty: @csrwng  The environment variables are `$AWS_ACCESS_KEY_ID      $AWS_SECRET_ACCESS_KEY` right? IMO it should include this information.

docs/developing.md
```diff
@@ -135,6 +135,31 @@ Alternatively, you can run the Dep command directly.
...
+
+### Generating a Certificate
+
+1. Ensure that the hiveutil binary is available (`make hiveutil`)
```
LalatenduMohanty: s/hiveutil/`hiveutil`/ , usually it is better to keep the binary names highlighted.

docs/developing.md
LalatenduMohanty: Looks like the example command needs be be fixed to `$ #bin/hiveutil certificate create ${CLUSTER_NAME} --base-domain ${BASE_DOMAIN}`
As the usage has `  --base-domain` instead of `--domain`
```
$ bin/hiveutil certificate create
Error: accepts 1 arg(s), received 0
Usage:
  hiveutil certificate create CLUSTER_NAME [flags]

Flags:
      --base-domain string   Base domain for the cluster (default "new-installer.openshift.com")
      --email string         email address to use for certificate (default "lmohanty@redhat.com")
  -h, --help                 help for create
      --output-dir string    Output directory for certs. Defaults to home directory (default "/home/lmohanty")
      --region string        AWS region for certificate hosted zone (default "us-east-1")
      --wait-time duration   amount of time to wait after creating dns entries so that they can be authenticated (default 30s)
```

csrwng:
Thx @LalatenduMohanty, comments addressed

csrwng:
Added utility for the additional CA configuration

docs/developing.md
```diff
@@ -90,6 +90,32 @@ spec:
...
+Prerequisites:
+* The `certbot` command must be available and in the path of your machine. You can install it by following the instructions at:
+  [https://certbot.eff.org/docs/install.html](https://certbot.eff.org/docs/install.html)
+* You must have credentials for AWS available in your command line, either by a configured `~/.aws/credentials` or environment variables (`AWS_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY`).
+
+### Generating a Certificate
+
+1. Ensure that the `hiveutil` binary is available (`make hiveutil`)
+2. Run: `hiveutil certificate create ${CLUSTER_NAME} --base-domain ${BASE_DOMAIN}` 
+   where CLUSTER_NAME is the name of your cluster and BASE_DOMAIN is the public DNS domain for your cluster (Defaults to `new-installer.openshift.com`)
```
dgoodwin: This is a great idea.

PR #503, opened 2019-08-14
Force reloading of APIServer on certs change

csrwng:
Whenever the serving cert secret is updated (as in renewal), we need
to let the Kube API server on the target cluster know so that the
new certs are applied. This fix uses a hash of all cert secrets
passed to the API server to set the value of the
"forceDeploymentReason" field on the KubeAPIServer resource on the
cluster.

### DIFF
--- a/pkg/controller/controlplanecerts/controlplanecerts_controller.go
+++ b/pkg/controller/controlplanecerts/controlplanecerts_controller.go
@@ -2,8 +2,11 @@ package controlplanecerts
 
 import (
 	"context"
+	"crypto/md5"
 	"fmt"
+	"io"
 	"reflect"
+	"sort"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -39,6 +42,8 @@ const (
 	certsNotFoundMessage = "One or more serving certificates for the cluster control plane are missing"
 	certsFoundReason     = "ControlPlaneCertificatesFound"
 	certsFoundMessage    = "Control plane certificates are present"
+
+	kubeAPIServerPatchTemplate = `[ {"op": "replace", "path": "/spec/forceRedeploymentReason", "value": %q } ]`
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -297,7 +297,19 @@ func (r *ReconcileControlPlaneCerts) generateControlPlaneCertsSyncSet(cd *hivev1
 		cdLog.WithError(err).Error("cannot add typemeta to syncset resources")
 		return nil, err
 	}
+	// kubeAPIServerPatch indicates sets the forceRedeploymentField on the kube API server cluster operator
```
dgoodwin: nit: "indicates sets"

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -297,7 +297,19 @@ func (r *ReconcileControlPlaneCerts) generateControlPlaneCertsSyncSet(cd *hivev1
...
+	// to a hash of all the cert secrets. If the content of the certs secrets changes, then the new
+	// hash value will force the kube API server to redeploy and apply the new certs.
```
dgoodwin: The in-cluster operator watches this annotation, and if it changes will automatically redeploy, but do nothing if the value stays steady?

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -375,3 +387,28 @@ func controlPlaneCertsSyncSetName(name string) string {
...
+		keys = append(keys, k)
+	}
+	sort.StringSlice(keys).Sort()
+	fmt.Fprintf(w, "%s/%s\n", secret.Namespace, secret.Name)
+	for _, k := range keys {
+		fmt.Fprintf(w, "%s: %x\n", k, secret.Data[k])
+	}
+}
+
+func secretsHash(secrets []*corev1.Secret) string {
```
dgoodwin: This func could probably use a unit test just to make sure we're getting consistent hashes, but then I'm not entirely sure if we can control the order of the secret keys in a test?

pkg/controller/controlplanecerts/controlplanecerts_controller.go
csrwng: Yes

csrwng:
Thx @dgoodwin, comments addressed

PR #504, opened 2019-08-15
Report target clusteroperator status

csrwng:


pkg/apis/hive/v1alpha1/clusterstate_types.go
```diff
@@ -0,0 +1,51 @@
...
+	Conditions []configv1.ClusterOperatorStatusCondition `json:"conditions,omitempty"`
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterState is the Schema for the clusterstates API
+// +k8s:openapi-gen=true
+// +kubebuilder:subresource:status
+type ClusterState struct {
```
dgoodwin: With this name it feels like it might be a nice home for ClusterVersionStatus that we currently store in ClusterDeployment.Status. In theory that may not be used and we could also move it here.

I like that these are done by two controllers, but then that's two controllers writing to the same type fairly often, I'm not sure if that's a good idea or not.

If we think it's ok to store cluster version status here someday (assuming it's a safe breaking API change), and possibly other state we might fetch from cluster in future, then I really like this naming.

If we don't like the sound of the above then this should probably be named something more specific like ClusterOperatorState.

csrwng: lol, the only reason I didn't name it ClusterOperatorState is that I didn't know what to name the thing that represents the status of a single ClusterOperator :)

csrwng: I think it makes sense that cluster version status be stored in this CRD and not on the clusterdeployment

csrwng:
should be ready for review

LalatenduMohanty:
Address https://jira.coreos.com/browse/CO-529

pkg/controller/clusterstate/clusterstate_controller_test.go
```diff
@@ -0,0 +1,246 @@
...
+				testClusterStateWithStatus(co("a"), co("b"), co("c")),
+				testClusterDeployment(),
+				testKubeconfigSecret(),
+			},
+			remote: []runtime.Object{co("a"), co("b"), uco("c")},
+			validate: func(t *testing.T, c client.Client, result reconcile.Result) {
+				st := cs(t, c)
+				validateStatus(t, st.Status, co("a"), co("b"), uco("c"))
+			},
+		},
```
staebler: Add unit tests for
1) removed remote clusteroperator,
2) added remote condition, and
3) removed remote condition.

pkg/controller/clusterstate/clusterstate_controller_test.go
```diff
@@ -0,0 +1,246 @@
...
+			t.Errorf("Operators in status don't match. Expected: %s, Got: %s", operators[i].Name, status.ClusterOperators[i].Name)
+			return
+		}
+		sort.Slice(status.ClusterOperators[i].Conditions, func(m, n int) bool {
+			return status.ClusterOperators[i].Conditions[m].Type < status.ClusterOperators[i].Conditions[n].Type
+		})
+		sort.Slice(operators[i].Status.Conditions, func(m, n int) bool {
+			return operators[i].Status.Conditions[m].Type < operators[i].Status.Conditions[n].Type
+		})
+		if !reflect.DeepEqual(status.ClusterOperators[i].Conditions, operators[i].Status.Conditions) {
```
staebler: `assert.ElementsMatch` will do this for you without needing to sort.

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	}()
+
+	// Fetch the ClusterState instance
+	st := &hivev1.ClusterState{}
+	err := r.Get(context.TODO(), request.NamespacedName, st)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Object not found, return.  Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers.
+			logger.Info("cluster state not found")
```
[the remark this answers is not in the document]
csrwng: Hmm, I like the idea of creating the ClusterState here, will change.

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	if removed.Len() > 0 {
+		changed = true
+		logger.Infof("Removed cluster operators: %v", removed.List())
+	}
+	if added.Len() > 0 {
+		changed = true
+		logger.Infof("Added cluster operators: %v", added.List())
+	}
+
+	for _, name := range same.List() {
```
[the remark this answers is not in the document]
csrwng: Yes that was the purpose. My thinking in having all this code here is that it could be valuable for debugging purposes to understand when a particular operator change was first observed on a cluster.

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,303 @@
...
+			changed = true
+		}
+	}
+	return changed
+}
+
+func operatorConditionChanged(logger log.FieldLogger, name string, existing, updated configv1.ClusterOperatorStatusCondition) bool {
+	changed := false
+	changes := []string{}
+	if existing.Status != updated.Status {
```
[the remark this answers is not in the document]
csrwng: I would like to log when a condition changes status, so maybe just log on that and not necessarily on every field?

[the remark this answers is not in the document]
staebler: That is a good compromise to limit the amount of code there just for logging.

pkg/controller/clusterstate/clusterstate_controller.go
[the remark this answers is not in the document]
staebler: OK. I can live with that.

config/crds/hive_v1alpha1_clusterstate.yaml
```diff
@@ -0,0 +1,76 @@
...
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
```
LalatenduMohanty: I understand that these are mostly copied from standard templates, but these links `https://git.k8s.io/community/contributors/devel/api-conventions.md#resources` does not exist right now. Not sure what is the correct thing to do.

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+)
+
+const (
+	controllerName = "clusterState"
+)
+
+var (
+	statusUpdateInterval = 10 * time.Minute
```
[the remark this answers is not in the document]
LalatenduMohanty: Can we make the status update interval configurable somehow?

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,303 @@
...
+		})
+	}
+	if operatorStatesChanged(logger, st.Status.ClusterOperators, operatorStates) {
+		st.Status.ClusterOperators = operatorStates
+		err = r.updateStatus(r, st)
+		if err != nil {
+			logger.WithError(err).Errorf("failed to update cluster operator state")
+			return reconcile.Result{}, err
+		}
+		logger.Info("clusterState has been updated")
```
LalatenduMohanty: IMO the code which update or adds cluster operator state should be in a separate function. 
```
operatorStates := make([]hivev1.ClusterOperatorState, 0, len(clusterOperators.Items))
	for _, clusterOperator := range clusterOperators.Items {
		operatorStates = append(operatorStates, hivev1.ClusterOperatorState{
			Name:       clusterOperator.Name,
			Conditions: clusterOperator.Status.Conditions,
		})
	}
	if operatorStatesChanged(logger, st.Status.ClusterOperators, operatorStates) {
		st.Status.ClusterOperators = operatorStates
		err = r.updateStatus(r, st)
		if err != nil {
			logger.WithError(err).Errorf("failed to update cluster operator state")
		}
		logger.Info("clusterState has been updated")
```

pkg/controller/clusterstate/clusterstate_controller.go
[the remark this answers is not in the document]
csrwng: We can do a followup and make it configurable if the default doesn't work for us. Otherwise, it's one more knob for which we have to come with guidance on how to tune.

config/crds/hive_v1alpha1_clusterstate.yaml
csrwng: these aren't copied, these are generated by the controller-tools crd generation tool. We should likely update to a more recent version.

csrwng:
@LalatenduMohanty @staebler thx for the reviews, comments addressed

config/crds/hive_v1alpha1_clusterstate.yaml
LalatenduMohanty: > e should likely update to a more recent version.



> We should likely update to a more recent version.

Can we have a Jira for this? You can assign this to me.

pkg/controller/clusterstate/clusterstate_controller.go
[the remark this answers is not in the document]
LalatenduMohanty: Agree.

config/crds/hive_v1alpha1_clusterstate.yaml
csrwng: https://jira.coreos.com/browse/CO-535

config/crds/hive_v1alpha1_clusterstate.yaml
```diff
@@ -0,0 +1,85 @@
...
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: clusterstates.hive.openshift.io
+spec:
+  group: hive.openshift.io
+  names:
+    kind: ClusterState
+    plural: clusterstates
```
LalatenduMohanty: IMO we should add a `shortNames` here as it will be used more often I guess. May be `cs`. But this is just a wish list and not blocker to merge the PR

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	case err != nil:
+		logger.WithError(err).Error("Error getting cluster deployment")
+		return reconcile.Result{}, err
+	}
+	if !st.DeletionTimestamp.IsZero() {
+		logger.Debug("ClusterState resource has been deleted")
+		return reconcile.Result{}, nil
+	}
+	if st.Status.LastUpdated != nil && time.Since(st.Status.LastUpdated.Time) < statusUpdateInterval {
+		nextUpdateWait := statusUpdateInterval - time.Since(st.Status.LastUpdated.Time) + 60*time.Second
```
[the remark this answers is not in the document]
csrwng: Thinking more about it, no benefit... it will get reqeued anyway. I was thinking of the case where the amount of time elapsed is just shy of the interval, that it would maybe make little sense to RequeueAfter with a very small duration. But I guess it doesn't matter. Will remove.

csrwng:
last round or review comments addressed

csrwng:
@LalatenduMohanty I removed the short name for the type because I found out it's already used by `ComponentStatus`

pkg/controller/clusterstate/clusterstate_controller.go
```diff
@@ -0,0 +1,304 @@
...
+		return reconcile.Result{}, nil
+	case err != nil:
+		logger.WithError(err).Error("Error getting cluster deployment")
+		return reconcile.Result{}, err
+	}
+	if !st.DeletionTimestamp.IsZero() {
+		logger.Debug("ClusterState resource has been deleted")
+		// If the previous cluster state is getting deleted, requeue in a few seconds
+		// to check again and recreate it if necessary.
+		return reconcile.Result{RequeueAfter: 60 * time.Second}, nil
```
LalatenduMohanty: Now that we are waiting for 60 seconds (with hard coded value), we need to add a log info  here. At least someone going through the log can figure out why the next reconcile  happening after 60 second.

csrwng: Added log line

PR #506, opened 2019-08-19
hive-frontend add events to role

maorfr:
This pR allows the hive-frontend ServiceAccount to manage Events.
This is required for https://jira.coreos.com/browse/APPSRE-882

Please let us know if that is the correct way to achieve this.

Thanks!


LalatenduMohanty:
It should have been a single commit we do not need two commits for the change, missed that in the review.

maorfr:
sorry about that :|

PR #507, opened 2019-08-19
Fix typo in operator role for syncset admission controller

staebler:
Change the permissions for the operator role to use `sync.admission.hive.operator.io` instead of `rsync.admission.hive.operator.io`.


staebler:
Thanks, @LalatenduMohanty. I have fixed the other reference now, too.

PR #509, opened 2019-08-21
wrapping around event handler that adds logging

staebler:
I found this helpful when I was verifying some behavior. I though it might be nice to keep in our toolbox.


csrwng:
Very nice 
/lgtm

PR #510, opened 2019-08-23
Update makefile to run golangci lint bydefault

LalatenduMohanty:
Also updated the installation method to make it work in CI systems as per https://github.com/golangci/golangci-lint#ci-installation

Commented some of the check in `golangci.yml` because either these need more effort to fix or not critical as compared to making `golangci-lint` default in `make test`. 

Once  `golangci-lint` is enabled it will help fixing incoming code for lint issues and we can gradually add more checks by removing the comments in  `golangci.yml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,10 @@ SRC_DIRS = pkg contrib
 GOFILES = $(shell find $(SRC_DIRS) -name '*.go' | grep -v bindata)
 VERIFY_IMPORTS_CONFIG = build/verify-imports/import-rules.yaml
 
+FIRST_GOPATH:=$(firstword $(subst :, ,$(shell go env GOPATH)))
+GOLANGCI_LINT_BIN=$(FIRST_GOPATH)/bin/golangci-lint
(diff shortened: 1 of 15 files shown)


csrwng:
/lgtm cancel
@LalatenduMohanty looks like the linter is not happy when run by CI:
```
/go/bin/golangci-lint run -c ./golangci.yml ./pkg/... ./cmd/... ./contrib/...
pkg/controller/clusterdeployment/clusterdeployment_controller.go:1494:65: `retrived` is a misspelling of `retrieved` (misspell)
			return "", errors.Wrap(err, "global pull secret could not be retrived")
			                                                             ^
```

LalatenduMohanty:
> @LalatenduMohanty looks like the linter is not happy when run by CI:
The last PR that got merged had a spelling mistake. :)

LalatenduMohanty:
@dgoodwin This is ready for merge.

LalatenduMohanty:
@dgoodwin You need to add lgtm again as, we merged (from the last time unit test had ran for this PR) some code with lint issues and the unit test failed again.

PR #512, opened 2019-08-26
Manually manage controller RBAC.

dgoodwin:
Removing all use of kubebuilder:rbac annotations. We have multiple
components in this repo, with all others using manual RBAC. Annotations
not buying us much and somewhat misleading now, so keeping the generated
yaml and manually managing from here on out.


dgoodwin:
Sorry didn't realize this now was conflicting. Updated and needs another lgtm.

PR #514, opened 2019-08-29
Moving the installation of golangci-lint to Dockerfile

LalatenduMohanty:
Sending the PR to just change the Dockerfile as without this is merged the unit tests will never pass for the Makefile change in https://github.com/openshift/hive/pull/528

Fixes part of  https://jira.coreos.com/browse/CO-539

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -21,3 +21,7 @@ RUN export version=1.0.8 && \
 
 # Get rid of "go: disabling cache ..." errors.
 RUN mkdir -p /go && chgrp -R root /go && chmod -R g+rwX /go
+
+RUN export GOLANGCI_LINT_VERSION=v1.17.1 &&\
+    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s $(GOLANGCI_LINT_VERSION) && \
+    mv ./bin/golangci-lint /usr/local/bin/golangci-lint


build/build-image/Dockerfile
```diff
@@ -21,3 +21,8 @@ RUN export version=1.0.8 && \
 
 # Get rid of "go: disabling cache ..." errors.
 RUN mkdir -p /go && chgrp -R root /go && chmod -R g+rwX /go
+
+RUN export GOLANGCI_LINT_VERSION=v1.17.1 &&\
+    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s $(GOLANGCI_LINT_VERSION) && \
+    mv ./bin/golangci-lint /usr/local/golangci-lint && \
```
joelddiaz: don't you want this in /usr/local/bin so it's in one of the default bin search paths?

staebler: You don't need to move the binary. You can still use the `-b` flag to place it in whichever bin directory you want. My point before was that we do not need to parse GOPATH to put it in the go bin.
Also, I think that the default `/bin` is in the bin search path and it is fine to leave the binary there.

LalatenduMohanty: Right, I have similar understanding but I am getting. Going to try few things. 
```
make: golangci-lint: Command not found
make: *** [lint] Error 127
golangci-lint run -c ./golangci.yml ./pkg/... ./cmd/... ./contrib/...
---
```

LalatenduMohanty: yes, I wanted put it in `/usr/local/bin` but mistakenly out it in `/usr/local`

LalatenduMohanty:
@staebler @joelddiaz need the `lgtm` again.

PR #516, opened 2019-08-29
Add ssh-agent to hive container for log gathering.

dgoodwin:
This was present in the Dockerfile.dev we were using, but not in the
final layer we based on for official builds. Unfortunately RHEL based
containers cannot run yum during builds due to lack of entitlements.
With this PR we lose the ability to build the main Dockerfile locally,
but we still have the much faster Dockerfile.dev and associated make
targets.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,11 @@ RUN go build -o bin/hiveutil github.com/openshift/hive/contrib/cmd/hiveutil
 RUN go build -o bin/hiveadmission github.com/openshift/hive/cmd/hiveadmission
 RUN go build -o bin/hive-operator github.com/openshift/hive/cmd/operator
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base as hive
+FROM registry.access.redhat.com/ubi7/ubi
+
+# ssh-agent required for gathering logs in some situations:
+RUN if ! rpm -q openssh-clients; then yum install -y openssh-clients && yum clean all && rm -rf /var/cache/yum/*; fi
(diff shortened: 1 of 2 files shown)


dgoodwin:
The CI system can't yum install with those base images either: https://ci.ext.devshift.net/job/openshift-hive-gh-pr-check/895/console

dgoodwin:
Interestingly it's the app-sre jenkins that is failing, the regular OpenShift CI is able to build it. I am testing use of the universal base image.

dgoodwin:
Switched to UBI.

Dockerfile
```diff
@@ -7,7 +7,11 @@ RUN go build -o bin/hiveutil github.com/openshift/hive/contrib/cmd/hiveutil
 RUN go build -o bin/hiveadmission github.com/openshift/hive/cmd/hiveadmission
 RUN go build -o bin/hive-operator github.com/openshift/hive/cmd/operator
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base as hive
+FROM registry.access.redhat.com/ubi7/ubi
```
staebler: `registry.svc.ci.openshift.org/ocp/4.2:base` is built from that same ubi7 image.

dgoodwin: It cannot be pulled without auth, I think that's true for all ocp images.

staebler: Ah. I remember you saying that before. Sorry for being so hard-headed.

dgoodwin: No problem!

PR #518, opened 2019-08-30
fix clusterprovision creation race

staebler:
Use expectations to ensure that clusterDeployment controller does not create multiple overlapping clusterprovisions.

There is currently a race condition where the controller will create a second clusterprovision if the first clusterprovision has not been added to the local cache by the time the clusterdeployment is reconciled again after creating the first clusterprovision. To prevent this, these changes use expectations to block the controller from creating the second clusterprovision unitl the controller sees that the first clusterprovision has been created.

When the clusterDeployment controller creates a clusterprovision, the controller will increase the number of creation expectations for the clusterdeployment. When the controller sees an add event for the clusterprovision, the controller will reduce the number of creation expectations for the clusterdeployment. During a reconcile loop, the controller will not look at clusterprovisions while there are outstanding creation expectations for the clusterdeployment being reconciled.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -44,6 +44,9 @@ import (
 	"github.com/openshift/hive/pkg/install"
 )
 
+// controllerKind contains the schema.GroupVersionKind for this controller type.
+var controllerKind = hivev1.SchemeGroupVersion.WithKind("ClusterDeployment")
+
 const (
 	controllerName     = "clusterDeployment"
 	defaultRequeueTime = 10 * time.Second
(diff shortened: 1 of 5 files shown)


dgoodwin:
Working through trying to understand this but initial thought, could you update the git commit msg on the second commit with some heavier details on what this is doing, how it works and why it's needed in this specific situation.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -566,8 +582,11 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		cdLog.WithError(err).Error("could not set the owner ref on provision")
 		return reconcile.Result{}, err
 	}
+
+	r.expectations.ExpectCreations(types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}.String(), 1)
 	if err := r.Create(context.TODO(), provision); err != nil {
 		cdLog.WithError(err).Error("could not create provision")
+		r.expectations.CreationObserved(types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}.String())
```
dgoodwin: Something that's confusing me a little bit here, it looks like the intent here is to expect the creation of the cluster provision, but the expectation itself is using the cluster deployment's name. Should this be the provision's name? Or is that an issue with the random component of their names.

staebler:
> Working through trying to understand this but initial thought, could you update the git commit msg on the second commit with some heavier details on what this is doing, how it works and why it's needed in this specific situation.

Do you mean the first commit?

staebler:
I need to copy the controller expectations code from `k8s.io/kubernetes/pkg/controller` into our codebase. That package is bringing in too much other stuff.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: The expectations are set up to expect a creation on behalf of the clusterdeployment. The key is always the resource being reconciled, and not the resource being created.

dgoodwin:
+1 copying sounds good.

I meant the commit with your code :) Whichever it was.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: gotcha thanks.

staebler:
> +1 copying sounds good.
> 
> I meant the commit with your code :) Whichever it was.

Sure. I will improve on the commit message.

dgoodwin:
This is great, thank you.

/lgtm

PR #520, opened 2019-08-30
Remove SyncSetStatus and SelectorSyncSetStatus from clusterdeployment

staebler:
These are not present in any clusterdeployments in stage or production.

Both Stage and Production give the following output.
```
$ oc get cd --all-namespaces -o json | jq ".items[].status.syncSetStatus | select(. != null)"
$ oc get cd --all-namespaces -o json | jq ".items[].status.selectorSyncSetStatus | select(. != null)"
```


staebler:
c07804a2...fc934293
* Rebased onto master

PR #521, opened 2019-08-30
remove deletion of legacy hiveadmission daemonset

staebler:
The code to delete the legacy hiveadmission daemonset has been out on stage and production, so those daemonsets should no longer exist.


dgoodwin:
/lgtm

Thanks for cleaning this up.

PR #522, opened 2019-08-30
Local e2e example and instructions

csrwng:
- Modifies the e2e script to allow overriding of environment variables. 
- Adds instructions on how to run the e2e test locally 
- Provides a sample script to run the e2e test locally
- Modifies the `go test` call in the Makefile for post-install and post-deploy tests so that results are not cached

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,11 +51,11 @@ test-e2e:
 
 .PHONY: test-e2e-postdeploy
 test-e2e-postdeploy:
-	go test -timeout 0 ./test/e2e/postdeploy/...
+	go test -timeout 0 -count=1 ./test/e2e/postdeploy/...
 
 .PHONY: test-e2e-postinstall
 test-e2e-postinstall:
-	go test -timeout 0 ./test/e2e/postinstall/...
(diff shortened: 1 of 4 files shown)


dgoodwin:
Great step, thanks!

/lgtm

I think we should do a follow up at some point to skip the hive installation. Personally I'd be looking to use this after writing code and I likely already have Hive running, maybe from source, and I want to see that it can successfully provision a cluster and pass the tests on it.

csrwng:
> I think we should do a follow up at some point to skip the hive installation. Personally I'd be looking to use this after writing code and I likely already have Hive running, maybe from source, and I want to see that it can successfully provision a cluster and pass the tests on it.

You can already do that (and maybe I should've added it to the doc). Once you have deployed Hive, you can verify that things are working well with:
`go test ./test/e2e/postdeploy/...`

If you've created a clusterdeployment and it has been installed, you can run the postinstall tests with:
`go test ./test/e2e/postinstall/...`

dgoodwin:
Oooo ok that's awesome, thanks!

PR #524, opened 2019-09-03
Refactor AWS APIs to separate package, remove OpenStack and Libvirt.

dgoodwin:
This is to better diff and sync with the installer, and potentially to
begin reusing their validation code.

OpenStack and Libvirt are removed and were not actually supported with
Hive.


dgoodwin:
Sorry needs another lgtm @staebler , was a compile error lurking in e2e tests.

PR #525, opened 2019-09-04
Fix crash on nil remote MachineSet ProviderSpec.

dgoodwin:


pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -193,6 +193,11 @@ func (r *ReconcileRemoteMachineSet) syncMachineSets(
 	// we extract installer image refs.
 	var amiID string
 	for _, ms := range remoteMachineSets.Items {
+		if ms.Spec.Template.Spec.ProviderSpec.Value == nil {
+			cdLog.WithField("machineSet", ms.Name).Warn("machine set has nil ProviderSpec")
```
staebler: Is this something that deserves a warning? It is valid for a machine set to have an empty provider spec. Are we assuming that the user is not allowed to create their own machine sets?

staebler: This leads me to the concern that this code that searches for the AMI ID may not be resilient enough to user-created machinesets, where the user may use a different or invalid AMI ID.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -493,6 +498,10 @@ func fetchAvailabilityZones(client awsclient.Client, region string) ([]string, e
 func decodeAWSMachineProviderSpec(rawExt *runtime.RawExtension, scheme *runtime.Scheme) (*awsprovider.AWSMachineProviderConfig, error) {
 	codecFactory := serializer.NewCodecFactory(scheme)
 	decoder := codecFactory.UniversalDecoder(awsprovider.SchemeGroupVersion)
+	if rawExt == nil {
```
staebler: This is redundant with the check added in `syncMachineSets`. My preference is to keep the check here and remove the check in `syncMachineSets`.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -193,6 +193,11 @@ func (r *ReconcileRemoteMachineSet) syncMachineSets(
...
 	var amiID string
 	for _, ms := range remoteMachineSets.Items {
+		if ms.Spec.Template.Spec.ProviderSpec.Value == nil {
+			cdLog.WithField("machineSet", ms.Name).Warn("machine set has nil ProviderSpec")
+			// keep looking for an AMI
+			continue
+		}
 		awsProviderSpec, err := decodeAWSMachineProviderSpec(ms.Spec.Template.Spec.ProviderSpec.Value, r.scheme)
 		if err != nil {
 			cdLog.WithError(err).Error("error decoding AWSMachineProviderConfig")
```
staebler: This should probably not be an error log. It should not be the job of hive to report errors in user-configured machinesets. Also, the function should not return an error in this case. It should just skip the machineset.

pkg/controller/remotemachineset/remotemachineset_controller.go
dgoodwin: Hive presently assumes full control of machine sets in any Hive managed cluster. In future I could see someone requesting disabling this or supporting user created, but not yet.

As such seeing a machine set with no ProviderSpec is pretty weird and for our purposes, we should probably be getting a warning in our logs.

dgoodwin:
Updated to warn and continue instead of err, and dropped the nil check at the higher level.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -493,6 +493,10 @@ func fetchAvailabilityZones(client awsclient.Client, region string) ([]string, e
 func decodeAWSMachineProviderSpec(rawExt *runtime.RawExtension, scheme *runtime.Scheme) (*awsprovider.AWSMachineProviderConfig, error) {
 	codecFactory := serializer.NewCodecFactory(scheme)
 	decoder := codecFactory.UniversalDecoder(awsprovider.SchemeGroupVersion)
+	if rawExt == nil {
+		// If the machine set has no ProviderSpec, return an empty object.
+		return &awsprovider.AWSMachineProviderConfig{}, nil
```
staebler: Do you want this to return an error now so that it is output as a warning log back in `syncMachineSets`?

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -311,6 +314,37 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+			expectedRemoteMachineSets: func() *machineapi.MachineSetList {
+				return &machineapi.MachineSetList{
+					Items: []machineapi.MachineSet{
+						*testMachineSet("foo-12345-worker-us-east-1a", "worker", true, 1, 0),
+						*testMachineSet("foo-12345-worker-us-east-1b", "worker", true, 1, 0),
+						*testMachineSet("foo-12345-worker-us-east-1c", "worker", true, 1, 0),
+					},
+				}
+			}(),
+			skipAMICheck: true,
```
staebler: Why not adjust the expected machine set to have a nil provider spec? That is what is expected after all.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -77,7 +77,10 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 		localExisting             []runtime.Object
 		remoteExisting            []runtime.Object
 		expectErr                 bool
+		expectDecodeErr           bool
 		expectedRemoteMachineSets *machineapi.MachineSetList
+		// Hack to avoid an assertion in the nil ProviderSpec test, until we decide if we should
```
staebler: Remove this comment.

dgoodwin:
Updated again, had to redo the test approach as a result of the change to returning an error from decode, it now uses just one machineset in the test, thus no AMI can be found and the overall decode errors, avoiding all the problems in the test harness. This is sufficient to expose the bug in that we can get through the loop for a bad machineset without crashing.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -373,23 +411,23 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 									t.Errorf("machineset %v has unexpected taints:\nexpected: %v\nactual: %v", eMS.Name, eMS.Spec.Template.Spec.Taints, rMS.Spec.Template.Spec.Taints)
 								}
 
-								rAWSProviderSpec, err := decodeAWSMachineProviderSpec(
+								rAWSProviderSpec, _ := decodeAWSMachineProviderSpec(
 									rMS.Spec.Template.Spec.ProviderSpec.Value, scheme.Scheme)
 								log.Debugf("remote AWS: %v", printAWSMachineProviderConfig(rAWSProviderSpec))
 								assert.NoError(t, err)
```
staebler: Remove this line.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -373,23 +411,23 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
 								assert.NotNil(t, rAWSProviderSpec)
-								eAWSProviderSpec, err := decodeAWSMachineProviderSpec(
+
+								eAWSProviderSpec, _ := decodeAWSMachineProviderSpec(
 									eMS.Spec.Template.Spec.ProviderSpec.Value, scheme.Scheme)
+								log.Debugf("expected AWS: %v", printAWSMachineProviderConfig(eAWSProviderSpec))
 								assert.NoError(t, err)
```
staebler: Remove this line.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -345,9 +378,14 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 
 			if err != nil && !test.expectErr {
 				t.Errorf("unexpected error: %v", err)
+				return
 			}
-			if err == nil && test.expectErr {
-				t.Errorf("expected error but got none")
+			if test.expectErr {
```
staebler: I find it a bit clearer if you move the `if test.expectErr` above the preceding if.

```
if test.expectErr {
  assert.Error(t, err, "expected error")
  return
}
if !assert.NoError(t, err, "unexpected error") {
  return
}

File: pkg/controller/remotemachineset/remotemachineset_controller_test.go [hunk 10]
```diff
@@ -311,6 +313,37 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+				testSecret(sshKeySecret, sshKeySecretKey, testName),
+			},
+			remoteExisting: []runtime.Object{
+				func() *machineapi.MachineSet {
+					ms := testMachineSet("foo-12345-worker-us-east-1c", "worker", true, 1, 0)
+					ms.Spec.Template.Spec.ProviderSpec.Value = nil
+					return ms
+				}(),
+			},
+			expectedRemoteMachineSets: func() *machineapi.MachineSetList {
```
staebler: We don't need this. It will never be used since the test will short-circuit after checking the returned error.

staebler:
> Updated again, had to redo the test approach as a result of the change to returning an error from decode, it now uses just one machineset in the test, thus no AMI can be found and the overall decode errors, avoiding all the problems in the test harness. This is sufficient to expose the bug in that we can get through the loop for a bad machineset without crashing.

This is *well* beyond the scope of this PR, but couldn't we directly compare the actual machine set and the expected machine set, rather going through all the pain of comparing each individual element of the machine sets? A side benefit beyond vastly reducing the code would also be that we would not have to rely on using code from the component that is under test to verify behavior.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
dgoodwin: removed

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -343,11 +363,16 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
 			})
 
+			if test.expectErr {
+				if err == nil {
+					t.Errorf("expected error but got none")
+				}
+				// Should not proceed with test validations if we expected an error from the reconcile.
+				return
+			}
 			if err != nil && !test.expectErr {
```
staebler: The check of `!test.expectErr` is redundant. The test already returns earlier if `test.expectErr` is true.

PR #529, opened 2019-09-05
fix gathering of install logs on e2e teardown

staebler:
In the teardown of e2e tests, look for the install pod using label selectors instead of expecting a fixed name. Also, gather the install logs from the clusterprovision instead of looking for a configmap.


dgoodwin:
great catch.

/lgtm

PR #535, opened 2019-09-09
create dummy clusterprovision for legacy clusterdeployment

staebler:
Redo of https://github.com/openshift/hive/pull/533 with a fix for the hot loop creating. The clusterDeployment controller will attempt to adopt an existing provision prior to creating a dummy provision. The controller will also use expectations to protect against creating a second clusterprovision in the interregnum between creation of the first and fetching of it from the API server.

---
If a clusterdeployment was installed prior to the changes that brought
clusterprovisions to be, then the clusterdeployment will not have a
clusterprovision associated with it. This is not a problem for normal
operation. However, if the clusterdeployment is backed up and restored,
then the clusterdeployment will lose its infra ID and kubeconfig
secret reference, since those are stored in the status. Normally,
after a restore, these items--along with others--are copied back into
the clusterdeployment from the clusterprovision. So, in order for the
clusterdeployment to work after being restored, the controller needs
to create a dummy clusterprovision for the clusterdeployment with
the pertinent data to be restored.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -395,20 +395,25 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			r.cleanupInstallLogPVC(cd, cdLog)
 			return reconcile.Result{}, nil
 		}
-		if cd.Status.Provision == nil {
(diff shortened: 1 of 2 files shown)


dgoodwin:
/hold

Do we want the revert to go out first?

LalatenduMohanty:
@dgoodwin @staebler What is the plan for this PR?

PR #539, opened 2019-09-10
Change Hive DR to use checkpoint object

twiest:
- [x] Create a checkpoint CRD that is namespace scoped.
- [x] Change Hive DR to use the namespace scoped checkpoint CR.
      Since the namespace is the thing being backed up, this allows
      for the tracking of the checksum of what was backed up as well
      as when it was backed up to be stored in a single object in the
      namespace.
- [x] Clean up any code that is no longer being used given the new
      algorithm.
- [x] Add tests for the new code.
- [x] Manually test the system against a real kube cluster.


pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
 
 	return r.Create(context.TODO(), backup)
 }
 
-func (r *ReconcileBackup) getModifiedHiveObjectsInNamespace(namespace string) ([]*hiveObject, error) {
-	nsLogger := addNamespaceLoggerFields(r.logger, namespace)
-
-	filteredHiveObjects := []*hiveObject{}
+func addNamespaceLoggerFields(logger log.FieldLogger, namespace string) log.FieldLogger {
+	return logger.WithField("namespace", namespace)
```
staebler: This field should be included in the logger for all of the logs in `Reconcile`. It should be added at the same time as the fields in `addReconcileRequestLoggerFields` are added. At that point, there is no need for either this function or `addReconcileRequestLoggerFields` as they would each only be called once and the replacement code inline would be just as concise.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
-		if err != nil {
-			hobjLogger.WithError(err).Error("Failed update.")
-			return err // If we error at all, then we should just return. The NS will be requeued in the future.
-		}
+	if err != nil {
+		nsLogger.WithError(err).Error("Failed to create or update CheckPoint object in namespace.")
+		return err
 	}
 
+	// Everything succeeded!
```
staebler: I'm not sure that this comment is especially helpful. If anything, I would rather see the following.

```
if err != nil {
  nsLogger.WithError(err).Error(...)
}

return err
```

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
-}
-
-func getRuntimeObjects(kclient client.Client, typesToList []runtime.Object, namespace string) ([]runtime.Object, error) {
-	nsObjects := []runtime.Object{}
-
-	for _, t := range typesToList {
-		listObj := t.DeepCopyObject()
-		if err := kclient.List(context.TODO(), listObj, client.InNamespace(namespace)); err != nil {
-			return nil, err
+// calculateChecksum computes the runtime object's hash WITHOUT the status as status is not used to determine if the object has changed.
```
staebler: We also need to clear out `ResourceVersion` from the metadata. That field will be incremented with changes to the status, effecting a new backup when the status changes.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
+			meta = &t.ObjectMeta
+			spec = &t.Spec
+		case *hivev1.SyncSet:
+			meta = &t.ObjectMeta
+			spec = &t.Spec
+		case *hivev1.DNSZone:
+			meta = &t.ObjectMeta
+			spec = &t.Spec
+		default:
+			r.logger.Infof("Unknown Type: %T", object)
```
staebler: This would at least be a Warning.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
-		if err := kclient.List(context.TODO(), listObj, client.InNamespace(namespace)); err != nil {
-			return nil, err
+// calculateChecksum computes the runtime object's hash WITHOUT the status as status is not used to determine if the object has changed.
+// We never return an error from this function as that could lead to a
+// situation where other objects change, but because one of the objects
+// checksum errored, then the other objects won't get backed up.
+// Instead for objects that error, we return a constant checksum value.
+// This function should rarely, if ever, return an error.
+func (r *ReconcileBackup) calculateObjectsChecksumWithoutStatus(objects ...runtime.Object) string {
+	checksums := []string{}
```
staebler: Create this slice with the expected size.
```
checksums := make(string[], len(objects))
for i, object := range objects {
   ...
   checksums[i] = checksum
}
```

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
+	for _, object := range objects {
+		var meta *metav1.ObjectMeta
+		var spec interface{}
+		var checksum = ""
+		var err error
```
staebler: Move this declaration down to where it is used.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
+			spec = &t.Spec
+		case *hivev1.SyncSet:
+			meta = &t.ObjectMeta
+			spec = &t.Spec
+		case *hivev1.DNSZone:
+			meta = &t.ObjectMeta
+			spec = &t.Spec
+		default:
+			r.logger.Infof("Unknown Type: %T", object)
+			checksum = errChecksum
```
staebler: If we do the following here, then we do not have to carry the `checksum` variable and do not have to have the check later of whether the checksum is the error checksum.
```
checksum[i] = errChecksum   // or append if not taking my other suggestion
continue
```

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -169,74 +185,124 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string) error {
...
-		nsObjects = append(nsObjects, list...)
+		checksums = append(checksums, checksum)
 	}
 
-	return nsObjects, nil
+	// This ensures that for the same set of objects we generate the same combined checksum.
+	// Otherwise a different order could cause the checksum to be different every time.
+	sort.Strings(checksums)
+
+	combinedChecksum, err := controllerutils.GetChecksumOfObjects(checksums)
```
staebler: The effect of creating a checksum is the same, but I think you actually want to do the following here. Or did you mean to call `GetChecksumOfObject`?
```
GetChecksumOfObjects(checksums...)
```

pkg/test/generic/generic.go
```diff
@@ -42,26 +40,3 @@ func WithAnnotationsPopulated() Option {
 		}
 	}
 }
-
-// ChecksumFunc defines a function signature for checksumming objects.
-type ChecksumFunc func(runtime.Object) string
```
staebler: The `ChecksumOfObjectFunc` and `ChecksumOfObjectsFunc` in `pkg/controller/utils/utils.go` are also not used by anything any more.

Also, while in that file, `GetChecksumOfObjects` can call `GetChecksumOfObject` to produce the same result without duplicating code.
```
func GetChecksumOfObjects(objects ...interface{}) (string, error) {
  return GetCheckumOfObject(objects)
}
```

pkg/controller/velerobackup/velerobackup_controller.go
twiest: @staebler Should the logger with the namespace field be passed from Reconcile down into each function it calls?

Or should r.logger be updated to include the field?

Not sure if r.logger should be updated as I believe that's an object that is shared between all calls to reconcile, which can be run simultaneously.

staebler: Definitely don't want to mutate `r.logger`. Whether you pass the logger down as a parameter or set it up in each function is your choice. The former aligns with how it is done in other controllers. The latter makes calling signatures cleaner, until there are fields that you want to include that are not already available in the called function.

staebler: Looking more closely at the logging, I see now that the logger used in `Reconcile` does indirectly include the namespace in the "request" field. It would be better, however, if the namespace were consistent throughout all of the logs generated by this controller. So either all of the logs use the "request" field or all of the logs use the "namespace" field.

There are also logs like [here](https://github.com/openshift/hive/blob/2913069b622b7d9fbc27db14d34e5b0dbdfe5395/pkg/controller/velerobackup/velerobackup_controller.go#L282) that do not include either field. It should include whichever you choose as well.

twiest:
@staebler discovered a bug while testing (wasn't setting the namespace correctly in the `ObjectReference`). Bug is fixed now, but the fix removed your LGTM.

PTAL.

/hold cancel

PR #542, opened 2019-09-13
Modifying some debug messages to info in dnszone controller

LalatenduMohanty:
It will help getting more useful information in the controller logs


dgoodwin:
It looks like I failed to submit my comment last week, but in addition to this it appears we have no logging when we fail to resolve the hostname. Please locate that point in the file and add an info line there as well, that will help us if this pops up again.

LalatenduMohanty:
> It looks like I failed to submit my comment last week, but in addition to this it appears we have no logging when we fail to resolve the hostname. Please locate that point in the file and add an info line there as well, that will help us if this pops up again.

I have added a log info for `SOA record for DNS zone not available`, please take a look.

PR #545, opened 2019-09-17
Adding SyncSetFailureCondition to cluster deployment

LalatenduMohanty:
SyncSetFailureCondition will be added if any SyncSet failed during application

### DIFF
--- a/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1alpha1/clusterdeployment_types.go
@@ -255,6 +255,9 @@ const (
 
 	// ProvisionFailedCondition indicates that a provision failed
 	ProvisionFailedCondition ClusterDeploymentConditionType = "ProvisionFailed"
+
+	// SyncSetFailedCondition indicates if any syncset for a cluster deployment failed
+	SyncSetFailedCondition ClusterDeploymentConditionType = "SyncSetFailed"
 )
 
 // AllClusterDeploymentConditions is a slice containing all condition types. This can be used for dealing with
@@ -268,6 +271,7 @@ var AllClusterDeploymentConditions = []ClusterDeploymentConditionType{
 	InstallFailingCondition,
 	DNSNotReadyCondition,
 	ProvisionFailedCondition,
+	SyncSetFailedCondition,
 }
 
 // +genclient
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -215,6 +215,15 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to SyncSetInstance
+	err = c.Watch(&source.Kind{Type: &hivev1.SyncSetInstance{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &hivev1.ClusterDeployment{},
+	})
+	if err != nil {
+		return fmt.Errorf("cannot start watch on syncset instance: %v", err)
+	}
+
 	return nil
 }
 
@@ -390,6 +399,17 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 	}
 
 	if cd.Spec.Installed {
+
+		// update SyncSetFailedCondition status condition
+		cdLog.Info("Check if any syncsetinstance Failed")
+		updateCD, err := r.setSyncSetFailedCondition(cd, cdLog)
+		if err != nil {
+			cdLog.WithError(err).Error("Error updating SyncSetFailedCondition status condition")
+			return reconcile.Result{}, err
+		} else if updateCD {
+			return reconcile.Result{}, nil
+		}
+
 		if cd.Status.InstalledTimestamp != nil {
 			cdLog.Debug("cluster is already installed, no processing of provision needed")
 			r.cleanupInstallLogPVC(cd, cdLog)
@@ -1771,4 +1791,101 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
 			pLog.WithError(err).Error("failed to delete old provision")
 		}
 	}
+}
+
+// getAllSyncSetInstances returns all syncset instances for a specific cluster deployment
+func (r *ReconcileClusterDeployment) getAllSyncSetInstances(cd *hivev1.ClusterDeployment) ([]*hivev1.SyncSetInstance, error) {
+
+	list := &hivev1.SyncSetInstanceList{}
+	err := r.List(context.TODO(), list, client.InNamespace(cd.Namespace))
+	if err != nil {
+		return nil, err
+	}
+
+	syncSetInstances := []*hivev1.SyncSetInstance{}
+	for i, syncSetInstance := range list.Items {
+		if syncSetInstance.Spec.ClusterDeployment.Name == cd.Name {
+			syncSetInstances = append(syncSetInstances, &list.Items[i])
+		}
+	}
+	return syncSetInstances, nil
+}
+
+// checkForFailedSyncSetInstance returns true if it finds failed syncset instance
+func checkForFailedSyncSetInstance(syncSetInstances []*hivev1.SyncSetInstance) bool {
+
+	for _, syncSetInstance := range syncSetInstances {
+		if checkSyncSetConditionsForFailure(syncSetInstance.Status.Conditions) {
+			return true
+		}
+		for _, r := range syncSetInstance.Status.Resources {
+			if checkSyncSetConditionsForFailure(r.Conditions) {
+				return true
+			}
+		}
+		for _, p := range syncSetInstance.Status.Patches {
+			if checkSyncSetConditionsForFailure(p.Conditions) {
+				return true
+			}
+		}
+		for _, s := range syncSetInstance.Status.SecretReferences {
+			if checkSyncSetConditionsForFailure(s.Conditions) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// checkSyncSetConditionsForFailure returns true when the condition contains hivev1.ApplyFailureSyncCondition
+// and condition status is equal to true
+func checkSyncSetConditionsForFailure(conds []hivev1.SyncCondition) bool {
+	for _, c := range conds {
+		if c.Status != corev1.ConditionTrue {
+			continue
+		}
+		switch c.Type {
+		case hivev1.ApplyFailureSyncCondition, hivev1.DeletionFailedSyncCondition, hivev1.UnknownObjectSyncCondition:
+			return true
+		}
+	}
+	return false
+}
+
+// setSyncSetFailedCondition returns true when it sets or updates the hivev1.SyncSetFailedCondition
+func (r *ReconcileClusterDeployment) setSyncSetFailedCondition(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (bool, error) {
+	// get all syncset instances for this cluster deployment
+	syncSetInstances, err := r.getAllSyncSetInstances(cd)
+	if err != nil {
+		cdLog.WithError(err).Error("Unable to list related syncset instances for cluster deployment")
+		return false, err
+	}
+
+	isFailedCondition := checkForFailedSyncSetInstance(syncSetInstances)
+
+	status := corev1.ConditionFalse
+	reason := "SyncSetApplySuccess"
+	message := "SyncSet apply is successful"
+	if isFailedCondition {
+		status = corev1.ConditionTrue
+		reason = "SyncSetApplyFailure"
+		message = "One of the SyncSetInstance apply has failed"
+	}
+	conds, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+		cd.Status.Conditions,
+		hivev1.SyncSetFailedCondition,
+		status,
+		reason,
+		message,
+		controllerutils.UpdateConditionNever,
+	)
+	if !changed {
+		return false, nil
+	}
+	cd.Status.Conditions = conds
+	if err := r.Status().Update(context.TODO(), cd); err != nil {
+		cdLog.WithError(err).Error("error updating syncset failed condition")
+		return false, err
+	}
+	return true, nil
 }
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -37,20 +37,21 @@ import (
 )
 
 const (
-	testName              = "foo-lqmsh"
-	testClusterName       = "bar"
-	testClusterID         = "testFooClusterUUID"
-	testInfraID           = "testFooInfraID"
-	provisionName         = "foo-lqmsh-random"
-	legacyInstallJobName  = "foo-lqmsh-install"
-	imageSetJobName       = "foo-lqmsh-imageset"
-	testNamespace         = "default"
-	metadataName          = "foo-lqmsh-metadata"
-	sshKeySecret          = "ssh-key"
-	pullSecretSecret      = "pull-secret"
-	globalPullSecret      = "global-pull-secret"
-	adminKubeconfigSecret = "foo-lqmsh-admin-kubeconfig"
-	adminKubeconfig       = `clusters:
+	testName                = "foo-lqmsh"
+	testClusterName         = "bar"
+	testClusterID           = "testFooClusterUUID"
+	testInfraID             = "testFooInfraID"
+	provisionName           = "foo-lqmsh-random"
+	legacyInstallJobName    = "foo-lqmsh-install"
+	imageSetJobName         = "foo-lqmsh-imageset"
+	testNamespace           = "default"
+	testSyncsetInstanceName = "testSSI"
+	metadataName            = "foo-lqmsh-metadata"
+	sshKeySecret            = "ssh-key"
+	pullSecretSecret        = "pull-secret"
+	globalPullSecret        = "global-pull-secret"
+	adminKubeconfigSecret   = "foo-lqmsh-admin-kubeconfig"
+	adminKubeconfig         = `clusters:
 - cluster:
     certificate-authority-data: JUNK
     server: https://bar-api.clusters.example.com:6443
@@ -1082,6 +1083,47 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "setSyncSetFailedCondition should be present",
+			existing: []runtime.Object{
+				testInstalledClusterDeployment(time.Now()),
+				createSyncSetInstanceObj(hivev1.ApplyFailureSyncCondition),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				if assert.NotNil(t, cd, "missing clusterdeployment") {
+					cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailedCondition)
+					if assert.NotNil(t, cond, "missing SyncSetFailedCondition status condition") {
+						assert.Equal(t, corev1.ConditionTrue, cond.Status, "did not get expected state for SyncSetFailedCondition condition")
+					}
+				}
+			},
+		},
+		{
+			name: "setSyncSetFailedCondition value should be corev1.ConditionFalse",
+			existing: []runtime.Object{
+
+				func() runtime.Object {
+					cd := testInstalledClusterDeployment(time.Now())
+					cd.Status.Conditions = append(
+						cd.Status.Conditions,
+						hivev1.ClusterDeploymentCondition{
+							Type:   hivev1.SyncSetFailedCondition,
+							Status: corev1.ConditionTrue,
+						},
+					)
+					return cd
+				}(),
+				createSyncSetInstanceObj(hivev1.ApplySuccessSyncCondition),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailedCondition)
+				if assert.NotNil(t, cond, "missing SyncSetFailedCondition status condition") {
+					assert.Equal(t, corev1.ConditionFalse, cond.Status, "did not get expected state for SyncSetFailedCondition condition")
+				}
+			},
+		},
 	}
 
 	for _, test := range tests {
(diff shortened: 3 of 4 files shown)


pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -247,6 +247,9 @@ const (
 
 	// ProvisionFailedCondition indicates that a provision failed
 	ProvisionFailedCondition ClusterDeploymentConditionType = "ProvisionFailed"
+
+	// SyncSetFailureCondition indicates if any syncset for a cluster deployment failed
+	SyncSetFailureCondition ClusterDeploymentConditionType = "false"
```
joelddiaz: you typically want a human readable text here "SyncSetFailed" (or something like that)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -215,6 +215,12 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to SyncSetInstance
+	err = c.Watch(&source.Kind{Type: &hivev1.SyncSetInstance{}}, &handler.EnqueueRequestForObject{})
```
joelddiaz: you want to queue up a ClusterDeployment here instead of the SyncSetInstance object

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1659,3 +1675,69 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
+		cdLog.WithError(err).Error("unable to list related sync set instances for cluster deployment")
+		return false, err
+	}
+
+	isfailedCondition := checkFailureSyncCondition(syncSetInstances)
+
+	// if failure found, add the cluster deployment status condition SyncSetFailureCondition
+	if isfailedCondition == true {
+		cd.Status.Conditions = controllerutils.SetClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailureCondition,
+			corev1.ConditionTrue, "SyncSetApplyFailure", "One of the SyncSetInstance apply has failed", controllerutils.UpdateConditionAlways)
```
joelddiaz: will UpdateConditionAlways put us in a hot loop by always updating the LastProbeTime?

joelddiaz: also, since ths function (updateSyncSetFailureCondition) is returning the bool indicating whether the condition was modified, you might want to look into the SetClusterDeploymentConditionWithChangeCheck() call since it will handle notifying you if any updates were made.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1659,3 +1675,69 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
+		// update the SyncSetFailureCondition if it exists on the cluster deployment
+		cd.Status.Conditions = controllerutils.SetClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailureCondition,
+			corev1.ConditionFalse, "SyncSetApplySuccess", "SyncSet apply is successful", controllerutils.UpdateConditionAlways)
+		err := r.Status().Update(context.TODO(), cd)
+		if err != nil {
+			cdLog.WithError(err).Error("error updating cluster deployment with SyncSetFailureCondition condition (= false)")
+			return false, err
+		}
+	}
+	return true, nil
```
joelddiaz: always returning true here for the non-error case means the caller will do an early exit of the controller. i think you want to use the results of UpdateClusterConditionWithChangeCheck()

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
LalatenduMohanty: Changed to `SyncSetFailedCondition` to keep the name consistent with other condition names.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: > will UpdateConditionAlways put us in a hot loop by always updating the LastProbeTime?

Right, I had plan to add the check existing value and update only if required.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: Right, I have refactored the code to fix it.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1750,3 +1769,82 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
+	// check if hivev1.SyncSetFailedCondition already set on the cluster deployment
+	currentCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailedCondition)
+
+	if currentCond == nil && isfailedCondition == false {
+		return false, nil
+	}
+
+	// add/update hivev1.SyncSetFailedCondition to true
+	if isfailedCondition == true {
+		if currentCond == nil || currentCond.Status == corev1.ConditionFalse {
```
LalatenduMohanty: With current code `lastTransitionTime:` wont change unless there is a change in condition status. I guess that is fine.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
joelddiaz: the SetClusterDeploymentConditionWithChangeCheck() will do a lot of the work of only updating when necessary and notifying you if something changed (so you know whether to call the Status().Update() ). is there any particular reason to replicate some of the functionality in that call here instead of just calling SetClusterDeploymentConditionWithChangeCheck()?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1750,3 +1769,85 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
```
joelddiaz: the SetClusterDeploymentConditionWithChangeCheck() will do a lot of the work of only updating when necessary and notifying you if something changed (so you know whether to call the Status().Update() ). is there any particular reason to replicate some of the functionality in that call here instead of just calling SetClusterDeploymentConditionWithChangeCheck()?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1082,6 +1084,28 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "setSyncSetFailedCondition should be added for failed syncsetinstance",
+			existing: []runtime.Object{
+				func() runtime.Object {
```
staebler: Use `testInstalledClusterDeployment(time.Now)`.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1843,3 +1867,39 @@ func getProvisions(c client.Client) []*hivev1.ClusterProvision {
 	}
 	return provisions
 }
+
+func createSyncSetInstanceObj(syncCondType hivev1.SyncConditionType) *hivev1.SyncSetInstance {
+	ssi := &hivev1.SyncSetInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testSyncsetInstanceName,
+			Namespace: constants.HiveNamespace,
```
staebler: The namespace should be `testNamespace`.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1843,3 +1867,39 @@ func getProvisions(c client.Client) []*hivev1.ClusterProvision {
...
+		},
+	}
+	ssi.Spec.ClusterDeployment.Name = testClusterName
```
staebler: The name of the cluster should be `testName`.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1082,6 +1084,28 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+				}(),
+				createSyncSetInstanceObj(hivev1.ApplyFailureSyncCondition),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailedCondition)
+				if cond == nil {
+					t.Error("Status condition SyncSetFailedCondition expected in cluster deployment")
+				} else {
+					if cond.Status != corev1.ConditionTrue {
```
staebler:
```
assert.Equal(t, corev1.ConditionTrue, cond.Status, ...)
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1843,3 +1867,39 @@ func getProvisions(c client.Client) []*hivev1.ClusterProvision {
...
+	ssi.Status = createSyncSetInstanceStatus(syncCondType)
+	return ssi
+}
+
+func createSyncSetInstanceStatus(syncCondType hivev1.SyncConditionType) hivev1.SyncSetInstanceStatus {
+	conditionTime := metav1.NewTime(time.Now())
+	var ssiStatus corev1.ConditionStatus
+	var condType hivev1.SyncConditionType
+	if syncCondType == hivev1.ApplyFailureSyncCondition {
+		ssiStatus = corev1.ConditionFalse
```
staebler: This status should be `corev1.ConditionTrue`. The condition is true when there is a failure applying the sync set.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1843,3 +1867,39 @@ func getProvisions(c client.Client) []*hivev1.ClusterProvision {
...
+		condType = syncCondType
+	} else {
+		ssiStatus = corev1.ConditionTrue
+		condType = hivev1.ApplySuccessSyncCondition
```
staebler: Why is this set to `hivev1.ApplySuccsessSyncCondition` instead of `syncCondType`?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -37,20 +37,22 @@ import (
...
-	sshKeySecret          = "ssh-key"
-	pullSecretSecret      = "pull-secret"
-	globalPullSecret      = "global-pull-secret"
-	adminKubeconfigSecret = "foo-lqmsh-admin-kubeconfig"
-	adminKubeconfig       = `clusters:
+	testName        = "foo-lqmsh"
+	testClusterName = "bar"
+	testClusterID   = "testFooClusterUUID"
+	testInfraID     = "testFooInfraID"
+
```
staebler: What is the purpose of this new blank line?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1750,3 +1770,86 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
 		}
 	}
 }
+
+// getAllSyncSetInstances return all syncset instances for a cluster deployment
+func (r *ReconcileClusterDeployment) getAllSyncSetInstances(cd *hivev1.ClusterDeployment) ([]*hivev1.SyncSetInstance, error) {
+
+	list := &hivev1.SyncSetInstanceList{}
+	err := r.Client.List(context.TODO(), list, client.InNamespace(cd.Namespace))
```
staebler: Use `r.List` instead of `r.Client.List`.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1750,3 +1770,86 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
+	if err != nil {
+		return nil, err
+	}
+
+	syncSetInstances := []*hivev1.SyncSetInstance{}
+	for i, syncSetInstance := range list.Items {
+		fmt.Println("syncsetinstance name: ", syncSetInstance.Name)
+		if syncSetInstance.Spec.ClusterDeployment.Name == cd.Name {
```
staebler: We should add the `constants.ClusterDeploymentNameLabel` label to the syncsetinstance so that the `List` returns only the relevant resources. Then, this name comparison would not be needed. It is good practice to have that label on the syncsetinstance anyway.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1750,3 +1770,86 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
+
+	return syncSetInstances, nil
+}
+
+// checkFailureSyncCondition return true when syncsetinstance status condition has hivev1.ApplyFailureSyncCondition
+func checkFailureSyncCondition(syncSetInstances []*hivev1.SyncSetInstance) bool {
+	for _, syncSetInstance := range syncSetInstances {
+		for _, condition := range syncSetInstance.Status.Conditions {
+			fmt.Println("Condition Type: ", condition.Type)
+			if condition.Type == hivev1.ApplyFailureSyncCondition {
```
staebler: You need to check `condition.Status` as well. When the syncsetinstance is applied successfully, the syncsetinstance will still have this condition, but it will have a status of false.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: That is how `lastTransitionTime` should behave. It is the last time that the condition transitioned from one status to another.

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
```diff
@@ -247,6 +247,9 @@ const (
 
 	// ProvisionFailedCondition indicates that a provision failed
 	ProvisionFailedCondition ClusterDeploymentConditionType = "ProvisionFailed"
+
+	// SyncSetFailedCondition indicates if any syncset for a cluster deployment failed
+	SyncSetFailedCondition ClusterDeploymentConditionType = "false"
```
staebler: s/false/SyncSetFailed/

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: Just keep the code more readable. Do you think it makes any difference?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: This is information to me. I was not aware of this.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
LalatenduMohanty: Are you sure about this as I can see the cluster deployment name is ` testClusterName`
```
func testClusterDeployment() *hivev1.ClusterDeployment {
	cd := testEmptyClusterDeployment()

	cd.Spec = hivev1.ClusterDeploymentSpec{
		ClusterName: testClusterName,
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: I am calling `SetClusterDeploymentCondition` which calls `SetClusterDeploymentConditionWithChangeCheck` to do the update. So finally the update does not happen is the condition is not getting changed  or the condition true. The code I have is complimentary to `SetClusterDeploymentConditionWithChangeCheck`. I have done similar in case of unreachable controller.

pkg/apis/hive/v1alpha1/clusterdeployment_types.go
LalatenduMohanty: fixed it.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: +1, sounds good to me. IMO I should fix this post this PR

LalatenduMohanty: Also we should add the label to the syncset objects as well, isn't? @abutcher WDYT?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: The point is that there is no need to have all of this code here since `SetClusterDeploymentConditionWithChangeCheck` gives you everything that you need.

Everything after `cdLog.Infof("isFailedCondition: %v", isFailedCondition)` can be replaced with the following, which is more concise and arguably more readable
```
	status := corev1.ConditionFalse
	reason := "SyncSetApplySuccess"
	message := "SyncSet apply is successful"
	if isFailedCondition {
		status = corev1.ConditionTrue
		reason = "SyncSetApplyFailure"
		message = "One of the SyncSetInstance apply has failed"
	}
	conds, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
		cd.Status.Conditions,
		hivev1.SyncSetFailedCondition,
		status,
		reason,
		message,
		controllerutils.UpdateConditionNever,
	)
	if !changed {
		return false, nil
	}
	cd.Status.Conditions = conds
	if err := r.Status().Update(context.TODO(), cd); err != nil {
		cdLog.WithError(err).Error("error updating syncset failed condition")
		return false, err
	}
	return true, nil
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
staebler: How is it more readable if the code does something unexpected like use a hardcoded condition type instead of the condition type passed in as an argument? As a reader, it makes me have to pause to reason about why the `ApplySuccessSyncCondition` condition type is special.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: I don't know that the syncset resources should be labeled unless we are going to use that to associate the syncset with a clusterdeployment instead of using the reference in the spec of the syncset. The syncset resource is different that the syncsetinstance since one is created by the user and the other is created by the controller.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1082,6 +1083,50 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+		{
+			name: "setSyncSetFailedCondition should be present",
+			existing: []runtime.Object{
+				testInstalledClusterDeployment(time.Now()),
+				createSyncSetInstanceObj(hivev1.ApplyFailureSyncCondition),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.SyncSetFailedCondition)
+				if cond == nil {
```
staebler: You do not need to have an additional check of whether `cond` is `nil`.
```
if assert.NotNil(t, cond, ...) {
   assert.Equal(t, corev1.ConditionTrue, cond.Status, ...)
}

File: pkg/controller/clusterdeployment/clusterdeployment_controller_test.go [hunk 19]
```diff
@@ -1082,6 +1083,50 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+		},
+		{
+			name: "setSyncSetFailedCondition value should be corev1.ConditionFalse",
+			existing: []runtime.Object{
+
+				func() runtime.Object {
+					cd := testInstalledClusterDeployment(time.Now())
+					var conditions []hivev1.ClusterDeploymentCondition
+					cd.Status.Conditions = append(
+						conditions,
```
staebler: Append to `cd.Status.Conditions` instead of creating an empty slice to which to append. If you really want to ensure that `cd.Status.Conditions` contains *only* the one condition, then create a slice with that one condition and don't use append at all.
```
cd.Status.Conditions = append(cd.Status.Conditions, hivev1.ClusterDeploymentCondition{...})
```
or
```
cd.Stauts.Conditions = []hivev1.ClusterDeploymentCondition{{...}}
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1750,3 +1770,85 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
...
+			syncSetInstances = append(syncSetInstances, &list.Items[i])
+		}
+	}
+	return syncSetInstances, nil
+}
+
+// checkFailureSyncCondition return true when syncsetinstance status condition has hivev1.ApplyFailureSyncCondition
+func checkFailureSyncCondition(syncSetInstances []*hivev1.SyncSetInstance) bool {
+	for _, syncSetInstance := range syncSetInstances {
+		for _, condition := range syncSetInstance.Status.Conditions {
```
staebler: Need to search through the conditions for each of the resources, patches, and secret references. Also, need to look for `UnknownObjectSyncCondition` and `DeletionFailedSyncCondition` conditions types, too.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: Good point

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: Trying to understand  `UnknownObjectSyncCondition` and `DeletionFailedSyncCondition` . Thanks

LalatenduMohanty: I think `UnknownObjectSyncCondition` should be part of status condistions which we are going through, refer  https://github.com/openshift/hive/blob/master/pkg/controller/syncsetinstance/syncsetinstance_controller.go#L508

LalatenduMohanty: `DeletionFailedSyncCondition` as it is part of  `resources` https://github.com/openshift/hive/blob/master/pkg/controller/syncsetinstance/syncsetinstance_controller.go#L395-L423

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: @staebler Agree. I am going to use the above.

LalatenduMohanty:
/hold cancel 
@staebler @joelddiaz it is ready for review

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1082,6 +1083,45 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
 			},
 		},
+		{
+			name: "setSyncSetFailedCondition should be present",
+			existing: []runtime.Object{
+				testInstalledClusterDeployment(time.Now()),
+				createSyncSetInstanceObj(hivev1.ApplyFailureSyncCondition),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
```
joelddiaz: might want to assert that we didn't get back a nil clusterDeployment from getCD()

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1082,6 +1083,46 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+				assert.NotNil(t, cd, "cluster depployment can not be nil")
```
staebler: Should check the return from the assert and only use `cd` if the return is `true`.

staebler: s/depployment/deployment/

LalatenduMohanty:
@staebler fixed the comments, but it needs the `lgtm` again.

PR #547, opened 2019-09-17
Extend default lease duration, renew deadline and retry period for op…

abutcher:
…erator and controllers.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -30,8 +30,11 @@ import (
 )
 
 const (
-	defaultLogLevel         = "info"
-	leaderElectionConfigMap = "hive-controllers-leader"
+	defaultLogLevel             = "info"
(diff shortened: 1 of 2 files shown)


dgoodwin:
Looks good thank you. With these hard coded you could also do 2 * time.Minute or 90*time.Second, but this is ok as is.

/lgtm

dgoodwin:
@jmelis FYI this is our attempt to stop the controller restarts in stage. We fail to get that leader election maybe 1-4 times a day, seemingly since that catastrophe last week but we dont have a lot of logs before that to work with. We do see it in prod occasionally as well though much more rarely. Prior to this change it would try to write to the configmap for leader election for 2 seconds, up to a max of 10, and sometimes that fails. This seems weird so we're suspicious something is wrong, but we don't have access to any metrics to see.

PR #550, opened 2019-09-23
update syncset to use secretReferennce

wanghaoran1988:


dgoodwin:
Great idea @wanghaoran1988 , I don't think we'd discussed this and definitely didn't have it planned.

The PR looks correct as is to me, the one thing I'm worried about is I don't see a lot of pre-existing e2e testing around this functionality, so the PR might merge despite a problem syncing  ingress or certs. I'm adding two Jira's for adding e2e for serving cert and ingress.

@wanghaoran1988 if we take this to stage would it be easy for you to verify creating a dedicated cluster there, and checking that the serving cert is correctly a letsencrypt cert, and the ingress route secrets made it successfully? If not I will ask around and get quota to test it myself.

/lgtm

wanghaoran1988:
> @wanghaoran1988 if we take this to stage would it be easy for you to verify creating a dedicated cluster there, and checking that the serving cert is correctly a letsencrypt cert, and the ingress route secrets made it successfully? If not I will ask around and get quota to test it myself.
> 
I can do a test after this is deployed in stage

dgoodwin:
Awesome thanks, once this merges it will go to stage automatically but I will post a comment here once it's confirmed running, it takes a few minutes.

PR #551, opened 2019-09-23
Adding check for unreachable condition

LalatenduMohanty:


dgoodwin:
Yup this looks correct.

/lgtm

PR #552, opened 2019-09-24
Revert "update syncset to use secretRefence"

wanghaoran1988:
This reverts commit 919be0510ac1b1d5e46b9a23ae8b0104d714121d.
I got error when testing:

> secrets \"haowang-test-primary-cert-bundle-secret\" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion certman.managed.openshift.io/v1alpha1 Kind CertificateRequest: no matches for kind \"CertificateRequest\" in version \"certman.managed.openshift.io/v1alpha1\"" syncsetinstance=uhc-staging-18bj1mflc7uemp0lvler2ggv5ln9bu9t/haowang-test-haowang-test-clusteringress-syncset

this seems because the target cluster didn't have teh CertificateRequest installed


dgoodwin:
I think your owner ref workaround looks good, will close this one for now.

PR #553, opened 2019-09-24
ignore ownerreference when using secretReferences in syncset

wanghaoran1988:


dgoodwin:
This looks correct to me.

/lgtm

/test unit

dgoodwin:
Ok I have no idea why the unit tests won't pass here. I have manually verified they work off this branch.

/override ci/prow/unit

PR #558, opened 2019-09-25
Remove ApplyOnce PatchApplyMode.

abutcher:
https://jira.coreos.com/browse/CO-499

### DIFF
--- a/config/crds/hive_v1_selectorsyncset.yaml
+++ b/config/crds/hive_v1_selectorsyncset.yaml
@@ -42,13 +42,6 @@ spec:
                     description: APIVersion is the Group and Version of the object
                       to be patched.
                     type: string
-                  applyMode:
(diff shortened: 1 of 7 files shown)


wanghaoran1988:
This is already being used by SRE team, I am still not clear why we need remove this, if we have to remove this, do we have a replaceable option for this?

jewzaam:
Commented in JIRA, we use ApplyOnce at this time and need to test / transition off of it before this can merge.
/hold

dgoodwin:
PR now targets V1, so we are going to merge and then fight about it with SRE if it becomes apparent they must have apply once patching in v1. :)

/lgtm

jewzaam:
@dgoodwin fwiw, we removed use of `ApplyOnce`: https://github.com/openshift/managed-cluster-config/pull/141

dgoodwin:
Awesome, thanks!

PR #560, opened 2019-09-25
Prevent crash on compute MachinePool with no platform.

dgoodwin:
Implemented via validation, as well as a safety check in the
remotemachineset_controller at the point we were hitting the nil
pointer.

### DIFF
--- a/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
+++ b/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
@@ -257,6 +257,12 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec 
 		if aws.Region == "" {
 			allErrs = append(allErrs, field.Required(awsPath.Child("region"), "must specify AWS region"))
 		}
+		for i, mp := range newObject.Spec.Compute {
+			computePath := specPath.Child("compute").Index(i)
+			if mp.Platform.AWS == nil {
+				allErrs = append(allErrs, field.Required(computePath.Child("aws"), "must specify platform for compute machine sets"))
+			}
+		}
 	}
 	if newObject.Spec.Platform.Azure != nil {
 		numberOfPlatforms++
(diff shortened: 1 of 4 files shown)


pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -285,6 +297,15 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec
 		if gcp.Region == "" {
 			allErrs = append(allErrs, field.Required(gcpPath.Child("region"), "must specify GCP region"))
 		}
+		// TODO: Uncomment once ready for GCP
```
staebler: Should uncomment now. We don't need to wait for GCP for this validation to be here.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -257,6 +257,12 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec
 		if aws.Region == "" {
 			allErrs = append(allErrs, field.Required(awsPath.Child("region"), "must specify AWS region"))
 		}
+		for i, mp := range newObject.Spec.Compute {
+			computePath := specPath.Child("compute").Index(i)
+			if mp.Platform.AWS == nil {
```
staebler: Why is this required? The user should be able to omit it and use the defaults.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
staebler: Oh, duh. The fields are not in the machine pool platform, yet.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
dgoodwin: Looking into this now, kinda looks like default machine pool platform may not be used.

staebler: The default machine pool is something that is applied by the installer. The order of precedence for the value of a field in the machine pool platform is as follows.
1) The value set in the machine pool platform for a specific machine pool.
2) The value set in the default machine pool platform for the cluster.
3) The default value for the field.

staebler: My concern is that the remotemachineset controller is not using the same algorithm for determining what values it expects on the resultant machine set.

PR #562, opened 2019-09-25
add missing conversion for non-aws machine pool platforms

staebler:
The azure and gcp platform-specific machine pool information was not being converted into the install config.


staebler:
This requires #556 to add the GCP field to the machine pool platform.

/hold

staebler:
#556 merged.

/hold cancel

PR #567, opened 2019-09-26
e2e tests for azure and gcp

staebler:
When running e2e tests, create a cluster on azure and gcp as well as aws.

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -30,7 +30,7 @@ spec:
         name: manager
         resources:
           requests:
-            cpu: 500m
+            cpu: 50m
             memory: 512Mi
         command:
           - /opt/services/manager
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -53,44 +53,27 @@ make deploy DEPLOY_IMAGE="${HIVE_IMAGE}"
 CLOUD_CREDS_DIR="${CLOUD_CREDS_DIR:-/tmp/cluster}"
 
 # Create a new cluster deployment
-export BASE_DOMAIN="${BASE_DOMAIN:-hive-ci.openshift.com}"
+export CLOUD="${CLOUD:-aws}"
 export CLUSTER_NAME="${CLUSTER_NAME:-hive-$(uuidgen | tr '[:upper:]' '[:lower:]')}"
-export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID:-$(cat ${CLOUD_CREDS_DIR}/.awscred | awk '/aws_access_key_id/ { print $3; exit; }')}"
-export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY:-$(cat ${CLOUD_CREDS_DIR}/.awscred | awk '/aws_secret_access_key/ { print $3; exit; }')}"
 export ARTIFACT_DIR="${ARTIFACT_DIR:-/tmp}"
 export SSH_PUBLIC_KEY_FILE="${SSH_PUBLIC_KEY_FILE:-${CLOUD_CREDS_DIR}/ssh-publickey}"
 export PULL_SECRET_FILE="${PULL_SECRET_FILE:-${CLOUD_CREDS_DIR}/pull-secret}"
 
 function teardown() {
-	INSTALL_JOB_NAME=$(oc get job -l "hive.openshift.io/cluster-deployment-name=${CLUSTER_NAME},hive.openshift.io/install=true" -o jsonpath='{.items[0].metadata.name}')
-	oc logs -c hive job/${INSTALL_JOB_NAME} &> "${ARTIFACT_DIR}/hive_install_job.log" || true
-	oc get clusterdeployment -A -o yaml &> "${ARTIFACT_DIR}/hive_clusterdeployment.yaml" || true
-	oc get clusterimageset -o yaml &> "${ARTIFACT_DIR}/hive_clusterimagesets.yaml" || true
-	oc get job ${INSTALL_JOB_NAME} -o yaml &> "${ARTIFACT_DIR}/hive_install_job.yaml" || true
-	echo "************* INSTALL JOB LOG *************"
-	if oc get clusterprovision -l "hive.openshift.io/cluster-deployment-name=${CLUSTER_NAME}" -o jsonpath='{.items[0].spec.installLog}' &> "${ARTIFACT_DIR}/hive_install_console.log"; then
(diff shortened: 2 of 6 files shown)


hack/e2e-test.sh
```diff
@@ -53,64 +53,101 @@ make deploy DEPLOY_IMAGE="${HIVE_IMAGE}"
...
-go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" create-cluster "${CLUSTER_NAME}" \
-	--ssh-public-key-file="${SSH_PUBLIC_KEY_FILE}" \
-	--pull-secret-file="${PULL_SECRET_FILE}" \
-	--base-domain="${BASE_DOMAIN}" \
-	--release-image="${RELEASE_IMAGE}" \
-	--install-once=true \
-	--uninstall-once=true
+export SRC_ROOT=$(git rev-parse --show-toplevel)
+
+for PLATFORM in "${PLATFORMS[@]}"
```
dgoodwin: I'm struggling with the thought of one e2e job that covers all platforms. On one hand this is definitely a quicker and easier path to it, and we just install Hive once and then use it to provision all three, which has some benefits in coverage.

It may however be very difficult to get PRs to merge when we need all three to pass and can't re-run specific clouds when they fail.

I'd probably lean towards the approach you've taken though and we can keep an eye on it.

hack/e2e-test.sh
```diff
@@ -53,64 +53,101 @@ make deploy DEPLOY_IMAGE="${HIVE_IMAGE}"
...
+	CLUSTER_NAME=${BASE_CLUSTER_NAME}-${PLATFORM}
+	case "$PLATFORM" in
+	"aws")
+		EXTRA_CREATE_CLUSTER_ARGS=" --base-domain=${AWS_BASE_DOMAIN} --creds-file=${CLOUD_CREDS_DIR}/.awscred"
+		;;
+	"azure")
+		EXTRA_CREATE_CLUSTER_ARGS=" --base-domain=${AZURE_BASE_DOMAIN} --creds-file=${CLOUD_CREDS_DIR}/osServicePrincipal.json"
+		;;
+	"gcp")
+		EXTRA_CREATE_CLUSTER_ARGS=" --base-domain=${GCP_BASE_DOMAIN} --creds-file=${CLOUD_CREDS_DIR}/gce.json --gcp-project-id=openshift-gce-devel-ci"
```
dgoodwin: What's getting used for GCE and Azure credentials when we run this? Something available to all e2e jobs?

staebler: I was hoping that the credentials would be there from these variables.
https://github.com/openshift/release/blob/dfd2beadbd400ce3ac680b2e3a430f58bcc5e63d/ci-operator/templates/openshift/installer/cluster-launch-installer-src.yaml#L157-L162
But it appears that the credentials are only created for the platform that the e2e tests are being run on. I need to dig into this more to see if and how we can get all of the credentials.

staebler: In light of the credentials problem, I think I will abandon the multiple-platforms-in-one-e2e-test-suite approach in favor of running multiple e2e test suites.

staebler: It is unfortunate that this will necessitate creating twice as many clusters, though.

staebler:
The e2e-azure and e2e-gcp tests won't pass until https://github.com/openshift/release/pull/5215 merges.

staebler:
**e2e-azure**
The hive-controllers pod cannot be scheduled on the masters due to there not being enough cpu.
```
"message": "0/6 nodes are available: 3 Insufficient cpu, 3 node(s) had taints that the pod didn't tolerate.",
"reason": "Unschedulable",
"status": "False",
"type": "PodScheduled"
```

**e2e-gcp**
TestManageMachineSets test is failing.
```
 --- FAIL: TestManageMachineSets (1208.53s)
    infra_test.go:67: Waiting for 3 infra machines to be created
    infra_test.go:80: timed out waiting for machines to be created
    infra_test.go:83: Waiting for nodes to be created
    infra_test.go:128: timed out waiting for nodes to be created
FAIL
FAIL	github.com/openshift/hive/test/e2e/postinstall/machinesets	1208.553s 
```

dgoodwin:
We also don't do remote machineset management on gcp or azure yet.

staebler:
> We also don't do remote machineset management on gcp or azure yet.

Good point. I changed the test to do nothing unless the cluster is installed on AWS.

staebler:
The azure and gcp e2e tests are both passing now.

config/manager/deployment.yaml
```diff
@@ -30,7 +30,7 @@ spec:
         name: manager
         resources:
           requests:
-            cpu: 500m
+            cpu: 50m
```
staebler: This change from 500m to 50m is so that Hive can be installed on Azure.

hack/e2e-test.sh
```diff
@@ -53,44 +53,27 @@ make deploy DEPLOY_IMAGE="${HIVE_IMAGE}"
...
+export CLOUD="${CLOUD:-aws}"
 export CLUSTER_NAME="${CLUSTER_NAME:-hive-$(uuidgen | tr '[:upper:]' '[:lower:]')}"
-export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID:-$(cat ${CLOUD_CREDS_DIR}/.awscred | awk '/aws_access_key_id/ { print $3; exit; }')}"
-export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY:-$(cat ${CLOUD_CREDS_DIR}/.awscred | awk '/aws_secret_access_key/ { print $3; exit; }')}"
 export ARTIFACT_DIR="${ARTIFACT_DIR:-/tmp}"
 export SSH_PUBLIC_KEY_FILE="${SSH_PUBLIC_KEY_FILE:-${CLOUD_CREDS_DIR}/ssh-publickey}"
 export PULL_SECRET_FILE="${PULL_SECRET_FILE:-${CLOUD_CREDS_DIR}/pull-secret}"
 
 function teardown() {
-	INSTALL_JOB_NAME=$(oc get job -l "hive.openshift.io/cluster-deployment-name=${CLUSTER_NAME},hive.openshift.io/install=true" -o jsonpath='{.items[0].metadata.name}')
```
staebler: The log capturing was moved out of the teardown so that any errors in the log capturing do not prevent the script from deleting the clusterdeployment. For some reason that I could not quickly figure out, when doing an `oc get -o jsonpath` that does not find any items, the script errors out instead of having the error contained to the if statement.

dgoodwin:
/lgtm

Huge win, thank you.

PR #568, opened 2019-09-27
Add ability to stop sync to target clusters

wanghaoran1988:


docs/annotations.md
```diff
@@ -0,0 +1,8 @@
+# Annotations
+
+This document provides a list of annotations used by hive CRDs
+
+
+| Annotation| Description | 
+| ---------- | ----------- |
+| hive.openshift.io/syncset-pause | When the value is "true", hive will stop syncing everything to target cluster including resources defined in `syncset` object, and remote machineset.  |
```
LalatenduMohanty: :+1:

docs/annotations.md
```diff
@@ -0,0 +1,8 @@
+# Annotations
+
+This document provides a list of annotations used by hive CRDs
```
LalatenduMohanty: s/hive/Hive/

pkg/constants/constants.go
```diff
@@ -39,6 +39,9 @@ const (
 
 	// CheckpointName is the name of the object in each namespace in which the namespace's backup information is stored.
 	CheckpointName = "hive"
+
+	// AnnotationSyncsetPause is a annotation used by clusterDeployment, if it's true, then we will disable syncing to a specific cluster
+	AnnotationSyncsetPause = "hive.openshift.io/syncset-pause"
```
LalatenduMohanty: s/AnnotationSyncsetPause/SyncsetPauseAnnotation/ . IMO `SyncsetPauseAnnotation` is a better name.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -128,6 +129,10 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
 		log.WithError(err).Error("error looking up cluster deployment")
 		return reconcile.Result{}, err
 	}
+	if cd.Annotations != nil && cd.Annotations[constants.AnnotationSyncsetPause] == "true" {
+		log.Debug("syncing to cluster is disabled")
```
LalatenduMohanty: I would prefer this to be a info message as this information should be available in the logs easily as we do not want users to forget about the annotation and then wondering why syncsets are not working.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -128,6 +129,10 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
 		log.WithError(err).Error("error looking up cluster deployment")
 		return reconcile.Result{}, err
 	}
+	if cd.Annotations != nil && cd.Annotations[constants.AnnotationSyncsetPause] == "true" {
```
LalatenduMohanty: Do we need to both `cd.Annotations != nil` and `cd.Annotations[constants.AnnotationSyncsetPause]`?

pkg/controller/remotemachineset/remotemachineset_controller.go
dgoodwin: Lets actually go Warn for this, it's a one off customization that should not be done and forgotten.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -198,6 +199,11 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, err
 	}
 
+	if cd.Annotations != nil && cd.Annotations[constants.AnnotationSyncsetPause] == "true" {
+		ssiLog.Debug("syncing to cluster is disabled")
```
dgoodwin: Warn here as well, I'd throw in "due to %s annotation" as well in both messages.

pkg/controller/remotemachineset/remotemachineset_controller.go
LalatenduMohanty: Also lets mention in the message that `SyncsetPauseAnnotation is present, hence syncing to the cluster deployment is disabled`

pkg/controller/remotemachineset/remotemachineset_controller.go
staebler: The check for `nil` is not necessary.

staebler: You probably want this to be the following for more flexibility.
```
if p, err := strconv.ParseBool(cd.Annotations; p && err != nil {
``

**Comment by wanghaoran1988** [comment]
Comments addressed, PTAL

File: pkg/controller/remotemachineset/remotemachineset_controller.go [hunk 7]
```diff
@@ -128,6 +129,10 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
 		log.WithError(err).Error("error looking up cluster deployment")
 		return reconcile.Result{}, err
 	}
+	if cd.Annotations[constants.SyncsetPauseAnnotation] == "true" {
+		log.Warn(constants.SyncsetPauseAnnotation, " is present, hence syncing to cluster is disabled")
```
dgoodwin: Usually would see this as log.Warnf("%s is present, hence ...", constants.SyncsetPauseAnnotation) but no idea if either form is better or why.

dgoodwin: Just a comment, not a necessary change.

pkg/controller/remotemachineset/remotemachineset_controller.go
dgoodwin: @wanghaoran1988 This comment was probably missed, the above format would be a little more flexible and let you be less strict on the format of "true" in the annotation.

PR #572, opened 2019-09-30
Update docs on using manageDNS with missing steps.

dgoodwin:


docs/using-hive.md
```diff
@@ -188,21 +188,41 @@ oc get nodes
...
 
 ```
-apiVersion: hive.openshift.io/v1alpha1
-kind: HiveConfig
-metadata:
-  name: hive
-spec:
-  managedDomains:
-  - hive1.example.com
+$ bin/hiveutil create-cluster --base-domain=devcluster.hive.example.com devcluster --manage-dns
```
staebler: Can you have this command match up with the names used in the preceding paragraph? I am having a hard time following along with how the base domain used by the installed relates to the domain of the resultant cluster.

docs/using-hive.md
```diff
@@ -188,21 +188,41 @@ oc get nodes
...
+     spec:
+       managedDomains:
+       - hive1.example.com
+       externalDNS:
+         aws:
+           credentials:
+             name: route53-aws-creds
+     ```
+
+You can now request clusters with clusterDeployment.spec.manageDNS=true and clusterDeployment.spec.baseDomain=mycluster.hive1.example.com. Hive will create a mycluster.hive1.example.com hosted zone and link it with your root hosted zone. Per-cluster DNS entries ("apps", "api", etc) will be created in the new mycluster.hive1.example.com hosted zone.
```
staebler: What does it mean to "link" one hosted zone to another?

staebler: What are "per-cluster DNS entries"? This implies to me that in `mycluster.hive1.example.com` there could be multiple "apps" entries, for example, one for each cluster.

staebler:
Sorry about this. But I would like a couple of my questions answered first before this is merged. I'll take ownership of providing a speedy lgtm afterwards.

/lgtm cancel

docs/using-hive.md
joelddiaz: yes, this should read --base-domain=devcluster.hive1.example.com

joelddiaz: perhaps this would read better if it was --base-domain=devdomain.hive1.example.com. this means that hive is responsible for making devdomain.hive1.example.com NS entry in the hive1.example.com Route53 zone that points to the hive-created devdomain.hive1.example.com Route53 zone.

docs/using-hive.md
joelddiaz: see above description of hive creating a new route53 zone and then hive making the necessary NS entries in the managed domain

joelddiaz: again, changing mycluster.hive1.example.com to mydomain.hive1.example.com would perhaps make it clearer that hive is potentially creating a new mydomain.hive1.example.com Route53 zone, and then making the needed NS entry in the hive1.example.com Route53 zone so that DNS resolution works.

dgoodwin: PR updated to try to clarify the steps hive will perform, and the linking.

docs/using-hive.md
```diff
@@ -188,19 +188,47 @@ oc get nodes
...
+$ bin/hiveutil create-cluster --base-domain=mydomain.hive.example.com mycluster --manage-dns
 ```
 
-You can now request clusters with clusterDeployment.spec.manageDNS=true and clusterDeployment.spec.baseDomain=mycluster.hive1.example.com. Hive will create the mycluster.hive1.example.com DNS zone, and the OpenShift installer will create DNS entries such as api.mycluser.hive1.example.com.
+Hive will then:
+
+  1. Create a mydomain.hive.example.com Route53 hosted zone.
+  1. Create NS records in the hive.example.com to forward DNS to the new mydomain.hive.example.com hosted zone.
+  1. Wait for the SOA record for the new domain to be resolvable, indicating that DNS is functioning.
+  1. Launch the install, which will create DNS entries for the new cluster ("\*.apps.mydomain.hive.example.com", "api.mydomain.hive.example.com", etc) in the new mydomain.hive.example.com hosted zone.
```
staebler: Here is the part that I do not understand. If the base domain passed to the installed is `mydomain.hive.example.com` and the name of the cluster is mycluster, won't the install create `*.apps.mycluster.mydomain.hive.example.com` et al?

dgoodwin: Not that I'm aware of, need someone to verify, I don't have an example handy.

docs/using-hive.md
```diff
@@ -188,19 +188,47 @@ oc get nodes
...
+     spec:
+       managedDomains:
+       - hive.example.com
+       externalDNS:
+         aws:
+           credentials:
+             name: route53-aws-creds
+     ```
+
+You can now request clusters with manageDNS enabled and a basedomain of mydomain.hive.example.com.
```
LalatenduMohanty: s/You can now request/You can now create/

docs/using-hive.md
```diff
@@ -188,19 +188,47 @@ oc get nodes
 
 Hive can optionally create delegated Route53 DNS zones for each cluster.
 
-To use this functionality you must first define the base domains you want to manage DNS for in your HiveConfig.
+To use this feature:
```
LalatenduMohanty: Should we create a sub topic for AWS as this very specific to it. We are going to support other cloud providers and the text will be different for them.

docs/using-hive.md
dgoodwin: Updated.

docs/using-hive.md
dgoodwin: As soon as we have another provider to document yes. Probably a separate markdown file at that point.

dgoodwin: Added a note.

docs/using-hive.md
dgoodwin: You are correct, if should be mycluster.mydomain. Updating.

staebler: I don't see how the cluster domain for a managed-DNS cluster would be any different than the cluster domain for a cluster that is not using managed DNS. In both cases, the cluster domain is CLUSTER_NAME.BASE_DOMAIN. So, in your example, the cluster domain would be `mycluster.mydomain.hive.example.com`.

joelddiaz: the difference with managed DNS enabled is that you can have subdomains created/managed by hive. given that we have hive.example.com managed, then hive will be able to create dev.hive.example.com, make the NS entry in hive.example.com to point to the new dev.hive.example.com, and then run the installer to create newcluster.dev.hive.example.com entries in dev.hive.example.com

PR #573, opened 2019-09-30
Add support for install retry cleanup of A records in managed DNSZones.

dgoodwin:
These are leaking due on occasion due to a bug in the installer leading
to an install that cannot complete.

Add code to deprovision A records after we run the installer's
deprovision code. A records are the only ones we're seeing an issue
with, others can be added in future if necessary.

This code is not linked to the standard deprovision for a cluster
deployment deletion. We are confident this is safe during an install,
we're not sure if it is during a deprovision.

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -268,6 +268,9 @@ func (c *awsClient) ChangeResourceRecordSets(input *route53.ChangeResourceRecord
 // For authentication the underlying clients will use either the cluster AWS credentials
 // secret if defined (i.e. in the root cluster),
 // otherwise the IAM profile of the master where the actuator will run. (target clusters)
+//
+// Pass a nil client, and empty secret name and namespace to load credentials from the standard
+// AWS environment variables.
 func NewClient(kubeClient client.Client, secretName, namespace, region string) (Client, error) {
 	awsConfig := &aws.Config{Region: aws.String(region)}
 
--- a/pkg/installmanager/dnscleanup.go
+++ b/pkg/installmanager/dnscleanup.go
@@ -1,0 +1,56 @@
+package installmanager
+
+import (
+	awsclient "github.com/openshift/hive/pkg/awsclient"
+
+	log "github.com/sirupsen/logrus"
+
+	awssdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+// cleanupDNSZone queries the Route53 zone and deletes any A records found. Other record
+// types may be added in the future, but right now this is the only one we're seeing
+// leak and conflict.
+// May no longer be necessary once https://jira.coreos.com/browse/CORS-1195 is fixed.
+func cleanupDNSZone(dnsZoneID, region string, logger log.FieldLogger) error {
+	zoneLogger := logger.WithField("dnsZoneID", dnsZoneID)
+	zoneLogger.Info("cleaning up DNSZone")
+
+	awsClient, err := awsclient.NewClient(nil, "", "", region)
+	if err != nil {
+		return err
+	}
+	recordsOutput, err := awsClient.ListResourceRecordSets(
+		&route53.ListResourceRecordSetsInput{
+			HostedZoneId: awssdk.String(dnsZoneID),
+		},
+	)
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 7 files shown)


pkg/awsclient/client.go
```diff
@@ -312,3 +315,36 @@ func NewClient(kubeClient client.Client, secretName, namespace, region string) (
 		tagClient:     resourcegroupstaggingapi.New(s),
 	}, nil
 }
+
+// NewClientFromEnvVars creates our client wrapper object for the actual AWS clients we use.
+// For authentication the underlying clients will use the standard AWS credentials env
+// vars.
+func NewClientFromEnvVars(region string) (Client, error) {
```
staebler: Can this function simply call `NewClient(nil, "", "", region)`?

Note that `session.NewSession` will read the environment variables for the access key.

pkg/installmanager/dnscleanup.go
```diff
@@ -0,0 +1,55 @@
...
+							Action:            awssdk.String("DELETE"),
+							ResourceRecordSet: r,
+						},
+					},
+				},
+				HostedZoneId: awssdk.String(dnsZoneID),
+			}
+			_, err := awsClient.ChangeResourceRecordSets(request)
+			if err != nil {
+				return err
```
staebler: Should this behave like the uninstaller does and loop until there are no failures instead of bailing out when there is a failure?

pkg/installmanager/installmanager.go
```diff
@@ -457,7 +458,31 @@ func runUninstaller(cd *hivev1.ClusterDeployment, infraID string, logger log.Fie
...
+			return err
+		}
+
+		// If we're managing DNS for this cluster, lookup the DNSZone and cleanup
+		// any leftover A records that may have leaked due to
+		// https://jira.coreos.com/browse/CORS-1195.
+		if cd.Spec.ManageDNS {
+			dnsZone := &hivev1.DNSZone{}
+			dnsZoneNamespacedName := types.NamespacedName{Namespace: cd.Namespace, Name: controllerutils.DNSZoneName(cd.Name)}
+			err := dynClient.Get(context.TODO(), dnsZoneNamespacedName, dnsZone)
```
staebler: Need to add permission [here](https://github.com/openshift/hive/blob/30a1011410d191c011ebba314c60c1b2ede677b7/pkg/controller/utils/sa.go#L27) in order to get DNSZone resources.

dgoodwin: Good catch, hadn't gotten far enough to hit that yet, thanks.

pkg/installmanager/dnscleanup.go
dgoodwin: I'm not sure, I think that's effectively what this will do now, we try to cleanup after a failure, and at the start of our next attempt. A failure to delete records here will error out the job and queue it up for another provision attempt. That gets us free backoff and realistically if you're failing to delete a DNS record more than once or twice, it's likely not going to resolve itself and will need manual intervention. In practice I think it's better to let that hit the provision backoff than infinite loop and hammer the API.

dgoodwin:
PR updated with removed NewClientForEnvVars and the missing permission.

pkg/installmanager/dnscleanup.go
staebler: My biggest concern was that the backoff escalates very quickly. So if there is a failed install followed by a transient error of some kind deleting the A record, then the backoff can quickly get into the hours range.

pkg/installmanager/dnscleanup.go
```diff
@@ -0,0 +1,55 @@
...
+
+	awssdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+// CleanupDNSZone queries the Route53 zone and deletes any A records found. Other record
+// types may be added in the future, but right now this is the only one we're seeing
+// leak and conflict.
+// May no longer be necessary once https://jira.coreos.com/browse/CORS-1195 is fixed.
+func CleanupDNSZone(dnsZoneID, region string, logger log.FieldLogger) error {
```
staebler: This functions does not need to be public.

dgoodwin:
Updated.

dgoodwin:
Not sure how I managed that. Squashed my commits and rebased them off master.

PR #574, opened 2019-09-30
accept a namespace to logextractor

staebler:
To be useful in the stage and production clusters, the logextractor script needs to be able to use clusterdeployments in non-default namespaces.

### DIFF
--- a/hack/logextractor.sh
+++ b/hack/logextractor.sh
@@ -1,21 +1,60 @@
 #!/bin/sh
 #
 # This script can be used to extract the failure cluster logs tarball from the PVC created for an install pod.
+# Note that there will only be failure logs if there was a failure installing and that the failure logs will
+# only be retained for 7 days after a successful install.
 #
-# Usage: ./logextractor.sh [CLUSTER_DEPLOYMENT_NAME] [DEST_DIR]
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
Used the PR to run the script and got following error. Not sure if it is an issue with the cluster (as I am doing 600 cluster deployment scale testing on it)
```
$ hack/logextractor.sh default/lalatendum-hive-1 logs_extractor_dir
Found PVC for cluster deployment: lalatendum-hive-1-install-logs
pod/hive-logextractor created
waiting for hive-logextractor pod to be ready
error: timed out waiting for the condition
copying logs to logs_extractor_dir
Error from server (BadRequest): pod hive-logextractor does not have a host assigned
deleting logextractor pod
pod "hive-logextractor" deleted
```

hack/logextractor.sh
```diff
@@ -2,20 +2,52 @@
 #
 # This script can be used to extract the failure cluster logs tarball from the PVC created for an install pod.
 #
```
LalatenduMohanty: Can we add a sentence here to mention that we need a failed install or install in progress for this script work?

staebler: Is that not implied by "extract the failure cluster logs tarball from the PVC created for an install pod"?

LalatenduMohanty: Nope, at least to me. +1 for explicitly mentioning it

dgoodwin:
cc @openshift/openshift-team-hive

PR #575, opened 2019-10-01
remove logging of install-config.yaml

staebler:
The Azure provisioning PR added logging of the install-config.yaml, which was supposed to have been removed before merging.


dgoodwin:
Ugh sorry about that.

/lgtm

PR #576, opened 2019-10-02
installmanager: gather logs reporting success with empty private ssh key file

staebler:
When the private ssh key file is empty, the installmanager is reporting that gathering the logs from the bootstrap node was successful even though the logs where not gathered. The `gatherBootstrapNodeLogs` function should be returning a non-nil error in this scenario.


dgoodwin:
Hah, well that's a slippery one.

/lgtm

PR #577, opened 2019-10-02
installmanager: wait-for install-complete after failed install

staebler:
If the `openshift-install create cluster` command fails after the bootstrapping has completed, then run `openshift-install wait-for install-complete` once to wait an additional 30 minutes for the cluster to finsih installing. It is not uncommon for a cluster install to time out when the cluster is at 99+% complete. This will give those clusters a chance to finish before tearing them down and
starting over.


staebler:
> Looks good otherwise, just a request for a log stmt.

I added code to emit a warn log entry when the `create cluster` command fails after bootstrapping has completed.

PR #578, opened 2019-10-02
remove unnecessary warning about dnszone not getting deleted

staebler:
The clusterDeployment controller emits a warning log when a dnszone for a clusterdeployment does not getting deleted when the clusterdeployment is deleted. However, this is expected behavior as the owned objects do not get deleted until the finalizers (other than the gc finalizer) are removed
from the owner.

In addition to no longer emitting the warning, the controller also requeues the clusterdeployment while waiting for the dnszone to be removed from storage. The controller will not get a watch event when the dnszone is removed from storage, so it neeeds to manually requeue.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1082,31 +1082,29 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.Clus
 	}
 	dnsZone := &hivev1.DNSZone{}
 	dnsZoneNamespacedName := types.NamespacedName{Namespace: cd.Namespace, Name: controllerutils.DNSZoneName(cd.Name)}
-	err := r.Get(context.TODO(), dnsZoneNamespacedName, dnsZone)
-	if err != nil && !apierrors.IsNotFound(err) {
-		cdLog.WithError(err).Error("error looking up managed dnszone")
-		return &reconcile.Result{}, err
-	}
-	if apierrors.IsNotFound(err) || !dnsZone.DeletionTimestamp.IsZero() {
-		cdLog.Debug("dnszone has been deleted or is getting deleted")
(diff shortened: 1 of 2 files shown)


staebler:
I would also like to mention that I do not see a reason to have the clusterDeployment controller manually delete the DNSZone. After the Hive finalizer is removed from the clusterDeployment, the DNSZone will get deleted automatically by the garbage collector. The dnszone controller does not need the clusterdeployment to still exist in order to do its clean up of the DNSZone.

dgoodwin:
This was added in d6bef9e223f8732bbf4674fdd18c29296daafae7 explicitly because the gc sometimes does not do what you describe, and clusters were getting stuck.

dgoodwin:
Conflicts :(

staebler:
> This was added in [d6bef9e](https://github.com/openshift/hive/commit/d6bef9e223f8732bbf4674fdd18c29296daafae7) explicitly because the gc sometimes does not do what you describe, and clusters were getting stuck.

Is this the same issue that we see sometimes with pods not getting garbage collected when the job is deleted? I wonder if this is any better in more recent versions of kube?

dgoodwin:
Still running on 3.11 in prod but we can always search for the log message to see if this is proc'ing anywhere.

staebler:
> Still running on 3.11 in prod but we can always search for the log message to see if this is proc'ing anywhere.

I am not sure which log message you are referring to.

dgoodwin:
The one this PR removed :) 	cdLog.Warn("managed dnszone did not get a deletionTimestamp when parent cluster deployment was deleted, deleting manually")

staebler:
> The one this PR removed :) cdLog.Warn("managed dnszone did not get a deletionTimestamp when parent cluster deployment was deleted, deleting manually")

That warning was being printed for every deprovision. That is why I deleted the message. It was not indicative of any warning condition.

staebler:
I did check in stage that it was indeed printed for every deprovision that had happened since the last restart of the Hive controllers.

PR #580, opened 2019-10-07
installmanager: fix check for whether bootstrap is complete

staebler:
The isBootstrapComplete check in InstallManager is returning true when there was an error validating that the bootstrap is complete. This is backwards: It should be returning false in this case.

### DIFF
--- a/pkg/installmanager/installmanager_test.go
+++ b/pkg/installmanager/installmanager_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


pkg/installmanager/installmanager_test.go
```diff
@@ -693,3 +694,37 @@ func TestInstallManagerSSH(t *testing.T) {
...
+		},
+		{
+			name:             "not complete",
+			errCode:          1,
+			expectedComplete: false,
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			dir, err := ioutil.TempDir("", "TestIsBootstrapComplete")
```
dgoodwin: Looks good just need to defer a cleanup on this dir.

staebler: Ah, yes. Good catch.

PR #571, opened 2019-09-27
Modifying admission validator names to make the plural names of CRDs to work

LalatenduMohanty:
The names need to be different else kubectl get <resource name>
will get rejected by the admission controller

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #582, opened 2019-10-08
webhook for clusterprovision validator

staebler:
There is a validator for clusterprovisions, but it is not hooked up with an webhook. This adds that webhook.


staebler:
This PR depends on #571 so that #571 does not need to make additional changes.

/hold

LalatenduMohanty:
I was also wondering why I do not see a validator for clusterprovisions. You need to run `make generate` to fix the unit tests.

PR #583, opened 2019-10-08
controller: no error message on creation when already exists

staebler:
There are numerous places where a controller tries to get a resource by a known name and, if the resource does not exist, then create said resource. This operation has a race condition where the controller may attempt to reconcile the controlling resource again prior to the local cache being populated with the resource recently created by the controller. When this happens, the controller spits out an error log entry, even though this is not indicative of a problem.

These changes provide a helper GetOrCreate function that can be used by a controller to attempt to get a resource or create it if it is not found. All of the places where a controller does a get/create are changed so that the log entry emitted on the conflicted create is an info-level entry instead of an error-level entry.

See https://jira.coreos.com/browse/CO-582.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -847,7 +847,7 @@ func (r *ReconcileClusterDeployment) createPVC(cd *hivev1.ClusterDeployment, cdL
 	}
 	err := r.Create(context.TODO(), pvc)
 	if err != nil {
-		cdLog.WithError(err).Error("error creating pvc")
+		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error creating pvc")
 	}
 	return err
 }
@@ -942,7 +942,7 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 
 		err = r.Create(context.TODO(), job)
 		if err != nil {
-			jobLog.WithError(err).Error("error creating job")
+			jobLog.WithError(err).Log(controllerutils.LogLevel(err), "error creating job")
 		} else {
 			// kickstartDuration calculates the delay between creation of cd and start of imageset job
 			kickstartDuration := time.Since(cd.CreationTimestamp.Time)
@@ -1184,7 +1184,7 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 			// requeue the clusterdeployment immediately to process the status of the deprovision request
 			return reconcile.Result{Requeue: true}, nil
 		case err != nil:
-			cdLog.WithError(err).Error("error creating deprovision request")
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error creating deprovision request")
 			// Check if namespace is terminated, if so we can give up, remove the finalizer, and let
 			// the cluster go away.
 			ns := &corev1.Namespace{}
@@ -1357,7 +1357,7 @@ func (r *ReconcileClusterDeployment) createManagedDNSZone(cd *hivev1.ClusterDepl
 
 	err := r.Create(context.TODO(), dnsZone)
 	if err != nil {
-		logger.WithError(err).Error("cannot create DNS zone")
+		logger.WithError(err).Log(controllerutils.LogLevel(err), "cannot create DNS zone")
 		return err
 	}
 	logger.Info("dns zone created")
--- a/pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
+++ b/pkg/controller/clusterdeprovisionrequest/clusterdeprovisionrequest_controller.go
@@ -167,7 +167,7 @@ func (r *ReconcileClusterDeprovisionRequest) Reconcile(request reconcile.Request
 		rLog.Debug("uninstall job does not exist, creating it")
 		err = r.Create(context.TODO(), uninstallJob)
 		if err != nil {
-			rLog.WithError(err).Errorf("error creating uninstall job")
+			rLog.WithError(err).Log(controllerutils.LogLevel(err), "error creating uninstall job")
 			return reconcile.Result{}, err
 		}
 		return reconcile.Result{}, nil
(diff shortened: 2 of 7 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1664,58 +1646,47 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
...
-		err = r.Create(context.TODO(), newPullSecretObj)
-		if err != nil {
-			return false, errors.Wrap(err, "error creating new pull secret object")
-		}
-		cdLog.WithField("secretName", mergedSecretName).Info("Created the merged pull secret object successfully")
+	existingPullSecret, ok := pullSecretSecret.Data[corev1.DockerConfigJsonKey]
+	if !ok {
+		return false, fmt.Errorf("Pull secret %s did not contain key %s", pullSecretSecret.Name, corev1.DockerConfigJsonKey)
+	}
+	if string(existingPullSecret) == pullSecret {
```
staebler: This was changed to compare the full pull secrets instead of hashes of the pull secrets. Comparing the hashes is only helpful if we already have at least one of the hashes. In this case, we only have the full strings of both and are creating both hashes. It is more efficient, then, to simply compare the full strings.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1664,58 +1646,47 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
...
+		client.ObjectKey{Namespace: cd.Namespace, Name: constants.GetMergedPullSecretName(cd)},
+		pullSecretSecret,
+		func() error {
+			pullSecretSecret.Type = corev1.SecretTypeDockerConfigJson
+			pullSecretSecret.StringData = map[string]string{corev1.DockerConfigJsonKey: pullSecret}
+			return errors.Wrap(
+				controllerutil.SetControllerReference(cd, pullSecretSecret, r.scheme),
+				"error setting controller reference on new merged pull secret",
+			)
+		},
```
LalatenduMohanty: IMHO I find the code relatively less readable as compared with the old code we had. I understand the motive behind it, but it is relatively harder to read.

staebler: Can you elaborate on why you find if less readable? Is it the `GetOrCreate` function in general? Or is there something about the flow of the logic in the function that is more difficult to follow?

I am inclined to keep the change for two reasons. (1) I would like to stay consistent across the code base with using `GetOrCreate`. (2) It is more concise, eliminating 28 lines of code.

LalatenduMohanty: When I start reading the function `updatePullSecretInfo` in the begining we have `GetOrCreate` which does not align with the function name. IMO the function name should be like `createOrUpdateMergePullSecret` instead of `updatePullSecretInfo`.

Second, it is not easy to understand the purpose of the anonymous function.  Also the `GetOrCreate` function too hard to understand.

staebler: > When I start reading the function `updatePullSecretInfo` in the begining we have `GetOrCreate` which does not align with the function name. IMO the function name should be like `createOrUpdateMergePullSecret` instead of `updatePullSecretInfo`.

This PR does not change the behavior of the `updatePullSecretInfo` function. If the name is not appropriate, then that is a problem that is outside of the scope of this PR.

staebler: > Second, it is not easy to understand the purpose of the anonymous function. Also the `GetOrCreate` function too hard to understand.

It sounds to me like your main gripe is with the `GetOrCreate` function in general, not just in this particular use of the function. My main goal with the function is to consolidate some of the logic and to provide a consistent way in which the common pattern of getting a resource or creating it if it does not exist can be written.

To be clear, given the choice between the two following alternatives, your choice would be the one that does not use `GetOrCreate`, right?

```
pullSecretSecret := &corev1.Secret{}
switch created, err := controllerutils.GetOrCreate(
	context.Background(),
	r.Client,
	client.ObjectKey{Namespace: cd.Namespace, Name: constants.GetMergedPullSecretName(cd)},
	pullSecretSecret,
	func() error {
		pullSecretSecret.Type = corev1.SecretTypeDockerConfigJson
		pullSecretSecret.StringData = map[string]string{corev1.DockerConfigJsonKey: pullSecret}
		return errors.Wrap(
			controllerutil.SetControllerReference(cd, pullSecretSecret, r.scheme),
			"error setting controller reference on new merged pull secret",
		)
	},
); {
case err != nil:
	return false, errors.Wrap(err, "error getting or creating merged pull secret secret")
case created:
	cdLog.WithField("secretName", pullSecretSecret.Name).Info("Created the merged pull secret object successfully")
	return true, nil
}
```
```
pullSecret := &corev1.Secret{}
switch err := r.Get(
    context.Background(),
    client.ObjectKey{Namespace: cd.Namespace, Name: constants.GetMergedPullSecretName(cd)},
    pullSecret,
); {
case apierrors.IsAlreadyExists(err):
    pullSecret.Namespace = cd.Namespace
    pullSecret.Name = constants.GetMergedPullSecretName(cd)
    pullSecretSecret.Type = corev1.SecretTypeDockerConfigJson
    pullSecretSecret.StringData = map[string]string{corev1.DockerConfigJsonKey: pullSecret}
    if err := r.Create(context.Background(), pullSecret); err != nil {
        return false, errors.Wrap(err, "error creating merged pull secret secret")
    } else {
        cdLog.WithField("secretName", pullSecretSecret.Name).Info("Created the merged pull secret object successfully")
        return true, nil
    }
case err != nil:
    return false, errors.Wrap(err, "error getting merged pull secret secret")
}
```

staebler:
/hold

I rebased this on top of #594. There will be conflicts between this PR and that PR, and that PR is likely to merge first.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Either is ok with me but if we're talking margins, the latter is a little more readable to me, it's a little more clear what everything is doing whereas with the former I think I'd often be going back to look at what that func was supposed to do in particular.

staebler: OK. Since the only ones expressing opinions are both in favor of the latter option, I will change to use that one.

staebler:
This PR has been trimmed down to only change log levels and not do any refactoring.

PR #585, opened 2019-10-09
delete managed DNS after deprovision

staebler:


dgoodwin:
So previous to these changes we would ensure deletion timestamp on the DNSZone before launching the deprovision job, which I guess implies that we would let the dnszone try to delete and fail until deprovision itself cleaned out the records?

This is getting through e2e because we don't use manage DNS, which is probably a big enough gap we should file a Jira and get that addressed. I will go add that.

I am still really curious why that explicit delete call is required every single time and GC isn't automatically deleting the dnszone via it's owner ref.

The change looks good, it seems logical to wait until the managed dnszone is completely gone before we remove our finalizer and I can't think of any way that's going to deadlock. 

Could you expand the commit msg before we merge though to link up with the commit that broke it and explain what happened.

staebler:
> I am still really curious why that explicit delete call is required every single time and GC isn't automatically deleting the dnszone via it's owner ref.

By design, the owned resources do not get deleted until all of the finalizers other than the GC finalizer are removed from the owner resource. In this example, the DNSZone will not get deleted until the `hive.openshift.io/deprovision` finalizer is removed from the clusterdeployment. But the `hive.openshift.io/deprovision` finalizer will not be removed until the DNSZone is deleted. If we remove the `ensureDNSManagedZoneDeleted` code, then the DNSZone will by contract get deleted after all of the non-GC finalizers are removed from the clusterdeployment.

staebler:
/close

Closed in favor of https://github.com/openshift/hive/pull/587.

PR #591, opened 2019-10-10
controller: stop cd reconcile on missing clusterimageset

staebler:
The clusterDeployment controller should not proceed with further reconcilation of a clusterdeployment if the controller finds that the clusterimageset is missing. Instead, the controller was stopping the first time when the condition was set indicating that the imageset is missing but was continuing on subsequent reconciliations since the clusterdeployment did not need to be modified in to set the condition after the first time.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -320,13 +320,6 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, r.Update(context.TODO(), cd)
 	}
 
-	imageSet, modified, err := r.getClusterImageSet(cd, cdLog)
-	if modified || err != nil {
-		return reconcile.Result{}, err
-	}
-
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -862,23 +862,23 @@ func (r *ReconcileClusterDeployment) getReleaseImage(cd *hivev1.ClusterDeploymen
...
-		return nil, false, nil
+		return nil, nil
 	}
 	imageSet := &hivev1.ClusterImageSet{}
-	err := r.Get(context.TODO(), types.NamespacedName{Name: cd.Spec.ImageSet.Name}, imageSet)
-	switch {
-	case apierrors.IsNotFound(err):
-		cdLog.WithField("clusterimageset", cd.Spec.ImageSet.Name).Warning("clusterdeployment references non-existent clusterimageset")
-		modified, err := r.setImageSetNotFoundCondition(cd, false, cdLog)
-		return nil, modified, err
```
staebler: The crux of the bug is here where `getClusterImageSet` would return `nil, false, nil` once the image set not found condition had been set. It should have been returning the IsNotFound error returned from the `Get` call.

PR #593, opened 2019-10-11
Removing outdated ValidatingWebhookConfigurations

LalatenduMohanty:
Fixes blocker issue https://jira.coreos.com/browse/CO-595

### DIFF
--- a/pkg/operator/hive/hiveadmission.go
+++ b/pkg/operator/hive/hiveadmission.go
@@ -18,6 +18,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	admregv1 "k8s.io/api/admissionregistration/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/operator/hive/hive.go
```diff
@@ -173,6 +174,30 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+		"clusterimagesets.admission.hive.openshift.io",
+		"dnszones.admission.hive.openshift.io",
+		"selectorsyncsets.admission.hive.openshift.io",
+		"syncsets.admission.hive.openshift.io",
+	}
+	for _, webhookName := range removeValidatingWebhooks {
+		webhookConfig := &admissionregistrationv1beta1.ValidatingWebhookConfiguration{}
+		err := r.Get(context.Background(), types.NamespacedName{Name: webhookName}, webhookConfig)
+		if err != nil {
+			hLog.Infof("Validating webhook does not exist: %v")
```
dgoodwin: Need to check the type of error here as the log message could be a bit inaccurate. If it doesn't exist don't log anything. Any other error should probably be logged at log.error.

pkg/operator/hive/hive.go
```diff
@@ -173,6 +174,30 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+			err = r.Delete(context.Background(), webhookConfig)
+			if err != nil {
+				hLog.WithError(err).WithField("ValidatingWebhookConfiguration", webhookConfig).Error(
+					"error deleting outdated ValidatingWebhookConfiguration")
+				return err
+			}
+			hLog.WithField("ValidatingWebhookConfiguration", webhookName).Info("deleted outdated ValidatingWebhookConfiguration")
+		}
+	}
+
```
dgoodwin: And technically all this code should be in hiveadmission.go as that is specific to that component.

Otherwise looks good!

pkg/operator/hive/hiveadmission.go
```diff
@@ -199,6 +200,30 @@ func (r *ReconcileHiveConfig) deployHiveAdmission(hLog log.FieldLogger, h *resou
...
+		"syncsets.admission.hive.openshift.io",
+	}
+	for _, webhookName := range removeValidatingWebhooks {
+		webhookConfig := &admissionregistrationv1beta1.ValidatingWebhookConfiguration{}
+		err := r.Get(context.Background(), types.NamespacedName{Name: webhookName}, webhookConfig)
+		if err != nil && !errors.IsNotFound(err) {
+			hLog.WithError(err).Error("error looking for obsolete ValidatingWebhookConfiguration")
+			return err
+		}
+		err = r.Delete(context.Background(), webhookConfig)
```
staebler: Should not try to delete if the get returned IsNotFound.

LalatenduMohanty:
@staebler @dgoodwin PTAL

PR #594, opened 2019-10-12
use info level for update conflicts

staebler:
It is not out of the ordinary for there to be conflicts when updating resources. Some resources are updated by multiple controllers running concurrently. As such, an update conflict should not be logged at the error level.


pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -159,38 +159,31 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
-		if err != nil {
-			cdLog.WithError(err).Error("error updating cluster deployment with unreachable condition (= true)")
-			return reconcile.Result{}, err
-		}
-		cdLog.Info("cluster is unreachable")
-		return reconcile.Result{}, nil
+		status = corev1.ConditionTrue
+		reason = "ErrorConnectingToCluster"
+		message = err.Error()
+		updateCheck = controllerutils.UpdateConditionIfReasonOrMessageChange
```
LalatenduMohanty: I like the code restructure but it is not related to the main fix PR is doing so should be a separate commit.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -351,7 +341,7 @@ func (r *ReconcileControlPlaneCerts) setCertsNotFoundCondition(cd *hivev1.Cluste
 		message = certsFoundMessage
 	}
 
-	cd.Status.Conditions = controllerutils.SetClusterDeploymentCondition(
+	conds, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
```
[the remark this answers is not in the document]
staebler: https://github.com/openshift/hive/pull/597

pkg/controller/unreachable/unreachable_controller.go
staebler: #598

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -384,7 +384,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 	if !controllerutils.HasFinalizer(cd, hivev1.FinalizerDeprovision) {
 		cdLog.Debugf("adding clusterdeployment finalizer")
 		if err := r.addClusterDeploymentFinalizer(cd); err != nil {
-			cdLog.WithError(err).Error("error adding finalizer")
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error adding finalizer")
```
LalatenduMohanty: @staebler can you put the example of log before and after this change. I am not sure what is the effect of `cdLog.WithError` in the logs.

LalatenduMohanty:
The PR looks good except the question I have https://github.com/openshift/hive/pull/594#discussion_r335409031.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: If the error is not a conflict error, then the log entry emitted is exactly the same.
If the error is a conflict error, then the only difference in the log entry omitted is that the level is info instead of error.

`WithError(err)` is identical to `WithField("error", err)`

PR #596, opened 2019-10-15
update install.md to include go lang pre-reqs when deploying

inderpaltiwana:
Updating the install Readme to include go lang dependency when deploying, and attempting to run the make deploy command.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -4,6 +4,7 @@
 
 * [kustomize](https://github.com/kubernetes-sigs/kustomize#kustomize)
 * [oc](https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/)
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -4,6 +4,7 @@
 
 * [kustomize](https://github.com/kubernetes-sigs/kustomize#kustomize)
 * [oc](https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/)
+* [go](https://github.com/golang/go)
```
LalatenduMohanty: We should add go version here i.e. s/go/go version 1.12/

inderpaltiwana: updated the readme go prerequisite to include go version 1.12 as suggested

PR #600, opened 2019-10-16
docs: updated some commands in syncset

gyliu513:


docs/syncset.md
```diff
@@ -127,8 +134,12 @@ The failure logs for syncset is present in Hive controller POD logs.
 
 To find the status of the syncset, check the corresponding syncsetinstance object in the cluster deployment namespace.
 
-```sh
-oc get syncsetinstance -n <namespace>
+```console
+$ oc get syncsetinstance -n <namespace>
```
LalatenduMohanty: We had used `sh` without the `$` in the command because it is easier for anyone to copy the command and paste it in any shell. Can you explain why you think `console` will be better here?

dgoodwin: Looking at the rendered page in github I can't see a lot of effective difference. I'm going to guess sh is more for bash scripts and console for console commands and thus might be more correct?

gyliu513: Here I want to make the document consistent, we should either have `$` or not have `$` for the whole page, so I was trying to make it more consistent, make sense? @LalatenduMohanty @dgoodwin

LalatenduMohanty: I would highly recommend that we remove the `$` as it will be helpful for people copying the command from the document and pasting them in a shell. Thanks for fixing it.

gyliu513: Done, thanks @LalatenduMohanty

PR #603, opened 2019-10-16
Update install log regexes configmap

LalatenduMohanty:


config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -25,3 +25,8 @@ data:
       - "waiting for Kubernetes API: context deadline exceeded"
       installFailingReason: KubeAPIWaitTimeout
       installFailingMessage: Timeout waiting for the Kubernetes API to begin responding
+    - name: MonitoringOperatorError
+      searchRegexStrings:
+      - "failed to initialize the cluster: Cluster operator monitoring is still updating"
+      installFailingReason: monitoring operator fails to become ready
```
staebler: The reason should be camel-case with no spaces.

See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

dgoodwin: We just had a race condition on this line. :)

PR #605, opened 2019-10-16
WIP: Manage dns e2e

dgoodwin:


contrib/pkg/adm/managedns/enable.go
```diff
@@ -0,0 +1,191 @@
...
+
+	var credsSecret *corev1.Secret
+
+	if o.Cloud == cloudAWS {
+		// Apply a secret for credentials to manage the root domain:
+		credsSecret, err = o.generateAWSCredentialsSecret()
+		if err != nil {
+			log.WithError(err).Fatal("error generating manageDNS credentials secret")
+		}
+		accessor, err := meta.Accessor(credsSecret)
```
staebler: Don't need to get an accessor. Can set the namespace directly with the following.
```
credsSecret.Namespace = hiveNamespace
```

contrib/pkg/adm/managedns/enable.go
```diff
@@ -0,0 +1,191 @@
...
+		if err != nil {
+			log.WithError(err).Errorf("Cannot create accessor for object of type %T", credsSecret)
+			return err
+		}
+		accessor.SetNamespace(hiveNamespace)
+		rh.ApplyRuntimeObject(credsSecret, scheme.Scheme)
+	}
+
+	hc.Spec.ExternalDNS = &hivev1.ExternalDNSConfig{
```
staebler: This should be inside the `if o.Cloud == cloudAWS` statement. It would then be nice to use `credsSecret.Name` instead of `manageDNSCredentialsSecret`.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -0,0 +1,191 @@
...
+	// create a default one once run.
+	hc := &hivev1.HiveConfig{}
+	err = dynClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hc)
+	if err != nil {
+		log.WithError(err).Fatal("error looking up HiveConfig 'hive'")
+	}
+
+	hc.Spec.ManagedDomains = args
+
+	var credsSecret *corev1.Secret
```
staebler: This declaration is unnecessary. The variable is only used inside the following if statement. There is no need for it to leak outside of that scope.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -0,0 +1,191 @@
...
+	}, nil
+}
+
+func (o *Options) getResourceHelper() (*resource.Helper, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		log.WithError(err).Error("Cannot get client config")
+		return nil, err
+	}
+	helper := resource.NewHelperFromRESTConfig(cfg, log.WithField("command", "create-cluster"))
```
staebler: Change the "command" field value.

dgoodwin:
Operator will deploy external DNS for us as soon as we make the config changes.

Thanks for the review, on a preliminary run now, will see how it goes.

staebler:
> Operator will deploy external DNS for us as soon as we make the config changes.
> 
> Thanks for the review, on a preliminary run now, will see how it goes.

Ah, makes sense. And we are actually publishing external-dns to registry.svc.ci.openshift.org/openshift/hive-v4.0:external-dns?

dgoodwin:
Bot is jammed and I don't know why, may have to open a new PR.

PR #606, opened 2019-10-16
Cleanup syncset secretreference secrets when ResourceApplyMode==Sync

abutcher:
https://jira.coreos.com/browse/CO-587

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -360,10 +360,14 @@ func (r *ReconcileSyncSetInstance) syncDeletedSyncSetInstance(ssi *hivev1.SyncSe
 
 	ssiLog.Info("deleting syncset resources on target cluster")
 	err = r.deleteSyncSetResources(ssi, dynamicClient, ssiLog)
-	if err == nil {
-		return reconcile.Result{}, r.removeSyncSetInstanceFinalizer(ssi, ssiLog)
+	if err != nil {
+		return reconcile.Result{}, err
 	}
-	return reconcile.Result{}, err
+	err = r.deleteSyncSetSecretReferences(ssi, dynamicClient, ssiLog)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, r.removeSyncSetInstanceFinalizer(ssi, ssiLog)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -360,10 +360,14 @@ func (r *ReconcileSyncSetInstance) syncDeletedSyncSetInstance(ssi *hivev1.SyncSe
...
 	ssiLog.Info("deleting syncset resources on target cluster")
 	err = r.deleteSyncSetResources(ssi, dynamicClient, ssiLog)
-	if err == nil {
-		return reconcile.Result{}, r.removeSyncSetInstanceFinalizer(ssi, ssiLog)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	err = r.deleteSyncSetSecretReferences(ssi, dynamicClient, ssiLog)
+	if err != nil {
+		return reconcile.Result{}, err
```
LalatenduMohanty: @abutcher Is it a safe assumption that secrets are the only objects that need to get attached to syncsets?

abutcher: Generic resources (objects) and secrets thus far.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -428,6 +432,36 @@ func (r *ReconcileSyncSetInstance) deleteSyncSetResources(ssi *hivev1.SyncSetIns
...
+		if err != nil {
+			switch {
+			case errors.IsNotFound(err):
+				secretLog.Debug("secret not found, nothing to do")
+			case errors.IsForbidden(err):
+				secretLog.WithError(err).Error("forbidden secret deletion, skipping")
+			default:
+				lastError = err
+				secretLog.WithError(err).Error("error deleting secret")
+				ssi.Status.SecretReferences[index].Conditions = r.setDeletionFailedSyncCondition(ssi.Status.SecretReferences[index].Conditions, fmt.Errorf("failed to delete secret: %v", err))
```
staebler: The controller never updates the status, so these conditions are lost.

staebler: It looks like the same things is done when deleting the resources.

PR #608, opened 2019-10-16
go lang in pre-reqs

inderpaltiwana:
updating the pre-reqs to include golang v1.12


inderpaltiwana:
Reattempted this after making a bit of a hash of it earlier in #596

PR #612, opened 2019-10-17
Skip minor OpenShift specific assets when running on vanilla Kube.

dgoodwin:
We deploy two ClusterRoles and ClusterRoleBindings intended to be mapped
to hive-admin and hive-reader groups in Hive clusters. Because these
groups only exist as OpenShift types, it appears using the OpenShift
variant of ClusterRole and ClusterRoleBinding is the only way to map to
them.

These two roles however are not required or useful for development, and
as such can be safely skipped.

hive-operator now checks if we're running on OpenShift by attempting to
list DeploymentConfigs. If this fails, these assets will not be
deployed and a warning is logged.

This change does not get the operator fully working on vanilla Kube,
next up will be fixed for hiveadmission certs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ VERIFY_IMPORTS_CONFIG = build/verify-imports/import-rules.yaml
 # To use docker build, specify BUILD_CMD="docker build"
 BUILD_CMD ?= imagebuilder
 
+DOCKER_CMD ?= docker
+
 # Image URL to use all building/pushing image targets
 IMG ?= hive-controller:latest
 
@@ -108,12 +110,6 @@ deploy: manifests install generate
 .PHONY: manifests
 manifests: crd
 
-# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-.PHONY: deploy-sd-dev
-deploy-sd-dev: crd
-	oc apply -f config/crds
-	kustomize build overlays/sd-dev | oc apply -f -
-
 # Generate CRD yaml from our api types:
 .PHONY: crd
 crd:
@@ -179,6 +175,12 @@ generate: $(GOPATH)/bin/mockgen
 .PHONY: docker-build
 docker-build:
 	$(BUILD_CMD) -t ${IMG} .
+
+# Build the docker image
+.PHONY: docker-dev-push
+docker-dev-push: build
+	$(DOCKER_CMD) build -t ${IMG} -f Dockerfile.dev .
+	$(DOCKER_CMD) push ${IMG}
 
 # Push the docker image
 .PHONY: docker-push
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -180,6 +176,12 @@ generate: $(GOPATH)/bin/mockgen
 docker-build:
 	$(BUILD_CMD) -t ${IMG} .
 
+# Build the docker image
+.PHONY: docker-dev-push
+docker-dev-push: build
```
staebler: Are these changes related to skipping the OpenShift-specific assets? Or are these changes for using the local image registry?

config/rbac/hive_admin_role_binding.yaml
```diff
@@ -1,3 +1,6 @@
+# NOTE: This binding uses the openshift apigroup as it is the only way to link
```
staebler: Would it be possible to use the `rbac.authorization.k8s.io` version in most cases and only the `authorization.openshift.io` version when running on OpenShift 3.11? The `rbac.authorization.k8s.io` version supports group. See https://github.com/openshift/api/blob/12f2193c359f84d4272827843d3efc2362d62344/authorization/v1/types.go#L397-L402.

Or is an OpenShift group different than a normal k8s group?

dgoodwin: I have no idea what the k8s group is, I can't find any such type in kind or openshift, I think it might be a separate concept and the two are very much different.

staebler: See https://kubernetes.io/docs/reference/access-authn-authz/authentication/

As far as I know, there are not resources that define groups. There are many parts of the system that determine which groups a user belongs to.

staebler: But the concept of a group should be the same regardless of which type of kube you are using.

dgoodwin: I don't want to risk cutting off our access in stg and prod, and I don't really want to invest the time in testing and debugging how this behaves if we make the switch. Propose we take this change and solve the kind problem for now, and invest in using just the rbac type in future if needed. These two roles are truly not needed in hive dev clusters.

staebler: That's fine.

It would not affect stage and prod. They would both continue to use the same that they are using now because they are running on an older version of kubernetes that does not support a group kind for the rbac role binding.

My overarching concern is trying to push back on keeping SRE-specific functionality in the Hive repo. If these roles are not something that is needed to run Hive, then they should not be part of the Hive repo. If they are only needed to support some operation that SRE is doing, then SRE should be responsible for maintaining and applying those role bindings.

dgoodwin: I agree we should keep SRE specifics out and this one is borderline, we considered that at the time but reasoned that this was generically useful enough to consider fair game for including in the main repo.

PR #613, opened 2019-10-17
Manage DNS E2E

dgoodwin:


dgoodwin:
time="2019-10-18T18:24:26Z" level=error msg="cannot create DNS zone" clusterDeployment=hive-ff7519fe-6ef2-4116-8a78-6e70d57a8b88 controller=clusterDeployment error="admission webhook \"dnszonevalidators.admission.hive.openshift.io\" denied the request: Failed validation: a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')" namespace=cluster-test zone=cluster-test/hive-ff7519fe-6ef2-4116-8a78-6e70d57a8b88-zone

baseDomain: HEgo.hive-ci.openshift.com

Switching this to generate a lowercase prefix.

dgoodwin:
Promising:

************* INSTALL JOB LOG *************
level=info msg="Waiting up to 30m0s for the cluster at https://api.hive-0ec509c6-6c04-4123-9b4d-75de3fae0e98.mabrfgix.hive-ci.openshift.com:6443 to initialize..."
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring is still updating"
Found PVC for cluster deployment: hive-0ec509c6-6c04-4123-9b4d-75de3fae0e98-install-logs

/test e2e

dgoodwin:
e2e passed. Squashed commits. Ready for review.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -0,0 +1,211 @@
...
+		log.WithError(err).Fatal("error updating HiveConfig")
+	}
+	log.Info("updated HiveConfig")
+
+	// Adding manageDNS to HiveConfig triggers a new rollout of the hiveadmission pods.
+	// To know when it's safe to proceed, we will list all hiveadmission pods and ensure
+	// they all have the expected env var set.
+	log.Info("waiting for new hiveadmission pods to deploy")
+	selector := map[string]string{"hiveadmission": "true"}
+	for i := 0; i < 24; i++ {
```
staebler: I would prefer to see this done with a `UntilWithSync` from k8s.io/client-go/tools/watch. It does not matter too much, though.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	err = dynClient.Update(context.Background(), hc)
+	if err != nil {
+		log.WithError(err).Fatal("error updating HiveConfig")
+	}
+	log.Info("updated HiveConfig")
+
+	// Adding manageDNS to HiveConfig triggers a new rollout of the hiveadmission pods.
+	// To know when it's safe to proceed, we will list all hiveadmission pods and ensure
```
staebler: It strikes me that it would have been better to check the rollout status of the deployment rather than checking the pods directly.

contrib/pkg/adm/managedns/enable.go
dgoodwin: Looks useful will try to remember that in future.

contrib/pkg/adm/managedns/enable.go
dgoodwin: I tried to get it to happen with the deployment but I couldn't see a way, there were only two conditions to work with and minavailable never flipped true due to how it does the rollout. It's likely possible with replicaset but it didn't look simple. I opted for the pods as it's quite simple and gives us clear info on the state we actually want. It also gets past any race conditions where the changes aren't reflected in the deployment/RS if we immediately check right after editing hiveconfig.

staebler: I was thinking something along the lines of the following.
1) Get the revision of the deployment.
2) Change the hiveconfig.
3) Wait for the deployment to have both a new revision and a condition of Available=true.

dgoodwin: What I was seeing however was that Available=true throughout the process, as long as there are min replicas ready, it's happy, and the deployment rolls out a new pod before replacing an old one so it was always marked available.

staebler: Interesting. Maybe it is the Progressing condition then instead. Nevertheless, the code is adequate the way it is now.

staebler: I do not know what `kubectl rollout` does under the covers.

dgoodwin:
[test]

PR #617, opened 2019-10-17
controllers: remove code to set spec installed when legacy status installed set

staebler:
All of the clusterdeployments in prod and stage that have status installed set also have spec installed set.


LalatenduMohanty:
Code looks fine to me but I am not sure if there is  something else we need to think before deploying this in production.

dgoodwin:
Looks good, just needs a conflict resolution.

PR #618, opened 2019-10-17
controllers: set .spec.installed last

staebler:
When an installation succeeds for a clusterdeployment, set the status fields in the clusterdeployment first and set the .spec.installed field last. This ensures that things like the kubeconfig are available when the clusterdeployment is marked as installed.
   
This was not done this way originally because the controller needed to accommodate legacy clusterdeployments that did not have a .spec.installed field. At this point, all clusterdeployments have the field and have it set appropriately.


staebler:
This PR is built on #617.

/hold

dgoodwin:
Looks good but conflicts.

PR #619, opened 2019-10-17
controllers: log NotFound errors at info level

staebler:


LalatenduMohanty:
I am not sure why @staebler's PRs are automatically approved, may be we need to add @staebler https://github.com/openshift/hive/blob/master/OWNERS

staebler:
/lgtm cancel

Conflicts need to be resolved

PR #621, opened 2019-10-17
Add cluster platform label to ClusterDeployment

cblecker:
ref: https://jira.coreos.com/browse/CO-604

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -320,6 +320,19 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, r.Update(context.TODO(), cd)
 	}
 
+	// Set platform label on the ClusterDeployment
+	if platform := getClusterPlatform(cd); cd.Labels[hivev1.HiveClusterPlatformLabel] != platform {
+		if cd.Labels == nil {
+			cd.Labels = make(map[string]string)
+		}
+		cd.Labels[hivev1.HiveClusterPlatformLabel] = platform
+		err := r.Update(context.TODO(), cd)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to set cluster platform label")
+		}
+		return reconcile.Result{}, err
+	}
+
 	if cd.DeletionTimestamp != nil {
 		if !controllerutils.HasFinalizer(cd, hivev1.FinalizerDeprovision) {
 			clearUnderwaySecondsMetrics(cd)
@@ -1891,4 +1904,17 @@ func (r *ReconcileClusterDeployment) setSyncSetFailedCondition(cd *hivev1.Cluste
 		return false, err
 	}
 	return true, nil
+}
+
+// getClusterPlatform returns the platform of a given ClusterDeployment
+func getClusterPlatform(cd *hivev1.ClusterDeployment) string {
+	switch {
+	case cd.Spec.Platform.AWS != nil:
+		return "aws"
+	case cd.Spec.Platform.Azure != nil:
+		return "azure"
+	case cd.Spec.Platform.GCP != nil:
+		return "gcp"
+	}
+	return "unknown"
 }
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -1124,6 +1124,20 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "Add cluster platform label",
+			existing: []runtime.Object{
+				testClusterDeploymentWithoutPlatformLabel(),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				if assert.NotNil(t, cd, "missing clusterdeployment") {
+					if assert.NotNil(t, cd.Labels, "missing labels map") {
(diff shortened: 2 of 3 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -320,6 +320,17 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, r.Update(context.TODO(), cd)
 	}
 
+	// Set platform label on the ClusterDeployment
+	platform := getClusterPlatform(cd)
+	if cd.Labels == nil {
+		cd.Labels = make(map[string]string)
+	}
+	cdPlatform, cdPlatformLabelExists := cd.Labels[hivev1.HiveClusterPlatformLabel]
+	if !cdPlatformLabelExists || cdPlatform != platform {
```
staebler: The check of `!cdPlatformLabelExists` is not necessary. If the label does not exist, then `cdPlatform` will be an empty string.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -320,6 +320,17 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, r.Update(context.TODO(), cd)
 	}
 
+	// Set platform label on the ClusterDeployment
+	platform := getClusterPlatform(cd)
+	if cd.Labels == nil {
```
staebler: From a code organization standpoint, I would rather see this inside the next if statement. In other words, only make the map if the platform label does not match the cd platform (understanding that if the labels field is nil that they won't match anyway).

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -320,6 +320,17 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+		cd.Labels = make(map[string]string)
+	}
+	cdPlatform, cdPlatformLabelExists := cd.Labels[hivev1.HiveClusterPlatformLabel]
+	if !cdPlatformLabelExists || cdPlatform != platform {
+		cd.Labels[hivev1.HiveClusterPlatformLabel] = platform
+		return reconcile.Result{}, r.Update(context.TODO(), cd)
```
staebler: Please log a non-nil error returned from `Update`. Use `controllerutils.LogLevel(err)` as the log level of the log entry emitted.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1402,6 +1419,12 @@ func testClusterDeploymentWithoutFinalizer() *hivev1.ClusterDeployment {
 	return cd
 }
 
+func testClusterDeploymentWithoutPlatformLabel() *hivev1.ClusterDeployment {
+	cd := testClusterDeployment()
+	cd.Labels = map[string]string{}
```
staebler: Make this more resilient to future changes in `testClusterDeployment`.
```
delete(cd.Labels, hivev1.HiveClusterPlatformLabel)
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1377,6 +1391,9 @@ func testClusterDeployment() *hivev1.ClusterDeployment {
 		},
 	}
 
+	cd.Labels = make(map[string]string)
```
staebler: To make this more resilient to future changes in `testEmptyClusterDeployment`, move the creation of the `Labels` map to `testEmptyClusterDeployment`.

cblecker:
@staebler Thanks! Feedback addressed 🚀

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -320,6 +320,19 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, r.Update(context.TODO(), cd)
 	}
 
+	// Set platform label on the ClusterDeployment
+	if platform := getClusterPlatform(cd); cd.Labels[hivev1.HiveClusterPlatformLabel] != platform {
```
LalatenduMohanty: This looks odd. If the cd already has a label and it does not match with `getClusterPlatform()` we should not overwrite it as this condition should not come. Which means there is bigger issue somewhere. So we need to add warning here that it does not match. We should only add the label when it does not exist.

staebler: I disagree. Hive should be authoritative on the label. If the user removes or changes the label, Hive should change it back. Hive could add a warning that it has been changed, or Hive could log an info whenever setting it, but Hive should always ensure that the label is set properly.

cblecker: I don't expect any process would change or alter this label, but that said, my thoughts were along the lines of what @staebler said.. this controller should be authoritative for the label, and replace it with what it knows to be the correct value.

We could possibly add a warning inside this if, that if the label is not an empty string, then log out something with the old and new values as we overwrite.. but personally I'm not sure that's necessary.

dgoodwin:
@LalatenduMohanty is out for a couple days and we're going to freeze soon until Tuesday. Lets take this and do a followup if more changes necessary.

/lgtm

pkg/controller/clusterdeployment/clusterdeployment_controller.go
LalatenduMohanty: I strongly feel we should add logging to at least tell user when we do this. So it would be explicit then implicit. Sending a PR ro fix it.

PR #623, opened 2019-10-18
Clarify the imagePullSecret location

clyang82:
Signed-off-by: clyang82 <clyang@cn.ibm.com>

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -7,7 +7,7 @@
   1. You will need a live and functioning Route53 DNS zone in the AWS account you will be installing the new cluster(s) into. For example if you own example.com, you could create a hive.example.com subdomain in Route53, and ensure that you have made the appropriate NS entries under example.com to delegate to the Route53 zone. When creating a new cluster, the installer will make future DNS entries under hive.example.com as needed for the cluster(s).
     * Note that there is an additional mode of DNS management where Hive can automatically create delegated zones for approved base domains. i.e. if hive.example.com exists already, you can specify a base domain of cluster1.hive.example.com on your ClusterDeployment, and Hive will create this zone for you, wait for it to resolve, and then proceed with installation. See below for additional info.
   1. Determine what OpenShift release image you wish to install, and possibly what Hive image you want to use to install it.
-  1. Create a Kubernetes secret containing a docker registry pull secret (typically obtained from [try.openshift.com](https://try.openshift.com)).
+  1. Create a Kubernetes secret containing a docker registry pull secret (typically obtained from [Infrastructure Provider](https://cloud.redhat.com/openshift/install/)).
         ```bash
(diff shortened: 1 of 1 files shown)


docs/using-hive.md
```diff
@@ -7,7 +7,7 @@
...
     * Note that there is an additional mode of DNS management where Hive can automatically create delegated zones for approved base domains. i.e. if hive.example.com exists already, you can specify a base domain of cluster1.hive.example.com on your ClusterDeployment, and Hive will create this zone for you, wait for it to resolve, and then proceed with installation. See below for additional info.
   1. Determine what OpenShift release image you wish to install, and possibly what Hive image you want to use to install it.
-  1. Create a Kubernetes secret containing a docker registry pull secret (typically obtained from [try.openshift.com](https://try.openshift.com)).
+  1. Create a Kubernetes secret containing a docker registry pull secret (typically obtained from [Infrastructure Provider](https://cloud.redhat.com/openshift/install/)).
```
dgoodwin: Could you elaborate on the thought process behind this change? I'm not sure why we'd change try.openshift.com to Infrastructure Provider. I'm also not sure if immediately sending people to a login page is preferable to the level above with a little info on OpenShift 4 at try.openshift.com.

clyang82: I cannot find the imagePullSecret from the try.openshift.com. And I think it is not easy for the beginner to find it by try.openshift.com. change it to infrastructure provider, the user should be easy to find the imagepullsecret for different provider. WDYT?

dgoodwin: try.openshift.com links to cloud.redhat.com/openshift/install at the top, but it contains a little more info about what's going on vs just dropping you at a RH login screen, which might be a little confusing to some. 

For infrastructure providers, the tokens from cloud.redhat.com are valid regardless of the underlying infrastructure. Now in theory, other uses of Hive might implement their own pull secret mechanisms, but this link in particular is meant to imply that typically you'd be obtaining this from Red Hat.

I think I'm ok with the link change, but how about we change to:

[Red Hat](https://cloud.redhat.com/openshift/install/)) - or -
[cloud.redhat.com](https://cloud.redhat.com/openshift/install/))

staebler: I disagree. The URL that we want to be providing to all users of OpenShift is https://try.openshift.com/. We should not provide a URIL that bypasses that initial landing page.

dgoodwin: Honestly this is probably correct, apologies @clyang82 but the current form is probably best, try.openshift.com is the landing page that we should be promoting and the link to cloud.redhat.com is quite prominent at the top of the page.

clyang82: That's OK. Thanks @staebler @dgoodwin for your effort. I am OK with this purpose to expose the landing page for all users.

PR #624, opened 2019-10-18
Fix bootstrap log gathering on 4.2 by using openshift-install gather …

dgoodwin:
…bootstrap.

For 4.1 we relied on terraform tfstate file to locate the bootstrap IP,
this file format has changed and we should not use an installer
implementation detail if possible.

In 4.2 openshift-install adds the "gather bootstrap" command to do this
work for us, without relying on any terraform internals.

This patch removes support for 4.1 bootstrap log gathering, we've never
seen it work in prod and 4.2 is out now. oc adm must-gather will still
be run against 4.1 clusters that make it past bootstrap, but still fail
to initialize.


dgoodwin:
I have tested this locally with kind and create-cluster --simulate-bootstrap-failure, it successfully grabbed logs.

The installer attempts to gather logs for you using local users ~/.ssh, but considering the turnaround time on testing an attempt at this I felt this was less error prone and more consistent with the current code.

pkg/installmanager/installmanager.go
```diff
@@ -763,23 +749,38 @@ func (m *InstallManager) gatherBootstrapNodeLogs(cd *hivev1.ClusterDeployment) e
...
-	_, err = m.runGatherScript(bootstrapIP, fetchLogsScriptTemplate, m.WorkDir)
+	m.log.Infof("copying log bundles from %s to %s", m.WorkDir, m.LogsDir)
+	logBundles, err := filepath.Glob(filepath.Join(m.WorkDir, "log-bundle-*.tar.gz"))
 	if err != nil {
-		m.log.Error("failed to run script for gathering logs")
+		m.log.WithError(err).Error("erroring globbing log bundles")
 		return err
 	}
+	for _, lb := range logBundles {
+		// Using mv here rather than reading them into memory to write them out again.
```
staebler: Comment says using mv but code uses cp.

dgoodwin: My bad should be mv.

dgoodwin: Updated.

PR #625, opened 2019-10-18
Add variant of kubefed's create-clusters.sh script.

dgoodwin:
What do we think about this? This is mostly a copy of kubefed's script I just included the wait util at the bottom directly.

First time usage is basically:

```
CREATE_INSECURE_REGISTRY=true CONFIGURE_INSECURE_REGISTRY_HOST=TRUE CONFIGURE_INSECURE_REGISTRY_CLUSTER=true NUM_CLUSTERS=1 hack/create-clusters.sh
```

After this you've got a registry you can use locally:

```
IMG=172.17.0.1:5000/hive:latest make docker-dev-push
DEPLOY_IMAGE=172.17.0.1:5000/hive:latest make deploy
```

Now if you want to redeploy your hive cluster, you need to drop a few options so you don't try to re-deploy the local registry:

```
CONFIGURE_INSECURE_REGISTRY_CLUSTER=true NUM_CLUSTERS=1 hack/create-clusters.sh
```

Do we like this approach? I'd almost like a script to setup local registry and configure the host, create clusters with kind directly, and a script to configure a kind cluster with the local registry.

Thoughts?

### DIFF
--- a/hack/create-clusters.sh
+++ b/hack/create-clusters.sh
@@ -1,0 +1,248 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


staebler:
My view is that we can take this PR as is and iterate on it.

Without giving it too much thought, what I would like to see is a make target for pushing to the local registry and a make target for creating a kind cluster.

Pushing to the local registry:
1) Create the registry if it does not exist.
2) Pushes to the registry.

Creating a kind cluster:
1) Create the cluster if it is not already created.
2) Create the registry if it does not exist.
3) Configure the registry on the cluster.

dgoodwin:
Ok lets start with this then.

PR #629, opened 2019-10-22
controllers: remove dead code in syncsetinstance controller

staebler:
There is dead code in the getSyncSetCommonSpec function that will never be reached.


staebler:
@joelddiaz Prow was not responding to my requests to retest, so I pushed up a rebased commit. Can you add your lgtm again?

PR #634, opened 2019-10-23
Vendor gcp clients

twiest:
- [x] Vendor gcp serviceusage, cloudresources and oauth clients
- [x] Create minimal gcpclient that can auth and list a resource


pkg/gcpclient/client.go
```diff
@@ -0,0 +1,76 @@
...
+	defer cancel()
+
+	return c.dnsClient.ManagedZones.List(c.projectName).Context(ctx).Do()
+}
+
+// NewClient creates our client wrapper object for interacting with GCP.
+func NewClient(projectName string, authJSON []byte) (Client, error) {
+	ctx := context.TODO()
+
+	// since we're using a single creds var, we should specify all the required scopes when initializing
```
joelddiaz: i feel like i should know the answer to this, but what exactly is the comment talking about?

twiest: This was copied from: https://github.com/openshift/cloud-credential-operator/blob/master/pkg/gcp/client.go#L195

PR #610, opened 2019-10-17
Add PersistentVolume in using hive section

clyang82:
Fixed #609
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

PR #642, opened 2019-10-30
Add script and documentation for vanilla kube (kind) deployment.

dgoodwin:
We were relying on OpenShift specifics to generate and inject a cert for
hiveadmission's webhooks. Add a script for devs looking to deploy on
kind which uses Kubernetes CSRs to request a cert, and then uses the
kube CA.crt for the webhooks.


dgoodwin:
This isn't quite right as we're triggering a "openshift 3.11" path to add the CAs to the webhooks, which is misleading, and I'm also not sure yet if adding those CAs is required. 

This PR does however work.

dgoodwin:
Implementation ideas from: https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/

dgoodwin:
Clarified the code, the CA injection does appear to still be required, and added a section in the dev docs on how to use kind + local insecure registry.

Ready for review.

docs/developing.md
```diff
@@ -32,6 +32,43 @@ cd $GOPATH/src/openshift
 git clone https://github.com/openshift/hive.git
 ```
 
+## Deploying with Kubernetes on Docker (kind)
+
+[Kind](https://github.com/kubernetes-sigs/kind) can be used as a lightweight development environment for deploying and testing Hive. The following instructions cover creating an insecure local registry (allowing for dramatically faster push/pull), and configuring your host OS, as well as the kind cluster to access it. This approch runs Hive in a container as you would in production, giving you the best coverage for manual testing.
+
+This approach requires [Docker](https://docs.docker.com/install/linux/docker-ce/fedora/). At present we do not have kind working with podman.
```
gyliu513: How about use `https://docs.docker.com/install` as the docker link, as someone may also want to run on Ubuntu etc.

docs/developing.md
```diff
@@ -32,6 +32,43 @@ cd $GOPATH/src/openshift
...
+
+For first time usage you can create the insecure registry and the kind cluster in one command:
+
+```bash
+CREATE_INSECURE_REGISTRY=true CONFIGURE_INSECURE_REGISTRY_HOST=TRUE CONFIGURE_INSECURE_REGISTRY_CLUSTER=true NUM_CLUSTERS=1 hack/create-clusters.sh
+```
+
+`docker ps` should now show you a "registry" and a "cluster1" container running.
+
```
gyliu513: I'd like an output of `docker ps` here.

docs/developing.md
```diff
@@ -32,6 +32,43 @@ cd $GOPATH/src/openshift
 git clone https://github.com/openshift/hive.git
 ```
 
+## Deploying with Kubernetes on Docker (kind)
```
gyliu513: /s/on/IN 

KIND: Kubernetes IN Docker

gyliu513:
Hmm, seems with `KIND`, hive admission pod failed to start with following errors:

```yaml
Events:
  Type     Reason       Age                   From                        Message
  ----     ------       ----                  ----                        -------
  Normal   Scheduled    5m24s                 default-scheduler           Successfully assigned hive/hiveadmission-dd7fb688f-d2hb4 to ocp-control-plane
  Warning  FailedMount  74s (x10 over 5m24s)  kubelet, ocp-control-plane  MountVolume.SetUp failed for volume "serving-cert" : secret "hiveadmission-serving-cert" not found
  Warning  FailedMount  64s (x2 over 3m21s)   kubelet, ocp-control-plane  Unable to mount volumes for pod "hiveadmission-dd7fb688f-d2hb4_hive(7f6908a5-fc80-11e9-9247-024231d31907)": timeout expired waiting for volumes to attach or mount for pod "hive"/"hiveadmission-dd7fb688f-d2hb4". list of unmounted volumes=[serving-cert]. list of unattached volumes=[serving-cert hiveadmission-token-46hfj]
```

The steps that I did is as follows:

```
kind create cluster --name hive
```

```
create a pv for ClusterDeployment
```

```
make deploy
```

@dgoodwin I did not found a place where the secret `hiveadmission-serving-cert` was created from source code, can you show some light for this? Thanks!

docs/developing.md
dgoodwin: Oops :), will fix, thanks!

docs/developing.md
dgoodwin: Good idea.

docs/developing.md
dgoodwin: :+1:

dgoodwin:
@gyliu513 the ./hack/hiveadmission-dev-cert.sh script will create that secret for you.

The kind create command is contained into the script in this PR as well which layers in the setup for the local insecure registry which is *really* handy for quickly building an image and deploying. It currently creates cluster1, I will change that to hive.

What was the PV for, log gathering? I thought I'd seen that work without doing anything, but I would have to verify.

In a followup I hope to split the scripts into on that sets up the registry and configures localhost, and one that creates a hive cluster with a given name and configures it for the registry.

dgoodwin:
I split the scripts into two, one to create the local insecure registry, and one to create a kind cluster with a given name and configure it to talk to the registry. The latter can be used to create or recreate as many clusters as you want.

gyliu513:
@dgoodwin the pv was introduced by https://github.com/openshift/hive/pull/610/files , and it is used by the `ClusterDeployment` to persist the installation logs.

gyliu513:
@dgoodwin the update looks good to me, thanks!

dgoodwin:
> @dgoodwin the pv was introduced by https://github.com/openshift/hive/pull/610/files , and it is used by the `ClusterDeployment` to persist the installation logs.

I have to check, I thought I've seen the log gathering work out of the box with a bare kind cluster. If not I will add this step, thanks!

docs/developing.md
```diff
@@ -32,6 +32,56 @@ cd $GOPATH/src/openshift
 git clone https://github.com/openshift/hive.git
 ```
 
+## Deploying with Kubernetes In Docker (kind)
```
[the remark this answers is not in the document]
dgoodwin: No I don't think kind instructions are a good fit for installing, better in developing.md IMO.

joelddiaz:
@dgoodwin is this PR ready to merge?

dgoodwin:
It is yes.

PR #643, opened 2019-10-31
docs: Corrected aws environment var.

gyliu513:
/cc @LalatenduMohanty


dgoodwin:
Good catch, thanks!

/lgtm

/ok-to-test

PR #645, opened 2019-10-31
fix resync

mjudeikis:
Fix resync of Syncsets

@staebler  is this what you had in mind?

/cc @dgoodwin @abutcher


dgoodwin:
Thanks @mjudeikis

PR #646, opened 2019-10-31
docs: debug hive does not request scale down to 0.

gyliu513:
Hive will be restarted automatically after edit finished.

/cc @LalatenduMohanty

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -83,12 +83,6 @@ We use a hiveutil subcommand for the install-manager, in pods and thus in an ima
 
 ## Enable Debug Logging In Hive Controllers
 
-Scale down the Hive operator to zero
-
-```bash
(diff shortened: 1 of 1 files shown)


dgoodwin:
/hold

Sorry I think the docs might be correct, there are two Deployments in play. hive-operator is responsible for creating and maintaining hive-controllers, if you just edit hive-controllers and the operator is still running, on the next reconciling it will revert your change and you'll lose your debug logging. 

Unless I'm mistaken and the semantics of kubectl apply won't clobber that new option, but from what I recall it would.

staebler:
Oh, right. I missed that important detail.

/lgtm cancel

gyliu513:
Ah, I see, thanks @dgoodwin and @staebler , let me close this PR :-)

PR #648, opened 2019-10-31
Refactor dnszone controller into the actuator pattern

twiest:
- [x] Create an actuator interface
- [x] Refactor zonereconciler into an AWS actuator that implements the
      actuator interface.
- [x] Refactor unit tests to be split between the dnszone controller and
      the AWS actuator
- [x] Ensure unit test coverage has not decreased.
- [x] Manually test dnszone controller with AWS actuator


twiest:
This PR is ready to be reviewed. Holding it until manual testing is completed.

/hold

dgoodwin:
Matt covered what I had in my review and then some. Only additional comment I had was to consider adding a compile time check that the objects meet the Actuator interface. https://stackoverflow.com/questions/10498547/ensure-a-type-implements-an-interface-at-compile-time-in-go

pkg/controller/dnszone/awsactuator_test.go
```diff
@@ -0,0 +1,174 @@
...
+			//assertErrorNilOrMessage(t, err, tc.expectedErrString)
+			if tc.expectedErrString == "" {
+				// Only check this if we're expected to succeed.
+				assert.Equal(t, expectedAWSActuator, zr)
+			}
+		})
+	}
+}
+
+func mockZoneExists(expect *mock.MockClientMockRecorder, zone *hivev1.DNSZone) {
```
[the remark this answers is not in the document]
twiest: @staebler and I discussed this in person and we'll look to add this in a future PR.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -0,0 +1,462 @@
...
+	logger log.FieldLogger
+
+	// awsClient is a utility for making it easy for controllers to interface with AWS
+	awsClient awsclient.Client
+
+	// scheme is the controller manager's scheme
+	scheme *runtime.Scheme
+
+	// currentHostedZone is the returned hosted zone from route53
+	currentHostedZone *route53.HostedZone
```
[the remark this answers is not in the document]
twiest: Done.

twiest:
@staebler I believe I've addressed your feedback. I gave up on the `DeepCopy()` idea, so this should be coded how you suggested.

I also manually tested the corner case that we updated the logic on and it worked as expected.

/hold cancel

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -199,13 +172,102 @@ func (r *ReconcileDNSZone) Reconcile(request reconcile.Request) (reconcile.Resul
...
 	if err != nil {
 		dnsLog.WithError(err).Error("Encountered error while attempting to reconcile")
 	}
 	return result, err
 }
 
+// ReconcileDNSProvider attempts to make the current state reflect the desired state. It does this idempotently.
+func (r *ReconcileDNSZone) ReconcileDNSProvider(actuator Actuator, dnsZone *hivev1.DNSZone) (reconcile.Result, error) {
+	r.logger.Debug("Retrieving current state")
+	err := actuator.Refresh()
```
staebler: `Refresh` is sticking out to me as something that is not well defined. For one thing, why does refresh fetch the zone but not the name servers. For another thing, it smells to me that `Refresh` is something that needs to be called immediately after creating the actuator and before doing anything else.

I have two alternatives in mind.
1) Have the `Exists` function fetch the zone ID and store it.
2) Have the `NewAWSActuator` function fetch the zone ID and store it. This is my preference.

In either case, there is no reason to store the tags: They can be fetched when needed in the `syncTags` function.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -236,21 +298,191 @@ func shouldSync(desiredState *hivev1.DNSZone) (bool, time.Duration) {
...
+		}
 
-	if dnsZone != nil && dnsZone.Spec.AWS != nil {
-		secretName = dnsZone.Spec.AWS.AccountSecret.Name
-		regionName = dnsZone.Spec.AWS.Region
+		return NewAWSActuator(dnsLog, secret, r.scheme, dnsZone, awsclient.NewClientFromSecret)
 	}
 
-	awsClient, err := r.awsClientBuilder(r.Client, secretName, dnsZone.Namespace, regionName)
+	return nil, fmt.Errorf("error: unable to determine which actuator to use")
```
staebler: Remove "error: " from the message. That is just noise in the error message.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -236,21 +298,191 @@ func shouldSync(desiredState *hivev1.DNSZone) (bool, time.Duration) {
 	return false, delta
 }
 
-// getAWSClient generates an awsclient
-func (r *ReconcileDNSZone) getAWSClient(dnsZone *hivev1.DNSZone, dnsLog log.FieldLogger) (awsclient.Client, error) {
-	// This allows for using host profiles for AWS auth.
-	var secretName, regionName string
+func (r *ReconcileDNSZone) getActuator(dnsZone *hivev1.DNSZone, dnsLog log.FieldLogger) (Actuator, error) {
+	if dnsZone.Spec.AWS != nil && dnsZone.Spec.AWS.AccountSecret.Name != "" {
```
staebler: It should be a separate error for the case where the spec stipulates AWS but the secret name is missing, rather than lumping the two together in a "unable to determine which actuator to use" error.

You probably don't even need to check that the name is not empty. If the name is empty, then the Get of the secret will throw an error anyway.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -0,0 +1,476 @@
...
+		logger.WithError(err).Error("cannot list tags for zone")
+		return nil, err
+	}
+	logger.WithField("tags", tagsString(resp.ResourceTagSet.Tags)).Debug("retrieved zone tags")
+	return resp.ResourceTagSet.Tags, nil
+}
+
+// Create makes an AWS Route53 hosted zone given the DNSZone object.
+func (actuator *AWSActuator) Create() error {
+	logger := actuator.logger.WithField("zone", actuator.dnsZone.Spec.Zone)
```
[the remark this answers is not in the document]
twiest: Totally agree. Would you mind capturing this in a card so we don't forget?   I'm more than happy to do the work.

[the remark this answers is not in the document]
staebler: https://jira.coreos.com/browse/CO-629

staebler:
@twiest Would you mind separating the vendor changes into its own commit. Could it be a revert of 10329f22fa1d56dcfb910c17bb7f5df195401bfb?

twiest:
> @twiest Would you mind separating the vendor changes into their own commit. Could it be a revert of [10329f2](https://github.com/openshift/hive/commit/10329f22fa1d56dcfb910c17bb7f5df195401bfb)?

Sure :)

pkg/controller/dnszone/dnszone_controller.go
twiest: I have the other changes done, but I don't fully understand what you're looking for on this one.

Are you looking to eliminate `Refresh()`?

What is the purpose of fetching and storing zoneID in `NewAWSActuator`? Like what does it accomplish?

The `syncTags` part is clear.

staebler: Yes, my suggestions is to eliminate `Refresh`. My issue is that `NewAWSActuator` does not yield anything usable. The actuator created is not usable until `Refresh` is called. If the user is required to call `Refresh` immediately after calling `NewAWSActuator`, then why not include the functionality of `Refresh` in the `NewAWSActuator` function and skip the extra function call.

twiest: Ok. Why store `zoneID` instead of `currentHostedZone`?

Would we just look up the hosted zone each time?

Is this to ensure the hosted zone isn't "stale" ?

staebler: I don't care much whether the actuator stores the entire hosted zone or just the ID. For what it's worth, though, nothing in the hosted zone other than the ID is used in the actuator.

twiest: You're right! I hadn't noticed that. :)

twiest: So I've been thinking about this change this weekend and while I see your point on Refresh, I'm just not convinced that putting that functionality into `NewAWSActuator` is a good idea.

1. `NewAWSActuator` is basically a constructor. I don't like the idea of having the constructor call out to the cloud.
1. I was also hoping to not significantly alter the flow control / logic of the current dnszone controller for the actuator refactor. Adopting the existing logic gives us the highest chance of not introducing bugs.
1. I'd like more time to think of a better way to address this concern. But I don't think we can afford to take that time right now. If we want to get the GCP actuator done this sprint, I really need to start focusing on it.

Proposal: I'll create a Jira card to think of and implement something to address this concern. Between now and when we do that card I'll think about a way to address the concern without having the constructor reach out to the cloud. To be clear, this card would also handle the `syncTags` part as well since the two seem related.

I have the other feedback that you've pointed out addressed (except putting the re-vendor of the SOA check deps into its own commit, which I can do today). I haven't pushed any of those changes yet.

@staebler @dgoodwin thoughts?

staebler: 1. I agree about getting this in without making any more changes around `Refresh` at this point.
2. Yes, `NewAWSActuator` is a constructor. An object returned by a constructor should be usable. This constructor does not return a usable object until `Refresh` is called. That is my issue.
3. I do not see where putting the `Refresh` logic in a stand-alone function or in the constructor changes the flow/logic in any way. I could make an argument that adding the `NewAWSActuator` as an additional function while having `Refresh` replace `getCurrentState` actually makes more changes than having `NewAWSActuator` be a replacement for `getCurrentState`.
4. If the concern is that it is peculiar to have the constructor make calls out to the cloud, then you could have each of the public functions that requires a refresh in order to be usable (1) check whether a refresh has been executed and (2) execute the refresh if needed.
5. I would not bother created a Jira card. I doubt that the card would ever be worked on. Once this PR merges, that is going to be the interface for the Actuator. And this is fine: It is just an internal interface used by one component so it does not need to be perfect.

twiest:
Holding until we can talk about the `Refresh` thing with @dgoodwin. 

/hold

pkg/controller/dnszone/awsactuator.go
```diff
@@ -0,0 +1,488 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	awsclient "github.com/openshift/hive/pkg/awsclient"
+)
+
+const (
+	hiveDNSZoneTag                  = "hive.openshift.io/dnszone"
+	domainAvailabilityCheckInterval = 30 * time.Second
```
staebler: These constants should be in dnszone_controller.go

dgoodwin:
Yeah lets just merge it, the Refresh concern is valid but it is an internal interface and I'm pretty sure we've made that mistake elsewhere. Lets keep things moving and we can worry about it next time someone is in there.

PR #650, opened 2019-11-02
docs: added a note for the creation of secret for AWS

gyliu513:
The secret will be created from either AWS credential file or from
the env varsi, it was already handled in the create-cluster logic.

/cc @dgoodwin

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -13,8 +13,9 @@
         ```
   1. Create a Kubernetes secret containing your AWS credentials:
         ```bash
-        oc create secret generic test-creds --from-literal=aws_secret_access_key=$AWS_SECRET_ACCESS_KEY --from-literal=aws_access_key_id=$AWS_ACCESS_KEY_ID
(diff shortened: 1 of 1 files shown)


docs/using-hive.md
```diff
@@ -11,10 +11,6 @@
         ```bash
         oc create secret generic mycluster-pull-secret --from-file=.dockerconfigjson=/path/to/pull-secret --type=kubernetes.io/dockerconfigjson
         ```
-  1. Create a Kubernetes secret containing your AWS credentials:
-        ```bash
-        oc create secret generic test-creds --from-literal=aws_secret_access_key=$AWS_SECRET_ACCESS_KEY --from-literal=aws_access_key_id=$AWS_ACCESS_KEY_ID
-        ```
```
dgoodwin: We view create-cluster as a developer tool, and it might be good to keep this step documented here so external users of Hive's API get a clear picture of each step involved. Could we perhaps keep this block but add a NOTE indicating this will be done automatically if using hiveutil create-cluster.

gyliu513: Make sense, thanks @dgoodwin

gyliu513:
@dgoodwin can you enable `ok-to-test` for this PR? Thanks!

PR #652, opened 2019-11-04
Move API from v1alpha1 to v1.

dgoodwin:
This only moves the types as they are currently to the new version. We
will begin updating for V1 shortly.

Likewise we will soon restore v1alpha1 types and the migration code +
aggregated APIserver to handle them.


dgoodwin:
I've created a v1 branch off master, and this PR to merge the first work *into* that branch, rather than master.

dgoodwin:
My bad, I think I've got them all now.

```
$ git grep v1alpha1 | grep -v vendor | grep -v Gopkg.lock                  (hive/172-17-0-4:6443//hive)
cmd/manager/main.go:	crv1alpha1 "k8s.io/cluster-registry/pkg/apis/clusterregistry/v1alpha1"
cmd/manager/main.go:			if err := crv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
config/templates/hive-csv-template.yaml:apiVersion: operators.coreos.com/v1alpha1
hack/olm-registry/olm-artifacts-template.yaml:- apiVersion: operators.coreos.com/v1alpha1
hack/olm-registry/olm-artifacts-template.yaml:- apiVersion: operators.coreos.com/v1alpha1
```

dgoodwin:
make verify and make test are clean.

PR #654, opened 2019-11-05
add a default instance type for GCP

joelddiaz:
mirror what we do for AWS and provide a default instance type for GCP when creating a clusterDeployment

### DIFF
--- a/contrib/pkg/createcluster/gcp.go
+++ b/contrib/pkg/createcluster/gcp.go
@@ -16,7 +16,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


staebler:
It is not sufficient to fill out the field when using hiveutl create-cluster. If the field needs to be filled out for Hive to work, then the field needs to be made mandatory. Otherwise a user creating a clusterdeployment may created an accepted clusterdeployment that will fail once the cluster has been installed.

joelddiaz:
/close
we'll use the installer's functions to fill in defaults for the installconfig

PR #657, opened 2019-11-06
fix byo installer image

mjudeikis:
Fix BYO-Installer image overrides in `ClusterDeployment` and `ImageSet`

/cc @staebler @dgoodwin

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -855,16 +855,26 @@ func (r *ReconcileClusterDeployment) statusUpdate(cd *hivev1.ClusterDeployment, 
 }
 
 func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDeployment, imageSet *hivev1.ClusterImageSet, releaseImage string, cdLog log.FieldLogger) (reconcile.Result, error) {
-	if len(cd.Spec.Images.InstallerImage) > 0 {
-		cdLog.WithField("image", cd.Spec.Images.InstallerImage).
-			Debug("setting status.InstallerImage to the value in spec.images.installerImage")
-		cd.Status.InstallerImage = &cd.Spec.Images.InstallerImage
-		return reconcile.Result{}, r.statusUpdate(cd, cdLog)
-	}
+	// we set InstallerImage only if it is not set already to the same image
 	if imageSet != nil && imageSet.Spec.InstallerImage != nil {
-		cd.Status.InstallerImage = imageSet.Spec.InstallerImage
-		cdLog.WithField("imageset", imageSet.Name).Debug("setting status.InstallerImage using imageSet.Spec.InstallerImage")
-		return reconcile.Result{}, r.statusUpdate(cd, cdLog)
+		if cd.Status.InstallerImage != nil && *cd.Status.InstallerImage == *imageSet.Spec.InstallerImage {
+			cdLog.WithField("imageset", imageSet.Name).Debug("status.InstallerImage already set to imageSet.Spec.InstallerImage")
+		} else {
+			cd.Status.InstallerImage = imageSet.Spec.InstallerImage
+			cdLog.WithField("imageset", imageSet.Name).Debug("setting status.InstallerImage using imageSet.Spec.InstallerImage")
+			return reconcile.Result{}, r.statusUpdate(cd, cdLog)
+		}
+	}
+	// cd.Spec.Images.InstallerImage takes over ImageSet.InstallerImage
+	if len(cd.Spec.Images.InstallerImage) > 0 {
+		if cd.Status.InstallerImage != nil && *cd.Status.InstallerImage == cd.Spec.Images.InstallerImage {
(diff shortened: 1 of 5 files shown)


mjudeikis:
Tested, works. if you fine with the proposed solution I might add unit tests too

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -855,16 +856,26 @@ func (r *ReconcileClusterDeployment) statusUpdate(cd *hivev1.ClusterDeployment,
...
-			Debug("setting status.InstallerImage to the value in spec.images.installerImage")
-		cd.Status.InstallerImage = &cd.Spec.Images.InstallerImage
-		return reconcile.Result{}, r.statusUpdate(cd, cdLog)
-	}
+	// we set InstallerImage only if it is not set already to the same image
 	if imageSet != nil && imageSet.Spec.InstallerImage != nil {
-		cd.Status.InstallerImage = imageSet.Spec.InstallerImage
-		cdLog.WithField("imageset", imageSet.Name).Debug("setting status.InstallerImage using imageSet.Spec.InstallerImage")
-		return reconcile.Result{}, r.statusUpdate(cd, cdLog)
+		if cd.Status.InstallerImage != nil && strings.EqualFold(*cd.Status.InstallerImage, *imageSet.Spec.InstallerImage) {
```
staebler: Are image URI case-insensitive?

pkg/imageset/updateinstaller.go
```diff
@@ -222,7 +222,10 @@ func (o *UpdateInstallerImageOptions) updateInstallerImage(installerImage, cliIm
 		logger.WithError(err).Error("failed to get ClusterDeployment")
 		return err
 	}
-	cd.Status.InstallerImage = &installerImage
+	// Set installer image only if it is not overrided by ImageSet or CD
```
staebler: I am concerned that this check, as is, will preserve an out-of-date installer image even when the imageset job yields a different installer image.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
mjudeikis: I would assume so, it just url

pkg/imageset/updateinstaller.go
mjudeikis: If we dont have this check, it will always override installer image to one from CLI. So we good to remove this feature at all, as it would always ignore values from ImageSet. 

The installer image is good only for cluster install. So its SRE responsibility to remove it from  from ImageSet when it is not needed anymore, and any new clusters will fallback into default logic. If we would use it again and again, I would be worried about this.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: There are certainly parts of the URI that are case-sensitive, such as the path. I don't see any reason to compare in a case-insensitive fashion here rather than just using the equality operator.

Interestingly, `oc` requires that the URI be all lowercase.
```
$ oc adm release info quay.io/openshift-release-dev/OCP-release:4.2.1
error: repository name must be lowercase
```

pkg/imageset/updateinstaller.go
staebler: What if we had the update-installer-image command take a parameter for the installer image override. If that parameter is set, then the commands writes that installer image to the clusterdeployment status instead of writing the installer image extracted from the release image. Bonus points if the extract-image script skips extracting the installer image when there is an override.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
mjudeikis: I was thinking something under these lines https://stackoverflow.com/questions/15641694/are-uris-case-insensitive but I can change

pkg/imageset/updateinstaller.go
mjudeikis: sgtm

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: My reading of that link supports my contention that the comparison should be done in a case-sensitive fashion.

mjudeikis:
@staebler should be better now... I think...

pkg/imageset/updateinstaller.go
```diff
@@ -179,22 +181,28 @@ func (o *UpdateInstallerImageOptions) Run() error {
...
 	if success {
 		o.log.Debugf("the oc release info command was successful")
-		installerImageFileName := path.Join(o.WorkDir, "installer-image.txt")
-		installerImageBytes, err := ioutil.ReadFile(installerImageFileName)
-		if err != nil {
-			o.log.WithError(err).Error("could not read install image file")
-			return err
+		var installerImage, cliImage string
+		if len(o.InstallerImage) > 0 {
+			o.log.Debugf("installer image override provided - skip installer image resolution")
```
staebler: This is not actually skipping the installer image resolution. The resolution is being done in the other container of the imageset pod.

dgoodwin:
Hey @mjudeikis is this still on the table to get merged? Still a little feedback pending.

staebler:
/close

The installer-image override has been removed. See https://github.com/openshift/hive/pull/742.

PR #658, opened 2019-11-07
[WIP] Adding in-cluster Prometheus monitoring

LalatenduMohanty:
Will be only enabled for OpenShift4 clusters


joelddiaz:
@LalatenduMohanty this all looks reasonable to me. still WIP b/c you're doing some testing?

dgoodwin:
Am I correct we should close this as it doesn't actually work until 4.3?

PR #660, opened 2019-11-08
Add InstallConfigSecret for v1 API.

dgoodwin:
To better support installer features as they land without requiring
exposure in Hive, we now accept an InstallConfig secret to be passed
directly to the installer. A secret was chosen as the InstallConfig
contains ssh key and pull secrets.

Some fields are now no longer required in the ClusterDeployment.

Updates create-cluster command and remotemachineset controller for this
change.


pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -286,7 +292,6 @@ var AllClusterDeploymentConditions = []ClusterDeploymentConditionType{
 // ClusterDeployment is the Schema for the clusterdeployments API
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="ClusterName",type="string",JSONPath=".spec.clusterName"
```
staebler: Should we leave this column while we still have clusterName in the spec?

contrib/pkg/createcluster/create.go
```diff
@@ -249,7 +252,12 @@ func (o *Options) Validate(cmd *cobra.Command) error {
 type cloudProvider interface {
 	generateCredentialsSecret(o *Options) (*corev1.Secret, error)
 
-	addPlatformDetails(o *Options, cd *hivev1.ClusterDeployment) error
+	addPlatformDetails(
+		o *Options,
+		cd *hivev1.ClusterDeployment,
+		installConfig *installertypes.InstallConfig,
+		controlPlanePool,
```
staebler: Why are the machine pools passed in separately rather than using the fields from the install config?

contrib/pkg/createcluster/aws.go
```diff
@@ -40,36 +43,58 @@ func (p *awsCloudProvider) generateCredentialsSecret(o *Options) (*corev1.Secret
...
+		InstanceType: "m4.xlarge",
+		EC2RootVolume: awsinstallertypes.EC2RootVolume{
 			IOPS: 100,
 			Size: 22,
 			Type: "gp2",
 		},
 	}
 
-	cd.Spec.ControlPlane.Platform = hivev1.MachinePoolPlatform{
+	controlPlanePool.Platform = installertypes.MachinePoolPlatform{
```
staebler: Can this be the following? Or are you trying to ensure also that no other platforms are set?
```
controlPlanePool.Platform.AWS = mpp
computePool.Platform.AWS = mpp
```

contrib/pkg/createcluster/create.go
```diff
@@ -325,19 +333,71 @@ func (o *Options) getResourceHelper() (*resource.Helper, error) {
 func (o *Options) GenerateObjects() ([]runtime.Object, error) {
 	result := []runtime.Object{}
 
-	pullSecret, err := o.generatePullSecret()
+	pullSecretSecret, err := o.generatePullSecret()
+	if err != nil {
+		return nil, err
+	}
+
+	pullSecret, err := o.getPullSecret()
```
staebler: Can we pass `pullSecret` in as a parameter to `generatePullSecret`? It looks strange to have `generatePullSecret` followed by `getPullSecret` with them having no visible relationship.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -115,12 +106,27 @@ type ClusterDeploymentSpec struct {
 
 	// Installed is true if the cluster has been installed
 	Installed bool `json:"installed"`
+
+	// Provisioning contains settings used only for initial cluster provisioning.
+	Provisioning Provisioning `json:"provisioning"`
```
staebler: Should this be optional to accommodate adopting installed clusters?

pkg/install/generate_test.go
```diff
@@ -10,6 +10,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	testName      = "foo"
```
staebler: These are both only used in one place. Can we use the string literals there rather than declaring constants?

pkg/installmanager/installmanager.go
```diff
@@ -231,26 +232,16 @@ func (m *InstallManager) Run() error {
...
 	m.waitForInstallerBinaries()
 
-	// Generate an install-config.yaml:
-	sshKey := os.Getenv("SSH_PUB_KEY")
-	pullSecret := os.Getenv("PULL_SECRET")
-	m.log.Info("generating install config")
-	ic, err := install.GenerateInstallConfig(cd, sshKey, pullSecret, true)
+	m.log.Info("copying install-config.yaml")
+	destInstallConfigPath := filepath.Join(m.WorkDir, "install-config.yaml")
+	cmd := exec.Command("cp", m.InstallConfigMountPath, destInstallConfigPath)
```
staebler: Should we copy the entire /installconfig directory over and not be proscriptive about what files are needed? I can think of a case where a future version of openshift-install uses different files.

contrib/pkg/createcluster/create.go
dgoodwin: Ugh I have absolutely no idea, will adjust, didn't even see that.

pkg/apis/hive/v1/clusterdeployment_types.go
dgoodwin: Yes good idea.

pkg/install/generate_test.go
dgoodwin: I was taught to avoid magic values but I don't even do this consistently. Will move them inline.

contrib/pkg/createcluster/create.go
dgoodwin: will do

pkg/install/generate_test.go
staebler: I agree that magic values can be problematic. I'm not sure that applies the same to test code. I prefer to see the magic values right where they are being used for tests. That may be just me, though.

pkg/installmanager/installmanager.go
staebler: That's reasonable.

dgoodwin:
feedback addressed.

contrib/pkg/createcluster/create.go
```diff
@@ -325,20 +331,72 @@ func (o *Options) getResourceHelper() (*resource.Helper, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	controlPlanePool := &installertypes.MachinePool{
+		Name:     "master",
+		Replicas: pointer.Int64Ptr(3),
+	}
+	computePool := &installertypes.MachinePool{
+		Name:     "worker",
+		Replicas: pointer.Int64Ptr(o.WorkerNodes),
```
staebler:
```
Replicas: &o.WorkerNodes,
```

contrib/pkg/createcluster/create.go
```diff
@@ -325,20 +331,72 @@ func (o *Options) getResourceHelper() (*resource.Helper, error) {
...
+	cd, err := o.GenerateClusterDeployment(pullSecretSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	sshPublicKey, err := o.getSSHPublicKey()
+	if err != nil {
+		return nil, err
+	}
+	controlPlanePool := &installertypes.MachinePool{
```
staebler: Don't need `controlPlanePool` and `computePool` variables. This info can be set directly in `installConfig` when it is initialized.

contrib/pkg/createcluster/create.go
```diff
@@ -373,11 +431,32 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
 		}
 	}
 
-	result = append(result, cd)
+	result = append(result, installConfigSecret, cd)
```
staebler: Should the install config secret be excluded when IncludeSecrets is false?

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -115,12 +106,28 @@ type ClusterDeploymentSpec struct {
...
+	Provisioning *Provisioning `json:"provisioning"`
+}
+
+// Provisioning contains settings used only for initial cluster provisioning.
+type Provisioning struct {
+	// InstallConfigSecret is the reference to a secret that contains an openshift-install
+	// InstallConfig. This file will be passed through directly to the installer.
+	// Any version of InstallConfig can be used, provided it can be parsed by the openshift-install
+	// version for the release you are provisioning.
+	InstallConfigSecret corev1.LocalObjectReference `json:"installConfigSecret,omitempty"`
```
staebler: This should not be `omitempty`. It is required when the `Provisioning` field is non-nil.

contrib/pkg/createcluster/create.go
dgoodwin: Tough one, but looking at that option and what it's for today it feels more like it's global files you use across most of your clusters, whereas the install-config is heavily generated and previous embedded in the cluster deployment. I think it would cause problems for someone using include secrets rather than help them. For now I'd say lets keep it outside that concept.

dgoodwin:
Updated with latest feedback.

contrib/pkg/createcluster/create.go
staebler: I am not sure what the purpose of the include secrets is, so I will defer to you.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -115,12 +106,28 @@ type ClusterDeploymentSpec struct {
 
 	// Installed is true if the cluster has been installed
 	Installed bool `json:"installed"`
+
+	// Provisioning contains settings used only for initial cluster provisioning.
+	// May be unset in the case of adopted clusters.
+	Provisioning *Provisioning `json:"provisioning"`
```
staebler: This should be omitempty.

dgoodwin:
Updated.

dgoodwin:
Forcing as we can't get past the appsre job that shouldn't be running on this branch.

PR #661, opened 2019-11-08
Added ssh-public-key secret creation

mvazquezc:
In the docs is not specified how to generate the secret for the ssh-publickey.

I'm not sure if this is automatically created when using hiveutil, if that's the case let me know and I'll add a note like the one used for AWS credentials.

Thanks,

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -11,7 +11,7 @@
         ```bash
         oc create secret generic mycluster-pull-secret --from-file=.dockerconfigjson=/path/to/pull-secret --type=kubernetes.io/dockerconfigjson
         ```
-  1. Create a Kubenertes secret containing a ssh key pair (typically generated with `ssh-keygen`)
+  1. Create a Kubernetes secret containing a ssh key pair (typically generated with `ssh-keygen`)
         ```bash
         oc create secret generic mycluster-ssh-key --from-file=ssh-privatekey=/path/to/private/key --from-file=ssh-publickey=/path/to/public/key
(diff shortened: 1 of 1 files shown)


docs/using-hive.md
```diff
@@ -11,6 +11,10 @@
         ```bash
         oc create secret generic mycluster-pull-secret --from-file=.dockerconfigjson=/path/to/pull-secret --type=kubernetes.io/dockerconfigjson
         ```
+  1. Create a Kubenertes secret containing a ssh public key (typically generated with `ssh-keygen`)
```
dgoodwin: s/Kubenertes/Kubernetes

docs/using-hive.md
```diff
@@ -11,6 +11,10 @@
...
+        ```bash
+        oc create secret generic mycluster-ssh-key --from-file=ssh-publickey=/path/to/public/key
+        ```
```
dgoodwin: Could you also add --from-file=ssh-privatekey=/path/to/private/key. Only specifying public should configure the nodes so you can ssh to them, but if you also specify private you gain automatic fetching of logs from the nodes in the event of an install failure.

Could you also note that this step is optional, and can be done automatically with hiveutil create-cluster via the --ssh-public-key-file and --ssh-private-key-file arguments.

mvazquezc: Added a new commit with your suggestions.

dgoodwin:
Perfect, thank you.

/lgtm

PR #662, opened 2019-11-08
move cluster metadata from status to spec in clusterdeployment

staebler:
Also, removed some legacy components that are not relevant in v1.
 * A restored clusterdeployment does not need to get its cluster metadata from the clusterprovision since the cluster metadata in the spec will be restored with the clusterdeployment.
 * A dummy clusterprovision does not need to be created for a legacy clusterdeployment since the clusterprovision is no longer needed to restore cluster metadata.
 * The installed field in the clusterdeployment status is removed as it was deprecated already in v1alpha1.

https://jira.coreos.com/browse/CO-617


staebler:
Passes `make verify`, `make lint`, and `make test`.

staebler:
https://jira.coreos.com/browse/CO-617

pkg/apis/hive/v1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -421,9 +422,24 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateUpdate(admissionSpec
...
 
-	if oldObject.Spec.Installed && !newObject.Spec.Installed {
-		allErrs := field.ErrorList{}
-		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "installed"), newObject.Spec.Installed, "cannot make uninstalled once installed"))
+	allErrs := field.ErrorList{}
+	specPath := field.NewPath("spec")
+
+	if newObject.Spec.Installed {
+		if newObject.Spec.ClusterMetadata != nil {
+			if oldObject.Spec.Installed {
```
dgoodwin: This might be worth a comment, I stared at this for a couple minutes thinking something was wrong, it wasn't, but it's a little unclear that this is ensuring clustermetadata is immutable once set.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -384,33 +383,9 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
-		case cd.Status.InfraID != "":
-			// TODO: Remove the code for creating a dummy clusterprovision once
-			// all legacy clusterdeployments have had clusterprovisions created
-			// for them.
-			cdLog.Info("cluster was created prior to clusterprovisions, creating dummy clusterprovision")
-			return r.createClusterProvisionForLegacyCD(cd, cdLog)
-		default:
-			cdLog.Warn("cluster is installed but does not have an infra ID or a clusterprovision")
-			return reconcile.Result{}, nil
-		}
```
dgoodwin: This is a nice little bonus.

pkg/apis/hive/v1/validating-webhooks/clusterdeployment_validating_admission_hook.go
staebler: A comment sounds like a great idea.

PR #664, opened 2019-11-11
allow user to add manifests to be installed

staebler:
Adds ManifestsConfigMap to the clusterdeployment. This points to a configmap containing manifests that the user wants the installer to include.

Adds the --manifests option to hiveutl create-cluster. This specifies a directory containing manifests. Each file in that directory will be added to the manifests configmap.

https://jira.coreos.com/browse/CO-616


staebler:
/hold

This PR is built on top of #660.

pkg/installmanager/installmanager.go
```diff
@@ -513,6 +516,18 @@ func (m *InstallManager) generateAssets(provision *hivev1.ClusterProvision) erro
 		}
 	}
 
+	if src := m.ManifestsMountPath; isDirNonEmpty(src) {
+		m.log.Info("copying user-provided manifests")
+		dest := filepath.Join(m.WorkDir, "manifests")
+		out, err := exec.Command("bash", "-c", fmt.Sprintf("cp %s %s", filepath.Join(src, "*"), dest)).CombinedOutput()
```
dgoodwin: Slight different approach to the other places where we copy in installmanager, any reason to prefer one vs the other?

pkg/installmanager/installmanager.go
```diff
@@ -513,6 +516,18 @@ func (m *InstallManager) generateAssets(provision *hivev1.ClusterProvision) erro
...
+		fmt.Printf("%s\n", out)
```
dgoodwin: intentional print? Looks like it might be. Should it go through the logger though?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -547,13 +547,6 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		},
 	}
 
-	if v, ok := cd.Annotations[constants.InstallFailureTestAnnotation]; ok {
-		if provision.Annotations == nil {
-			provision.Annotations = map[string]string{}
-		}
-		provision.Annotations[constants.InstallFailureTestAnnotation] = v
-	}
```
dgoodwin: Glad you caught this as well.

pkg/installmanager/installmanager.go
staebler: Yes. This was intentional to get the error output if the copy failed. I did not run it through the logger because it may be multiline, which does not output well.

pkg/installmanager/installmanager.go
staebler: I did not see any other places in installmanager where all files in a directory are copied. Can you point out where that is done?

dgoodwin: https://github.com/openshift/hive/blob/v1/pkg/installmanager/installmanager.go#L237

staebler: That copies a single file with a well-known name. In the case of manifests, the installmanager does not know the names of the files to copy. It wants to copy all of the files in the directory.

PR #667, opened 2019-11-12
docs: hiveadmission-dev-cert.sh request cfssl and cfssljson.

gyliu513:
/cc @dgoodwin


hack/hiveadmission-dev-cert.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
gyliu513: @dgoodwin `pushd` request `bash.`

dgoodwin:
/lgtm

/ok-to-test

Thanks!

PR #668, opened 2019-11-12
types: change names of references to include the resource type

staebler:
* SSHKey moved under Provisioning. Limited to only the private key. Renamed to SSHPrivateKeySecret.
* Credentials under PlatformSecrets moved to Platform. Renamed to CredentialsSecret.
* FederatedClusterRef renamed to FederatedCluster.
* SecretRef under CertificateBundleSpec renamed to CertificateSecret.

https://jira.coreos.com/browse/CO-618


staebler:
I don't think that these changes are right. According to the kubernetes API conventions [1], the names of the reference fields should end in "Ref". For example, `CredentialsSecret` should be `CredentialsSecretRef`.

staebler:
Closing this as the tooling is confused due to initially having the target branch as master instead of v1.

/close

PR #669, opened 2019-11-12
types: change names of references to include the resource type

staebler:
* SSHKey moved under Provisioning. Limited to only the private key. Renamed to SSHPrivateKeySecret.
* Credentials under PlatformSecrets moved to Platform. Renamed to CredentialsSecret.
* FederatedClusterRef renamed to FederatedCluster.
* SecretRef under CertificateBundleSpec renamed to CertificateSecret.

https://jira.coreos.com/browse/CO-618


staebler:
/hold

This is built on top of #660.

staebler:
I don't think that these changes are right. According to the kubernetes API conventions [1], the names of the reference fields should end in "Ref". For example, CredentialsSecret should be CredentialsSecretRef.

[1] https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -122,6 +112,14 @@ type Provisioning struct {
...
 	// version for the release you are provisioning.
 	InstallConfigSecret corev1.LocalObjectReference `json:"installConfigSecret"`
+
+	// SSHPrivateKeySecret is the reference to the secret that contains the private SSH key to use
+	// for access to compute instances. This private key should correspond to the public key included
+	// in the InstallConfig. The private key is used by Hive to gather logs on the target cluster if
+	// there are install failures.
+	// The SSH private key is expected to be in the secret data under the "ssh-privatekey" key.
+	// +optional
+	SSHPrivateKeySecret *corev1.LocalObjectReference `json:"sshPrivateSecretKey,omitempty"`
```
dgoodwin: Got a little nervous about this change after hearing some of what Eric said in the meetings yesterday. (that we don't need to deal in private keys) I think our plan is ok though, if you don't want to give us a private key, give the InstallConfig the public key, and us nothing?

Or should be backtrack here and keep it as a ssh-key that might contain a public and/or private key key.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -181,7 +170,7 @@ type ClusterDeploymentStatus struct {
 
 	// FederatedClusterRef is the reference to the federated cluster resource associated with
 	// this ClusterDeployment.
-	FederatedClusterRef *corev1.ObjectReference `json:"federatedClusterRef,omitempty"`
+	FederatedCluster *corev1.ObjectReference `json:"federatedCluster,omitempty"`
```
dgoodwin: Ugh we should remove this entirely, even if we keep the code it should be axed from the V1 API until it's useful.

dgoodwin:
Needs conflicts resolved.

pkg/apis/hive/v1/clusterdeployment_types.go
staebler: I'll remove the federation stuff as its own PR.

staebler:
> Needs conflicts resolved.

Conflicts have been resolved now.

PR #681, opened 2019-11-13
Requeue successful ssi for resync.

abutcher:
https://jira.coreos.com/browse/CO-612

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -264,7 +264,29 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 	}
 
 	ssiLog.Info("done reconciling syncsetinstance")
-	return reconcile.Result{}, applyErr
+	if applyErr != nil {
+		return reconcile.Result{}, applyErr
+	}
+
+	reapplyDuration := r.ssiReApplyDuration(ssi)
+	return reconcile.Result{RequeueAfter: reapplyDuration}, nil
+}
+
+// ssiReApplyDuration returns the shortest time.Duration to meet reapplyInterval from successfully applied
+// resources, patches and secretReferences in the SyncSetInstance status.
+func (r *ReconcileSyncSetInstance) ssiReApplyDuration(ssi *hivev1.SyncSetInstance) time.Duration {
+	timeSinceOldestApply := time.Duration(0)
+	for _, statuses := range [][]hivev1.SyncStatus{ssi.Status.Resources, ssi.Status.Patches, ssi.Status.SecretReferences} {
+		for _, status := range statuses {
+			if applySuccessCondition := controllerutils.FindSyncCondition(status.Conditions, hivev1.ApplySuccessSyncCondition); applySuccessCondition != nil {
+				since := time.Since(applySuccessCondition.LastProbeTime.Time)
+				if since > timeSinceOldestApply {
+					timeSinceOldestApply = since
+				}
+			}
+		}
+	}
+	return reapplyInterval - timeSinceOldestApply
 }
 
 func (r *ReconcileSyncSetInstance) getClusterDeployment(ssi *hivev1.SyncSetInstance, ssiLog log.FieldLogger) (*hivev1.ClusterDeployment, error) {
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -268,13 +268,13 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, applyErr
 	}
 
-	reapplyDuration := r.ssiReApplyDuration(ssi)
+	reapplyDuration := ssiReApplyDuration(ssi)
 	return reconcile.Result{RequeueAfter: reapplyDuration}, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -264,7 +264,10 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 	}
 
 	ssiLog.Info("done reconciling syncsetinstance")
-	return reconcile.Result{}, applyErr
+	if applyErr != nil {
+		return reconcile.Result{}, applyErr
+	}
+	return reconcile.Result{RequeueAfter: reapplyInterval}, nil
```
staebler: The requeue-after time should not always be `reapplyInterval`. It should be the amount of time until the next resource in the syncsetinstance needs to be applied.

abutcher: @staebler updated with a function that walks the `applySuccessConditions` and generates the shortest time to meet the regular reapply interval

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -264,7 +264,43 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
...
+	reapplyDuration := r.ssiReApplyDuration(ssi)
+	return reconcile.Result{RequeueAfter: reapplyDuration}, nil
+}
+
+// ssiReApplyDuration returns the shortest time.Duration to meet reapplyInterval from successfully applied
+// resources, patches and secretReferences in the SyncSetInstance status.
+func (r *ReconcileSyncSetInstance) ssiReApplyDuration(ssi *hivev1.SyncSetInstance) time.Duration {
+	requeueAfter := reapplyInterval
+	for _, resourceStatus := range ssi.Status.Resources {
+		if applySuccessCondition := controllerutils.FindSyncCondition(resourceStatus.Conditions, hivev1.ApplySuccessSyncCondition); applySuccessCondition != nil {
```
staebler: Don't duplicate this block of code.

```
for _, statuses := range [][]hivev1.SyncStatus{ssi.Status.Resource, ssi.Status.Patches, ssi.Status.SecretReferences} {
  for _, resourceStatus := range statuses {
    ...
  }
}
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -264,7 +264,43 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
...
+	}
+
+	reapplyDuration := r.ssiReApplyDuration(ssi)
+	return reconcile.Result{RequeueAfter: reapplyDuration}, nil
+}
+
+// ssiReApplyDuration returns the shortest time.Duration to meet reapplyInterval from successfully applied
+// resources, patches and secretReferences in the SyncSetInstance status.
+func (r *ReconcileSyncSetInstance) ssiReApplyDuration(ssi *hivev1.SyncSetInstance) time.Duration {
+	requeueAfter := reapplyInterval
```
staebler: I would find it a little easier to rationalize about if this where `timeSinceOldestApply` instead of `requeueAfter`. That would keep from having to do subtractions in the comparison loop. The return from the `ssiReApplyDuration` would be `reapplyInterval - timeSinceOldestApply`.

abutcher:
Working on adding some tests

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -1503,3 +1503,40 @@ func computeHash(data interface{}) string {
...
+	ssi.Status = successfulResourceStatus(resources...)
+
+	oldestResourceApplyTimes := []time.Time{
+		tenMinutesAgo,
+		fortyFiveMinutesAgo,
+		ninetyMinutesAgo,
+	}
+
+	for _, oldestResourceApplyTime := range oldestResourceApplyTimes {
+		ssi.Status.Resources[0].Conditions[0].LastProbeTime = metav1.NewTime(oldestResourceApplyTime)
```
staebler: This test is only setting the time in the first resource. We do not need to test that multiple different times in the first resource works. We need to test that the right time is picked regardless of which resource has the latest time and regardless of where in the conditions slice the apply-success condition is.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -1503,3 +1503,40 @@ func computeHash(data interface{}) string {
 	}
 	return fmt.Sprintf("%x", md5.Sum(b))
 }
+
+func TestSSIReApplyDuration(t *testing.T) {
+	startProbe := time.Now()
+
+	tenMinutesAgo := time.Unix(metav1.NewTime(time.Now().Add(-10*time.Minute)).Unix(), 0)
```
staebler: The `Unix` cruft is not necessary and will cause flakiness in the test.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
abutcher: Modified this to change the time for each resource in the loop since I had 3 example times.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
abutcher: Removed.

PR #682, opened 2019-11-13
make Provisioning and InstallConfigSecretRef.Name required

gregsheremeta:
make Provisioning and InstallConfigSecretRef.Name required


dgoodwin:
I thought these were optional for the adoption use case.

dgoodwin:
Oh I see its only  for uninstalled clusters.

gregsheremeta:
@dgoodwin should I edit the subject? "... for uninstalled clusters."

dgoodwin:
Nah its ok.

PR #684, opened 2019-11-14
docs: updated e2e test steps.

gyliu513:
/cc @dgoodwin 

@dgoodwin there are also some other problems for the current e2e section, it only mentions how to test via AWS, but not for other platforms, I think the document also needs to be updated to clarify how to run e2e for other platforms, like Azure, GCP etc. Comments?


LalatenduMohanty:
> I think the document also needs to be updated to clarify how to run e2e for other platforms, like Azure, GCP etc. Comments?

Agree. This is something we need to fix.

PR #688, opened 2019-11-14
Fix missing details in Azure install-config and CD.Spec.

dgoodwin:


dgoodwin:
Just waiting on a successful azure deploy but it's at 99% and looking good.

dgoodwin:
Hmm where's our CI...

staebler:
This looks good. I'll wait for e2e-azure test to compete before giving a lgtm.

staebler:
The e2e failure is from the changes that @joelddiaz made for SSH private key secret. https://github.com/openshift/hive/pull/686

PR #695, opened 2019-11-18
Prevent running a ClusterDeprovision job if owning cluster is not del…

dgoodwin:
…eted.

### DIFF
--- a/pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
+++ b/pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
@@ -2,6 +2,8 @@ package clusterdeprovision
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -132,6 +134,30 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
 
 	if instance.Status.Completed {
 		rLog.Debug("clusterdeprovision is complete, skipping")
+		return reconcile.Result{}, nil
+	}
+
+	// Check if there is a ClusterDeployment owning this Deprovision, if so look it up and
+	// make sure it has a deletion timestamp. Otherwise bail out as a safety check.
+	oRef := metav1.GetControllerOf(instance)
+	if oRef == nil {
+		// TODO: this was once supported to cleanup self-managed "preserveOnDelete" clusters,
+		// but the feature was killed off. For now we'd rather not open the door to dangling
+		// ClusterDeprovisions with no associated cluster.
+		rLog.Warn("ClusterDeprovision does not have an owning ClusterDeployment")
+		return reconcile.Result{}, nil
+	}
+	if oRef.Kind != "ClusterDeployment" || !strings.HasPrefix(oRef.APIVersion, "hive.openshift.io") {
+		rLog.Warnf("ClusterDeprovision has a non-ClusterDeployment owner: %v", oRef)
+		return reconcile.Result{}, nil
+	}
+	cd := &hivev1.ClusterDeployment{}
+	if err := r.Get(context.TODO(), types.NamespacedName{Namespace: instance.Namespace, Name: oRef.Name}, cd); err != nil {
+		rLog.Error("error looking up ClusterDeployment that owns ClusterDeprovision")
+		return reconcile.Result{}, fmt.Errorf("error looking up ClusterDeployment that owns ClusterDeprovision")
+	}
+	if cd.DeletionTimestamp == nil {
+		rLog.Error("ClusterDeprovision created for ClusterDeployment that has not been deleted")
 		return reconcile.Result{}, nil
 	}
 
--- a/pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
+++ b/pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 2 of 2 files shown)


dgoodwin:
Adds a check to make sure the ClusterDeployment (should one exist) that owns a ClusterDeprovosion has been deleted. Prevents an unlikely case where someone created a ClusterDeprovision manually.

Not sure if this makes sense, I'm more worried about something mistakenly issuing a ClusterDeployment delete tbh.

dgoodwin:
Also InstallManager also has a guard against running if the ClusterDeployment is set installed, so the provision case should be protected.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,25 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, nil
 	}
 
+	// Check if there is a ClusterDeployment owning this Deprovision, if so look it up and
+	// make sure it has a deletion timestamp. Otherwise bail out as a safety check.
+	for _, oRef := range instance.OwnerReferences {
```
staebler: `GetControllerOf` from "k8s.io/apimachinery/pkg/apis/meta/v1"

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,25 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, nil
 	}
 
+	// Check if there is a ClusterDeployment owning this Deprovision, if so look it up and
+	// make sure it has a deletion timestamp. Otherwise bail out as a safety check.
```
staebler: The controller should also bail out if the deprovision is **not owned** by a ClusterDeployment.

dgoodwin: I wasn't sure where to go with that as technically at one point this was designed and used to be able to cleanup clusters that were provisioned with hive and the preserveOnDelete option, then later deprovision them without there being a cluster deployment. Feature was axed in OCM but might make a reappearance so it was theoretically possible that there could be a detached unassociated Deprovision.

staebler: I'm not keen on being in the business of providing a means to delete a random cluster not controller by Hive. I would rather prevent the deletion now and relax the restriction in the future if we decide to explicitly allow the feature.

dgoodwin: Deal.

staebler: Not thinking about it too much, but I think I would feel better if, to cover the use case described, the user had to create a new clusterdeployment that adopted the existing cluster.

staebler:
@dgoodwin Should this go in master?

dgoodwin:
Pretty grey, but I think v1 is ok, it's not critical to get it out to master and I mostly want it there when we get to migration work.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,32 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
...
+	oRef := metav1.GetControllerOf(instance)
+	if oRef == nil {
+		// TODO: this was once supported to cleanup self-managed "preserveOnDelete" clusters,
+		// but the feature was killed off. For now we'd rather not open the door to dangling
+		// ClusterDeprovisions with no associated cluster.
+		rLog.Warn("ClusterDeprovision does not have an owning ClusterDeployment")
+		return reconcile.Result{}, fmt.Errorf("ClusterDeprovision does not have an owning ClusterDeployment")
```
staebler: This should not return an error. There is no reason to resync the resource.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,32 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
...
+	}
+	if oRef.Kind != "ClusterDeployment" || !strings.HasPrefix(oRef.APIVersion, "hive.openshift.io") {
+		rLog.Warnf("ClusterDeprovision has a non-ClusterDeployment owner: %v", oRef)
+		return reconcile.Result{}, fmt.Errorf("ClusterDeprovision has a non-ClusterDeployment owner: %v", oRef)
```
staebler: Same thing about not returning an error.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,32 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
...
+	}
+	cd := &hivev1.ClusterDeployment{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: instance.Namespace, Name: oRef.Name}, cd)
+	if err != nil {
+		rLog.Error("error looking up ClusterDeployment that owns ClusterDeprovision")
+		return reconcile.Result{}, nil
```
staebler: This **should** return an error.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,32 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
...
+	}
+	if cd.DeletionTimestamp == nil {
+		err = fmt.Errorf("ClusterDeprovision created for ClusterDeployment that has not been deleted")
+		rLog.WithError(err).Error("skipping")
+		return reconcile.Result{}, err
```
staebler: This should not return an error.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,32 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
...
+	}
+	cd := &hivev1.ClusterDeployment{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: instance.Namespace, Name: oRef.Name}, cd)
+	if err != nil {
+		rLog.Error("error looking up ClusterDeployment that owns ClusterDeprovision")
+		return reconcile.Result{}, nil
+	}
+	if cd.DeletionTimestamp == nil {
+		err = fmt.Errorf("ClusterDeprovision created for ClusterDeployment that has not been deleted")
+		rLog.WithError(err).Error("skipping")
```
staebler: I don't see any need to use `WithError` here. Make the message of the log entry the text of what you have as the error.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -135,6 +137,32 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
...
+		// ClusterDeprovisions with no associated cluster.
+		rLog.Warn("ClusterDeprovision does not have an owning ClusterDeployment")
+		return reconcile.Result{}, fmt.Errorf("ClusterDeprovision does not have an owning ClusterDeployment")
+	}
+	if oRef.Kind != "ClusterDeployment" || !strings.HasPrefix(oRef.APIVersion, "hive.openshift.io") {
+		rLog.Warnf("ClusterDeprovision has a non-ClusterDeployment owner: %v", oRef)
+		return reconcile.Result{}, fmt.Errorf("ClusterDeprovision has a non-ClusterDeployment owner: %v", oRef)
+	}
+	cd := &hivev1.ClusterDeployment{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: instance.Namespace, Name: oRef.Name}, cd)
```
staebler: `if err := ...; err != nil {` ?

dgoodwin:
Updated.

PR #697, opened 2019-11-19
Add hiveutil create-cluster support for adopting clusters.

dgoodwin:
--adopt options allow you to adopt an already installed cluster into
Hive for purposes of using SyncSets, and Hive deprovision code.

This feature is likely most helpful for Hive developers who need are not
looking to explicitly test provisioning and would like to quickly get
valid clusters into hive without waiting for a full provision.

Adds documentation for using a kind cluster as your adopted
ClusterDeployment.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -131,6 +131,12 @@ type Options struct {
 	SimulateBootstrapFailure bool
 	WorkerNodes              int64
 	ManifestsDir             string
+	Adopt                    bool
+	AdoptAdminKubeConfig     string
+	AdoptInfraID             string
+	AdoptClusterID           string
+	AdoptAdminUsername       string
+	AdoptAdminPassword       string
 
 	// Azure
 	AzureBaseDomainResourceGroupName string
@@ -210,6 +216,14 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=gcp --gcp-project-id=PROJECT_ID`,
 	flags.Int64Var(&opt.WorkerNodes, "workers", 3, "Number of worker nodes to create.")
 	flags.StringVar(&opt.ManifestsDir, "manifests", "", "Directory containing manifests to add during installation")
 
+	// Flags related to adoption.
+	flags.BoolVar(&opt.Adopt, "adopt", false, "Enable adoption mode for importing a pre-existing cluster into Hive. Will require additional flags for adoption info.")
+	flags.StringVar(&opt.AdoptAdminKubeConfig, "adopt-admin-kubeconfig", "", "Path to a cluster admin kubeconfig file for a cluster being adopted. (required if using --adopt)")
+	flags.StringVar(&opt.AdoptInfraID, "adopt-infra-id", "", "Infrastructure ID for this cluster's cloud provider. (required if using --adopt)")
+	flags.StringVar(&opt.AdoptClusterID, "adopt-cluster-id", "", "Cluster UUID used for telemetry. (required if using --adopt)")
+	flags.StringVar(&opt.AdoptAdminUsername, "adopt-admin-username", "", "Username for cluster web console administrator. (optional)")
+	flags.StringVar(&opt.AdoptAdminPassword, "adopt-admin-password", "", "Password for cluster web console administrator. (optional)")
+
 	// Azure flags
 	flags.StringVar(&opt.AzureBaseDomainResourceGroupName, "azure-base-domain-resource-group-name", "os4-common", "Resource group where the azure DNS zone for the base domain is found")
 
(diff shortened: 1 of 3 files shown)


contrib/pkg/createcluster/create.go
```diff
@@ -210,6 +216,14 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=gcp --gcp-project-id=PROJECT_ID`,
 	flags.Int64Var(&opt.WorkerNodes, "workers", 3, "Number of worker nodes to create.")
 	flags.StringVar(&opt.ManifestsDir, "manifests", "", "Directory containing manifests to add during installation")
 
+	// Flags related to adoption.
+	flags.BoolVar(&opt.Adopt, "adopt", false, "Enable adoption mode for importing a pre-existing cluster into Hive. Will require additional flags for adoption info.")
+	flags.StringVar(&opt.AdoptAdminKubeConfig, "adopt-admin-kubeconfig", "", "Path to a cluster admin kubeconfig file for a cluster being adopted. (required if using --adopt)")
+	flags.StringVar(&opt.AdoptInfraID, "adopt-infra-id", "", "Infrastructure ID for this cluster's cloud provider. (required if using --adopt)")
+	flags.StringVar(&opt.AdoptClusterID, "adopt-cluster-id", "", "Cluster UUID used for telemetry. (required if using --adopt)")
```
staebler: Is ClusterID required? We use it for deprovisioning by tag with the following comment. Is that comment still applicable?
```
// Also cleanup anything with the tag for the legacy cluster ID (credentials still using this for example)
```

contrib/pkg/createcluster/create.go
```diff
@@ -458,6 +499,52 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
...
+	}
+
+	adminKubeconfigSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-adopted-admin-kubeconfig", cd.Name),
+			Namespace: cd.Namespace,
+		},
+		Data: map[string][]byte{
+			"kubeconfig":     kubeconfigBytes,
+			"raw-kubeconfig": kubeconfigBytes,
```
staebler: Where does "raw-kubeconfig" come from?

contrib/pkg/createcluster/create.go
dgoodwin: If you go back to 4.1 I *think* it's still applicable. The globally unique identifier for a cluster felt important enough to keep as required even if we could dance away from it, may prove useful in relating with other systems.

contrib/pkg/createcluster/create.go
dgoodwin: I am honestly not sure what that's doing there, just that it's what our controllers / installmanager uploads today and I wanted to make sure to match it. It appears to be identical to what goes into the other key.

contrib/pkg/createcluster/create.go
staebler: I guess my question is more about whether it is important enough to **require** it for a dev-centric adopt feature. Or can we say a little extra work for a dev when creating an adopted cluster by allowing it to be omitted.

contrib/pkg/createcluster/create.go
staebler: I ask because I don't see the installmanager adding that and I don't see it in the example secrets that I looked at.

dgoodwin: Hmm I'm getting it with my secrets on a cluster created by hive.

contrib/pkg/createcluster/create.go
dgoodwin: My thought was yes, by requiring it we're making whoever is adopting think about it, if it's a dev we know we can fake it, anyone else using this would need to make that determination themselves, but I kinda like that this forces them to acknowledge it.

dgoodwin:
Regarding adopt-cluster I considered this, and we might want to do that at some point, it's just a much bigger PR whereas adding it this way slides in pretty smoothly. It seems like we do want most if not all of what create-cluster does, I actually thought we landed last week at InstallConfigs were going to be required for all clusters but I cannot remember why.

contrib/pkg/createcluster/create.go
staebler: I found where this is done now.
https://github.com/openshift/hive/blob/a057c9fcff2fd64fc2c7555b522b3109263f4a12/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L985
I'm still trying to figure out *why* it is done.

contrib/pkg/createcluster/create.go
staebler: Makes sense.

staebler:
> Regarding adopt-cluster I considered this, and we might want to do that at some point, it's just a much bigger PR whereas adding it this way slides in pretty smoothly. It seems like we do want most if not all of what create-cluster does, I actually thought we landed last week at InstallConfigs were going to be required for all clusters but I cannot remember why.

I think you are right about requiring installconfigs. This all seems reasonable.

contrib/pkg/createcluster/create.go
staebler: The raw one is the original created by the installer. The other is the one fix the additional CAs added in.

dgoodwin: Ohhhh right, that's it. I remember now.

PR #698, opened 2019-11-19
Add a gcp actuator to the dnszone controller

twiest:
- [x] Add a gcp actuator to the dnszone controller
- [x] Manually test that the gcp actuator is working
- [x] Add Unit tests for the gcp actuator


twiest:
@staebler Still working on the manual tests, but I think this is ready for a first pass code review.

pkg/apis/hive/v1alpha1/dnszone_types.go
```diff
@@ -50,6 +54,16 @@ type AWSResourceTag struct {
...
 }
 
+// GCPDNSZoneSpec contains GCP-specific DNSZone specifications
+type GCPDNSZoneSpec struct {
+	// Credentials references a secret that will be used to authenticate with
+	// GCP CloudDNS. It will need permission to create and manage CloudDNS Hosted Zones.
+	// Secret should have a key named 'osServiceAccount.json'.
+	// The credentials must specify the project to use.
+	// +optional
+	Credentials corev1.LocalObjectReference `json:"credentials,omitempty"`
```
twiest: I want to be consistent here, but we have 2 patterns now.

Above in the `AWSDNSZoneSpec`, it references the secret as `AccountSecret`.

But in the [`PlatformSecrets`](https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1alpha1/gcp/platform.go#L27) and [`ExternalDNSGCPConfig`](https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1alpha1/hiveconfig_types.go#L117) it refers to them as `Credentials`.

I followed the latter pattern here, but is that the correct decision?

Should we fix [`AWSDNSZoneSpec`](https://github.com/openshift/hive/blob/v1/pkg/apis/hive/v1/dnszone_types.go#L31-L42) in v1 to be consistent as well?

Also, not sure we need `Region` in there since Route53 is regionless. Should we remove that for v1 too?

To be clear, I'm not suggesting making those changes in this PR.

@staebler @dgoodwin thoughts?

dgoodwin: +1 for fixing AWS to be consistent with Credentials.

dgoodwin: Also this should probably be CredentialsSecretRef in both cases per what we did most recently in v1.

twiest: @dgoodwin do we want it to be `CredentialsSecretRef` even in the v1alpha1 version?

twiest: I filed [this bug](https://jira.coreos.com/browse/CO-635) to fix `AWSDNSZoneSpec` in the v1 api.

dgoodwin: I would yeah, we're inconsistent in v1alpha1 anyhow and it's one less thing we need to change after this merges into v1.

twiest: Done.

pkg/apis/hive/v1alpha1/dnszone_types.go
```diff
@@ -50,6 +54,16 @@ type AWSResourceTag struct {
 	Value string `json:"value"`
 }
 
+// GCPDNSZoneSpec contains GCP-specific DNSZone specifications
+type GCPDNSZoneSpec struct {
+	// CredentialsSecretRef references a secret that will be used to authenticate with
+	// GCP CloudDNS. It will need permission to create and manage CloudDNS Hosted Zones.
+	// Secret should have a key named 'osServiceAccount.json'.
+	// The credentials must specify the project to use.
+	// +optional
```
staebler: This is not optional, is it?

pkg/apis/hive/v1alpha1/dnszone_types.go
```diff
@@ -50,6 +54,16 @@ type AWSResourceTag struct {
...
+	CredentialsSecretRef corev1.LocalObjectReference `json:"credentialsSecretRef,omitempty"`
```
staebler: Assuming that this is not actually optional, this should not be `omitempty`.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -2,7 +2,9 @@ package dnszone
 
 import (
 	"context"
+	"errors"
 	"fmt"
+
```
staebler: Remove this blank line.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -2,7 +2,9 @@ package dnszone
 
 import (
 	"context"
+	"errors"
```
staebler: I would prefer that we use "github.com/pkg/errors" as it includes some more functionality than the stock "error" package.

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -0,0 +1,176 @@
+package dnszone
+
+import (
+	"errors"
```
staebler: "github.com/pkg/errors"?

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Delete implements the Delete call of the actuator interface
+func (a *GCPActuator) Delete() error {
+	if a.managedZone == nil {
+		return errors.New("managedZone is unpopulated")
+	}
+
+	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone).WithField("zoneName", a.managedZone.Name)
+	logger.Info("Deleting managedzone")
```
staebler: "managed zone" is used in the rest of the log entries.

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -0,0 +1,176 @@
...
+
+// GetNameServers implements the GetNameServers call of the actuator interface
+func (a *GCPActuator) GetNameServers() ([]string, error) {
+	if a.managedZone == nil {
+		return nil, errors.New("managedZone is unpopulated")
+	}
+
+	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone)
+	result := a.managedZone.NameServers
+	logger.WithField("nameservers", strings.Join(result, ",")).Debug("found managed zone name servers")
```
staebler: I don't think that we need to generate a string for the nameservers field. If we use `result` directly as the value of the field, then the field will have all of the name servers included. This would (1) make the code simpler and (2) avoid an unnecessary string formatting when not running the logger at debug level.

pkg/constants/constants.go
```diff
@@ -59,6 +59,9 @@ const (
 	// requesting that their domains be managed must have a base domain
 	// that is a direct child of one of the valid domains.
 	ManagedDomainsFileEnvVar = "MANAGED_DOMAINS_FILE"
+
+	// GCPCredentialsName is the name of the GCP credentials file or secret key.
+	GCPCredentialsName = "osServiceAccount.json"
```
staebler: +1 to moving this into the global constants.

mjudeikis:
Minor updates to https://github.com/openshift/hive/blob/master/docs/using-hive.md would be good :)
Testing now

mjudeikis:
Diff to "hack it" to working state:
https://gist.github.com/mjudeikis/134ad1d029fb54012a75ab01055e7936
@staebler PR to fix "dots" https://github.com/openshift/hive/pull/700

twiest:
> Diff to "hack it" to working state:
> https://gist.github.com/mjudeikis/134ad1d029fb54012a75ab01055e7936
> @staebler PR to fix "dots" #700

@mjudeikis Hey, thanks for testing this out and for working on the above Gist. Would you mind submitting a PR for the above gist once this PR has merged?

twiest:
@staebler ok, tests are in. I copied the tests from AWS and converted them to GCP. So we have the exact same code coverage for both actuators now.

As we discussed last week, in a future PR I'll implement the "FakeActuator" pattern so that we don't have to implement the same set of tests for each cloud.

I think this is ready to go.

PR #699, opened 2019-11-19
Update SelectorSyncSet label selector doc.

abutcher:


staebler:
> /lgtm
> 
> Was this the issue Nimrod was hitting?

Yes

PR #702, opened 2019-11-21
 merge master into v1

staebler:
includes changes in (105cae8, a057c9f]


staebler:
I am looking to minimize the number of commits included in the massively-conflicting merge that will happen when we merge https://github.com/openshift/hive/pull/694 into v1.

PR #704, opened 2019-11-21
merge master into v1

staebler:
includes changes in (a057c9f, 6231266]


staebler:
I want to bring #701 to v1 so that I can use it for work to remove the DNSEndpoint type.

PR #706, opened 2019-11-22
types: remove dnsendpoint resource

staebler:
The link to the parent domain will be done with the DNSZone resource instead of creating a DNSEndpoint resource.


pkg/operator/hive/hive.go
```diff
@@ -149,7 +149,6 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		"config/crds/hive_v1_clusterdeployment.yaml",
 		"config/crds/hive_v1_clusterdeprovision.yaml",
 		"config/crds/hive_v1_clusterimageset.yaml",
-		"config/crds/hive_v1_dnsendpoint.yaml",
```
dgoodwin: Seems like the operator should cleanup this CRD if it exists.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -2,12 +2,14 @@ package dnsendpoint
 
 import (
 	"context"
+	"fmt"
```
dgoodwin: Is there a better name for this controller now?

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -158,61 +161,60 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
...
+		if err := r.nameServerQuery.Delete(rootDomain, domain, currentNameServers); err != nil {
+			dnsLog.WithError(err).Error("error deleting NS record")
+			return reconcile.Result{}, err
+		}
+		r.nameServerScraper.RemoveEndpoint(domain)
+	}
+
+	status := corev1.ConditionFalse
+	reason := "ParentLinkNotCreated"
+	message := "Parent link has not bee created"
```
dgoodwin: been

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -158,61 +161,60 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
...
-	if err := r.nameServerQuery.Delete(rootDomain, domain, currentNameServers); err != nil {
-		dnsLog.WithError(err).Error("error deleting NS record")
-		return err
-	}
-	r.nameServerScraper.RemoveEndpoint(domain)
-	controllerutils.DeleteFinalizer(instance, hivev1.FinalizerDNSEndpoint)
-	if err := r.Update(context.Background(), instance); err != nil {
-		dnsLog.WithError(err).Log(controllerutils.LogLevel(err), "error deleting finalizer")
-		return err
+	switch {
```
dgoodwin: I find this switch a little bit hard to follow, I think because we go from two clauses inclusing !isDeleted and then switch to deleting for the default, looking at it I guess that's because we either have a deletionTimestamp or we have no desiredNameServers. A comment clearing this up might help. 

Also I can't immediately see what the situation is where we'd have no desired name servers, a comment there might be helpful as well.

pkg/apis/hive/v1/dnsendpoint_types.go
```diff
@@ -1,82 +0,0 @@
...
-	// RecordType type of record, e.g. CNAME, A, SRV, TXT etc
-	RecordType string `json:"recordType,omitempty"`
-	// TTL for the record
-	RecordTTL TTL `json:"recordTTL,omitempty"`
-	// Labels stores labels defined for the Endpoint
-	// +optional
-	Labels Labels `json:"labels,omitempty"`
-	// ProviderSpecific stores provider specific config
-	// +optional
-	ProviderSpecific ProviderSpecific `json:"providerSpecific,omitempty"`
```
dgoodwin: This data doesn't seem to have been moved to DNSZone, I assume it just wasn't necessary for us to track?

pkg/controller/dnsendpoint/dnsendpoint_controller.go
staebler: Yes, there is a better name. It may even be better to just include this functionality in the DNSZone controller. However, I wanted to limit the changes for this PR. I do plan on, at least, renamed this controller in future work.

pkg/operator/hive/hive.go
staebler: OK

pkg/controller/dnsendpoint/dnsendpoint_controller.go
staebler: The scenario where there would be no desired name servers is if the hosted zone was deleted or lost for some reason and maybe could not be recreated. The DNSZone controller would clear the name servers from the status of the DNSZone resource. The DNSEndpoint controller at that point should delete the NS records since they are no longer relevant.

pkg/apis/hive/v1/dnsendpoint_types.go
staebler: Correct. And that is my main motivation for removing the DNSEndpoint CRD. It is filled with fields that Hive does not care about--or that will always be the same value for Hive. After removing all of the unnecessary fields from the DNSEndpoint CRD, we would be left with just the `DNSName` and `Targets` fields, which are both already present in the the DNSZone CRD.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
dgoodwin: Would still like to see a comment here just clarifying what's going on.

pkg/operator/hive/hive.go
dgoodwin: Looks like no update yet, and conflicts now in place.

staebler: Yes, no updates yet. This PS has not been a high priority for me. I'll get back to it soon.

staebler:
@dgoodwin I have updated this PR to
1. include the requested comments and
2. delete the DNSEndpoint CRD.

I had been putting off working on this PR because it had not been passing the e2e tests and I did not have the time to track down what was wrong. After I rebased, it did pass the e2e tests. But I am not convinced that the pass was not a fluke.

dgoodwin:
More successes, feeling ok to let this one go through?

staebler:
> More successes, feeling ok to let this one go through?

Yes. I had been waiting for a e2e-gcp pass, too. I have that now as well.

PR #708, opened 2019-11-22
do not allow more than one compute machine pool on GCP

joelddiaz:
Need to figure out how we handle extra compute machine pools. Either reuse existing subnets, or handle creating of new subnets as needed.

For now simply only allow the single original machine pool.

### DIFF
--- a/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
+++ b/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
@@ -607,8 +607,26 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
 
 	vErrs = append(vErrs, validateMachinePool(cd, &cd.Spec.ControlPlane, field.NewPath("spec", "controlPlane"))...)
 
+	vErrs = append(vErrs, validateComputeMachinePoolLength(cd)...)
+
 	for i, mp := range cd.Spec.Compute {
 		vErrs = append(vErrs, validateMachinePool(cd, &mp, field.NewPath("spec", "compute").Index(i))...)
+	}
+
+	return vErrs
+}
+
+func validateComputeMachinePoolLength(cd *hivev1.ClusterDeployment) field.ErrorList {
+	vErrs := field.ErrorList{}
+
+	if cd.Spec.Platform.GCP != nil {
+		// Only restrict the length of compute machine pools on GCP.
+		// Restriction can be removed once proper handling of reuse or creation
+		// of necessary subnets is resolved.
+		if len(cd.Spec.Compute) > 1 {
+			path := field.NewPath("spec")
+			vErrs = append(vErrs, field.Invalid(path, cd.Spec.Compute, "length of compute machine pools must not be greather than 1"))
+		}
 	}
 
 	return vErrs
--- a/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
+++ b/pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
@@ -607,7 +607,7 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
 
 	vErrs = append(vErrs, validateMachinePool(cd, &cd.Spec.ControlPlane, field.NewPath("spec", "controlPlane"))...)
 
-	vErrs = append(vErrs, validateComputeMachinePoolLength(cd)...)
+	vErrs = append(vErrs, validateComputeMachinePoolRequirements(cd)...)
 
 	for i, mp := range cd.Spec.Compute {
 		vErrs = append(vErrs, validateMachinePool(cd, &mp, field.NewPath("spec", "compute").Index(i))...)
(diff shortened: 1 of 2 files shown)


pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,31 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
...
 	}
 
 	return vErrs
 }
 
+func validateComputeMachinePoolLength(cd *hivev1.ClusterDeployment) field.ErrorList {
+	vErrs := field.ErrorList{}
+
+	if cd.Spec.Platform.GCP != nil {
+		// Only restrict the length of compute machine pools on GCP.
```
staebler: Don't we also need to restrict the name of the machine pool to "worker"?

joelddiaz: yes, it does appear that the installer forces us to use the name "worker"

joelddiaz:
@staebler ok, we're now enforcing the name 'worker' and checking for reuse of names

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,57 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
...
 
+func validateComputeMachinePoolRequirements(cd *hivev1.ClusterDeployment) field.ErrorList {
+	vErrs := field.ErrorList{}
+	path := field.NewPath("spec")
+
+	if cd.Spec.Platform.GCP != nil {
+		// Only restrict the length of compute machine pools on GCP.
+		// Restriction can be removed once proper handling of reuse or creation
+		// of necessary subnets is resolved.
+		if len(cd.Spec.Compute) > 1 {
```
staebler: Collapse both of these if blocks into a simpler validation that the name is "worker".

```
for i, mp := range cd.Spec.Compute {
  if mp.Name != "worker" {
    vErrs = append(vErrs, field.Invalid(computePath.Index(i).Child("name"), mp.Name, "name of the compute pool must be \"worker\"")
  }
}
```

If we only allow "worker" pools and the names must be unique, then there cannot be more than one pool.

If there can only be one pool and the only valid name is "worker", then if we have more than zero pools, then there must be a pool named "worker".

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,57 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
...
+
+	// ensure machinePool names are unique
+	names := map[string]int{}
+	for _, mp := range cd.Spec.Compute {
+		names[mp.Name]++
+	}
+	for k, v := range names {
+		if v > 1 {
+			validationMsg := fmt.Sprintf("compute pool with name %s used more than once", k)
+			vErrs = append(vErrs, field.Invalid(path, cd.Spec.Compute, validationMsg))
```
staebler: This is not using the right path.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,57 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
...
+			// make sure we have a compute pool named 'worker'
+			foundWorkerPool := false
+			for _, mp := range cd.Spec.Compute {
+				if mp.Name == "worker" {
+					foundWorkerPool = true
+					break
+				}
+			}
+			if !foundWorkerPool {
+				vErrs = append(vErrs, field.Invalid(path, cd.Spec.Compute, "must have one compute pool named 'worker'"))
```
staebler: This is not using the right field path.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,57 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
...
+	vErrs = append(vErrs, validateComputeMachinePoolRequirements(cd)...)
+
 	for i, mp := range cd.Spec.Compute {
 		vErrs = append(vErrs, validateMachinePool(cd, &mp, field.NewPath("spec", "compute").Index(i))...)
 	}
 
 	return vErrs
 }
 
+func validateComputeMachinePoolRequirements(cd *hivev1.ClusterDeployment) field.ErrorList {
```
staebler: Pass in the field path for the compute field to the function.

staebler: Pass in the slice of machine pools rather than the full clusterDeployment.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,57 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
...
+				}
+			}
+			if !foundWorkerPool {
+				vErrs = append(vErrs, field.Invalid(path, cd.Spec.Compute, "must have one compute pool named 'worker'"))
+			}
+		}
+	}
+
+	// ensure machinePool names are unique
+	names := map[string]int{}
```
staebler:
```
names := map[string]bool{}
for i, mp := range cd.Spec.Compute {
  if names[mp.Name] {
    vErrs = append(vErrs, field.Invalid(path.Index(i).Child("name"), mp.Name, "name must be unique")
  }
  names[mp.Name] = true
}
```

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
joelddiaz: i'll lose the platform info. while GCP happens to require MachinePool.Platform to be populated today (b/c we need the instance type) this may not always be the case

staebler: Fair enough. Although I think it should **always** be the case that the Platform must be present matching the platform from the clusterDeployment. In fact, that should be validated by the webhook, too.

pkg/apis/hive/v1alpha1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -607,13 +607,42 @@ func validateMachinePools(cd *hivev1.ClusterDeployment) field.ErrorList {
 
 	vErrs = append(vErrs, validateMachinePool(cd, &cd.Spec.ControlPlane, field.NewPath("spec", "controlPlane"))...)
 
+	vErrs = append(vErrs, validateComputeMachinePoolRequirements(cd, field.NewPath("spec", "compute"))...)
```
staebler: Since the entire clusterDeployment is being passed in, then the field path should actually be omitted. Sorry to steer you wrong there.

PR #709, opened 2019-11-22
Document deploying a temp prometheus to view hive metrics.

dgoodwin:


config/prometheus/prometheus-configmap.yaml
```diff
@@ -0,0 +1,20 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: prometheus-config
+  namespace: hive
+data:
+  prometheus.yml: |+
+    global:
+      scrape_interval:     15s
```
LalatenduMohanty: The config I used had `evaluation_interval: 15s` , not sure if that will be something useful.

docs/developing.md
```diff
@@ -306,8 +311,17 @@ Run the Hive e2e script:
...
 
-### TIP
+## Viewing Metrics with Prometheus
 
-* The Dep cache located under *_$GOPATH/pkg/dep_*.
-* If you see any Dep errors during `make vendor`, you can remove local cached directory and try again.
+Hive publishes a number of metrics that can be scraped by prometheus. If you do not have an in-cluster prometheus that can scrape hive's endpoint, you can deploy a stateless prometheus pod in the hive namespace with:
+
+```
+kubectl apply -f config/prometheus
```
LalatenduMohanty: s/kubectl/oc/

LalatenduMohanty: As we have used `oc` in our documentation.

LalatenduMohanty:
tested this on an OpenShift cluster and it works fine.

dgoodwin:
Updated.

config/prometheus/prometheus-configmap.yaml
dgoodwin: # How frequently to evaluate rules.
  [ evaluation_interval: <duration> | default = 1m ]

That presumably is for alerting rules? I don't think devs will even need this, we should probably just keep it bare minimum for simple prom testing.

PR #710, opened 2019-11-22
enable managedDNS for GCP

staebler:


staebler:
I need to use managed DNS for the e2e gcp tests, too. I can do that as part of this PR, or I can do it as a follow-up PR.

staebler:
> I need to use managed DNS for the e2e gcp tests, too. I can do that as part of this PR, or I can do it as a follow-up PR.

I have added it as part of this PR.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -117,23 +120,39 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+		hc.Spec.ExternalDNS = &hivev1.ExternalDNSConfig{
+			GCP: &hivev1.ExternalDNSGCPConfig{
+				Credentials: corev1.LocalObjectReference{Name: manageDNSCredentialsSecret},
+			},
+		}
+	default:
+		log.WithField("cloud", o.Cloud).Fatal("unsupported cloud")
 	}
 
+	log.Infof("created Route53 credentials secret: %s", credsSecret.Name)
```
dgoodwin: -Route53

dgoodwin:
Could you also update the using-hive.md docs as this mentions not being available for GCP and is currently Route53 specific.

PR #713, opened 2019-11-25
V1 DNS Field Updates

dgoodwin:
Some small API adjustments to for consistency in V1.


pkg/controller/dnszone/awsactuator.go
```diff
@@ -51,14 +54,9 @@ func NewAWSActuator(
...
-	var regionName string
-
-	if dnsZone != nil && dnsZone.Spec.AWS != nil {
-		regionName = dnsZone.Spec.AWS.Region
-	}
-
-	awsClient, err := awsClientBuilder(secret, regionName)
+	// Route53 is a regionless service, we specify a default region just for the purpose of creating
+	// our client configuration.
+	awsClient, err := awsClientBuilder(secret, defaultRegionEndpoint)
```
dgoodwin: Here be dragons. Is this what we expect if we remove the region?

staebler: That is what the DNSEndpoint code does now. I don't know if this will work for special regions like gov-cloud, though.

dgoodwin: Ok good, we can cross that bridge if we come to it, probably via a HiveConfig option.

PR #716, opened 2019-11-27
hive-frontend ClusterRoleBinding fix

maorfr:
attempts to fix #715

### DIFF
--- a/config/rbac/hive_frontend_role_binding.yaml
+++ b/config/rbac/hive_frontend_role_binding.yaml
@@ -7,6 +7,8 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: hive-frontend
+groupNames:
+- hive-frontend
 subjects:
 - kind: ServiceAccount
   name: hive-frontend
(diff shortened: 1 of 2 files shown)


config/rbac/hive_frontend_role_binding.yaml
```diff
@@ -7,6 +7,8 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: hive-frontend
+groupNames:
```
staebler: This is not a valid field for ClusterRoleBinding in rbac.authorization.k8s.io/v1.

staebler: This is a field from ClusterRoleBinding in authorization.openshift.io/v1, which is not what is being used here. Even in that, the field is legacy and has been superseded by the use of a Group subject, as was done in #715.

dgoodwin:
Just as an FYI there's some really funky stuff going on with the auth types where they're aliased and not separate types. I can't remember all the details but there's some weird magic going on behind the scenes.

maorfr:
i'll look into this again tomorrow morning. thanks!

config/rbac/hive_frontend_role_binding.yaml
maorfr: submitted a fix in #719

PR #719, opened 2019-11-28
fix hive-frontend ClusterRoleBinding

maorfr:
fixes #716


maorfr:
note that even after #715 was merged, the ClusterRoleBinding was not updated to include the group

PR #720, opened 2019-11-28
Add SyncSet metrics.

dgoodwin:
Adds four new metrics, and SyncSetInstance.Status.Applied for a simpler
API to determine if everything in the SyncSet is applying properly.
Currently you must interpret at least 4 conditions, some of which are
errors and one of which is success. The new field is displayed as a
column in kubectl get output.

hive_selectorsyncset_clusters_total is a gauge labeled by name showing
the number of clusters that matched the label selector.
hive_selectorsyncset_clusters_unapplied_total is similar but for the
number that are not currently applying. Both are calculated by
examining SyncSetInstances.

hive_syncsets_total is a gauge that shows the total number of SyncSets,
and hive_syncsets_unapplied_total the number that are not currently
applying. This is also calculated by examining SyncSetInstances.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ docker-build:
 
 # Build the docker image
 .PHONY: docker-dev-push
-docker-dev-push: build
+docker-dev-push: manifests generate build
 	$(DOCKER_CMD) build -t ${IMG} -f Dockerfile.dev .
 	$(DOCKER_CMD) push ${IMG}
 
--- a/pkg/apis/hive/v1alpha1/syncsetinstance_types.go
+++ b/pkg/apis/hive/v1alpha1/syncsetinstance_types.go
@@ -82,6 +82,9 @@ type SyncSetInstanceStatus struct {
 	// when a resource type cannot be determined from a SyncSet resource.
 	// +optional
 	Conditions []SyncCondition `json:"conditions,omitempty"`
+
+	// Applied will be true if all resources, patches, or secrets have successfully been applied on last attempt.
+	Applied bool `json:"applied"`
 }
 
 // +genclient
@@ -90,6 +93,7 @@ type SyncSetInstanceStatus struct {
 // SyncSetInstance is the Schema for the syncsetinstances API
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Applied",type="boolean",JSONPath=".status.applied"
 // +kubebuilder:resource:path=syncsetinstances,shortName=ssi
 type SyncSetInstance struct {
 	metav1.TypeMeta   `json:",inline"`
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -61,6 +61,22 @@ var (
 		Name: "hive_imageset_jobs",
 		Help: "Total number of imageset jobs running by cluster type and state.",
 	}, []string{"cluster_type", "state"})
+	metricSelectorSyncSetClustersTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_selectorsyncset_clusters_total",
+		Help: "Total number of SyncSetInstances that match the label selector for each SelectorSyncSet.",
+	}, []string{"name"})
+	metricSelectorSyncSetClustersUnappliedTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_selectorsyncset_clusters_unapplied_total",
+		Help: "Total number of SyncSetInstances that match the label selector for each SelectorSyncSet but have not successfully applied all resources/patches/secrets.",
+	}, []string{"name"})
+	metricSyncSetsTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "hive_syncsets_total",
+		Help: "Total number of SyncSetInstances referencing non-selector SyncSets.",
+	})
+	metricSyncSetsUnappliedTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "hive_syncsets_unapplied_total",
+		Help: "Total number of SyncSetsInstances referencing non-selector SyncSets that have not successfully applied all resources/patches/secrets.",
+	})
 
 	// MetricClusterDeploymentProvisionUnderwaySeconds is a prometheus metric for the number of seconds
 	// between when a still provisioning cluster was created and now.
@@ -103,6 +119,10 @@ func init() {
 	metrics.Registry.MustRegister(metricInstallJobsTotal)
 	metrics.Registry.MustRegister(metricUninstallJobsTotal)
 	metrics.Registry.MustRegister(metricImagesetJobsTotal)
+	metrics.Registry.MustRegister(metricSelectorSyncSetClustersTotal)
+	metrics.Registry.MustRegister(metricSelectorSyncSetClustersUnappliedTotal)
+	metrics.Registry.MustRegister(metricSyncSetsTotal)
+	metrics.Registry.MustRegister(metricSyncSetsUnappliedTotal)
 	metrics.Registry.MustRegister(MetricControllerReconcileTime)
 
 	metrics.Registry.MustRegister(MetricClusterDeploymentProvisionUnderwaySeconds)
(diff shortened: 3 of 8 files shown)


dgoodwin:
cc @jharrington22, the selectorsyncset metrics are what SRE probably wants to alert on, I don't think Hive uses them internally so a problem with those will almost always indicate a problem coming out of SRE's git repo.

Makefile
```diff
@@ -87,6 +87,8 @@ hiveadmission:
 # Run against the configured Kubernetes cluster in ~/.kube/config
 .PHONY: run
 run: generate fmt vet
+	oc scale -n hive deployment.v1.apps/hive-controllers --replicas=0
```
gyliu513: Why did the makefile need such change? How does this work if I do not have a OCP cluster installed when run makefile?

Also if only scale `hive-controllers` to 0, does the `hive-operator` will create new instance for hive controllers?

dgoodwin: Normal path to install is make deploy, it *might* be possible to just make install the crds and then make run but I don't think we document that path anywhere. In general it should be safe to assume hive-controllers is there.

I don't *think* the hive-operator will fight for scale but I'm not positive, if it reconciled a HiveConfig it might flip it back.

I'm going to remove these changes for now until I'm more sure they should be there.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -189,6 +189,12 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return r.syncDeletedSyncSetInstance(ssi, ssiLog)
 	}
 
+	// Assume applied is false until we successfully apply everything to cluster.
+	ssi.Status.Applied = false
+	original := ssi.DeepCopy()
+	// Always update status if anything has changed
+	defer r.updateSyncSetInstanceStatus(ssi, original, ssiLog)
```
twiest: Scenario:
- A sync has applied a syncsetinstance successfully
- We run this code again, with 0 changes to the syncsetinstance
- The sync this time, though, fails to apply that syncsetinstance

Won't that set `.status.applied` to false, even though this syncsetinstance was successfully sync'd earlier?



Also, what happens if updateSyndSetInstanceStatus errors out?

dgoodwin: Good question, so yes, in this metric right now, it would almost be better worded as "applying", rather than "applied". It's an indicator that this syncset is currently working ok. If nothing has changed but for some reason we can't apply, it's going to start reporting an issue. I think this is probably desirable and ok. We can structure the alerts such that we need it to be failing for some period of time before it triggers I think, items might show up briefly if there's an issue and then clear within a few seconds or minutes on requeue.

Does the behavior seem ok? And if so should I rename to applying or some other term?

twiest: I guess it depends on what we're trying to do. I'm fine with this behavior. Is this something that SRE is asking for?

If so, I think SRE would like to use this in order to know if the syncsetinstance has ever been applied successfully. I think "was the last sync successful" is important, but less meaningful to them (I could be wrong).

Assuming I'm wrong, I'd probably rename this to `status.lastSyncSucceeded` (very wordy) or maybe just `status.syncSucceeded`. If the behavior stays the same, I think `status.applied` is not an accurate name and will likely confuse people.

dgoodwin: I've been raising that all clusters have had broken syncsets for months to them, and offered to get them a little more info on what specific syncsets are busted in the metrics instead of just X clusters have something wrong.

I'm not wondering if maybe there's enough info available for me to add a label indicating if this instance has *ever* synced to the cluster or not. (I think we can safely skip the complexity of determining if the current hash has ever synced successfully, unless that's really easy) Will check with @abutcher

pkg/controller/metrics/metrics.go
```diff
@@ -265,13 +285,61 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+		return
+	}
+
+	sssInstancesTotal := map[string]int{}
+	sssInstancesUnappliedTotal := map[string]int{}
+
+	ssInstancesTotal := 0
+	ssInstancesUnappliedTotal := 0
+	for _, ssi := range ssis.Items {
+		if ssi.Spec.SelectorSyncSet != nil {
```
staebler: This can be more concise. You do not need the code that initializes the map entry to 0.
```
if sss := ssi.Spec.SelectorSyncSet; sss != nil {
  sssInstancesTotal[sss.Name]++
  if ssi.Status.Applied {
    sssInstancesUnappliedTotal[sss.Name]++
  }
}
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
staebler: This may have been lost in the rest of the discussion from the comments from @twiest, so I will reiterate that the code needs to handle properly an error returned from `updateSyncSetInstanceStatus`.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -80,6 +81,7 @@ func TestSyncSetReconcile(t *testing.T) {
 		expectDeleted          []deletedItemInfo
 		expectSSIDeleted       bool
 		expectErr              bool
+		expectApplied          *bool
```
staebler: Why is this using a bool pointer? Why not always test the expected value of `applied` in every test case?

pkg/operator/hive/hive.go
```diff
@@ -156,6 +156,7 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		"config/crds/hive_v1alpha1_selectorsyncset.yaml",
 		"config/crds/hive_v1alpha1_syncidentityprovider.yaml",
 		"config/crds/hive_v1alpha1_syncset.yaml",
+		"config/crds/hive_v1alpha1_syncsetinstance.yaml",
```
staebler: We are missing more than just this yaml file. Should we instead apply all of the yaml files in the `config/crds` directory?

pkg/controller/metrics/metrics.go
dgoodwin: Much better, thanks.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
dgoodwin: I actually can't see where this was mentioned in the PR, was it slack perhaps? 

Can we still defer and just log the error? Or do we need to be able to requeue in the event of an error updating status, in which case this needs a rethink.

staebler: Mentioned here.
| Also, what happens if updateSyndSetInstanceStatus errors out?

We need to re-queue in the event of an error updating the status.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
dgoodwin: Been awhile but I think I got confused around when we do or don't lookup the instance to do test.validate and this, non-pointer looks like it works fine, updated.

dgoodwin: Not pushed yet, but updated. :) Will wrap up the last one tomorrow.

pkg/controller/metrics/metrics.go
staebler: I see that you made the change in the `else` block but not in the `if` block. Was that an oversight?

dgoodwin:
PR updated.

pkg/apis/hive/v1alpha1/syncsetinstance_types.go
```diff
@@ -82,6 +82,9 @@ type SyncSetInstanceStatus struct {
 	// when a resource type cannot be determined from a SyncSet resource.
 	// +optional
 	Conditions []SyncCondition `json:"conditions,omitempty"`
+
+	// Applied will be true if all resources, patches, or secrets have successfully been applied on last reconcile.
```
staebler: This should be "on the last attempt to apply them" rather than "on last reconcile". There could be reconciles without applies.

pkg/controller/metrics/metrics.go
```diff
@@ -265,13 +285,55 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+	sssInstancesTotal := map[string]int{}
+	sssInstancesUnappliedTotal := map[string]int{}
+
+	ssInstancesTotal := 0
+	ssInstancesUnappliedTotal := 0
+	for _, ssi := range ssis.Items {
+		if sss := ssi.Spec.SelectorSyncSet; sss != nil {
+			// Process SyncSetInstances with a SelectorSyncSet reference:
+			sssInstancesTotal[sss.Name]++
+
```
staebler: I would like to see this empty line removed to make the if block look the same as the else block.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -189,6 +189,10 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return r.syncDeletedSyncSetInstance(ssi, ssiLog)
 	}
 
+	// Assume applied is false until we successfully apply everything to cluster.
+	ssi.Status.Applied = false
```
staebler: Since we are not doing the defer any more, this does not need to be done up here any more. We can keep the deep copy where it was before and omit the code that explicitly sets `Applied` to false.

pkg/controller/metrics/metrics.go
dgoodwin: I just pushed recently, not sure if you saw that version, but the only else I see in this area wasn't doing the same thing. (just simple counters going up, no map)

staebler: Yes, I was looking at the version before the push. I see that it is updated now.

pkg/apis/hive/v1alpha1/syncsetinstance_types.go
dgoodwin: Updated with a slightly more brief version.

dgoodwin:
Updated again.

staebler:
@dgoodwin If you get a chance before this merges, can you squash the commits?

dgoodwin:
done

PR #721, opened 2019-11-28
Add basic support for bare metal cluster provisioning.

dgoodwin:
Adds a new platform, currently just an empty struct, for BareMetal.

If we're provisioning bare metal we need to lookup a ref for a different
installer image. Using this image the openshift-install binary will be
in the exact same location with the exact same name as we normally use,
so we just need to swap out the image reference and everything else can
remain the same.

There is a pending need for installing mkisofs which is hopefully going
to be handled by the requirement being removed upstream soon.

### DIFF
--- a/pkg/apis/hive/v1/baremetal/machinepool.go
+++ b/pkg/apis/hive/v1/baremetal/machinepool.go
@@ -1,0 +1,6 @@
+package baremetal
+
+// MachinePoolPlatform stores the configuration for a machine pool
+// installed on bare metal.
+type MachinePoolPlatform struct {
+}
--- a/pkg/apis/hive/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/hive/v1/zz_generated.deepcopy.go
@@ -8,6 +8,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	aws "github.com/openshift/hive/pkg/apis/hive/v1/aws"
 	azure "github.com/openshift/hive/pkg/apis/hive/v1/azure"
+	baremetal "github.com/openshift/hive/pkg/apis/hive/v1/baremetal"
 	gcp "github.com/openshift/hive/pkg/apis/hive/v1/gcp"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -1790,6 +1791,11 @@ func (in *Platform) DeepCopyInto(out *Platform) {
 		in, out := &in.GCP, &out.GCP
 		*out = new(gcp.Platform)
 		(*in).DeepCopyInto(*out)
+	}
+	if in.BareMetal != nil {
+		in, out := &in.BareMetal, &out.BareMetal
+		*out = new(baremetal.Platform)
+		**out = **in
 	}
 	return
 }
--- a/pkg/imageset/generate.go
+++ b/pkg/imageset/generate.go
@@ -1,6 +1,7 @@
 package imageset
 
 import (
+	"fmt"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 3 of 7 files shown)


pkg/apis/hive/v1/zz_generated.deepcopy.go
```diff
@@ -1791,6 +1792,11 @@ func (in *Platform) DeepCopyInto(out *Platform) {
 		*out = new(gcp.Platform)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.BareMetal != nil {
+		in, out := &in.BareMetal, &out.BareMetal
+		*out = new(baremetal.Platform)
+		**out = **in
```
gyliu513: Just a question here, why not `DeepCopyInto` like other cloud providers?

dgoodwin: This is generated code, I honestly am not sure why it's doing that differently, possibly because it's just an empty struct today.

pkg/apis/hive/v1/baremetal/machinepool.go
```diff
@@ -0,0 +1,6 @@
+package baremetal
+
+// MachinePoolPlatform stores the configuration for a machine pool
+// installed on bare metal.
+type MachinePoolPlatform struct {
```
dhellmann: If the MachinePool is related to a MachineSet, does this need to have fields like the provider spec for a machine set? The fields for bare metal are in https://github.com/openshift/cluster-api-provider-baremetal/blob/master/pkg/apis/baremetal/v1alpha1/baremetalmachineproviderspec_types.go#L33

joelddiaz: The way things currently stand with MachinePools is that one MachinePool can represent multiple MachineSets. So a 1-to-many relationship between MachinePools->MachineSets (although you can definitely define a MachinePool with a list of AvailabilityZones of exactly one AZ, and that will result in one single MachineSet being generated).

dgoodwin: Remote machine pool generation (day 2) should be disabled by default until we add support, so we can add this as followup.

pkg/imageset/generate.go
```diff
@@ -23,9 +24,12 @@ type ImageSpec struct {
...
 
 const (
-	extractImageScript = `#/bin/bash
+	// extractImageScriptTemplate is a minimal shell script we run in the imageset job to determine the images to
+	// use for various OpenShift components involved in the provisioning. A string value will be formatted in
+	// for which installer image we need to use. (regular vs baremetal)
+	extractImageScriptTemplate = `#/bin/bash
 echo "About to run oc adm release info"
-if oc adm release info --image-for="installer" --registry-config "${PULL_SECRET}" "${RELEASE_IMAGE}" > /common/installer-image.txt 2> /common/error.log; then
+if oc adm release info --image-for="%s" --registry-config "${PULL_SECRET}" "${RELEASE_IMAGE}" > /common/installer-image.txt 2> /common/error.log; then
```
staebler: It would be nice to use [go templates](https://golang.org/pkg/text/template) for this to make it a little easier to maintain, given the length of the template. Although, admittedly, that may be overkill for a template with only a single parameter. With a templates, you could embed in the template the image keys to use and pass the platform as a parameter to the template instead.

staebler:
I don't know if it is intentional at this point, but clusterdeployments for bare metal will be rejected by the validating webhook. See https://github.com/openshift/hive/blob/dbba4e8a98ad315f9f6863ca7df493b5480bfbc5/pkg/apis/hive/v1/validating-webhooks/clusterdeployment_validating_admission_hook.go#L244-L293.

staebler:
And associated machinepools will be rejected as well. See https://github.com/openshift/hive/blob/dbba4e8a98ad315f9f6863ca7df493b5480bfbc5/pkg/apis/hive/v1/validating-webhooks/machinepool_validating_admission_hook.go#L253-L277.

dgoodwin:
Not intentional.

/hold

dgoodwin:
/hold cancel

We'll get it as followup.

PR #722, opened 2019-11-29
Add scripting and notes from first scale test.

dgoodwin:
Not in perfect state but wanted to get the tooling involved committed so
we can evolve it over time as more tests are run.


hack/scaletest/README.md
```diff
@@ -0,0 +1,79 @@
+# Goal
+
+Simulate 1000 clusters in Hive and observe performance. Because we do not have the budget or resources to actually provision 1000 clusters, we will provision one, and adopt 1000 copies of it into Hive. For the purposes of all API communication we will be talking to one cluster. Hive runs 5 threads per controller, so we should not be making excessive concurrent requests to the cluster. However we need to observe for problems from this caveat with regards to SyncSets possibly competing over the same resources and thrashing. If syncing the same resources however, we should always expect the same result, so thrashing should not be possible.
```
LalatenduMohanty: s/we will provision one, and adopt 1000 copies of it into Hive/we will provision one, and adopt 1000 copies of its clusterdeployment into Hive for testing/s

hack/scaletest/syncset-template.yaml
```diff
@@ -0,0 +1,29 @@
...
+      metadata:
+        name: scaletest-${RESOURCE_INDEX}-configmap
+        namespace: default
+        labels:
+          scaletest: "true"
+      data:
+        foo: bar
+    clusterDeploymentSelector:
+      matchLabels:
+        scaletest: "true"
```
LalatenduMohanty: :+1:

PR #723, opened 2019-12-02
Allow bare metal validation to pass and begin execution.

dgoodwin:
Required an additional change to install a library that is unfortunately
not available in the UBI. Modified to match what the bare metal installer
image itself is using.


Dockerfile
```diff
@@ -7,11 +7,15 @@ RUN go build -o bin/hiveutil github.com/openshift/hive/contrib/cmd/hiveutil
 RUN go build -o bin/hiveadmission github.com/openshift/hive/cmd/hiveadmission
 RUN go build -o bin/hive-operator github.com/openshift/hive/cmd/operator
 
-FROM registry.access.redhat.com/ubi7/ubi
+#FROM registry.access.redhat.com/ubi7/ubi
```
twiest: Is the intention to go back to UBI at some point?

If not, should we just remove this line instead of commenting it out?

dgoodwin: Mixed on this, I'd kinda like it to be kept in our face a little.

dgoodwin:
@twiest @staebler this is everything I've had to do to get bare metal working. Could you let me know what you think is merging?

pkg/installmanager/installmanager.go
```diff
@@ -72,6 +72,7 @@ const (
 	sshCopyTempFile                     = "/tmp/ssh-privatekey"
 	defaultInstallConfigMountPath       = "/installconfig/install-config.yaml"
 	defaultManifestsMountPath           = "/manifests"
+	hiveSSHKnownHostsAnnotation         = "hive.openshift.io/ssh-known-host"
```
dgoodwin: This is pretty hacky and I don't think it's the permanent solution, we'll have libvirt in a pod eventually,I could rename the annotation to indicate it's not something you should consider permanent?

pkg/installmanager/installmanager.go
```diff
@@ -223,6 +224,23 @@ func (m *InstallManager) Run() error {
...
+		// Not a fatal error.
+		m.log.WithError(sshAgentSetupErr).Info("unable to initialize ssh agent")
+	} else {
+		sshCleanupFunc, sshAgentSetupErr := m.initSSHAgent(sshKeyPath)
+		defer sshCleanupFunc()
+		if sshAgentSetupErr != nil {
+			// Not a fatal error.
+			m.log.WithError(sshAgentSetupErr).Info("ssh agent is not initialized")
+		}
+	}
```
dgoodwin: Moving SSH agent setup code here, we are assuming the SSH private key can also be used to hit libvirt over ssh as I didn't want to add another SSH key to the API, as this is also not likely permanent. We now always initialize SSH if we can, and if we cannot it's not a fatal error.

I will run a test that log gathering still works and report back.

pkg/installmanager/installmanager.go
staebler: We can leave the annotation name as is. I don't think there is any danger of someone using this and us dropping support for it.

pkg/installmanager/installmanager.go
staebler: Should this be an error if it is a bare-metal install? Or is it OK to wait until the later failure to error out if we cannot initialize SSH?

pkg/installmanager/installmanager.go
```diff
@@ -232,12 +250,20 @@ func (m *InstallManager) Run() error {
 	cmd := exec.Command("cp", m.InstallConfigMountPath, destInstallConfigPath)
 	err = cmd.Run()
 	if err != nil {
-		log.WithError(err).Errorf("error copying install-config.yaml from %s to %s",
+		m.log.WithError(err).Errorf("error copying install-config.yaml from %s to %s",
```
staebler: I ran into this same deficiency elsewhere. It is terribly unfortunate that logrus has the same bare functions that can be called outside of an explicit logger.

pkg/installmanager/installmanager.go
```diff
@@ -705,75 +735,27 @@ func (m *InstallManager) gatherClusterLogs(cd *hivev1.ClusterDeployment) error {
...
-
-	m.log.Info("attempting to gather logs from cluster bootstrap node")
-
+func (m *InstallManager) initSSHKey() (string, error) {
 	m.log.Debug("checking for SSH private key")
 	sshPrivKeyPath := os.Getenv("SSH_PRIV_KEY_PATH")
 	if sshPrivKeyPath == "" {
-		m.log.Warn("cannot gather logs as SSH_PRIV_KEY_PATH is unset or empty")
-		return fmt.Errorf("cannot gather logs as SSH_PRIV_KEY_PATH is unset or empty")
+		m.log.Warn("cannot configure SSH agent as SSH_PRIV_KEY_PATH is unset or empty")
```
staebler: This should probably not be a warning any more. Otherwise, we will always warn when the user does not provide a private key--for all installs.

pkg/installmanager/installmanager.go
```diff
@@ -705,75 +735,27 @@ func (m *InstallManager) gatherClusterLogs(cd *hivev1.ClusterDeployment) error {
...
-	newSSHPrivKeyPath, err := m.chmodSSHPrivateKey(sshPrivKeyPath)
-	if err != nil {
-		return err
-	}
-
-	if fileInfo.Size() == 0 {
+		m.log.WithError(err).Error("error stat'ing file containing private key")
+		return "", err
+	} else if fileInfo.Size() == 0 {
 		m.log.Warn("cannot gather logs/tarball as ssh private key file is empty")
```
staebler: This log entry should mention initializing SSH agent rather than gathering logs.

pkg/installmanager/installmanager.go
```diff
@@ -1133,6 +1147,19 @@ func waitForProvisioningStage(provision *hivev1.ClusterProvision, m *InstallMana
 	return errors.Wrap(err, "ClusterProvision did not transition to provisioning stage")
 }
 
+func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
```
staebler: Long term, this function should be resilient to the directory and the file already existing.

pkg/installmanager/installmanager.go
dgoodwin: Yeah it will quickly error elsewhere if SSH isn't available, so this form leaves us a little closer to long term form.

dgoodwin: (the SSH need won't be permanent I should have mentioned)

pkg/installmanager/installmanager.go
dgoodwin: The file write should overwrite as it was, and I switched to os.MkdirAll which should return nothing in the event of the dir(s) already existing.

pkg/installmanager/installmanager.go
dgoodwin: done

dgoodwin:
@staebler @twiest this one is updated per review, would like to get merged soon.

pkg/installmanager/installmanager.go
staebler: By resilient to an already existing file, I meant that it should *not* overwrite the file. It should only add the additional known host to the file. But that is a problem that we can address another day.

PR #726, opened 2019-12-03
Add documentation for secretReferences

matt-simons:


docs/syncset.md
```diff
@@ -41,6 +41,18 @@ spec:
     patch: |-
       { "data": { "foo": "new-bar" } }
     patchType: merge
+    
+  secretReferences:
+  - source:
+      kind: Secret
```
staebler: The kind and apiVersion can be omitted in both the source and the target.

dgoodwin: Yeah those fields would be unused here and they should probably be removed.

dgoodwin:
Thanks a bunch for adding this @matt-simons

docs/syncset.md
matt-simons: cool, removed those

PR #732, opened 2019-12-06
allow machine pools to be auto-scaled

staebler:
This adds an autoscaling field to MachinePools. Under the autoscaling field are the minReplicas and maxReplicas fields. The autoscaling field is mutually exclusive with the replicas field. The user can either specify a static number of replicas or specify that the cluster should take care of auto-scaling between the min and max replicas.

When auto-scaling is enabled for a machine pool, Hive will ensure that the cluster has a "default" ClusterAutoscaler resource with scale down enabled. Hive will also create MachineAutoscaler resources for each of the MachineSets for the MachinePool. The minReplicas and maxReplicas specified in the MachinePool will be distributed evenly across all of the MachineAutoscaler resources.

For example, let us say that minReplicas is 5 and maxReplicas is 10 for a MachinePool using 3 zones. Since there are 3 zones, there will be 3 MachineSets and 3 MachineAutoscalers. The first MachineAutoscaler will specify a minimum of 2 replicas and a maximum of 4 replicas. The second MachineAutoscaler will specify a minimum of 2 replicas and a maximum of 3 replicas. The last MachineAutoscaler will specify a minimum of 1 replica and a maximum of 3 replicas. In total, the minimum replicas specified by the MachineAutoscalers will be (2+2+1=) 5 and the maximum replicas will be (4+3+3=) 10.

In order for the autoscaler to consider a MachineSet for scaling, that MachineSet must specify at least 1 replica. Consequently, the minimum number of replicas is bounded on the lower end by the number of zones used for the MachinePool. If the zones are specified explicitly in the MachinePool, then the validator will ensure that the minimum replicas specified are sufficient. If the zones are not specified explicitly, then the controller will put a condition on the MachinePool if the minimum replicas are not sufficient. In the latter case, the controller will not create any MachineSets until the user specifies a sufficient number of minimum replicas.


staebler:
/hold

This depends on #727.

staebler:
I forgot that I still had work to do on the e2e tests for auto-scaling.

PR #735, opened 2019-12-07
allow machine pools to be auto-scaled

staebler:
This add an autoscaling field to MachinePools. Under the autoscaling field are the minReplicas and maxReplicas fields. The autoscaling field is mutually exclusive with the replicas field. The user can either specify a static number of replicas or specify that the cluster should take care of auto-scaling between the min and max replicas.

When auto-scaling is enabled for a machine pool, Hive will ensure that the cluster has a "default" ClusterAutoscaler resource with scale down enabled. Hive will also create MachineAutoscaler resources for each of the MachineSets for the MachinePool. The minReplicas and maxReplicas specified in the MachinePool will be distributed evenly across all of the MachineAutoscaler resources.

For example, let us say that minReplicas is 5 and maxReplicas is 10 for a MachinePool using 3 zones. Since there are 3 zones, there will be 3 MachineSets and 3 MachineAutoscalers. The first MachineAutoscaler will specify a minimum of 2 replicas and a maximum of 4 replicas. The second MachineAutoscaler will specify a minimum of 2 replicas and a maximum of 3 replicas. The last MachineAutoscaler will specify a minimum of 1 replica and a maximum of 3 replicas. In total, the minimum replicas specified by the MachineAutoscalers will be (2+2+1=) 5 and the maximum replicas will be (4+3+3=) 10.

In order for the autoscaler to consider a MachineSet for scaling, that MachineSet must specify at least 1 replica. Consequently, the minimum number of replicas is bounded on the lower end by the number of zones used for the MachinePool. If the zones are specified explicitly in the MachinePool, then the validator will ensure that the minimum replicas specified are sufficient. If the zones are not specified explicitly, then the controller will put a condition on the MachinePool if the minimum replicas are not sufficient. In the latter case, the controller will not create any MachineSets until the user specifies a sufficient number of minimum replicas.


test/e2e/postinstall/machinesets/infra_test.go
```diff
@@ -155,15 +161,159 @@ func TestNewMachinePool(t *testing.T) {
...
+							},
+						},
+					}},
+				},
+			},
+		},
+	}
+	if err := rc.Create(context.TODO(), busyboxDeployment); err != nil {
+		t.Fatalf("cannot create busybox deployment: %v", err)
+	}
```
dgoodwin: This is a nice bit of unexpected test coverage.

Could you add some comments to elaborate on how the test works? It looks like we're autoscaling based on replicas needed, which confused me a little as I thought this operated on CPU load. I guess requesting 50 is just a high number to ensure for whatever reason, we want to max the full 12, so it probably is still using CPU or memory to determine?

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -213,85 +213,199 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
+	logger.Infof("generated %v worker machine sets", len(generatedMachineSets))
+
+	return generatedMachineSets, nil
+}
+
+func (r *ReconcileRemoteMachineSet) ensureEnoughReplicas(
+	pool *hivev1.MachinePool,
+	generatedMachineSets []*machineapi.MachineSet,
+	logger log.FieldLogger,
+) (*reconcile.Result, error) {
```
dgoodwin: Could use a comment on why this returns an optional result pointer, as well as an error, looks like it's to avoid a requeue if you just used errrors to stop the reconcile loop, as we know there's no point in re-reconciling here?

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -300,41 +414,62 @@ func (r *ReconcileRemoteMachineSet) syncMachineSets(
...
-					// If the length of both labels is zero, then they match, even if one is a nil map and the other is an empty map.
-					if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; (len(rl) != 0 || len(l) != 0) && !reflect.DeepEqual(rl, l) {
-						msLog.WithField("desired", l).WithField("observed", rl).Info("labels out of sync")
-						rMS.Spec.Template.Spec.Labels = l
+				} else {
+					// If minReplicas==maxReplicas, then the autoscaler will ignore the machineset,
+					// even if the replicas in the machineset is not equal to the min and max.
+					// To ensure that the replicas falls within min and max regardless, Hive needs
+					// to set the replicas to explicitly be within the desired range.
+					min, max := getMinMaxReplicasForMachineSet(pool, generatedMachineSets, i)
```
dgoodwin: Good catch.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -352,65 +487,225 @@ func (r *ReconcileRemoteMachineSet) syncMachineSets(
...
+	tm.SetGroupVersionKind(autoscalingv1.SchemeGroupVersion.WithKind("ClusterAutoscaler"))
+	if err := remoteClusterAPIClient.List(
+		context.Background(),
+		remoteClusterAutoscalers,
+		client.UseListOptions(&client.ListOptions{Raw: &metav1.ListOptions{TypeMeta: tm}}),
+	); err != nil {
+		logger.WithError(err).Error("unable to fetch remote cluster autoscalers")
+		return err
+	}
+	logger.Infof("found %v remote cluster autoscalers", len(remoteClusterAutoscalers.Items))
```
dgoodwin: Should we warn if this is over 1?

test/e2e/postinstall/machinesets/infra_test.go
staebler: Yes, I will add some comments.

It is autoscaling on CPU load. For AWS, each machine has 4 CPU. To get the autoscaler to build out 12 machines, the CPU demand needs to be at least 48. Each pod in the deployment is requesting 2 CPU. The deployment requests an excessive amount of pods because (1) it does not matter if the deployment replicas is too high and (2) I want to make it a bit more resilient to changes in instance types across platforms and over time where the ultimate machine may have more or less than 4 CPU.

pkg/controller/remotemachineset/remotemachineset_controller.go
staebler: Yes, it is so that we do not requeue in the case where the replicas count is too low. A requeue will never succeed until the machinepool is updated.

I will add a comment.

pkg/controller/remotemachineset/remotemachineset_controller.go
staebler: I don't know that it is Hive's job to warn about this. I am hesitant to add warning log events to Hive for things that do not affect Hive's behavior. Any ClusterAutoscaler that is not named "default" is (1) ignored by Hive and (2) ignored by the autoscaler.

staebler:
@dgoodwin I have added comments were requested. Let me know if you think there need some more color.

dgoodwin:
/lgtm

Thanks!

PR #736, opened 2019-12-09
makefile: do not install mockgen automatically

staebler:
As with golintci-lint, it should be left up to the developer to manually install mockgen on their local development environment.

Also, provide a means for running builds locally using the build image.
 * make build-build-image will build the build image
 * ./hack/make script will run make in a container using the build image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ test-e2e-postinstall:
 
 # Builds all of hive's binaries (including utils).
 .PHONY: build
-build: $(GOPATH)/bin/mockgen manager hiveutil hiveadmission operator
+build: manager hiveutil hiveadmission operator
 
 
 # Build manager binary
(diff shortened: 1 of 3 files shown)


staebler:
I need to move the changes to the build image to a separate PR that will need to merge before this PR.

staebler:
/hold

This requires #737.

Makefile
```diff
@@ -207,9 +207,6 @@ buildah-dev-push: build
 buildah-push: buildah-build
 	$(SUDO_CMD) buildah push ${IMG}
 
-$(GOPATH)/bin/mockgen:
-	go get -u github.com/golang/mock/mockgen/...
```
dgoodwin: Could you add this to the developing docs, we don't mention mockgen there at all today.

staebler: Will do.

staebler:
/hold cancel

#737 has merged.

PR #738, opened 2019-12-09
types: move provision images

staebler:
In ClusterDeployment, move the fields in `.spec.images` to `.spec.provisioning` and the `spec.imageSetRef` field to `.spec.provisioning.imageSetRef`.


staebler:
This PR silently fixed a bug in `hiveutil create-cluster` where an imageset was not being used even though the `--use-image-set` flag was true (which is the default). This exposed another bug in the clusterdeployment controller where it was accepting a blank installer image from the clusterimageset. This PR has been updated to fix that second bug as well.

PR #739, opened 2019-12-10
Disable deprovision for baremetal clusters.

dgoodwin:
There is currently no deprovision path for baremetal. Skip launching a
job so we don't need to manually remove finalizers for time being.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1113,6 +1113,16 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1113,6 +1113,16 @@ func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.Clu
 		return reconcile.Result{}, err
 	}
 
+	// We do not yet support deprovision for BareMetal, for now skip deprovision and remove finalizer.
+	if cd.Spec.Platform.BareMetal != nil {
+		cdLog.Info("skipping deprovision for BareMetal cluster, removing finalizer")
+		err = r.removeClusterDeploymentFinalizer(cd)
```
staebler: Oh, one more thing. We should limit the scope of variables as much as possible. Use `err := `.

dgoodwin:
Updated.

PR #740, opened 2019-12-10
[WIP] Recover owner references for Hive owned objects

twiest:
- [x] When creating an owner reference, also create a label
- [x] When the label exists, but not the owner reference, restore the
      owner reference
- [x] Add unit tests
- [x] Manually test

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -50,6 +50,15 @@ const (
 
 	// GCPCredentialsName is the name of the GCP credentials file or secret key.
 	GCPCredentialsName = "osServiceAccount.json"
+
+	// ClusterDeploymentOwnerLabel is the label that is used to identify the owner of a runtime object.
+	ClusterDeploymentOwnerLabel = "hive.openshift.io/owner-cluster-deployment-name"
+
+	// ClusterProvisionOwnerLabel is the label that is used to identify the owner of a runtime object.
+	ClusterProvisionOwnerLabel = "hive.openshift.io/owner-cluster-provision-name"
+
+	// ClusterDeprovisionOwnerLabel is the label that is used to identify the owner of a runtime object.
+	ClusterDeprovisionOwnerLabel = "hive.openshift.io/owner-cluster-deprovision-name"
 )
 
 // GetMergedPullSecretName returns name for merged pull secret name per cluster deployment
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -17,6 +17,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -43,7 +44,19 @@ import (
 )
 
 // controllerKind contains the schema.GroupVersionKind for this controller type.
-var controllerKind = hivev1.SchemeGroupVersion.WithKind("ClusterDeployment")
+var (
+	controllerKind                    = hivev1.SchemeGroupVersion.WithKind("ClusterDeployment")
+	ownedByClusterDeploymentListTypes = []runtime.Object{
+		&batchv1.JobList{},
+		&corev1.PersistentVolumeClaimList{},
+		&corev1.SecretList{},
+		&hivev1.ClusterProvisionList{},
+		&hivev1.ClusterDeprovisionList{},
+		&hivev1.DNSZoneList{},
+		&hivev1.ClusterStateList{},
+		&hivev1.SyncSetList{},
+	}
+)
 
 const (
 	controllerName     = "clusterDeployment"
@@ -282,7 +295,86 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
 		return reconcile.Result{}, err
 	}
 
+	rolResult, err := r.reconcileOwnerReferences(constants.ClusterDeploymentOwnerLabel, cd, cdLog)
+	if err != nil {
+		cdLog.WithError(err).Error("Failed to reconcile objects owned by this cluster deployment.")
+		return reconcile.Result{}, err
+	}
+
+	if rolResult != nil {
+		cdLog.Info("Reconciled objects owned by this cluster deployment.")
+		return *rolResult, nil
+	}
+
 	return r.reconcile(request, cd, cdLog)
+}
+
+func (r *ReconcileClusterDeployment) reconcileOwnerReferences(ownerLabel string, cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (*reconcile.Result, error) {
+	selector := map[string]string{ownerLabel: cd.Name}
+	objects, err := controllerutils.GetRuntimeObjects(r, ownedByClusterDeploymentListTypes, client.InNamespace(cd.Namespace), client.MatchingLabels(selector))
+	if err != nil {
+		errors.Wrap(err, "Failed to list objects owned by ClusterDeployment")
+		return nil, err
+	}
+
+	allErrors := []error{}
+	objectsModified := false
+
+	for _, obj := range objects {
+		tmpModified, err := r.checkAndFixOwnerReferences(cd, obj, cdLog)
+		if tmpModified {
+			objectsModified = true
+		}
+		if err != nil {
+			allErrors = append(allErrors, err)
+		}
+	}
+
+	if len(allErrors) != 0 {
+		return nil, utilerrors.NewAggregate(allErrors)
+	}
+
+	if objectsModified {
+		cdLog.Debug("We successfully reconciled all owner references, and some were fixed.")
+		return &reconcile.Result{
+			RequeueAfter: defaultRequeueTime,
+		}, nil
+	}
+
+	cdLog.Debug("We successfully reconciled all owner references, and none needed to be fixed.")
+	return nil, nil
+}
+
+func (r *ReconcileClusterDeployment) checkAndFixOwnerReferences(cd *hivev1.ClusterDeployment, obj runtime.Object, cdLog log.FieldLogger) (bool, error) {
+	meta, ok := obj.(metav1.Object)
+	if !ok {
+		return false, fmt.Errorf("Failed converting runtime object to meta object: %v", obj)
+	}
+
+	objLog := cdLog.WithField("ownedObject", meta.GetName())
+
+	for _, ownerRef := range meta.GetOwnerReferences() {
+		if *ownerRef.Controller &&
+			ownerRef.APIVersion == cd.APIVersion &&
+			ownerRef.Kind == cd.Kind &&
+			ownerRef.Name == cd.Name {
+
+			objLog.Debug("We found the owner reference. Nothing more to do on this one.")
+			return false, nil
+		}
+	}
+
+	objLog.Debug("Restore the owner reference")
+	if err := controllerutil.SetControllerReference(cd, meta, r.scheme); err != nil {
+		return false, errors.Wrapf(err, "could not restore the owner reference: %v", meta.GetName())
+	}
+	err := r.Update(context.TODO(), obj)
+	if err != nil {
+		return false, errors.Wrapf(err, "could not update the object with the restored owner reference: %v", meta.GetName())
+	}
+
+	objLog.Info("We successfully fixed the owner reference for this object.")
+	return true, nil
 }
 
 func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (result reconcile.Result, returnErr error) {
@@ -530,6 +622,7 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		provision.Spec.PrevInfraID = &cd.Spec.ClusterMetadata.InfraID
 	}
 
+	controllerutils.SetOwnerLabel(constants.ClusterDeploymentOwnerLabel, cd, provision)
 	if err := controllerutil.SetControllerReference(cd, provision, r.scheme); err != nil {
 		cdLog.WithError(err).Error("could not set the owner ref on provision")
 		return reconcile.Result{}, err
@@ -785,10 +878,12 @@ func (r *ReconcileClusterDeployment) createPVC(cd *hivev1.ClusterDeployment, cdL
 	}
 
 	cdLog.WithField("pvc", pvc.Name).Info("creating persistent volume claim")
+	controllerutils.SetOwnerLabel(constants.ClusterDeploymentOwnerLabel, cd, pvc)
 	if err := controllerutil.SetControllerReference(cd, pvc, r.scheme); err != nil {
 		cdLog.WithError(err).Error("error setting controller reference on pvc")
 		return err
 	}
+
 	err := r.Create(context.TODO(), pvc)
 	if err != nil {
 		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error creating pvc")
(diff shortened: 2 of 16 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
+	meta, ok := obj.(metav1.Object)
+	if !ok {
+		return false, fmt.Errorf("Failed converting runtime object to meta object: %v", obj)
+	}
+
+	for _, ownerRef := range meta.GetOwnerReferences() {
+		if *ownerRef.Controller &&
+			ownerRef.APIVersion == cd.APIVersion &&
+			ownerRef.Kind == cd.Kind &&
+			ownerRef.Name == cd.Name {
```
twiest: Is there a better way to see if an object is owned by another object?

I looked in kube and couldn't find anything that did that (I may have missed it).

pkg/constants/constants.go
```diff
@@ -58,6 +58,9 @@ const (
 
 	// GCPCredentialsName is the name of the GCP credentials file or secret key.
 	GCPCredentialsName = "osServiceAccount.json"
+
+	// OwnerLabel is the label that is used to identify the owner of a runtime object.
+	OwnerLabel = "hive.openshift.io/owner"
```
staebler: I don't think that you want to use a generic label. This is going to conflict for the types that may be owned by multiple different types of owners. For example, jobs can be owned by either a clusterdeployment or a clusterprovision. What you should be looking for is a specific label placed on a resource that is doing a specific task. For example, to re-establish the owner reference for the imageset job, you should be looking for a job with a labels "hive.openshift.io/imageset=true" and "hive.openshift.io/cluster-deployment-name=CD_NAME", where CD_NAME is the name of the clusterdeployment being reconciled.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
 	}
 
+	rolResult, err := r.reconcileOwnerReferences(cd, cdLog)
+	if err != nil {
+		cdLog.WithError(err).Error("Failed to reconcile objects owned by this cluster deployment.")
+		return reconcile.Result{}, err
+	}
+
+	if rolResult != nil {
+		cdLog.WithError(err).Error("Reconciled objects owned by this cluster deployment.")
```
staebler: This should not be an error-level message. Or, at a minimum, it should not use `WithError` since `err` will always be nil.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
+	}
+
+	if len(allErrors) != 0 {
+		return nil, utilerrors.NewAggregate(allErrors)
+	}
+
+	if objectsModified {
+		cdLog.Debug("We successfully reconciled all owner references, and some were fixed.")
+		return &reconcile.Result{
+			RequeueAfter: defaultRequeueTime,
```
staebler: Why does this need to wait to requeue instead of requeueing immediately?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: https://github.com/kubernetes/apimachinery/blob/adc6f4cd9e7d28fc1c1e5efc658d940d55c0f356/pkg/apis/meta/v1/controller_ref.go#L33

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
+	objLog.Debug("Restore the owner reference")
+	if err := controllerutil.SetControllerReference(cd, meta, r.scheme); err != nil {
+		return false, errors.Wrapf(err, "could not restore the owner reference: %v", obj)
+	}
+	err := r.Update(context.TODO(), obj)
+	if err != nil {
+		return false, errors.Wrapf(err, "could not update the object with the restored owner reference: %v", obj)
+	}
+
+	objLog.Debug("We successfully fixed the owner reference for this object.")
```
staebler: This should probably be Info level.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
+			ownerRef.Name == cd.Name {
+
+			objLog.Debug("We found the owner reference. Nothing more to do on this one.")
+			return false, nil
+		}
+	}
+
+	objLog.Debug("Restore the owner reference")
+	if err := controllerutil.SetControllerReference(cd, meta, r.scheme); err != nil {
+		return false, errors.Wrapf(err, "could not restore the owner reference: %v", obj)
```
staebler: It is overkill to dump the entire contents of the object into the error. For secrets, we do not want to write the contents of the secrets into the logs.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
+	}
+	err := r.Update(context.TODO(), obj)
+	if err != nil {
+		return false, errors.Wrapf(err, "could not update the object with the restored owner reference: %v", obj)
```
staebler: Same here about dumping the entire object into the logs.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -280,9 +293,87 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
...
+			RequeueAfter: defaultRequeueTime,
+		}, nil
+	}
+
+	cdLog.Debug("We successfully reconciled all owner references, and none needed to be fixed.")
+	return nil, nil
+}
+
+func (r *ReconcileClusterDeployment) checkAndFixOwnerReferences(cd *hivev1.ClusterDeployment, obj runtime.Object, cdLog log.FieldLogger) (bool, error) {
+	objLog := cdLog.WithField("ownedObject", obj)
```
staebler: Do not add the entire object as a field to the logger.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -10,7 +10,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	batchv1 "k8s.io/api/batch/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
```
staebler: To be consistent with other files, let's use "apierrors" instead of "kerrors".

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -44,7 +45,19 @@ import (
...
+var (
+	controllerKind                    = hivev1.SchemeGroupVersion.WithKind("ClusterDeployment")
+	ownedByClusterDeploymentListTypes = []runtime.Object{
+		&batchv1.JobList{},
+		&corev1.PersistentVolumeClaimList{},
+		&corev1.SecretList{},
+		&hivev1.ClusterProvisionList{},
+		&hivev1.ClusterDeprovisionList{},
+		&hivev1.DNSZoneList{},
+		&hivev1.ClusterStateList{},
```
staebler: The controller that creates the object should be the one responsible for ensuring the owner reference.

pkg/installmanager/installmanager.go
```diff
@@ -959,6 +959,7 @@ func uploadAdminKubeconfig(provision *hivev1.ClusterProvision, m *InstallManager
 		},
 	}
 
+	controllerutils.SetOwnerLabel(provision, kubeconfigSecret)
```
staebler: Not a concern for this PR, but this does bring to light a bit of an issue. I think that we want the ownership of the kubeconfig and admin password secrets to be passed from the ClusterProvision to the ClusterDeployment once the provision has completed.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
twiest: Is that always going to be possible?

What about a situation where a controller creates the object, but doesn't watch the owner object?

What will reconcile it then?

Also, is this a change from what we discussed, or did I just not understand you before?

I thought you were saying that the owner of the object should reconcile the owned objects.

staebler: > Is that always going to be possible?
>
> What about a situation where a controller creates the object, but doesn't watch the owner object?

What would prompt a controller to create an object on behalf of owner that the controller is not reconciling? I cannot think of a scenario. How would a controller know to create an object for the owner in that case?

> What will reconcile it then?
> 
> Also, is this a change from what we discussed, or did I just not understand you before?
> 
> I thought you were saying that the owner of the object should reconcile the owned objects.

This is what I recall us discussing. I don't know what it means for "the owner of the object [to] reconcile". Objects do no reconcile: Controllers reconcile. Typically, there is a single controller that reconciles a particular object. However, Hive has multiple controllers that reconcile the same object, presumably in order to modularize features. If a controller is creating an object as part of a feature that is isolated to that controller, then it breaks modularity to leak the responsibility of maintaining the owner relationship of that object to a different controller.

twiest: Interesting, I must have misunderstood you.

What you're proposing sounds a lot like what I was initially proposing (the controller of the owned object reconciles the ownership), but I thought you suggested to do it this way.

Let's discuss tomorrow at stand up.

staebler: It seems like there is still a misunderstanding. I am not suggesting that the controller reconciling the owned object does anything. The controller that created the object should be responsible for maintaining the owner relationship. The controller that created the object should not be the controller that reconciles the object. For example, the clusterDeployment controller reconciles clusterDeployment resources. That controller creates clusterProvision  resources owned by clusterDeployment resources. The clusterDeployment controller should maintain the owner relationship between the clusterProvision resource and the clusterDeployment resource.

staebler: Let's look at another example where the clusterDeployment controller should not be maintaining the relationship even though the owner is a ClusterDeployment resource. The clusterState controller reconciles ClusterDeployment resources and creates ClusterState resources owned by ClusterDeployment resources. The clusterState controller should be the controller responsible for maintaining the owner relationship between ClusterState resources and ClusterDeployment resources.

dgoodwin:
My understanding from our call this morning with some examples.

Firstly many of us learned that we probably should be propagating all labels from a "parent" object down to a "child" object when we create it, per kube patterns. We don't do this today in many cases but probably should, so we'll assume that is the case.

Second we discussed using two labels for any scenario of ownership. One essentially indicates the type or origin of the ownership, the second indicates the specific owner.

Example 1: imageset job, created by the clusterdeployment controller.
hive.openshift.io/cluster-deployment-name=dgoodwin-gcp (already exists today)
hive.openshift.io/imageset=true (already exists today, indicates this is an imageset job)
The cluster deployment controller can process jobs, it knows that only it creates imageset jobs, it can tell if this job is from the cluster deployment we're reconciling, so it can restore the owner ref.

Example 2: provision job, created by clusterprovision controller.
hive.openshift.io/cluster-deployment-name=dgoodwin-gcp (exists today, propagated, but not really needed for our controller to restore)
hive.openshift.io/cluster-provision=dgoodwin-gcp-0-xgjz9
hive.openshift.io/install=true
The clusterprovision controller can see install=true so it knows this is a job it should restore owner reference for, and can see who the owner is from cluster-provision.

Example 3: ClusterState, created by clusterstate controller.
This resource has NO labels today, it's owner ref is cluster deployment.
Per the above it should get:
hive.openshift.io/cluster-provision=dgoodwin-gcp-0-xgjz9
I don't see the need for another label here, there should never be a ClusterState created for any other reason. Thoughts?

Example 4: SyncSet, created by user but handled by syncset controller.
Has *NO* labels today. Actually has no owner ref either. This is probably correct, I believe we target user created things to not be garbage collected if the cluster deployment is deleted, and can be created in any order.
No owner ref resoration needed. Cluster deployment is part of spec.

Example 5: SyncSetInstance, created by syncsetinstance controller.
Interestingly owned by ClusterDeployment, not it's parent SyncSet.
No labels today.
Per above should be modified to have: 
hive.openshift.io/cluster-deployment-name=dgoodwin-gcp
I don't see a need for an additional label here either.

Are we agreed we don't need a second label if there's only one way a resource should be created?

CC @twiest @staebler @csrwng

staebler:
> My understanding from our call this morning with some examples.
> 
> Firstly many of us learned that we probably should be propagating all labels from a "parent" object down to a "child" object when we create it, per kube patterns. We don't do this today in many cases but probably should, so we'll assume that is the case.
> 
> Second we discussed using two labels for any scenario of ownership. One essentially indicates the type or origin of the ownership, the second indicates the specific owner.

As shown in the examples, two labels is not always necessary. And there could be scenarios where two labels is not sufficient. Also, labels are convenient but not strictly necessary. For example, the ClusterState owner reference could be restored using the reference to the ClusterDeployment in the spec of the ClusterState. The labels are only a convenience to make it easier to filter, both in the controller and for the user.

> Example 1: imageset job, created by the clusterdeployment controller.
> hive.openshift.io/cluster-deployment-name=dgoodwin-gcp (already exists today)
> hive.openshift.io/imageset=true (already exists today, indicates this is an imageset job)
> The cluster deployment controller can process jobs, it knows that only it creates imageset jobs, it can tell if this job is from the cluster deployment we're reconciling, so it can restore the owner ref.
> 
> Example 2: provision job, created by clusterprovision controller.
> hive.openshift.io/cluster-deployment-name=dgoodwin-gcp (exists today, propagated, but not really needed for our controller to restore)
> hive.openshift.io/cluster-provision=dgoodwin-gcp-0-xgjz9
> hive.openshift.io/install=true
> The clusterprovision controller can see install=true so it knows this is a job it should restore owner reference for, and can see who the owner is from cluster-provision.
> 
> Example 3: ClusterState, created by clusterstate controller.
> This resource has NO labels today, it's owner ref is cluster deployment.
> Per the above it should get:
> hive.openshift.io/cluster-provision=dgoodwin-gcp-0-xgjz9
> I don't see the need for another label here, there should never be a ClusterState created for any other reason. Thoughts?

The label should be `hive.openshift.io/cluster-deployment-name=dgoodwin-gcp`.

> Example 4: SyncSet, created by user but handled by syncset controller.
> Has _NO_ labels today. Actually has no owner ref either. This is probably correct, I believe we target user created things to not be garbage collected if the cluster deployment is deleted, and can be created in any order.
> No owner ref resoration needed. Cluster deployment is part of spec.

Correct. It is the responsibility of the user to create owner references if they would like to have the SyncSet deleted when the ClusterDeployment is deleted. Hive will take no responsibility in restoring the user-added owner references.

 
> Example 5: SyncSetInstance, created by syncsetinstance controller.
> Interestingly owned by ClusterDeployment, not it's parent SyncSet.
> No labels today.
> Per above should be modified to have:
> hive.openshift.io/cluster-deployment-name=dgoodwin-gcp
> I don't see a need for an additional label here either.

The SyncSetInstance should be owned by both the SyncSet **and** the ClusterDeployment. It is a deficiency that it is not currently owned by both.

> Are we agreed we don't need a second label if there's only one way a resource should be created?

I agree that we do not **need** a label. I would argue that we **want** a label.
 
> CC @twiest @staebler @csrwng

twiest:
@dgoodwin @staebler @csrwng I created a [spreadsheet](https://docs.google.com/spreadsheets/d/1Zj9SadVzoLTyOrU8HD3absxvNyiwUj227yWX4EZf3iI/edit#gid=0) to map out the controller ownership creation calls.

I'd like to discuss the "proposed labels" on Monday and hammer out exactly what we'd like there.

But generally speaking, does this follow the pattern we discussed?

PR #741, opened 2019-12-10
Restore adding apply error message to syncsetinstance status condition.

abutcher:
The tempfile this aims to filter out comes from https://github.com/openshift/hive/blob/master/pkg/resource/helper.go#L15 https://github.com/openshift/hive/blob/master/pkg/resource/apply.go#L35.


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -856,6 +858,32 @@ func (r *ReconcileSyncSetInstance) clearUnknownObjectSyncCondition(syncSetCondit
...
+// Ex:
+//     applyErr="error when creating \"/tmp/apply-475927931\": namespaces \"openshift-am-config\" not found"
+// Returns:
+//     "namespaces \"openshift-am-config\" not found"
+func filterApplyError(applyErr string) string {
+	if !strings.Contains(applyErr, applyTempFile) {
+		return applyErr
+	}
+
+	splitErr := strings.Split(applyErr, " ")
```
staebler: How would you feel about just replacing the filename with a constant string like "file"? Or do you explicitly want to remove the "error when creating" part, too?

See https://play.golang.org/p/m3UreAoTzG2

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -871,7 +899,7 @@ func (r *ReconcileSyncSetInstance) setApplySyncConditions(resourceSyncConditions
 		// TODO: we cannot include the actual error here as it currently contains a temp filename which always changes,
```
staebler: This TODO should get removed or replaced with a comment describing why the filter is needed.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -1539,3 +1539,16 @@ func TestSSIReApplyDuration(t *testing.T) {
 		}
 	}
 }
+
+func TestFilterApplyError(t *testing.T) {
+	err1 := "error when creating \"/tmp/apply-475927931\": namespaces \"openshift-am-config\" not found"
+	expectedFilteredErr1 := "namespaces \"openshift-am-config\" not found"
+	if filterApplyError(err1) != expectedFilteredErr1 {
+		t.Fatalf("expected temporary file to be trimmed from error message")
```
staebler: This should use `t.Errorf` instead. You want the second part of the test to run even when the first part fails.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -1539,3 +1539,16 @@ func TestSSIReApplyDuration(t *testing.T) {
...
+	}
+
+	err2 := "unable to parse \"[ { \\\"op\\\": \\\"replace\\\", \\\"path\\\": \\\"/data/foo\\\", \\\"value\\\": \\\"baz-json\\\" ]\": yaml: did not find expected ',' or '}'"
```
staebler: You can save yourself a bunch of escapes here by using \`unable to parse ...\` instead of "unable to parse ...".

pkg/controller/syncsetinstance/syncsetinstance_controller.go
staebler: If you did want to remove the "error when creating" part, then I suggest something like the following.

https://play.golang.org/p/zgvVEm2-K9X

My concern is that splitting and rejoining is not very efficient.

abutcher: I updated it to remove the "error when creating" part using your second example as to me it seems related to creating the contents of the tempfile which is extra information as far as applying goes.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
abutcher: Removed.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
abutcher: Changed to `t.Errorf`.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
abutcher: Shortened. This was just an error that we generate from syncset I use as a failing patch test.

dgoodwin:
CC @matt-simons error reporting should be back in action for SyncSets, this will hopefully provide a lot more info.

matt-simons:
Fantastic 👍

PR #745, opened 2019-12-12
remove gcp project_id from clusterdeployment

staebler:
The GCP Project ID can be obtained from the GCP credentials. So, rather than requiring that the user supply the Project ID again in the ClusterDeployment, Hive will get the Project ID from the credentials.

https://jira.coreos.com/browse/CO-636

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -140,10 +140,8 @@ bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=aws myclu
 Credentials will be read from `~/.azure/osServicePrincipal.json` typically created via the `az login` command.
 
 ```bash
-bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=azure --azure-base-domain-resource-group-name=myresourcegroup --release-image=registry.svc.ci.openshift.org/origin/release:4.2 mycluster
+bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=azure --azure-base-domain-resource-group-name=myresourcegroup mycluster
 ```
-
-`--release-image` is used above as Azure installer support is only present in 4.2 dev preview builds.
 
 #### Create Cluster on GCP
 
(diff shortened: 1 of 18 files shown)


docs/using-hive.md
```diff
@@ -140,11 +140,9 @@ bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=aws myclu
 Credentials will be read from `~/.azure/osServicePrincipal.json` typically created via the `az login` command.
 
 ```bash
-bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=azure --azure-base-domain-resource-group-name=myresourcegroup --release-image=registry.svc.ci.openshift.org/origin/release:4.2 mycluster
+bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=azure --azure-base-domain-resource-group-name=myresourcegroup mycluster
```
joelddiaz: sneaky :)

staebler: The default release image is a 4.2 image now, so explicitly specifying the release image is no longer necessary.

staebler:
> did we ever figure out what a cross-project credentials file would look like (can they even exist)?

I don't know what a cross-project credentials file would look like. I vaguely remember that a creds file could have the project ID omitted, but I may be confusing GCP with Azure. In any event, I don't think it is too much of a burden to require the user to provide a project-specific credentials file.

joelddiaz:
> I don't know what a cross-project credentials file would look like. I vaguely remember that a creds file could have the project ID omitted, but I may be confusing GCP with Azure. In any event, I don't think it is too much of a burden to require the user to provide a project-specific credentials file.
> 
> Also, we can always add back the project ID as an optional field later if we find that there are users that need it.
> 
> Although I guess there is the concern that the user could create a cluster under the project specified in the install-config and then not be able to destroy the cluster if the project specified in the credentials is different.

I'm not sure that we actually can create creds that do not have the project set in the credentials (I just remember discussing the possibility).
I'm fine dealing with any consequences down the road as I prefer to not having us duplicating info in the spec that is already stored somewhere else.

staebler:
This is stuck with a missing image.

/retest

PR #752, opened 2019-12-12
Fix baremetal provisioning when run on OpenShift.

dgoodwin:
This was fine in Kube but in OpenShift both ssh to libvirt, and .cache
image copying broke due to the security constraints in OpenShift.

This change adds a home dir, changes it's permissions so it will work
with a random UID, and adds that random runtime UID to /etc/passwd so
SSH will work.

Also moved mounting of the ssh private key such that it will always
happen even if log gathering is disabled.

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -1149,15 +1149,40 @@ func waitForProvisioningStage(provision *hivev1.ClusterProvision, m *InstallMana
 
 func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
 
-	if err := os.MkdirAll("/root/.ssh", 0700); err != nil {
-		m.log.WithError(err).Error("error creating /root/.ssh directory")
+	// Add our potentially random UID to /etc/passwd so ssh works. Need to shell out here as
+	// the go libraries for user info appear to rely on /etc/passwd, which our user is not in.
+	// TODO: temporary work so we can use libvirt over ssh, this whole function can go once we're no longer doing that.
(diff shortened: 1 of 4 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -1149,15 +1149,40 @@ func waitForProvisioningStage(provision *hivev1.ClusterProvision, m *InstallMana
 
 func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
 
-	if err := os.MkdirAll("/root/.ssh", 0700); err != nil {
-		m.log.WithError(err).Error("error creating /root/.ssh directory")
+	// Add our potentially random UID to /etc/passwd so ssh works. Need to shell out here as
+	// the go libraries for user info appear to rely on /etc/passwd, which our user is not in.
+	// TODO: temporary work so we can use libvirt over ssh, this whole function can go once we're no longer doing that.
+	out, err := exec.Command("id", "-u").Output()
```
dhellmann: I think you can get this info from the user package. https://golang.org/pkg/os/user/#Current

dgoodwin: It appeared to not work if the user is not in passwd. :( I tried this in one of the earlier attempts.

dgoodwin:
For what it's worth most of this originates from the OpenShift documentation on how to work with arbitrary user IDs assigned by the platform: https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html#openshift-container-platform-specific-guidelines. The script there I ported into go to avoid complication around the entry point. But overall shouldn't be permanent.

PR #754, opened 2019-12-13
api url override

staebler:
Add a feature for overriding the API URL that Hive uses to communicate with the remote cluster. This is set via the .spec.controlPlaneConfig.apiURLOverride field in the ClusterDeployment.

When the field is set, the unreachable controller will attempt to connect to the override URL when checking for whether the remote cluster is reachable. If the override URL is reachable, then all other controllers that connect to the remote cluster will use the override URL. If the override URL is not reachable, then the unreachable controller will attempt to connect to the initial API URL set when the remote cluster was installed. If the initial URL is reachable, then all other controllers will use the initial URL when connecting to the remote cluster. If both the override URL and the initial URL are unreachable, then the unreachable controller will mark the remote cluster as unreachable. No other controllers will attempt to connect to the remote cluster in this case.

The "Unreachable" condition is true when both the initial URL and the override URL are unreachable. The message will include the errors from both connection attempts.

The "UnreachableAPIURLOverride" condition is true when the override URL is set and unreachable. The message will include only the error from the connection attempt using the override URL.

https://jira.coreos.com/browse/CO-669


staebler:
/hold

I want to add some unit tests to `remoteclient.builder` before this merges.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -345,6 +350,11 @@ type ControlPlaneConfigSpec struct {
 	// ServingCertificates specifies serving certificates for the control plane
 	// +optional
 	ServingCertificates ControlPlaneServingCertificateSpec `json:"servingCertificates,omitempty"`
+
+	// APIURLOverride is the URL that Hive will use to communicate with the API server
+	// of the remote cluster.
```
dgoodwin: Suggest something like: APIURLOverride is the optional URL override that Hive will transition to communicate with the API server of the remote cluster, once we are able to determine it is active.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -992,38 +994,34 @@ func (r *ReconcileClusterDeployment) fixupAdminKubeconfigSecret(secret *corev1.S
...
-		err = remoteClusterAPIClient.Get(context.Background(),
-			types.NamespacedName{Namespace: "openshift-console", Name: "console"}, routeObject)
-		if err != nil {
-			cdLog.WithError(err).Error("error fetching remote route object")
-			return err
-		}
-		cdLog.Debugf("read remote route object: %s", routeObject)
-		cd.Status.WebConsoleURL = "https://" + routeObject.Spec.Host
+	// The secondary API URL is used here because we always want to use the initial API URL immediately after installation.
+	remoteClientBuilder := r.remoteClusterAPIClientBuilder(cd).UseSecondaryAPIURL()
```
dgoodwin: This comment and code had me confused initially, I assumed SecondaryAPIURL meant OverrideAPIURL, I see that's not the case. The comment could possibly use a little clarification but the code is probably about as clear as it can be.

Are we guaranteed that at this point, the original API URL is still functional and no operators could have replaced it yet? It seems unlikely, but is it guaranteed?

staebler: In all but a restore scenario, we are guaranteed that Hive has not laid down anything that would make the original API URL no longer accessible. At this point, the .spec.installed field is still false, so no other controllers should be trying to communicate with the remote cluster.

That being said, I think it will be better to try both URLs to cover any scenarios where one may be inaccessible.

staebler: I reworked the conditions so that the clsuterdeployment controller does not need to explicitly use one of the URLs. The controller will now use the active URL.

staebler:
@dgoodwin I put up some new changes. Given the size of this PR, I kept the changes in a separate commit. Let me know when you've had a chance to review them, so that I can squash the commit.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -234,9 +234,9 @@ const (
 	// UnreachableCondition indicates that Hive is unable to establish an API connection to the remote cluster.
 	UnreachableCondition ClusterDeploymentConditionType = "Unreachable"
 
-	// UnreachableAPIURLOverrideCondition indicates that Hive is unable to establish an API connection to the remote
-	// cluster using the API URL override.
-	UnreachableAPIURLOverrideCondition ClusterDeploymentConditionType = "UnreachableAPIURLOverride"
+	// ActiveAPIURLOverrideCondition indicates that Hive is communicating with the remote cluster using the
+	// API URL overide.
```
dgoodwin: s/overide/override

staebler:
> just a typo, lets squash and merge.

fixed and squashed

staebler:
/hold cancel

Unit tests have been added.

PR #756, opened 2019-12-13
install: use merged pull secret in install-config

staebler:
https://jira.coreos.com/browse/CO-678

### DIFF
--- a/pkg/installmanager/testdata/pull-secret.json
+++ b/pkg/installmanager/testdata/pull-secret.json
@@ -1,0 +1,8 @@
+{
+  "auths": {
+    "cloud.openshift.com": {
+      "auth": "some-auth",
+      "email": "hive-user@hive.example.com"
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


pkg/installmanager/testdata/pull-secret.json
```diff
@@ -0,0 +1,8 @@
+{
```
joelddiaz: i'm not 100% clear on how all this merged pull secret stuff works, so forgive me if i'm missing the big picture

the way i'm reading the purpose of this file is to represent the generated merged pull secret for test cases. if so, can we name this file 'merged-pull-secret.json'? and could we have it have two pull secret entries to make it clear that this represents the state post-merge?

staebler: The installmanager does not have a concept of a merged pull secret. The installmanager just gets a pull secret file. There is no pre- and post-merge pull secret here.

joelddiaz: i now see that's the domain of the clusterdeployment controller

PR #758, opened 2019-12-16
Multi cloud dns management

joelddiaz:
https://issues.redhat.com/browse/CO-631

allow managing dns for multiple cloud providers simultaneously

update hiveconfig to allow defining multiple entries for each managed domain where each entry can contain a list of domains and any cloud-specific data (ie credentials) for managing those domains.

update the hive operator to send a JSON representation of the hiveconfig.spec.managedDomains fields as the configmap to be mounted for the dnsendpoint controller.

update hiveutil to process updates to hiveconfig.spec.managedDomains.

modify the dnsendpoint controller to read the new format of the file containing the managed domains configuration (now just a JSON dump of hiveConfig.Spec.ManagedDomains).

modify ReconcileDNSEndpoint struct to maintain a list of domains being managed.

add test cases to ensure proper processing of the updated configuration file format.

update existing test cases.

add DomainNotManagedCondition. this can happen in the event that a DNSZone object that was created when the hiveConfig has support for domain x.y.z, but an updated hiveConfig no longer has a config for x.y.z. Set a condition on the DNSZone to note this. also, allow dnsendpoint controller to always run so that it can set the DomainNotManaged condition when HiveConfig has no ManagedDomains configuration

handle updating an existing NS entry in GCP. when restarting hive, the dnsendpoint controller will re-reconcile all DNSZone objects. this leads down the path of creating/updating the DNS entries out in GCP as needed. the GCP nameserver code (for the dnsendpoint controller) would attempt to recreate the NS records, and that would error and bubble up back to the dnsendpoint controller, causing us to never complete reconciliation of the DNSZone. add a new UpdateResourceRecordSet call to the GCP client, and use it to allow the update path to succeed.

add live test case to cover the new update path.


joelddiaz:
@dgoodwin @staebler curious what your thoughts are on the expected behavior for these cases

https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR191
https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR213

staebler:
> @dgoodwin @staebler curious what your thoughts are on the expected behavior for these cases
> 
> https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR191
> https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR213

@joelddiaz They should not return an error. An error will cause the DNSZone to be re-queued. But the DNSZone will always fail, so there is no benefit in reconciling it again.

pkg/apis/hive/v1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -50,9 +50,13 @@ func NewClusterDeploymentValidatingAdmissionHook() *ClusterDeploymentValidatingA
 	if err != nil {
 		logger.WithError(err).Fatal("Unable to read managedDomains file")
 	}
-	logger.WithField("managedDomains", strings.Join(managedDomains, ",")).Info("Read managed domains")
+	domains := []string{}
+	for _, md := range managedDomains {
+		domains = append(domains, md.Domains...)
+	}
+	logger.WithField("managedDomains", strings.Join(domains, ",")).Info("Read managed domains")
```
staebler: I don't think we need to do a `strings.Join` here. The formatting of the field will already display the domains.

pkg/manageddns/manageddns.go
```diff
@@ -20,15 +21,16 @@ func ReadManagedDomainsFile() ([]string, error) {
...
-		s := scanner.Text()
-		s = strings.TrimSpace(s)
-		if s == "" {
-			continue
-		}
-		domains = append(domains, s)
+
+	domains := []hivev1.ManageDNSConfig{}
+
+	fileBytes, err := ioutil.ReadAll(file)
```
staebler: Use `ioutil.ReadFile` and get rid of the call to `os.Open`.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -89,21 +54,90 @@ func Add(mgr manager.Manager) error {
...
+		}
+	}
+
+	return nil
+}
+
+func newReconciler(mgr manager.Manager, kubeClient client.Client) (*ReconcileDNSEndpoint, []chan event.GenericEvent, error) {
+	nameServerScrapers := map[string]*nameServerScraper{}
+	nameServerQueryers := map[string]nameserver.Query{}
+	nameServerChangeNotifiers := []chan event.GenericEvent{}
```
staebler: Don't need a separate channel for each managed domain. All of the scrapers can write to the same channel.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -145,10 +179,19 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, nil
 	}
 
-	domain := instance.Spec.Zone
-	dnsLog = dnsLog.WithField("domain", domain)
+	fullDomain := instance.Spec.Zone
+	splitDomain := strings.Split(fullDomain, ".")
+	rootDomain := strings.Join(splitDomain[1:], ".")
```
staebler: This is not a sufficient method of determining the root domain. You could have a full domain of `abc.xyz.example.com` linked to the root domain of `example.com`.

I suggest that, instead of storing the scrapers and queries in maps, we store them in a slice. To find the right one to use, the controller will need to iterate over all of the scrapers until it finds one that works for the domain.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -89,21 +54,90 @@ func Add(mgr manager.Manager) error {
...
 // ReconcileDNSEndpoint reconciles a DNSEndpoint object
 type ReconcileDNSEndpoint struct {
 	client.Client
-	scheme            *runtime.Scheme
-	logger            log.FieldLogger
-	nameServerScraper *nameServerScraper
-	nameServerQuery   nameserver.Query
+	scheme               *runtime.Scheme
+	logger               log.FieldLogger
+	nameServerScraperMap map[string]*nameServerScraper
```
staebler: Combine the scrapers and the queries into the same data structure.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -221,18 +270,17 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
...
-	if gcpCredsSecretName != "" {
-		logger.Infof("using gcp creds for external DNS stored in %q secret", gcpCredsSecretName)
-		return nameserver.NewGCPQuery(c, gcpCredsSecretName)
+	if managedDomain.GCP != nil {
+		secretName := managedDomain.GCP.CredentialsSecretRef.Name
+		logger.Infof("using gcp creds for managed domain stored in %q secret", secretName)
+		return nameserver.NewGCPQuery(c, secretName)
 	}
-
+	logger.Warn("unsupported cloud for managing DNS")
```
staebler: This should be an error-level log entry. It is a logical error in the Hive code.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -162,7 +205,13 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
 	}
 
 	desiredNameServers := sets.NewString(instance.Status.NameServers...)
-	rootDomain, currentNameServers := r.nameServerScraper.GetEndpoint(domain)
+	_, currentNameServers := scraper.GetEndpoint(fullDomain) // remove returning root???
+
+	nsQuery, ok := r.nameServerQueryMap[rootDomain]
+	if !ok {
+		dnsLog.Warn("no name server query client found for domain")
```
staebler: This should be an error-level log entry. It is a failure of the validation to get a domain that is not covered by one of the managed domains.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -145,10 +179,19 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
...
+	fullDomain := instance.Spec.Zone
+	splitDomain := strings.Split(fullDomain, ".")
+	rootDomain := strings.Join(splitDomain[1:], ".")
+
+	dnsLog = dnsLog.WithField("domain", fullDomain)
 
-	if !r.nameServerScraper.HasBeenScraped(domain) {
+	scraper, ok := r.nameServerScraperMap[rootDomain]
+	if !ok {
+		dnsLog.Warn("no scraper for domain found, skipping reconcile")
```
staebler: This should be an error-level log entry. It is a failure of the validation to get a domain that is not covered by one of the managed domains.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -89,21 +54,90 @@ func Add(mgr manager.Manager) error {
...
+
+	logger := log.WithField("controller", controllerName)
+
+	managedDomains, err := manageddns.ReadManagedDomainsFile()
+	if err != nil {
+		logger.WithError(err).Error("could not read managed domains file")
+		return nil, nameServerChangeNotifiers, errors.Wrap(err, "could not read managed domains file")
+	}
+	if len(managedDomains) == 0 {
+		logger.Info("no managed domains for external DNS; controller disabled")
```
staebler: Remove ` for external DNS`.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -89,21 +54,90 @@ func Add(mgr manager.Manager) error {
...
+		return nil, nameServerChangeNotifiers, nil
+	}
+
+	for _, md := range managedDomains {
+		nameServerQuery := createNameServerQuery(kubeClient, logger, md)
+		if nameServerQuery == nil {
+			logger.Warnf("no platform found for managed DNS; controller disabled for domains %v", md.Domains)
```
staebler: Put the domains in as a logger field rather than in the warning text.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -89,21 +54,90 @@ func Add(mgr manager.Manager) error {
...
+			nameServerChanges <- event.GenericEvent{
+				Meta: &metav1.ObjectMeta{
+					Namespace: objectKey.Namespace,
+					Name:      objectKey.Name,
+				},
+			}
+		}
+		nameServerScraper := newNameServerScraper(logger, nameServerQuery, md.Domains, registerNameServerChange)
+		if err := mgr.Add(nameServerScraper); err != nil {
+			logger.WithError(err).Warnf("skipping handling of domains %v as unable to add name server scraper", md.Domains)
```
staebler: Put in domains as a logger field.

joelddiaz:
> > @dgoodwin @staebler curious what your thoughts are on the expected behavior for these cases
> > https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR191
> > https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR213
> 
> @joelddiaz They should not return an error. An error will cause the DNSZone to be re-queued. But the DNSZone will always fail, so there is no benefit in reconciling it again.

Right, that was my thinking in not returning an error. I guess the only way we would end up in this situation is a DNSZone that was created when domain x.y.z was managed, then later hiveconfig was updated to no longer support x.y.z.

Does setting a condition on the DNSZone sound reasonable?

staebler:
> > > @dgoodwin @staebler curious what your thoughts are on the expected behavior for these cases
> > > https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR191
> > > https://github.com/openshift/hive/pull/758/files#diff-f3f1cb0eefaf77b85b906e14900cca3dR213
> > 
> > 
> > @joelddiaz They should not return an error. An error will cause the DNSZone to be re-queued. But the DNSZone will always fail, so there is no benefit in reconciling it again.
> 
> Right, that was my thinking in not returning an error. I guess the only way we would end up in this situation is a DNSZone that was created when domain x.y.z was managed, then later hiveconfig was updated to no longer support x.y.z.
> 
> Does setting a condition on the DNSZone sound reasonable?

Sounds reasonable to me.

pkg/controller/dnsendpoint/dnsendpoint_controller_test.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/openshift/hive/pkg/apis"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
-	"github.com/openshift/hive/pkg/controller/dnsendpoint/nameserver"
+	////"github.com/openshift/hive/pkg/controller/dnsendpoint/nameserver"
```
staebler: Remove line.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -59,45 +58,44 @@ func Add(mgr manager.Manager) error {
...
-	nameServerScrapers := map[string]*nameServerScraper{}
-	nameServerQueryers := map[string]nameserver.Query{}
-	nameServerChangeNotifiers := []chan event.GenericEvent{}
+type nameServerTool struct {
+	scraper     *nameServerScraper
+	queryClient nameserver.Query
+}
+
+func newReconciler(mgr manager.Manager, kubeClient client.Client) (*ReconcileDNSEndpoint, chan event.GenericEvent, error) {
+	nameServerChangeNotifier := make(chan event.GenericEvent, 1024)
```
staebler: Don't need to create this until the end of this function. The function does not need to return a non-nil notifier when it returns a nil reconciler.

joelddiaz:
@staebler i've gone through your feedback (thanks!), went ahead an implemented the new condition (that required an update to our vendored testify/assert package as i was getting panics while running my new test cases).
lastly, i ran into an issue when restarting the controller. it wouldn't properly handle GCP DNSZone objects as the create/update path the controller took would only work for create. after some research, i came up with a series of API calls that allow updates to an existing NS record.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -145,10 +179,34 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
...
+		rootDomain, currentNameServers = nst.scraper.GetEndpoint(fullDomain)
+		if rootDomain != "" {
+			nsTool = r.nameServerTools[i]
+			break
+		}
+	}
+
+	if rootDomain == "" {
+		dnsLog.WithField("domain", fullDomain).Error("no scraper for domain found, skipping reconcile")
+		_, err := updateMissingDomainClientCondition(r.Client, dnsLog, instance, true)
```
staebler: This condition will not get set in the case where the managed DNS was removed completely from HiveConfig, since there will be no reconciler running. I think that we want to keep the reconciler running now even when managed DNS is disabled.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
```diff
@@ -221,18 +267,78 @@ func (r *ReconcileDNSEndpoint) Reconcile(request reconcile.Request) (reconcile.R
...
+}
+
+func updateMissingDomainClientCondition(c client.Client, logger log.FieldLogger, dnsZone *hivev1.DNSZone, missing bool) (bool, error) {
+	var status corev1.ConditionStatus
+	var reason string
+	var message string
+	if missing {
+		status = corev1.ConditionTrue
+		reason = "MissingDomainClient"
+		message = "Failed to find NameServer client for domain"
```
staebler: The messages for this condition should be understandable to the end user. The message should talk about there not be a managed DNS domain rather than a NameServer client.

pkg/apis/hive/v1/dnszone_types.go
```diff
@@ -133,6 +133,9 @@ const (
 	ZoneAvailableDNSZoneCondition DNSZoneConditionType = "ZoneAvailable"
 	// ParentLinkCreatedCondition is true if the parent link has been created
 	ParentLinkCreatedCondition DNSZoneConditionType = "ParentLinkCreated"
+	// MissingDomainClientCondition is true if we try to reconcile a DNSZone and the controller
+	// cannot find a nameServerScraper/nameServer Query client
+	MissingDomainClientCondition DNSZoneConditionType = "MissingDomainClient"
```
staebler: As an end user, I don't think that I would know what is meant be "MissingDomainClient".

pkg/controller/dnsendpoint/nameserver/gcp.go
```diff
@@ -194,7 +195,35 @@ func (q *gcpQuery) queryNameServer(gcpClient gcpclient.Client, managedZone strin
 
 // createNameServers creates the name servers for the specified domain in the specified managed zone.
 func (q *gcpQuery) createNameServers(gcpClient gcpclient.Client, managedZone string, domain string, values sets.String) error {
-	return gcpClient.AddResourceRecordSet(managedZone, q.resourceRecordSet(domain, values))
+	// First check whether there is an existing NS record
```
staebler: I would like to see attempt the create first and only query for existing records if the create fails with a status that indicates that the records already exist. See Delete for a similar example.

I expect that update is a lot more rare than insert, so I want to avoid performing a query every time when it is usually unneeded. I am a bit concerned about why the DNSEndpoint controller needed to update the NS records n restart.

pkg/controller/dnsendpoint/dnsendpoint_controller.go
joelddiaz: definitely. will modify.

pkg/controller/dnsendpoint/nameserver/gcp.go
joelddiaz: the code falls through https://github.com/openshift/hive/blob/v1/pkg/controller/dnsendpoint/dnsendpoint_controller.go#L172 which leads to the Create() call

joelddiaz:
@staebler i've renamed the condition to something more user-friendly, and changed the flow of the update handling to always just try a create first (and added some live tests to cover the update path)

pkg/controller/dnsendpoint/nameserver/gcp.go
staebler: Right, which it should not be. It should get caught by the case right above that where the current nameservers are equal to the desired nameservers. Since it is not, there is a bug. We can create a Jira card to take that bug on separate from this PR, if you think that the bug is not caused by the changes from this PR.

pkg/controller/dnsendpoint/nameserver/gcp_live_test.go
```diff
@@ -139,3 +145,26 @@ func (s *LiveGCPTestSuite) getCUT() *gcpQuery {
 		},
 	}
 }
+
+func (s *LiveGCPTestSuite) testCreateThenUpdate(tc *testCreateAndDeleteCase) {
```
staebler: Use a different type for the parameter to this function. Something like `testCreateThenUpdateCase`.

pkg/controller/dnsendpoint/nameserver/gcp_live_test.go
```diff
@@ -139,3 +145,26 @@ func (s *LiveGCPTestSuite) getCUT() *gcpQuery {
...
+	err := cut.Create(s.rootDomain, domain, sets.NewString(tc.createValues...))
+	if s.NoError(err, "unexpected error creating NS") {
+		defer func() {
+			err := cut.Delete(s.rootDomain, domain, sets.NewString(tc.deleteValues...))
+			s.NoError(err, "unexpected error deleting NS")
+		}()
+	}
+
+	// now test updating by re-issuing a Create()
+	err = cut.Create(s.rootDomain, domain, sets.NewString(tc.createValues...))
```
staebler: Change this to use a `tc.updateValues`, and add a case where the update is changing the values.

pkg/controller/dnsendpoint/nameserver/gcp.go
joelddiaz: I tried running this on the v1 branch, and created a GCP dnszone, and then restarted the controller and see the controller spin as it keeps trying to re-create the NS entries.

```
INFO[0005] reconciling dns endpoint                      controller=dnsendpoint dnszone=testzoneg namespace=hive
INFO[0005] creating NS record                            controller=dnsendpoint dnszone=testzoneg domain=test.jdomain.gcp.hive.openshift.com namespace=hive
ERRO[0006] error creating NS record                      controller=dnsendpoint dnszone=testzoneg domain=test.jdomain.gcp.hive.openshift.com error="error creatin
g the name server: googleapi: Error 409: The resource 'entity.change.additions[0]' named 'test.jdomain.gcp.hive.openshift.com. (NS)' already exists, alreadyExist
s" namespace=hive
INFO[0006] reconcile complete                            controller=dnsendpoint dnszone=testzoneg domain=test.jdomain.gcp.hive.openshift.com elapsed=413.038206ms
 namespace=hive
INFO[0008] reconciling dns endpoint                      controller=dnsendpoint dnszone=testzoneg namespace=hive
INFO[0008] creating NS record                            controller=dnsendpoint dnszone=testzoneg domain=test.jdomain.gcp.hive.openshift.com namespace=hive
ERRO[0008] error creating NS record                      controller=dnsendpoint dnszone=testzoneg domain=test.jdomain.gcp.hive.openshift.com error="error creatin
g the name server: googleapi: Error 409: The resource 'entity.change.additions[0]' named 'test.jdomain.gcp.hive.openshift.com. (NS)' already exists, alreadyExist
s" namespace=hive
INFO[0008] reconcile complete                            controller=dnsendpoint dnszone=testzoneg domain=test.jdomain.gcp.hive.openshift.com elapsed=394.647008ms
 namespace=hive
```
So this doesn't appear to be new behavior. I gave a quick glance at the code and it appears to be due to this line needing to run before a re-reconcile can skip the Create() path
https://github.com/openshift/hive/blob/v1/pkg/controller/dnsendpoint/dnsendpoint_controller.go#L178

joelddiaz:
@staebler ok, i updated the test cases to do a more proper update flow (separate commit). if this all looks good, then i'll squash and repush

pkg/controller/dnsendpoint/nameserver/gcp.go
staebler: I created https://issues.redhat.com/browse/CO-696 to track this.

staebler: It appears that the gpc nameserver scaper is just plain broken.

staebler:
> @staebler ok, i updated the test cases to do a more proper update flow (separate commit). if this all looks good, then i'll squash and repush

Squash away!

PR #760, opened 2019-12-17
Add controller and operator health check

abutcher:
Add basic health endpoint w/ liveness checks in templates.

```
$ curl -v localhost:8080/debug/health
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /debug/health HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 2
< Content-Type: application/json; charset=utf-8
< Date: Tue, 17 Dec 2019 18:50:47 GMT
< 
* Connection #0 to host localhost left intact
{}
```


Gopkg.lock
```diff
@@ -241,6 +241,18 @@
   revision = "d2133a1ce379ef6fa992b0514a77146c60db9d1c"
   version = "v1.1.0"
 
+[[projects]]
```
staebler: The changes in this file should go in the other commit.

cmd/manager/main.go
```diff
@@ -125,6 +127,10 @@ func newRootCommand() *cobra.Command {
 				log.Fatal(err)
 			}
 
+			log.Printf("Starting healthz endpoint.")
+
+			go http.ListenAndServe(":8080", nil)
```
staebler: Would be nice to include a comment that associates this with go-healthcheck so that in 2 years we'll be able to figure out why this line of code is needed.

cmd/manager/main.go
```diff
@@ -125,6 +127,10 @@ func newRootCommand() *cobra.Command {
 				log.Fatal(err)
 			}
 
+			log.Printf("Starting healthz endpoint.")
```
staebler: s|healthz|debug/health|

config/manager/deployment.yaml
```diff
@@ -40,4 +40,10 @@ spec:
         env:
         - name: CLI_CACHE_DIR
           value: /var/cache/kubectl
+        livenessProbe:
+          httpGet:
+            path: /debug/health
+            port: 8080
+          failureThreshold: 1
+          periodSeconds: 10
```
staebler: Add an `initialDelaySeconds` so that the liveness probe is not racing with the manager startup.

abutcher: Added a 10 second delay. Let me know if you think that's enough.

Gopkg.lock
abutcher: Moved to other commit.

cmd/manager/main.go
```diff
@@ -125,6 +127,11 @@ func newRootCommand() *cobra.Command {
 				log.Fatal(err)
 			}
 
+			// Start http server which will enable the /debug/health handler from go-healthcheck
+			log.Printf("Starting debug/health endpoint.")
```
dgoodwin: Any reason this is using printf vs one of the normal log levels?

abutcher: All the other “starting up” lines were printf. I can info log them.

abutcher: Updated!

staebler: For reference, the difference between `Printf` and `Infof` is that `Printf` gets output to the log regardless of what the log level is set to. They both yield an info-level log entry, though.

PR #761, opened 2019-12-19
generate clientset

staebler:
The clientset is useful in setting up an aggregated API server for v1alpha1 that proxies to the v1 CRDs.


pkg/apis/apis.go
```diff
@@ -1,5 +1,6 @@
 // Generate deepcopy for apis
 //go:generate go run ../../vendor/k8s.io/code-generator/cmd/deepcopy-gen/main.go -O zz_generated.deepcopy -i ./... -h ./../../hack/boilerplate.go.txt
+//go:generate go run ../../vendor/k8s.io/code-generator/cmd/client-gen/main.go --clientset-name "clientset" --input "github.com/openshift/hive/pkg/apis/hive/v1" --input-base "" --output-package "github.com/openshift/hive/pkg/client/clientset-generated" -h ./../../hack/boilerplate.go.txt
```
joelddiaz: the line above this is related to the generated deepcopy bits.

this line here is related to the generated client. can we add a comment above this line indicating its purpose?

PR #767, opened 2020-01-09
Add configurable syncset reapply interval.

abutcher:


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -269,13 +274,25 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, applyErr
 	}
 
-	reapplyDuration := ssiReApplyDuration(ssi)
+	reapplyDuration := ssiReApplyDuration(ssi, r.reapplyInterval)
 	return reconcile.Result{RequeueAfter: reapplyDuration}, nil
 }
 
+func getReApplyInterval(log log.FieldLogger) time.Duration {
+	if envReapplyInterval := os.Getenv(reapplyIntervalEnvKey); len(envReapplyInterval) > 0 {
```
staebler: Let's take this opportunity to be consistent about whether we want to use "Reapply" or "ReApply". It looks like the function names use "ReApply" and the variable names use "Reapply". Let's pick one.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -71,19 +72,41 @@ type Applier interface {
...
-	return AddToManager(mgr, NewReconciler(mgr))
+	logger := log.WithField("controller", controllerName)
+
+	// Get hiveConfig and parse hiveConfig.Spec.SyncSetReapplyInterval
+	tempClient, err := client.New(mgr.GetConfig(), client.Options{Scheme: mgr.GetScheme()})
+	if err != nil {
+		return err
+	}
+	hiveConfig := &hivev1.HiveConfig{}
+	err = tempClient.Get(context.TODO(), types.NamespacedName{Name: hiveConfigName}, hiveConfig)
```
staebler: The controllers should never get the HiveConfig. The hive-operator should pass the values from the HiveConfig into the controllers via environment variables or command-line arguments .

abutcher: Updated to set an environment variable from hiveconfig via the operator

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -71,19 +73,30 @@ type Applier interface {
...
-// Add creates a new SyncSet controller and adds it to the manager with default RBAC. The manager will set fields on the
+// Add creates a new SyncSetInstance controller and adds it to the manager with default RBAC. The manager will set fields on the
 // controller and start it when the manager starts.
 func Add(mgr manager.Manager) error {
-	return AddToManager(mgr, NewReconciler(mgr))
+	logger := log.WithField("controller", controllerName)
+	reapplyInterval := defaultReapplyInterval
+	var err error
+	if envReapplyInterval := os.Getenv(reapplyIntervalEnvKey); len(envReapplyInterval) > 0 {
+		reapplyInterval, err = time.ParseDuration(envReapplyInterval)
```
staebler: You do not want to change `reapplyInterval` when there is an error parsing the environment variable.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -71,19 +73,30 @@ type Applier interface {
...
+		if err != nil {
+			log.WithError(err).Errorf("unable to parse %s: %s", reapplyIntervalEnvKey, envReapplyInterval)
```
staebler: Log `envReappyInterval` as a field rather than in the error message.
```
log.WithError(err).WithField("reapplyInterval", envReapplyInterval).Errorf("unable to parse %s", reapplyIntervalEnvKey)
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -71,19 +73,30 @@ type Applier interface {
...
+		}
+	}
+	log.Infof("SyncSetReapplyInterval: %s", reapplyInterval.String())
```
staebler: Log the reapply interval as a field.
```
log.WithField("reapplyInterval", reapplyInterval).Info("Reapply interval set")
```
Note also that the "SyncSet" in the log message is redundant with the field that specifies the controller name.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -21,11 +21,12 @@ import (
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
-	"k8s.io/client-go/rest"
 	"strings"
 	"testing"
 	"time"
 
+	"k8s.io/client-go/rest"
```
staebler: Move this down in with the other k8s.io imports.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
abutcher: I should have been returning the error here. In that case it would be okay to change `reapplyInterval` right?

staebler: Yes.

staebler: While in that code, can you also bring `var err error` done into the scope that it will be used.

PR #770, opened 2020-01-09
WIP: V1 Migration Tooling

dgoodwin:


hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+
+oc scale -n hive deployment.v1.apps/hive-operator --replicas=0
+oc scale -n hive deployment.v1.apps/hive-controllers --replicas=0
+oc scale -n hive deployment.v1.apps/hiveadmission --replicas=0
+
+# TODO: wait until no pods are running
+
+mkdir -p $WORKDIR/
+
+HIVE_TYPES=( checkpoints clusterdeployments clusterdeprovisionrequests clusterimagesets clusterprovisions clusterstates dnsendpoints dnszones hiveconfig selectorsyncidentityprovider selectorsyncset syncidentityprovider syncsetinstance syncset )
```
staebler: If you don't want to hard-code this,
```
readarray -t HIVE_TYPES <<< \
"$(oc get crd -o json | \
jq -r '.items[] | select(.spec.group="hive.openshift.io") | select(.spec.version="v1alpha1") | .spec.names.plural')"
```

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+for i in "${HIVE_TYPES[@]}"
+do
+	:
+	echo "Storing all ${i} in ${WORKDIR}/${i}.yaml"
+	oc get ${i}.hive.openshift.io -A -o json | jq '.items | .[] |
+		del(.status) |
```
staebler: I don't think that we want to strip out the status. We will need it in some cases.

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+		del(.metadata.annotations."kubectl.kubernetes.io/last-applied-configuration") |
+		del(.metadata.creationTimestamp) |
```
staebler: Do we really need to strip this kind of stuff out? I would expect the API server to strip it out for us when we go to create the objects. Or is the purpose of stripping it out to reduce the size of the yaml?

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+		del(.metadata.generation) |
+		del(.metadata.resourceVersion) |
+		del(.metadata.ownerReferences) |
```
staebler: I would expect us to keep the owner references in the yaml so that we can use them to re-apply the owner references on the restored resources.

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+# TODO: wait until no pods are running
+
+mkdir -p $WORKDIR/
+
+HIVE_TYPES=( checkpoints clusterdeployments clusterdeprovisionrequests clusterimagesets clusterprovisions clusterstates dnsendpoints dnszones hiveconfig selectorsyncidentityprovider selectorsyncset syncidentityprovider syncsetinstance syncset )
+for i in "${HIVE_TYPES[@]}"
+do
+	:
+	echo "Storing all ${i} in ${WORKDIR}/${i}.yaml"
+	oc get ${i}.hive.openshift.io -A -o json | jq '.items | .[] |
```
staebler: You can simplify `.items | .[]` down to `.items[]`.

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+
+for i in "${HIVE_TYPES[@]}"
+do
+	:
+	# Only attempt to remove finalizers from types we found results for:
+	if [ -s ${WORKDIR}/${i}.yaml ]
+	then
+		echo "Removing finalizers from all ${i}"
+		oc get ${i}.hive.openshift.io -A -o json | jq '.items | .[] |
+			.metadata.finalizers = null' | oc apply -f -
```
staebler: Is there a reason to explicitly set the finalizers to null in the yaml being applied rather than deleting the field?

hack/v1migration/phase2.sh
```diff
@@ -0,0 +1,48 @@
...
+
+HIVE_TYPES=( checkpoints clusterdeployments clusterdeprovisionrequests clusterimagesets clusterprovisions clusterstates dnsendpoints dnszones hiveconfig selectorsyncidentityprovider selectorsyncset syncidentityprovider syncsetinstance syncset )
+for i in "${HIVE_TYPES[@]}"
+do
+	:
+	echo "Deleting all ${i}"
+	oc delete ${i}.hive.openshift.io -A --all
+done
+
+oc delete customresourcedefinition.apiextensions.k8s.io/checkpoints.hive.openshift.io
```
staebler: Why is this not being done in an array loop over HIVE_TYPES?

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# TODO: wait until no pods are running
+
+mkdir -p $WORKDIR/
+
+HIVE_TYPES=( checkpoints clusterdeployments clusterdeprovisionrequests clusterimagesets clusterprovisions clusterstates dnsendpoints dnszones hiveconfig selectorsyncidentityprovider selectorsyncset syncidentityprovider syncsetinstance syncset )
+for i in "${HIVE_TYPES[@]}"
+do
+	:
+	echo "Storing all ${i} in ${WORKDIR}/${i}.yaml"
```
staebler: How would you feel about storing each resource in its own file rather than one file per kind?

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+	oc get ${i}.hive.openshift.io -A -o json | jq '.items | .[] |
+		del(.status) |
+		del(.metadata.annotations."kubectl.kubernetes.io/last-applied-configuration") |
+		del(.metadata.creationTimestamp) |
+		del(.metadata.generation) |
+		del(.metadata.resourceVersion) |
+		del(.metadata.ownerReferences) |
+		del(.metadata.selfLink) |
+		del(.metadata.uid)' > ${WORKDIR}/${i}.yaml
```
staebler: This file is json not yaml (understanding that json is a subset of yaml).

hack/v1migration/phase1.sh
```diff
@@ -0,0 +1,50 @@
...
+done
+
```
staebler: I would like an added check here that fetches the resources again and compares them to what is stored in the yaml before proceeding.

hack/v1migration/phase2.sh
```diff
@@ -0,0 +1,48 @@
...
+    exit 1
+fi
+
+
+HIVE_TYPES=( checkpoints clusterdeployments clusterdeprovisionrequests clusterimagesets clusterprovisions clusterstates dnsendpoints dnszones hiveconfig selectorsyncidentityprovider selectorsyncset syncidentityprovider syncsetinstance syncset )
+for i in "${HIVE_TYPES[@]}"
+do
+	:
+	echo "Deleting all ${i}"
+	oc delete ${i}.hive.openshift.io -A --all
```
staebler: Before deleting, I need two checks.
(1) The user must acknowledge that they meant to run the delete.
(2) We must check that the resource stored in yaml and matches the yaml.

hack/v1migration/README.md
```diff
@@ -0,0 +1,20 @@
...
+  1. Run ./hack/v1migration/phase2.sh [workdir] to delete all Hive CRs and CRDs.
+  1. Deploy Hive v1
+  1. ./hack/make-apiserver-certs.sh
+  1. ./hack/deploy-apiserver.sh
+  1. Scale everything down to 0 again, but keep hiveadmission and hiveapi up.
+  1. kubectl apply -f [workdir] # run twice due to problems recognizing types initially?
+
+## TODO
+
+  1. Restored ClusterDeployment has no infraID.
```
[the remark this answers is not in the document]
dgoodwin: Ah so not until the controllers are up again.

hack/v1migration/phase1.sh
dgoodwin: My understanding was we were going to rollout @twiest 's work and let that use labels to restore owner refs.

hack/v1migration/phase1.sh
dgoodwin: These examples are partly based on our scripts for duplicating clusters, I actually thought the apiserver would reject a create with a creationtimestamp, uid, etc. I'd have to test though.

hack/v1migration/phase1.sh
dgoodwin: I believe I tested this and found it did not clear them unless they're explicitly set to null. That might have been related to using apply though and it's possible a patch of some kind would avoid it.

hack/v1migration/phase2.sh
dgoodwin: That did not occur to me, will update it.

hack/v1migration/phase1.sh
dgoodwin: I'd like to keep the scripts as simple as possible, is this easy to do? I can't immediately think of a way aside from querying one at a time, the I'm not crazy about the jq solutions to split I'm seeing.

Are there specific benefits? I couldn't think of any.

hack/v1migration/phase1.sh
dgoodwin: I think I'm going to change it to backup everything but the last-applied annotation, and then strip out what we need with jq when submitting later.

hack/v1migration/phase1.sh
dgoodwin: What is the thought behind this one?

hack/v1migration/phase2.sh
dgoodwin: (1) makes sense, what do you mean by (2) and why?

dgoodwin: I think I figured it out, probably to see if anything restored finalizers on the objects, in which case we're in trouble.

Agreed, but we should probably skip this in phase 1.

hack/v1migration/phase1.sh
dgoodwin: I may have figured out why on the other script but is there still a reason to do this here, immediately after fetching the yaml, but before we've even removed finalizers?

dgoodwin:
Updated with several fixes from the review, per slack convo handing over to @staebler

hack/v1migration/README.md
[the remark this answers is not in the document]
staebler: The controllers being up won't help.

hack/v1migration/phase1.sh
staebler: That would be my preference. I would like to preserve everything that we can.

hack/v1migration/phase1.sh
staebler: Ah, you may be right.

hack/v1migration/phase1.sh
staebler: That will require us to apply the labels to the v1alpha1 objects. It doesn't strike me as too much harder to restore the owner references directly than it is to apply the labels. And then we do not *need* to have Thomas's work.

hack/v1migration/phase1.sh
staebler: Sounds reasonable.

hack/v1migration/phase1.sh
staebler: My two motivations are (1) making it easier to manually validate the data collected and (2) making it easier to re-apply the resources back.

hack/v1migration/phase1.sh
staebler: It would be a safety check that verifies that everything was correctly stored in the files before we start doing destructive things.

hack/v1migration/phase2.sh
staebler: A safety check to make sure that we have actually stored the resource to file before we delete it. In the extreme, I am thinking of a case where phase 2 is accidentally executed before phase 1.

staebler:
/close

expanded upon in #776

PR #772, opened 2020-01-10
WIP: Baremetal Provisioning Libvirt In a Pod

dgoodwin:


pkg/install/generate.go
```diff
@@ -291,11 +291,45 @@ func InstallerPodSpec(
...
+							// host
+							Value: "16509",
+						},
+					},
+				*/
+			},
+		)
+		// TODO: host networking is used to access the networks for target clusters.
+		// instead we should provide access: https://docs.openshift.com/container-platform/4.2/networking/multiple-networks/attaching-pod.html
+		hostNetwork := true
```
staebler: `hostNetwork = true`

pkg/install/generate.go
```diff
@@ -291,11 +291,45 @@ func InstallerPodSpec(
...
 	}
 
+	var hostNetwork bool
+	if cd.Spec.Platform.BareMetal != nil {
+		containers = append(containers,
+			corev1.Container{
+				Name: "libvirt",
+				// TODO: need an official image for this
+				Image:           "docker.io/djzager/libvirt:netstart",
+				ImagePullPolicy: corev1.PullAlways,
```
joelddiaz: do we always want 'PullAlways'? or is this temporary?

dgoodwin: Temporary, not sure what the longterm image we use here will be, although I think we usually use PullAlways don't we?

joelddiaz: i was just thinking if whatever container repo that houses these images becomes inaccessible, does that mean that installations grind to a halt even though the image might be cached?

dgoodwin: Could be a problem, but across the board as I believe we're using this for almost everything.

dgoodwin: Seems like we want this for development setups so we always get our latest image, but in prod/stage we use hashes so IfNotPresent would work fine there. We should probably make this configurable somewhere, HiveConfig perhaps.

dgoodwin:
This work might be dead for time being, and at very least we need to restore the ssh approach I removed in here. Closing for now and focusing on solidifying the ssh libvirt approach.

PR #777, opened 2020-01-17
Extract URLs from kubeconfig for adopted clusters without a provision.

dgoodwin:
Adopted clusters do not have a ClusterProvision, but this code was the
only place we set Status.APIURL and Status.WebConsoleURL.

Will now ensure these are set even if the ClusterDeployment is marked
installed.

The ClusterProvision processing still has similar code just to avoid an
extra status update on intial processing of a completed install.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -285,6 +285,25 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
 	return r.reconcile(request, cd, cdLog)
 }
 
+func (r *ReconcileClusterDeployment) postProcessAdminKubeconfig(cd *hivev1.ClusterDeployment,
+	cdLog log.FieldLogger) error {
+
+	adminKubeconfigSecret := &corev1.Secret{}
+	if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name}, adminKubeconfigSecret); err != nil {
+		cdLog.WithError(err).Error("failed to get admin kubeconfig secret")
+		return err
+	}
+	if err := r.fixupAdminKubeconfigSecret(adminKubeconfigSecret, cdLog); err != nil {
+		cdLog.WithError(err).Error("failed to fix up admin kubeconfig secret")
+		return err
+	}
+	if err := r.setAdminKubeconfigStatus(cd, cdLog); err != nil {
+		cdLog.WithError(err).Error("failed to set admin kubeconfig status")
+		return err
+	}
+	return nil
+}
+
 func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (result reconcile.Result, returnErr error) {
 	// Set platform label on the ClusterDeployment
 	if platform := getClusterPlatform(cd); cd.Labels[hivev1.HiveClusterPlatformLabel] != platform {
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -387,6 +387,26 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 
 		cdLog.Debug("cluster is already installed, no processing of provision needed")
 		r.cleanupInstallLogPVC(cd, cdLog)
+
+		if cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name != "" && (cd.Status.WebConsoleURL == "" || cd.Status.APIURL == "") {
```
staebler: Check `cd.Spec.ClusterMetadata != nil` also.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -387,6 +387,26 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+			adminKubeconfigSecret := &corev1.Secret{}
+			if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name}, adminKubeconfigSecret); err != nil {
```
staebler: Can we create a function to call to do this work that can be called from `reconcileCompletedProvision` as well rather than duplicating the code?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: Add a unit test that triggers this.

dgoodwin:
Updated.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -667,19 +701,13 @@ func (r *ReconcileClusterDeployment) reconcileCompletedProvision(cd *hivev1.Clus
...
-			cdLog.WithError(err).Error("failed to get admin kubeconfig secret")
-			return reconcile.Result{}, err
-		}
-		if err := r.fixupAdminKubeconfigSecret(adminKubeconfigSecret, cdLog); err != nil {
-			cdLog.WithError(err).Error("failed to fix up admin kubeconfig secret")
-			return reconcile.Result{}, err
-		}
-		if err := r.setAdminKubeconfigStatus(cd, cdLog); err != nil {
-			cdLog.WithError(err).Error("failed to set admin kubeconfig status")
+		err := r.postProcessAdminKubeconfig(cd, cdLog)
```
staebler: The controller will end up updating the status of the cluster deployment twice--once in this function call and once a few lines later. I suggest that we move the status update out of the `postProcessAdminKubeconfig` method.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -387,6 +410,17 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 
 		cdLog.Debug("cluster is already installed, no processing of provision needed")
 		r.cleanupInstallLogPVC(cd, cdLog)
+
+		if cd.Spec.ClusterMetadata != nil &&
+			cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name != "" &&
+			(cd.Status.WebConsoleURL == "" || cd.Status.APIURL == "") {
```
staebler: This is not something that I expect us to do for this PR, but we should get in the habit of enforcing that status fields are set to the correct value on each reconcile instead of just simply set to something. So, here in particular, it would be better for the controller to determine the web console URL on every reconcile and check that the value set in the `WebConsoleURL` matches the correct value.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Updated.

PR #779, opened 2020-01-22
Add labels used for ownership reconciliation

twiest:
- [x] Add relevant labels where controller owners are added so that
      controller ownership relationships can be reconciled
- [x] Consolidate cluster-deployment-name label to a single golang constant
- [x] Add unit tests to ensure labels are set correctly
- [x] Manually test that labels are applied.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -27,8 +27,62 @@ const (
 	// UninstallJobLabel is the label used for artifacts specific to Hive cluster deprovision.
 	UninstallJobLabel = "hive.openshift.io/uninstall"
 
-	// ClusterDeploymentNameLabel is the label that is used to identify the installer pod of a particular cluster deployment
+	// ClusterDeploymentNameLabel is the label that is used to identify a relationship to a given cluster deployment object.
 	ClusterDeploymentNameLabel = "hive.openshift.io/cluster-deployment-name"
+
+	// ClusterDeprovisionNameLabel is the label that is used to identify a relationship to a given cluster deprovision object.
+	ClusterDeprovisionNameLabel = "hive.openshift.io/cluster-deprovision-name"
+
+	// ClusterProvisionNameLabel is the label that is used to identify a relationship to a given cluster provision object.
+	ClusterProvisionNameLabel = "hive.openshift.io/cluster-provision-name"
+
+	// PVCTypeLabel is the label that is used to identify what a PVC is being used for.
+	PVCTypeLabel = "hive.openshift.io/pvc-type"
+
+	// PVCTypeInstallLogs is used as a value of PVCTypeLabel that says the PVC specifically stores installer logs.
+	PVCTypeInstallLogs = "installlogs"
+
+	// JobTypeLabel is the label that is used to identify what a Job is being used for.
+	JobTypeLabel = "hive.openshift.io/job-type"
+
+	// JobTypeImageSet is used as a value of JobTypeLabel that says the Job is specifically running to determine which imageset to use.
+	JobTypeImageSet = "imageset"
+
+	// JobTypeDeprovision is used as a value of JobTypeLabel that says the Job is specifically running the deprovisioner.
+	JobTypeDeprovision = "deprovision"
+
+	// JobTypeProvision is used as a value of JobTypeLabel that says the Job is specifically running the provisioner.
+	JobTypeProvision = "provision"
+
+	// DNSZoneTypeLabel is the label that is used to identify what a DNSZone is being used for.
+	DNSZoneTypeLabel = "hive.openshift.io/dnszone-type"
+
+	// DNSZoneTypeChild is used as a value of DNSZoneTypeLabel that says the DNSZone is specifically used as the forwarding zone for the target cluster.
+	DNSZoneTypeChild = "child"
+
+	// SecretTypeLabel is the label that is used to identify what a Secret is being used for.
+	SecretTypeLabel = "hive.openshift.io/secret-type"
+
+	// SecretTypeMergedPullSecret is used as a value of SecretTypeLabel that says the secret is specifically used for storing a pull secret.
+	SecretTypeMergedPullSecret = "merged-pull-secret"
+
+	// SecretTypeKubeConfig is used as a value of SecretTypeLabel that says the secret is specifically used for storing a kubeconfig.
+	SecretTypeKubeConfig = "kubeconfig"
+
+	// SecretTypeKubeAdminCreds is used as a value of SecretTypeLabel that says the secret is specifically used for storing kubeadmin credentials.
+	SecretTypeKubeAdminCreds = "kubeadmincreds"
+
+	// SyncSetTypeLabel is the label that is used to identify what a SyncSet is being used for.
+	SyncSetTypeLabel = "hive.openshift.io/syncset-type"
+
+	// SyncSetTypeControlPlaneCerts is used as a value of SyncSetTypeLabel that says the syncset is specifically used to distribute control plane certificates.
+	SyncSetTypeControlPlaneCerts = "controlplanecerts"
+
+	// SyncSetTypeRemoteIngress is used as a value of SyncSetTypeLabel that says the syncset is specifically used to distribute remote ingress information.
+	SyncSetTypeRemoteIngress = "remoteingress"
+
+	// SyncSetTypeIdentityProvider is used as a value of SyncSetTypeLabel that says the syncset is specifically used to distribute identity provider information.
+	SyncSetTypeIdentityProvider = "identityprovider"
 
 	// GlobalPullSecret is the environment variable for controllers to get the global pull secret
 	GlobalPullSecret = "GLOBAL_PULL_SECRET"
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -567,6 +567,8 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		provision.Spec.PrevInfraID = &cd.Spec.ClusterMetadata.InfraID
 	}
 
+	cdLog.WithField("derivedObject", provision.Name).Debug("Setting label on derived object")
+	controllerutils.AddLabel(provision, constants.ClusterDeploymentNameLabel, cd.Name)
 	if err := controllerutil.SetControllerReference(cd, provision, r.scheme); err != nil {
 		cdLog.WithError(err).Error("could not set the owner ref on provision")
 		return reconcile.Result{}, err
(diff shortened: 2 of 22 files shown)


twiest:
@dgoodwin @staebler FYI, this is ready for review. I'm putting a hold on it so that I can do manual testing today.

/hold

pkg/constants/constants.go
```diff
@@ -27,9 +27,63 @@ const (
...
+	JobTypeDeprovision = "deprovision"
+
+	// JobTypeProvision is used as a value of JobTypeLabel that says the Job is specifically running the provisioner.
+	JobTypeProvision = "provision"
+
+	// DNSZoneTypeLabel is the label that is used to identify what a DNSZone is being used for.
+	DNSZoneTypeLabel = "hive.openshift.io/dnszone-type"
+
+	// DNSZoneTypeTargetCluster is used as a value of DNSZoneTypeLabel that says the DNSZone is specifically used as the forwarding zone for the target cluster.
+	DNSZoneTypeTargetCluster = "targetcluster"
```
dgoodwin: What do folks think of "forwarding" for the value here instead.

twiest: @dgoodwin we can definitely do that. Just so it's clear, this is the dns zone that is forwarded to from the main "parent" zone.

So maybe `forwarded` would be clearer?

Here are some other alternatives that I thought of:
- `managedcluster`
- `target`
- `child` # since the zone forwarding to it is called the `parent` zone

joelddiaz: 'child' seems reasonable since we already have the 'LinkToParent' terminoligy

twiest:
Manual testing complete (and successful). Canceling hold.

/hold cancel

pkg/constants/constants.go
twiest: @dgoodwin I'm ready to change this to whatever your decision is.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -564,6 +564,8 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		provision.Spec.PrevInfraID = &cd.Spec.ClusterMetadata.InfraID
 	}
 
+	cdLog.WithField("ownedObject", provision.Name).Debugf("Setting labels that are used to restore ownership of the object")
```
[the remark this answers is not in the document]
twiest: The important part is that the `provision` is having a label put on saying it's related to the cluster deployment (in this case, specifically owned by it).

What verbiage would you prefer?

Is this better?
```golang
cdLog.WithField("ownedObject", provision.Name).Debugf("Setting ownership label of the object")
```

Or would you prefer this?
```golang
cdLog.WithField("relatedObject", provision.Name).Debug("Setting label on related object")
```

pkg/constants/constants.go
dgoodwin: "child" works for me. Thanks!

pkg/controller/clusterdeployment/clusterdeployment_controller.go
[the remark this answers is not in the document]
dgoodwin: I'd go with derived object verbiage.

twiest:
@dgoodwin @staebler I believe I've addressed your feedback. PTAL.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -143,6 +143,17 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			validate: func(c client.Client, t *testing.T) {
 				provisions := getProvisions(c)
 				assert.Len(t, provisions, 1, "expected provision to exist")
+
+				pvc := &corev1.PersistentVolumeClaim{}
+				err := c.Get(context.TODO(), client.ObjectKey{Name: testInstallLogPVC().Name, Namespace: testNamespace}, pvc)
+				assert.NoError(t, err)
+				if pvc.Labels == nil {
+					t.Error("missing cluster deployment name label")
```
staebler: This is already covered in the else case. You do not need any special code for when the labels are nil.

pkg/controller/utils/utils.go
```diff
@@ -122,3 +122,16 @@ func LogLevel(err error) log.Level {
 		err = cause
 	}
 }
+
+// AddLabel sets the specified label on the object. If the label key is already present, then it is over written with the value. If the labels map is nil, a new labels map is created.
+func AddLabel(object metav1.Object, key, value string) {
+	objLabels := object.GetLabels()
+	if objLabels == nil {
+		object.SetLabels(map[string]string{key: value})
```
staebler: Sorry, my suggestion of using `labels` directly when passed into `SetLabels` does not work as well here now that the function is `AddLabel`.

I suggest the following.
```
if objLabels == nil {
  objLabels = map[string]string}{}
}
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
twiest: Won't that just be a nil dereference error?

Maybe I misunderstood you, but I thought your feedback was "If the labels are empty, I think we would rather see asserts that the specific labels we are interesting in are missing rather than that the labels map is empty."

Do we no longer want to see "asserts that the specific labels we are interesting in are missing"?

staebler: You will not have a nil dereference error. Getting an entry from a nil map returns the default value: It does not panic. See https://play.golang.org/p/iY5Wdz3BONO.

twiest: Oh wow, that's interesting, I had no idea. Thanks, will fix!

staebler: Sorry for the confusion. My comments could have been a lot clearer along the way.

twiest:
@staebler ok, feedback addressed, PTAL.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -440,6 +446,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			validate: func(c client.Client, t *testing.T) {
 				deprovision := getDeprovision(c)
 				assert.NotNil(t, deprovision, "expected deprovision request")
+				assert.Equal(t, testClusterDeployment().Name, deprovision.Labels[constants.ClusterDeploymentNameLabel], "incorrect cluster deployment name label")
```
staebler: This should only be run when the previous assert passes.
```
if assert.NotNil(t, deprovision, ...) {
  assert.Equal(t, testClusterDeployment().Name, ...)
}
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -586,6 +597,9 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			validate: func(c client.Client, t *testing.T) {
 				zone := getDNSZone(c)
 				assert.NotNil(t, zone, "dns zone should exist")
+
+				assert.Equal(t, testClusterDeployment().Name, zone.Labels[constants.ClusterDeploymentNameLabel], "incorrect cluster deployment name label")
```
staebler: Only when `zone` is not nil.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -472,6 +479,10 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 						break
 					}
 				}
+
+				// Ensure job type labels are set correctly
+				assert.Equal(t, testClusterDeployment().Name, job.Labels[constants.ClusterDeploymentNameLabel], "incorrect cluster deployment name label")
```
staebler: Only when `job` is not nil.

pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
```diff
@@ -259,6 +261,9 @@ func validateJobExists(t *testing.T, c client.Client) {
 	if err != nil {
 		t.Errorf("cannot get expected job")
 	}
+
+	assert.Equal(t, testClusterDeprovision().Name, job.Labels[constants.ClusterDeprovisionNameLabel], "incorrect cluster deprovision name label")
```
staebler: Only when `job` is not nil.

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -65,6 +65,12 @@ func TestClusterProvisionReconcile(t *testing.T) {
 			expectedStage:         hivev1.ClusterProvisionStageInitializing,
 			expectNoJobReference:  true,
 			expectPendingCreation: true,
+			validate: func(c client.Client, t *testing.T) {
+				job := getJob(c)
+
+				assert.Equal(t, testProvision().Name, job.Labels[constants.ClusterProvisionNameLabel], "incorrect cluster provision name label")
```
staebler: Only when `job` is not nil.

pkg/controller/clusterstate/clusterstate_controller_test.go
```diff
@@ -70,6 +71,8 @@ func TestClusterStateReconcile(t *testing.T) {
 			validate: func(t *testing.T, c client.Client, result reconcile.Result) {
 				st := cs(t, c)
 				assert.NotNil(t, st, "clusterstate should have been created")
+
+				assert.Equal(t, testClusterDeployment().Name, st.Labels[constants.ClusterDeploymentNameLabel], "incorrect cluster deployment name label")
```
staebler: Only when `st` is not nil.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
dgoodwin: May want to use testify 'require' module here, all the same assertions but the test fails immediately when one fails. (provided you don't want to keep validating other things, that may not be the case in a validate() function)

twiest: done

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
twiest: done

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
twiest: done

pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
twiest: done

pkg/controller/clusterprovision/clusterprovision_controller_test.go
twiest: done

pkg/controller/clusterstate/clusterstate_controller_test.go
twiest: done

twiest:
@staebler Ok, I addressed the feedback using the `require` package as suggested by @dgoodwin. PTAL.

twiest:
Need to investigate if we missed a label on syncsetinstance.

/hold

PR #781, opened 2020-01-22
handle changing managed domains in hiveConfig

joelddiaz:
The hive operator would not relaunch the hive-controllers deployment when the only change to the HiveConfig was adding/removing an existing managed domain entry (leaving at least one managed domain).

Update the logic to create a new configmap with a new name (so that the controllers cannot see the old configmap contents on controller restart), and detect when changes are needed so that the hive-controllers deployment is re-deployed as necessary.

A notable change is that a HiveConfig with no managed domain entries will now have a configmap with an empty list of managed domains used in the hive-controllers deployment (the dns controllers already handle the case of an empty list of managed domains). Previously the hive-operator would simply not create the configmap at all.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2301,6 +2301,7 @@
     "k8s.io/kubernetes/pkg/apis/core/validation",
     "k8s.io/kubernetes/pkg/capabilities",
     "k8s.io/kubernetes/pkg/client/metrics/prometheus",
+    "k8s.io/kubernetes/pkg/kubectl",
     "k8s.io/kubernetes/pkg/kubectl/cmd/apply",
     "k8s.io/kubernetes/pkg/kubectl/cmd/patch",
     "k8s.io/kubernetes/pkg/kubectl/cmd/util",
--- a/contrib/pkg/adm/managedns/enable.go
+++ b/contrib/pkg/adm/managedns/enable.go
@@ -11,19 +11,28 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
+	clientwatch "k8s.io/client-go/tools/watch"
+	"k8s.io/kubernetes/pkg/kubectl"
+
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
-	contributils "github.com/openshift/hive/contrib/pkg/utils"
+	hiveutils "github.com/openshift/hive/contrib/pkg/utils"
 	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
 	gcputils "github.com/openshift/hive/contrib/pkg/utils/gcp"
 	"github.com/openshift/hive/pkg/apis"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	hiveclient "github.com/openshift/hive/pkg/client/clientset-generated/clientset"
 	"github.com/openshift/hive/pkg/constants"
 	"github.com/openshift/hive/pkg/resource"
 )
@@ -38,9 +47,12 @@ managed domains, create a credentials secret for your cloud provider, and link i
 the ExternalDNS section of HiveConfig.
 `
 const (
-	cloudAWS      = "aws"
-	cloudGCP      = "gcp"
-	hiveNamespace = "hive"
+	cloudAWS                = "aws"
+	cloudGCP                = "gcp"
+	hiveNamespace           = "hive"
+	hiveAdmissionDeployment = "hiveadmission"
+	hiveConfigName          = "hive"
+	waitTime                = time.Minute * 2
 )
 
 // Options is the set of options to generate and apply a new cluster deployment
@@ -48,6 +60,9 @@ type Options struct {
 	Cloud     string
 	CredsFile string
 	homeDir   string
+
+	dynamicClient dynamic.Interface
+	hiveClient    *hiveclient.Clientset
 }
 
 // NewEnableManageDNSCommand creates a command that generates and applies artifacts to enable managed
@@ -68,14 +83,13 @@ func NewEnableManageDNSCommand() *cobra.Command {
 			if err := opt.Validate(cmd); err != nil {
 				return
 			}
-			dynClient, err := contributils.GetClient()
-			if err != nil {
-				log.WithError(err).Fatal("error creating kube clients")
+
+			if err := opt.setupLocalClients(); err != nil {
+				log.WithError(err).Fatal("error creating hive client")
 			}
 
-			err = opt.Run(dynClient, args)
-			if err != nil {
-				log.WithError(err).Error("Error")
+			if err := opt.Run(args); err != nil {
+				log.WithError(err).Fatal("Failed while deploying updated managed dns")
 			}
 		},
 	}
@@ -102,7 +116,7 @@ func (o *Options) Validate(cmd *cobra.Command) error {
 }
 
 // Run executes the command
-func (o *Options) Run(dynClient client.Client, args []string) error {
+func (o *Options) Run(args []string) error {
 	if err := apis.AddToScheme(scheme.Scheme); err != nil {
 		return err
 	}
@@ -113,8 +127,7 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
 
 	// Update the current HiveConfig, which should always exist as the operator will
 	// create a default one once run.
-	hc := &hivev1.HiveConfig{}
-	err = dynClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hc)
+	hc, err := o.hiveClient.HiveV1().HiveConfigs().Get(hiveConfigName, metav1.GetOptions{})
 	if err != nil {
 		log.WithError(err).Fatal("error looking up HiveConfig 'hive'")
 	}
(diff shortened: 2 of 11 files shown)


pkg/operator/hive/hive.go
```diff
@@ -86,9 +86,7 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		hiveContainer.Env = append(hiveContainer.Env, syncsetReapplyIntervalEnvVar)
 	}
 
-	if len(instance.Spec.ManagedDomains) > 0 {
-		addManagedDomainsVolume(&hiveDeployment.Spec.Template.Spec)
-	}
+	addManagedDomainsVolume(&hiveDeployment.Spec.Template.Spec, mdConfigMap.Name)
```
staebler: Is there a technical reason why we had to change this behavior? It seems strange to me to have an empty configmap.

pkg/operator/hive/manageddns.go
```diff
@@ -37,38 +42,105 @@ func (r *ReconcileHiveConfig) teardownLegacyExternalDNS(hLog log.FieldLogger) er
...
+	key := client.ObjectKey{Namespace: constants.HiveNamespace, Name: "managed-domains"}
+	if err := resource.DeleteAnyExistingObject(r, key, &corev1.ConfigMap{}, hLog); err != nil {
+		hLog.WithError(err).Error("failed to delete legacy managed domains configmap")
+	}
+	return nil
+}
+
+// configureManagedDomains will create a new configmap holding the managed domains settings (if necessary), or simply
+// return the current configmap of the current deployment if the settings it contains match the desired settings.
+// It will also return the old configmap if a new one was created (so that it can be garbage collected).
```
staebler: The function never returns the old configmap. It returns either the new one or the existing retained one.

pkg/operator/hive/manageddns.go
```diff
@@ -37,38 +42,105 @@ func (r *ReconcileHiveConfig) teardownLegacyExternalDNS(hLog log.FieldLogger) er
...
+
+	err := r.List(context.TODO(), configMapList, client.MatchingLabels(labelSelector))
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list config maps for managed domains")
+	}
+
+	for i, cm := range configMapList.Items {
+		if reflect.DeepEqual(cmData, cm.Data) {
+			currentConfigMap = &configMapList.Items[i]
+			continue
```
staebler: This continue is unnecessary.

pkg/operator/hive/manageddns.go
```diff
@@ -37,38 +42,105 @@ func (r *ReconcileHiveConfig) teardownLegacyExternalDNS(hLog log.FieldLogger) er
...
+	labelSelector := map[string]string{configMapLabel: "true"}
+
+	currentConfigMap := &corev1.ConfigMap{}
+
+	err := r.List(context.TODO(), configMapList, client.MatchingLabels(labelSelector))
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list config maps for managed domains")
+	}
+
+	for i, cm := range configMapList.Items {
```
staebler: This leaves multiple configmaps if there are multiple that have the correct, current data. Coming out of this function, there should only be a single configmap.

pkg/operator/hive/manageddns.go
```diff
@@ -37,38 +42,105 @@ func (r *ReconcileHiveConfig) teardownLegacyExternalDNS(hLog log.FieldLogger) er
...
+}
+
+// getCurrentConfigMap will see if any existing configmap (for managed domains) already has the necessary
+// settings. It will also delete any configmaps (for managed domains) that have out-of-date contents
+// (so that the configmaps are not orphaned as config changes happen).
+func (r *ReconcileHiveConfig) getCurrentConfigMap(cmData map[string]string, logger log.FieldLogger) (*corev1.ConfigMap, error) {
+	configMapList := &corev1.ConfigMapList{}
+	labelSelector := map[string]string{configMapLabel: "true"}
+
+	currentConfigMap := &corev1.ConfigMap{}
```
staebler: Move this declaration after the call to `List` since it is not used until then.

staebler: When there is not a good current configmap, return nil instead of an empty configmap.

pkg/operator/hive/manageddns.go
```diff
@@ -37,38 +42,105 @@ func (r *ReconcileHiveConfig) teardownLegacyExternalDNS(hLog log.FieldLogger) er
...
-	_, err = h.ApplyRuntimeObject(cm, scheme.Scheme)
-	return errors.Wrap(err, "error applying managed domains configmap")
+	newConfigMapData := map[string]string{managedDomainsConfigMapKey: string(domains)}
+
+	currentConfigMap, err := r.getCurrentConfigMap(newConfigMapData, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	if !reflect.DeepEqual(newConfigMapData, currentConfigMap.Data) {
```
staebler: We have already done this deep equal check in `getCurrentConfigMap`. There is no need to do it again. If you return nil from `getCurrentConfigMap` when there is no good configmap, then you can check for nil here to determine whether you need to create a new configmap.

pkg/operator/hive/manageddns.go
```diff
@@ -37,38 +42,105 @@ func (r *ReconcileHiveConfig) teardownLegacyExternalDNS(hLog log.FieldLogger) er
...
+			}
+		}
+	}
+
+	return currentConfigMap, nil
 }
 
-func addManagedDomainsVolume(podSpec *corev1.PodSpec) {
+func addManagedDomainsVolume(podSpec *corev1.PodSpec, configMapName string) {
+	mountPath := "/data/config"
```
staebler:
```
const mountPath = "/data/config"
```

pkg/operator/hive/hive.go
joelddiaz: i though it made the code easier to always assume there will be a configmap. i originally was preserving the behavior of not using a configmap in the absence of any managed domains, and just was dealing with a lot of special case if-no-managed-domains code. I can restore the old behavior if the empty configmap is really that surprising.

pkg/operator/hive/manageddns.go
joelddiaz: yep, that comment was for a previous version of behavior. will fix.

pkg/operator/hive/manageddns.go
joelddiaz: i didn't see a way that the operator could end up with multiple configmaps with correct, current data (unless someone comes in external to the code and starts making these configmaps) (or maybe the operator has multiple replicas running)

pkg/operator/hive/hive.go
staebler: OK. If it trickles down to a lot of places where you need to deal with the special case, then there is not too much harm in having an empty configmap.

pkg/operator/hive/manageddns.go
staebler: I agree that it is unlikely. However, you are already looping through all of the configmaps so I figured it wouldn't be too hard to also ensure that we don't have duplicates. It is fine as is.

joelddiaz:
@staebler i think i've addressed all the PR feedback (minus the bit about not using a configmap at all in the absense of any managed domains in hiveconfig)

pkg/operator/hive/manageddns.go
joelddiaz: I changed the flow to catch this case.

dgoodwin:
Needs a squash, removal of WIP tag, and passing e2e.

staebler:
The e2e test failures are legitimate. Retesting will not help.

```
Creating cluster deployment
time="2020-01-30T19:51:42Z" level=error msg="running the apply command failed" command=create-cluster error="error when creating \"/tmp/apply-974266781\": admission webhook \"clusterdeploymentvalidators.admission.hive.openshift.io\" denied the request: The base domain must be a child of one of the managed domains for ClusterDeployments with manageDNS set to true" stderr= stdout=
```

contrib/pkg/adm/managedns/enable.go
```diff
@@ -191,15 +206,59 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
-	for _, ev := range p.Spec.Containers[0].Env {
-		if ev.Name == envVar {
-			return true
+func getUpdatedManagedDNSConfig(dynClient client.Client, previousGeneration int64) (string, error) {
+	deployment := &appsv1.Deployment{}
+	key := types.NamespacedName{Name: hiveAdmissionDeployment, Namespace: hiveNamespace}
+
+	log.Info("waiting for new hiveadmission deployment")
+	for i := 0; i < 24; i++ {
+		err := dynClient.Get(context.TODO(), key, deployment)
```
staebler: Use an new empty deployment each loop through. A `Get` does not clear out fields.

staebler: It would be nice if we got in the habit of using a watch for things like this instead of looping. See https://github.com/kubernetes/client-go/blob/1a26190bd76a9017e289958b9fba936430aa3704/tools/watch/until.go#L129.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -191,15 +206,59 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+		if err != nil {
+			log.WithError(err).Fatal("error getting hiveadmission deployment")
+			return "", err
```
staebler: Is there a race condition where this may run prior to the operator creating the hiveadmission deployment for the first time? If that does happen, then the test will fail here since the `Get` will return a NotFound error.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -165,7 +174,13 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
 
 	// Adding manageDNS to HiveConfig triggers a new rollout of the hiveadmission pods.
 	// To know when it's safe to proceed, we will list all hiveadmission pods and ensure
-	// they all have the expected env var set.
+	// they all reference the new configmap.
+	log.Debug("getting new configmap with updated dns data")
+	dnsConfigMapName, err := getUpdatedManagedDNSConfig(dynClient, currHiveadmissionGeneration)
```
staebler: The hiveutil command will return an error if the managed DNS is unchanged because it will time out waiting for an update to the generation of the hiveadmission deployment.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -191,15 +206,59 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+	}
+
+	log.Error("waited for new hiveadmission deployment, but it never appeared")
+	return "", fmt.Errorf("waited for new hiveadmission deployment, but it never appeared")
+}
+
+func podUsingConfigMap(p corev1.Pod, cmName string) bool {
+	for _, vol := range p.Spec.Volumes {
+		if vol.Name == constants.ManagedDomainsVolumeName {
+			if vol.ConfigMap.Name == cmName {
```
staebler: Can this be `return vol.ConfigMap.Name == cmName`? Is there an expectation that there could be multiple volumes with the name of `ManagedDomainsVolumeName`?

joelddiaz: I would hope not. I'll fix it up to just do the equality return immediately.

contrib/pkg/adm/managedns/enable.go
joelddiaz: Yes! I'll change it to catch the IsNotFound case and keep waiting for a new deployment to appear.

joelddiaz:
@staebler i made a bunch of changes so that we can store status into the HiveConfig object. I haven't squashed, so it's all in the most recent commit on this PR. I think your proposed approach makes things easier to reason about, PTAL.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -165,18 +168,39 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+		log.WithError(err).Fatal("gave up waiting for HiveConfig to be processed")
+	}
+
+	log.Debug("getting new configmap with updated dns data")
+	dnsConfigMapName, err := getUpdatedManagedDNSConfig(dynClient)
+	if err != nil {
+		log.WithError(err).Fatal("failed to get new configmap name")
+	}
+
+	if err := waitForHiveAdmissionPods(dynClient, dnsConfigMapName); err != nil {
```
staebler: Do we need to do this advanced wait any more? Can we simply wait for the hiveadmission deployment to roll out?

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -57,6 +57,13 @@ type HiveConfigStatus struct {
 	// configmap data from the openshift-config-managed namespace. When the configmap changes,
 	// admission is redeployed.
 	AggregatorClientCAHash string `json:"aggregatorClientCAHash,omitempty"`
+
+	// LastGenerationObserved will record the most recently processed HiveConfig object's generation.
+	LastGenerationObserved int64 `json:"lastGenerationObserved,omitempty"`
+
+	// ConfigApplied will be set by the hive operator to indicate whether or not the LastGenerationObserved
+	// was successfully reconciled.
+	ConfigApplied bool `json:"configApplied,omitempty"`
```
staebler: It would be nice if this was a condition that stipulated why operator failed to apply the config. But that can be written up as a separate Jira card.

dgoodwin:
I have pointed this at master instead of v1 branch.

joelddiaz:
@staebler some major changes with these most recent commits.

Primary change is using some of functionality from what powers 'kubectl rollout status --wait' (and all the changes that required).
Also addressed the renaming of the status field to 'observedGeneration'.
Now rebased on the master branch as well.

This will need some squashing, but I'm keeping them separate for now to hopefully make clear the most recent changes.

Gopkg.lock
```diff
@@ -2301,6 +2301,7 @@
     "k8s.io/kubernetes/pkg/apis/core/validation",
     "k8s.io/kubernetes/pkg/capabilities",
     "k8s.io/kubernetes/pkg/client/metrics/prometheus",
+    "k8s.io/kubernetes/pkg/kubectl",
```
staebler: How is there a change to `Gopkg.lock` without any change to vendor files?

contrib/pkg/adm/managedns/enable.go
```diff
@@ -160,112 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+			return dynClient.Resource(gvr).Namespace(constants.HiveNamespace).List(options)
+
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			options.FieldSelector = fieldSelector
+			return dynClient.Resource(gvr).Namespace(constants.HiveNamespace).Watch(options)
+		},
+	}
+
+	preconditionFunc := func(store cache.Store) (bool, error) {
```
staebler: I don't think we need this precondition for our use case. We **want** to wait for the timeout if the deployment does not exist already.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -160,112 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			options.FieldSelector = fieldSelector
+			return hiveClient.HiveV1().HiveConfigs().Watch(options)
+		},
 	}
 
-	for i, vol := range deployment.Spec.Template.Spec.Volumes {
-		if vol.Name == constants.ManagedDomainsVolumeName {
-			return deployment.Spec.Template.Spec.Volumes[i].ConfigMap.Name, nil
+	preconditionFunc := func(store cache.Store) (bool, error) {
```
staebler: We do no need this precondition.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -160,112 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+	if err := scheme.Scheme.Convert(unstDeployment, deployment, nil); err != nil {
+		log.WithError(err).Error("failed to convert unstructed deployment to deployment")
+		return err
+	}
+
+	statusViewer := &kubectl.DeploymentStatusViewer{}
+
+	fieldSelector := fields.OneTermEqualSelector("metadata.name", hiveAdmissionDeployment).String()
+
+	lw := &cache.ListWatch{
```
staebler: Use `NewFilteredListWatchFromClient` from k8s.io/client-go/tools/cache.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -160,112 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
+func waitForHiveConfigToBeProcessed(hiveClient *hiveclient.Clientset) error {
+	// Wait for the HiveConfig.Status.ObservedGeneration to match
+	// the HiveConfig's Generation (and status showing ConfigApplied == true).
 
-	if err := dynClient.Get(context.TODO(), key, deployment); err != nil {
-		log.WithError(err).Fatal("error getting hiveadmission deployment")
-		return "", err
+	fieldSelector := fields.OneTermEqualSelector("metadata.name", hiveConfigName).String()
+
+	lw := &cache.ListWatch{
```
staebler: Use NewFilteredListWatchFromClient from k8s.io/client-go/tools/cache.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -160,112 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
-		case err != nil:
-			log.WithError(err).Error("error getting HiveConfig object")
-			return err
-		case hc.Generation == hc.Status.LastGenerationObserved && hc.Status.ConfigApplied:
-			return nil
-		}
+	_, err = clientwatch.UntilWithSync(ctx, lw, &unstructured.Unstructured{}, preconditionFunc, func(e watch.Event) (bool, error) {
+		switch t := e.Type; t {
+		case watch.Added, watch.Modified:
+			_, done, err := statusViewer.Status(e.Object.(runtime.Unstructured), deployment.Generation)
```
staebler: I would pass in a revision of 0 instead of the revision from the deployment. We do not care about reaching that particular revision: We only care about reaching a point where the observedGeneration and generation are equal.

This also means that we do not need to get the deployment prior to the call to `UntilWithSync`.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -160,112 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
-	return fmt.Errorf("waited for HiveConfig to be procssed, but it never completed")
+	log.Debug("done waiting for hiveadmission pods")
+	return err
 }
 
-func getUpdatedManagedDNSConfig(dynClient client.Client) (string, error) {
-	key := types.NamespacedName{Name: hiveAdmissionDeployment, Namespace: hiveNamespace}
-	deployment := &appsv1.Deployment{}
+func waitForHiveConfigToBeProcessed(hiveClient *hiveclient.Clientset) error {
+	// Wait for the HiveConfig.Status.ObservedGeneration to match
```
staebler: Put this comment on the function itself rather than inside the function.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -320,3 +378,30 @@ func (o *Options) getResourceHelper() (*resource.Helper, error) {
 	helper := resource.NewHelperFromRESTConfig(cfg, log.WithField("command", "adm manage-dns enable"))
 	return helper, nil
 }
+
+func (o *Options) setupLocalClients() error {
+	log.Debug("creating cluster client config")
+	rules := clientcmd.NewDefaultClientConfigLoadingRules()
+	kubeconfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, &clientcmd.ConfigOverrides{})
+	cfg, err := kubeconfig.ClientConfig()
```
staebler: `GetClientConfig` from contrib/pkg/utils

Gopkg.lock
staebler: I do see that those files are already in vendor.

joelddiaz: i had originally vendored some additional stuff from k8s.io/kubernetes/pkg/kubectl, but ended up not using it, so when i re-vendored, this is what was left over. i'm quite confused, but each current run of 'dep ensure' keeps that change in there.

contrib/pkg/adm/managedns/enable.go
joelddiaz: maybe i'm missing something here. why would a deployment not exist? this is the hiveadmission pods we're talking about which should always be running on a healthy instance of hive, no?

contrib/pkg/adm/managedns/enable.go
```diff
@@ -157,47 +171,159 @@ func (o *Options) Run(dynClient client.Client, args []string) error {
...
-	log.Info("Hive is now ready to create clusters with manageDNS=true")
 
-	return nil
+	ctx, cancel := context.WithTimeout(context.Background(), waitTime)
+	defer cancel()
+
+	_, err = clientwatch.UntilWithSync(ctx, lw, &unstructured.Unstructured{}, preconditionFunc, func(e watch.Event) (bool, error) {
+		switch t := e.Type; t {
+		case watch.Added, watch.Modified:
+			_, done, err := statusViewer.Status(e.Object.(runtime.Unstructured), deployment.Generation)
```
staebler: This conversion from a `runtime.Object` to a `runtime.Unstructured` should have an OK check.

contrib/pkg/adm/managedns/enable.go
staebler: This hiveutil command could be run prior to the hive-operator creating the hiveadmission deployments for the first time.

contrib/pkg/adm/managedns/enable.go
```diff
@@ -381,12 +343,9 @@ func (o *Options) getResourceHelper() (*resource.Helper, error) {
 
 func (o *Options) setupLocalClients() error {
 	log.Debug("creating cluster client config")
-	rules := clientcmd.NewDefaultClientConfigLoadingRules()
-	kubeconfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, &clientcmd.ConfigOverrides{})
-	cfg, err := kubeconfig.ClientConfig()
+	cfg, err := hiveutils.GetClientConfig()
 	if err != nil {
 		log.WithError(err).Error("cannot obtain client config")
-		return err
```
staebler: Was the remove to the return intentional?

joelddiaz: absolutely not. good catch.

PR #782, opened 2020-01-22
GCP Multi-MachinePool support with single character leases.

dgoodwin:
Implement a leasing system to support multiple machinepools. "w" is reserved for the "worker" pool from the installer. "m" is reserved for masters to avoid colliding with master machines, even though they are not represented by a MachineSet yet.

GCP will be limited to 35 machine pools. (a-z0-9 minus "m")


dgoodwin:
Need to review code myself and do real world testing still so a bit rough.

dgoodwin:
Wondering if we should add labels or spec/status fields to help link MachineSets to their MachinePool.

staebler:
> Wondering if we should add labels or spec/status fields to help link MachineSets to their MachinePool.

I am not exactly following what you are proposing here. Does any of the following cover it.
1. We currently set the `hive.openshift.io/machine-pool` label on the MachineSets that Hive manages. The value of the label is the pool name.
2. The MachinePool status has the names of all of the MachineSets that comprise the MachinePool.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -1,13 +1,21 @@
 package remotemachineset
 
 import (
+	"context"
 	"fmt"
+	"math/rand"
+	"strings"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
staebler: Put these k8s imports with the other k8s imports below.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -17,25 +25,32 @@ import (
...
+	client    client.Client
+	gcpClient gcpclient.Client
+	logger    log.FieldLogger
 }
 
 var _ Actuator = &GCPActuator{}
 
 // NewGCPActuator is the constructor for building a GCPActuator
-func NewGCPActuator(gcpCreds *corev1.Secret, logger log.FieldLogger) (*GCPActuator, error) {
+func NewGCPActuator(client client.Client, gcpCreds *corev1.Secret, logger log.FieldLogger) (*GCPActuator, error) {
```
staebler: Did you intend to store `client` in the created actuator?

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -84,7 +105,12 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 		computePool.Platform.GCP.Zones = zones
 	}
 
-	installerMachineSets, err := installgcp.MachineSets(cd.Spec.ClusterMetadata.InfraID, ic, computePool, imageID, pool.Spec.Name, "worker-user-data")
+	// Installer code uses the first character of the computePool.Name in the MachineSet name,
+	// so we need to fake it to get it to use our leaseChar.
+	computePool.Name = string(leaseChar)
+	// Assuming all machine pools are workers at this time.
+	role := "worker"
```
staebler:
```
const role = worker
```

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+
+// obtainLease uses the Hive MachinePoolNameLease resource to obtain a unique, single character
+// for use in the name of the machine pool. We are severely restricted on name lengths on GCP
+// and effectively have one character of flexibility with the naming convention originating in
+// the installer.
+func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, logger log.FieldLogger) (rune, error) {
+	logger.Debugf("obtaining lease for pool: %s", pool.Name)
+
+	cd := &hivev1.ClusterDeployment{}
+	err := a.client.Get(context.TODO(), types.NamespacedName{Namespace: pool.Namespace, Name: pool.Spec.ClusterDeploymentRef.Name}, cd)
```
staebler: Why do we need to fetch the clusterdeployment? It was passed in to the `GenerateMachineSets` function which calls this function.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	if err != nil {
+		return 'X', err
+	}
+
+	leases := &hivev1.MachinePoolNameLeaseList{}
+	// TODO: pass ^^ to find command?
```
staebler: What does this comment mean?

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	err = a.client.List(context.TODO(), leases, client.UseListOptions(&client.ListOptions{Namespace: pool.Namespace}))
```
staebler: Replace
```
client.UseListOptions(...)
```
with
```
client.InNamespace(pool.Namespace)
```

staebler: Add a label to the lease so that you can limit the list to only leases for the clusterdeployment via the `client.MatchingLabels` function.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
 		return "", errors.New(msg)
 	}
 }
+
+// obtainLease uses the Hive MachinePoolNameLease resource to obtain a unique, single character
+// for use in the name of the machine pool. We are severely restricted on name lengths on GCP
+// and effectively have one character of flexibility with the naming convention originating in
+// the installer.
+func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, logger log.FieldLogger) (rune, error) {
```
staebler: Return a `string` rather than a `rune`. Returning a `rune` (1) unnecessarily limits what this function can do in the future and (2) provides no benefit over a `string` as it will be converted to a `string` by the caller.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+
+	leases := &hivev1.MachinePoolNameLeaseList{}
+	// TODO: pass ^^ to find command?
+	err = a.client.List(context.TODO(), leases, client.UseListOptions(&client.ListOptions{Namespace: pool.Namespace}))
+	if err != nil {
+		return 'X', err
+	}
+	logger.Debugf("found %d leases in namespace", len(leases.Items))
+	for _, l := range leases.Items {
+		for _, ownerRef := range l.OwnerReferences {
```
staebler: Do not use owner references as authoritative: Use a label instead.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+				ownerRef.Kind == "MachinePool" &&
+				ownerRef.Name == pool.Name {
+				logger.Info("machine pool already has lease: %s", l.Name)
+				return rune(l.Name[len(l.Name)-1]), nil
+			}
+		}
+	}
+
+	// Pool does not have a lease yet, lookup all currently available lease chars
+	availLeaseChars, err := a.findAvailableLeaseChars(cd)
```
staebler: Why not pass in the `MachinePoolNameLeaseList` to `findAvailableLeaseChars` rather than listing again?

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	if err != nil {
+		return 'X', err
+	}
+	logger.Debugf("availLeaseChars: %v", availLeaseChars)
+
+	// If the pool.Spec.Name == "worker", we want to preseve this MachinePool's "w" character that
+	// the installer would have selected so we do not cycle all worker nodes. In theory, multiple
+	// MachinePools could specify this in their spec.Name, in which case we'll start erroring
```
staebler: This is not true. The pool name is unique for a given clusterdeployment because the name of the MachinePool must be $CLUSTERDEPLOYMENTNAME-$POOLNAME.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	// because the first one already grabbed "w".
+	var leaseRune rune
+	if pool.Spec.Name == "worker" {
```
staebler: I would rather special-case this earlier rather than leasing "w". In other words, if the pool name is "worker", then automatically use "w" for the lease char, do not look through existing leases, and do not create a lease. Remove "w" from the list of valid lease chars.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+				leaseRune = 'w'
+				logger.Debug("selecting lease char 'w' for original worker pool")
+			}
+		}
+	} else {
+		// Choose a random entry in the available chars to limit collisions while processing
+		// multiple machine pools at the same time. In this case the subsequent attempts to create
+		// the lease will fail due to name collisions and re-reconcile.
+		logger.Debug("selecting random lease char from available")
+		rand.Seed(time.Now().UnixNano())
```
staebler: I think that you want this already seeded rather than seeding anew on every use.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+		leaseRune = availLeaseChars[rand.Intn(len(availLeaseChars))]
+	}
+	logger.Debugf("selected lease char: %s", string(leaseRune))
+
+	leaseName := fmt.Sprintf("%s-%s", cd.Spec.ClusterMetadata.InfraID, string(leaseRune))
+	// Attempt to claim the lease:
+	l := &hivev1.MachinePoolNameLease{
```
staebler: Don't we need to add the owner reference? And, preferably, the labels.

In addition to the owner reference to the MachinePool alluded to elsewhere in the code, there should also be an owner reference to the ClusterDeployment. We want the lease to expire when either the MachinePool or the ClusterDeployment are deleted.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+		chars[char] = true
+	}
+	return chars
+}
+
+func (a *GCPActuator) findAvailableLeaseChars(cd *hivev1.ClusterDeployment) ([]rune, error) {
+	availChars := stringToRuneSet(validLeaseChars)
+
+	leases := &hivev1.MachinePoolNameLeaseList{}
+	err := a.client.List(context.TODO(), leases, client.UseListOptions(&client.ListOptions{Namespace: cd.Namespace}))
```
staebler: `client.InNamespace`

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	if err != nil {
+		return []rune{}, err
```
staebler: There is no need to allocate an array here or in any of the other returns where the function is returning an error.
```
return nil, err
```

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	for _, lease := range leases.Items {
+		if len(lease.OwnerReferences) != 1 {
+			return []rune{}, fmt.Errorf("MachinePoolNameLease %s has unexpected number of owner refs: %d", lease.Name, len(lease.OwnerReferences))
+		}
+
+		if lease.OwnerReferences[0].APIVersion != "hive.openshift.io/v1" || lease.OwnerReferences[0].Kind != "MachinePool" {
+			return []rune{}, fmt.Errorf("MachinePoolNameLease %s has unexpected owner ref: %v", lease.Name, lease.OwnerReferences[0])
+		}
+
+		// TODO: should we store the cluster deployment infra ID on the lease to make this unnecessary?
```
staebler: Yes. It should be stored in a label, although I don't follow why the infra ID is necessary and not just the clusterdeployment name.

But, we still need to check that the name of the lease is in the correct format. We are relying on that name to provide the uniqueness of the lease, so it **must** be correct.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +168,126 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+			// doesn't match this cluster
+			continue
+		}
+
+		// Lease name is [infraid]-[x], we need to parse out the 'x'.
+		char := lease.Name[len(lease.Name)-1]
+		delete(availChars, rune(char))
+	}
+
+	keys := make([]rune, 0, len(availChars))
```
staebler:
```
keys := make([]rune, len(availChars))
i := 0
for k := range availChars {
  keys[i] = k
  i++
}
```

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: I did yes. Is there a reason not to? Viewing it as an extension of the actuator and it needs several client calls now, seemed more logical than passing it around several times.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: passing the lease list to the function that also does the same query later, which you caught with a comment there as well.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: Ah yes! Thanks.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: The idea here was to error if two pools had spec.name = worker, but if that's not possible then I'm with you.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: Whoops, moved to NewActuator.

pkg/controller/remotemachineset/gcpactuator.go
staebler: No. What I meant was that you are not setting the client when I think that you meant to.

dgoodwin: LOL that's a good one, thanks. Would have hit that fast in real world  testing.

pkg/controller/remotemachineset/gcpactuator.go
staebler: I don't know that we need extra protection for this specific to the GCP case. There is already protection in the admission web hook validators that rejects duplicates, and that is generic for all platforms.

pkg/controller/remotemachineset/gcpactuator.go
staebler: I would rather do it in `cmd/manager/main.go` or in an `init()` function.

pkg/controller/remotemachineset/gcpactuator.go
staebler: Ah, I misread your response. We are on the same page. Disregard my response.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: Moved to main.go.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: Adding a check that everything but the last char matches infraid- which should cover us in all cases. (and a test)

dgoodwin:
Updated with everything but the expectations, I'll work off your past PRs for that first chance I get.

twiest:
@dgoodwin hey, above you said this PR was "rough". It also needs a rebase.

Are you looking for a PR review or do you want some more time to get it ready first?

dgoodwin:
Working on implementing the expectations but I'm hitting some things that I don't understand.

If I'm following the problem is reconciling a new pool, we list all leases in this namespace, find none owned by our pool, and create a new one. We then return from the obtainLease function but proceed to create the remote machine pools, assuming that our lease is now in place.

When combined with your comment that we *still* would need to handle two leases for one pool even with expectations I don't think we've implemented leases properly, and cleaning this up could be a pretty serious problem. 

First question would be how is this still possible with expectations?

Second question would be if there is some way to better approach this. If we stored the lease character onto status as soon as we obtain it, would this bypass the cache creation problem? Because frankly I'd massively prefer that to expectations plus handling whatever scenario could lead to two leases for one pool.

dgoodwin:
@twiest needs more time, I've done the rebase locally but working on the expectations problem first. Rest of the code is ok for review though.

dgoodwin:
Clarification from slack, expectations *should* prevent there ever being two leases for one pool. @staebler 's point was more in regards to defensive programming. I am working on the expectations and will give some thought to whether or not we can reasonably handle two leases for one pool. (or should just start erroring and let an admin sort it out)

dgoodwin:
I think this is ready for review, updated with expectations and tested to be working. I do need to go make sure we're hitting this as hard as other platforms in e2e.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
 		return "", errors.New(msg)
 	}
 }
+
+// obtainLease uses the Hive MachinePoolNameLease resource to obtain a unique, single character
+// for use in the name of the machine pool. We are severely restricted on name lengths on GCP
+// and effectively have one character of flexibility with the naming convention originating in
+// the installer.
+func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (string, bool, error) {
```
staebler: Either (1) add details in the function description about what the second return value is or (2) name the return values.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+				ownerRef.Kind == "MachinePool" &&
+				ownerRef.Name == pool.Name {
+
+				logger.Debugf("machine pool already has lease: %s", l.Name)
+
+				// Ensure the lease name is in the format we expect, we know everything up to
+				// the last character.
+				expectedLeasePrefix := fmt.Sprintf("%s-", cd.Spec.ClusterMetadata.InfraID)
+				if l.Name[0:len(l.Name)-1] != expectedLeasePrefix {
+					return "", false, fmt.Errorf("lease %s did not match expected lease name format (%s[CHAR])", l.Name, expectedLeasePrefix)
```
staebler: Should Hive do something to resolve this?

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -1,62 +1,111 @@
...
 	installertypesgcp "github.com/openshift/installer/pkg/types/gcp"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
 	"github.com/openshift/hive/pkg/gcpclient"
 )
 
+const (
+	// Omit m, the installer used this for the master machines.
+	validLeaseChars = "abcdefghijklnopqrstuvwxyz0123456789"
```
staebler: The `w` character needs to be removed from the valid lease chars.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	if pool.Spec.Name == "worker" {
+		leaseRune = 'w'
+		logger.Debug("selecting lease char 'w' for original worker pool")
+	} else {
+		// Pool does not have a lease yet, lookup all currently available lease chars
+		availLeaseChars, err := a.findAvailableLeaseChars(cd, leases)
+		if err != nil {
+			return "", false, err
+		}
+		logger.Debugf("availLeaseChars: %v", availLeaseChars)
```
staebler: I do not think that this will print anything meaningful. I think it will only print the address of the slice--not the contents.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+					Controller: pointer.BoolPtr(true),
+				},
+			},
+		},
+	}
+	logger.Debug("adding expectation for lease creation for this pool")
+	a.expectations.ExpectCreations(types.NamespacedName{Namespace: pool.Namespace, Name: pool.Name}.String(), 1)
+	err = a.client.Create(context.TODO(), l)
+	if err != nil {
+		return "", false, err
```
staebler: If there is an error, you need to remove the expected creation.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+					Kind:       "MachinePool",
+					Name:       pool.Name,
+					UID:        pool.UID,
+					Controller: pointer.BoolPtr(true),
+				},
+			},
+		},
+	}
+	logger.Debug("adding expectation for lease creation for this pool")
+	a.expectations.ExpectCreations(types.NamespacedName{Namespace: pool.Namespace, Name: pool.Name}.String(), 1)
```
staebler: It smells to me a bit that the responsibility for expectations is split between the controller and the actuator. I can live with it for now, but I reserve the right to revisit it at some point.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -1,62 +1,111 @@
...
 	}
 
+	leaseChar, leaseCreated, err := a.obtainLease(pool, cd, logger)
+	if err != nil {
+		logger.WithError(err).Error("error obtaining pool name lease")
+		return nil, false, err
+	}
+
+	if leaseCreated {
+		return nil, false, fmt.Errorf("new machine pool name lease created, waiting until creation is observed")
```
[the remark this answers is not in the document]
dgoodwin: You're quite right, I did this both ways, put the proceed flag back in, and then forgot to take out the error.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+		i++
+	}
+
+	return keys, nil
+}
+
+func lookupProjectID(kubeClient client.Client, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (string, error) {
+
+	credsSecret := &corev1.Secret{}
+	if err := kubeClient.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Spec.Platform.GCP.CredentialsSecretRef.Name}, credsSecret); err != nil {
```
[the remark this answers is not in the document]
dgoodwin: Awesome, missed that, thanks!

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +195,145 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+		client.MatchingLabels(map[string]string{constants.ClusterDeploymentNameLabel: cd.Name}))
+	if err != nil {
+		return "", false, err
+	}
+	logger.Debugf("found %d leases in namespace", len(leases.Items))
+	for _, l := range leases.Items {
+		logger.Debugf("checking lease: %s", l.Name)
+		for _, ownerRef := range l.OwnerReferences {
+			logger.Debugf("checking owner ref: %v", ownerRef)
+			if ownerRef.APIVersion == "hive.openshift.io/v1" &&
```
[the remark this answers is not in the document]
dgoodwin: Good idea, I will add the label to the selector and then skip the whole for loop.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: Vars named.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: It printed the contents of the slice but not as characters. Removing as that's not useful either. :)

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: Yeah the thought process here was making all of this, leases and expectations, available to another actuator if they wanted to make use of it. I'm open to changing it in future though as I'm not sure this is the best either.

pkg/controller/remotemachineset/gcpactuator.go
dgoodwin: I'm not sure what it would be. I don't think I'd dare delete it and let it create a new one as we might nuke a machine set. It felt like a pretty fringe/unexpected thing to happen, and a fair candidate to refuse to reconcile and require manual intervention from an admin who can figure out what happened.

dgoodwin:
Review feedback addressed and e2e testing enabled for GCP machinesets, which got a little bigger than I'd have liked. I was having a hard time seeing what was going on when waiting for timeouts so I reworked the tests to use logrus, added -v in the Makefile targets, and switched to using testify instead of just t.Fail.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -200,37 +201,35 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	if len(leases.Items) > 1 {
+		return "", false, fmt.Errorf("found %d leases for pool, manual intervention required", len(leases.Items))
+	} else if len(leases.Items) == 1 {
+		l := leases.Items[0]
+		logger.Debugf("machine pool already has lease: %s", l.Name)
+
+		// Ensure the lease name is in the format we expect, we know everything up to
+		// the last character.
+		expectedLeasePrefix := fmt.Sprintf("%s-", cd.Spec.ClusterMetadata.InfraID)
+		if l.Name[0:len(l.Name)-1] != expectedLeasePrefix {
```
staebler: The `0` can be omitted.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -200,37 +201,35 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+	logger.Debugf("found %d leases for pool", len(leases.Items))
+	if len(leases.Items) > 1 {
+		return "", false, fmt.Errorf("found %d leases for pool, manual intervention required", len(leases.Items))
+	} else if len(leases.Items) == 1 {
+		l := leases.Items[0]
+		logger.Debugf("machine pool already has lease: %s", l.Name)
+
+		// Ensure the lease name is in the format we expect, we know everything up to
+		// the last character.
+		expectedLeasePrefix := fmt.Sprintf("%s-", cd.Spec.ClusterMetadata.InfraID)
```
staebler: I think this can be made a little easier to read in the following way. Take it or leave it. Your choice.

```
leaseChar := l.Name[len(l.Name)-1:]
expectedLeaseName := fmt.Sprintf("%s-%s", cd.Spec.ClusterMetadata.InfraID, leaseChar)
if expectedLeastName != l.Name {
  return "", false, fmt.Errorf(...)
}
return leaseChar, false, nil
```

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -200,37 +201,35 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+		if l.Name[0:len(l.Name)-1] != expectedLeasePrefix {
+			return "", false, fmt.Errorf("lease %s did not match expected lease name format (%s[CHAR])", l.Name, expectedLeasePrefix)
 		}
+
+		return string(l.Name[len(l.Name)-1]), false, nil
```
staebler:
```
return l.Name[len(l.Name)-1:], false, nil
```

Makefile
```diff
@@ -54,11 +54,11 @@ test-e2e:
 
 .PHONY: test-e2e-postdeploy
 test-e2e-postdeploy:
-	go test -timeout 0 -count=1 ./test/e2e/postdeploy/...
+	go test -v -timeout 0 -count=1 ./test/e2e/postdeploy/...
```
staebler: This will fill up the ci e2e logs with test output for successful tests, which is generally not what we want.

test/e2e/common/machinepoolnamelease.go
```diff
@@ -0,0 +1,71 @@
...
+
+	log "github.com/sirupsen/logrus"
+
+	"k8s.io/client-go/rest"
+	clientcache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+)
+
+// WaitForMachinePoolNameLease waits for a lease to appear in the given namespace for the given pool.
+func WaitForMachinePoolNameLease(cfg *rest.Config, namespace, poolName string, timeOut time.Duration) (string, error) {
```
staebler: This function strikes me as a lot more complicated than it needs to be. Can we accomplish the same thing more easily with `clientwatch.UntilWithSync`?

test/e2e/postinstall/machinesets/infra_test.go
```diff
@@ -22,12 +25,19 @@ import (
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	hivev1aws "github.com/openshift/hive/pkg/apis/hive/v1/aws"
 	hivev1azure "github.com/openshift/hive/pkg/apis/hive/v1/azure"
+	hivev1gcp "github.com/openshift/hive/pkg/apis/hive/v1/gcp"
 	"github.com/openshift/hive/test/e2e/common"
 )
 
+const (
+	workerPoolName       = "worker"
```
staebler: Either
`workerPoolName` and `infraPoolName`
or
`workerMachinePoolName` and `infraMachinePoolName`.

test/e2e/postinstall/machinesets/infra_test.go
```diff
@@ -41,59 +51,62 @@ func TestScaleMachinePool(t *testing.T) {
 	c := common.MustGetClient()
 
 	// Scale down
-	pool := common.GetMachinePool(cd, "worker")
-	if pool == nil {
-		t.Fatal("worker machine pool does not exist")
-	}
+	pool := common.GetMachinePool(cd, workerPoolName)
+	require.NotNil(t, pool, "worker machine pool does not exist: ", workerPoolName)
```
staebler: My preference would be to use `NotNilf` here instead.

Makefile
dgoodwin: I don't think it's going to actually fill anything, just add some context but I'm fine with removing and going back to only on failures.

test/e2e/common/machinepoolnamelease.go
dgoodwin: It's done, maybe a little cleaner.

dgoodwin:
Updates pushed.

I am fine with removing the -v from e2e makefile targets but I am very much in favor of the switch to logrus. t.Log is a common problem for go programmers and surfaces especially in long running tests, you see nothing until the test ends. That is the situation I was in fixing gcp and it's a pain. It's a popular request to change in go itself, but they can't because they have to first print pass or fail, and things depend on that output, so they cannot change it now. 

logrus gets us output on the fly, and respects -v, you will still see nothing for passing tests and everything once they fail without it. Meanwhile developers can use -v and have some understanding of what's going on in a test that can take 5-10 minutes. It's objectively better and I can't think of a downside. We're also already using loggers in the e2e common code, and mixing that with t.Log is weird.

IMO we should not be using t.Log anywhere, and opt for loggers instead.

dgoodwin:
Leaving WIP until there's a go-ahead to squash and tests are clear.

staebler:
> logrus gets us output on the fly, and respects -v, you will still see nothing for passing tests and everything once they fail without it. Meanwhile developers can use -v and have some understanding of what's going on in a test that can take 5-10 minutes. It's objectively better and I can't think of a downside. We're also already using loggers in the e2e common code, and mixing that with t.Log is weird.

Ah, yes, you are correct. Logrus does respect the `-v` flag. I was confused with my IDE experience where the `-v` flag is added automatically. I retract my objection to using logrus.

test/e2e/common/machinepoolnamelease.go
```diff
@@ -0,0 +1,78 @@
...
+// WaitForMachinePoolNameLease waits for a lease to appear in the given namespace for the given pool.
+func WaitForMachinePoolNameLease(config *rest.Config, namespace, poolName string, timeOut time.Duration) (string, error) {
+	logger := log.WithField("pool", poolName)
+	logger.Infof("Waiting for MachinePoolNameLease")
+	scheme, err := hivev1.SchemeBuilder.Build()
+
+	waitContext, cancel := context.WithTimeout(context.Background(), timeOut)
+	defer cancel()
+
+	gvk, err := apiutil.GVKForObject(&hivev1.MachinePoolNameLease{}, scheme)
```
staebler: You do not need all the scheme and codec cruft.

```
hiveClient, err := hiveclient.NewForConfig(config)
if err != nil {
  return "", errors.Wrap(err, "could not create the Hive client")
}

restClient := hiveClient.HiveV1().RESTClient()
```

test/e2e/common/machinepoolnamelease.go
```diff
@@ -0,0 +1,78 @@
...
+			restClient,
+			"machinepoolnameleases",
+			namespace,
+			optionsModifier,
+		),
+		&hivev1.MachinePoolNameLease{},
+		nil,
+		func(event watch.Event) (bool, error) {
+			switch event.Type {
+			case watch.Added, watch.Modified, watch.Deleted:
```
staebler: I'm not sure that you need to consider `watch.Deleted`.

test/e2e/common/machinepoolnamelease.go
```diff
@@ -0,0 +1,78 @@
...
+				lease, ok = event.Object.(*hivev1.MachinePoolNameLease)
```
staebler: I don't think that you need to do this part here. I would move it to after the `UnitlWithSync` call. If there is an object received in the event that is not a MachinePoolNameLease, then that is reason enough to fail the test.

```
event, err := clientwatch.UntilWithSync(
  ...
  func(event watch.Event) (bool, error) {
    return event.Type == watch.Added || event.Type == watch.Modified
  },
)
if err != nil {
  return "", errors.Wrap(err, "timeout waiting for MachinePoolNameLease")
}
lease, ok := event.Object.(*hivev1.MachinePoolNameLease)
if !ok {
  return "", errors.Fmt("expected a MachinePoolNameLease object but got a %T object instead", event.Object)
}
logger.Infof("found MachinePoolName lease: %s", lease.Name)
return lease.Name, nil
```

test/e2e/common/machinepoolnamelease.go
dgoodwin: Excellent, thanks.

test/e2e/common/machinepoolnamelease.go
dgoodwin: Looks better.

dgoodwin:
Feedback addressed, squashed into one commit. Removing WIP.

test/e2e/common/machinepoolnamelease.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "could not create the Hive client")
+	}
+
+	restClient := hiveClient.HiveV1().RESTClient()
+	optionsModifier := func(options *metav1.ListOptions) {
+		options.LabelSelector = fmt.Sprintf("%s=%s", constants.MachinePoolNameLabel, poolName)
+	}
+
+	var lease *hivev1.MachinePoolNameLease
```
staebler: Can you remove these two declarations?

dgoodwin: done

dgoodwin:
One of the errors:

```
time="2020-03-05T01:39:43Z" level=error msg="running the apply command failed" controller=hive error="error when creating \"/var/cache/kubectl/apply-342300959\": clusterroles.rbac.authorization.k8s.io \"hive-admin\" is forbidden: user \"system:serviceaccount:hive:hive-operator\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:hive\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"hive.openshift.io\"], Resources:[\"machinepoolnameleases\"], Verbs:[\"get\" \"list\" \"watch\"]}" stderr= stdout=
time="2020-03-05T01:39:43Z" level=error msg="error applying asset" asset=config/rbac/hive_admin_role.yaml controller=hive error="error when creating \"/var/cache/kubectl/apply-342300959\": clusterroles.rbac.authorization.k8s.io \"hive-admin\" is forbidden: user \"system:serviceaccount:hive:hive-operator\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:hive\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"hive.openshift.io\"], Resources:[\"machinepoolnameleases\"], Verbs:[\"get\" \"list\" \"watch\"]}"
time="2020-03-05T01:39:43Z" level=error msg="error deploying Hive" controller=hive error="error when creating \"/var/cache/kubectl/apply-342300959\": clusterroles.rbac.authorization.k8s.io \"hive-admin\" is forbidden: user \"system:serviceaccount:hive:hive-operator\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:hive\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"hive.openshift.io\"], Resources:[\"machinepoolnameleases\"], Verbs:[\"get\" \"list\" \"watch\"]}"
```

I've updated PR to add the permission to the operator role.

A number of test runs didn't make it past bootstrap, but sometimes they did and then hit this, last round all made it past so I'm hoping that was a flake in the CI system somewhere.

staebler:
> I've updated PR to add the permission to the operator role.

Do you think that there is any reason for hive_reader or hive_frontend to have at least GET permissions for machinepoolleasename?

dgoodwin:
I can't think of a reason they need it at this time, but we can add it in future if so.

Unit test failure is legit, there's a flake, if I run it enough times I can see that fail locally as well. Working on this now.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -142,3 +196,129 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
...
+// and effectively have one character of flexibility with the naming convention originating in
+// the installer.
+func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (leaseChar string, leaseCreated bool, leaseErr error) {
+	logger.Debugf("obtaining lease for pool: %s", pool.Name)
+
+	leases := &hivev1.MachinePoolNameLeaseList{}
+	err := a.client.List(context.TODO(), leases, client.InNamespace(pool.Namespace),
+		client.MatchingLabels(map[string]string{
+			constants.ClusterDeploymentNameLabel: cd.Name,
+			constants.MachinePoolNameLabel:       pool.Name,
```
dgoodwin: Just found a bug with this refactor here, in limiting only to this pool, we later use this list to calc all the leases in use and thus always find nothing, resulting in possibility of hitting a char already in use error.

dgoodwin:
Operator still missing the permission, not sure what's up yet.

dgoodwin:
/test e2e-azure 
/test e2e-gcp

AWS passed

dgoodwin:
e2e and e2e-gcp are passing, azure failed for gateway timeouts, and previously for lack of IPs which I reported to dev productivity.

Needs an /lgtm, quick as possible. :)

PR #783, opened 2020-01-22
Remove use of ObjectReference type in Hive v1 API

csrwng:
Changes the reference type used in SyncSets and Checkpoints. 
Renames the existing SecretReference type to SecretMapping and introduces a new SecretReference type that will refer to secrets.


pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -214,32 +214,32 @@ func rawExtensionsFromClusterDeployment(rContext *reconcileContext) []runtime.Ra
...
-		secretReference := hivev1.SecretReference{
-			Source: corev1.ObjectReference{
+		secretMapping := hivev1.SecretMapping{
+			Source: hivev1.SecretReference{
 				Namespace: cbSecret.Namespace,
 				Name:      cbSecret.Name,
 			},
-			Target: corev1.ObjectReference{
+			Target: hivev1.SecretReference{
 				Namespace: remoteIngressConrollerSecretsNamespace,
```
staebler: Note, for myself so that I can find it later, that there is a typo here where Controller is incorrectly spelled Conroller.

csrwng:
Thx @staebler , comments addressed. I do like the `Secrets` field name used in both the spec and status, so I changed it to that.

PR #785, opened 2020-01-24
allow hive operator to deploy/tear down the v1alpha1 aggregated api

joelddiaz:
Introduce HiveAPIEnabled boolean to HiveConfig.

Add static yaml Service, APIService, Deployment, ServiceAccount, and ClusterRoleBindings to config/apiserver/.

Ensure 'make deploy' will push out the new ClusterRoleBindings by adding it to the kustomize pipeline.

Allow the hive operator to bring up/tear down the aggregated v1alpha1 components based on the HiveConfig boolean.

Change the existing hack scripts for deploying the api server so that the certs are not created under config/* so that they don't get generated/saved into the repo.


config/apiserver/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+      apiserver: "true"
+  template:
+    metadata:
+      labels:
+        api: hiveapi
+        apiserver: "true"
+    spec:
+      containers:
+      # By default we will use the latest CI images published from hive master:
+      - image: registry.svc.ci.openshift.org/openshift/hive-v4.0:hive
```
dgoodwin: Shouldn't this be registry.svc.ci.openshift.org/openshift/hivev1:hive? I'm worried something's hiding issues in your test flow because I think this is the v1alpha1 image that shouldn't have the binary we're trying to run here.

staebler: This should be the one published from v1 not master.

config/apiserver/apiserver.template.yaml
```diff
@@ -109,13 +109,13 @@ objects:
         volumes:
         - name: apiserver-certs
           secret:
-            secretName: ${SVC_NAME}
```
staebler: Is this file still used for anything?

hack/deploy-apiserver.sh
```diff
@@ -4,7 +4,7 @@ export SVC_NAME="${SVC_NAME:-hiveapi}"
 export SVC_NAMESPACE="${SVC_NAMESPACE:-hive}"
 
 oc process --local=true -f "config/apiserver/apiserver.template.yaml" \
-  -p CA_CRT="$(base64 -w0 config/apiserver/certificates/apiserver_ca.crt)" \
-  -p TLS_CRT="$(base64 -w0 config/apiserver/certificates/apiserver.crt)" \
-  -p TLS_KEY="$(base64 -w0 config/apiserver/certificates/apiserver.key)" \
+  -p CA_CRT="$(base64 -w0 hiveapi-certs/apiserver_ca.crt)" \
```
staebler: Is this file still used?

hack/make-apiserver-certs.sh
```diff
@@ -1,30 +1,31 @@
 #!/usr/bin/env bash
+set -e
 
 export SVC_NAME="${SVC_NAME:-hiveapi}"
 export SVC_NAMESPACE="${SVC_NAMESPACE:-hive}"
 
-mkdir -p "config/apiserver/certificates"
+mkdir "hiveapi-certs"
```
staebler: I am not sure why these needed to move. The config/apiserver/certificates folder was already being ignored by git. And .gitignore has not been updated to ignore the new hiveapi-certs folder.

pkg/operator/hive/hiveapi.go
```diff
@@ -0,0 +1,142 @@
...
+	is311, err := r.is311(hLog)
+	if err != nil {
+		hLog.Error("error detecting 3.11 cluster")
+		return err
+	}
+	// If we're running on vanilla Kube (mostly devs using kind), or OpenShift 3.x, we
+	// will not have access to the service cert injection we normally use. Lookup
+	// the cluster CA and inject into the APIServer.
+	// NOTE: If this is vanilla kube, you will also need to manually create a certificate
+	// secret, see hack/hiveadmission-dev-cert.sh for example of creating the secret.
```
staebler: Should direct to  hack/make-apiserver-certs.sh instead.

config/apiserver/apiserver.template.yaml
joelddiaz: i haven't killed off the current hack/deploy-apiserver.sh, so it still uses the template

hack/make-apiserver-certs.sh
joelddiaz: ignored by git, but hack/update-bindata.sh was picking up those files (mostly just trying to follow what we've done for hiveadmission files)

config/apiserver/deployment.yaml
joelddiaz: i've been testing with my own custom containers, since that's where all the new hive operator code is

hack/make-apiserver-certs.sh
staebler: Ah, OK.

hack/deploy-apiserver.sh
joelddiaz: once we retire the ./hack/make-apiserver-certs.sh and ./hack/deploy-apiserver.sh scripts, these should go away

config/apiserver/apiservice.yaml
```diff
@@ -0,0 +1,19 @@
...
+    apiserver: "true"
+  annotations:
+    service.alpha.openshift.io/inject-cabundle: "true"
+spec:
+  group: hive.openshift.io
+  groupPriorityMinimum: 2000
+  service:
+    name: hiveapi
+    namespace: hive
+    port: 443
```
staebler: As far as I can tell, `port` is not a valid field.

hack/make-apiserver-certs.sh
staebler: Please add the hiveapi-certs directory to .gitignore.

dgoodwin:
Good catch on having it use the running hive image as we do for hive-controllers.

/lgtm

joelddiaz:
@dgoodwin fyi, i had to rebase to fix a merge conflict in hack/deploy-apiserver.sh so need a new lgtm

joelddiaz:
@dgoodwin sorry, forgot to push the generated pkg/operator/assets/bindata.go with the previous merge conflict resolution

config/apiserver/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+        apiserver: "true"
+    spec:
+      containers:
+      # By default we will use the latest CI images published from hive master:
+      - image: registry.svc.ci.openshift.org/openshift/hivev1:hive
+        name: apiserver
+        resources:
+          requests:
+            cpu: 100m
+            memory: 20Mi
```
staebler: The request and limit for memory should be 100Mi and 200Mi now.

config/apiserver/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+          limits:
+            cpu: 100m
+            memory: 30Mi
+        command:
+          - "/opt/services/hive-apiserver"
+        args:
+          - "start"
+          - "--tls-cert-file=/apiserver.local.config/certificates/tls.crt"
+          - "--tls-private-key-file=/apiserver.local.config/certificates/tls.key"
```
staebler: Add the `--logtostderr` argument.

config/apiserver/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+            cpu: 100m
+            memory: 100Mi
+          limits:
+            cpu: 100m
+            memory: 200Mi
+        command:
+          - "/opt/services/hive-apiserver"
+        args:
+          - "start"
+          - "--alsologtostderr"
```
staebler: Out of curiosity, what do we get out of logging to file as well as stderr?

joelddiaz: none. :) i just copy-pasted the wrong CLI argument from the apiserver help output

joelddiaz: i went ahead and changed it to --logtostderr (as originally intended)

PR #787, opened 2020-01-24
Add cd name to sample syncsets.

abutcher:
To keep syncsets created in the same namespace from colliding.

selectorsyncsets are applied to the same label so they don't need a prefix.


dgoodwin:
/lgtm

Probably won't affect anyone today as our instructions indicate to make multiple clusters in separate namespaces.

PR #788, opened 2020-01-27
Add maintenanceMode support in HiveConfig.

dgoodwin:
If set to true, will scale down hive-controllers to 0 replicas. Only
useful in extreme situations where we need to ensure the controllers are
not manipulating data, as in our upcoming v1 migration.

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -130,6 +130,12 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 
 	r.includeGlobalPullSecret(hLog, h, instance, hiveDeployment)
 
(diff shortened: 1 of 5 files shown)


pkg/operator/hive/hive.go
```diff
@@ -130,6 +130,12 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 
 	r.includeGlobalPullSecret(hLog, h, instance, hiveDeployment)
 
+	if instance.Spec.MaintenanceMode != nil && *instance.Spec.MaintenanceMode {
+		hLog.Warn("maintenanceMode enabled in HiveConfig, setting hive-controllers replicas to 0")
+		replicas := int32(0)
+		hiveDeployment.Spec.Replicas = &replicas
```
staebler: This can be simplified somewhat to `pointer.Int32Ptr(0)`. But it is not necessary.

PR #790, opened 2020-01-29
apiserver: watch support

staebler:


staebler:
Built on #789

staebler:
> I did notice that my v1alpha1 watches would drop me back to shell after a fairly short period of time, is that expected with kg clusterdeployment.v1alpha1.hive.openshift.io -w ? No info in the hiveapi pod logs related.

@dgoodwin I also notice that the watch closes after 1-1.5 minutes. This is due to one of the underlying watches closing. Watches of v1 resources close after 1-2 minutes as well. This may be something generic to CRs or something that we are doing wrong with our CRDs.

staebler:
> This PR does not contain any hacks for resource version correct? I can't see it and in testing it doesn't seem to, it just reports the same version from the v1 type.

That is correct. This PR does not have any code for fixing up the resource version.

pkg/hive/apiserver/registry/clusterdeployment/watch.go
```diff
@@ -0,0 +1,206 @@
...
+			pw.logger.Debug("received on secret channel")
+		case sourceEvent, ok = <-machinePoolSource:
+			pw.logger.Debug("received on machinepool channel")
+		}
+		if !ok {
+			pw.logger.Debug("shutting down receive")
+			break
+		}
+
+		if pw.stopping() {
```
joelddiaz: is this necessary since the for loop is already stopping on this condition, or could the for be an infinite loop?

staebler: Yes, the for loop should be able to be an infinite loop. There is a happens-before edge in `stopping()` that guarantees that the channels are all closed when the `stopped` field is true. So, if we remove the `stopping()` check from the for loop, then if the code starts a new loop when the watch has been stopped, the select statement will return immediately with `ok` set to false due to the channels being closed.

staebler:
94b74aaf...09728ea0
* Rebased onto latest v1
09728ea0...7edc0d7d
* Removed unnecessary `pw.stopping()` check in for loop as pointed out [here](https://github.com/openshift/hive/pull/790#discussion_r373674945).

PR #792, opened 2020-01-29
Add an install log scan regex for AWS NAT gateway limit exceeded errors.

dgoodwin:
Surfaces as: "Error: Error creating NAT Gateway:
NatGatewayLimitExceeded: Performing this operation would exceed the limit
of 5 NAT gateways"

Scanning just for the AWS error string seems easiest and safe.


staebler:
/lgtm cancel

Need to fix the manifests.

dgoodwin:
My bad, updated now.

PR #793, opened 2020-01-30
chore: label SyncSetInstances with (Selector)SyncSetRef

djzager:
Add a label to SyncSetInstances with the SelectorSyncSet or SyncSet they
are associated with. This makes it possible, knowing the (Selector)SyncSet, to find SyncSetInstances that were created.

### DIFF
--- a/pkg/controller/syncset/syncset_controller.go
+++ b/pkg/controller/syncset/syncset_controller.go
@@ -396,6 +396,9 @@ func (r *ReconcileSyncSet) syncSetInstanceForSyncSet(cd *hivev1.ClusterDeploymen
 			Name:            syncSetInstanceNameForSyncSet(cd, syncSet),
 			Namespace:       cd.Namespace,
 			OwnerReferences: []metav1.OwnerReference{*cdRef},
+			Labels: map[string]string{
+				"SyncSetName": syncSet.Name,
+			},
 		},
 		Spec: hivev1.SyncSetInstanceSpec{
 			ClusterDeploymentRef: corev1.LocalObjectReference{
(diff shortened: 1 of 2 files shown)


pkg/controller/syncset/syncset_controller.go
```diff
@@ -396,6 +396,9 @@ func (r *ReconcileSyncSet) syncSetInstanceForSyncSet(cd *hivev1.ClusterDeploymen
 			Name:            syncSetInstanceNameForSyncSet(cd, syncSet),
 			Namespace:       cd.Namespace,
 			OwnerReferences: []metav1.OwnerReference{*cdRef},
+			Labels: map[string]string{
+				"SyncSetName": syncSet.Name,
```
staebler: Use the label key of `hive.openshift.io/sync-set-name`.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -421,6 +424,9 @@ func (r *ReconcileSyncSet) syncSetInstanceForSelectorSyncSet(cd *hivev1.ClusterD
 			Name:            syncSetInstanceNameForSelectorSyncSet(cd, selectorSyncSet),
 			Namespace:       cd.Namespace,
 			OwnerReferences: []metav1.OwnerReference{*cdRef},
+			Labels: map[string]string{
+				"SelectorSyncSetName": selectorSyncSet.Name,
```
staebler: Use the label key of `hive.openshift.io/selector-sync-set-name`.

twiest:
Yeah, I agree. Thanks @djzager.

I added a [PR](https://github.com/openshift/hive/pull/794) that adds similar functionality plus cleans up a few other issues with the labels.

Does that PR address what you were looking for?

PR #794, opened 2020-01-30
Add labels for syncsetinstance

twiest:
- [x] Add labels for syncsetinstance
- [x] Ensure syncset and selectorsyncset labels are passed down to
      syncsetinstance
- [x] Update unit tests to check for labels.

### DIFF
--- a/pkg/controller/syncset/syncset_controller.go
+++ b/pkg/controller/syncset/syncset_controller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
+	k8slabels "k8s.io/kubernetes/pkg/util/labels"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -23,6 +24,7 @@ import (
 
 	"github.com/openshift/hive/pkg/apis/helpers"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 )
 
@@ -386,12 +388,14 @@ func findSyncSetInstanceForSelectorSyncSet(name string, syncSetInstances []*hive
 }
 
 func (r *ReconcileSyncSet) syncSetInstanceForSyncSet(cd *hivev1.ClusterDeployment, syncSet *hivev1.SyncSet) (*hivev1.SyncSetInstance, error) {
+	ssiLogger := r.logger.WithField("syncSetName", syncSet.Name)
 	cdRef := metav1.NewControllerRef(cd, hivev1.SchemeGroupVersion.WithKind("ClusterDeployment"))
 	hash, err := r.computeHash(syncSet.Spec)
 	if err != nil {
 		return nil, err
 	}
-	return &hivev1.SyncSetInstance{
+
+	syncSetInstance := &hivev1.SyncSetInstance{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            syncSetInstanceNameForSyncSet(cd, syncSet),
 			Namespace:       cd.Namespace,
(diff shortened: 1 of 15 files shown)


pkg/controller/syncset/syncset_controller.go
```diff
@@ -386,16 +387,28 @@ func findSyncSetInstanceForSelectorSyncSet(name string, syncSetInstances []*hive
...
 func (r *ReconcileSyncSet) syncSetInstanceForSyncSet(cd *hivev1.ClusterDeployment, syncSet *hivev1.SyncSet) (*hivev1.SyncSetInstance, error) {
+	ssiLogger := r.logger.WithField("syncsetName", syncSet.Name)
 	cdRef := metav1.NewControllerRef(cd, hivev1.SchemeGroupVersion.WithKind("ClusterDeployment"))
 	hash, err := r.computeHash(syncSet.Spec)
 	if err != nil {
 		return nil, err
 	}
+
+	ssiLogger.Debug("Apply syncset labels to syncsetinstance")
+	newLabels := controllerutils.CloneLabels(syncSet.Labels)
```
twiest: @staebler I added this because we discussed last week that "child" objects should inherit their parent's labels.

Is this correct or should I remove this?

pkg/controller/syncset/syncset_controller.go
```diff
@@ -411,16 +424,28 @@ func (r *ReconcileSyncSet) syncSetInstanceForSyncSet(cd *hivev1.ClusterDeploymen
...
 func (r *ReconcileSyncSet) syncSetInstanceForSelectorSyncSet(cd *hivev1.ClusterDeployment, selectorSyncSet *hivev1.SelectorSyncSet) (*hivev1.SyncSetInstance, error) {
+	ssiLogger := r.logger.WithField("selectorSyncsetName", selectorSyncSet.Name)
 	cdRef := metav1.NewControllerRef(cd, hivev1.SchemeGroupVersion.WithKind("ClusterDeployment"))
 	hash, err := r.computeHash(selectorSyncSet.Spec)
 	if err != nil {
 		return nil, err
 	}
+
+	ssiLogger.Debug("Apply selectorsyncset labels to syncsetinstance")
+	newLabels := controllerutils.CloneLabels(selectorSyncSet.Labels)
```
twiest: Same question here (should it inherit its parent's labels)?

pkg/controller/syncset/syncset_controller.go
staebler: I was wrong about inheriting parents' labels. It is not widespread practice. I gave the example of labels on a deployment getting inherited down through the replicaset, job, and pod. That is not the case. The labels that get passed down are the labels in the pod spec template of the deployment.

I still think it is a good idea to inherit some labels. For example, I like all of the resources related to a specific cluster deployment to have a label with the name of the cluster deployment. However, I can be persuaded that we should not inherit **all** labels.

For what it's worth, I found some functions in the k8s.io/kubernetes repo that may be helpful. For example, there is already a function to do essentially what `controllerutils.AddLabel` does. See https://github.com/kubernetes/kubernetes/blob/b7394102d6ef778017f2ca4046abbaa23b88c290/pkg/util/labels/labels.go#L56.

dgoodwin:
All looks fine to me, but holding off merge to see if the kube controllerutils function is worth using.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -386,16 +387,28 @@ func findSyncSetInstanceForSelectorSyncSet(name string, syncSetInstances []*hive
...
 
 func (r *ReconcileSyncSet) syncSetInstanceForSyncSet(cd *hivev1.ClusterDeployment, syncSet *hivev1.SyncSet) (*hivev1.SyncSetInstance, error) {
+	ssiLogger := r.logger.WithField("syncsetName", syncSet.Name)
 	cdRef := metav1.NewControllerRef(cd, hivev1.SchemeGroupVersion.WithKind("ClusterDeployment"))
 	hash, err := r.computeHash(syncSet.Spec)
 	if err != nil {
 		return nil, err
 	}
+
+	ssiLogger.Debug("Apply syncset labels to syncsetinstance")
```
csrwng: the debug stmt might be more useful if you say what label(s) you're talking about.
(same on lines 403, 434, 441)

pkg/controller/syncset/syncset_controller.go
csrwng: I wonder if in this case, the right labels to clone would be the ones from the clusterdeployment (or both syncset and clusterdeployment). As it is, I don't see how the syncsetinstance would get a label saying that what clusterdeployment it is for.

twiest:
@csrwng @staebler @dgoodwin I believe I've addressed your feedback. PTAL.

pkg/controller/syncset/syncset_controller.go
```diff
@@ -23,7 +23,10 @@ import (
 
 	"github.com/openshift/hive/pkg/apis/helpers"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+
+	k8slabels "k8s.io/kubernetes/pkg/util/labels"
```
staebler: Put this import with the group of `k8s.io` imports.

twiest: Fixed.

pkg/controller/syncset/syncset_controller.go
twiest: fixed.

PR #795, opened 2020-02-03
Formalize addition of ssh known hosts in install manager.

dgoodwin:
Originally done as part of a PoC for bare metal provisioning with a
dedicated libvirt hypervisor host, we now wish to make this a more long
term hive feature.

As a first step in this, this change removes the ssh known host
annotation and makes it a formal part of the Hive API. The hosts are now
represented in in spec.Provisioning.SSHKnownHosts. This is generic and
not tied to bare metal provisioning as we're not sure the extent we may
need ssh in installs in the future. We already use ssh to gather logs on
failures and while this does not currently require known hosts entries,
it seems best to keep the door open.

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -75,7 +75,7 @@ const (
 	defaultInstallConfigMountPath       = "/installconfig/install-config.yaml"
 	defaultPullSecretMountPath          = "/pullsecret/" + corev1.DockerConfigJsonKey
 	defaultManifestsMountPath           = "/manifests"
-	hiveSSHKnownHostsAnnotation         = "hive.openshift.io/ssh-known-host"
+	defaultHomeDir                      = "/home/hive" // Used if no HOME env var set.
 )
 
 var (
@@ -267,8 +267,13 @@ func (m *InstallManager) Run() error {
 	}
 	m.log.Infof("copied %s to %s", m.InstallConfigMountPath, destInstallConfigPath)
 
-	if knownHost, ok := cd.Annotations[hiveSSHKnownHostsAnnotation]; ok {
-		err = m.setupSSHKnownHost(knownHost)
+	if cd.Spec.Provisioning != nil && len(cd.Spec.Provisioning.SSHKnownHosts) > 0 {
+		err = m.setupSSHUser()
+		if err != nil {
+			m.log.WithError(err).Error("error setting up SSH known_hosts")
+			return err
+		}
+		err = m.writeSSHKnownHosts(getHomeDir(), cd.Spec.Provisioning.SSHKnownHosts)
 		if err != nil {
 			m.log.WithError(err).Error("error setting up SSH known_hosts")
 			return err
@@ -1169,11 +1174,10 @@ func waitForProvisioningStage(provision *hivev1.ClusterProvision, m *InstallMana
 	return errors.Wrap(err, "ClusterProvision did not transition to provisioning stage")
 }
 
-func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
+func (m *InstallManager) setupSSHUser() error {
 
 	// Add our potentially random UID to /etc/passwd so ssh works. Need to shell out here as
 	// the go libraries for user info appear to rely on /etc/passwd, which our user is not in.
-	// TODO: temporary work so we can use libvirt over ssh, this whole function can go once we're no longer doing that.
 	out, err := exec.Command("id", "-u").Output()
 	if err != nil {
 		m.log.WithError(err).Error("error running id -u")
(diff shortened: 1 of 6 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -267,8 +266,13 @@ func (m *InstallManager) Run() error {
...
 
-	if knownHost, ok := cd.Annotations[hiveSSHKnownHostsAnnotation]; ok {
-		err = m.setupSSHKnownHost(knownHost)
+	if cd.Spec.Provisioning != nil && len(cd.Spec.Provisioning.SSHKnownHosts) > 0 {
+		err = m.setupSSHUser()
+		if err != nil {
+			m.log.WithError(err).Error("error setting up SSH known_hosts")
+			return err
+		}
+		err = m.writeSSHKnownHosts("/home/hive", cd.Spec.Provisioning.SSHKnownHosts)
```
dhellmann: When we were working on the libvirt integration we ended up needing to change the home directory of the user so it pointed to a shared volume. Should this read the HOME environment variable instead of assuming "/home/hive"?

pkg/installmanager/installmanager.go
```diff
@@ -1194,17 +1197,21 @@ func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
...
-	sshDir := "/home/hive/.ssh"
+func (m *InstallManager) writeSSHKnownHosts(homeDir string, knownHosts []string) error {
+	sshDir := filepath.Join(homeDir, ".ssh")
 	if err := os.MkdirAll(sshDir, 0700); err != nil {
 		m.log.WithError(err).Errorf("error creating %s directory", sshDir)
 		return err
 	}
-	if err := ioutil.WriteFile(filepath.Join(sshDir, "known_hosts"), []byte(knownHost), 0644); err != nil {
+	allKnownHosts := strings.Join(knownHosts[:], "\n")
+	if err := ioutil.WriteFile(filepath.Join(sshDir, "known_hosts"), []byte(allKnownHosts), 0644); err != nil {
```
dhellmann: Is there any chance this file would ever exist and contain content that needs to be preserved?

pkg/installmanager/installmanager.go
```diff
@@ -1194,17 +1197,21 @@ func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
...
 
-	sshDir := "/home/hive/.ssh"
+func (m *InstallManager) writeSSHKnownHosts(homeDir string, knownHosts []string) error {
+	sshDir := filepath.Join(homeDir, ".ssh")
 	if err := os.MkdirAll(sshDir, 0700); err != nil {
 		m.log.WithError(err).Errorf("error creating %s directory", sshDir)
 		return err
 	}
-	if err := ioutil.WriteFile(filepath.Join(sshDir, "known_hosts"), []byte(knownHost), 0644); err != nil {
+	allKnownHosts := strings.Join(knownHosts[:], "\n")
```
staebler: Why do we need to slice `knownHosts`?

pkg/installmanager/installmanager.go
```diff
@@ -1194,17 +1197,21 @@ func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
...
+	if err := ioutil.WriteFile(filepath.Join(sshDir, "known_hosts"), []byte(allKnownHosts), 0644); err != nil {
 		m.log.WithError(err).Error("error writing ssh known_hosts")
 		return err
 	}
-	m.log.Infof("Wrote known host to %s/known_hosts: %s", sshDir, knownHost)
+	m.log.Infof("Wrote known hosts to %s/known_hosts: %v", sshDir, knownHosts)
```
staebler: I recommend putting `knownHosts` in a log field rather than in the log message.

pkg/installmanager/installmanager_test.go
```diff
@@ -647,6 +651,55 @@ func TestInstallManagerSSH(t *testing.T) {
...
+			expectedFile: `192.168.86.100 ecdsa-sha2-nistp256 FOOBAR
+192.168.86.101 ecdsa-sha2-nistp256 FOOBAR2
+192.168.86.102 ecdsa-sha2-nistp256 FOOBAR3`,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			tempDir, err := ioutil.TempDir("", "installmanagersshknownhosts")
+			require.NoError(t, err)
```
staebler: Add messages to the require and assert calls to aid in readability of test failures.

pkg/installmanager/installmanager_test.go
```diff
@@ -647,6 +651,55 @@ func TestInstallManagerSSH(t *testing.T) {
...
+			defer os.RemoveAll(tempDir)
+
+			im := InstallManager{
+				log:      log.WithField("test", test.name),
+				LogLevel: "debug",
```
staebler: Neither `LogLevel` nor `WorkDir` are used by this test. Why set them? Moreover, if we want to set `WorkDir`, it should be a different directory than the directory passed as an argument to `writeSSHKnownHosts`.

pkg/installmanager/installmanager.go
staebler: +1. See https://github.com/openshift/hive/pull/723#discussion_r356062465.

pkg/installmanager/installmanager.go
dgoodwin: Bad copy paste off the interweb.

pkg/installmanager/installmanager.go
dgoodwin: Good idea, adjusting...

pkg/installmanager/installmanager.go
dgoodwin: If it's in the container image something is really wrong. If it's mounted in via kube we're going to fail to be able to write it. (IIRC) I don't really want to take on the edge cases around newlines at the end of file or not for something that really shouldn't be a concern. 

How about a hard error if the file exists before we go to write and we can deal with it if that comes up.

dgoodwin: OR we're using /output as HOME and re-trying a failed install. :) But in this case we still want to overwrite rather than append or check if we need to append, which would lean me towards keeping it as is. Thoughts?

dgoodwin:
Updates pushed for everything but the possibility of a pre-existing known hosts file.

pkg/installmanager/installmanager.go
staebler: I don't understand the second scenario. However, I am fine with either a hard error or a warning log entry if the file exists.

dhellmann: The only reasons for the file to exist that I could think of are the existing volume and a custom image. In the case of the volume, it seems like you're right and we should consider that a re-try attempt and write over the file. So, do we support custom images for any reason?

dgoodwin: Custom images, only for devs testing changes really.

dhellmann: I don't think we need to spend a lot of energy on custom developer images, so if we don't support them for users I think it's safe to always try to overwrite the file.

staebler: Yeah, the more I think about this, the more that I am OK with doing nothing here. We don't need to do anything to detect an existing file.

staebler: Just don't ever try to run this binary outside of a container. ;)

dhellmann: :boom:

dgoodwin: That's what backups are for. :)

Thanks, feedback addressed then.

pkg/installmanager/installmanager.go
```diff
@@ -1188,23 +1192,27 @@ func (m *InstallManager) setupSSHKnownHost(knownHost string) error {
 		return err
 	}
 	defer f.Close()
-	passwdLine := fmt.Sprintf("default:x:%s:0:default user:/home/hive:/sbin/nologin\n", uid)
+	passwdLine := fmt.Sprintf("default:x:%s:0:default user:%s:/sbin/nologin\n", getHomeDir(), uid)
```
staebler: Did the order here get mixed up? The first `%s` should be the uid and the second should be the home directory, right?

dgoodwin: Wow, that would now have been fun to debug. Nice catch.

dgoodwin: Fixed.

PR #797, opened 2020-02-04
apiserver: clusterdeployment resource version amalgamation

staebler:
Set the resource version of v1alpha1 clusterdeployment to be a hyphen-delimited concatenation of the resource versions from the v1 clusterdeployment, the install-config secret, and the v1 machinepools.

For example,
if the v1 clusterdeployment has a resource version of "cd",
the install-config secret has a resource version of "ic",
the worker machine pool has a resource version of "worker",
and the infra machine pool has a resource version of "infra",
then the resource version of the v1alpha1 clusterdeployment will be "cd-ic-infra-worker".

Note that the machine pool resource versions are ordered alphabetically by the name of the machinepools.

### DIFF
--- a/pkg/hive/apiserver/registry/util/convert.go
+++ b/pkg/hive/apiserver/registry/util/convert.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"fmt"
+	"strings"
 
 	netopv1 "github.com/openshift/cluster-network-operator/pkg/apis/networkoperator/v1"
 	"github.com/openshift/installer/pkg/ipnet"
@@ -35,9 +36,15 @@ func CheckpointFromHiveV1(in *hivev1.Checkpoint, out *hiveapi.Checkpoint) error 
 func ClusterDeploymentToHiveV1(in *hiveapi.ClusterDeployment, sshKey string, out *hivev1.ClusterDeployment, installConfig *installtypes.InstallConfig, machinePools *[]*hivev1.MachinePool) error {
 	in = in.DeepCopy()
 
+	// Preserve the resource version since the v1alpha1 resource version is an amalgamation of resource versions from
+	// multiple resources.
+	outResourceVersion := out.ResourceVersion
+
 	if err := hiveconversion.Convert_v1alpha1_ClusterDeployment_To_v1_ClusterDeployment(in, out, nil); err != nil {
 		return err
 	}
+
+	out.ResourceVersion = outResourceVersion
 
 	installConfig.ObjectMeta.Name = in.Spec.ClusterName
 	installConfig.TypeMeta.APIVersion = installtypes.InstallConfigVersion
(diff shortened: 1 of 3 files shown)


csrwng:
Hey @staebler so would it be possible to render a previous resource version if you add a new machine pool and then delete it? (I don't remember what was concluded from that discussion)

staebler:
> Hey @staebler so would it be possible to render a previous resource version if you add a new machine pool and then delete it? (I don't remember what was concluded from that discussion)

@csrwng Yes, it will be possible to render a previous resource version. We deemed this scenario to be acceptable given the way that machine pools are being used in prod. The machine pools will be created when the cluster deployment is created and will not be deleted until the cluster deployment is deleted.

@dgoodwin Is that how you recollect things?

dgoodwin:
Yes, I am ok with that tradeoff for a limited period of time, it should not be an issue in the real world.

pkg/hive/apiserver/registry/util/convert.go
```diff
@@ -157,13 +164,21 @@ func ClusterDeploymentToHiveV1(in *hiveapi.ClusterDeployment, sshKey string, out
...
+func ClusterDeploymentFromHiveV1(in *hivev1.ClusterDeployment, installConfig *installtypes.InstallConfig, machinePools []*hivev1.MachinePool, installConfigResourceVersion string, out *hiveapi.ClusterDeployment) error {
 	in = in.DeepCopy()
 
 	if err := hiveconversion.Convert_v1_ClusterDeployment_To_v1alpha1_ClusterDeployment(in, out, nil); err != nil {
 		return err
 	}
 
+	resourceVersions := make([]string, 2+len(machinePools))
+	resourceVersions[0] = in.ResourceVersion
+	resourceVersions[1] = installConfigResourceVersion
```
csrwng: should you have a place holder for when the installConfigResourceVersion is blank?

staebler: The placeholder is the empty string. What would we get out of a non-empty placeholder?

csrwng: Just less confusion when you examine a resource and it has `aaa--bbb` instead of `aaa-NONE-bbb` (the latter is more explicit imho)

staebler: My preference would be to leave it as an empty string.

1) There should rarely if ever be a case where there is not an install-config secret.
2) The resource version is not something that should usually have to be inspected by a user.
3) This would require additional code in the creation of the resource version as well as the parsing of the resource version for its constituent parts.

csrwng: 👍

staebler:
@csrwng I stupidly rebased this onto the latest v1, losing your lgtm.

staebler:
> /lgtm for Cesar. :)

If you really intended to tag this as lgtm, then the /lgtm needs to be on its own line.

PR #799, opened 2020-02-05
controllers: return when syncsetinstance controller cannot build restconfig

staebler:


dgoodwin:
I've changed this to now go to master.

/lgtm

PR #809, opened 2020-02-08
Migrate hive objects to have expected v1 labels

twiest:
- [x] Migrate hive objects to have expected v1 labels
- [x] Test manually
- [x] Add unit tests

### DIFF
--- a/contrib/pkg/labelobjects/labelobjects.go
+++ b/contrib/pkg/labelobjects/labelobjects.go
@@ -1,0 +1,918 @@
+package labelobjects
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	contributils "github.com/openshift/hive/contrib/pkg/utils"
+	"github.com/pkg/errors"
+	logger "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hivev1alpha1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
+	"github.com/openshift/hive/pkg/constants"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8slabels "k8s.io/kubernetes/pkg/util/labels"
+
+	hivecontrolplanecerts "github.com/openshift/hive/pkg/controller/controlplanecerts"
+	hiveremoteingress "github.com/openshift/hive/pkg/controller/remoteingress"
+	hivesyncidentityprovider "github.com/openshift/hive/pkg/controller/syncidentityprovider"
+	hiveimageset "github.com/openshift/hive/pkg/imageset"
+)
+
+// Options contains the options for labeling related Hive objects
+type Options struct {
+	Logger logger.FieldLogger
+	DryRun bool
+}
+
+// NewLabelObjectsCommand adds a subcommand for labeling Hive objects in a cluster.
+func NewLabelObjectsCommand() *cobra.Command {
+	opt := &Options{}
+	logLevel := "info"
+	cmd := &cobra.Command{
+		Use:   "label-objects",
+		Short: "Labels related Hive objects in the cluster",
+		Run: func(cmd *cobra.Command, args []string) {
+			// Set log level
+			level, err := logger.ParseLevel(logLevel)
+			if err != nil {
+				logger.WithError(err).Error("Cannot parse log level")
+				os.Exit(1)
+			}
+
+			log := logger.NewEntry(&logger.Logger{
+				Out: os.Stdout,
+				Formatter: &logger.TextFormatter{
+					FullTimestamp: true,
+				},
+				Hooks: make(logger.LevelHooks),
+				Level: level,
+			})
+			opt.Logger = log
+
+			kubeclient, err := contributils.GetClient()
+			if err != nil {
+				log.WithError(err).Fatal("error creating kube clients")
+			}
+
+			if _, err := opt.run(kubeclient); err != nil {
+				log.WithError(err).Fatal("Runtime error")
+			}
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&logLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
+	flags.BoolVar(&opt.DryRun, "dry-run", false, "Determines if the migration should change values in the cluster")
+
+	return cmd
+}
+
+type labelMigrator struct {
+	client    client.Client
+	dryRun    bool
+	logger    logger.FieldLogger
+	succeeded bool
+
+	clusterProvisionsSuccessCount               int
+	installLogPVCsSuccessCount                  int
+	imagesetJobsSuccessCount                    int
+	clusterDeprovisionsSuccessCount             int
+	dnsZonesSuccessCount                        int
+	mergedPullSecretsSuccessCount               int
+	clusterDeprovisionJobsSuccessCount          int
+	clusterProvisionJobsSuccessCount            int
+	clusterStatesSuccessCount                   int
+	controlPlaneCertificateSyncSetsSuccessCount int
+	remoteIngressSyncSetsSuccessCount           int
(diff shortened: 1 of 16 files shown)


contrib/pkg/labelobjects/labelobjects.go
```diff
@@ -0,0 +1,928 @@
...
+	}
+
+	for _, item := range all.Items {
+		objLogger := l.newObjectLogger(item.Name, item.Namespace, "PersistentVolumeClaim")
+
+		val, ok := item.Labels[constants.InstallJobLabel]
+		if !ok || val != "true" {
+			// Not a PVC that we're interested in. Move on.
+			continue
+		}
```
dgoodwin: I think you could bypass this label check if you wanted by filtering in the List: https://github.com/openshift/hive/blob/master/pkg/controller/metrics/metrics.go#L232

contrib/pkg/labelobjects/labelobjects.go
```diff
@@ -0,0 +1,928 @@
...
+	err := l.client.List(context.TODO(), all)
+	if err != nil {
+		l.imagesetJobsFailureCount++
+		return errors.Wrap(err, "Failed listing Jobs")
+	}
+
+	for _, item := range all.Items {
+		objLogger := l.newObjectLogger(item.Name, item.Namespace, "Job")
+
+		val, ok := item.Labels[hiveimageset.ImagesetJobLabel]
```
dgoodwin: Similar to above, could probably filter in List.

contrib/pkg/labelobjects/labelobjects.go
```diff
@@ -0,0 +1,928 @@
...
+			continue
+		}
+
+		name, found := findControllerOwnerNameByKind(&item, "ClusterDeployment")
+		if !found {
+			continue
+		}
+
+		item.Labels = k8slabels.AddLabel(item.Labels, constants.ClusterDeploymentNameLabel, name)
+		item.Labels = k8slabels.AddLabel(item.Labels, constants.SecretTypeLabel, constants.SecretTypeMergedPullSecret)
```
dgoodwin: This looks a little off, it seems odd to assume that any secret of type dockerconfigjson, with a ClusterDeployment owner, is a merged pull secret, but on closer inspection that does appear to be correct, normal non-merged pull secrets would not be owned by their ClusterDeployment. Just calling it out in case anyone sees a flaw in my logic.

contrib/pkg/labelobjects/labelobjects.go
```diff
@@ -0,0 +1,928 @@
...
+	err := l.client.List(context.TODO(), all)
+	if err != nil {
+		l.clusterDeprovisionJobsFailureCount++
+		return errors.Wrap(err, "Failed listing Jobs")
+	}
+
+	for _, item := range all.Items {
+		objLogger := l.newObjectLogger(item.Name, item.Namespace, "Job")
+
+		val, ok := item.Labels[constants.UninstallJobLabel]
```
dgoodwin: List on labels.

contrib/pkg/labelobjects/labelobjects.go
```diff
@@ -0,0 +1,928 @@
...
+	err := l.client.List(context.TODO(), all)
+	if err != nil {
+		l.clusterProvisionJobsFailureCount++
+		return errors.Wrap(err, "Failed listing Jobs")
+	}
+
+	for _, item := range all.Items {
+		objLogger := l.newObjectLogger(item.Name, item.Namespace, "Job")
+
+		val, ok := item.Labels[constants.InstallJobLabel]
```
dgoodwin: List on labels.

contrib/pkg/labelobjects/labelobjects.go
staebler: It is probably true for the Dedicated clusters. It is not necessarily true for all users of Hive. For example, I would not be surprised if MCM put an owner ref from the non-merged pull secret to the clusterdeployment so that the secret was cleaned up when the clusterdeployment was.

staebler: Forget to mention in the comment that since it is true for Dedicated, then we are probably good. Unless we are expecting other users of v1alpha1 to use our migration steps to migrate to v1--which scares me.

staebler: But it may be better to use the same pattern that is being used for the Hive-managed syncsets. Namely, we can look for the merged pull secret by the expected name of the secret.

twiest: I agree. Thanks for the feedback.

twiest:
@dgoodwin @staebler Thanks for the feedback, I believe I've addressed your comments. PTAL.

dgoodwin:
Apologies not sure if I should have done this but I changed the base to master, now you have a conflict. :) Sorry!

contrib/pkg/labelobjects/labelobjects_test.go
```diff
@@ -0,0 +1,1188 @@
...
+			existingObjects: []runtime.Object{
+				buildClusterProvision(clusterProvisionBase(), genericClusterProvision(generic.WithControllerOwnerReference(fakeClusterDeployment()))),
+				persistentvolumeclaim.Build(persistentVolumeClaimBase(),
+					persistentvolumeclaim.Generic(generic.WithLabel(constants.InstallJobLabel, "true")),
+					persistentvolumeclaim.Generic(generic.WithControllerOwnerReference(fakeClusterDeployment()))),
+				job.Build(
+					jobBase(),
+					job.Generic(generic.WithNamePostfix("1")),
+					job.Generic(generic.WithLabel(hiveimageset.ImagesetJobLabel, "true")),
+					job.Generic(generic.WithControllerOwnerReference(fakeClusterDeployment()))),
```
dgoodwin: To follow this I cloned the branch and clicked through roughly 6 functions to see what was going on to build this test job.

For discussion purposes, compare to:

```go
testJob(
    "somejob-1", 
    namespace, 
    map[string]string{hiveimageset.ImagesetJobLabel: "true"}, 
    fakeClusterDeployment())
```

Maybe it's just because this is how we typically create these test objects but I find this clearer and easier to understand without jumping around very far. I feel like these builder functions are just another way to do things with roughly the same amount of code, more indirection, and another process the reader has to understand to work with it.

Granted it does mean "testJob" functions littered throughout various test modules.

What do we think, does this provide benefits I'm missing?

twiest: I see what you're saying, but I think that may just be a familiarity thing. It took me awhile to feel comfortable with it when @staebler suggested using it, but now it's second nature.

The problem with what you listed is that the `testJob` function parameters aren't flexible (you have a fixed number of params that must be passed in).

For example, if you wanted some jobs to have a label and some other jobs to have a controller reference, but neither to have both, you'd need multiple `testJob` functions.

With the builder pattern on the other hand, you just put in whatever builder functions you want and the object is created exactly as you've asked.

The builders are also very easy to extend without affecting existing users. Alternatively if you change the signature of the `testJob` function, you must now go change all callers and possibly declare more variables to pass in to the function.

When building the extensions, they can also live out in a package that all tests can import or alternatively if it's something just for your tests in this one package, you can just include the builder function in that one package.

It's really nice to only have to code the builder functions once, and then we can use them in any of our tests to build up any type of object that we need.

Also, using the `job.Generic` pattern allows me to use a bunch of general functions that work on any object that implements `metav1.Object` (which all core and hive objects do).

When we originally talked about using a pattern like this, we discussed using either this syntax (which we got from the pattern used for list options in controller runtime), or possibly using a fluent pattern (like what logrus uses).

Here's roughly the equivalent in a fluent syntax:
```golang
jobBuilder
    .Base()
    .WithNamePostfix("1")
    .WithLabel(hiveimageset.ImagesetJobLabel, "true")
    .WithControllerOwnerReference(fakeClusterDeployment())
```

Would this have made more sense?

In other words, is the objection to having the builder pattern, or is it the syntax of the builders?

dgoodwin: The objection was to using a builder pattern in general, which I generally consider for construction of complex objects, and I don't see the go structs in our tests as qualifying for that when were setting metadata and labels. Personally I would stick to the existing methods, or inline the structs if we need a lot of flexibility. 

This is good context above though and if you both are on board, based it off established controller runtime patterns and enjoy using it then I'm ok with it.

PR #811, opened 2020-02-10
Remove install of libvirt deps temporarily.

dgoodwin:
This was done for bare metal install work. We'll need another solution,
the rpms cannot be installed from UBI, or from RHEL images if you are
not entitled.

This was working in CI because they replace the FROM, but would fail
locally, or if run in the appsre job.


dgoodwin:
ART replaces all FROM lines so I think we should switch to CentOS for the base image upstream, let ART replace it down. Will test this once we're through migration.

PR #813, opened 2020-02-10
migration: do not mutate object when doing a dry run delete

staebler:


dgoodwin:
/lgtm

Are there bash fixes needed as well or was this it?

staebler:
Are there bash fixes needed as well or was this it?

No bash fixes. The issue was that the delete was not respecting the dry-run field set in the delete options. 

Outside of that, the hiveutil command was incorrectly removing finalizers during a dry-run as well. I know that the Hive admission webhooks reject dry-run updates. So, rather than working through everything, I have gone ahead and not skipped the update and delete commands entirely when doing a dry run.

PR #817, opened 2020-02-10
More fixes for hive RBAC.

dgoodwin:


pkg/operator/hive/hiveapi.go
```diff
@@ -30,7 +30,7 @@ func (r *ReconcileHiveConfig) deployHiveAPI(hLog log.FieldLogger, h *resource.He
 
 	err := util.ApplyAsset(h, "config/apiserver/hiveapi-cluster-role-binding.yaml", hLog)
 	if err != nil {
-		return err
+		hLog.WithError(err).Error("unable to apply hiveapi cluster role binding")
```
staebler: Should we take out this change? This was only in there to try to brute force the start of the apiserver yesterday.

staebler:
This is included in #818 

/close

PR #818, opened 2020-02-11
Fix missing bindata in rushed merges yesterday.

dgoodwin:


dgoodwin:
Added another fix for the bad rolebinding that broke e2e run. Also a rush merge from yesterday.

dgoodwin:
Fixes #821

PR #819, opened 2020-02-11
Separate cluster ssh key from libvirt connection ssh key.

dgoodwin:
For PoC purposes we assumed to use the same ssh key for dropping on the
cluster hosts as we do for connecting to libvirt for bare metal
provisioning. There is no good reason to do this so for a medium term
solution this change adds a new SSH key that can be specified for
libvirt provisioning.


dgoodwin:
needs #818

pkg/installmanager/installmanager.go
```diff
@@ -78,6 +78,8 @@ const (
 	defaultPullSecretMountPath          = "/pullsecret/" + corev1.DockerConfigJsonKey
 	defaultManifestsMountPath           = "/manifests"
 	defaultHomeDir                      = "/home/hive" // Used if no HOME env var set.
+	sshPrivKeyPathEnvVar                = "SSH_PRIV_KEY_PATH"
```
staebler: Should we put these definitions in a common place that can be used by `pkg/install/generate.go` as well.

pkg/installmanager/installmanager.go
```diff
@@ -231,19 +233,37 @@ func (m *InstallManager) Run() error {
...
-		sshCleanupFunc, sshAgentSetupErr := m.initSSHAgent(sshKeyPath)
+		sshKeyPaths = append(sshKeyPaths, sshKeyPath)
+	}
+
+	// setup the bare metal provisioning libvirt ssh key if possible
+	if cd.Spec.Platform.BareMetal != nil {
+		libvirtSSHKeyPath, libvirtSSHAgentSetupErr := m.initSSHKey(libvirtSSHPrivKeyPathEnvVar)
+		if libvirtSSHAgentSetupErr != nil {
+			// Not a fatal error.
+			m.log.WithError(libvirtSSHAgentSetupErr).Info("unable to initialize host ssh key")
```
staebler: Can we make this info message distinguishable from the one for the other ssh key? They both say "host ssh key".

pkg/installmanager/installmanager.go
```diff
@@ -231,19 +233,37 @@ func (m *InstallManager) Run() error {
...
+		} else {
+			sshKeyPaths = append(sshKeyPaths, libvirtSSHKeyPath)
+		}
+	}
+
+	if len(sshKeyPaths) > 0 {
+		m.log.Info("initializing ssh agent with %d keys", len(sshKeyPaths))
```
staebler: This needs to be `Infof`.

pkg/installmanager/installmanager.go
```diff
@@ -796,7 +825,7 @@ func (m *InstallManager) initSSHKey() (string, error) {
 
 }
 
-func (m *InstallManager) initSSHAgent(sshKeyPath string) (func(), error) {
+func (m *InstallManager) initSSHAgent(sshKeyPaths ...string) (func(), error) {
```
staebler: Is this function variadic for any reason other than not wanting to change the tests to pass it a slice?

I would prefer to see this as `initSSHAgent(sshKeyPaths []string)` unless we really want the variadic feature.

pkg/apis/hive/v1/baremetal/platform.go
```diff
@@ -1,5 +1,12 @@
...
 
+import corev1 "k8s.io/api/core/v1"
+
 // Platform stores the global configuration for the cluster.
 type Platform struct {
+	// LibvirtSSHPrivateKeySecretRef is the reference to the secret that contains the private SSH key to use
+	// for access to the libvirt provisioning host.
+	// The SSH private key is expected to be in the secret data under the "ssh-privatekey" key.
+	// +optional
+	LibvirtSSHPrivateKeySecretRef *corev1.LocalObjectReference `json:"libvirtSSHPrivateKeySecretRef,omitempty"`
```
staebler: Is this truly optional? Is this to cover the case where the ssh key provided in the `.spec.provisioning.sshPrivateKeySecretRef` is the same key used for libvirt? If so, I think that I would prefer that the user still explicitly specified the ssh key here.

pkg/installmanager/installmanager.go
```diff
@@ -760,16 +789,16 @@ func (m *InstallManager) gatherClusterLogs(cd *hivev1.ClusterDeployment) error {
...
 
-func (m *InstallManager) initSSHKey() (string, error) {
+func (m *InstallManager) initSSHKey(envVar string) (string, error) {
 	m.log.Debug("checking for SSH private key")
-	sshPrivKeyPath := os.Getenv("SSH_PRIV_KEY_PATH")
+	sshPrivKeyPath := os.Getenv(envVar)
 	if sshPrivKeyPath == "" {
-		m.log.Warn("cannot configure SSH agent as SSH_PRIV_KEY_PATH is unset or empty")
-		return "", fmt.Errorf("cannot configure SSH agent as SSH_PRIV_KEY_PATH is unset or empty")
+		m.log.Warnf("cannot configure SSH agent as %s is unset or empty", envVar)
```
staebler: This ends up generating a warn log entry here followed immediately by an info log entry in the caller with the same text.

pkg/installmanager/installmanager.go
```diff
@@ -861,11 +890,14 @@ func (m *InstallManager) initSSHAgent(sshKeyPath string) (func(), error) {
...
-	cmd := exec.Command(bin, sshKeyPath)
-	cmd.Stderr = os.Stderr
-	if err := cmd.Run(); err != nil {
-		m.log.WithError(err).Errorf("failed to add private key: %v", sshKeyPath)
-		return sshAgentCleanup, err
+	for _, sshKeyPath := range sshKeyPaths {
+		cmd := exec.Command(bin, sshKeyPath)
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			m.log.WithError(err).Errorf("failed to add private key: %v", sshKeyPath)
```
staebler: Use `WithField("key", sshKeyPath)` for consistency with the later `Info` call.

pkg/apis/hive/v1/baremetal/platform.go
dgoodwin: It's optional only because I don't think it will be required long term. The whole libvirt host to provision thing should go away someday so it seemed best to make this optional. If we can easily make it mandatory now and optional later then that would be fine, but initially I wouldn't think it ideal to change things back and forth in the API like that.

pkg/installmanager/installmanager.go
dgoodwin: Yeah just less test churn and seemed a little more concise. Updating to use slice, but what was the motivation here, just to avoid unexpected things if someone ever adds another arg?

dgoodwin:
Changes pushed minus making the libvirt ssh key required.

pkg/apis/hive/v1/baremetal/platform.go
dgoodwin: Hmm, should we leave it optional and validate that it's set today?

staebler: I think that it is easy to go from mandatory to optional but not the other way around. So if we start with optional, then it will always be optional.

dgoodwin: Ok updated to make required.

PR #820, opened 2020-02-11
Switch to FROM centos:7 and restore libvirt dependencies.

dgoodwin:
Needs #818


gregsheremeta:
what's the reason?

dgoodwin:
libvirt deps cannot be installed in the universal base image, they're not available, and use of any other RHEL based build will fail locally and in appsre pipelines due to lack of entitlements.

ART docs appear to indicate our upstream images can and should be based on centos. They override the FROM during CI builds so they're ultimately RHEL but using centos upstream lets things like this work.

dgoodwin:
Need an ack here before bare metal works again.

PR #822, opened 2020-02-11
migration: support oc 3.11

staebler:
Replace uses of "-A" with "--all-namespaces".

Replace "oc delete -A" with a delete in each individual namespace.


hack/v1migration/delete.sh
```diff
@@ -59,17 +59,18 @@ then
...
-if [[ -z $dry_run ]]
-then
-  echo "Deleting install jobs"
-  oc delete jobs -l "hive.openshift.io/install=true" -A
-fi
+delete_jobs() {
+  local labels=${1:?must specify a label for the jobs to delete}
+  oc get jobs -l "${labels}" --all-namespaces -ojsonpath='{.items[*].metadata.namespace}' | \
+    sort -u | \
+    xargs -i oc delete pods -n {} -l "${labels}" "${dry_run}"
```
dgoodwin: This looks like it should be jobs not pods?

PR #825, opened 2020-02-12
apiserver: fix rbac and port used

staebler:
Change the port that the apiserver is listening on to 10443 to match the target port set for the service.

Add a ClusterRoleBinding for the apiserver service account to system:auth-delegator.

Remove the code in hive-operator that was trying to set up for RBAC for the apiserver service account. This will always fail because the operator does not have the necessary permissions to pass to the apiserver. The RBAC needs to be set up for the apiserver in the same way that it is set up for the hiveadmission.

### DIFF
--- a/config/apiserver/deployment.yaml
+++ b/config/apiserver/deployment.yaml
@@ -34,7 +34,7 @@ spec:
         args:
           - "start"
           - "--v=2"
-          - "--secure-port=10433"
+          - "--secure-port=10443"
           - "--logtostderr"
           - "--tls-cert-file=/apiserver.local.config/certificates/tls.crt"
           - "--tls-private-key-file=/apiserver.local.config/certificates/tls.key"
--- a/pkg/operator/hive/hiveapi.go
+++ b/pkg/operator/hive/hiveapi.go
@@ -28,18 +28,11 @@ func (r *ReconcileHiveConfig) deployHiveAPI(hLog log.FieldLogger, h *resource.He
 		return r.tearDownHiveAPI(hLog)
 	}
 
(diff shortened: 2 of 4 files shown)


pkg/operator/hive/hiveapi.go
```diff
@@ -28,18 +28,11 @@ func (r *ReconcileHiveConfig) deployHiveAPI(hLog log.FieldLogger, h *resource.He
...
 	}
 
-	err := util.ApplyAsset(h, "config/apiserver/hiveapi-cluster-role-binding.yaml", hLog)
-	if err != nil {
-		return err
-	}
-
-	err = util.ApplyAsset(h, "config/apiserver/service.yaml", hLog)
-	if err != nil {
+	if err := util.ApplyAsset(h, "config/apiserver/service.yaml", hLog); err != nil {
```
joelddiaz: are we not applying the role binding from inside the hive operator b/c OLM is expected to perform this?

staebler: Yes. The hive operator does not have the RBAC permissions that the apiserver needs so cannot give the apiserver those permissions.

staebler: The OLM is doing this same task already for the hive admission service account.

config/apiserver/deployment.yaml
```diff
@@ -34,7 +34,7 @@ spec:
         args:
           - "start"
           - "--v=2"
-          - "--secure-port=10433"
+          - "--secure-port=10443"
```
dgoodwin: Doh.

pkg/operator/hive/hiveapi.go
dgoodwin: That is totally not what I remembered but absolutely correct.

PR #826, opened 2020-02-13
Remove now obsolete toleration edit in remoteingress controller.

dgoodwin:
This was done to force the rollout of ingress config, which should no
longer be needed as of 4.2.0.

The code in question is now causing a problem for hive users who need to
adjust default router tolerations to land them on specific nodes, also
being done by syncsets, and now both syncsets are fighting for the
tolerations causing the pods to restart often, impacting running
services.

With this change, remoteingress will no longer work seamlessly on 4.1
OpenShift clusters, manual restarts may be required. We do not believe
this will impact many users as 4.2 and 4.3 are both available now.

### DIFF
--- a/pkg/controller/remoteingress/remoteingress_controller.go
+++ b/pkg/controller/remoteingress/remoteingress_controller.go
@@ -49,8 +49,6 @@ const (
 	ingressCertificateNotFoundReason = "IngressCertificateNotFound"
 	ingressCertificateFoundReason    = "IngressCertificateFound"
 
-	ingressSecretTolerationKey = "hive.openshift.io/ingress"
-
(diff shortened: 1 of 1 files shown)


staebler:
> With this change, remoteingress will no longer work seamlessly on 4.1
> OpenShift clusters, manual restarts may be required. We do not believe
> this will impact many users as 4.2 and 4.3 are both available now.

Have we checked to see if there are indeed production users that are running 4.1? Do we even have insight into that?

dgoodwin:
@mwoodson confirmed no prod users are on 4.1 anymore, they upgrade the entire fleet fairly often.

PR #829, opened 2020-02-13
apiserver: fix rbac for olm deploy

staebler:
Follow the pattern that is used for hiveadmission to set up the RBAC for hiveapi.


staebler:
I also deleted some role bindings for hiveadmission that were redundant with the permissions that were already being granted in the system:openshift:hive:hiveadmission clusterrole.

PR #830, opened 2020-02-14
restart pods with injected certs when the certs rotate 

staebler:
The hiveadmission and hiveapi pods use certs that may change while the pods are running. When the certs change, the pods will no longer be able to do their work. The pods need to be restarted to pick up the new certs.

### DIFF
--- a/pkg/certs/certs.go
+++ b/pkg/certs/certs.go
@@ -1,0 +1,48 @@
+package certs
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+)
+
+var (
+	filenames = []string{"tls.crt", "tls.key"}
+)
+
+// TerminateOnCertChanges terminates the process when either tls.crt or tls.key in the specified certs directory change.
+func TerminateOnCertChanges(certsDir string) {
+	obs, err := fileobserver.NewObserver(10 * time.Second)
+	if err != nil {
+		log.WithError(err).Fatal("could not set up file observer to check for cert changes")
+	}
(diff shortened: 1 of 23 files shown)


staebler:
https://issues.redhat.com/browse/CO-782

pkg/certs/certs.go
```diff
@@ -0,0 +1,49 @@
...
+		if err != nil {
+			log.WithError(err).WithField("filename", fullFilename).Warning("Unable to read initial content of cert file")
+			continue // intentionally ignore errors
+		}
+		files[fullFilename] = fileBytes
+	}
+	obs.AddReactor(func(filename string, action fileobserver.ActionType) error {
+		log.WithField("filename", filename).Info("exiting because cert changed")
+		// TODO: Exit more cleanly, allowing for cleanup by closing contexts.
+		os.Exit(1)
```
joelddiaz: Why exit non-zero? I wouldn't expect the cert change to be considered an error condition.

staebler: Makes sense. I will change it.

pkg/certs/certs.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil
+	}, files, fullFilenames...)
+
+	stop := make(chan struct{})
+	go func() {
+		log.WithField("files", fullFilenames).Info("running file observer")
+		obs.Run(stop)
+		log.Error("file observer stopped")
```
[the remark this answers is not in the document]
staebler: Yeah, that seems like a good idea.

staebler:
> this all looks good to me. can we squash the last two commits?

I can squash the last commit. I prefer to keep the vendor commit separate in order to isolate the noise from vendor changes from the signal of Hive code changes.

dgoodwin:
Awesome, glad to see this one fixed.

PR #823, opened 2020-02-12
docs: update using-hive.md for v1, add hiveutil doc

gregsheremeta:
docs: update using-hive.md for v1, add hiveutil doc

Fixes: https://github.com/openshift/hive/issues/780
Fixes: https://issues.redhat.com/browse/CO-715

PR #837, opened 2020-02-18
hack: script for extracting the kubeconfig for a clusterdeployment

staebler:
This is something that I do often enough and that is a long enough command that I thought it would be beneficial to create a script for.

### DIFF
--- a/hack/get-kubeconfig.sh
+++ b/hack/get-kubeconfig.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+#
+# This script will print to stdout the kubeconfig for a ClusterDeployment.
+#
+
+set -e
+
+usage(){
+	echo "Usage: $0 [CLUSTERDEPLOYMENT_NAMESPACE/]CLUSTERDEPLOYMENT_NAME"
+	exit 1
+}
+
+
+# Get the name and namespace for the clusterdeployment to duplicate.
+CD_NAME_ARG=${1?missing ClusterDeployment name}
+CD_NAME_ARR=(${CD_NAME_ARG//// })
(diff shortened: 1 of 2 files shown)


hack/get-kubeconfig.sh
```diff
@@ -0,0 +1,35 @@
...
+
+usage(){
+	echo "Usage: $0 [CLUSTERDEPLOYMENT_NAMESPACE/]CLUSTERDEPLOYMENT_NAME"
+	exit 1
+}
+
+
+# Get the name and namespace for the clusterdeployment to duplicate.
+CD_NAME_ARG=${1?missing ClusterDeployment name}
+CD_NAME_ARR=(${CD_NAME_ARG//// })
```
gregsheremeta: how does this work? i'm not great at bash

gregsheremeta:
looks good to me. Hopefully #823 beats this, though :)

staebler:
> looks good to me. Hopefully #823 beats this, though :)

I'll hold this to wait for #823. This PR will be easy to rebase if needed.

/hold

hack/get-kubeconfig.sh
staebler: * `{string/substring/replacement}` replaces the first occurrence of *substring* with *replacement*
* `(string1 string2 string3)` creates an array with the elements in the list.

So, `(${CD_NAME_ARG//// })` replaces the first forward slash `/` in `CD_NAME_ARG` with a space then creates an array where the first element is everything before the new space and the second element is everything after the new space.

dgoodwin:
LGTM and will be handy.

PR #839, opened 2020-02-18
Add documentation on bare metal provisioning with Hive.

dgoodwin:


gregsheremeta:
if @staebler would like to ack #823, you'll need to rebase :)

docs/using-hive.md
```diff
@@ -15,7 +15,7 @@
         ```bash
         oc create secret generic mycluster-ssh-key --from-file=ssh-privatekey=/path/to/private/key --from-file=ssh-publickey=/path/to/public/key
         ```
-     **NOTE**: This step is optional. This will be done automatically if using `hiveutil create-cluster` with `--ssh-public-key-file` and `--ssh-private-key-file` arguments. 
+     **NOTE**: This step is optional. This will be done automatically if using `hiveutil create-cluster` with `--ssh-public-key-file` and `--ssh-private-key-file` arguments.
```
gregsheremeta: this section is gone in #823

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+      ironic_endpoint: "http://172.22.0.3:6385/v1/"
+      ironic_inspector_endpoint: "http://172.22.0.3:5050/v1/"
+      cache_url: "http://192.168.111.1/images"
+      rhcos_image_url: "https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.3/43.81.201911192044.0/x86_64/rhcos-43.81.201911192044.0-openstack.x86_64.qcow2.gz"
+kind: ConfigMap
+metadata:
+  name: my-baremetal-cluster-install-manifests
+```
+
+Create a ClusterDeployment, note the libvirtSSHPrivateKeyRef and sshKnownHosts for bare metal:
```
gregsheremeta: it's nice to use backticks around CR types and properties, so
`ClusterDeployment`, `libvirtSSHPrivateKeyRef`, etc

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+
+At present this feature requires a separate pre-existing libvirt provisioning host to run the bootstrap node. This host will require very specific network configuration that far exceeds the scope of Hive documentation. See *LINK TBD* for more information.
+
+![Bare metal provisioning with one hypervisor host per cluster](hive-baremetal-hypervisor-per-cluster.png)
+
+![Bare metal provisioning with a shared hypervisor host](hive-baremetal-shared-hypervisor.png)
+
+To provision bare metal clusters with Hive:
+
+Create a Secret containing a [bare metal enabled InstallConfig](https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md#install-config).
```
dhellmann: We could mention here that the install-config must include the libvirtURI, unless you think that's implicit based on line 164?

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+
+Create a Secret containing the SSH private key that can connect to your libvirt provisioning host, without a passphrase.
+
+```yaml
+- apiVersion: v1
```
dhellmann: The `-` at the start of this line and the indentation look suspicious.

docs/using-hive.md
dgoodwin: My editor is setup to trim trailing whitespace so it's picking this up on it's own. :( Will deal with it on merge conflict.

docs/using-hive.md
dgoodwin: That does look nice, thanks.

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
 bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=gcp mycluster
 ```
 
+#### Create Cluster on Bare Metal
+
+Hive supports bare metal provisioning as provided by [openshift-install](https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md)
+
+At present this feature requires a separate pre-existing libvirt provisioning host to run the bootstrap node. This host will require very specific network configuration that far exceeds the scope of Hive documentation. See *LINK TBD* for more information.
```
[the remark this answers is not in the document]
dgoodwin: Thanks Doug! Added to my next commit.

docs/using-hive.md
dgoodwin: Yeah probably copied out of a list, bringing it all in one level, thanks!

docs/using-hive.md
dgoodwin: Added.

dhellmann:
This looks good!

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+      provisioning_interface: "enp1s0"
+      provisioning_ip: "172.22.0.3/24"
+      dhcp_range: "172.22.0.10,172.22.0.100"
+      deploy_kernel_url: "http://172.22.0.3:6180/images/ironic-python-agent.kernel"
+      deploy_ramdisk_url: "http://172.22.0.3:6180/images/ironic-python-agent.initramfs"
+      ironic_endpoint: "http://172.22.0.3:6385/v1/"
+      ironic_inspector_endpoint: "http://172.22.0.3:5050/v1/"
+      cache_url: "http://192.168.111.1/images"
+      rhcos_image_url: "https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.3/43.81.201911192044.0/x86_64/rhcos-43.81.201911192044.0-openstack.x86_64.qcow2.gz"
+kind: ConfigMap
```
staebler: Personally, I would re-order the fields to make it a little easier to read. Use the same order that is used in the nested configmap.
```
kind:
apiVersion:
metadata:
data:
```

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+metadata:
+  name: my-baremetal-cluster-install-manifests
```
staebler: Use the `mynamespace` namespace to be consistent with the rest of the resources for the clusterdeployment.

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+
+Create a `Secret` containing a [bare metal enabled InstallConfig](https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md#install-config). This `InstallConfig` must contain a `libvirtURI` property pointing to the provisioning host.
+
+Create a `Secret` containing the SSH private key that can connect to your libvirt provisioning host, without a passphrase.
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: provisioning-host-ssh-private-key
```
staebler: Use the `mynamespace` namespace.

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+  name: my-baremetal-cluster-install-manifests
+```
+
+Create a `ClusterDeployment`, note the `libvirtSSHPrivateKeyRef` and `sshKnownHosts` for bare metal:
+
+```yaml
+apiVersion: hive.openshift.io/v1
+kind: ClusterDeployment
+metadata:
+  name: my-baremetal-cluster
```
staebler: Use the `mynamespace` namespace.

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+  annotations:
+    hive.openshift.io/try-install-once: "true"
```
staebler: Is the try-install-once annotation something necessary for a baremetal install? Or is it just a remnant of testing?

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+spec:
+  baseDomain: test.example.com
+  clusterName: my-baremetal-cluster
+  controlPlaneConfig:
+    servingCertificates: {}
+  platform:
+    bareMetal:
+      libvirtSSHPrivateKeyRef:
```
staebler: `libvirtSSHPrivateKeySecretRef`

docs/using-hive.md
```diff
@@ -157,6 +157,102 @@ Credentials will be read from `~/.gcp/osServiceAccount.json`, this can be create
...
+  provisioning:
+    installConfigSecretRef:
+      name: my-baremetal-cluster-install-config
+    sshPrivateKeySecretRef:
+      name: my-baremetal-hosts-ssh-private-key
+    manifestsConfigMapRef:
+      name: my-baremetal-cluster-install-manifests
+    imageSetRef:
+      name: my-clusterimageset
+    sshKnownHosts:
```
staebler: It may be helpful to give a little explanation for this field in this section.

docs/using-hive.md
dgoodwin: I'm unsure on this, there is no deprovision process for bare metal anywhere, so I'm not positive retry makes sense. If it goes bad it usually hangs indefinitely. I think try once is ok to recommend for now.

dgoodwin: This could go either way. It probably can't hurt to take it out. I will remove for now.

docs/using-hive.md
dgoodwin: LOL, this is why my property wasn't showing up in cluster during demo. Thx.

docs/using-hive.md
staebler: If it does not make sense to try the install more than once for bare metal, should we build that into Hive directly rather than requiring the user to add the annotation?

dgoodwin: I'm really not sure, if we bake it in and change our minds, it's a pretty dramatic change in behavior that could cause some problems. I'm going to keep the annotation for now and ask around for more info, we can followup if necessary.

dgoodwin:
Updates pushed.

dgoodwin:
Updated now that the other docs PR has merged. Sorry, needs another lgtm. :(

docs/using-hive.md
twiest: I agree with @staebler. If it's a known behavior, then we should not expect the end user to supply the annotations each time they want to install with bare metal.

dhellmann: Retrying won't require deprovisioning, but the types of failures we see mean it's not likely to work just by trying again. Someone usually has to go fix a switch or other infrastructure setting to get it to work.

I don't know what that means about the UX of making try-install-once the default for only one platform. Maybe the place to make that decision is in ACM, rather than Hive?

PR #840, opened 2020-02-18
installmanager: remove manual adding to /etc/passwd

staebler:
As of OpenShift 4.3, there is no need for the installmanager to manually manipulate the /etc/passwd file. The version of cri-o included in OpenShift 4.3 will automatically add the user to the /etc/passwd file.

Also, remove build/hive/Dockerfile which does not appear to be used anywhere.

https://issues.redhat.com/browse/CO-770

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,24 +22,14 @@ COPY --from=builder /go/src/github.com/openshift/hive/bin/hiveutil /usr/bin
 COPY --from=builder /go/src/github.com/openshift/hive/bin/hive-operator /opt/services
 COPY --from=builder /go/src/github.com/openshift/hive/bin/hive-apiserver /opt/services/
 
-# In the event of failures we run ssh commands to execute the script which gathers logs
-# from all systems in the cluster, and scp's those from the bootstrap node into the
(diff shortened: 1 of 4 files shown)


twiest:
Looks good to me. Not adding the tag so that the second reviewer can add the tag.

joelddiaz:
Do we need to worry about 4.2?

gregsheremeta:
> Do we need to worry about 4.2?

Great question @joeldiaz. SD isn't done upgrading to 4.3, and i don't know what ACM will support but I'd guess older versions. And then open source users.

staebler:
> > Do we need to worry about 4.2?
> 
> Great question @joelddiaz. SD isn't done upgrading to 4.3, and i don't know what ACM will support but I'd guess older versions. And then open source users.

This is only relevant for bare-metal installations. And even then it is only germane to the cluster running Hive and not the cluster installed by Hive.

dgoodwin:
Yes we only need to support bare metal provisioning 4.3+ (and I doubt it's even present in 4.2). This should have no impact on dedicated.

sdodson:
/retitle Bug 1793316: installmanager: remove manual adding to /etc/passwd

PR #841, opened 2020-02-19
Fix hiveapi-cluster-role-binding.yaml

berndonline:
Hello,
Small fix for hiveapi-extension-apiserver-authentication-reader rolebinding otherwise rolebinding will fail getting applied. Wrong namespace selected and kube-system was defined under roleRef but not supported option.
Please review and merge.
Thx,
Bernd


berndonline:
Didn't saw this rolebinding was being removed. Close the PR 👍

PR #843, opened 2020-02-21
ensure hive puts a domain into the apiserver object

joelddiaz:
The domain field must not include the protocol or port information from the clusterDeployment.status.APIURL field.

https://issues.redhat.com/browse/CO-799

### DIFF
--- a/pkg/controller/controlplanecerts/controlplanecerts_controller.go
+++ b/pkg/controller/controlplanecerts/controlplanecerts_controller.go
@@ -5,6 +5,7 @@ import (
 	"crypto/md5"
 	"fmt"
 	"io"
+	"net/url"
 	"sort"
 	"time"
 
@@ -33,6 +34,7 @@ import (
 	"github.com/openshift/hive/pkg/constants"
 	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	"github.com/openshift/hive/pkg/remoteclient"
 	"github.com/openshift/hive/pkg/resource"
 )
 
@@ -66,11 +68,16 @@ func Add(mgr manager.Manager) error {
 func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 	logger := log.WithField("controller", controllerName)
 	helper := resource.NewHelperWithMetricsFromRESTConfig(mgr.GetConfig(), controllerName, logger)
-	return &ReconcileControlPlaneCerts{
+	r := &ReconcileControlPlaneCerts{
 		Client:  controllerutils.NewClientWithMetricsOrDie(mgr, controllerName),
 		scheme:  mgr.GetScheme(),
 		applier: helper,
 	}
+	r.remoteClientBuilder = func(cd *hivev1.ClusterDeployment) remoteclient.Builder {
+		return remoteclient.NewBuilder(r.Client, cd, controllerName)
+	}
+
+	return r
 }
 
 // AddToManager adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -95,8 +102,9 @@ var _ reconcile.Reconciler = &ReconcileControlPlaneCerts{}
 // ReconcileControlPlaneCerts reconciles a ClusterDeployment object
 type ReconcileControlPlaneCerts struct {
 	client.Client
-	scheme  *runtime.Scheme
-	applier applier
+	scheme              *runtime.Scheme
+	applier             applier
+	remoteClientBuilder func(*hivev1.ClusterDeployment) remoteclient.Builder
 }
 
 // Reconcile reads that state of the cluster for a ClusterDeployment object and makes changes based on the state read
@@ -127,6 +135,10 @@ func (r *ReconcileControlPlaneCerts) Reconcile(request reconcile.Request) (recon
 	}
 	// If the clusterdeployment is deleted, do not reconcile.
 	if cd.DeletionTimestamp != nil {
+		return reconcile.Result{}, nil
+	}
+
+	if !cd.Spec.Installed {
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -361,6 +357,12 @@ func (r *ReconcileControlPlaneCerts) setCertsNotFoundCondition(cd *hivev1.Cluste
 	return true, r.Status().Update(context.TODO(), cd)
 }
 
+// defaultControlPlaneDomain will return the string representing the DNS entry
+// for the cluster's API endpoing
```
staebler: s/endpoing/endpoint/

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -361,6 +357,12 @@ func (r *ReconcileControlPlaneCerts) setCertsNotFoundCondition(cd *hivev1.Cluste
...
+func defaultControlPlaneDomain(cd *hivev1.ClusterDeployment) string {
+	return fmt.Sprintf("api.%s.%s", cd.Spec.ClusterName, cd.Spec.BaseDomain)
```
staebler: I do not like that this is hard-coded. This will be a nightmare if the installer changes what the API endpoint is at some point.

dgoodwin: Agreed, the Status value was extracted from kubeconfig IIRC and should be resilient to installer changes, more commit details would help here.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -285,9 +286,13 @@ func (r *ReconcileControlPlaneCerts) generateControlPlaneCertsSyncSet(cd *hivev1
 		if cd.Status.APIURL == "" {
 			return nil, errors.New("API URL not set in ClusterDeployment")
 		}
+		apidomain, err := defaultControlPlaneDomain(cd.Status.APIURL)
```
staebler: The `cd.Status.APIURL` is not necessarily the API URL from the kubeconfig. It could be the override API URL. Although, I think that it is a deficiency that it could be the override API URL. Since https://github.com/openshift/hive/pull/777, the `cd.Status.APIURL` could be set after the clusterdeployment has been set to installed.

staebler: In any event, I think we want to grab the API URL directly from the kubeconfig rather than relying on the field set in the clusterdeployment status.
```
remoteclient.NewBuilder(...).UseSecondaryAPIURL().APIURL()
```

pkg/controller/controlplanecerts/controlplanecerts_controller.go
joelddiaz: now using url parsing to pull out the hostname/domain

joelddiaz:
@staebler should this controller bother to reconcile before the clusterDeployment has reached the "install completed" state?

joelddiaz:
@staebler i went ahead and started using the remoteclient builder to fetch the URL as suggested

i also put in the check early in the controller to not bother reconciling a ClusterDeployment until it is marked 'installed' WDYT?

pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
```diff
@@ -310,6 +357,15 @@ func (f *fakeClusterDeploymentWrapper) defaultCert(name, secretName string) *fak
 	return f
 }
 
+func (f *fakeClusterDeploymentWrapper) adminKubeconfig() *fakeClusterDeploymentWrapper {
+	f.cd.Spec.ClusterMetadata = &hivev1.ClusterMetadata{
```
staebler: May as well make this the default for the ClusterDeployment. You are already setting the ClusterDeployment to installed by default.

Athough, my preference would be that you use a mock remoteclientbuilder rather than writing a kubeconfig secret.
1. This will allow you to more easily test other unhappy scenarios such as a bad URL or errors from the remoteclientbuilder.
2. This will keep the code similar to the other controllers that use the mock remoteclientbuilder in their unit tests.

pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
```diff
@@ -124,21 +114,23 @@ func TestReconcileControlPlaneCerts(t *testing.T) {
...
 					namedCert("cert2", "bar.com", "secret2").
-					adminKubeconfig().
 					obj(),
 				fakeCertSecret("secret0"),
 				fakeCertSecret("secret1"),
 				fakeCertSecret("secret2"),
-				testAdminKubeconfigSecret(),
+			},
+			setupRemoteClient: func(mockController *gomock.Controller) *mockremoteclient.MockBuilder {
+				mockRemoteClient := mockremoteclient.NewMockBuilder(mockController)
```
staebler: Move this code into the test proper. Use a boolean in the test case to determine whether the calls to the mock remote client builder are expected. Or use a string for the API URL to return from the mock call as a determiner of whether the calls are expected.

pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
```diff
@@ -81,18 +68,21 @@ func TestReconcileControlPlaneCerts(t *testing.T) {
...
 			},
 			validate: func(t *testing.T, c client.Client, applied []runtime.Object) {
 				cd := getFakeClusterDeployment(t, c)
 				assert.Empty(t, cd.Status.Conditions, "no conditions should be set")
 
-				expectedAPIDomain, err := defaultControlPlaneDomain(cd, c)
-				if err != nil {
-					t.Fatalf("unexpectedly failed to decode API URL for cluster deployment: %v", err)
-				}
+				expectedAPIDomain := fakeAPIURLDomain
```
staebler: Remove this variable. Pass `fakeAPIURLDomain` directly to `additionalCert`.

pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
```diff
@@ -124,21 +114,23 @@ func TestReconcileControlPlaneCerts(t *testing.T) {
...
```
staebler: Remove this variable.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -97,8 +99,9 @@ var _ reconcile.Reconciler = &ReconcileControlPlaneCerts{}
 // ReconcileControlPlaneCerts reconciles a ClusterDeployment object
 type ReconcileControlPlaneCerts struct {
 	client.Client
-	scheme  *runtime.Scheme
-	applier applier
+	scheme              *runtime.Scheme
+	applier             applier
+	remoteClientBuilder func(client.Client, *hivev1.ClusterDeployment) remoteclient.Builder
```
staebler: My personal preference would be to remove the `client.Client` parameter from this function. That will match what is done in the other controllers. If you feel strongly about keeping it this way, then I can live with it, though,

pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
joelddiaz: i thought this approach (of allowing per-test-case remote client setup) would allow more flexibility if/when we want tests to cover error paths. although, admittedly, right now it isn't being used

joelddiaz: i'll change it so that the test case contains a remoteClientResponse struct to allow defining the url returned and any possible error

joelddiaz:
@staebler PTAL the most recent commit addressing your feedback. then i can squash.

staebler:
> @staebler PTAL the most recent commit addressing your feedback. then i can squash.

@joelddiaz I am not seeing any recent commits addressing my feedback.

joelddiaz:
> @joelddiaz I am not seeing any recent commits addressing my feedback.
It certainly helps if I actually push the commit :)

PR #844, opened 2020-02-23
migration: changes to support hive-int

staebler:
* Fix differences seen in namespace when comparing cluster-scoped resources.
* Ignore not-found errors when trying to restore owner references on DNSEndpoints and install-log PVCs.
* Fix delete_jobs script function.
* Do not do a dry-run delete of jobs as that is not supported with oc 3.11.
* Make a separate step for removing the validating webhooks. The hiveadmission will not work once the Hive CSV is deleted. So the validating webhooks need to be deleted prior to run the delete.sh script.
* Make the scaling scripts more granular. Deleting the Hive CSV will delete the hive-operator deployment.
* Accept a missing dpeloyment as an acceptably scaled-down deployment.

### DIFF
--- a/contrib/pkg/v1migration/restoreownerrefs.go
+++ b/contrib/pkg/v1migration/restoreownerrefs.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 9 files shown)


contrib/pkg/v1migration/restoreownerrefs.go
```diff
@@ -164,3 +171,12 @@ func isNamespaceScoped(kind string) bool {
 		return true
 	}
 }
+
+func isRefFromPVCToCD(ref ownerRef) bool {
+	return ref.Resource == "persistentvolumeclaims" && ref.Group == "" &&
+		ref.Ref.Kind == "ClusterDeployment" && ref.Group == hivev1alpha1.SchemeGroupVersion.String()
```
dgoodwin: Requires ref.Group == "" && ref.Group == hivev1alpha1, looks like a typo here, guessing the latter one should be ref.Ref.Group == hivev1alpha1

hack/v1migration/README.md
```diff
@@ -2,13 +2,17 @@
 
   1. Deploy master (v1alpha1) to a cluster.
   1. Create a cluster deployment and wait for completion.
-  1. Run `./hack/v1migration/scaledown.sh` to scale down Hive deployments.
+  1. Run `oc patch hiveconfig hive --type='merge' -p $'spec:\n maintenanceMode: true'`
```
dgoodwin: Nice.

contrib/pkg/v1migration/restoreownerrefs.go
staebler: Yes. Good catch. Looking through the owner-refs.yaml file, I don't see any refs from PVCs captured. So I need to look into why that was missed.

Changed to `ref.Ref.APIVersion == hivev1alpha1.SchemeGroupVersion.String()`.

Also, in `isRefFromDNSEndpoint`, the check for the group was wrong, since the group does not contain the version. I changed that to `ref.Group == hivev1alpha1.GroupName`.

PR #850, opened 2020-02-25
Updates install dependencies.

jtudelag:
yq is needed.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -5,6 +5,7 @@
 * [kustomize](https://github.com/kubernetes-sigs/kustomize#kustomize)
 * [oc](https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/)
 * [go (version 1.12)](https://github.com/golang/go)
+* [yq](https://github.com/mikefarah/yq)
 
 ## Deployment Options
 


dgoodwin:
Is this because of it's use in hack/olm-registry-deploy.sh or somewhere else? I can't find much. I also don't have it installed. :)

jtudelag:
@dgoodwin , yes, here: https://github.com/openshift/hive/blob/master/hack/olm-registry-deploy.sh#L41

dgoodwin:
Can you clarify why you are using this script? This is used for our pipelines that deploy to dedicated, definitely not something I would expect a normal user or developer of hive to need to worry about.

jtudelag:
@dgoodwin just wanted to deploy hive to play with it. Is there any easier way to do it?

dgoodwin:
Today we use the methods referenced in the documentation, the OLM bundle building is more complex and only used in our pipelines right now, though we do plan to package as an OLM community operator and have work underway for that.

Recommended way today is to use the make deploy targets.

jtudelag:
Thanks. Well, we can close this if you want then.

PR #854, opened 2020-02-26
support aws china for managed dns

staebler:
Use a custom endpoint for route53 when the region is a "cn-" region.

A region field has been added to .spec.managedDomains[].aws in HiveConfig. When the managed domain is handled by AWS China, the region field should be set to "cn-northwest-1" so that the route53 requests are sent to AWS China instead of regular AWS.

https://issues.redhat.com/browse/CO-769

### DIFF
--- a/pkg/apis/hive/v1/dnszone_types.go
+++ b/pkg/apis/hive/v1/dnszone_types.go
@@ -45,6 +45,12 @@ type AWSDNSZoneSpec struct {
 	// to these tags,the DNS Zone controller will set a hive.openhsift.io/hostedzone tag
 	// identifying the HostedZone record that it belongs to.
 	AdditionalTags []AWSResourceTag `json:"additionalTags,omitempty"`
+
+	// Region is the AWS region to use for route53 operations.
+	// This defaults to us-east-1.
+	// For AWS China, use cn-northwest-1.
+	// +optional
+	Region string `json:"region,omitempty"`
 }
 
 // AWSResourceTag represents a tag that is applied to an AWS cloud resource
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -27,6 +28,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/route53/route53iface"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+
+	"github.com/openshift/hive/pkg/constants"
 )
 
 const (
@@ -52,7 +55,7 @@ func init() {
 
 // Client is a wrapper object for actual AWS SDK clients to allow for easier testing.
 type Client interface {
-	//EC2
+	// EC2
 	DescribeAvailabilityZones(*ec2.DescribeAvailabilityZonesInput) (*ec2.DescribeAvailabilityZonesOutput, error)
 	DescribeImages(*ec2.DescribeImagesInput) (*ec2.DescribeImagesOutput, error)
 	DescribeVpcs(*ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error)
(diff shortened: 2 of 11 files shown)


staebler:
cc @dofinn

pkg/awsclient/client.go
```diff
@@ -339,3 +345,14 @@ func NewClientFromSecret(secret *corev1.Secret, region string) (Client, error) {
...
 	}, nil
 }
+
+func awsChinaEndpointResolver(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
+	if service != route53.EndpointsID || region != constants.AWSChinaRoute53Region {
+		return endpoints.DefaultResolver().EndpointFor(service, region, optFns...)
+	}
+
+	return endpoints.ResolvedEndpoint{
+		URL:         "https://api.route53.cn",
```
staebler: Should this be `https://route53.amazonaws.com.cn`?

wanghaoran1988: @staebler Yes, we should use https://route53.amazonaws.com.cn, this is from AWS SA:

```
For the China Route53, there are 2 endpoints,  see below

    1  api.route53.cn

    2 route53.amazonaws.com.cn

Based on the info from service team , both 2 endpoints are workable after GA. But the new endpoint(the 2nd ) is recommended.
```

dofinn:
> cc @dofinn

/lgtm :)

pkg/awsclient/client.go
dgoodwin: Waiting on this change but otherwise LGTM.

staebler: I made this change and verified that everything is still working.

pkg/apis/hive/v1/dnszone_types.go
```diff
@@ -45,6 +45,12 @@ type AWSDNSZoneSpec struct {
 	// to these tags,the DNS Zone controller will set a hive.openhsift.io/hostedzone tag
 	// identifying the HostedZone record that it belongs to.
 	AdditionalTags []AWSResourceTag `json:"additionalTags,omitempty"`
+
+	// Region is the AWS region to use for route53 operations.
+	// This defaults to us-east-1.
+	// For AWS China, use cn-northwest-1.
```
joelddiaz: Is it is a hard requirement for this field to be 'cn-northwest-1' when wanting to interact with AWS China? It appears that putting in 'cn-north-1' would also result in using the alternative API endpoint (with the region overridden to use 'cn-northwest-1' for the created AWS client).

staebler: That was an open question that I had that I have not gotten an answer for yet. I don't have the capability to really test out whether it must be `cn-northwest-1`, can be `cn-northwest-1`, or must be `cn-north-1` when the cluster is being installed into `cn-north-1`. If I understood, @dofinn correctly, he indicated to me that it can be `northwest-1`

joelddiaz: I can confirm that the client requires it to be set to `cn-northwest-1`. I'm just observing that if you put `cn-north-1` in the DNSZone, the client gets created with a region set to `cn-northwest-1` b/c the DNSZone region has the magic `cn-` prefix.

dofinn: AWS China tends to have a default region endpoint depending on the service. Unfortunately, its best to figure this out using trial and error at the moment. `aws --debug route53 --endpoint-url https://api.route53.cn list-hosted-zones` is very helpful.

staebler: So, @joelddiaz, I am a bit confused. Are you suggesting that there are changes that need to be made to the code? The way I see the code working now is that the clusterdeployment controller always sets the region in the DNSZone to `cn-northwest-1` if the region of the ClusterDeployment is a China region. If I understand your second comment correctly, that is what we want to happen since the region used **must** be `cn-northwest-1` when making route53 requests. If a user creates their own DNSZone, then they can set the region to whatever they like and that will be used to make the route53 requests--for better or worse.

wanghaoran1988: We should use "cn-northwest-1", just get confirmation from AWS China SA, this service is hosted in "cn-northwest-1" region, here is some test I did"

```
Failed when using cn-north-1:
aws route53  list-hosted-zones  --region cn-north-1 --endpoint-url https://route53.amazonaws.com.cn 
aws route53   list-tags-for-resources  --region cn-north-1 --endpoint-url https://route53.amazonaws.com.cn --resource-id <id> --resource-type hostedzone
Succeed with cn-northwest-1:
aws route53  list-hosted-zones  --region cn-northwest-1 --endpoint-url https://route53.amazonaws.com.cn --> succeed
aws route53   list-tags-for-resources  --region cn-northwest-1 --endpoint-url https://route53.amazonaws.com.cn --resource-id <id> --resource-type hostedzone
```

joelddiaz: I'm sorry, I misread where the `cn-` prefix checking is going on. I see now that the clusterdeployment controller will put in `cn-northwest-1` for any clusterDeployment with the `cn-` region.

PR #858, opened 2020-02-27
Add additional CAs to admin kubeconfig even when status URLs set.

dgoodwin:
We discovered that in situations where the status URLs get set, but at a time when the additional CAs were not configured properly, the cluster goes into a state where it's admin kubeconfig will never be updated. Because it lacks the correct CAs, the cluster is permanently marked unreachable and hive cannot communicate with it.

This fix adjusts the admin kubeconfig reconciliation such that it always runs even if status URLs are already set. The postProcessAdminKubeconfig method was broken down as it mixed implicit updates with updates the caller was expected to do, and we require different assumptions regarding the status URLs. Methods renamed to be more clear as to what they're actually doing.

Added two tests which only check that we copied the kubeconfig data to the raw field, indicating that the additional CAs would be added if any were configured, when the status URLs are set and not set. There did not appear to be any coverage of this functionality beforehand.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -284,7 +284,7 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (resul
 	return r.reconcile(request, cd, cdLog)
 }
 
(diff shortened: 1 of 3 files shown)


staebler:
This looks good for resolving the issue on hive-int.

Long term, the controller should only have one place where it adjusts the kubeconfig. Also, the controller should have the same behavior for the console and API URLs, where the controller enforces that the values are correct in the status and not just present.

/lgtm

PR #861, opened 2020-02-27
add support for machinepools on azure

staebler:
* Add an Azure actuator for the remotemachinesets controller.
* Add an Azure client wrapper for making requests to Azure. The only request this supports right now is listing resource SKUs, as this is all that is needed for machineset support.
* Turn on Azure MachinePool e2e tests.

https://issues.redhat.com/browse/CO-574


staebler:
53eeb975...30dec0ad
* Increased the time to wait for the expected machines when running the e2e machinesets tests. The time was increased from 5 minutes to 10 minutes. It is taking a bit over 5 minutes on Azure to delete a machine.

pkg/controller/remotemachineset/azureactuator.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		computePool.Platform.Azure.Zones = zones
+	}
+
+	// The imageID parameter is not used. The image is determined by the infraID.
+	const imageID = ""
+	// The role must be "worker" so that the machines get hooked up to the existing subnet.
+	const role = "worker"
+
+	installerMachineSets, err := installazure.MachineSets(cd.Spec.ClusterMetadata.InfraID, ic, computePool, imageID, role, "worker-user-data")
```
dgoodwin: worker-user-data should probably be treated the same as the other const's just above.

pkg/controller/remotemachineset/azureactuator_test.go
```diff
@@ -0,0 +1,168 @@
...
+			expectedMachineSetReplicas: map[string]int64{
+				generateAzureMachineSetName("zone1"): 3,
+			},
+		},
+		{
+			name:              "generate machinesets across zones",
+			clusterDeployment: testAzureClusterDeployment(),
+			pool:              testAzurePool(),
+			mockAzureClient: func(mockCtrl *gomock.Controller, client *mockazure.MockClient) {
+				mockListResourceSKUs(mockCtrl, client, []string{"zone1", "zone2", "zone3"})
```
dgoodwin: Could you add or expand this to have more zones than needed and assert the latter ones are set to 0.

staebler:
30dec0ad...a547a581
* Made a const for user data so that it can have a comment describing why it is hard-coded to "worker-user-data". Resolves https://github.com/openshift/hive/pull/861#discussion_r385885799.
* Added TestAzureActuator/more_zones_than_replicas and TestAzureActuator/list_zones_returns_zero test cases. Resolves https://github.com/openshift/hive/pull/861#discussion_r385887189

staebler:
Auto-scaling may not be working correctly. I need to check this.

```
 --- FAIL: TestAutoscalingMachinePool (1639.00s)
    infra_test.go:254: waiting for Hive to create cluster autoscaler
    infra_test.go:323: timed out waiting for machines to be created
    infra_test.go:326: timed out waiting for nodes to be created
FAIL
FAIL	github.com/openshift/hive/test/e2e/postinstall/machinesets	2589.467s 
```

staebler:
002da1ea...6c003e27
* Changed the busybox deployment in `TestAutoscalingMachinePool` to use a CPU quantity of 1 instead of 2. The pods were not being schedule on Azure because the machines were not large enough to accommodate the CPU request--even though the nodes claimed to have a capacity of 2.

staebler:
The e2e tests on azure are passing now.

PR #862, opened 2020-02-28
WIP: Cluster Leasing

dgoodwin:
Prototype of an implementation for maintaining a hot pool of clusters, and allowing users to request access to one and be granted it immediately, rather than waiting for a full provisioning.

Introduces ClusterPool CRD. Represents a cloud provider and region, and the number of clusters we'd like to keep "hot" in the pool. Controller will ensure we have them created, or if we have too many, scale down. In the future I'd like to implement overrides for the desired size based on day of week, and time of day. CP's can specify the default "DeleteAfter" annotation to be set on the cluster.

Example:

```yaml
kind: ClusterPool
metadata:
  name: aws-us-east-1
  namespace: hive
spec:
  platform:
    aws:
      credentialsSecretRef:
        name: dgoodwin1-aws-creds
      region: us-east-1
  size: 2
  baseDomain: new-installer.openshift.com
  deleteAfter: 4h
```

This re-uses the hiveutil create-cluster code which already does most of what we need. (generating an InstallConfig, etc)

## TODOs

- [ ] move to a v1alpha1 in a separate temporary apigroup
- [ ] reap namespaces that have our label but no longer have a cluster deployment
- [ ] implement expectations for the pool's cluster deployment creations
- [ ] examine atomicity of the create process, namespace, copy secret, make cluster...
- [x] cluster pool unit tests
- [ ] cluster pool validation
- [ ] implement release image lookup (optional, maybe in future? point to specific cluster imageset for now)
- [ ] determine if we should be re-using platform in the lease pool and assuming cred secrets are in the hive namespace
- [ ] set a maximum number of clusters *from* the pool to prevent users grabbing too many, set condition on ClusterPool when in this state


dgoodwin:
Some required hiveutil work for this broken out into #880

PR #866, opened 2020-03-03
Set region label on Clusterdeployment CR

dofinn:
This PR mimics the logic behind setting the platformLabel in the
Clusterdeployment CR. Previously present strings were also moved to
const's in the same package.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -66,7 +66,7 @@ const (
 
 	platformAWS       = "aws"
 	platformAzure     = "azure"
-	platformGCP       = "hcp"
+	platformGCP       = "gcp"
 	platformBaremetal = "baremetal"
 	platformUnknown   = "unknown"
 	regionUnknown     = "unknown"
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -63,6 +63,13 @@ const (
 
 	deleteAfterAnnotation    = "hive.openshift.io/delete-after" // contains a duration after which the cluster should be cleaned up.
 	tryInstallOnceAnnotation = "hive.openshift.io/try-install-once"
+
+	platformAWS       = "aws"
+	platformAzure     = "azure"
+	platformGCP       = "GCP"
+	platformBaremetal = "baremetal"
+	platformUnknown   = "unknown platform"
+	regionUnknown     = "unknown region"
 )
 
 var (
@@ -340,6 +347,23 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to set cluster platform label")
 		}
 		return reconcile.Result{}, err
+	}
+
+	// Set region label on the ClusterDeployment
+	if region := getClusterRegion(cd); cd.Spec.Platform.BareMetal == nil && cd.Labels[hivev1.HiveClusterRegionLabel] != region {
+		if cd.Labels == nil {
+			cd.Labels = make(map[string]string)
+		}
+		if cd.Labels[hivev1.HiveClusterRegionLabel] != "" {
+			cdLog.Warnf("changing the value of %s from %s to %s", hivev1.HiveClusterRegionLabel,
+				cd.Labels[hivev1.HiveClusterRegionLabel], region)
+		}
+		cd.Labels[hivev1.HiveClusterRegionLabel] = region
+		err := r.Update(context.TODO(), cd)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to set cluster region label")
+		}
+		return reconcile.Result{}, nil
 	}
 
 	if cd.DeletionTimestamp != nil {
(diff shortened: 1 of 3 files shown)


dofinn:
We should look to have `getClusterPlatform` and `getClusterRegion` return errors in future instead of "unknown" strings.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -342,6 +349,23 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, err
 	}
 
+	// Set region label on the ClusterDeployment
+	if region := getClusterRegion(cd); cd.Spec.Platform.BareMetal == nil && cd.Labels[hivev1.HiveClusterRegionLabel] != region {
```
dgoodwin: I don't think we need to check BareMetal is nil, just getClusterRegion, see if the label is not what we expect. If expected is "unknown" that's fine.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -342,6 +349,23 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+		if cd.Labels[hivev1.HiveClusterRegionLabel] != "" {
+			cdLog.Warnf("changing the value of %s from %s to %s", hivev1.HiveClusterRegionLabel,
+				cd.Labels[hivev1.HiveClusterRegionLabel], region)
+		}
+		cd.Labels[hivev1.HiveClusterRegionLabel] = region
+		err := r.Update(context.TODO(), cd)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to set cluster region label")
+		}
+		return reconcile.Result{}, nil
```
dgoodwin: Your update error is getting ignored, needs to be returned.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -342,6 +349,23 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+			cd.Labels = make(map[string]string)
+		}
+		if cd.Labels[hivev1.HiveClusterRegionLabel] != "" {
+			cdLog.Warnf("changing the value of %s from %s to %s", hivev1.HiveClusterRegionLabel,
+				cd.Labels[hivev1.HiveClusterRegionLabel], region)
+		}
+		cd.Labels[hivev1.HiveClusterRegionLabel] = region
+		err := r.Update(context.TODO(), cd)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to set cluster region label")
```
dgoodwin: Not familiar with this LogLevel(err) syntax, but in Hive codebase this should probably be cdLog.WithError(err).Error("failed to set cluster region label").

staebler: No, using `controllerutils.LogLevel(err)` is correct here. This logs update errors that are not controller errors--such as conflict errors--at the info level instead of the error level.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: I disagree. I do not think that we should be adding a region label for platforms where region is not a concept.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dofinn: good catch, i changed this while debugging a test.

dofinn:
> These changes look mostly fine. Can you give some more context for why this label is needed?

This is to enable per region deployments for SelectorSyncSets. For example, if we do not wish to deploy splunk-forwarder-operator in AWS China regions (cn-north-1 || cn-northwest-1)

staebler:
> > These changes look mostly fine. Can you give some more context for why this label is needed?
> 
> This is to enable per region deployments for SelectorSyncSets. For example, if we do not wish to deploy splunk-forwarder-operator in AWS China regions (cn-north-1 || cn-northwest-1)

We litigated this for the platform label and agreed.that the platform label is generally useful enough to have Hive apply it. I am not entirely convinced here. It seems like what you want is a label that indicates whether the cluster is deployed in AWS China or not. Is the region useful outside of the peculiarities of AWS China?

dofinn:
> > > These changes look mostly fine. Can you give some more context for why this label is needed?
> > 
> > 
> > This is to enable per region deployments for SelectorSyncSets. For example, if we do not wish to deploy splunk-forwarder-operator in AWS China regions (cn-north-1 || cn-northwest-1)
> 
> We litigated this for the platform label and agreed.that the platform label is generally useful enough to have Hive apply it. I am not entirely convinced here. It seems like what you want is a label that indicates whether the cluster is deployed in AWS China or not. Is the region useful outside of the peculiarities of AWS China?

You are correct that the driver behind this was to enable the determination of AWS China regions. But this also enables deployment configurations to be platform & region selective which may be used for canary deployments and management of deployments that rely on X services in X region. Some regions may not have the same services available as others per cloud provider. The choice of region is more cloud agnostic vs label of "aws-china"

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1052,6 +1052,19 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
 		},
+		{
+			name: "Add cluster region label",
+			existing: []runtime.Object{
+				testClusterDeploymentWithoutRegionLabel(),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				if assert.NotNil(t, cd, "missing clusterdeployment") {
+					//					assert.Equal(t, getClusterRegion(cd), cd.Labels[hivev1.HiveClusterRegionLabel], "incorrect cluster region label")
```
staebler: Remove this line.

dofinn: apologies, resolved as per latest push.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -63,6 +63,13 @@ const (
 
 	deleteAfterAnnotation    = "hive.openshift.io/delete-after" // contains a duration after which the cluster should be cleaned up.
 	tryInstallOnceAnnotation = "hive.openshift.io/try-install-once"
+
+	platformAWS       = "aws"
+	platformAzure     = "azure"
+	platformGCP       = "hcp"
```
lwan-wanglin: Is "hcp" correct? Whether this value is "gcp"?

dofinn: yes, resolved as per last push.

PR #872, opened 2020-03-04
migration: add certificaterequests to owner refs to save

staebler:
On the hive-int cluster, there are certificaterequests with owner references to ClusterDeployments. We need to save and restore those owner references during migration.


staebler:
This is an erroneous failure.

```
golangci-lint run -c ./golangci.yml ./pkg/... ./cmd/... ./contrib/...
pkg/controller/dnszone/test_helpers.go:74:2: var `validGCPSecret` is unused (unused)
```

PR #876, opened 2020-03-06
migration: remove initial fetch from delete process

staebler:
* Use a Patch instead of an Update to remove finalizers. This eliminates the need to do the initial Get of the resource from the cluster.
* Eliminate the pre-conditions check on the Delete, in case it takes some extra time to validate the pre-conditions.
* Remove the deleting of jobs from the delete.sh script as that has been moved to the delete_hive_jobs.sh script, run separately.

### DIFF
--- a/hack/v1migration/delete.sh
+++ b/hack/v1migration/delete.sh
@@ -17,20 +17,6 @@ then
   exit 1
 fi
 
-delete_jobs() {
-  local labels=${1:?must specify a label for the jobs to delete}
-  oc get jobs -l "${labels}" --all-namespaces -o json | \
-    jq -r '.items[]?.metadata.namespace' | \
-    sort -u | \
-    xargs -i oc delete jobs -n {} -l "${labels}"
-}
-
-echo "Deleting install jobs"
-delete_jobs "hive.openshift.io/install=true"
-
-echo "Deleting imageset jobs"
-delete_jobs "hive.openshift.io/imageset=true"
-
 # shellcheck source=hivetypes.sh
 source "$(dirname "$0")/hivetypes.sh"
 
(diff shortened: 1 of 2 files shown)


hack/v1migration/delete.sh
```diff
@@ -17,20 +17,6 @@ then
...
-    sort -u | \
-    xargs -i oc delete jobs -n {} -l "${labels}"
-}
-
-echo "Deleting install jobs"
-delete_jobs "hive.openshift.io/install=true"
-
-echo "Deleting imageset jobs"
-delete_jobs "hive.openshift.io/imageset=true"
-
```
dgoodwin: Worst case I can think of is provision jobs, recently launched, haven't yet written an infra ID back. We leave the job running, we do our save to disk without an infra ID, it writes it shortly thereafter. We restore without an infra ID, the job will fail at some point as it can't write back status, we replace with a new install job but can't cleanup the successful live cluster.

Pretty fringe concern, what do we think?

staebler: 1) The install job sets the infra ID on the ClusterProvision. The install job then waits for the ClusterProvision to transition to the provisioning stage. The install job will not create any cloud resources until the ClusterProvision has made that transition. The transition is done by the controllers after the infra ID has been saved to the ClusterDeployment. So as long as the hive controllers are not running (which the script already verifies), then there is no concern of leaking cloud resources.
2) The jobs should already have been deleted in a previous step of the migration process. If the jobs came back, then there are other problems.

staebler: So, to be clear, this change does not remove the step of deleting the Hive jobs from the migration process. This change removes the redundant step that was incorrectly left in from https://github.com/openshift/hive/pull/863/commits/ef523170f85b610a6a3285e64467918a41fe828d.

dgoodwin: Ah I see now, thanks.

dgoodwin:
This is an LGTM for me.

PR #877, opened 2020-03-06
Add HiveConfig.Spec.DeprovisionsDisabled.

dgoodwin:
Adding for safety net in scenarios where you'd like to make sure no
deprovision jobs can accidentally trigger.

### DIFF
--- a/pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
+++ b/pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
@@ -3,6 +3,8 @@ package clusterdeprovision
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
```diff
@@ -139,6 +160,12 @@ func (r *ReconcileClusterDeprovision) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, nil
 	}
 
+	// Check if deprovisions are currently disabled: (originates in HiveConfig in real world)
+	if r.deprovisionsDisabled {
+		rLog.Warn("deprovisions are currently disabled in HiveConfig, skipping")
+		return reconcile.Result{}, nil
+	}
```
twiest: Should we move this under the ClusterDeployment deletion timestamp check so that we get those log messages even when deprovisioning is disabled?

So right above these lines:
```golang
	// Generate an uninstall job
	rLog.Debug("generating uninstall job")
```

I'm just thinking that these might be a good warning / error messages to seee if something mangles the data during the migration.

pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
```diff
@@ -196,8 +207,9 @@ func TestClusterDeprovisionReconcile(t *testing.T) {
 func testClusterDeprovision() *hivev1.ClusterDeprovision {
 	return &hivev1.ClusterDeprovision{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      testName,
-			Namespace: testNamespace,
+			Name:        testName,
+			Namespace:   testNamespace,
+			Annotations: map[string]string{},
```
twiest: Just curious, why was this change added?  Were we missing Annotations before?

I don't see you using them anywhere else in this PR.

dgoodwin: Ah I had a test for the approval annotation prior which was removed, this is still in. I'll kill it.

pkg/controller/clusterdeprovision/clusterdeprovision_controller.go
dgoodwin: Sounds good to me, updated.

twiest:
+1 from me. Not adding the tag so someone else can review. LMK if you want me to just add the tag and I will.

PR #879, opened 2020-03-06
migration: process for fixing bad restores mid-migration

staebler:
Provide tooling for fixing resources that did not restore properly during a migration.

The process is the following.
1. Create restore data for only the namespace with problems so that fixes can be made in isolation to the resources in that namespace.
1. Clear out Hive resources from the problem namespace.
1. Restore the isolated data for that namespace with whatever changes needed to be made.

### DIFF
--- a/hack/v1migration/README.md
+++ b/hack/v1migration/README.md
@@ -16,6 +16,7 @@
   1. Run `oc patch hiveconfig hive --type='merge' -p $'spec:\n maintenanceMode: true'`
   1. Run `./hack/v1migration/scaledown.sh hive-operator` to scale down Hive operator.
   1. Run `./hack/v1migration/restore.sh [workdir]` to restore all data from disk.
+  1. Run `./hack/v1migration/validate.sh [workdir]` to validate that everything was restored and is not deleted.
   1. Run `./hack/v1migration/scaleup.sh hive-operator` to scale up Hive operator.
   1. Run `oc patch hiveconfig hive --type='merge' -p $'spec:\n maintenanceMode: false'`
 
--- a/hack/v1migration/README.md
+++ b/hack/v1migration/README.md
@@ -18,3 +18,15 @@
   1. Run `./hack/v1migration/restore.sh [workdir]` to restore all data from disk.
   1. Run `./hack/v1migration/scaleup.sh hive-operator` to scale up Hive operator.
   1. Run `oc patch hiveconfig hive --type='merge' -p $'spec:\n maintenanceMode: false'`
+
+
+# Resolving Problems during Restore
+
+If there was a problem during restore, then we should delete the resources in that namesapce
+that were restored, fix the data being restored, and run the restore again for just that namespace.
+
+  1. `./hack/v1migration/hive_resources_in_namespace.sh [workdir] [namespace] > [namespace_workdir]/resources.json`
+  1. `./hack/v1migration/hive_owner_refs_in_namespace.sh [workdir] [namespace] > [namespace_workdir]/owner-ref.json`
+  1. `./hack/v1migration/delete_in_namespace.sh [namespace]`
+  1. Make whatever changes are needed to resources saved in `[namespace_workdir]`.
+  1. `./hack/v1migration/restore.sh [namespace_workdir]`
--- a/hack/v1migration/delete_in_namespace.sh
+++ b/hack/v1migration/delete_in_namespace.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-NAMESPACE=${1:?must specificy namespace}
+NAMESPACE=${1:?must specify namespace}
 echo "Using namespace: $NAMESPACE"
 
 # shellcheck source=hivetypes.sh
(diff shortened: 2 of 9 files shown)


hack/v1migration/README.md
```diff
@@ -18,3 +18,15 @@
...
   1. Run `oc patch hiveconfig hive --type='merge' -p $'spec:\n maintenanceMode: false'`
+
+
+# Resolving Problems during Restore
+
+If there was a problem during restore, then we should delete the resources in that namesapce
+that were restored, fix the data being restored, and run the restore again for just that namespace.
+
+  1. `./hack/v1migration/hive_resources_in_namespace.sh [workdir] [namespace] > [namespace_workdir]/resources.json`
+  1. `./hack/v1migration/hive_owner_refs_in_namespace.sh [workdir] [namespace] > [namespace_workdir]/owner-ref.json`
```
dgoodwin: If we hit something like a validation error during restore, no resource created at all, we'd want to manually setup a restore dir correct? I almost wonder if we want a script that scans the global backup dir and saved out the per namespace resources, rather than pulling them from the namespace we failed to restore.

staebler:
Added a validation script to be run before restarted the Hive controllers. The validation ensures (1) that all of the resources that were saved during migration have been restored and (2) that no Hive resources have deletion timestamps.

staebler:
Updated with a fix to the delete_in_namespace.sh script to also delete machinepools and installconfig secrets.

/hold cancel

dgoodwin:
Looks good to me with limited bash experience.

@twiest leaving to you for lgtm.

hack/v1migration/hive_resources_in_namespace.sh
```diff
@@ -0,0 +1,18 @@
...
+if [[ ! -d "$WORKDIR" ]]
+then
+  echo "Directory $WORKDIR does not exist."
+  exit 1
+fi
+
+for f in "${WORKDIR}"/*.json
+do
+  if [[ "$f" =~ (hiveconfigs)|(dnsendpoints)|(owner-refs).json ]]; then continue; fi
+  <"$f" jq --arg NAMESPACE "$NAMESPACE" 'select(.metadata.namespace==$NAMESPACE)'
```
twiest: Same question as above about using redirection instead of filename at the end.

hack/v1migration/validate.sh
```diff
@@ -0,0 +1,42 @@
...
+
+TEMPDIR="$(mktemp -d)"
+
+for t in "${HIVE_TYPES[@]}"
+do
+  if [[ "$t" == "dnsendpoints" ]]; then continue; fi
+
+  echo "Validating ${t}"
+
+  <"${WORKDIR}/$t.json" jq '.metadata.namespace + "/" + .metadata.name' | sort > "${TEMPDIR}/$t.orig"
```
twiest: Same question as above about redirecting instead of using filename at the end of the command.

hack/v1migration/validate.sh
```diff
@@ -0,0 +1,42 @@
...
+  echo "Diff in resources"
+  diff "${TEMPDIR}/$t.orig" "${TEMPDIR}/$t.restored"
+
+  echo "Deleted resources"
+  oc get "$t.v1alpha1.hive.openshift.io" --all-namespaces -o json | jq '.items[] | select(.metadata.deletionTimestamp) | .metadata.namespace + "/" + .metadata.name'
+done
+
+echo "Validating machinepools"
+
+<"${WORKDIR}/clusterdeployments.json" jq '.metadata.namespace + "/" + .metadata.name + "-" + .spec.compute[]?.name' | sort > "${TEMPDIR}/machinepools.orig"
```
twiest: Same question about redirecting

hack/v1migration/delete_in_namespace.sh
```diff
@@ -0,0 +1,26 @@
...
+# shellcheck source=hivetypes.sh
+source "$(dirname "$0")/hivetypes.sh"
+
+echo "Deleting machinepools"
+oc delete machinepools -n "${NAMESPACE}" --all --cascade=false
+
+echo "Deleting installconfig secrets"
+oc get clusterdeployments -n "${NAMESPACE}" -ojson | \
+  jq -r '.items[].spec.provisioning.installConfigSecretRef.name' | \
+  xargs -i oc delete secret {} -n "${NAMESPACE}"
```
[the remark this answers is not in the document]
staebler: The aggregated API server creates these secrets. They do not exist in the v1alpha1 API.

hack/v1migration/hive_owner_refs_in_namespace.sh
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR=${1:?must specify working directory}
+NAMESPACE=${2:?must specificy namespace}
+
+if [[ ! -d "$WORKDIR" ]]
+then
+  echo "Directory $WORKDIR does not exist."
+  exit 1
+fi
+
+<"${WORKDIR}/owner-refs.json" jq --arg NAMESPACE "${NAMESPACE}" 'select(.namespace==$NAMESPACE)'
```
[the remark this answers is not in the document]
staebler: Mostly out of habit to stay consistent with piping from `oc` output. If you feel strongly about it, then I can change it.

staebler:
Added to this PR to fix two other issues found during testing.
* Delete the installconfig secret after the clusterdeployment in `delete_in_namespace.sh` script. When the installconfig secret is deleted first, then the `oc delete` command may fail when trying to watch for the clusterdeployment to be fully removed.
* Sort the machinepools created during a clusterdeployment create so that the resourceVersion returned from the create is correct.

PR #880, opened 2020-03-08
Refactor hiveutil create-cluster code into a package.

dgoodwin:
In preparation for cluster pools, we will want a way to generate a ClusterDeployment and all related objects in the core hive code. We however do not want this code reading from the filesystem or environment variables, or assuming the use of apply.

Move parts of hiveutil's create-cluster command which generates the runtime.Objects into a new package. A Generator object can be created and configured incrementally prior to generating all the required objects. Callers likely don't care what the objects are, they just configure the generator, get the list of objects, and then create or apply them.

In the midst of this there is a new option for re-using a credentials secret that already exists.

Still needs validation and unit tests.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -3,6 +3,10 @@ package createcluster
 import (
 	"encoding/json"
 	"fmt"
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
+	gcputils "github.com/openshift/hive/contrib/pkg/utils/gcp"
+	"github.com/openshift/hive/pkg/clusterresource"
+	"github.com/openshift/hive/pkg/gcpclient"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -10,7 +14,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -22,11 +25,7 @@ import (
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
-
-	"github.com/openshift/installer/pkg/ipnet"
-	installertypes "github.com/openshift/installer/pkg/types"
 
 	"github.com/openshift/hive/pkg/apis"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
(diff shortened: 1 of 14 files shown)


dgoodwin:
This could use an initial pass, tests are light and will be adding more, code is in the shape I want.

pkg/clusterresource/builder.go
```diff
@@ -0,0 +1,492 @@
...
+			Kind:       "Secret",
+			APIVersion: corev1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      o.getSSHPrivateKeySecretName(),
+			Namespace: o.Namespace,
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			"ssh-privatekey": o.SSHPrivateKey,
```
twiest: Should this be a constant maybe? 

We're using it in multiple places in the code base (like in `pkg/install/generate.go`). Maybe this should be added to our hive constants file and referenced by other places?

pkg/clusterresource/builder.go
```diff
@@ -0,0 +1,492 @@
...
+			APIVersion: corev1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      o.getServingCertSecretName(),
+			Namespace: o.Namespace,
+		},
+		Type: corev1.SecretTypeTLS,
+		StringData: map[string]string{
+			"tls.crt": o.ServingCert,
+			"tls.key": o.ServingCertKey,
```
twiest: Same question about if these. These are also used in multiple places in the code base. Might be good to constify (definitely not a word) them.

pkg/clusterresource/builder.go
```diff
@@ -0,0 +1,492 @@
...
+
+func (o *Builder) generateAdminKubeconfigSecret() *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      o.getAdoptAdminKubeconfigSecretName(),
+			Namespace: o.Namespace,
+		},
+		Data: map[string][]byte{
+			"kubeconfig":     o.AdoptAdminKubeconfig,
+			"raw-kubeconfig": o.AdoptAdminKubeconfig,
```
twiest: Same constify question.

contrib/pkg/createcluster/create.go
```diff
@@ -737,98 +544,33 @@ func (o *Options) generateManifestsConfigMap() (*corev1.ConfigMap, error) {
...
-func (o *Options) configureImages(cd *hivev1.ClusterDeployment) (*hivev1.ClusterImageSet, error) {
+func (o *Options) configureImages(generator *clusterresource.Builder) (*hivev1.ClusterImageSet, error) {
 	if len(o.ClusterImageSet) > 0 {
-		cd.Spec.Provisioning.ImageSetRef = &hivev1.ClusterImageSetReference{
-			Name: o.ClusterImageSet,
-		}
+		generator.ImageSet = o.ClusterImageSet
 		return nil, nil
 	}
+	// TODO: move release image lookup code to the cluster library
```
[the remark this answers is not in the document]
dgoodwin: Yup future todo.

dgoodwin:
Changes addressed. Squashed commits into one.

PR #883, opened 2020-03-11
Fix crash when using more than 35 GCP MachinePools.

dgoodwin:
Code did not properly handle the case when no lease characters are
available.

Added a short docs blurb, and a condition to be set when this situation
arises.

cc @lwan-wanglin

### DIFF
--- a/pkg/controller/remotemachineset/gcpactuator.go
+++ b/pkg/controller/remotemachineset/gcpactuator.go
@@ -3,18 +3,16 @@ package remotemachineset
 import (
 	"context"
 	"fmt"
+	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
 	"math/rand"
-
-	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
-
-	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	installgcp "github.com/openshift/installer/pkg/asset/machines/gcp"
@@ -91,14 +89,13 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 		return nil, false, errors.New("MachinePool is not for GCP")
 	}
 
-	leaseChar, leaseCreated, err := a.obtainLease(pool, cd, logger)
+	leaseChar, proceed, err := a.obtainLease(pool, cd, logger)
 	if err != nil {
-		logger.WithError(err).Error("error obtaining pool name lease")
+		logger.WithError(err).Log(controllerutils.LogLevel(err), "error obtaining pool name lease")
 		return nil, false, err
 	}
 
-	if leaseCreated {
-		logger.Debug("new machine pool name lease created, waiting until creation is observed")
+	if !proceed {
 		return nil, false, nil
 	}
 
@@ -201,7 +198,7 @@ func (a *GCPActuator) getImageID(cd *hivev1.ClusterDeployment, logger log.FieldL
 // for use in the name of the machine pool. We are severely restricted on name lengths on GCP
 // and effectively have one character of flexibility with the naming convention originating in
 // the installer.
-func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (leaseChar string, leaseCreated bool, leaseErr error) {
+func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (leaseChar string, proceed bool, leaseErr error) {
 	logger.Debugf("obtaining lease for pool: %s", pool.Name)
 
 	leases := &hivev1.MachinePoolNameLeaseList{}
@@ -223,7 +220,7 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
 			if expectedLeaseName != l.Name {
 				return "", false, fmt.Errorf("lease %s did not match expected lease name format (%s[CHAR])", l.Name, expectedLeaseName)
 			}
-			return leaseChar, false, nil
+			return leaseChar, true, nil
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -244,6 +242,41 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
...
+				hivev1.NoMachinePoolNameLeasesAvailable,
+				corev1.ConditionTrue,
+				"AllCharactersInUse",
+				"All characters already in use for GCP MachinePoolNameLeases",
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				pool.Status.Conditions = conds
+				if err := a.client.Status().Update(context.Background(), pool); err != nil {
+					logger.WithError(err).Error("failed to update MachinePool conditions")
```
staebler: The level of the log entry should be set using `controllerutils.LogLevel(err)`.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -244,6 +242,41 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
...
+			corev1.ConditionFalse,
+			"CharactersAvailable",
+			"Characters are available for GCP MachinePoolNameLease",
+			controllerutils.UpdateConditionNever,
+		)
+		if changed {
+			pool.Status.Conditions = conds
+			err := a.client.Status().Update(context.Background(), pool)
+			if err != nil {
+				logger.WithError(err).Error("failed to update MachinePool conditions")
```
staebler: Here also use `controllerutils.LogLevel`.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -244,6 +242,41 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
 		if err != nil {
 			return "", false, err
 		}
+		if len(availLeaseChars) == 0 {
+			logger.Warn("no GCP MachinePoolNameLease characters available, setting condition")
+			conds, changed := controllerutils.SetMachinePoolConditionWithChangeCheck(
+				pool.Status.Conditions,
+				hivev1.NoMachinePoolNameLeasesAvailable,
+				corev1.ConditionTrue,
+				"AllCharactersInUse",
```
staebler: I would rather the reason and message not mention characters and instead say out of machine pool names. That way, if we end up having to use this for some other platform that may have more than 1 character, we can still use the same reason and message.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -244,6 +242,41 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
...
+				"All characters already in use for GCP MachinePoolNameLeases",
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				pool.Status.Conditions = conds
+				if err := a.client.Status().Update(context.Background(), pool); err != nil {
+					logger.WithError(err).Error("failed to update MachinePool conditions")
+				}
+			}
+			return "", false, fmt.Errorf("no lease chars available, unable to assign MachinePoolNameLease to pool %s", pool.Name)
```
staebler: Should this return an error? I think it would be better to watch machinepoolnameleases in the controller and requeue the machinepools on changes to leases.

dgoodwin:
Feedback addressed.

dgoodwin:
Nope just spotted a bug with the change to "proceed" from obtainLeaseChar.

/hold

dgoodwin:
Alright should be correct now, it got quite a bit more complicated but I believe the tests are sorted and correct now, and have completed manual testing as well.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -91,14 +89,13 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 		return nil, false, errors.New("MachinePool is not for GCP")
 	}
 
-	leaseChar, leaseCreated, err := a.obtainLease(pool, cd, logger)
+	leaseChar, proceed, err := a.obtainLease(pool, cd, logger)
 	if err != nil {
 		logger.WithError(err).Error("error obtaining pool name lease")
```
staebler: Use `controllerutils.LogError` here, too.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -244,6 +241,42 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
...
+				hivev1.NoMachinePoolNameLeasesAvailable,
+				corev1.ConditionTrue,
+				"OutOfMachinePoolNames",
+				"All machine pool names are in use",
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				pool.Status.Conditions = conds
+				if err := a.client.Status().Update(context.Background(), pool); err != nil {
+					logger.WithError(err).Log(controllerutils.LogLevel(err), "failed to update MachinePool conditions")
```
staebler: An error should be returned here. Also, logging here is redundant with the logging that will be done in the caller.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -244,6 +241,42 @@ func (a *GCPActuator) obtainLease(pool *hivev1.MachinePool, cd *hivev1.ClusterDe
...
+			corev1.ConditionFalse,
+			"MachinePoolNamesAvailable",
+			"Machine pool names available",
+			controllerutils.UpdateConditionNever,
+		)
+		if changed {
+			pool.Status.Conditions = conds
+			err := a.client.Status().Update(context.Background(), pool)
+			if err != nil {
+				logger.WithError(err).Log(controllerutils.LogLevel(err), "failed to update MachinePool conditions")
```
staebler: This logging is redundant with the logging done in the caller.

pkg/controller/remotemachineset/leaseexceptions.go
```diff
@@ -40,6 +44,47 @@ func (h *machinePoolNameLeaseEventHandler) Create(e event.CreateEvent, q workque
...
+		logger.WithError(err).Log(controllerutils.LogLevel(err),
+			"unable to list machine pools for cluster")
+		return
+	}
+	logger.Debugf("found %d MachinePools for cluster", len(clusterMachinePools.Items))
+	for _, mp := range clusterMachinePools.Items {
+		if mp.Spec.ClusterDeploymentRef.Name != lease.Labels[constants.ClusterDeploymentNameLabel] {
+			continue
+		}
+		h.EnqueueRequestForOwner.Delete(e, q)
```
staebler: Remove this line. This will requeue the owner of the deleted machinepoolnamelease.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -87,6 +87,13 @@ func Add(mgr manager.Manager) error {
 		return err
 	}
 
+	// Watch for changes to MachinePoolNameLeases
+	err = c.Watch(&source.Kind{Type: &hivev1.ClusterDeployment{}}, &handler.EnqueueRequestsFromMapFunc{
```
staebler: Remove this. This is an exact replica of the previous block of code, except for an incorrect change to the comment.

dgoodwin:
Updates pushed.

PR #884, opened 2020-03-11
WIP: new lambda for building running instance reports

joelddiaz:
* A new AWS Lambda function written in Python that can be used to get a list of running instances from a list of regions, and email a report out.
* A CloudWatch event to periodically (once a day) run the Lambda, and provide the Lambda with the values it needs to perform its function (ie recipients, regions to search through, etc).
* Static AWS policies that grant the necessary permissions to the Lambda.
* An Ansible playbook to create all these objects in AWS and stitch them together.

The create/update flow is to run the playbook while providing the list of recipients who will recieve the report and the email to use as the from address (for bounce back emails).

`./upload-lambda.yaml --extra-vars '{"recipients": ["one@example.com", "two@example.com"], "fromemail": "reply@example.com"}'`

### DIFF
--- a/hack/lambda/periodic_lambda_function.py
+++ b/hack/lambda/periodic_lambda_function.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+import boto3
+import json
+import io
+
+def lambda_handler(event, context):
+
+    print(event)
+    running_instances = {}
+
+    for region in event["regions"]:
+        ec2client = boto3.client('ec2', region)
+
+        response = ec2client.describe_instances()
+        
+        running_instances[region] = []
+
+        for resp in response["Reservations"]:
+            for instance in resp["Instances"]:
+                if instance["State"]["Name"] == "running":
+                    for tag in instance["Tags"]:
+                        if tag["Key"] == "Name":
+                            running_instances[region].append(tag["Value"])
+
+        running_instances[region].sort()
+
+    instances_formatted = build_instance_email_text(running_instances)
+
+    send_email(event["recipients"], event["fromemail"], instances_formatted, event["emailregion"])
+
+    return {
+        'statusCode': 200,
+        'body': json.dumps('Lambda complete!')
+    }
+
+def build_instance_email_text(instances):
+    buf = io.StringIO()
+
+    for region in instances.keys():
+        buf.write("Region: {}\n".format(region))
+        for instance in instances[region]:
+            buf.write("{}\n".format(instance))
+
+        buf.write("\n")
+        
+    return buf.getvalue()
+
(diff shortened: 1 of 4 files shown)


dgoodwin:
Very nice.

Feels like it's a sketchy fit for Hive repo, I don't think it has anything hive specific in there, should we spin up another repo somewhere?

Where would GCP and Azure fit into this? I'm using GCP more and more myself. Would we want one job dealing with all cloud providers? I'm assuming it could if we injected creds?

hack/lambda/periodic_lambda_function.py
```diff
@@ -0,0 +1,75 @@
...
+
+    print("Email response: {}".format(response))
+
+def main():
+    event = {"regions": ["us-east-1", "us-east-2"],
+            "recipients": ["someone@example.com"],
+            "fromemail": "return@example.com",
+            "emailregion": "us-east-1",
+            }
+    result = lambda_handler(event, "")
```
dgoodwin: Does executing main actually try to send an email to example.com?

hack/lambda/periodic_lambda_function.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    print("Email response: {}".format(response))
+
+def main():
+    event = {"regions": ["us-east-1", "us-east-2"],
+            "recipients": ["someone@example.com"],
+            "fromemail": "return@example.com",
+            "emailregion": "us-east-1",
+            }
```
dgoodwin: Disgregard previous comment, looks like this is just printing the email contents.

joelddiaz:
> Very nice.
> 
> Feels like it's a sketchy fit for Hive repo, I don't think it has anything hive specific in there, should we spin up another repo somewhere?
> 
> Where would GCP and Azure fit into this? I'm using GCP more and more myself. Would we want one job dealing with all cloud providers? I'm assuming it could if we injected creds?

I did feel strange putting this code in the hive repo. We can certainly move this to somewhere more appropriate.

I suppose one of the nice things about having the AWS code run on AWS is that the creds are not stored anywhere. We simply define the policy that applies to the Lambda and the Lambda gets those permissions whenever it runs. And whoever runs the Ansible script is ultimately the only link in the chain that needs credentials to set everything up.

If we used AWS Lambda to query the status of GCP and/or Azure we'd have to bake in creds. But we would be able to send out only a single notice instead of one per cloud.

dgoodwin:
It would probably be similarly easier to query GCP from a cloud func, but we'd probably have to fight the email approval battle again.

joelddiaz:
closing as we will house this code in another repo

PR #885, opened 2020-03-12
docs: proposal for inter-cluster moves

staebler:


docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## Process
+
+There is a controller that will reconcile ClusterDeployments and watch ClusterDeployments
+and ClustereRelocators. For ClusterDeployments that match with a ClusterRelocator, the
+controller will move the ClusterDeployment to the Hive instance associated with the kubeconfig
+referenced in the ClusterRelocator.
+
+1. Add the `hive.openshift.io/relocating=<relocator-name>` annotation to the ClusterDeployment. This informs other controllers not to take action on the cluster.
```
dgoodwin: It would be nice if we could make this more generic as this could be generally useful. Is the relocator name actually needed? Could we go with something like hive.openshift.io/disabled or something similar.

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+controller from running its finalizer code for the ClusterDeployment.
+1. Create a namespace (or project?) for the ClusterDeployment in the remote cluster.
+1. Copy Secrets, ConfigMaps, MachinePools, SyncSets, and SyncIdentityProviders associated
+with the ClusterDeployment to the remote cluster.
+1. Copy the ClusterDeployment to the remote cluster.
+1. Delete the ClusterDeployment.
+1. Remove the `hive.openshift.io/relocating` annotation. Replace it with a
+`hive.openshift.io/relocated=<relocator-name>` annotation. This frees the clusterdeployment
+controller to run its finalizer code for the ClusterDeployment but informs the controller
+to run perform a deprovision.
```
dgoodwin: "to not perform a deprovision" I assume.

dgoodwin: Would it be any safer to reconcile and ensure there are no finalizers set if relocated is set? (rather than acting on finalizers and checking if the annotations is set in that process) Or perhaps both.

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## Concerns
+1. How does the dnszone controller get informed that the DNS records should be preserved?
+1. What should happen if there are SelectorSyncSets in the source Hive instance that are not present
+in the destination Hive instance?
```
dgoodwin: Could we skip SyncSetInstances entirely? They should all be recreated in the new cluster per whatever SyncSets are present there.

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+1. Copy the ClusterDeployment to the remote cluster.
+1. Delete the ClusterDeployment.
+1. Remove the `hive.openshift.io/relocating` annotation. Replace it with a
+`hive.openshift.io/relocated=<relocator-name>` annotation. This frees the clusterdeployment
+controller to run its finalizer code for the ClusterDeployment but informs the controller
+to run perform a deprovision.
+
+
+## Concerns
+1. How does the dnszone controller get informed that the DNS records should be preserved?
```
dgoodwin: If the new controller reconciles on ClusterDeployment, it can determine if there's a DNSZone for the cluster or not, and then perhaps just ensure the same annotations are applied on the DNSZone and respected by it's controller?

docs/move_clusters.md
staebler: The name is needed to cover the case where the ClusterRelocator is deleted in the middle of a relocation. We already have generic ways of disabled via the syncset-pause annotation and the preserveOnDelete field.

staebler: I did not want to re-use those two things because I wanted to preserve them during the move. For example, if the syncset-pause annotation was set prior to the move, then it should remain set after the move.

docs/move_clusters.md
staebler: I do not like the idea of having two controllers that could both remove the same finalizer. I much prefer to keep the controller that adds the finalizer as the only one that removes the finalizer. Or maybe I am missing something in what you are suggesting. Would you mind elaborating a bit more?

docs/move_clusters.md
staebler: Yes, the idea is to skip them. That is the crux of this concern, though. Let's assume that the source Hive instance has a SelectorSyncSet that lays down resource A but that the destination Hive instance does not have said SelectrSyncSet. When the cluster is relocated, it will still have resource A even though its new home does not have a SelectorSyncSet to manage it.

docs/move_clusters.md
staebler: I think I am more keen to have the dnszone controller look up the ClusterDeployment and check its annotations. If the new controller applies the annotations to DNSZone then there is more surface area for a problem during the period between when the annotations are adjusted on ClusterDeployment and when the annotations are adjusted on DNSZone. I'll have to think about whether that is a problem or not.

docs/move_clusters.md
dgoodwin: Those are not quite as ubiquitous as I'd like (MachinePools exempted for example), but if the name is useful then that's ok.

docs/move_clusters.md
dgoodwin: Ah I see, and if we relocate the SyncSetInstances they would get pruned due to the absence of the SelectorSyncSet, which I'm hoping would cleanup said resources? If so maybe it makes sense to sync them over and let them get cleaned up.

CC @abutcher

docs/move_clusters.md
dgoodwin: It could still be the same controller I think, I'm basically wondering if it's best to have the cluster deployment controller (1) detect relocated annotation being set, and thus clear the normal deprovision finalizer, or (2) to leave the normal deprovision finalizer and let that code trigger, checking if the relocated annotation is being set within. Both sounds appealing, or rather (3) if relocated is set clear finalizer, and in deprovision code which should never run, also check if relocated is set and skip or block the deprovision with lots of errors and alarm bells.

docs/move_clusters.md
staebler: | Those are not quite as ubiquitous as I'd like (MachinePools exempted for example)

I don't understand that statement. Could you rephrase it?

dgoodwin: We can disable syncsets and deprovision, but not remotemachineset controller as an example, I was thinking it might be handy to have a way to disable ALL controllers. (not what this change is about so that can be a discussion for another day if we need it and re-using this annotation feels too weird)

staebler: The syncset-pause annotation **does** disable the remotemachineset controller.

dgoodwin: Ah I see, bad example then, but my point was a mechanism to disable *all* controllers.

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+details about the destination cluster and label selectors to match against the cluster to
+move.
+
+## ClusterRelocator CRD
+
+A ClusterRelocator is cluster-scoped.
+
+This is an example of a ClusterRelocator that may be used to move all of the ClusterDeployments
+with the label `hive-instance: hive2` to the remote cluster accessed via the kubeconfig secret
+at "hive2-remote-kubecofig-secret" in the "hive" namespace.
```
dgoodwin: @jmelis raised a good point here, are we ok for networking? Will this require vpc peering to get from priv cluster to priv cluster?

CC @mwoodson @twiest

staebler: Are you suggesting that the Hive instances would be run in 2 separate private clusters? The relocation does not interact with the cluster created by Hive.

staebler: Or are you suggesting that VPC peering would be set up so that the first Hive instance can communicate with the private cluster and would need to be changed so that the second Hive instance can communicate with the private cluster? I do assume that there will be other parts outside of Hive that will need to do work to relocate.

dgoodwin: Each Hive instance in a separate private cluster, connectivity between the two may not be a given to do the push from A to B may not be a given and require some additional thought.

staebler: OK. Well that is a major restriction that would make this entire proposal unworkable. So I will stop moving forward on any implementation of this until we can get an answer on whether it is feasible.

dgoodwin: It should be fine, it sounds like it's just VPC peering but just need to hear back from those involved if that is feasible.

gregsheremeta: cc @pbergene and @rogbas too
> @jmelis raised a good point here, are we ok for networking? Will this require vpc peering to get from priv cluster to priv cluster?

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+This may be useful if the user want to replace a Hive instance with a new one or if the user
+wants to implement sharding between Hive instances.
+
+Hive will manage all of the moving for the user. The user will create a CR indicating with
+details about the destination cluster and label selectors to match against the cluster to
+move.
+
+## ClusterRelocator CRD
+
+A ClusterRelocator is cluster-scoped.
```
gregsheremeta: KB asked me privately if Hive could grow a feature someday to have multiple Hives installed in a single cluster, i.e. one on hot standby. I guess this would be cheaper than having an entire second Hive cluster on standby. IMO you lose fault tolerance if the cluster dies -- I'd rather just have a hot standby cluster. Thought I would mention it in case this is easy to accomplish without adding a cluster-scoped resource.

docs/move_clusters.md
dgoodwin: And to clarify, no need to stop on impl, we just need to coordinate. This is the best option as it lets us rebalance as well as migrate, we don't want anything else IMO.

jmelis: I agree, I don't expect any major issues with this VPC peering, but it needs some thinking as @dgoodwin has mentioned

docs/move_clusters.md
staebler: The issue with that is the same as using the syncset-pause annotation for relocation. It requires extra mechanisms to determine whether the annotation was present before relocation and therefore needs to be retained after relocation.

Is there a use case for disabling reconciliation of all controllers outside of the relocation use case?

docs/move_clusters.md
staebler: I don't see much difference in maintainability between having a single annotation that every controller has to check and having a single function that every controller has to call.

For the purposes of example, I have changed current code like the following
```
if cd.Annotations[constants.SyncsetPauseAnnotation] == "true" {
```
to
```
if controllerutils.ShouldSyncCluster(cd, ssiLog) {
```

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Summary
+
+There is a need to be able to move Hive-managed clusters from one Hive instance to another.
+This may be useful if the user want to replace a Hive instance with a new one or if the user
+wants to implement sharding between Hive instances.
+
+Hive will manage all of the moving for the user. The user will create a CR indicating with
+details about the destination cluster and label selectors to match against the cluster to
+move.
```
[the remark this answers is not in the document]
staebler: I will try to reword this to make it more clear. Thanks for your suggestions.

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+and ClustereRelocators. For ClusterDeployments that match with a ClusterRelocator, the
+controller will move the ClusterDeployment to the Hive instance associated with the kubeconfig
+referenced in the ClusterRelocator.
+
+1. Add the `hive.openshift.io/relocating=<relocator-name>` annotation to the ClusterDeployment. This informs other controllers not to take action on the cluster.
+This will block syncsets and machinesets from being applied and will block the clusterdeployment
+controller from running its finalizer code for the ClusterDeployment.
+1. Create a namespace (or project?) for the ClusterDeployment in the remote cluster.
+1. Copy Secrets, ConfigMaps, MachinePools, SyncSets, and SyncIdentityProviders associated
+with the ClusterDeployment to the remote cluster.
```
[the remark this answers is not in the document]
staebler: My hope is that I can move just the things needed by Hive. My initial proof-of-concept moved an entire namespace at a time so that I did not have to worry about which resources in the namespace were needed. But that was when the move was done as an external tool. Now that the move is done for a specific ClusterDeployment and we support multiple ClusterDeployments in a single namespace, things are more tricky.

docs/move_clusters.md
gregsheremeta: > I'm not sure how this would help. What use cases are we trying to address with this?
> Hive relies on kube for HA.

That was pretty much my thought.

dgoodwin: Yeah on the surface I don't think this makes sense either, just bump your replicas and use leader election, there's nothing having another set of pods running would buy you. 

Separate hot backup hive clusters also sounds troubling, we don't want to be writing our own replication and failover code.

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
+# Moving Hive-managed Clusters
+
+## Summary
+
+There is a need to be able to move Hive-managed clusters from one Hive instance to another.
+This may be useful if the user want to replace a Hive instance with a new one or if the user
+wants to implement sharding between Hive instances.
+
+Hive will manage all of the moving for the user. The user will create a CR indicating with
```
csrwng, line 9: s/with details/details/

docs/move_clusters.md
```diff
@@ -0,0 +1,59 @@
...
+## Process
+
+There is a controller that will reconcile ClusterDeployments and watch ClusterDeployments
+and ClustereRelocators. For ClusterDeployments that match with a ClusterRelocator, the
+controller will move the ClusterDeployment to the Hive instance associated with the kubeconfig
+referenced in the ClusterRelocator.
+
+1. Add the `hive.openshift.io/relocating=<relocator-name>` annotation to the ClusterDeployment. This informs other controllers not to take action on the cluster.
+This will block syncsets and machinesets from being applied and will block the clusterdeployment
+controller from running its finalizer code for the ClusterDeployment.
```
csrwng, line 44: Will the annotation be applied to a clusterdeployment in any state? or will we wait for an ongoing provision/deprovision to finish before applying the annotation?

docs/move_clusters.md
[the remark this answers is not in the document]
csrwng: What about resources created by someone other than us? but still relevant to the cluster, like certman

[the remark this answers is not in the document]
csrwng: Also, will we get some unexpected behavior from controllers that manage these resources as the resources are created but the clusterdeployment is not copied over yet?

docs/move_clusters.md
staebler: If a provision is ongoing, then the clusterdeployment controller will stop the provisioning when relocation commences.

If deprovisioning is ongoing, then the ClusterDeployment is not eligible to be relocated. We should not relocate deleted resources.

docs/move_clusters.md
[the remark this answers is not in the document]
staebler: Custom resources not managed by Hive will not be relocated by Hive.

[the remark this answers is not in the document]
staebler: The guidance that I would give to users with resources that reference Hive resources is to add a finalizer to the resource that the user owns and add an owner reference to the ClusterDeployment. When the ClusterDeployment has been relocated, then the user's controller can move the resource to the new Hive instance.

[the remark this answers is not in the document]
staebler: Alternatively, I could see some new CRD linked to the ClusterDeployment that reference a resource that needs to be moved in the event of a relocation. Users could add a CR to have their resource included in the relocation.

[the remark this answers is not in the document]
csrwng: That sounds cool, though it'd be good to understand what would have to be moved to see if such a mechanism makes sense.

gregsheremeta:
should we merge this, since it's coded?

PR #887, opened 2020-03-16
OLM artifacts: switch managed dns config to an array

gregsheremeta:
OLM artifacts: switch managed dns config to an array

To more easily support multiple managed domains and dns providers,
switch managed dns config to an array.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -10,14 +10,17 @@ parameters:
   value: staging
 - name: IMAGE_TAG
   value: latest
-- name: EXTERNAL_DNS_AWS_CREDS
-  value: external-dns-route53-admin
-- name: EXTERNAL_DNS_MANAGED_DOMAIN
-  value: s1.openshiftapps.com
 - name: ADDITIONAL_CERTIFICATE_AUTHORITY
   value: letsencrypt-ca
 - name: GLOBAL_PULL_SECRET
   value: ""
+- name: MANAGED_DOMAINS
+  value:
+  - aws:
+      credentialsSecretRef:
+        name: my-dns-creds
+    domains:
+    - openshift.example.com
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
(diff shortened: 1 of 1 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -55,11 +55,6 @@ objects:
   spec:
     additionalCertificateAuthoritiesSecretRef:
     - name: ${ADDITIONAL_CERTIFICATE_AUTHORITY}
-    managedDomains:
-    - aws:
-        credentialsSecretRef:
-          name: ${EXTERNAL_DNS_AWS_CREDS}
-      domains:
-      - ${EXTERNAL_DNS_MANAGED_DOMAIN}
+    managedDomains: "${{MANAGED_DOMAINS}}"
```
maorfr: The new parameter should be also added to the parameters section.
I suggest to add it with a default value. The reason is that this PR will be the first to be merged, while the saas repo still does not contain the parameter.

maorfr:
This PR should be merged **after** https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3681 and **before** https://gitlab.cee.redhat.com/service/saas-hive/merge_requests/37

hack/olm-registry/olm-artifacts-template.yaml
gregsheremeta: done

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -10,14 +10,17 @@ parameters:
   value: staging
 - name: IMAGE_TAG
   value: latest
-- name: EXTERNAL_DNS_AWS_CREDS
-  value: external-dns-route53-admin
-- name: EXTERNAL_DNS_MANAGED_DOMAIN
-  value: s1.openshiftapps.com
```
maorfr: i propose we keep those around to avoid failing on redundant parameters in the saas repo :man_facepalming:

dgoodwin: These should not be present in the public Hive repo IMO. ACM folks picked them up accidentally and then were asking what they were used for.

gregsheremeta: removed

PR #891, opened 2020-03-17
modify openshift detection logic

joelddiaz:
Rather than list DeploymentConfig objects to detect whetther or not hive is running on openshift, query the available resources for the DeploymentConfig group/version to determine OpenShift vs non-OpenShift.

This should silence some of the log output from the operator that looks like:

`E0303 16:28:56.009055 1 reflector.go:329] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:204: expected type *v1beta1.CustomResourceDefinition, but watch event object had type *unstructured.Unstructured`

by having the controller-runtime not open a watch on the DeploymentConfigs.


dgoodwin:
[retest]

PR #892, opened 2020-03-17
add config setting for cli image

dhellmann:
When using hive in a disconnected bare metal environment the user
needs to be able to specify the location of the CLI image. Add a field
to the config spec and use it to set the existing environment variable
via the deployment for the controller.

### DIFF
--- a/pkg/apis/hive/v1/hiveconfig_types.go
+++ b/pkg/apis/hive/v1/hiveconfig_types.go
@@ -57,6 +57,10 @@ type HiveConfigSpec struct {
 
 	// DeprovisionsDisabled can be set to true to block deprovision jobs from running.
 	DeprovisionsDisabled *bool `json:"deprovisionsDisabled,omitempty"`
+
+	// ImageExtractionCLIImage can be set to override the location of
+	// the image with the oc command line tool.
+	ImageExtractionCLIImage string `json:"imageExtractionCLIImage,omitempty"`
 }
 
 // HiveConfigStatus defines the observed state of Hive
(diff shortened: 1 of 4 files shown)


pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -57,6 +57,10 @@ type HiveConfigSpec struct {
 
 	// DeprovisionsDisabled can be set to true to block deprovision jobs from running.
 	DeprovisionsDisabled *bool `json:"deprovisionsDisabled,omitempty"`
+
+	// CLIImage can be set to override the location of the image with
+	// the oc command line tool.
+	CLIImage string `json:"cliImage,omitempty"`
```
staebler: I am a little concerned about the name of this variable. The CLI image term is overloaded in Hive, where it could mean the CLI image used to extract images from the release image or the CLI image used by the installmanager to run oc adm commands. I think we should be a bit more specific with the name of this variable here since this is exposed directly to the end user. How do you feel about `ImageExtractionCLIImage`? It is a bit wordy.

dgoodwin: +1, that name sounds ok to me.

Change looks great in general, thanks for wiring this up @dhellmann

dhellmann: I've updated the name.

dhellmann:
What should I do about documentation for this new option?

dgoodwin:
At present we just go-doc our API types. I don't think this needs  anything in official docs at this time.

/lgtm

PR #893, opened 2020-03-18
Vendor latest installer master for deprovision code.

dgoodwin:
Overrides are required for several transient dependencies to be able to
compile, I believe because the installer has switched to go modules.

Tested deprovision of both 4.4 and 4.5 AWS clusters after this change.


dgoodwin:
4.3.0 good as well.

staebler:
> 4.3.0 good as well.

/lgtm

PR #894, opened 2020-03-19
Validate SyncSet ResourceApplyMode

abutcher:
https://issues.redhat.com/browse/CO-830

SyncSet [resourceApplyMode](https://github.com/openshift/hive/blob/master/docs/syncset.md#syncset-object-definition) is case sensitive and we only support `""`, `"Upsert"` (default) and `"Sync"` modes so we should validate them in the API.

Ex.
```
➜  oc create -f ~/syncsets/sss.yaml
The SelectorSyncSet "mygroup" is invalid: spec.ResourceApplyMode: Unsupported value: "sync": supported values: "Upsert", "Sync"
```

### DIFF
--- a/pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
+++ b/pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
@@ -42,6 +42,21 @@ var validPatchTypes = map[string]bool{
 
 var validPatchTypeSlice = []string{"json", "merge", "strategic"}
 
+var (
+	validResourceApplyModes = map[hivev1.SyncSetResourceApplyMode]bool{
+		hivev1.UpsertResourceApplyMode: true,
+		hivev1.SyncResourceApplyMode:   true,
+	}
+
+	validResourceApplyModeSlice = func() []string {
+		v := make([]string, 0, len(validResourceApplyModes))
+		for m := range validResourceApplyModes {
+			v = append(v, string(m))
+		}
+		return v
+	}()
+)
+
 // SyncSetValidatingAdmissionHook is a struct that is used to reference what code should be run by the generic-admission-server.
 type SyncSetValidatingAdmissionHook struct{}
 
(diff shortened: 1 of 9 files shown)


gregsheremeta:
can you add some information to the commit and the PR:
 - what is this mode?
 - what are we adding it?

 thanks! :smile:

pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -42,6 +42,13 @@ var validPatchTypes = map[string]bool{
 
 var validPatchTypeSlice = []string{"json", "merge", "strategic"}
 
+var validResourceApplyModes = map[string]bool{
```
staebler: Use the actual `SyncSetResourceApplyMode` values.

```
var (
	validResourceApplyModes = map[hivev1.SyncSetResourceApplyMode]bool{
		hivev1.UpsertResourceApplyMode: true,
		hivev1.SyncResourceApplyMode: true,
	}

	validResourceApplyModeSlice = func() []string {
		v := make([]string, 0, len(validResourceApplyModes))
		for m := range validResourceApplyModes {
			v = append(v, string(m))
		}
		return v
	}()
)
```

pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -247,6 +256,14 @@ func validatePatches(patches []hivev1.SyncObjectPatch, fldPath *field.Path) fiel
 	return allErrs
 }
 
+func validateResourceApplyMode(resourceApplyMode hivev1.SyncSetResourceApplyMode, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if resourceApplyMode != "" && !validResourceApplyModes[string(resourceApplyMode)] {
```
staebler: If the keys of `resourceApplyMode` are `SyncSetResourceApplyMode` objects, then you do not need the cast to `string`.

staebler:
These comments need to change, too. https://github.com/openshift/hive/blob/584da850d20727863b3b7655ff5637bb9baa03d9/pkg/apis/hive/v1/syncset_types.go#L133-L135

staebler:
And here. https://github.com/openshift/hive/blob/584da850d20727863b3b7655ff5637bb9baa03d9/pkg/apis/hive/v1/syncset_types.go#L188-L190

staebler:
And here. https://github.com/openshift/hive/blob/584da850d20727863b3b7655ff5637bb9baa03d9/pkg/apis/hive/v1/syncsetinstance_types.go#L49-L51

And may as well here. https://github.com/openshift/hive/blob/584da850d20727863b3b7655ff5637bb9baa03d9/pkg/controller/syncsetinstance/syncsetinstance_controller.go#L622

staebler:
We need a plan for dealing with pre-existing resources that will fail the new validation.

On hive-int, hive-stage, and hive-prod, there are resources that use "", "sync", and "Sync".
```sh
$ (oc get selectorsyncset -ojson; oc get syncset -A -ojson) | jq -r '.items[].spec.resourceApplyMode' | sort -u
null
sync
Sync
```

jewzaam:
/hold

OSD uses "sync" in literally every SSS, we need to update all these before we can swap over.  PRs are up and linked in https://issues.redhat.com/browse/OSD-3163

jewzaam:
/hold cancel

enough things are using "Sync" now we can turn on the validation in production, anything that isn't already updated has promotions including the fix coming very soon

PR #895, opened 2020-03-23
Improved clearing of provision/deprovision underway metrics.

dgoodwin:
Situations were encountered where the edge-trigger event for clearing
the metrics was being missed, resulting in alerts that would never clear
until the pod was restarted.

Delete was discovered and allows us to clear a metric, regardless if it
exists or not. This is far preferable to setting to 0 in very specific
scenarios and then stopping reporting.

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -83,7 +83,7 @@ var (
 	MetricClusterDeploymentProvisionUnderwaySeconds = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "hive_cluster_deployment_provision_underway_seconds",
-			Help: "Length of time a cluster has been provisioning. Goes to 0 on successful install and then will no longer be reported.",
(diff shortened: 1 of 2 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -187,16 +187,25 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+				} else {
+
+					// For deprovisioning clusters we report the seconds since
+					// cluster was deleted. clusterdeployment_controller should delete this
+					// when removing the finalizer.
+					MetricClusterDeploymentDeprovisioningUnderwaySeconds.WithLabelValues(
+						cd.Name,
+						cd.Namespace,
+						GetClusterDeploymentType(&cd)).Set(
+						time.Since(cd.CreationTimestamp.Time).Seconds())
```
2uasimojo, line 208: Oops. Copypasta.
```suggestion
						time.Since(cd.DeletionTimestamp.Time).Seconds())
```
Fixed via #1981.

PR #896, opened 2020-03-23
imageset: do not use oc adm to get installer and cli images 

staebler:
Instead of using oc adm, load the release image in a container and parse the installer and cli images from the image-references file.


staebler:
https://issues.redhat.com/browse/CO-843

pkg/imageset/updateinstaller.go
```diff
@@ -137,91 +120,71 @@ func (o *UpdateInstallerImageOptions) Validate() error {
...
+	imageStreamData, err := ioutil.ReadFile(filepath.Join(o.WorkDir, imageReferencesFilename))
+	if err != nil {
+		return errors.Wrapf(err, "could not read %s file", imageReferencesFilename)
+	}
+	is := &imageapi.ImageStream{}
+	if err := yaml.Unmarshal(imageStreamData, &is); err != nil {
+		return errors.Wrap(err, "unable to load release image-references")
+	}
+	if is.Kind != "ImageStream" || is.APIVersion != "image.openshift.io/v1" {
+		return errors.Wrap(err, "unrecognized image-references in release payload")
```
dgoodwin: Can you confirm this works as far back as 4.1? We could have the discussion if not but it's a lot easier if we can just keep supporting 4.1.

staebler: Yep. I will check on that.

staebler: This worked with https://registry.svc.ci.openshift.org/v2/ocp/release/manifests/4.1.0-0.nightly-2020-03-21-052817. Testing did reveal a problem where the pull secret was not being used. I updated the code to add the pull secret to the pod spec.

PR #899, opened 2020-03-24
operator bundle 1.0.0

gregsheremeta:
operator bundle 1.0.0

The CRDs are copied -- no need to review those


bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,620 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    categories: OpenShift Optional
+    certified: 'false'
+    containerImage: quay.io/openshift-hive/hive:1.0.0
```
dgoodwin: this should probably be v1.0.0 as well, I recall things can get weird out there in stange places and i think this is consistent with most openshift images

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,620 @@
...
+    createdAt: '2020-03-23T22:36:00Z'
+    description: OpenShift cluster provisioning and management at scale.
+    support: Devan Goodwin
```
dgoodwin: Not entirely comfortable with this one :) but it's ok for now.

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,620 @@
...
+  name: hive-operator.v1.0.0-1
```
dgoodwin: What's the -1 for here?

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,620 @@
...
+  namespace: hive
+spec:
+  customresourcedefinitions:
+    owned:
+    - description: Configuration for the Hive Operator
+      displayName: Hive Config
+      kind: HiveConfig
+      name: hiveconfigs.hive.openshift.io
+      version: v1
```
dgoodwin: If we *only* reference the HiveConfig CRD here, I have a suspicion all the other files in this PR should not be included and would not be used.

bundle/hive-package.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+- currentCSV: hive-operator.v1.0.0-1
+  name: alpha
+defaultChannel: alpha
+packageName: hive-operator
```
dgoodwin: So our bundle/ dir will just grow over time with each new version, and this file is the central entrypoint for the version graph, do I have that right? This is quite cool if so.

version.txt
```diff
@@ -0,0 +1,2 @@
+1.0.0
```
dgoodwin: 1.0.0 or v1.0.0?

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,620 @@
+apiVersion: operators.coreos.com/v1alpha1
```
dgoodwin: We should watch out for v1alpha1, might be ok now but there should be a more stable version by now, I would be surprised if not.

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: I was going to attempt to delete them, but was thinking then i need to tweak permissions down below and that scared me off

bundle/hive-package.yaml
gregsheremeta: yes

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: ok

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: mentioned this on slack. It's a release number -- the CSV release version, akin to an RPM release. I was thinking the packaging is independent of the payload. We might want to tweak something in the CSV only, and keep the binary the same, so 1.0.0-1 --> 1.0.0-2. Maybe not needed if we're checking in the CSV here. If we don't use a release number, then any small CSV tweak (like change the icon) requires a version bump for the hive image itself.

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: I'm happy to change it -- what to?

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: ok will change

version.txt
gregsheremeta: will change to v1.0.0

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
dgoodwin: Nah it's ok for now.

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
dgoodwin: This would imply editing directly in the community repo, and then overwriting later? I'm worried our dependency graph will break here if we start doing things outside what is committed. Is that possible? We need to consistently know what version we're replacing if I understand this right, in which case I think any CSV changes needed should be tagged and build as a new version of hive.

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
dgoodwin: I believe rbac like most kube things is tolerant of types not existing yet, so if it was rbac to crds that don't exist yet that worried you, i think that's ok. I've seen RBAC let me mistakenly grant permissions to things that don't exist many times. :)

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: > This would imply editing directly in the community repo, and then overwriting later

nope I was thinking that any small edits in /bundle wouldn't necessitate a version bump. New icon is a good example:
1. edit the CSV in this repo, set the icon
2. bump the release number, but not the hive version (1.0.0-1 --> 1.0.0-2)
3. commit CSV here
4. commit CSV to community-operators repo

This means no new tag after a minor CSV change, which I feel meh about. If you feel sfaer to bind operator and operand version (and icon change means 1.0.0 --> 1.0.1), I'm ok with it too.

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: ok i'll test with deleting them

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: oops already changed it

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: done

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: done

bundle/1.0.0-1/hive-operator.v1.0.0-1.clusterserviceversion.yaml
gregsheremeta: `MachinePool is not in the list of crds in pkg/operator/hive/hive.go, line 172`

will be fixed later

version.txt
gregsheremeta: done

gregsheremeta:
@dgoodwin ready for re-review

bundle/1.0.0/hive-operator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,620 @@
+apiVersion: operators.coreos.com/v1alpha1
```
dgoodwin: Really sorry Greg I know you're rushing, can we get this into bundle/v1.0.0/, from past experience it's best we get a consistent way of writing the version otherwise we'll all be guessing and trial and erroring for all time. We should use v1.0.0 everywhere.

PR #900, opened 2020-03-24
Dynamically apply all hive CRDs from the operator.

dgoodwin:
Static list was often out of date and currently missing MachinePool.
Switch to reading all CRD files from the expected directory and applying
each.

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -6,8 +6,10 @@ import (
 	"crypto/md5"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
@@ -24,7 +26,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


pkg/operator/hive/hive.go
```diff
@@ -5,7 +5,9 @@ import (
 	"context"
 	"crypto/md5"
 	"fmt"
+	"github.com/pkg/errors"
```
staebler: This import should go down in the section with logrus.

pkg/operator/hive/hive.go
```diff
@@ -202,6 +185,22 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+	// as it requires a significant privilege escalation we would rather
+	// leave in the hands of OLM.
+	crdFiles, err := assets.AssetDir("config/crds")
+	if err != nil {
+		return errors.Wrap(err, "error listing CRD asset files")
+	}
+	for _, a := range crdFiles {
+		err = util.ApplyAsset(h, filepath.Join("config/crds", a), hLog)
+		if err != nil {
+			return err
```
staebler: Wrap this error around a message indicating which crd could not be applied.

pkg/operator/hive/hive.go
```diff
@@ -202,6 +185,22 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+	// the latest in the operator to ensure updates roll out.
+	// TODO: Attempt removing this once Hive is running purely on 4.x,
+	// as it requires a significant privilege escalation we would rather
+	// leave in the hands of OLM.
+	crdFiles, err := assets.AssetDir("config/crds")
+	if err != nil {
+		return errors.Wrap(err, "error listing CRD asset files")
+	}
+	for _, a := range crdFiles {
+		err = util.ApplyAsset(h, filepath.Join("config/crds", a), hLog)
```
staebler: Would you object to doing `if err := ...; err != nil {`?

pkg/operator/hive/hive.go
```diff
@@ -227,7 +226,7 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 	for _, isName := range removeImageSets {
 		clusterImageSet := &hivev1.ClusterImageSet{}
 		err := r.Get(context.Background(), types.NamespacedName{Name: isName}, clusterImageSet)
-		if err != nil && !errors.IsNotFound(err) {
+		if err != nil && !apierrors.IsNotFound(err) {
```
staebler: Although not really related to your changes, the `err != nil` is unnecessary.

dgoodwin: This boils down to a is err.(type) which makes me nervous but how about we just delete the whole block as it's ancient cleanup. :)

staebler: Even better!

Although I would like to know more about what makes you nervous about removing the `err != nil`.

dgoodwin:
Updated. We can't apply everything in rbac without careful thought as part of that is not applied by our operator rather OLM, and I believe there are random names involved, it's just these 'addon' roles we deliver. We can look at this in future but for now I'm happy just to get the main CRD pain point taken care of.

pkg/operator/hive/hive.go
dgoodwin: a nil err and err.(type) is not something i would ever see and assume was ok. It appears it's fine but it just looks like something that could cause a panic and crash. It reads better IMO with the != nil guard.

staebler: OK. It is guaranteed from the language to not panic. It is perfectly valid to do nil.(type).

PR #902, opened 2020-03-25
initial support for OpenStack

joelddiaz:
- [x] add openstack platform details to API/types
- [x] extend webhook validation
- [x] add deprovisioning support
- [x] add provisioning support
- [x] add hiveutil create-cluster support

Because there is no management of DNS when installing on OpenStack, the DNS setup for the API endpoint for the cluster must be performed in advance. On the OpenStack cluster create or pick a Floating IP that will be used for the installation. For a cluster being installed into `stack.example.com` with cluster name `mycluster`, the DNS A record named `api.mycluster.stack.example.com` should point to the Floating IP created/noted previously.

Also, because there is no management of DNS, after the cluster is installed, create or pick another Floating IP. Make the DNS entry `*.apps.mycluster.stack.example.com` pointing to the Floating IP. Then associate the apps Floating IP with the ingress port created during cluster installation:

```
INGRESS_PORT=$(openstack port list --format value -c Name | awk "/mycluster.*-ingress-port/ {print}")
openstack floating ip set --port ${INGRESS_PORT} FLOATING_IP_FOR_APPS
```


config/crds/hive_v1_clusterdeployment.yaml
```diff
@@ -226,7 +226,7 @@ spec:
                         will be created.
                       type: string
                   type: object
-                baremetal:
+                bareMetal:
```
dgoodwin: This looks to be undoing an intentional change, this should be "baremetal"

joelddiaz: Yep. Must have snuck in with the rebase.

gregsheremeta:
what's the status of this one?

dgoodwin:
I've been trying to test but stuck getting CRC + libvirt up. Hopefully soon.

joelddiaz:
updated PR to catch move from dep to go modules

pkg/clusterresource/builder.go
```diff
@@ -101,6 +101,12 @@ type Builder struct {
 	// ReleaseImage is a specific OpenShift release image to install this cluster with. Will override
 	// ImageSet.
 	ReleaseImage string
+
+	// MachineNetwork is the subnet to use for the cluster's machine network.
+	MachineNetwork string
+
+	// SkipMachinePoolGeneration is set to skip generating MachinePool objects
+	SkipMachinePoolGeneration bool
```
dgoodwin, line 109: What's the use case for this option. I can't immediately think of why we'd want to skip the machine pools.

contrib/pkg/utils/openstack/openstack.go
```diff
@@ -0,0 +1,33 @@
...
+	"os"
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/hive/pkg/constants"
+)
+
+// GetCreds reads OpenStack credentials either from either the specified credentials file,
+// ~/.config/openstack/clouds.yaml, or /etc/openstack/clouds.yaml
```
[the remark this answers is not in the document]
joelddiaz: Yes, I got it from https://docs.openstack.org/python-openstackclient/pike/configuration/index.html#clouds-yaml

pkg/apis/hive/v1/openstack/machinepools.go
```diff
@@ -0,0 +1,43 @@
...
+	FlavorName string `json:"type"`
+
+	// RootVolume defines the root volume for instances in the machine pool.
+	// The instances use ephemeral disks if not set.
+	// +optional
+	RootVolume *RootVolume `json:"rootVolume,omitempty"`
+}
+
+// Set sets the values from `required` to `a`.
+func (o *MachinePool) Set(required *MachinePool) {
```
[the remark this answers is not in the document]
joelddiaz: No. I just copied the file directly from the installer https://github.com/openshift/installer/blob/master/pkg/types/openstack/machinepool.go

pkg/clusterresource/builder.go
joelddiaz: I put it there b/c our current OpenStack support has no support for MachinePools

dgoodwin:
/lgtm

Will iterate if any changes are needed once I'm able to get up and running on OpenStack.

contrib/pkg/deprovision/openstack.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// openStackOptions is the set of options to deprovision an OpenStack cluster
+type openStackOptions struct {
+	logLevel string
+	infraID  string
+	cloud    string
+}
+
+// NewDeprovisionOpenStackCommand is the entrypoint to create the GCP deprovision subcommand
```
gregsheremeta, line 22: s/GCP/OpenStack/

joelddiaz:
ok, i sqashed (and grabbed @gregsheremeta 's feedback in there as well)

PR #903, opened 2020-03-26
Modified the Deployment version in e2e test postinstall file

lwan-wanglin:
The apiVersion "extensions/v1beta1" got deprecated,so changed it from extensions/v1beta1 to apps/v1.


dgoodwin:
/lgtm

Thanks Lin!

PR #906, opened 2020-03-26
Update to secretMappings in syncset doc.

abutcher:
Update our syncset docs to reflect the name change from `secretReferences` to `secretMappings` introduced in https://github.com/openshift/hive/pull/783.

/cc @csrwng


dgoodwin:
This caused https://issues.redhat.com/browse/CO-839 didn't it. I'm going to assign to you if you don't mind abutcher and move to code review.

PR #907, opened 2020-03-26
Stop treating syncset applyErrs as controller errors.

abutcher:
https://issues.redhat.com/browse/CO-793

Return `reconcile.Result{Requeue: true}` when an error is encountered applying a syncset instead of returning the error. This will backoff as if we had returned the error but will no longer result in a controller error for metrics. An `applyErr`can occur if we can't apply the resource/patch or if we can't read a local secret wrt `secretMappings`.

An `applyErr` is still logged as an error in several places. I can change this if we'd rather treat as debug logs.

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -296,7 +296,9 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 
 	ssiLog.Info("done reconciling syncsetinstance")
 	if applyErr != nil {
-		return reconcile.Result{}, applyErr
+		ssiLog.WithError(applyErr).Warn("failed to apply, requeueing")
+		// Set requeue to true instead of returning the applyErr to requeue the syncsetinstance.
+		return reconcile.Result{Requeue: true}, nil
 	}
 
 	reapplyDuration := r.ssiReapplyDuration(ssi)
@@ -426,7 +428,7 @@ func (r *ReconcileSyncSetInstance) deleteSyncSetResources(ssi *hivev1.SyncSetIns
 			WithField("kind", resourceStatus.Kind)
 		gv, err := schema.ParseGroupVersion(resourceStatus.APIVersion)
 		if err != nil {
-			itemLog.WithError(err).Error("cannot parse resource apiVersion, skipping deletion")
+			itemLog.WithError(err).Warn("cannot parse resource apiVersion, skipping deletion")
 			continue
 		}
 		gvr := gv.WithResource(resourceStatus.Resource)
@@ -437,10 +439,10 @@ func (r *ReconcileSyncSetInstance) deleteSyncSetResources(ssi *hivev1.SyncSetIns
 			case errors.IsNotFound(err):
 				itemLog.Debug("resource not found, nothing to do")
 			case errors.IsForbidden(err):
-				itemLog.WithError(err).Error("forbidden resource deletion, skipping")
+				itemLog.WithError(err).Warn("forbidden resource deletion, skipping")
 			default:
 				lastError = err
-				itemLog.WithError(err).Error("error deleting resource")
+				itemLog.WithError(err).Warn("error deleting resource")
 				ssi.Status.Resources[index].Conditions = r.setDeletionFailedSyncCondition(ssi.Status.Resources[index].Conditions, fmt.Errorf("failed to delete resource: %v", err))
 			}
 		}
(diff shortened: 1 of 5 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -290,7 +290,9 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 
 	ssiLog.Info("done reconciling syncsetinstance")
 	if applyErr != nil {
-		return reconcile.Result{}, applyErr
+		ssiLog.Debugf("failed to apply, requeueing: %s", applyErr)
```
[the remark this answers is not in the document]
dgoodwin: Info or warn sounds good to me.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -290,7 +290,9 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
...
+		// Set requeue to true instead of returning the applyErr to requeue the syncsetinstance.
+		return reconcile.Result{Requeue: true}, nil
```
dgoodwin: Really glad to learn we can do this and still get backoff.

staebler: Agree. I am using that behavior elsewhere too. At some point, we should make a push to address how all of our controllers return errors. I suppose that will be needed as part of the upgrade of controller-runtime.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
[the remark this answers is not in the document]
abutcher: I've gone with warn in all cases where we'd log an error while applying.

gregsheremeta:
what the status of this one?

abutcher:
@gregsheremeta ready for another review

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -640,29 +639,31 @@ func TestSyncSetReconcile(t *testing.T) {
 				remoteClusterAPIClientBuilder: func(*hivev1.ClusterDeployment) remoteclient.Builder { return mockRemoteClientBuilder },
 				reapplyInterval:               2 * time.Hour,
 			}
-			_, err := r.Reconcile(reconcile.Request{
+			rec, _ := r.Reconcile(reconcile.Request{
```
staebler, line 642: Assert that the error is not nil--using `assert.NotNil`.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
```diff
@@ -640,29 +639,31 @@ func TestSyncSetReconcile(t *testing.T) {
...
-			if !test.expectErr && err != nil {
-				t.Fatalf("unexpected error: %v", err)
-			} else if test.expectErr && err == nil {
-				t.Fatal("expected error not returned")
+
+			if !test.expectRequeue && rec.Requeue {
+				t.Fatalf("unexpected syncsetinstance requeue")
+			} else if test.expectRequeue && !rec.Requeue {
+				t.Fatalf("expected syncsetinstance to be requeued")
 			}
```
staebler, line 653: Replace this with `assert.Equal`.

pkg/resource/apply.go
```diff
@@ -57,7 +57,7 @@ func (r *Helper) Apply(obj []byte) (ApplyResult, error) {
 	if err != nil {
 		r.logger.WithError(err).
 			WithField("stdout", ioStreams.Out.(*bytes.Buffer).String()).
-			WithField("stderr", ioStreams.ErrOut.(*bytes.Buffer).String()).Error("running the apply command failed")
+			WithField("stderr", ioStreams.ErrOut.(*bytes.Buffer).String()).Warn("running the apply command failed")
```
staebler, line 60: These changes to the resource helpers have some more far-reaching implications as they also affect calls made in hive-operator, I believe. It's probably OK still, though. Ideally, the resource helper would not be the one doing the logging, since it cannot know the severity, but that is well beyond something that we should tackle for this PR. I would rather have the hive-operator log warnings than have the syncsetinstance controller log errors.

pkg/controller/syncsetinstance/syncsetinstance_controller_test.go
abutcher: We're not testing any Reconcile errors currently so I'm now ensuring the err is `nil`. Most error cases are related to getting resources we need to continue reconciling the ssi.

PR #908, opened 2020-03-30
cherry-pick #896 to OCM 4.4.0

dhellmann:


dgoodwin:
Meh no tests on this branch. We should probably do something about that at some point.

PR #909, opened 2020-03-30
controller: use requeue as backoff for unreachable controller

staebler:
Use the unreachable controller work queue backoff as the means of backing off from attempting to connect to an unreachable remote cluster.

If an API URL override is set, continue attempting to connect to the cluster using the work queue backoff rather than falling back to the 2-hour check used when the cluster is reachable.

When a reachable ClusterDeployment is synced in the unreachable controller, requeue the ClusterDeployment so that it is re-synced for the 2-hour check.

In other controllers that connect to the remote cluster, set the unreachable condition if there is a failure to connect to the remote cluster. This will give a bit faster feedback when a cluster becomes unreachable, and will allow the controllers to fall back to the initial API URL more quickly.

https://issues.redhat.com/browse/CO-847

### DIFF
--- a/pkg/controller/clusterstate/clusterstate_controller_test.go
+++ b/pkg/controller/clusterstate/clusterstate_controller_test.go
@@ -163,7 +163,6 @@ func TestClusterStateReconcile(t *testing.T) {
 			mockCtrl := gomock.NewController(t)
 			defer mockCtrl.Finish()
 			mockRemoteClientBuilder := remoteclientmock.NewMockBuilder(mockCtrl)
-			mockRemoteClientBuilder.EXPECT().Unreachable().Return(false)
 			if !test.noRemoteCall {
 				mockRemoteClientBuilder.EXPECT().Build().Return(fake.NewFakeClient(test.remote...), nil)
 			}
@@ -233,6 +232,12 @@ func testClusterDeployment() *hivev1.ClusterDeployment {
 				},
 			},
 			Installed: true,
+		},
+		Status: hivev1.ClusterDeploymentStatus{
+			Conditions: []hivev1.ClusterDeploymentCondition{{
+				Type:   hivev1.UnreachableCondition,
+				Status: corev1.ConditionFalse,
+			}},
 		},
 	}
 }
--- a/pkg/controller/controlplanecerts/controlplanecerts_controller.go
+++ b/pkg/controller/controlplanecerts/controlplanecerts_controller.go
@@ -73,9 +73,6 @@ func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
 		scheme:  mgr.GetScheme(),
 		applier: helper,
 	}
-	r.remoteClientBuilder = func(cd *hivev1.ClusterDeployment) remoteclient.Builder {
-		return remoteclient.NewBuilder(r.Client, cd, controllerName)
-	}
 
 	return r
 }
@@ -102,9 +99,8 @@ var _ reconcile.Reconciler = &ReconcileControlPlaneCerts{}
 // ReconcileControlPlaneCerts reconciles a ClusterDeployment object
 type ReconcileControlPlaneCerts struct {
 	client.Client
-	scheme              *runtime.Scheme
-	applier             applier
-	remoteClientBuilder func(*hivev1.ClusterDeployment) remoteclient.Builder
+	scheme  *runtime.Scheme
+	applier applier
 }
 
 // Reconcile reads that state of the cluster for a ClusterDeployment object and makes changes based on the state read
@@ -380,11 +376,9 @@ func (r *ReconcileControlPlaneCerts) setCertsNotFoundCondition(cd *hivev1.Cluste
 // defaultControlPlaneDomain will attempt to return the domain/hostname for the secondary API URL
 // for the cluster based on the contents of the clusterDeployment's adminKubeConfig secret.
 func (r *ReconcileControlPlaneCerts) defaultControlPlaneDomain(cd *hivev1.ClusterDeployment) (string, error) {
-	remoteClient := r.remoteClientBuilder(cd)
-
-	apiurl, err := remoteClient.UseSecondaryAPIURL().APIURL()
+	apiurl, err := remoteclient.InitialURL(r.Client, cd)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to fetch secondary API URL")
+		return "", errors.Wrap(err, "failed to fetch initial API URL")
 	}
 
 	u, err := url.Parse(apiurl)
--- a/pkg/controller/unreachable/unreachable_controller.go
+++ b/pkg/controller/unreachable/unreachable_controller.go
@@ -46,8 +46,7 @@ import (
 const (
 	controllerName = "unreachable"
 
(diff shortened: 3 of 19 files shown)


pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -380,11 +376,9 @@ func (r *ReconcileControlPlaneCerts) setCertsNotFoundCondition(cd *hivev1.Cluste
 // defaultControlPlaneDomain will attempt to return the domain/hostname for the secondary API URL
 // for the cluster based on the contents of the clusterDeployment's adminKubeConfig secret.
 func (r *ReconcileControlPlaneCerts) defaultControlPlaneDomain(cd *hivev1.ClusterDeployment) (string, error) {
-	remoteClient := r.remoteClientBuilder(cd)
-
-	apiurl, err := remoteClient.UseSecondaryAPIURL().APIURL()
+	apiurl, err := remoteclient.InitialURL(r.Client, cd)
```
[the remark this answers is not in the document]
dgoodwin: Ah I think I see, initial is secondary isn't it.

pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -145,88 +144,84 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
-		if !isTimeForConnectivityCheck(cond) {
-			cdLog.WithFields(log.Fields{
-				"sinceLastTransition": time.Since(cond.LastTransitionTime.Time),
-				"sinceLastProbe":      time.Since(cond.LastProbeTime.Time),
-			}).Debug("skipping unreachable check")
-			return reconcile.Result{}, nil
-		}
+	wasUnreachable, lastCheck := remoteclient.Unreachable(cd)
+	wasPrimaryActive := remoteclient.IsPrimaryURLActive(cd)
+	connectivityCheckDelay := maxUnreachableDuration - time.Since(lastCheck)
```
dgoodwin: Could you commend this connectivityCheckDelay variable, it seems to be present in the most confusing lines below. I was wondering why we didn't handle this being < 0 here but I see that being checked below.

pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -145,88 +144,84 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
+
+	if !wasUnreachable && wasPrimaryActive && connectivityCheckDelay > 0*time.Second {
```
dgoodwin: Comments needed on this if. Wondering why wasPrimaryActive is called out here, and connectivityCheckDelay as well.

pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -145,88 +144,84 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
+	updateUnreachable := true
+	_, primaryErr := remoteClientBuilder.UsePrimaryAPIURL().Build()
+	if primaryErr != nil {
+		if hasOverride(cd) {
 			cdLog.WithError(primaryErr).Info("unable to create remote API client using API URL override")
-			activeAPIURLOverrideError = primaryErr
-			if _, secondaryErr := remoteClientBuilder.UseSecondaryAPIURL().Build(); secondaryErr != nil {
-				cdLog.WithError(secondaryErr).Warn("unable to create remote API client with either the initial API URL or the API URL override, marking cluster unreachable")
-				unreachableError = utilerrors.NewAggregate([]error{primaryErr, secondaryErr})
+			if wasUnreachable || wasPrimaryActive || connectivityCheckDelay <= 0*time.Second {
```
dgoodwin: This if could also use a comment.

pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -145,88 +144,84 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
+				if _, secondaryErr := remoteClientBuilder.UseSecondaryAPIURL().Build(); secondaryErr != nil {
```
dgoodwin: Got confused again at this point, throughout the PR I've seen initial, override, primary, secondary. Can these be collapsed into just two consistent terms?

Looking further maybe not. Initial and override I understand. It looks like primary is whatever is supposed to be active, and changes once the override is available. Secondary must be whatever is non-primary at that point in time?

pkg/controller/clusterstate/clusterstate_controller_test.go
```diff
@@ -234,6 +233,12 @@ func testClusterDeployment() *hivev1.ClusterDeployment {
 			},
 			Installed: true,
 		},
+		Status: hivev1.ClusterDeploymentStatus{
+			Conditions: []hivev1.ClusterDeploymentCondition{{
+				Type:   hivev1.UnreachableCondition,
+				Status: corev1.ConditionFalse,
+			}},
+		},
```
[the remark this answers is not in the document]
staebler: No, it is no longer true that a cluster can have no unreachable condition and be considered reachable. The probe time on the unreachable condition is needed to determine when to check again for connectivity. If a clusterdeployment does not have the condition, then controllers will not attempt to connect to the cluster.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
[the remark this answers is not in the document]
staebler: Yes. I broke out the `APIURL` function from the `remoteclient.Builder` interface. Consequently, I was no longer beholden to use the terms Primary, Secondary, and Active. Initial API URL is a bit more clear about what the intention is because, when there is no override URL, it is a bit convoluted to say that the secondary is also the initial.

pkg/controller/unreachable/unreachable_controller.go
staebler: I think you've got it.
**Initial**: what comes from the installer in the kubeconfig
**Active**: the API URL that is currently being used to connect to the remote cluster
**Primary**: the preferred API URL. If there is an override, this is the override URL. If there is not an override, this is the initial URL.
**Secondary**: the fallback API URL. If there is an override, this is the initial URL.

staebler: There are comments describing what Primary and Secondary mean in the `remoteclient.Builder` interface.
https://github.com/openshift/hive/blob/e503130882dce2c6945762439283e9a62d0467f8/pkg/remoteclient/remoteclient.go#L47-L53

staebler:
> That's a doozy for logic.

I agree. I thought hard about how I could make the flow less complicated. I am **very** open to any ideas anyone may have. I have added comments that will hopefully make things slightly easier to follow.

staebler:
Squashed the commits.

Also, fixed the bad logic for `connectivityRecheckNeeded`.
```
-       connectivityRecheckNeeded := !wasUnreachable && connectivityRecheckDelay > 0*time.Second
+       connectivityRecheckNeeded := wasUnreachable || connectivityRecheckDelay <= 0*time.Second
```

PR #910, opened 2020-03-30
Add ownership reconciliation based on labels

twiest:
- [x] Given specific labels of objects controlled by hive, ensure that the
  controller owner of the object is correctly set.
  - [x] If an object's owner doesn't exist, add it.
  - [x] If an object's owner is set incorrectly, fix it.
  - [x] If an object's owner is set correctly, do nothing.
- [x] Manually test the above scenarios
- [x] Check that calls match up with documented unique labels
- [x] Add unit tests

Add ownership reconciliation based on labels

### DIFF
--- a/pkg/controller/utils/ownership.go
+++ b/pkg/controller/utils/ownership.go
@@ -1,0 +1,110 @@
+package utils
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+
+	librarygocontroller "github.com/openshift/library-go/pkg/controller"
+)
+
+// OwnershipUniqueKey contains the uniquly identifiable pattern for ensuring ownership labels are correct applied for a type.
+type OwnershipUniqueKey struct {
+	LabelSelector map[string]string
+	TypeToList    runtime.Object
+}
+
+// ReconcileOwnerReferences ensures that given owner is in fact the actual owner for all types in typesToList given a specific labelSelector
+func ReconcileOwnerReferences(owner hivev1.MetaRuntimeObject, ownershipKeys []*OwnershipUniqueKey, kubeclient client.Client, scheme *runtime.Scheme, logger log.FieldLogger) error {
+	errlist := []error{}
+
+	for _, ownershipKey := range ownershipKeys {
+		objects, err := ListRuntimeObjects(kubeclient, []runtime.Object{ownershipKey.TypeToList}, client.MatchingLabels(ownershipKey.LabelSelector))
+		if err != nil {
+			errlist = append(errlist, errors.Wrap(err, "failed listing objects owned by clusterdeployment according to label"))
+			continue
+		}
+
+		for _, object := range objects {
+			mrObject, ok := object.(hivev1.MetaRuntimeObject)
+			if !ok {
+				// This should never happen since all hive objects implement both the meta and runtime object interfaces.
+				logger.Warnf("Failed converting object to MetaRuntimeObject")
+				continue
+			}
+
+			err := SyncControllerReference(owner, mrObject, kubeclient, scheme, logger)
+			if err != nil {
+				errlist = append(errlist, err)
+			}
+		}
+	}
+
+	return utilerrors.NewAggregate(errlist)
+}
+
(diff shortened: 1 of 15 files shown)


pkg/controller/utils/ownership.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		object.SetOwnerReferences(newOwnerReferences)
+	}
+
+	// The owner reference wasn't found, so we need to add it.
+	if err := controllerutil.SetControllerReference(owner, object, scheme); err != nil {
+		objectLogger.WithError(err).Warn("could not set the controller owner reference")
+		return nil // Not returning error because that could stop the overall reconciliation. Just logging a warning.
+	}
```
staebler: Can I offer an alternative to the logic in this function that leverages some existing code in library-go?
```
	// Remove any other controller ref
	for i, ref := range object.GetOwnerReferences() {
		if ref.Controller != nil && *ref.Controller {
			if ref.UID != owner.GetUID() {
				ownerRefs := object.GetOwnerReferences()
				ownerRefs[i] = ownerRefs[len(ownerRefs)-1]
				object.SetOwnerReferences(ownerRefs[:len(ownerRefs)-1])
			}
			break
		}
	}

	ownerRef := metav1.NewControllerRef(owner, ownerGVK)
	ownerRefsChanged := librarygocontroller.EnsureOwnerRef(object, *ownerRef)
	
	if ownerRefsChanged {
		objectLogger.Debug("Object has correct ownership. No changes necessary.")
		return nil
	}
```

pkg/controller/utils/ownership.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+
+			err := SyncControllerReference(owner, mrObject, kubeclient, scheme, logger)
+			if err != nil {
+				errlist = append(errlist, err)
+			}
+		}
+	}
+
+	if len(errlist) != 0 {
```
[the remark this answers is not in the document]
twiest: Awesome! Will fix.

pkg/controller/utils/ownership.go
twiest: Absolutely! The only reason that I didn't use it because I didn't previously know about it and I didn't find it when looking.

I much prefer using built in api calls. :)

Will Fix.

twiest:
> What are your thoughts on adding watches to controllers for the owned resources? For example, if there is a change to the merged pull secret, then it would be nice if the clusterdeployment were resynced. This is probably not too important for an owner that is updated often like clusterdeployment, but may be more important from something like clusterprovision.

I like the idea of immediate reconciliation of owner references.

We previously shied away from watching secrets because it would balloon our cache. Not sure if we're watching secrets now or not, but I know we discussed and rejected the idea in the past.  @dgoodwin what are your thoughts on this?

At any rate, @staebler if we decide to do that, would you mind if we do that as a separate PR / card? I feel this PR / card is already late. :)

dgoodwin:
I don't think we watch Secrets today, at least not anywhere I can find in a quick grep, and there are 6-8k in some of our envs. Definitely something we'd have to tread lightly with.

staebler:
I am good with postponed any additional changes to add watches.

As for caching, the controller-runtime caches every type that you Get or List already. So even if we are not explicitly watching, there is an implicit watch that is already being done.

twiest:
> As for caching, the controller-runtime caches every type that you Get or List already. So even if we are not explicitly watching, there is an implicit watch that is already being done.

For posterity, we discussed this in stand up today and if we can confirm that we're already caching all secrets because of our Get calls on secrets, then the team is ok with watching secrets.

However there were multiple anecdotal experiences on the team where watching secrets caused the memory of other controllers to balloon and cause problems.

joelddiaz:
> > As for caching, the controller-runtime caches every type that you Get or List already. So even if we are not explicitly watching, there is an implicit watch that is already being done.
> 
> For posterity, we discussed this in stand up today and if we can confirm that we're already caching all secrets because of our Get calls on secrets, then the team is ok with watching secrets.
> 
> However there were multiple anecdotal experiences on the team where watching secrets caused the memory of other controllers to balloon and cause problems.

FWIW, I don't remember specifically that cloud-cred-operator needed its memory allowance increased b/c of *watching* secrets. It could very well be that the increased memory was just as much a result of doing Get()s on secrets.

twiest:
@staebler feedback addressed. PTAL

pkg/controller/utils/ownership_test.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// Arrange
+			fakeKubeClient := fake.NewFakeClient(test.existingObjects...)
+			logger := log.WithField("fake", "fake")
+			ownershipKey := test.ownershipUniqueKeys[0]
```
staebler, line 173: It may be better to have the test cases use a `ownershipUniqueKey *OwnershipUniqueKey` instead of a slice. Otherwise, there is an implicit requirement on the test cases that there be exactly one element in the slice.

staebler:
> Let me know if you feel strongly about this and I'll revisit it.

Nope. I don't feel strongly about it. Leave it as is.

Issue #911, opened 2020-03-31
Bug - Invalid CSV file name in olm-registry-deploy.sh


ahmedwaleedmalik:
Hi,

The `CSV_FILE` field generated in [olm-registry-deploy.sh](https://github.com/openshift/hive/blob/master/hack/olm-registry-deploy.sh#L35) is invalid. Since, the field `full_version` in [generate-operator-bundle.py](https://github.com/openshift/hive/blob/master/hack/generate-operator-bundle.py#L33) is generated based on the following pattern:

`full_version = "%s.%s-sha%s" % (VERSION_BASE, git_num_commits, git_hash)`

And this is used as a prefix in the rendered `clusterserviceversion.yaml` file.

Error logs:
```
Generating CSV for version: 0.1.1798-shae5031308
Wrote ClusterServiceVersion: bundle/0.1.1798-shae5031308/hive-operator.v0.1.1798-shae5031308.clusterserviceversion.yaml
17:11:16 main [ERRO] open bundle/0.1.1798-e5031308/hive-operator.v0.1.1798-e5031308.clusterserviceversion.yaml: no such file or directory
```

PR for fix: https://github.com/openshift/hive/pull/912

PR #912, opened 2020-03-31
Fix CSV_FILE name substitution in hack/olm-registry-deploy.sh

ahmedwaleedmalik:
Fixes issue: https://github.com/openshift/hive/issues/911


gregsheremeta:
Thanks. I recommend using https://operatorhub.io/operator/hive-operator

dgoodwin:
This change looks correct, I believe olm-registry-deploy is not used by anything in the pipelines just Hive devs, thus why this has lingered so long.

/lgtm

PR #913, opened 2020-03-31
add version information to log output

dhellmann:
Add LDFLAGS to inject version details for each build based on the git
tag and commit SHA.

Add a hiveutil version subcommand.

Log the version info at the startup of all other commands.

### DIFF
--- a/cmd/hiveadmission/main.go
+++ b/cmd/hiveadmission/main.go
@@ -10,9 +10,11 @@ import (
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	hivevalidatingwebhooks "github.com/openshift/hive/pkg/apis/hive/v1/validating-webhooks"
 	"github.com/openshift/hive/pkg/certs"
+	"github.com/openshift/hive/pkg/version"
 )
 
 func main() {
+	log.Infof("Version: %s @ %s", version.String, version.Commit)
 	log.Info("Starting CRD Validation Webhooks.")
 
 	// TODO: figure out a way to combine logrus and klog logging levels. The team has decided that hardcoding this is ok for now.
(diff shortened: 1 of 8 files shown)


dhellmann:
The hiveutil command output looks like:

```
$ ./bin/hiveutil version
INFO[0000] openshift/hive v1.0.0-17-gf54e4f68a86d0fffc44bc1b7586ea3f087025794 @ f54e4f68a86d0fffc44bc1b7586ea3f087025794
```

The log output for the other services should look similar.

cmd/hiveadmission/main.go
```diff
@@ -10,10 +10,12 @@ import (
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	hivevalidatingwebhooks "github.com/openshift/hive/pkg/apis/hive/v1/validating-webhooks"
 	"github.com/openshift/hive/pkg/certs"
+	"github.com/openshift/hive/pkg/version"
 )
 
 func main() {
 	log.Info("Starting CRD Validation Webhooks.")
+	log.Infof("Version: %s @ %s", version.String, version.Commit)
```
staebler: Should this be moved up a line? I would think that the version should be the first thing output.

dhellmann: sure, I can do that

gregsheremeta:
I wanted to see this in action so i checked hive-operators on hive-stage, and I see:
`time="2020-04-02T11:35:08Z" level=info msg="Version: openshift/hive was not built with version info @ "`
@dhellmann @staebler @dgoodwin am I missing something?

staebler:
> I wanted to see this in action so i checked `hive-operator` on hive-stage, and I see:
> `time="2020-04-02T11:35:08Z" level=info msg="Version: openshift/hive was not built with version info @ "`
> @dhellmann @staebler @dgoodwin am I missing something?

You are not missing anything. The binaries for the image are not built using the make targets changed in this PR.

https://github.com/openshift/hive/blob/f0cf32526fe737f3cf0e5dd6c2e1075ca9ae74d0/Dockerfile#L5-L9

PR #914, opened 2020-03-31
make log redaction tests more robust

dhellmann:
Instead of looking for individual substrings in the cleaned text,
verify that the entire return value matches the expected value. This
makes future tests easier to add and makes the existing tests easier
to read.

### DIFF
--- a/pkg/installmanager/installmanager_test.go
+++ b/pkg/installmanager/installmanager_test.go
@@ -3,13 +3,13 @@ package installmanager
 import (
 	"context"
 	"fmt"
-	"github.com/stretchr/testify/require"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


pkg/installmanager/installmanager_test.go
```diff
@@ -400,62 +399,63 @@ level=info msg="Install complete!"
...
-				"testing %v: unexpected string found after cleaning",
-				test.name, testString)
-		}
-
-		for _, testString := range test.expectedStrings {
-			assert.True(t, strings.Contains(cleanedString, testString),
-				"testing %v: expected string %q not found after cleaning: %q became %q",
-				test.name, testString, test.sourceString, cleanedString)
-		}
+		assert.True(t, cleanedString == test.expectedString,
```
staebler: Use `assert.Equal`. Then you do not need to explicitly add the expected and actual to the message.

```
assert.Equal(t, test.expectedString, cleanedString, "unexpected cleaned string")
```

dhellmann: Done

PR #919, opened 2020-04-02
Add syncset and selectorsyncset test scripts

lwan-wanglin:
I write some e2e test scripts that maintain main  syncsets and selectorsyncsets  parts func verifying. I find most e2e test codes are using instance to create resources. This way is a little hard for me because I am not familiar with this part,so I choice another way to finish it and I used the ginkgo frame. I don't know whether this is ok or we'd better have a uniform format.
Please help to review.

### DIFF
--- a/test/e2e/common/commonutils.go
+++ b/test/e2e/common/commonutils.go
@@ -1,0 +1,105 @@
+package common
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	defTmpDir = "/tmp"
+)
+
+//RunShellCmdWithEnv is a function to run a shell command with new os environament variable and special dir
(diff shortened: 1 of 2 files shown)


csrwng:
@lwan-wanglin The ginkgo part is fine. The framework itself looks good. However, it'd be preferable to use the go client to make API calls instead of calling out to the `oc` command. To do this, you could copy (or vendor) the client utility from the origin extended tests: https://github.com/openshift/origin/blob/master/test/extended/util/client.go
And perform actions via the client as in:
https://github.com/openshift/origin/blob/97738ea1cbf17b45410616690abb3e3a244eb253/test/extended/builds/build_pruning.go#L81

csrwng:
You would need to add utility functions to the client to create a Hive client. Please let me know if you need help doing this.

lwan-wanglin:
> You would need to add utility functions to the client to create a Hive client. Please let me know if you need help doing this.

@csrwng Thank you for your advice, I will try it by myself firstly, I will ask help for you if I meet the trouble.

PR #920, opened 2020-04-02
Configurable separate namespaces for hive-operator and hive-controllers etc.

dgoodwin:
This change removes the assumption that all hive components run in the "hive" namespace. hive-operator can now be run in it's own namespace, and deploy hive itself into another namespace as configured by HiveConfig.Spec.TargetNamespace. For now we assume you will not change target namespace, we cannot validate this easily as hiveadmission does not exist at this point. If you change the target namespace you'll get a parallel hive install with both controllers listening. *Do not change the target namespace.*

OLM and kustomize still have roughly the same responsibility, deploying the hive-operator. Previously RBAC for controllers/hiveadmission/hiveapi was munged into this process, this had to be removed as kustomize and OLM both are not good at multi-namespace. As such there is much less in kustomize.yaml.

hive-operator now deploys the ClusterRoles and ClusterRoleBindings for controllers/admission/hiveapi. This required additional permissions be added to the operator, and some duplication and simplification was done particularly on the controllers role.

hive-controllers rbac was renamed from "manager-role" and "manager-role-binding" to be hive-controllers specific. This could avoid a complication where another operator attempted to use controller-runtime default rbac names and collides with ours. Additionally a hive-controller's service account was added as we should never be using the default service account for operators.

hive-operator previously created a default HiveConfig for you if none exists, this is no longer the case as it was discovered to be non-standard behavior. make deploy will do this for you now.

hive-operator now has to inject the configured namespace into a variety of the assets it deploys.

e2e tests now make use of the most complex scenario, hive-operator in it's own separate ns, and hive itself is moved to a non-"hive" namespace just to ensure this functionality is and stays working, despite the defaults of "hive".


staebler:
> hive-operator now deploys the ClusterRoles and ClusterRoleBindings for controllers/admission/hiveapi. This required additional permissions be added to the operator, and some duplication and simplification was done particularly on the controllers role.

This seems like a **big** change. I thought that we explicitly wanted OLM to handle dishing out permissions rather than giving the permissions to the operator.

config/kustomization.yaml
```diff
@@ -18,20 +18,15 @@ namespace: hive
...
+# This kustomize configuration mimic's what is deployed by OLM, essentially *only* what is required
+# for the hive-operator.
 resources:
-- ./namespace.yaml
-- ./rbac/hive_controllers_role.yaml
-- ./rbac/hive_controllers_role_binding.yaml
-- ./rbac/hive_frontend_role.yaml
-- ./rbac/hive_frontend_role_binding.yaml
+- ./hiveadmission/service-account.yaml
 - ./hiveadmission/hiveadmission_rbac_role.yaml
```
staebler, line 25: Why are we OK laying down the RBAC for hiveadmission and hiveapi but not for the controllers?

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -7,6 +7,11 @@ import (
 
 // HiveConfigSpec defines the desired state of Hive
 type HiveConfigSpec struct {
+
+	// TargetNamespace is the namespace where the core Hive components should be run. Defaults to "hive". Will be
+	// created if it does not already exist.
+	TargetNamespace string `json:"targetNamespace"`
```
staebler, line 13: `omitempty` since this is optional

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1564,7 +1564,7 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
 	globalPullSecretName := os.Getenv(constants.GlobalPullSecret)
 	var globalPullSecret string
 	if len(globalPullSecretName) != 0 {
-		globalPullSecret, err = controllerutils.LoadSecretData(r.Client, globalPullSecretName, constants.HiveNamespace, corev1.DockerConfigJsonKey)
+		globalPullSecret, err = controllerutils.LoadSecretData(r.Client, globalPullSecretName, controllerutils.GetHiveNamespace(), corev1.DockerConfigJsonKey)
```
staebler, line 1567: Shouldn't the global pull secret by in the namespace of the hive-operator? It is part of the operand.

dgoodwin:
> > hive-operator now deploys the ClusterRoles and ClusterRoleBindings for controllers/admission/hiveapi. This required additional permissions be added to the operator, and some duplication and simplification was done particularly on the controllers role.
> 
> This seems like a **big** change. I thought that we explicitly wanted OLM to handle dishing out permissions rather than giving the permissions to the operator.

Can you think of another option? OLM can't cross namespaces, you specify your permissions and a service account and it assumes to configure and bind to a service account in the namespace you're installing your operator. This would mean that hive must be in the namespace hive-operator is installed, and subsequently that all OCM components must be in one namespace, as they are planning to munge our CSV into theirs for productized releases.

The hive-operator already had permission to manage clusterroles and clusterrolebindings: https://github.com/openshift/hive/blob/master/config/operator/operator_role.yaml#L116

This allows hive-operator to fully provision hive, I'm actually not crazy about the previous implementation where some things were OLM and some were hive-operator. Now OLM configures hive-operator, and hive-operator configures everything else, it's cleaner and clearer. (with exception of CRDs which are configured by both, but on newer OLM we can make a desicion which way to go there) 

Indeed hive-operator is very powerful, but this does not need to trickle down to hive-controllers etc.

In future if OLM allows multi-namespace, we could revisit and revoke these permissions. Today I can't think of another alternative aside from jamming everything into one NS, which does not seem like a good idea.

pkg/operator/hive/operatorutils.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return nil
+}
+
+func dynamicDelete(dynamicClient dynamic.Interface, gvr schema.GroupVersionResource, name string, hLog log.FieldLogger) error {
+	rLog := hLog.WithFields(log.Fields{
+		"gvr":  gvr,
+		"name": name,
+	})
+	if _, err := dynamicClient.Resource(gvr).Get(name, metav1.GetOptions{}); err == nil {
```
staebler, line 60: I don't understand the purpose of doing a Get first. Why not just ignore a not-found error returned from the Delete for the case where the object does not exist? Also, the current approach is ignoring all errors from the Get, when I would think that it should only be ignoring the not-found errors.

gregsheremeta:
> Today I can't think of another alternative aside from jamming everything into one NS, which does not seem like a good idea.

Are there any specific things you're worried about?

On one hand, it's appealing to have the NS configurable. On the other hand, it's appealing to let OLM be the god object.

dgoodwin:
> > Today I can't think of another alternative aside from jamming everything into one NS, which does not seem like a good idea.
> 
> Are there any specific things you're worried about?
> 
> On one hand, it's appealing to have the NS configurable. On the other hand, it's appealing to let OLM be the god object.

I'd be ok with it for hive and hive-operator, but not when we get to the OCM use case where every component lands in one namespace, because we're munging into one CSV.

config/kustomization.yaml
dgoodwin: Oversight, forgot these ones in there.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: The two sentences here seem to conflict, hive-controllers is the operand right? As such I believe this is the right place for it.

This secret is created by an admin, added to HiveConfig (alongside targetNamespace), and we assume it's in targetNamespace. (right now) I believe it's only used by hive-controllers, and I don't think hive-controllers should need to be passed the operator's namespace, or read anything from it, it should be largely unaware of the operator's existence. Feels better here than the alternative.

pkg/operator/hive/operatorutils.go
dgoodwin: Good point will update.

dgoodwin:
Most feedback addressed, but we still need to settle what to do about who deploys hive operand RBAC. 

I believe that leaving this to OLM implies that all ACM components must land in the same namespace, it can't deploy RBAC for a service account other than where we're deploying hive-operator. I personally do not want to be called in to debug OCM/ACM problems where there are dozens of pods in one namespace all with access to the same secrets and configmaps. It feels like bad practice to stuff multiple independent components into one namespace.

This permission to grant RBAC is not new, hive-operator already had it as it's used to deploy the hive-admin/reader roles for us in prod/stage.

But more than that I want to try to make the case that this is preferable in general. We chose to have our own hive-operator layer to deploy, upgrade and orchestrate hive itself. Contrast with a world where hive-controllers+hiveadmission was deployed directly by OLM (hive is after all an operator in itself), and OLM deploying it might make sense someday, but that is not the design we chose and currently have today. 

With what we have today I prefer that OLM deploys hive-operator, and hive-operator deploys hive, and we stick to this as much as possible. Having hive-operator control the RBAC for hive then is cleaner and more logical than having the responsibilities cross. hive-operator has powerful permissions but that's necessary for it's job.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: Sorry, I used the wrong terminology. I was thinking that the HiveConfig was the operand. My thought was that the admin would not be creating or touching the targetNamespace namespace at all. All of the admin configuration is done relative to HiveConfig and the hive-operator. The hive-operator is the one that creates the targetNamespace namespace. Is the expectation that the admin would be create that namespace instead? Or is the expectation that the admin would be create the global-pull-secret secret after the hive-operator create that namespace?

dgoodwin: The hive-operator is the one that can create the target namespace if it doesn't exist, but the admin can pre-create and add their global pull secret there prior to creating HiveConfig if desired, or update HiveConfig post-install.

staebler: OK. I am fine with this.

dgoodwin:
E2E was just clear, I've squashed everything and pushed.

I've successfully installed these changes with OLM all to the hive namespace. I have not been able to test a 1.0.0 upgrade yet because of OLM concepts I'm still learning, regardless I think an install is a good indicator this is safe enough to go to stage.

arewm:
nudge on this PR so that we can get `ocm-4.4.0` updated. 😄

config/templates/hiveconfig.yaml
```diff
@@ -0,0 +1,21 @@
...
+    description: Namespace where core hive components will run.
+    value: hive
+
+objects:
+  - apiVersion: hive.openshift.io/v1
+    kind: HiveConfig
+    metadata:
+      name: hive
+    spec:
+      logLevel: debug
```
staebler, line 18: Do we really want the default here to be debug?

config/templates/hiveconfig.yaml
```diff
@@ -0,0 +1,21 @@
...
+      targetNamespace: ${HIVE_NS}
+      deprovisionsDisabled: false
+      maintenanceMode: false
```
staebler, line 21: These can be omitted.

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -7,6 +7,12 @@ import (
 
 // HiveConfigSpec defines the desired state of Hive
 type HiveConfigSpec struct {
+
+	// TargetNamespace is the namespace where the core Hive components should be run. Defaults to "hive". Will be
+	// created if it does not already exist.
+	// +optional
+	TargetNamespace string `json:"targetNamespace,omitempty"`
```
staebler, line 14: Fix up the comments for the reference fields in HiveConfig so that they explicitly state that the resources should be in the target namespace. For example, the comment for the `AdditionalCertificateAuthoritiesSecretRef` field states that the secret should be in the hive namespace. Alternatively (or additionally), add a comment to the `TargetNamespace` field to state that all resources referenced in the HiveConfig should be in the target namespace.

pkg/controller/clusterprovision/installlogmonitor.go
```diff
@@ -2,6 +2,7 @@ package clusterprovision
 
 import (
 	"context"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
```
staebler, line 5: This import should be in its own section at the end of the import list.

pkg/operator/hive/hive.go
```diff
@@ -5,6 +5,8 @@ import (
 	"context"
 	"crypto/md5"
 	"fmt"
+	"k8s.io/apimachinery/pkg/runtime"
```
staebler, line 8: These imports should be in a section between the section with logrus and the section with hive.

pkg/operator/hive/hive.go
```diff
@@ -143,37 +154,45 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
 	}
 
-	result, err := h.ApplyRuntimeObject(hiveDeployment, scheme.Scheme)
-	if err != nil {
-		hLog.WithError(err).Error("error applying deployment")
-		return err
+	// Load namespaced assets, decode them, set to our target namespace, and apply:
+	coreCodecs := serializer.NewCodecFactory(scheme.Scheme)
+	coreDecoder := coreCodecs.UniversalDecoder(corev1.SchemeGroupVersion)
+	namespacedAssets := map[string]runtime.Decoder{
```
staebler, line 160: What is the purpose of having a map with a decoder for the value when every entry in the map has the same decoder?

pkg/operator/hive/hive_controller.go
```diff
@@ -254,11 +245,28 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+	recorder := events.NewRecorder(r.kubeClient.CoreV1().Events(r.hiveOperatorNamespace), "hive-operator", &corev1.ObjectReference{
 		Name:      request.Name,
-		Namespace: constants.HiveNamespace,
+		Namespace: r.hiveOperatorNamespace,
 	})
 
+	// Ensure the target namespace for hive components exists and create if not:
+	hiveNSName := getHiveNamespace(instance)
+	hiveNamespace := &corev1.Namespace{}
+	if err := r.Client.Get(context.Background(), types.NamespacedName{Name: hiveNSName}, hiveNamespace); err != nil {
```
staebler, line 256: Since you do not need to use the `corev1.Namespace` object, don't bother doing a Get first. Create the namespace and handle the error condition when the namespace already exists.

pkg/operator/hive/hiveadmission.go
```diff
@@ -3,27 +3,22 @@ package hive
 import (
 	"context"
 	"fmt"
-
 	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
```
staebler, line 7: Put this import with the rest of the k8s.io imports.

pkg/operator/hive/hiveadmission.go
```diff
@@ -84,11 +99,7 @@ func (r *ReconcileHiveConfig) deployHiveAdmission(hLog log.FieldLogger, h *resou
 		hLog.WithError(err).Error("error applying deployment")
 		return err
 	}
-	hLog.Infof("deployment applied (%s)", result)
-
-	hLog.Debug("reading apiservice")
-	asset = assets.MustAsset("config/hiveadmission/apiservice.yaml")
-	apiService := util.ReadAPIServiceV1Beta1OrDie(asset, scheme.Scheme)
+	hLog.Infof("hiveadmission deployment applied (%s)", result)
```
staebler, line 102: Put the result in a field.

pkg/operator/hive/hiveapi.go
```diff
@@ -2,6 +2,7 @@ package hive
 
 import (
 	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
```
staebler, line 5: Move import.

pkg/operator/hive/hiveadmission.go
```diff
@@ -35,33 +30,53 @@ const (
...
+	hiveNSName := getHiveNamespace(instance)
+
+	// Load namespaced assets, decode them, set to our target namespace, and apply:
+	coreCodecs := serializer.NewCodecFactory(scheme.Scheme)
+	coreDecoder := coreCodecs.UniversalDecoder(corev1.SchemeGroupVersion)
+	namespacedAssets := map[string]runtime.Decoder{
+		"config/hiveadmission/service.yaml":         coreDecoder,
+		"config/hiveadmission/service-account.yaml": coreDecoder,
+	}
+	for assetPath, decoder := range namespacedAssets {
```
staebler, line 46: Decoder in map?

pkg/operator/hive/hiveapi.go
```diff
@@ -24,33 +24,65 @@ import (
...
-		return r.tearDownHiveAPI(hLog)
+		return r.tearDownHiveAPI(hLog, hiveNSName)
 	}
 
-	if err := util.ApplyAsset(h, "config/apiserver/service.yaml", hLog); err != nil {
-		return err
+	// Load namespaced assets, decode them, set to our target namespace, and apply:
+	coreCodecs := serializer.NewCodecFactory(scheme.Scheme)
+	coreDecoder := coreCodecs.UniversalDecoder(corev1.SchemeGroupVersion)
+	namespacedAssets := map[string]runtime.Decoder{
```
staebler, line 36: Decoder in map?

pkg/operator/hive/hive.go
```diff
@@ -143,37 +154,45 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		hiveDeployment.Spec.Replicas = &replicas
 	}
 
-	result, err := h.ApplyRuntimeObject(hiveDeployment, scheme.Scheme)
-	if err != nil {
-		hLog.WithError(err).Error("error applying deployment")
-		return err
+	// Load namespaced assets, decode them, set to our target namespace, and apply:
+	coreCodecs := serializer.NewCodecFactory(scheme.Scheme)
+	coreDecoder := coreCodecs.UniversalDecoder(corev1.SchemeGroupVersion)
```
staebler, line 159: I don't think that a decoder is what you want to use. I think you want to use a deserializer instead. See https://github.com/kubernetes/apimachinery/blob/591a38b7a7b7e4f327f8bdbb32dfb8f0247e411e/pkg/runtime/serializer/codec_factory.go#L253-L255.

test/e2e/common/utils.go
```diff
@@ -0,0 +1,24 @@
...
+func GetHiveNamespaceOrDie() string {
+	hiveNS := os.Getenv(constants.HiveNamespaceEnvVar)
+	if len(hiveNS) == 0 {
+		log.Fatalf("required environment variable is not defined: %s", constants.HiveNamespaceEnvVar)
+	}
+	return hiveNS
+}
+
+func GetHiveOperatorNamespaceOrDie() string {
+	hiveNS := os.Getenv(hive.HiveOperatorNamespaceEnvVar)
```
staebler, line 19: `hiveOperatorNS` or just `namespace`.

hack/generate-operator-bundle.py
```diff
@@ -61,33 +61,6 @@
             'serviceAccountName': 'hive-operator',
         })
 
-# Add our hive-controllers role to the CSV:
-with open('config/rbac/hive_controllers_role.yaml', 'r') as stream:
```
staebler, line 65: Does anything need to be done to clean up the old controller rules for the default service account? Or will that all be handle by OLM?

config/templates/hiveconfig.yaml
dgoodwin: For make deploy I could go either way, I'll flip to info.

pkg/operator/hive/hive.go
dgoodwin: When implementing, I wasn't aware it was all going to be core resources, and wanted to keep it flexible to be able to apply other things if necessary. I can't think of anything upcoming so will adjust this to not pass the decoder.

config/templates/hiveconfig.yaml
staebler: Would it make sense to make it configurable and use the `RUN_LOG_LEVEL` variable in the make deploy call?

pkg/operator/hive/hive.go
staebler: I think that you may be incorrectly assuming that the parameter to `UniversalDecoder` does. It does not determine which apis can be decoded. Instead, it tells the decoder to use specific versions instead of the internal version. Multiple api versions can be passed to the function. If there were types in another API group, you could still use the same decoder and pass all of the api versions that you were interested in to the function. With that said, I still think that you want the deserializer instead.

hack/generate-operator-bundle.py
dgoodwin: Should be cleaned out by OLM yes.

config/templates/hiveconfig.yaml
dgoodwin: Sounds handy, will add.

pkg/operator/hive/hive.go
dgoodwin: That is good to know and definitely not how I thought that worked.

Makefile
```diff
@@ -121,11 +130,14 @@ install: crd
...
 	# Deploy the operator manifests:
+	oc create namespace ${HIVE_OPERATOR_NS} || true
 	mkdir -p overlays/deploy
 	cp overlays/template/kustomization.yaml overlays/deploy
-	cd overlays/deploy && kustomize edit set image registry.svc.ci.openshift.org/openshift/hive-v4.0:hive=${DEPLOY_IMAGE}
+	cd overlays/deploy && kustomize edit set image registry.svc.ci.openshift.org/openshift/hive-v4.0:hive=${DEPLOY_IMAGE} && kustomize edit set namespace ${HIVE_OPERATOR_NS}
 	kustomize build overlays/deploy | oc apply -f -
 	rm -rf overlays/deploy
+	# Create a default basic HiveConfig so the operator will deploy Hive
+	oc process --local=true -p HIVE_NS=${HIVE_NS} -p LOG_LEVEL=${LOG_LEVEL} -f config/templates/hiveconfig.yaml | oc apply -f -
```
staebler, line 140: This still makes the debug the default log level. I can live with this for now. I don't know who else uses make deploy outside of OpenShift development.

pkg/operator/hive/hive_controller.go
```diff
@@ -5,12 +5,13 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
staebler, line 8: Import grouping.

pkg/operator/hive/hive_controller.go
```diff
@@ -254,11 +246,30 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+	if err := r.Client.Create(context.Background(), hiveNamespace); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			hLog.WithField("hiveNS", hiveNSName).Debug("target namespace already exists")
+		} else {
+			hLog.WithError(err).Error("error creating hive target namespace")
+			return reconcile.Result{}, err
+		}
+	} else {
+		hLog.WithField("hiveNS", hiveNSName).Info("target namespace created")
+		hiveNamespace.ObjectMeta.Name = hiveNSName
```
staebler, line 270: This line is unnecessary.

PR #922, opened 2020-04-02
update dockerfile to use make to build binaries

dhellmann:
The Dockerfile was using go build directly, which meant the version
information was not being injected in the way expected. Switching to
use make required adding a new "binaries" target that does not include
the "generate" dependency because the builder image does not include
mockgen. The old "build" target is left in place with a dependency on
generate and binaries to provide backwards compatibility.


dgoodwin:
Needs a rebase but LGTM.

dhellmann:
Rebased

dgoodwin:
/lgtm

Thanks once again Doug!

PR #927, opened 2020-04-04
Add an unofficial MachinePool AMI override annotation.

dgoodwin:
For very limited use case and thus not a part of our official API at
this time.

/cc @jeremyeder @derekwaynecarr

### DIFF
--- a/pkg/controller/remotemachineset/awsactuator.go
+++ b/pkg/controller/remotemachineset/awsactuator.go
@@ -33,16 +33,21 @@ type AWSActuator struct {
 var _ Actuator = &AWSActuator{}
 
 // NewAWSActuator is the constructor for building a AWSActuator
-func NewAWSActuator(awsCreds *corev1.Secret, region string, remoteMachineSets []machineapi.MachineSet, scheme *runtime.Scheme, logger log.FieldLogger) (*AWSActuator, error) {
+func NewAWSActuator(awsCreds *corev1.Secret, region string, pool *hivev1.MachinePool, remoteMachineSets []machineapi.MachineSet, scheme *runtime.Scheme, logger log.FieldLogger) (*AWSActuator, error) {
 	awsClient, err := awsclient.NewClientFromSecret(awsCreds, region)
 	if err != nil {
 		logger.WithError(err).Warn("failed to create AWS client")
 		return nil, err
 	}
-	amiID, err := getAWSAMIID(remoteMachineSets, scheme, logger)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


dgoodwin:
This change may make no sense, we read the AMI ID from what exists already, so if this is done in the installconfig and applied in cluster, we would inherit the override the user originally specified.

/hold

dgoodwin:
/unhold

After speaking quickly with Derek this change is probably still worth reviewing and merging. There may be issues in scale to 0 scenarios where the worker MachineSet no longer exists, and we'd then lose the AMI to use. We'll break in this case with current code, but none the less this change is a good override for that scenario.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -33,16 +33,22 @@ type AWSActuator struct {
...
 	if err != nil {
 		logger.WithError(err).Warn("failed to create AWS client")
 		return nil, err
 	}
-	amiID, err := getAWSAMIID(remoteMachineSets, scheme, logger)
-	if err != nil {
-		logger.WithError(err).Warn("failed to get AMI ID")
-		return nil, err
+	var amiID string
+	if pool.Annotations != nil && pool.Annotations[hivev1.MachinePoolImageIDOverrideAnnotation] != "" {
```
staebler, line 43: The check for nil is not necessary.

staebler: Then, this block can become the following.
```
amiID := pool.Annotations[hivev1.MachinePoolImageIDOverrideAnnotation]
if amiID != "" {
} else {
}
```

twiest: For my knowledge, why is the check for nil unnecessary?

twiest: Never mind, should have googled first: 

"Map types are reference types, like pointers or slices, and so the value of m above is nil; it doesn't point to an initialized map. A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic; don't do that. To initialize a map, use the built in make function:"

https://blog.golang.org/maps

dgoodwin: Aha so it's just writes that are a problem, good to know.

Updated.

dgoodwin:
Updated, needs another lgtm, sorry @twiest !

twiest:
No problem at all! :)

/lgtm

PR #929, opened 2020-04-06
controller: add jitter to syncset re-apply

staebler:
Add some jitter to the time when a syncset will be requeued in order to be re-applied. The intention is to spread out the re-apply of syncsets so that there are not spikes in re-apply activity.

https://issues.redhat.com/browse/CO-871

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -523,32 +526,47 @@ func findSyncStatus(status hivev1.SyncStatus, statusList []hivev1.SyncStatus) *h
...
+	}
+	applySuccessCondition := controllerutils.FindSyncCondition(status.Conditions, hivev1.ApplySuccessSyncCondition)
+	if applySuccessCondition == nil {
+		return time.Now()
+	}
+	return applySuccessCondition.LastProbeTime.Add(r.reapplyInterval)
+}
+
+func (r *ReconcileSyncSetInstance) setReapplyTime(status *hivev1.SyncStatus) {
+	wait := time.Duration((1 + 0.1 * rand.Float64()) * r.reapplyInterval.Seconds()) * time.Second
```
csrwng, line 562: Is there a reason this jitter would need to be done per item in the syncset? (and stored in the status?) Wouldn't it be simpler to apply a jitter value to the requeue time? (can still be a multiple of reapplyInterval)

csrwng: sgtm!

staebler: The code has been simplified to only jitter the requeue time.

PR #931, opened 2020-04-06
generate-operator-bundle.py improvements

gregsheremeta:
generate-operator-bundle.py improvements

- add argparsing
- add mode for operatorhub bundle generation
- refactor current mode for OpenShift Dedicated
- update some data in the CSV template
- change app-sre pipeline scripts to work with new generate-operator-bundle.py
  options and the new hive versioning scheme. This removes the old commit
  counting. It also removed the deletion of future commits from
  the bundle, because hive can never be rolled back, only forward.

/cc @dgoodwin

### DIFF
--- a/config/templates/hive-csv-template.yaml
+++ b/config/templates/hive-csv-template.yaml
@@ -4,23 +4,84 @@ metadata:
   name: hive-operator-0.0.1
   namespace: placeholder
   annotations:
-    categories: A list of comma separated categories that your operator falls under.
+    capabilities: Seamless Upgrades
+    categories: OpenShift Optional
     certified: "false"
     description: OpenShift cluster provisioning and management at scale.
     containerImage: quay.io/dgoodwin/hive:latest
     createdAt: "2019-02-25T14:29:00Z"
-    support: Devan Goodwin
+    repository: https://github.com/openshift/hive
+    support: OpenShift Hive Team
+    alm-examples: |-
+      [{"apiVersion":"hive.openshift.io/v1","kind":"HiveConfig","metadata":{"name":"hive"},"spec":{"managedDomains":[{"aws":{"credentialsSecretRef":{"name":"my-route53-creds"}},"domains":["my-base-domain.example.com"]}]}}]
 spec:
-  displayName: Hive
-  description: OpenShift cluster provisioning and management at scale.
+  displayName: OpenShift Hive
+  description: |-
+    OpenShift Hive is an operator that runs on top of Kubernetes/OpenShift. Hive can be used to provision
+    and perform initial configuration of OpenShift clusters.
+
+    For provisioning OpenShift, Hive uses the [OpenShift installer](https://github.com/openshift/installer).
+
+    ### Supported cloud providers
+    * AWS
+    * Azure
+    * Google Cloud Platform
+
+    In the future Hive will support more cloud providers.
+
+    ## Documentation
+
+    * [Quick Start Guide](https://github.com/openshift/hive/blob/master/docs/quick_start.md)
+    * [Using Hive](https://github.com/openshift/hive/blob/master/docs/using-hive.md)
+    * [Hiveutil CLI](https://github.com/openshift/hive/blob/master/docs/hiveutil.md)
+    * [Frequently Asked Questions](https://github.com/openshift/hive/blob/master/docs/FAQs.md)
+    * [Architecture](https://github.com/openshift/hive/blob/master/docs/architecture.md)
+
+    See the [project README](https://github.com/openshift/hive#documentation) for more documentation.
+
+    ## Post Install Configuration
+
+    After installing the OpenShift Hive operator, create a cluster-scoped `HiveConfig` CR to configure Hive.
+    Upon creation of `HiveConfig`, the operator will configure Hive's pods.
+
+    Example `HiveConfig`:
+    ```yaml
+    ---
+      apiVersion: hive.openshift.io/v1
+      kind: HiveConfig
+      metadata:
+        name: hive
+      spec:
+        managedDomains:
+        - aws:
+            credentialsSecretRef:
+              name: my-route53-creds
+          domains:
+          - my-base-domain.example.com
+    ```
+
+    ## Create a cluster
+
+    To create a cluster with Hive, create a `ClusterDeployment` CR. You can also use the
+    [`hiveutil` tool](https://github.com/openshift/hive/blob/master/docs/hiveutil.md)'s `create-cluster` command
+    to create clusters.
   keywords:
   - kubernetes
   - openshift
   - multi-cluster
   - cluster
+  links:
+  - name: OpenShift Hive GitHub
+    url: https://github.com/openshift/hive
+  - name: "OpenShift Hive: Cluster-as-a-Service"
+    url: https://www.openshift.com/blog/openshift-hive-cluster-as-a-service
+  - name: OpenShift
+    url: https://www.openshift.com/
+  maintainers:
+    - name: OpenShift Hive Team
   version: 0.0.1
   provider:
-    name: Red Hat, Inc
+    name: Red Hat
   maturity: alpha
   installModes:
   - type: OwnNamespace
@@ -28,9 +89,9 @@ spec:
   - type: SingleNamespace
     supported: true
   - type: MultiNamespace
-    supported: false
+    supported: true
   - type: AllNamespaces
-    supported: false
+    supported: true
   install:
     strategy: deployment
     spec:
@@ -46,14 +107,4 @@ spec:
       displayName: Hive Config
       kind: HiveConfig
       name: hiveconfigs.hive.openshift.io
-      version: v1
-    - description: Defines an OpenShift cluster to be created.
-      displayName: Cluster Deployment
-      kind: ClusterDeployment
-      name: clusterdeployments.hive.openshift.io
-      version: v1
-    - description: Defines a DNSZone to be managed in a cloud provider.
-      displayName: DNS Zone
-      kind: DNSZone
-      name: dnszones.hive.openshift.io
       version: v1
--- a/hack/app_sre_create_image_catalog.sh
+++ b/hack/app_sre_create_image_catalog.sh
@@ -6,11 +6,10 @@ BRANCH_CHANNEL="$1"
 QUAY_IMAGE="$2"
 
 GIT_HASH=`git rev-parse --short=7 HEAD`
-GIT_COMMIT_COUNT=`git rev-list 9c56c62c6d0180c27e1cc9cf195f4bbfd7a617dd..HEAD --count`
 
-# clone bundle repo
 SAAS_OPERATOR_DIR="saas-hive-operator-bundle"
-BUNDLE_DIR="$SAAS_OPERATOR_DIR/hive/"
+SAAS_OPERATOR_BUNDLE_DIR="$SAAS_OPERATOR_DIR/hive/"
+HIVE_BUNDLE_DIR="bundle/"
 
 rm -rf "$SAAS_OPERATOR_DIR"
 
@@ -19,67 +18,33 @@ git clone \
     https://app:${APP_SRE_BOT_PUSH_TOKEN}@gitlab.cee.redhat.com/service/saas-hive-operator-bundle.git \
     $SAAS_OPERATOR_DIR
 
-# remove any versions more recent than deployed hash
-REMOVED_VERSIONS=""
-if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
-    DEPLOYED_HASH=$(
-        curl -s 'https://gitlab.cee.redhat.com/service/saas-hive/raw/master/hive-services/hive.yaml' | \
-            docker run --rm -i evns/yq -r '.services[]|select(.name="hive").hash'
-    )
-
-    delete=false
-    for version in `ls $BUNDLE_DIR | sort -t . -k 3 -g`; do
-        # skip if not directory
-        [ -d "$BUNDLE_DIR/$version" ] || continue
-
-        if [[ "$delete" == false ]]; then
-            short_hash=$(echo $version | cut -d- -f2)
-            short_hash=${short_hash##sha}
-
-            if [[ "$DEPLOYED_HASH" == "${short_hash}"* ]]; then
-                delete=true
-            fi
-        else
-            rm -rf "$BUNDLE_DIR/$version"
-            REMOVED_VERSIONS="$version $REMOVED_VERSIONS"
-        fi
-    done
-fi
 
 # generate bundle
-PREV_VERSION=$(ls $BUNDLE_DIR | sort -t . -k 3 -g | tail -n 1)
+PREV_VERSION=$((cd $SAAS_OPERATOR_BUNDLE_DIR && find -type d | xargs -n 1 basename) | sort -V  | tail -n 1)
 
-./hack/generate-operator-bundle.py \
-    $BUNDLE_DIR \
-    $PREV_VERSION \
-    $GIT_COMMIT_COUNT \
-    $GIT_HASH \
-    $QUAY_IMAGE:$GIT_HASH
+./hack/generate-operator-bundle.py osd --previous-version $PREV_VERSION --hive-image $QUAY_IMAGE:$GIT_HASH --channel $BRANCH_CHANNEL
 
-NEW_VERSION=$(ls $BUNDLE_DIR | sort -t . -k 3 -g | tail -n 1)
+NEW_VERSION=$((cd $HIVE_BUNDLE_DIR && find -type d | xargs -n 1 basename) | sort -V  | tail -n 1)
 
 if [ "$NEW_VERSION" = "$PREV_VERSION" ]; then
     # stopping script as that version was already built, so no need to rebuild it
     exit 0
 fi
 
-# create package yaml
-cat <<EOF > $BUNDLE_DIR/hive.package.yaml
-packageName: hive-operator
-channels:
-- name: $BRANCH_CHANNEL
-  currentCSV: hive-operator.v${NEW_VERSION}
-EOF
+# copy new csv
+cp -ax $HIVE_BUNDLE_DIR/$NEW_VERSION $SAAS_OPERATOR_BUNDLE_DIR/
+
+# copy package
+cp $HIVE_BUNDLE_DIR/hive.package.yaml $SAAS_OPERATOR_BUNDLE_DIR/
 
 # add, commit & push
 pushd $SAAS_OPERATOR_DIR
 
 git add .
 
-MESSAGE="add version $GIT_COMMIT_COUNT-$GIT_HASH
+MESSAGE="add version $NEW_VERSION
 
-replaces $PREV_VERSION
-removed versions: $REMOVED_VERSIONS"
+replaces $PREV_VERSION"
 
 git commit -m "$MESSAGE"
 git push origin "$BRANCH_CHANNEL"
--- a/hack/generate-operator-bundle.py
+++ b/hack/generate-operator-bundle.py
@@ -4,88 +4,170 @@
 # into a directory, and composes the ClusterServiceVersion which needs bits and
 # pieces of our rbac and deployment files.
 #
-# Usage ./hack/generate-operator-bundle.py OUTPUT_DIR PREVIOUS_VERSION GIT_NUM_COMMITS GIT_HASH HIVE_IMAGE
+# Usage: hack/generate-operator-bundle.py (osd|operatorhub)
 #
-# Commit count can be obtained with: git rev-list 9c56c62c6d0180c27e1cc9cf195f4bbfd7a617dd..HEAD --count
-# This is the first hive commit, if we tag a release we can then switch to using that tag and bump the base version.
+#  See help for options.
+#
 
+import argparse
 import datetime
 import os
-import sys
-import yaml
 import shutil
+import subprocess
+import yaml
 
-# This script will append the current number of commits given as an arg
-# (presumably since some past base tag), and the git hash arg for a final
-# version like: 0.1.189-3f73a592
-VERSION_BASE = "0.1"
 
-if len(sys.argv) != 6:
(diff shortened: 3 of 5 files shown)


hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
-csv['metadata']['annotations']['createdAt'] = now.strftime("%Y-%m-%dT%H:%M:%SZ")
-
-# Write the CSV to disk:
-csv_filename = "hive-operator.v%s.clusterserviceversion.yaml" % full_version
-csv_file = os.path.join(version_dir, csv_filename)
-with open(csv_file, 'w') as outfile:
-    yaml.dump(csv, outfile, default_flow_style=False)
-print("Wrote ClusterServiceVersion: %s" % csv_file)
 
+logging.basicConfig()
```
gregsheremeta, line 24: not used, remove

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+#        hack/generate-operator-bundle.py generate-operatorhub
+#
+#  optional:
+#   --hive-image IMAGE
 #
-# Commit count can be obtained with: git rev-list 9c56c62c6d0180c27e1cc9cf195f4bbfd7a617dd..HEAD --count
-# This is the first hive commit, if we tag a release we can then switch to using that tag and bump the base version.
 
+import argparse
+import logging
```
gregsheremeta, line 15: not used. clean these up

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+    # bump the zstream
+    previous_z = prev_version[-1]
+    previous_z_int = int(previous_z)
+    new_z_int = previous_z_int + 1
+    new_version = "%s%s" % (prev_version[:-1], new_z_int)
+    return new_version
+
+
+def generate_osd_version(args=None):
+    result = subprocess.check_output("git describe", shell=True).strip()
```
dgoodwin, line 50: Damn, that's handy. Was not aware of this command.

Will this roll out safely as an upgrade? Considering they're on 0.1.x now it must be ok as we're going to 1.x.x.

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+def current_operatorhub_version(args=None):
+    result = subprocess.check_output("git describe --abbrev=0", shell=True).strip()
+    # remove the v -- OLM bundles need to start with the digit
+    result = result[1:]
+    return result
+
+
+def generate_operatorhub_version(args=None, prev_version=current_operatorhub_version()):
+    # bump the zstream
+    previous_z = prev_version[-1]
```
dgoodwin, line 42: This looks like it'll break when we go from 1.0.9 to 1.0.10.

dgoodwin: I think we'll eventually want a cli arg to override, but we can cross that bridge when we want to go to 2.x or something.

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+    previous_z_int = int(previous_z)
```
dgoodwin, line 43: suggest simplifying to something like int(prev_version.split('.')[-1]

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+
+parser = argparse.ArgumentParser(
+    prog='OpenShift Hive Operator CSV generator',
+    description='created CSV and related artifacts for operatorhub.io and OpenShift Dedicated.',
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+)
+
+subparsers = parser.add_subparsers()
+
+osd_parser = subparsers.add_parser('generate-osd', help='Generate CSV for OpenShift Dedicated deployment.')
```
dgoodwin, line 192: Stutters with the script name, maybe just generate-operator-bundle.py osd or operatorhub

config/templates/hive-csv-template.yaml
```diff
@@ -28,9 +28,9 @@ spec:
   - type: SingleNamespace
     supported: true
   - type: MultiNamespace
-    supported: false
+    supported: true
```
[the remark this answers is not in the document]
gregsheremeta: I'll PM you a slack thread that shows why I think we might be true for all.

config/templates/hive-csv-template.yaml
```diff
@@ -28,9 +28,9 @@ spec:
...
   - type: AllNamespaces
-    supported: false
+    supported: true
```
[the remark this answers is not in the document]
gregsheremeta: I'll PM you a slack thread that shows why I think we might be true for all.

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+    with open(csv_file, 'w') as outfile:
+        yaml.dump(csv, outfile, default_flow_style=False)
+    print("Wrote ClusterServiceVersion: %s" % csv_file)
+
+
+def update_package(version):
+    PACKAGE_FILE = 'bundle/hive.package.yaml'
+
+    with open(PACKAGE_FILE, 'r') as stream:
+        package = yaml.load(stream, Loader=yaml.SafeLoader)
```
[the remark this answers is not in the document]
gregsheremeta: generate it. done.

hack/generate-operator-bundle.py
gregsheremeta: great point. done.

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+
+subparsers = parser.add_subparsers()
+
+osd_parser = subparsers.add_parser('generate-osd', help='Generate CSV for OpenShift Dedicated deployment.')
+osd_parser.add_argument('--previous-version', dest='prev_version', metavar='VERSION', help='Previous version, the version being replaced')
+osd_parser.add_argument('--hive-image', dest='hive_image', metavar='IMAGE', help='The hive image to deploy', default=OSD_HIVE_IMAGE_DEFAULT)
+osd_parser.set_defaults(func=generate_csv_osd)
+
+operatorhub_parser = subparsers.add_parser('generate-operatorhub', help='Generate CSV for operatorhub deployment.')
+operatorhub_parser.add_argument('--hive_image', dest='hive_image', metavar='IMAGE', help='The hive image to deploy', default=OPERATORHUB_HIVE_IMAGE_DEFAULT)
```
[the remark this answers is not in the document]
gregsheremeta: done

hack/generate-operator-bundle.py
```diff
@@ -4,115 +4,199 @@
...
+    print("Generating CSV for OpenShift Dedicated")
+    version = generate_osd_version()
+    generate_csv_base(version, args.prev_version, args.hive_image)
+    update_package(version)
+
+
+def generate_csv_base(version, prev_version, hive_image):
+    print("Generating CSV for version: %s" % version)
+
+    BUNDLE_DIR = 'bundle'
```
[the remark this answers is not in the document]
gregsheremeta: fixed

config/templates/hive-csv-template.yaml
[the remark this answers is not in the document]
gregsheremeta: resolved

config/templates/hive-csv-template.yaml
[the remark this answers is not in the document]
gregsheremeta: resolved

gregsheremeta:
/hold
need to update app_sre_creation_image_catalog.sh too

hack/app_sre_create_image_catalog.sh
```diff
@@ -19,67 +18,33 @@ git clone \
     https://app:${APP_SRE_BOT_PUSH_TOKEN}@gitlab.cee.redhat.com/service/saas-hive-operator-bundle.git \
     $SAAS_OPERATOR_DIR
 
-# remove any versions more recent than deployed hash
```
gregsheremeta, line 22: I don't think we need this for hive. It cannot be rolled back. If something is deployed and we need an earlier commit, the current commit should be reverted in a new commit. Please correct me if I'm missing something.

hack/app_sre_create_image_catalog.sh
```diff
@@ -19,67 +18,33 @@ git clone \
...
-        else
-            rm -rf "$BUNDLE_DIR/$version"
-            REMOVED_VERSIONS="$version $REMOVED_VERSIONS"
-        fi
-    done
-fi
 
 # generate bundle
-PREV_VERSION=$(ls $BUNDLE_DIR | sort -t . -k 3 -g | tail -n 1)
+PREV_VERSION=$((cd $SAAS_OPERATOR_BUNDLE_DIR && find -type d | xargs -n 1 basename) | sort -V  | tail -n 1)
```
gregsheremeta, line 23: get only directories, -V version sort (0.1.1800, 1.0.0, 1.0.1 ...)

gregsheremeta:
/hold until I get acks from
/cc @jharrington22 @maorfr @jmelis 

Hi guys, Hive now publishes to operatorhub and is now version 1.0.0, with weekly tagged releases. I've changed the pipeline to generate CSV for both scenarios (osd mode, operatorhub mode) and adapt to the new version scheme.

gregsheremeta:
testing locally using my bundle repo fork...
runs of `hack/app_sre_create_image_catalog.sh staging quay.io/gshereme/hive` are successful:

```
time="2020-04-10T00:56:55Z" level=info msg=directory dir=manifests file=0.1.996-642daac load=package
time="2020-04-10T00:56:55Z" level=info msg=directory dir=manifests file=0.1.998-9514c58 load=package
time="2020-04-10T00:56:55Z" level=info msg=directory dir=manifests file=1.0.0-64-g69b3279a load=package
Removing intermediate container bc6173b44584
 ---> a73708035f5d
Step 4/4 : CMD ["registry-server", "-t", "/tmp/terminate.log"]
 ---> Running in 499de8ef4124
Removing intermediate container 499de8ef4124
 ---> ed24b1326815
Successfully built ed24b1326815
Successfully tagged quay.io/app-sre/hive-registry:staging-latest

# make a commit to hive, rerun

time="2020-04-10T00:58:39Z" level=info msg=directory dir=manifests file=0.1.998-9514c58 load=package
time="2020-04-10T00:58:39Z" level=info msg=directory dir=manifests file=1.0.0-64-g69b3279a load=package
time="2020-04-10T00:58:39Z" level=info msg=directory dir=manifests file=1.0.0-65-gc1cf6875 load=package
Removing intermediate container 79ab12f4dd82
 ---> 2f88eaf89d08
Step 4/4 : CMD ["registry-server", "-t", "/tmp/terminate.log"]
 ---> Running in d4f7bde3b6ad
Removing intermediate container d4f7bde3b6ad
 ---> 3888ba00eea3
Successfully built 3888ba00eea3
Successfully tagged quay.io/app-sre/hive-registry:staging-latest

# make a commit to hive, rerun
# here I tagged 1.0.1, simulating operatorhub release

time="2020-04-10T01:00:03Z" level=info msg=directory dir=manifests file=0.1.998-9514c58 load=package
time="2020-04-10T01:00:03Z" level=info msg=directory dir=manifests file=1.0.0-64-g69b3279a load=package
time="2020-04-10T01:00:03Z" level=info msg=directory dir=manifests file=1.0.0-65-gc1cf6875 load=package
time="2020-04-10T01:00:03Z" level=info msg=directory dir=manifests file=1.0.1 load=package
Removing intermediate container 624b1456ed0f
 ---> 9689671ff4be
Step 4/4 : CMD ["registry-server", "-t", "/tmp/terminate.log"]
 ---> Running in 3612c0aa2bc5
Removing intermediate container 3612c0aa2bc5
 ---> 7c6593a7e4de
Successfully built 7c6593a7e4de
Successfully tagged quay.io/app-sre/hive-registry:staging-latest

# make a commit to hive, rerun

time="2020-04-10T01:03:40Z" level=info msg=directory dir=manifests file=0.1.998-9514c58 load=package
time="2020-04-10T01:03:40Z" level=info msg=directory dir=manifests file=1.0.0-64-g69b3279a load=package
time="2020-04-10T01:03:40Z" level=info msg=directory dir=manifests file=1.0.0-65-gc1cf6875 load=package
time="2020-04-10T01:03:40Z" level=info msg=directory dir=manifests file=1.0.1 load=package
time="2020-04-10T01:03:40Z" level=info msg=directory dir=manifests file=1.0.1-1-gb2729dd7 load=package
Removing intermediate container e925ff6f2bae
 ---> 2ca113ae58a7
Step 4/4 : CMD ["registry-server", "-t", "/tmp/terminate.log"]
 ---> Running in f5e87112a7ef
Removing intermediate container f5e87112a7ef
 ---> bf68abba05aa
Successfully built bf68abba05aa
Successfully tagged quay.io/app-sre/hive-registry:staging-latest




commit log in bundle repo:

commit 96e5d6946be01893b4075d8471fc6e79c6e13c53 (origin/staging)
Author: Greg Sheremeta <gshereme@redhat.com>
Date:   Thu Apr 9 21:02:53 2020 -0400

    add version 1.0.1-1-gb2729dd7
    
    replaces 1.0.1

commit 2c69d4a007697ee09d5068a1e8f268bceb2eead0    ## here I tagged 1.0.1, simulating operatorhub release
Author: Greg Sheremeta <gshereme@redhat.com>
Date:   Thu Apr 9 20:59:16 2020 -0400

    add version 1.0.1
    
    replaces 1.0.0-65-gc1cf6875

commit d5028c9a07535a903c1485c7f03f6f5149ff5020
Author: Greg Sheremeta <gshereme@redhat.com>
Date:   Thu Apr 9 20:57:52 2020 -0400

    add version 1.0.0-65-gc1cf6875
    
    replaces 1.0.0-64-g69b3279a

commit 99be6d8feb68abdd3cfb628dea9d7ff5a7e8fad2
Author: Greg Sheremeta <gshereme@redhat.com>
Date:   Thu Apr 9 20:56:08 2020 -0400

    add version 1.0.0-64-g69b3279a
    
    replaces 0.1.1844-shad80f25a

commit e6b4bf63b7faa07bfc33624dbfa8d502f09c70a3 (HEAD -> staging, upstream/staging)
Author: App-SRE Team <sd-app-sre@redhat.com>
Date:   Thu Apr 9 13:51:12 2020 +0000

    add version 1844-d80f25a
    
    replaces 0.1.1841-sha4185b02
    removed versions:

```

gregsheremeta:
still need to update `olm-registry-deploy.sh`, missed that

hack/generate-operator-bundle.py
```diff
@@ -117,7 +117,9 @@ def generate_csv_base(version, prev_version, hive_image):
     csv['spec']['replaces'] = "hive-operator.v%s" % prev_version
 
     # Update the deployment to use the defined image:
-    image_ref = "%s:v%s" % (hive_image, version)
+    image_ref = hive_image
+    if not ":" in image_ref:
```
gregsheremeta, line 121: i'm not sure the best way to check a ref for a tag

maorfr: `git describe --exact-match --tags <ref>` perhaps?

gregsheremeta: Thanks! -- not git tag (English needs more words)

hive_image is a user-specified image ref e.g. `quay.io/gshereme/hive` and I need to know if there's a tag on the end e.g. `quay.io/gshereme/hive:sometag`. If there's not, use the version as the tag.

hack/app_sre_create_image_catalog.sh
jharrington22: So this exists to for production deploys only see [here](https://github.com/openshift/hive/blob/master/hack/app_sre_build_deploy.sh#L31). What it does is removes any versions in front of the current deployed hash so that when you this new CRD is push it "replaces" the last deployed version to prod. Stage pushes all versions, production only pushes those we promote in saas-hive. I believe you still want this unless you are pushing every version to prod.

hack/app_sre_create_image_catalog.sh
jharrington22: I did not know `sort` supported this `-V` option, nice!

hack/generate-operator-bundle.py
```diff
@@ -4,88 +4,172 @@
...
+    csv['spec']['install']['spec']['deployments'][0]['spec']['template']['spec']['containers'][0]['image'] = image_ref
+    csv['metadata']['annotations']['containerImage'] = image_ref
+
+    # Set the CSV createdAt annotation:
+    now = datetime.datetime.now()
+    csv['metadata']['annotations']['createdAt'] = now.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    # Write the CSV to disk:
+    csv_filename = "hive-operator.v%s.clusterserviceversion.yaml" % version
+    csv_file = os.path.join(version_dir, csv_filename)
```
jharrington22, line 132: So will OLM manage the CRD's or Hive still? If you are putting them in the bundle that means that OLM is going to deploy them? Don't you want to add them to `csv['spec']['customresourcedefinitions']['owned']` ?

hack/generate-operator-bundle.py
jharrington22: Since its a string these seems reasonable to me.

hack/app_sre_create_image_catalog.sh
jmelis: +1 to what @jharrington22 said

hack/generate-operator-bundle.py
```diff
@@ -4,88 +4,172 @@
...
 
+OSD_HIVE_IMAGE_DEFAULT = 'quay.io/app-sre/hive'
+OPERATORHUB_HIVE_IMAGE_DEFAULT = 'quay.io/openshift-hive/hive'
+CHANNEL_DEFAULT = 'alpha'
+BUNDLE_DIR = 'bundle'
+PACKAGE_FILE = os.path.join(BUNDLE_DIR, 'hive.package.yaml')
+
+
+def current_operatorhub_version():
+    result = subprocess.check_output("git describe --abbrev=0", shell=True).strip()
```
jmelis, line 28: `check_output` will return a `bytes` object, I think it would be safer to decode it so you don't have problems later on.

```
result = subprocess.check_output("git describe --abbrev=0", shell=True).strip().decode()
```

There is a similar line below

jmelis:
@gregsheremeta this is great, I've reviewed it and at least the osd part works great.

Regarding the idea of removing the part that prunes the CSVs that have not been deployed to prod, perhaps as Greg suggest we could move to that model, which would simplify things a lot.

In order to understand why we have that piece please look at this:
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md#skipping-updates

Especially these sentences:

> We traverse every version of the operator until we’re at the newest version the CatalogSource contains.
> But sometimes this is not a safe operation to perform. There will be cases where a published version of an operator should never be installed on a cluster if it hasn’t already (e.g. because that version introduces a serious vulnerability).

After reading that we decided that since OLM traverses every version that may install a bad version, we decided to start pruning so the prod catalog so it skips all version that have never been set as promoted to prod versions.

If you agree that we can go ahead and stop pruning that would basically mean that we only need to build one catalog which would be great in terms of simplifying the builds.

Also, it's worth noting that OLM now supports [skipRange](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md#skiprange), so we could start leveraging that to skip any bad versions that we may identify.

So the questions that I think we need to answer to determine whether we should stop pruning the CSVs are the following:

@dgoodwin: do you foresee any problems / risks if we when we promote to prod we traverse through all the versions in between? that means that OLM will install each version and immediately install the next one until it reaches the target one.

@jharrington22: same for aws-account-operator and the other SREP operators.

@ecordell: In our 3.11 hive clusters we are using `quay.io/operator-framework/olm@sha256:73d60e4f2adbc70ed8df93245fb2d83c9e0062489a22110d897b83c21918e101` which maps to OLM version: 0.13.0 / git commit: 16619cd27fa56af837c26a49ba9badc7b6566fd8. Can we start using `skipRange` there? I've looked inside that commit and I don't see the same support for `skipRange` as you have right now in HEAD (for example in that commit skipRange is not a table in the sqlite DB of the operator). For our OSDv4.3 we are using this OLM version: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:dbde7140eb5c05e05252e2bc2891f47761aecf465803d08831c6b20c84bebbb8`, which I don't know to what OLM version it maps to. Is `skipRange` fully supported here?

hack/generate-operator-bundle.py
jmelis: oh, I just realized this is executed by `py2.7` so what I said is inaccurate. Please ignore this.

gregsheremeta:
The pruning block isn't working and I don't think it ever has:
```
13:52:51 git add .
13:52:51 + git add .
13:52:52 warning: You ran 'git add' with neither '-A (--all)' or '--ignore-removal',
13:52:52 whose behaviour will change in Git 2.0 with respect to paths you removed.
13:52:52 Paths like 'hive/0.1.1846-sha7f123aa/hive-operator.v0.1.1846-sha7f123aa.clusterserviceversion.yaml' that are
13:52:52 removed from your working tree are ignored with this version of Git.
```

Thus the current flow is this. The catalog is built at build-master time. Every commit to master ends up in the catalog directory. In build-master, the csv 'replaces' field is always set to the sha of the current deployment.

Assume saas-herder points to currently deployed hash1. Since that deployment, we have 2 more commits. Catalog looks like:

hash0
hash1  replaces: hash0   <-- currently deployed
hash2  replaces: hash1
hash3  replaces: hash1

After hash3's build-master ran, the latest registry image contains all of those.

It's a new week -- time to update the deployment. We could pick either hash3 or hash2. If we pick hash3, hash2 is still in the catalog and registry image. OLM just ignores it because it makes its chain by starting at the channel's HEAD and walking back. So, channel references hash3, replaces hash1, ... hash2 is present but ignored.

tl;dr: we can remove that block and there is no effective change. If the pruner actually worked, hash2 would have been deleted from the catalog and wouldn't be deployable, so good thing it doesn't work :)

gregsheremeta:
/close

I'm going to split this up into smaller PRs, starting with #964.

PR #941, opened 2020-04-14
Update docs to specify supported version of kind

akhil-rane:
Currently, we support 0.5.1 version of kind. Our development doc does not mention
it. This PR updates the doc specifying the supported version of kind.

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -67,7 +67,7 @@ git clone https://github.com/openshift/hive.git
 
 ## Deploying with Kubernetes In Docker (kind)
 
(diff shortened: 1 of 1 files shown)


gregsheremeta:
Can you also add a note about it in the "Deploying with Kubernetes In Docker (kind)" section? It's easy to miss in the bullet list, especially for "speed readers" like me :laughing:

docs/developing.md
```diff
@@ -39,6 +39,7 @@
 ### External tools
 
 - [kustomize](https://github.com/kubernetes-sigs/kustomize#kustomize)
+- [kind](https://github.com/kubernetes-sigs/kind) (0.5.1)
```
staebler, line 42: I would rather see the version limitation spelled out explicitly in the `Deploying with Kubernetes In Docker` section. The kind tool is specific to that section.

staebler: In other words, I do not think that kind should be added to the `External Tools`.

akhil-rane: Done

PR #942, opened 2020-04-14
Fix bug in label ownership reconciliation

twiest:
- [x] Fix bug that causes objects in other namespaces to be set with the incorrect owner if the owner's name in multiple namespaces is the same.
- [x] Update unit tests for bug fix
- [x] Manually test bug fix


twiest:
@staebler @dgoodwin This PR reverts the revert for the label ownership reconciliation and fixes the bug where it thinks that objects in other namespaces are the valid owner.

PR #937, opened 2020-04-09
use standard makefile

staebler:
Use the makefile from https://github.com/openshift/build-machinery-go

PR #947, opened 2020-04-20
OpenStack MachinePool Support

dgoodwin:
WIP just because I need to write tests, but have to set this aside for a few days, main code should be ready for review.


pkg/apis/hive/v1/openstack/machinepools.go
```diff
@@ -5,7 +5,10 @@ package openstack
 type MachinePool struct {
 	// FlavorName defines the OpenStack Nova flavor.
 	// eg. m1.large
-	FlavorName string `json:"type"`
+	// The json key here differs from the installer which uses both "computeFlavor" and type "type" depending on which
+	// type you're looking at, and the resulting field on the MachineSet is "flavor". We are opting to stay consistent
+	// with the end result.
+	FlavorName string `json:"flavor"`
```
staebler, line 11: What is the reasoning behind keeping the name of the field `FlavorName` instead of `Flavor` to match the json key?

pkg/apis/hive/v1/openstack/platform.go
```diff
@@ -13,4 +13,7 @@ type Platform struct {
 	// Cloud will be used to indicate the OS_CLOUD value to use the right section
 	// from the cloud.yaml in the CredentialsSecretRef.
 	Cloud string `json:"cloud"`
+
+	// TrunkSupport indicates whether or not to use trunk ports in your OpenShift cluster.
+	TrunkSupport bool `json:"trunkSupport"`
```
staebler, line 18: Is this a required field? Or does it default to false?

pkg/clusterresource/builder.go
```diff
@@ -104,9 +104,6 @@ type Builder struct {
 
 	// MachineNetwork is the subnet to use for the cluster's machine network.
 	MachineNetwork string
-
-	// SkipMachinePoolGeneration is set to skip generating MachinePool objects
-	SkipMachinePoolGeneration bool
```
staebler, line 109: Why are we removing the option to skip machine pool generation?

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -0,0 +1,96 @@
+package remotemachineset
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
+	installosp "github.com/openshift/installer/pkg/asset/machines/openstack"
```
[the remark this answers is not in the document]
staebler: `installosp` matches what is used in other actuators. I think it is the right choice.

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -0,0 +1,96 @@
...
+	installertypes "github.com/openshift/installer/pkg/types"
+	installertypesosp "github.com/openshift/installer/pkg/types/openstack"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+)
+
+const (
+	workerUserDataSecretName = "worker-user-data"
```
staebler, line 18: If you want to make this a package-level constant, then (1) it should be moved to a separate file to be used by the other actuators and (2) the other actuators should be changed to use it.

staebler: It would be nice to do the same thing with `workerRole`, too, so that the OpenStack actuator is not getting the `workerRole` constant from the GCP actuator file.

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// OpenStackActuator encapsulates the pieces necessary to be able to generate
+// a list of MachineSets to sync to the remote cluster.
+type OpenStackActuator struct {
+	logger log.FieldLogger
+	region string
+	amiID  string
```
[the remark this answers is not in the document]
staebler: `region`, too

go.sum
```diff
@@ -1403,30 +1524,44 @@ github.com/openshift/api v0.0.0-20200116145750-0e2ff1e215dd/go.mod h1:fT6U/JfG8u
...
 github.com/openshift/api v3.9.1-0.20191111211345-a27ff30ebf09+incompatible/go.mod h1:dh9o4Fs58gpFXGSYfnVxGR9PnV53I8TW84pQaJDdGiY=
 github.com/openshift/baremetal-operator v0.0.0-20200206190020-71b826cc0f0a/go.mod h1:cXwn0hhgHpORjBasg0RnZwhKaJGy9+r6qgj0HCXrs/Y=
 github.com/openshift/build-machinery-go v0.0.0-20200205161356-ef115f5adc73/go.mod h1:1CkcsT3aVebzRBzVTSbiKSkJMsC/CASqxesfqEMfJEc=
 github.com/openshift/client-go v0.0.0-20190617165122-8892c0adc000/go.mod h1:6rzn+JTr7+WYS2E1TExP4gByoABxMznR6y2SnUIkmxk=
 github.com/openshift/client-go v0.0.0-20200116152001-92a2713fa240/go.mod h1:4riOwdj99Hd/q+iAcJZfNCsQQQMwURnZV6RL4WHYS5w=
+github.com/openshift/cloud-credential-operator v0.0.0-20200316201045-d10080b52c9e h1:2gyl9UVyjHSWzdS56KUXxQwIhENbq2x2olqoMQSA/C8=
 github.com/openshift/cloud-credential-operator v0.0.0-20200316201045-d10080b52c9e/go.mod h1:iPn+uhIe7nkP5BMHe2QnbLtg5m/AIQ1xvz9s3cig5ss=
 github.com/openshift/cluster-api v0.0.0-20190805113604-f8de78af80fc/go.mod h1:mNsD1dsD4T57kV4/C6zTHke/Ro166xgnyyRZqkamiEU=
 github.com/openshift/cluster-api v0.0.0-20191030113141-9a3a7bbe9258/go.mod h1:T18COkr6nLh9RyZKPMP7YjnwBME7RX8P2ar1SQbBltM=
```
staebler, line 1536: Can you run `go mod tidy` to remove these two lines?

I added this to the vendor target in #937.

https://github.com/openshift/hive/blob/e760d971aec2f70eef80ec2e317982b842543d8e/Makefile#L69

staebler: Bah, wrong two line. It should be the lines for the old github.com/openshift/cluster-api-provider-openstack version. But the point remains.

pkg/apis/hive/v1/openstack/platform.go
dgoodwin: I believe it defaults to false.

pkg/clusterresource/builder.go
dgoodwin: This was added in the interim while we waited for openstack machine pool support. I don't know of an actual use case to keep this, and if there's no use case it's a path I'd rather remove. A developer or advanced hiveutil user can do this by editing the -o yaml output.

dgoodwin:
PR updated for feedback and a unit test, there's not much involved here as there are no regions or zones to spread machinesets across, however the test still buys us coverage in case the installer starts using more of the fake platform we're populating without us realizing.

pkg/apis/hive/v1/openstack/platform.go
staebler: If there is a default, then this should be `omitempty` and `optional`.

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -0,0 +1,90 @@
...
+			OpenStack: &installertypesosp.Platform{
+				TrunkSupport: trunkSupportStr,
+			},
+		},
+	}
+
+	// The installer will assume to upload a copy of the RHCOS image for the release with this name.
+	// It is possible to override this in the installer with the OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
+	// env var, but we do not expose this in Hive, so it should be safe to assume the installers default
+	// name format. If we do add support for uploading images more efficiently (as it's a 2GB dl+ul per
```
staebler, line 80: We do expose a way to set this via the `.spec.provisioning.installerEnv`. But I agree with the conclusion that we are safe to assume the default name format.

pkg/controller/remotemachineset/openstackactuator_test.go
```diff
@@ -0,0 +1,96 @@
...
+			actuator := &OpenStackActuator{
+				logger: log.WithField("actuator", "openstackactuator_test"),
+			}
+
+			generatedMachineSets, _, err := actuator.GenerateMachineSets(test.clusterDeployment, test.pool, actuator.logger)
+
+			if test.expectedErr {
+				assert.Error(t, err, "expected error for test case")
+			} else {
+				validateOSPMachineSets(t, generatedMachineSets, test.expectedMachineSetReplicas)
```
staebler, line 53: Should have an `assert.NoError` here.

staebler:
/lgtm cancel

The e2e tests are broken. Maybe something got messed up with the short name for ClusterDeployment.

```
Getting ClusterDeployment hive-bd41905b-b8dd-4406-9d64-8fc7b869fd05. Try #59/60:
Failed, sleeping 10 seconds
Getting ClusterDeployment hive-bd41905b-b8dd-4406-9d64-8fc7b869fd05. Try #60/60:
Failed, exiting
```

config/crds/hive_v1_clusterdeployment.yaml
```diff
@@ -2,15 +2,13 @@ apiVersion: apiextensions.k8s.io/v1beta1
...
-  labels:
-    controller-tools.k8s.io: "1.0"
   name: clusterdeployments.hive.openshift.io
 spec:
   additionalPrinterColumns:
   - JSONPath: .spec.clusterName
     name: ClusterName
     type: string
-  - JSONPath: .metadata.labels.hive\.openshift\.io/cluster-type
+  - JSONPath: .metadata.labels[hive.openshift.io/cluster-type]
```
dgoodwin, line 11: The failure is this change in lines right here, the new syntax appears to break the additional printer columns, at least on the versions of kube we're running against. The result is no additional printer columns, which fails our test looking for BASEDOMAIN.

Derek mentioned this column was a weird thing to display awhile back and I'd been meaning to remove it since. Rather than fight this I'm going to just take it out.

staebler: Is there any other user that is relying on that column? Would it be easier to keep it in as the older version for now?

Checkout the master version of the file and update the CRDs from that version.

dgoodwin: I have doubts anyone is using it and I can't see where this would fit into any AppSRE/OCM/SRE-P scripting, but I don't think we should be considering additional printer columns to be an API. If this does happen to break some pipeline or automation, that should be addressed anyhow.

staebler: Here is an example of a place where additional printer columns are used by users. I am fine if we want to make this change, but I would rather make it as a separate PR rather than burying it in a larger PR.

https://github.com/openshift/managed-cluster-config/blob/0f29251676f95c9414c9326cc43afe652246cc0d/scripts/rotate-aws-creds.sh#L93

dgoodwin: I don't think that is using any of our additional printer columns. 

I will go get the master file copy and hand edit it for the new field and try to get this all updated properly, then submit a new PR.

staebler: You should not need to hand-edit anything.
```
git checkout master -- config/crds
make update-codegen-crds
```

dgoodwin: This picks up an unrelated change that removes syncsetinstance Applied printer column for some reason. I'll just got with the hand edit.

dgoodwin:
Updated for conflict resolution.

PR #952, opened 2020-04-21
Implement Hive uninstallation via deleting HiveConfig 'hive'.

dgoodwin:
If the HiveConfig named 'hive' (the only name we recognize), Hive
components will be torn down.

A few minor objects may remain but would be caught by namespace cleanup.
Critically we wipe out all RBAC, Deployments, and webhooks.


staebler:
Can we add cleanup to the e2e tests to verify that it is working?

dgoodwin:
E2E occurred to me, it's probably worth adding, I will do so.

I'm am unsure about a parent object to do GC. We don't want to have to manually add that owner ref to every resource as we'd be no better off in terms of missing things. We could add a util for the operator that did it for us before every Apply, but we'd have to be sure we didn't use it for CRDs or anything else we might mistakenly lump into the cleanup bucket that we don't want to. It looks like we can do owner refs to a cluster-scoped resource on both namespaced and cluster scoped resources, so HiveConfig itself might work but anything namespaced probably wouldn't. It would be a bit of a pain to roll out to pre-existing installations, but perhaps we wouldn't need to do that.

I'm torn, I think simpler might be good enough for now and we can worry about it if this proves insufficient.

pkg/operator/hive/hiveadmission.go
```diff
@@ -242,3 +246,44 @@ func (r *ReconcileHiveConfig) is311(hLog log.FieldLogger) (bool, error) {
...
+		err := dynamicDelete(r.dynamicClient, gvrnsn, hLog)
+		if err != nil {
+			return err
+		}
+	}
+
+	validatingWebhooks := &admregv1.ValidatingWebhookConfigurationList{}
+	err := r.Client.List(context.Background(), validatingWebhooks)
+	if err != nil {
+		hLog.WithError(err).Error("error listing ValidatingWebhookConfigurations")
```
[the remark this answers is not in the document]
dgoodwin: Updating but I wouldn't have thought IsAlreadyExists IsConflict or IsNotFound would not appear on List calls?

pkg/operator/hive/hiveadmission.go
```diff
@@ -242,3 +246,44 @@ func (r *ReconcileHiveConfig) is311(hLog log.FieldLogger) (bool, error) {
...
+		return err
+	}
+
+	// Delete any hive validating webhooks. Relies on the convention that they are named ending with hive.openshift.io.
+	for _, wh := range validatingWebhooks.Items {
+		if strings.HasSuffix(wh.Name, "hive.openshift.io") {
```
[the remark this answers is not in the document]
dgoodwin: Yeah probably, that or the owner ref to something cluster scoped.

pkg/operator/hive/operatorutils.go
```diff
@@ -57,15 +57,30 @@ func applyClusterRoleBindingAssetWithSubjectNamespaceOverride(h *resource.Helper
...
+func dynamicDelete(dynamicClient dynamic.Interface, gvrnsn gvrNSName, hLog log.FieldLogger) error {
 	rLog := hLog.WithFields(log.Fields{
-		"gvr":  gvr,
-		"name": name,
+		"resource": gvrnsn,
 	})
-	err := dynamicClient.Resource(gvr).Delete(name, &metav1.DeleteOptions{})
+	gvr := schema.GroupVersionResource{Group: gvrnsn.group, Version: gvrnsn.version, Resource: gvrnsn.resource}
+	var err error
+	if gvrnsn.namespace != "" {
```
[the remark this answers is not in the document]
dgoodwin: Ahhh nice, thanks.

pkg/operator/hive/hiveadmission.go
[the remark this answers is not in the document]
staebler: You are correct that there is nothing currently that would be returned from List that would be logged at anything other that Error. However, I would like the behavior to be that we **always** use `controllerutils.LogLevel`.

staebler:
> I'm torn, I think simpler might be good enough for now and we can worry about it if this proves insufficient.

It is the simpler that I am concerned about. Our disagreement is with which is simpler. To me, it seems much simpler to let the kube garbage collection clean things up for us rather than having to keep up-to-date a bunch of separated lists of resources that should be cleaned up. Everything created by the hive operator (with the exception of the hive namespace and the managed-domains configmap) is created through the same resource helper. It seems simple enough to me to add code to the resource helper to add the ownership when the resource is applied. This will (1) allows us to add more resources without having to remember to add cleanup code and (2) automatically fix up existing resources.

staebler:
> It looks like we can do owner refs to a cluster-scoped resource on both namespaced and cluster scoped resources, so HiveConfig itself might work but anything namespaced probably wouldn't.

Oh yeah, I had that backwards in my head. That is even better. That means that we do not even need to have a dummy object.

dgoodwin:
By simpler in this case I mean the one that is explicit and already implemented when we're on a deadline. :) I will see if I can get the HiveConfig owner ref in place today.

pkg/operator/hive/hiveadmission.go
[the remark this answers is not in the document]
dgoodwin: "for Client calls" I assume. We should doc this somewhere, probably in developing.md.

gregsheremeta:
Can you tweak the commit message?

> If the HiveConfig named 'hive' (the only name we recognize), Hive components will be torn down.

missing "is deleted"

dgoodwin:
Ok added e2e and then redid to use the hiveconfig owner refs, please let me know ASAP how this looks, hoping to deliver to ACM for tomorrow.

Will get the commit message updated when it's time to squash, thanks for catching!

gregsheremeta:
in a separate PR it would be good to document this well. We should at least clearly document what will be deleted and what will remain (e.g. "`ClusterDeployments` will not be deleted")

pkg/operator/util/apply.go
```diff
@@ -21,3 +37,78 @@ func ApplyAsset(h *resource.Helper, assetPath string, hLog log.FieldLogger) erro
...
+	return nil
+}
+
+// ApplyRuntimeObjectWithGC adds an OwnerReference to the HiveConfig on the runtime object, and applies it to the cluster.
+func ApplyRuntimeObjectWithGC(h *resource.Helper, runtimeObj runtime.Object, hc *hivev1.HiveConfig) (resource.ApplyResult, error) {
+	obj, err := meta.Accessor(runtimeObj)
+	if err != nil {
+		panic(err)
+	}
+	ownerRef := v1.OwnerReference{
```
staebler, line 99: Do we want to block owner deletion? I am not sure whether we do or not. What were your thoughts here?

pkg/operator/util/apply.go
```diff
@@ -21,3 +37,78 @@ func ApplyAsset(h *resource.Helper, assetPath string, hLog log.FieldLogger) erro
...
+		return errors.Wrapf(err, "unable to apply asset: %s", roleBindingAssetPath)
+	}
+	return nil
+}
+
+// ApplyRuntimeObjectWithGC adds an OwnerReference to the HiveConfig on the runtime object, and applies it to the cluster.
+func ApplyRuntimeObjectWithGC(h *resource.Helper, runtimeObj runtime.Object, hc *hivev1.HiveConfig) (resource.ApplyResult, error) {
+	obj, err := meta.Accessor(runtimeObj)
+	if err != nil {
+		panic(err)
```
staebler, line 97: Do we really want to panic here? Is the intention that we want hive-operator to crash-loop when there are bad assets?

dgoodwin: I added the error return after and forgot to replace the panic, will fix when I squash.

pkg/operator/util/apply.go
dgoodwin: From the GC docs: "If an object’s ownerReferences field is set by a controller (such as Deployment or ReplicaSet), blockOwnerDeletion is set automatically and you do not need to manually modify this field."

That's probably our situation as well. I'll add it.

dgoodwin:
PR updated, thanks for quick review.

dgoodwin:
There is a bug with the openshift specific stuff that's hidden on kind, fixing now.

dgoodwin:
Hopefully this does the trick.

pkg/operator/util/apply.go
```diff
@@ -21,3 +40,79 @@ func ApplyAsset(h *resource.Helper, assetPath string, hLog log.FieldLogger) erro
...
+	}
+	// This assumes we have full control of owner references for these resources the operator creates.
+	obj.SetOwnerReferences([]v1.OwnerReference{ownerRef})
+	return h.ApplyRuntimeObject(runtimeObj, scheme.Scheme)
+}
+
+func readRuntimeObject(assetPath string) (runtime.Object, error) {
+	codec := serializer.NewCodecFactory(scheme.Scheme)
+	// Additional SchemeGroupVersions may need to be added here if the operator needs to start deploying new types.
+	return runtime.Decode(codec.UniversalDecoder(rbacv1.SchemeGroupVersion, orbacv1.GroupVersion), assets.MustAsset(assetPath))
```
staebler, line 117: You want `rbacv1` and `orbacv1` added to the scheme not added as preferred versions to use. If they are not in the scheme, then the decoder will not know how to decode them.

staebler: Also, for what it is worth, you should not have to decode to an structured type to add the owner reference. You should be able to decode to an unstructured type.

dgoodwin: I am very confused, my first attempt at fixing this error (which I got for the Kube type) was to add to the scheme in cmd/operator/main.go, but this did not work. I then started passing the scheme all around from manager down to this function to make sure the right one was in use, but that also didn't work. Adding as an argument to UniversalDecoder was when it finally fixed itself, and then I backed out the scheme registration and arguments and it kept working. 

In current state the Kube RBAC is not added to scheme, at least not in main.go, and yet those resources are able to be applied. 

Will keep digging. I am setting up an OpenShift cluster to run Hive so I can test myself and save some time, I'll definitely try scheme registration in main.go, but confused as to what's going on given what I saw.

go.sum
```diff
@@ -11,6 +11,7 @@ cloud.google.com/go v0.44.1/go.mod h1:iSa0KzasP4Uvy3f1mN/7PiObzGgflwredwwASm/v6A
 cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxKY=
 cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
 cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
+cloud.google.com/go v0.49.0 h1:CH+lkubJzcPYB1Ggupcq0+k8Ni2ILdG2lYjDIgavDBQ=
```
staebler, line 14: None of the changes in this file are needed. We can leave them for now. But we should clean them up later.

PR #954, opened 2020-04-22
Fix label ownership apiversion bug

twiest:
- [x] Fix label ownership bug where if the apiversion differs it won't
      change the controller owner.
- [x] Add test case to ensure bug doesn't come back
- [x] Change unit tests to be more easily expanded

### DIFF
--- a/pkg/apis/hive/v1/register.go
+++ b/pkg/apis/hive/v1/register.go
@@ -14,8 +14,14 @@ import (
 )
 
 var (
+	// HiveAPIGroup is the group that all hive objects belong to in the API server.
+	HiveAPIGroup = "hive.openshift.io"
+
+	// HiveAPIVersion is the api version that all hive objects are currently at.
+	HiveAPIVersion = "v1"
+
 	// SchemeGroupVersion is group version used to register these objects
-	SchemeGroupVersion = schema.GroupVersion{Group: "hive.openshift.io", Version: "v1"}
+	SchemeGroupVersion = schema.GroupVersion{Group: HiveAPIGroup, Version: HiveAPIVersion}
 
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
 	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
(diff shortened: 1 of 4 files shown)


pkg/test/dnszone/dnszone.go
```diff
@@ -1,11 +1,10 @@
 package dnszone
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
 	"github.com/openshift/hive/pkg/test/generic"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
staebler, line 7: This import should stay where it was. The k8s imports should come in a section before the hive imports.

pkg/test/dnszone/dnszone.go
```diff
@@ -52,6 +51,17 @@ func WithControllerOwnerReference(owner metav1.Object) Option {
 	return Generic(generic.WithControllerOwnerReference(owner))
 }
 
+// WithControllerOwnerReferenceAPIVersion sets the api version on the controller owner reference.
+func WithControllerOwnerReferenceAPIVersion(apiVersion string) Option {
```
staebler, line 55: This function is very specific to the ownership tests. Can we leave it in that file rather than adding it here?

pkg/test/dnszone/dnszone.go
twiest: I think my editor probably did that (probably because I removed a line that used it so it deleted it, and then I put the line back and it re-added it). At any rate, this wasn't an intentional change. Will fix.

pkg/test/dnszone/dnszone.go
twiest: Sure, but I reserve the right to put it back if I end up needing it in a different package. :)

twiest:

twiest:
I tested that the new logic does indeed set the apiversion correctly on hive objects that we're reconciling.

It also was showing the correct behavior on objects that were already set correctly (no modifications).

/hold cancel

PR #955, opened 2020-04-23
Implement Hive uninstallation via deleting HiveConfig 'hive'.

dgoodwin:
If the HiveConfig named 'hive' (the only name we recognize) is deleted,
Hive components will be torn down.

This is accomplished by automatically adding owner references to most
resources the hive-operator creates, to the HiveConfig hive. Once
deleted we let kubernetes garbage collection take care of cleaning
everything up.


dgoodwin:
This is a cherry-pick of https://github.com/openshift/hive/pull/952 for OCM, mostly just to make sure e2e tests pass after backport.

dgoodwin:
Added a fix for the OpenShift specific error.

Issue #774, opened 2020-01-10
adoption requires cloud credentials


mhrivnak:
I ran `make build` and `make deploy` from the v1 branch. I then tried to "adopt" an existing cluster and got an error message about missing AWS credentials. Cloud credentials should not be required for adoption. I saw a similar message if I specified the cloud as "azure" or "gcp".

```
$ bin/hiveutil create-cluster --adopt --adopt-admin-kubeconfig=/tmp/cluster0.config --adopt-cluster-id=cluster0 --adopt-infra-id=cluster0 cluster0
ERRO[0000] Cannot load AWS credentials                  
ERRO[0000] Error                                         error="open /home/mhrivnak/.aws/credentials: no such file or directory"
```

mhrivnak:
I added `--include-secrets=false` to the command and that enabled it to succeed. That causes the create command to skip [this block of code](https://github.com/openshift/hive/blob/5520264/contrib/pkg/createcluster/create.go#L471-L493), which tries to get the cloud credentials by calling `o.cloudProvider.generateCredentialsSecret(o)`

staebler:
Cloud credentials are required for Hive to manage a cluster. The credentials are not used solely during provisioning: The credentials are used for on-going operations and for de-provisioning as well.

mhrivnak:
The docs describe adopting a "kind" cluster for dev purposes, which presumably cannot take advantage of any cloud credentials. Seems like there should be a way to do that without being required to provide credentials.

https://github.com/openshift/hive/blob/v1/docs/developing.md#adopting-clusterdeployments

staebler:
> The docs describe adopting a "kind" cluster for dev purposes, which presumably cannot take advantage of any cloud credentials.

Yes, however, even when adopting a "kind" cluster, the aws creds are required. In the example provided, the "kind" cluster is adopted as an AWS cluster.

dhellmann:
There won't be any cloud credentials for a bare metal cluster. How do we plan to deal with that?

mhrivnak:
> > The docs describe adopting a "kind" cluster for dev purposes, which presumably cannot take advantage of any cloud credentials.
> 
> Yes, however, even when adopting a "kind" cluster, the aws creds are required. In the example provided, the "kind" cluster is adopted as an AWS cluster.

[...]

PR #957, opened 2020-04-27
Add "local" cloud provider option

jeefy:
Currently if you follow the docs in [/docs/developing.md] to create and adopt a cluster locally, the process will fail. The adoption process makes the assumption that the cluster being adopted is within a cloud provider and will attempt to parse credentials that may or may not exist.

This adds the `local` cloud provider and stubs out required objects required for cluster adoption. It also updates the development docs to use the new flag.

Tested this locally and it works great! (Though I had to downgrade my version of `kind`. I might shoot another patch forward that supports 0.5.x and and 0.7.0+)

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -3,17 +3,18 @@ package createcluster
 import (
 	"encoding/json"
 	"fmt"
-	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
-	gcputils "github.com/openshift/hive/contrib/pkg/utils/gcp"
-	openstackutils "github.com/openshift/hive/contrib/pkg/utils/openstack"
-	"github.com/openshift/hive/pkg/clusterresource"
-	"github.com/openshift/hive/pkg/gcpclient"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"os/user"
 	"path/filepath"
 	"strings"
+
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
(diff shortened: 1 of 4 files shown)


jeefy:
/hold 

Found a cleaner way to do this. :)

jeefy:
/hold cancel

Quick refactor to make the implementation cleaner. :)

staebler:
I am not keen on having a "local" cloud provider option in hiveutil when Hive in general does not have a "local" cloud provider.

jeefy:
Originally I had a separate "--adopt-local" flag, but it didn't seem like a
clean option vs. adding a "local" cloud provider.

I'm happy to swap back to the previous method if it's the preferred option,
it just will add some code complexity.

vkareh:
Thanks @jeefy! This flag can help clusters-service when developing features that rely on modifying ClusterDeployment and SyncSets and not have to wait for a full cloud provisioning before testing the local feature.

gregsheremeta:
Related: https://github.com/openshift/hive/issues/774

gregsheremeta:
@jeefy @vkareh  Does the comment here https://github.com/openshift/hive/issues/774#issuecomment-573227963 help?

jeefy:
That totally helps! If that's the preferred method for testing hive locally
then I'll nuke this PR and submit a new one to update the developer docs
(since I'm kind of on this train)

I do see some merit in the idea of adding a "local" cloud provider though,
and the UX seems neater. But I defer to y'all. :)

gregsheremeta:
> I'll nuke this PR and submit a new one to update the developer docs

Please do.

/close

PR #958, opened 2020-04-28
Delete successful install pods and old failed provisions

akhil-rane:
This PR ensures that the install pod gets deleted after successful cluster
provision. It also deletes failed cluster provisions that are more than 7 days
old.

cc: @staebler @gregsheremeta

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -4,6 +4,11 @@ set -e
 
 max_tries=60
 sleep_between_tries=10
+# Set timeout for the cluster deployment to install
+# timeout = sleep_between_cluster_deployment_status_checks * max_cluster_deployment_status_checks
+max_cluster_deployment_status_checks=90
+sleep_between_cluster_deployment_status_checks="1m"
+
 component=hive
 local_hive_image=$(eval "echo $IMAGE_FORMAT")
 export HIVE_IMAGE="${HIVE_IMAGE:-$local_hive_image}"
@@ -208,13 +213,26 @@ echo ""
 echo "Events in cluster namespace"
 oc get events -n ${CLUSTER_NAMESPACE}
 
-echo "Waiting for install job to start and complete"
-
+echo "Waiting for the ClusterDeployment ${CLUSTER_NAME} to install"
 INSTALL_RESULT=""
-if go run "${SRC_ROOT}/contrib/cmd/waitforjob/main.go" "${CLUSTER_NAME}" "install"
+
+i=1
+while [ $i -le ${max_cluster_deployment_status_checks} ]; do
+  IS_CLUSTER_DEPLOYMENT_INSTALLED=$(oc get cd ${CLUSTER_NAME} -n ${CLUSTER_NAMESPACE} -o json | jq .spec.installed )
+  if [[ "${IS_CLUSTER_DEPLOYMENT_INSTALLED}" == "true" ]] ; then
+    INSTALL_RESULT="success"
+    break
+  fi
+  sleep ${sleep_between_cluster_deployment_status_checks}
+  echo "Still waiting for the ClusterDeployment ${CLUSTER_NAME} to install. Status check #${i}/${max_cluster_deployment_status_checks}... "
+  i=$((i + 1))
+done
+
+if [[ "${INSTALL_RESULT}" == "success" ]]
 then
-	echo "ClusterDeployment ${CLUSTER_NAME} was installed successfully"
-	INSTALL_RESULT="success"
+  echo "ClusterDeployment ${CLUSTER_NAME} was installed successfully"
+else
+  echo "Timed out waiting for the ClusterDeployment ${CLUSTER_NAME} to install"
 fi
 
 # Capture install logs
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -495,6 +495,13 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			return reconcile.Result{}, nil
 		}
 
+		// delete failed provisions which are more than 7 days old
+		existingProvisions, err := r.existingProvisions(cd, cdLog)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.deleteOldFailedProvisions(existingProvisions, cdLog)
+
 		cdLog.Debug("cluster is already installed, no processing of provision needed")
 		r.cleanupInstallLogPVC(cd, cdLog)
 
@@ -1767,6 +1774,20 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
 		pLog.Info("Deleting old provision")
 		if err := r.Delete(context.TODO(), provision); err != nil {
 			pLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to delete old provision")
+		}
+	}
+}
+
+// deleteOldFailedProvisions deletes the failed provisions which are more than 7 days old
+func (r *ReconcileClusterDeployment) deleteOldFailedProvisions(provs []*hivev1.ClusterProvision, cdLog log.FieldLogger) {
+	cdLog.Debugf("Deleting failed provisions which are more than 7 days old")
+	for _, provision := range provs {
+		if provision.Spec.Stage == hivev1.ClusterProvisionStageFailed && time.Since(provision.CreationTimestamp.Time) > (7*24*time.Hour) {
+			pLog := cdLog.WithField("provision", provision.Name)
+			pLog.Info("Deleting failed provision")
+			if err := r.Delete(context.TODO(), provision); err != nil {
+				pLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to delete failed provision")
+			}
 		}
 	}
 }
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -1331,6 +1331,48 @@ func TestDeleteStaleProvisions(t *testing.T) {
 	}
 }
 
+func TestDeleteOldFailedProvisions(t *testing.T) {
+	apis.AddToScheme(scheme.Scheme)
+	cases := []struct {
+		name                                    string
+		totalProvisions                         int
+		failedProvisionsMoreThanSevenDaysOld    int
+		expectedNumberOfProvisionsAfterDeletion int
+	}{
+		{
+			name:                                    "One failed provision more than 7 days old",
+			totalProvisions:                         2,
+			failedProvisionsMoreThanSevenDaysOld:    1,
+			expectedNumberOfProvisionsAfterDeletion: 1,
+		},
+		{
+			name:                                    "No failed provision more than 7 days old",
+			totalProvisions:                         2,
+			failedProvisionsMoreThanSevenDaysOld:    0,
+			expectedNumberOfProvisionsAfterDeletion: 2,
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			provisions := make([]runtime.Object, tc.totalProvisions)
+			for i := 0; i < tc.totalProvisions; i++ {
+				if i < tc.failedProvisionsMoreThanSevenDaysOld {
+					provisions[i] = testOldFailedProvision(time.Now().Add(-7*24*time.Hour), i)
+				} else {
+					provisions[i] = testOldFailedProvision(time.Now(), i)
+				}
+			}
+			fakeClient := fake.NewFakeClient(provisions...)
+			rcd := &ReconcileClusterDeployment{
+				Client: fakeClient,
+				scheme: scheme.Scheme,
+			}
+			rcd.deleteOldFailedProvisions(getProvisions(fakeClient), log.WithField("test", "TestDeleteOldFailedProvisions"))
+			assert.Len(t, getProvisions(fakeClient), tc.expectedNumberOfProvisionsAfterDeletion, "unexpected provisions kept")
+		})
+	}
+}
+
 func testEmptyClusterDeployment() *hivev1.ClusterDeployment {
 	cd := &hivev1.ClusterDeployment{
 		TypeMeta: metav1.TypeMeta{
@@ -1497,6 +1539,14 @@ func testFailedProvisionTime(time time.Time) *hivev1.ClusterProvision {
 			LastTransitionTime: metav1.NewTime(time),
 		},
 	}
+	return provision
+}
+
+func testOldFailedProvision(time time.Time, attempt int) *hivev1.ClusterProvision {
+	provision := testProvision()
+	provision.Name = fmt.Sprintf("%s-%02d", provision.Name, attempt)
+	provision.CreationTimestamp.Time = time
+	provision.Spec.Stage = hivev1.ClusterProvisionStageFailed
 	return provision
 }
 
--- a/pkg/controller/clusterprovision/clusterprovision_controller.go
+++ b/pkg/controller/clusterprovision/clusterprovision_controller.go
@@ -173,7 +173,15 @@ func (r *ReconcileClusterProvision) Reconcile(request reconcile.Request) (reconc
 			return r.reconcileRunningJob(instance, pLog)
 		}
 		return r.transitionStage(instance, hivev1.ClusterProvisionStageFailed, "NoJobReference", "Missing reference to install job", pLog)
-	case hivev1.ClusterProvisionStageComplete, hivev1.ClusterProvisionStageFailed:
+	case hivev1.ClusterProvisionStageComplete:
+		pLog.Debugf("ClusterProvision is %s", instance.Spec.Stage)
+		if instance.Status.JobRef != nil && time.Since(instance.CreationTimestamp.Time) > (24*time.Hour) {
+			return r.deleteInstallJob(instance, pLog)
+		}
+		// installJobDeletionRecheckDelay will be duration between current time and expected install job deletion time (provision creation time + 24 hours)
+		installJobDeletionRecheckDelay := instance.CreationTimestamp.Time.Add(24 * time.Hour).Sub(time.Now())
+		return reconcile.Result{RequeueAfter: installJobDeletionRecheckDelay}, nil
+	case hivev1.ClusterProvisionStageFailed:
 		pLog.Debugf("ClusterProvision is %s. Nothing more to do", instance.Spec.Stage)
 		return reconcile.Result{}, nil
 	default:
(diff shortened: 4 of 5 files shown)


akhil-rane:
/hold

putting hold since this PR is blocked by [`SDA-1969`](https://issues.redhat.com/browse/SDA-1969)

config/controllers/hive_controllers_role.yaml
```diff
@@ -38,10 +38,18 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
```
staebler, line 41: The controllers should not need delete rights to the pods directly. All of the deletes should be to jobs instead of pods.

config/operator/operator_role.yaml
```diff
@@ -145,6 +145,7 @@ rules:
   - get
   - list
   - watch
+  - delete
```
staebler, line 148: The operator should not need to delete pods.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -484,6 +486,13 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, nil
 	}
 
+	// delete failed provisions which are more than 7 days old
+	existingProvisions, err := r.existingProvisions(cd, cdLog)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	r.deleteOldFailedProvisions(existingProvisions, cdLog)
```
staebler, line 494: Do this in the `if cd.Spec.Installed` block. There is already code to clean up ClusterProvisions as new ClusterProvisions are created.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -496,6 +505,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		}
 
 		cdLog.Debug("cluster is already installed, no processing of provision needed")
+		r.cleanupInstallPod(cd, cdLog)
```
staebler, line 508: The clusterdeployment controller should not be doing anything with the install pods. It should be the clusterprovision controller that deletes the install pod when the provision is successful.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1505,6 +1531,53 @@ func (r *ReconcileClusterDeployment) cleanupInstallLogPVC(cd *hivev1.ClusterDepl
 	return nil
 }
 
+// deleteInstallPods deletes the install pods for the given provision
+func (r *ReconcileClusterDeployment) deleteInstallPods(provision *hivev1.ClusterProvision, pLog log.FieldLogger) error {
```
staebler, line 1535: The controller should not be deleting the install pod directly. It should be deleting the install job, which will cascade the delete to the install pod. If this is done by the clusterprovision controller, then there is no need to search for the job: The ClusterProvision has a reference to the job in its status.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1505,6 +1531,53 @@ func (r *ReconcileClusterDeployment) cleanupInstallLogPVC(cd *hivev1.ClusterDepl
...
+		pLog.WithError(err).Error("unexpected error generating label selector to fetch install pods")
+		return err
+	}
+	labelSelector := labels.NewSelector().Add(*requirement)
+	listOptions := &client.ListOptions{
+		LabelSelector: labelSelector,
+	}
+
+	podList := &corev1.PodList{}
+	if err := r.List(context.TODO(), podList, listOptions); err != nil {
```
staebler, line 1550: For reference, this can be made simpler.
```suggestion
	podList := &corev1.PodList{}
	if err := r.List(
		context.TODO(),
		podList,
		client.InNamespace(provision.Namespace),
		client.MatchingLabels{"job-name": install.GetInstallJobName(provision)},
	); err != nil {
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1780,6 +1853,21 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
 	}
 }
 
+// deleteOldFailedProvisions deletes the failed provisions which are more than 7 days old
+func (r *ReconcileClusterDeployment) deleteOldFailedProvisions(provs []*hivev1.ClusterProvision, cdLog log.FieldLogger) {
+	cdLog.Infof("Deleting failed provisions which are more than 7 days old")
+	for _, provision := range provs {
+		failedCond := controllerutils.FindClusterProvisionCondition(provision.Status.Conditions, hivev1.ClusterProvisionFailedCondition)
```
staebler, line 1860: You do not need to use the condition. You can use `provision.Spec.Stage`.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1334,6 +1359,48 @@ func TestDeleteStaleProvisions(t *testing.T) {
...
+					provisions[i] = testDeleteOldFailedProvisions(time.Now(), i)
+				}
+			}
+			fakeClient := fake.NewFakeClient(provisions...)
+			rcd := &ReconcileClusterDeployment{
+				Client: fakeClient,
+				scheme: scheme.Scheme,
+			}
+			rcd.deleteOldFailedProvisions(getProvisions(fakeClient), log.WithField("test", "TestDeleteOldFailedProvisions"))
+			assert.Equal(t, tc.expectedNumberOfProvisionsAfterDeletion, len(getProvisions(fakeClient)), "unexpected provisions kept")
```
staebler, line 1399:
```suggestion
			assert.Len(t, getProvisions(fakeClient), tc.expectedNumberOfProvisionsAfterDeletion, "unexpected provisions kept")
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1474,6 +1541,7 @@ func testProvision() *hivev1.ClusterProvision {
 
 func testSuccessfulProvision() *hivev1.ClusterProvision {
 	provision := testProvision()
+	provision.ObjectMeta.Labels[jobNameLabel] = install.GetInstallJobName(provision)
```
staebler, line 1544: Why would the provision be labelled with the name of the job?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1503,6 +1571,28 @@ func testFailedProvisionTime(time time.Time) *hivev1.ClusterProvision {
 	return provision
 }
 
+func testDeleteOldFailedProvisions(time time.Time, attempt int) *hivev1.ClusterProvision {
```
staebler, line 1574: Rename the function to `testOldFailedProvision` since it is building a single old, failed provision. The name of the function as is states that the function is going to test deleting old, failed provisions.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
akhil-rane: condition also gives me `last transition time` which I use to precisely calculate when exactly status changed to failed. I can also use `creation timestamp` for the provision. Should not make much difference when we are considering 7 days time frame.

staebler: Ah, of course. Leave it as is.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
akhil-rane: My bad, its not needed

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -290,6 +290,15 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 
 func (r *ReconcileClusterProvision) reconcileSuccessfulJob(instance *hivev1.ClusterProvision, job *batchv1.Job, pLog log.FieldLogger) (reconcile.Result, error) {
 	pLog.Info("install job succeeded")
+	pLog.Info("deleting successful install job")
+	// deleting install job with background propagation policy to cascade delete install pod
+	deletePropagationBackground := metav1.DeletePropagationBackground
+	deleteOptions := &client.DeleteOptions{PropagationPolicy: &deletePropagationBackground}
+	if err := r.Delete(context.TODO(), job, deleteOptions); err != nil {
```
staebler, line 297:
```suggestion
	if err := r.Delete(context.TODO(), job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
```

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -290,6 +290,15 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
+		jLog := pLog.WithField("job", job.Name)
+		jLog.WithError(err).Error("error deleting successful install job")
```
staebler, line 299:
```suggestion
		pLog.WithField("job", job.Name).
			WithError(err).
			Log(controllerutils.LogLevel(err), "error deleting successful install job")
```

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -102,12 +102,13 @@ func TestClusterProvisionReconcile(t *testing.T) {
 			expectedStage: hivev1.ClusterProvisionStageInitializing,
 		},
 		{
-			name: "completed job",
+			name: "removed completed job after successful install",
```
staebler, line 105: Keep the old name. The test is doing more than ensuring that the completed job is removed. It is also validating that the stage is transitioned correctly.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -290,6 +290,15 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 
 func (r *ReconcileClusterProvision) reconcileSuccessfulJob(instance *hivev1.ClusterProvision, job *batchv1.Job, pLog log.FieldLogger) (reconcile.Result, error) {
 	pLog.Info("install job succeeded")
+	pLog.Info("deleting successful install job")
```
staebler, line 293: I think that we want to delete the job as a separate step after the ClusterProvision has transitioned to the Complete stage. Hive will need to delete jobs for existing ClusterProvisions that are already at the Complete stage.

The following test needs to be changed.
https://github.com/openshift/hive/blob/2bcc601e897555008f6f0e70df804da51462af3f/pkg/controller/clusterprovision/clusterprovision_controller_test.go#L130-L137

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -456,3 +459,19 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
 		},
 	}
 }
+
+// deleteInstallJob deletes the install job of a successful provision
+func (r *ReconcileClusterProvision) deleteInstallJob(provision *hivev1.ClusterProvision, pLog log.FieldLogger) (reconcile.Result, error) {
+	pLog.Info("deleting successful install job")
```
staebler, line 465: This should be debug not info. This will write on every reconcile of a completed provision.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -456,3 +459,19 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
...
+	job := &batchv1.Job{}
+	if err := r.Get(context.TODO(), types.NamespacedName{Name: provision.Status.JobRef.Name, Namespace: provision.Namespace}, job); err != nil {
+		pLog.WithError(err).Log(controllerutils.LogLevel(err), "could not get install job")
+		return reconcile.Result{}, err
```
staebler, line 469: Do not return an error when the job has already been deleted.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -456,3 +459,19 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
...
+	}
+	// deleting install job with background propagation policy to cascade delete install pod
+	if err := r.Delete(context.TODO(), job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
+		pLog.WithField("job", job.Name).WithError(err).Log(controllerutils.LogLevel(err), "error deleting successful install job")
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
```
staebler, line 476: Clear the job reference after the job has been deleted.

pkg/controller/clusterprovision/clusterprovision_controller.go
akhil-rane: Instead do we want to run `deleteInstallJob` only if JobRef is present? If we are clearing JobRef than we don't have to run this code on every reconcile

staebler: Yes. If we are only running this until the job reference is removed, then Info is a good log level.

staebler:
I think that we need to do some work on the e2e test given that the install pod is going to be deleted right away when it succeeds. The e2e test waits for the install job to be successful. I don't think that we need to rely on looking at the install job at all. I think we can accept a successful ClusterProvision or an installed ClusterDeployment as sufficient

https://github.com/openshift/hive/blob/2bcc601e897555008f6f0e70df804da51462af3f/hack/e2e-test.sh#L213-L218

pkg/controller/clusterprovision/clusterprovision_controller.go
akhil-rane: I think returning error here is fine since we will executing this code only once. Correct me If I am wrong

pkg/controller/clusterprovision/clusterprovision_controller.go
akhil-rane: Done

hack/e2e-test.sh
```diff
@@ -4,6 +4,11 @@ set -e
 
 max_tries=60
 sleep_between_tries=10
+# Set timeout for the cluster deployment to install
+# timeout = sleep_between_cluster_deployment_status_checks * max_cluster_deployment_status_checks
+max_cluster_deployment_status_checks=90
+sleep_between_cluster_deployment_status_checks="1m"
```
akhil-rane, line 10: Not sure what values to set here.

hack/e2e-test.sh
```diff
@@ -208,13 +213,26 @@ echo ""
...
 
-echo "Waiting for install job to start and complete"
-
+echo "Waiting for the ClusterDeployment ${CLUSTER_NAME} to install"
 INSTALL_RESULT=""
-if go run "${SRC_ROOT}/contrib/cmd/waitforjob/main.go" "${CLUSTER_NAME}" "install"
+
+i=1
+while [ $i -le ${max_cluster_deployment_status_checks} ]; do
+  IS_CLUSTER_DEPLOYMENT_INSTALLED=$(oc get cd ${CLUSTER_NAME} | awk 'NR == 2 {print $4}')
```
staebler, line 221: Need to specify the namespace.

staebler, line 221: I would rather we use `jq` instead of `awk`. We do not consider the printer columns to be a hard API.

hack/e2e-test.sh
staebler: These seem like decent enough values.

hack/e2e-test.sh
```diff
@@ -208,13 +213,26 @@ echo ""
...
+  if [[ "${IS_CLUSTER_DEPLOYMENT_INSTALLED}" == "true" ]] ; then
+    INSTALL_RESULT="SUCCESS"
```
staebler, line 223: Changing to all capitals "SUCCESS" is going to make the later check fail.
https://github.com/openshift/hive/blob/d9348de8ae632eb236aec1b24bcd078d903e7867/hack/e2e-test.sh#L241

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1771,6 +1778,21 @@ func (r *ReconcileClusterDeployment) deleteStaleProvisions(provs []*hivev1.Clust
 	}
 }
 
+// deleteOldFailedProvisions deletes the failed provisions which are more than 7 days old
+func (r *ReconcileClusterDeployment) deleteOldFailedProvisions(provs []*hivev1.ClusterProvision, cdLog log.FieldLogger) {
+	cdLog.Infof("Deleting failed provisions which are more than 7 days old")
```
staebler, line 1783: This should be debug level since it is run on every reconcile of an installed ClusterDeployment.

pkg/controller/clusterprovision/clusterprovision_controller.go
staebler: If we return an error here when the job has already been deleted, then `provision.Status.JobRef` will never be cleared.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -456,3 +462,25 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
...
+	pLog.Info("deleting successful install job")
+	job := &batchv1.Job{}
+	if err := r.Get(context.TODO(), types.NamespacedName{Name: provision.Status.JobRef.Name, Namespace: provision.Namespace}, job); err != nil {
+		pLog.WithError(err).Log(controllerutils.LogLevel(err), "could not get install job")
+	}
+	// deleting install job with background propagation policy to cascade delete install pod
+	if err := r.Delete(context.TODO(), job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
+		pLog.WithField("job", job.Name).WithError(err).Log(controllerutils.LogLevel(err), "error deleting successful install job")
+		return reconcile.Result{}, err
+	}
```
staebler, line 477: `Delete` can only be called if there was not an error returned from `Get`. Otherwise, the `Delete` call will error.

```suggestion
	switch err := r.Get(context.TODO(), types.NamespacedName{Name: provision.Status.JobRef.Name, Namespace: provision.Namespace}, job); {
	case apierrors.IsNotFound(err):
		pLog.Info("install job has already been deleted")
	case err != nil:
		pLog.WithError(err).Log(controllerutils.LogLevel(err), "could not get install job")
		return reconcile.Result{}, err
	default:
		// deleting install job with background propagation policy to cascade delete install pod
		if err := r.Delete(context.TODO(), job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
			pLog.WithField("job", job.Name).WithError(err).Log(controllerutils.LogLevel(err), "error deleting successful install job")
			return reconcile.Result{}, err
		}
	}
```

akhil-rane:
> These changes look good to me.
> 
> I think that there is still an open question of whether our users would prefer to keep a success install pod around for 24 hours or whether the logs stored in the ClusterProvision are sufficient.

Looks like we need to keep successful install pod for 24 hours. I will make the changes.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -173,7 +173,13 @@ func (r *ReconcileClusterProvision) Reconcile(request reconcile.Request) (reconc
 			return r.reconcileRunningJob(instance, pLog)
 		}
 		return r.transitionStage(instance, hivev1.ClusterProvisionStageFailed, "NoJobReference", "Missing reference to install job", pLog)
-	case hivev1.ClusterProvisionStageComplete, hivev1.ClusterProvisionStageFailed:
+	case hivev1.ClusterProvisionStageComplete:
+		pLog.Debugf("ClusterProvision is %s", instance.Spec.Stage)
+		if instance.Status.JobRef != nil && time.Since(instance.CreationTimestamp.Time) > (24*time.Hour) {
+			return r.deleteInstallJob(instance, pLog)
+		}
+		return reconcile.Result{}, nil
```
staebler, line 181: If the install job needs to be deleted in the future, the `reconcile.Result` should have `RequeueAfter` set to the duration remaining until the delete.

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -128,12 +128,14 @@ func TestClusterProvisionReconcile(t *testing.T) {
 			expectedFailReason: unknownReason,
 		},
 		{
-			name: "keep job after success",
+			name: "removed job after success",
```
staebler, line 131: Re-work this test case so that we test keeping the job within 24 hours and deleting the job after 24 hours.

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -198,7 +210,11 @@ func TestClusterProvisionReconcile(t *testing.T) {
 
 			result, err := rcp.Reconcile(reconcileRequest)
 
-			assert.Equal(t, test.expectedReconcileResult, result, "unexpected reconcile result")
+			if test.expectedRequeueAfter {
+				assert.NotEqual(t, time.Duration(0), result.RequeueAfter, "expected requeue after duration in reconcile result")
+			} else {
+				assert.Equal(t, time.Duration(0), result.RequeueAfter, "expected no requeue after duration in reconcile result")
+			}
```
akhil-rane, line 217: It is difficult to set exact expected RequeueAfter duration so just checking whether it is greater than 0 or not. Can it be improved?

staebler: Store the time before the test structs are created. Store the time after the reconcile is run. Use that to determine a window of values that the RequeueAfter value should lie within.

akhil-rane: Done

akhil-rane:
@staebler this is up for review again

akhil-rane:
/hold cancel

Removing hold as discussed in the scrum

PR #960, opened 2020-04-28
vendor: update go.sum

staebler:
The go.sum file is hashes for modules where the version does not
have `/go.sum` appended at the end.

Ran the following.
```
  go clean --modcache
  rm go.sum
  go mod tidy
```


staebler:
These changes were included in #952 

/close

PR #961, opened 2020-04-28
Detect serving cert changes in operator instead of hiveadmission.

dgoodwin:
Trying an alternate approach for detecting changes to the serving cert
by monitoring the secret in the hive-operator, and on any change we
reconcile. Reconcile will then hash the secret data and store in an
annotation on the pod template in the hiveadmission deployment. Any
change to this value will result in a new rollout.

### DIFF
--- a/pkg/operator/hive/hive_controller.go
+++ b/pkg/operator/hive/hive_controller.go
@@ -32,11 +32,14 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -52,6 +55,9 @@ const (
 
 	// HiveOperatorNamespaceEnvVar is the environment variable we expect to be given with the namespace the hive-operator is running in.
 	HiveOperatorNamespaceEnvVar = "HIVE_OPERATOR_NS"
+
+	// watchResyncInterval is used for a couple handcrafted watches we do with our own informers.
+	watchResyncInterval = 30 * time.Minute
 )
 
 // Add creates a new Hive Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -62,7 +68,7 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileHiveConfig{Client: mgr.GetClient(), scheme: mgr.GetScheme(), restConfig: mgr.GetConfig()}
+	return &ReconcileHiveConfig{Client: mgr.GetClient(), scheme: mgr.GetScheme(), restConfig: mgr.GetConfig(), mgr: mgr}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -72,6 +78,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+
+	// Provide a ref to the controller on the reconciler, which is used to establish a watch on
+	// secrets in the hive namespace, which isn't known until we have a HiveConfig.
+	r.(*ReconcileHiveConfig).ctrlr = c
 
 	r.(*ReconcileHiveConfig).kubeClient, err = kubernetes.NewForConfig(mgr.GetConfig())
 	if err != nil {
@@ -121,7 +131,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err == nil {
 		log.Debugf("the %s namespace exists, setting up a watch for configmaps on it", managedConfigNamespace)
 		// Create an informer that only listens to events in the OpenShift managed namespace
-		kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(r.(*ReconcileHiveConfig).kubeClient, 10*time.Minute, kubeinformers.WithNamespace(managedConfigNamespace))
+		kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(r.(*ReconcileHiveConfig).kubeClient, watchResyncInterval, kubeinformers.WithNamespace(managedConfigNamespace))
 		configMapInformer := kubeInformerFactory.Core().V1().ConfigMaps().Informer()
 		mgr.Add(&informerRunnable{informer: configMapInformer})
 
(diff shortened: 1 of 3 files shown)


pkg/operator/hive/hive_controller.go
```diff
@@ -201,24 +208,26 @@ var _ reconcile.Reconciler = &ReconcileHiveConfig{}
...
+	dynamicClient                     dynamic.Interface
+	restConfig                        *rest.Config
+	hiveImage                         string
+	hiveOperatorNamespace             string
+	hiveImagePullPolicy               corev1.PullPolicy
+	syncAggregatorCA                  bool
+	managedConfigCMLister             corev1listers.ConfigMapLister
+	ctrlr                             controller.Controller
+	servingCertSecretWatchEstablished bool
+	mgr                               manager.Manager
```
dgoodwin, line 225: Last three fields are the new ones.

pkg/operator/hive/hive_controller.go
```diff
@@ -246,13 +255,63 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+		// force a deployment rollout.
+		err = r.ctrlr.Watch(&source.Informer{Informer: secretsInformer}, handler.Funcs{
+			CreateFunc: func(e event.CreateEvent, q workqueue.RateLimitingInterface) {
+				log.Info("eventHandler CreateFunc")
+				q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Name: hiveConfigName}})
+			},
+			UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+				log.Info("eventHandler UpdateFunc")
+				q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Name: hiveConfigName}})
+			},
```
dgoodwin, line 285: Tested and this does appear to work even though it's configured later than most watches are. It does not fire for secrets in other namespaces, and it fires for secrets in this namespace but filters out via the predicate.

pkg/operator/hive/hiveadmission.go
```diff
@@ -142,8 +140,28 @@ func (r *ReconcileHiveConfig) deployHiveAdmission(hLog log.FieldLogger, h *resou
...
+		if err := r.Client.Get(context.Background(), types.NamespacedName{Namespace: hiveNSName, Name: hiveAdmissionServingCertSecretName}, servingCertSecret); err != nil {
+			hLog.WithError(err).WithField("secretName", hiveAdmissionServingCertSecretName).Log(
+				controllerutils.LogLevel(err), "error getting serving cert secret")
+		}
+		hLog.Info("Running on OpenShift 4, hashing serving cert secret onto a hiveadmission deployment annotation")
+		certSecretHash := computeSecretDataHash(servingCertSecret.Data)
+		if hiveAdmDeployment.Spec.Template.Annotations == nil {
+			hiveAdmDeployment.Spec.Template.Annotations = map[string]string{}
+		}
+		hiveAdmDeployment.Spec.Template.Annotations[servingCertSecretHashAnnotation] = certSecretHash
```
staebler, line 155: Does this need to be gated on whether Hive is running on OpenShift 4? Even when Hive is running on OpenShift 3.11 or another kube flavor, the same serving cert is in use.

pkg/operator/hive/hive_controller.go
```diff
@@ -245,13 +254,63 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+		}, predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				// WARNING: can't check namespace for same reason as above, we don't have a HiveConfig
+				// to definitively know what namespace Hive is deployed to.
+				log.WithField("predicateResponse", e.Meta.GetName() == hiveAdmissionServingCertSecretName).Debug("secret CreateEvent")
+				return e.Meta.GetName() == hiveAdmissionServingCertSecretName
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				// WARNING: can't check namespace for same reason as above, we don't have a HiveConfig
```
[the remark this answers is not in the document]
dgoodwin: My bad, this was leftover from before I tried moving it into the reconcile loop itself.

pkg/operator/hive/hiveadmission.go
dgoodwin: Good catch, I did not realize that.

dgoodwin:
Updated. Certs function is restored.

For the deletion of HiveConfig, when reconcile and find no HiveConfig "hive", the var is now cleared. The watch lives on, I can't see any way to delete it, so if you then go edit hiveadmission-serving-cert we will reconcile, lookup HiveConfig, log that it's deleted and return.

If you reinstall to the same ns the watch is recreated but it does not appear to trigger duplicate reconciles.

If you reinstall to *another* namespace, you now have two watches for the secrets in either ns. Editing either will trigger a normal reconcile. I think we can live with this as it's harmless, extremely unlikely to occur, and fixed as soon as you restart the pod.

pkg/operator/hive/hive_controller.go
```diff
@@ -245,13 +255,19 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, nil
 	}
 
+	hiveNSName := getHiveNamespace(instance)
+
+	err = r.establishSecretWatch(hLog, hiveNSName)
```
staebler, line 260: `if err := ...; err != nil {`

pkg/operator/hive/hive_controller.go
```diff
@@ -337,6 +353,54 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
 }
 
+func (r *ReconcileHiveConfig) establishSecretWatch(hLog *log.Entry, hiveNSName string) error {
+	// We need to establish a watch on Secret in the Hive namespace, one time only. We do not know this namespace until
+	// we have a HiveConfig.
+	if !r.servingCertSecretWatchEstablished {
+		hLog.WithField("namespace", hiveNSName).Info("establishing watch on secrets in hive namespace")
+
+		// Create an informer that only listens to events in the OpenShift managed namespace
+		kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(r.kubeClient, 10*time.Minute,
```
staebler, line 363: 10 minutes seems like a very low resync period. Does that controller-runtime end up setting a larger resync period? Or will the hive-operator be resync HiveConfig every 10 minutes?

pkg/operator/hive/hive_controller.go
```diff
@@ -337,6 +353,54 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+			kubeinformers.WithNamespace(hiveNSName))
+		secretsInformer := kubeInformerFactory.Core().V1().Secrets().Informer()
+		err := r.mgr.Add(&informerRunnable{informer: secretsInformer})
```
staebler, line 366: `if err := ...; err != nil {`

pkg/operator/hive/hive_controller.go
```diff
@@ -337,6 +353,54 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+		if err != nil {
+			hLog.WithError(err).Error("error adding secret informer to manager")
+			return err
+		}
+
+		// Watch Secrets in hive namespace, so we can detect changes to the hiveadmission serving cert secret and
+		// force a deployment rollout.
+		err = r.ctrlr.Watch(&source.Informer{Informer: secretsInformer}, handler.Funcs{
+			CreateFunc: func(e event.CreateEvent, q workqueue.RateLimitingInterface) {
+				log.Debug("eventHandler CreateFunc")
```
staebler, line 376: Should this use `hLog` instead of `log`? It would also be nice if the message were a bit more descriptive--at least mentioning that it is the event handler for the secret watcher.

pkg/operator/hive/hive_controller.go
dgoodwin: I picked this up from the configmap watch for kube-apiserver-aggregator-client-ca above. I can bump them both to 30.

pkg/operator/hive/hive_controller.go
dgoodwin: Yup another remnant from when it was outside the reconciler. Updated.

dgoodwin:
Updated.

pkg/operator/hive/hive_controller.go
```diff
@@ -231,6 +243,7 @@ func (r *ReconcileHiveConfig) Reconcile(request reconcile.Request) (reconcile.Re
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
 			hLog.Debug("HiveConfig not found, deleted?")
+			r.servingCertSecretWatchEstablished = false
```
staebler, line 246: This should only be cleared if the `HiveConfig` deleted was named `hive`. Maybe we should move the name check to before the `Get`.

dgoodwin: Moved above and changed to not use instance.Name.

pkg/operator/hive/hive_controller.go
```diff
@@ -200,29 +210,40 @@ var _ reconcile.Reconciler = &ReconcileHiveConfig{}
...
 	hLog := log.WithField("controller", "hive")
 	hLog.Info("Reconciling Hive components")
 
 	// Fetch the Hive instance
 	instance := &hivev1.HiveConfig{}
+
+	// We only support one HiveConfig per cluster, and it must be called "hive". This prevents installing
+	// Hive more than once in the cluster.
+	if request.NamespacedName.Name != hiveConfigName {
+		hLog.WithField("hiveConfig", request.NamespacedName.Name).Warn(
```
staebler, line 242: We can leave as is, but you do not need to include `NamespacedName`.

```suggestion
	if request.Name != hiveConfigName {
		hLog.WithField("hiveConfig", request.Name).Warn(
```

dgoodwin:
My long run test was successful, pods rotated and API stayed up without error.

PR #962, opened 2020-04-30
Fix syncset unapplied metrics not clearing.

dgoodwin:
When a selector sync set has failing instances, but all failure clear
and all instance has been applied, the metric was not being cleared.

This change will not clear unapplied metrics for any currently known
selector sync set if they do not have any failing instances.

NOTE: we still would have an issue if the selector sync set was entirely
deleted. This metric would not clear until the pod is restarted.

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -334,6 +334,15 @@ func (mc *Calculator) calculateSelectorSyncSetMetrics(mcLog log.FieldLogger) {
 	}
 	for k, v := range sssInstancesTotal {
 		metricSelectorSyncSetClustersTotal.WithLabelValues(k).Set(float64(v))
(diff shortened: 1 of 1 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -334,6 +334,15 @@ func (mc *Calculator) calculateSelectorSyncSetMetrics(mcLog log.FieldLogger) {
 	}
 	for k, v := range sssInstancesTotal {
 		metricSelectorSyncSetClustersTotal.WithLabelValues(k).Set(float64(v))
+		// If this selector sync set currently has no unapplied instances, ensure any past unapplied metric is cleared:
+		if unapplied := sssInstancesUnappliedTotal[k]; unapplied == 0 {
```
staebler, line 338:
```suggestion
		if sssInstancesUnappliedTotal[k] == 0 {
```

PR #964, opened 2020-04-30
add new operator bundle generate script for operatorhub deploys

gregsheremeta:
Add new generate-operator-bundle-operatorhub.py script. This script
is adapted from the existing hack/generate-operator-bundle.py, and
is capable of generating a bundle for both operatorhub and OSD. For
now we're only using it for operatorhub and keeping the OSD pipeline
unchanged. At a later date we'll adapt the OSD pipeline and rename
this script.

/cc @dgoodwin

### DIFF
--- a/hack/generate-operator-bundle-operatorhub.py
+++ b/hack/generate-operator-bundle-operatorhub.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python
+#
+# Generate an operator bundle for publishing to OLM. Copies appropriate files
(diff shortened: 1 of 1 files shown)


gregsheremeta:
extracting this from #931 and closing that. It got a little gross in there after we discovered some pipeline issues. I'm going to take a fresh approach to the pipeline, but want to at least get this script in the repo since I use it every week.

gregsheremeta:
@dgoodwin you already fully reviewed this in #931. No changes to it.

PR #965, opened 2020-04-30
WIP: New osd pipeline

gregsheremeta:
WIP prototype. Just interested if you think it could work.
ignore generate-operator-bundle-operatorhub.py, I'm just basing on that commit.

new OpenShift Dedicated pipeline that builds bare-minimum catalogs

Create an OLM CatalogSource / registry with only two bundles --
the currently deployed version (old), and the version to upgrade to (new).
The old version previously replaced an older version when it was installed,
but here we strip that 'replaces:' from its CSV to make it the new head
of the replaces chain on every deploy.

Previously we had a difficult-to-maintain chain of every version ever:

```
v1 <---- channel head
and next time...
v1 <---- v2 (replaces v1) <---- channel head
and next time...
v1 <---- v2 (replaces v1) <---- v3 (replaces v2) <---- channel head
and next time...
v1 <---- v2 (replaces v1) <---- v3 (replaces v2) <---- v4 (replaces v3) <---- channel head
```

Now:

```
v1 <---- v2 (replaces v1) <---- channel head
and next time...
v2 <---- v3 (replaces v2) <---- channel head
and next time...
v3 <---- v4 (replaces v3) <---- channel head
and so on.
```

/cc @dgoodwin @jmelis @jharrington22

### DIFF
--- a/hack/app_sre_create_image_catalog_new.sh
+++ b/hack/app_sre_create_image_catalog_new.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+set -exv
+
+BRANCH_CHANNEL="production"
+QUAY_IMAGE="quay.io/gshereme/hive"
+
+if [[ -z "$OLD_DEPLOYMENT" ]]; then
+    echo "OLD_DEPLOYMENT must be defined" 1>&2
+    exit 1
+fi
+
+if [[ -z "$NEW_DEPLOYMENT" ]]; then
+    echo "NEW_DEPLOYMENT must be defined" 1>&2
+    exit 1
+fi
+
+OLD_BUNDLE_DIR=$(mktemp -d)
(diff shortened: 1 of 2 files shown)


gregsheremeta:
/hold
probably never merging this. Prototype only.

hack/app_sre_create_image_catalog_new.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+set -exv
+
+BRANCH_CHANNEL="production"
+QUAY_IMAGE="quay.io/gshereme/hive"
+
+OLD_DEPLOYMENT=d80f25ace2566e6724b3825726bd82820c6b2588
```
gregsheremeta, line 8: temporary hardcode. TODO change to read the current deployment's SHA

PR #968, opened 2020-05-04
remove operator bundle and version.txt

gregsheremeta:
Now that we are supporting multiple operator channels that
build from branches other than master (`ocm-1.0`), it doesn't
make sense to maintain the bundle here in `master`. Remove it
and let https://github.com/operator-framework/community-operators
be the source of truth.

/cc @dgoodwin


dgoodwin:
RIP

/lgtm

PR #974, opened 2020-05-05
Add annotation based option to set additional waits to complete cluster provision

akhil-rane:
This PR provides an option to set additional 'wait-for install-complete'
executions after a timed out cluster initialization by using annotation on the
Cluster Deployments

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -89,26 +89,27 @@ var (
 // InstallManager coordinates executing the openshift-install binary, modifying
 // generated assets, and uploading artifacts to the kube API after completion.
 type InstallManager struct {
-	log                      log.FieldLogger
-	LogLevel                 string
-	WorkDir                  string
-	LogsDir                  string
-	ClusterID                string
-	ClusterName              string
-	ClusterProvisionName     string
-	Namespace                string
-	InstallConfigMountPath   string
-	PullSecretMountPath      string
-	ManifestsMountPath       string
-	DynamicClient            client.Client
-	cleanupFailedProvision   func(dynamicClient client.Client, cd *hivev1.ClusterDeployment, infraID string, logger log.FieldLogger) error
-	updateClusterProvision   func(*hivev1.ClusterProvision, *InstallManager, provisionMutation) error
-	readClusterMetadata      func(*hivev1.ClusterProvision, *InstallManager) ([]byte, *installertypes.ClusterMetadata, error)
-	uploadAdminKubeconfig    func(*hivev1.ClusterProvision, *InstallManager) (*corev1.Secret, error)
-	uploadAdminPassword      func(*hivev1.ClusterProvision, *InstallManager) (*corev1.Secret, error)
-	readInstallerLog         func(*hivev1.ClusterProvision, *InstallManager, bool) (string, error)
-	waitForProvisioningStage func(*hivev1.ClusterProvision, *InstallManager) error
-	isGatherLogsEnabled      func() bool
+	log                              log.FieldLogger
+	LogLevel                         string
+	WorkDir                          string
+	LogsDir                          string
+	ClusterID                        string
+	ClusterName                      string
+	ClusterProvisionName             string
+	Namespace                        string
+	InstallConfigMountPath           string
+	PullSecretMountPath              string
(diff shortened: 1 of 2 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -89,26 +89,27 @@ var (
...
+	DynamicClient                    client.Client
+	cleanupFailedProvision           func(dynamicClient client.Client, cd *hivev1.ClusterDeployment, infraID string, logger log.FieldLogger) error
+	updateClusterProvision           func(*hivev1.ClusterProvision, *InstallManager, provisionMutation) error
+	readClusterMetadata              func(*hivev1.ClusterProvision, *InstallManager) ([]byte, *installertypes.ClusterMetadata, error)
+	uploadAdminKubeconfig            func(*hivev1.ClusterProvision, *InstallManager) (*corev1.Secret, error)
+	uploadAdminPassword              func(*hivev1.ClusterProvision, *InstallManager) (*corev1.Secret, error)
+	readInstallerLog                 func(*hivev1.ClusterProvision, *InstallManager, bool) (string, error)
+	waitForProvisioningStage         func(*hivev1.ClusterProvision, *InstallManager) error
+	isGatherLogsEnabled              func() bool
+	waitForInstallCompleteExecutions int
```
akhil-rane, line 112: This is the new field added to the InstallManager struct

pkg/installmanager/installmanager.go
```diff
@@ -626,8 +635,10 @@ func (m *InstallManager) provisionCluster() error {
 
 	if err := m.runOpenShiftInstallCommand("create", "cluster"); err != nil {
 		if m.isBootstrapComplete() {
-			m.log.WithError(err).Warn("provisioning cluster failed after completing bootstrapping, waiting longer for install to complete")
-			err = m.runOpenShiftInstallCommand("wait-for", "install-complete")
+			for i := 0; i < m.waitForInstallCompleteExecutions; i++ {
+				m.log.WithError(err).Warn("provisioning cluster failed after completing bootstrapping, waiting longer for install to complete")
```
akhil-rane, line 639: Not sure whether we can have a different message here for every additional wait

pkg/installmanager/installmanager.go
```diff
@@ -383,6 +384,14 @@ func (m *InstallManager) Run() error {
 	}
 
 	m.log.Info("provisioning cluster")
+
+	if waitForInstallCompleteExecutions, ok := cd.Annotations[constants.WaitForInstallCompleteExecutionsAnnotation]; ok {
+		m.waitForInstallCompleteExecutions, err = strconv.Atoi(waitForInstallCompleteExecutions)
+		if err != nil {
+			m.log.WithError(err).Error("error fetching additional wait-for install-complete executions")
```
dgoodwin, line 391: Suggest Errorf and "error parsing integer from %s annotation"

pkg/installmanager/installmanager.go
dgoodwin: That's a good idea, would Warnf("[current msg] (%d/%d)", i+1, waitForInstallCompleteExecutions) work?

staebler: Put the current iteration being performed in a log field.
```suggestion
				m.log.WithField("waitIteration", i).WithError(err).Warn("provisioning cluster failed after completing bootstrapping, waiting longer for install to complete")
```

pkg/installmanager/installmanager.go
staebler, line 391: Include the value of the annotation in the log message.
```suggestion
			m.log.WithField("value", waitForInstallCompleteExecutions).WithError(err).Error("error fetching additional wait-for install-complete executions")
```

pkg/installmanager/installmanager.go
```diff
@@ -626,8 +635,10 @@ func (m *InstallManager) provisionCluster() error {
 
 	if err := m.runOpenShiftInstallCommand("create", "cluster"); err != nil {
 		if m.isBootstrapComplete() {
```
staebler, line 637: Add a check here for if `m.waitForInstallCompleteExecutions` is greater than zero. If it is not, then there is no need to run `m.isBootstrapComplete()`. That will keep the logs a bit cleaner.

PR #976, opened 2020-05-07
Backport several fixes for OCM

dgoodwin:
Several commits here to backport fixes that were lost when we recreated the OCM branch, or added after, which are required for OCM.


dgoodwin:
With exception of the vendor commit which was to resolve problems after cherry-picking to a different gomod base, these changes have already been reviewed, this PR is primarily to get a clean CI run.

Issue #833, opened 2020-02-17
Update hack scripts to work with latests versions of Kind


jparrill:
Hi there,

There is some modifications needed to be done over the `hack/....sh` files to make it work wih kind latests version.

E.G: `kind get kubeconfig-path` does not exists anymore and also the kustomize part is not fully compatible with the latests k8s version.

dgoodwin:
Yes I saw breakage with latest kind awhile back, nobody has had time to investigate yet. We've been using 0.5.1 in meantime.

jparrill:
I'm still working in this PR in order to solve the kind-control-plane machine issue that occurs on hive pull image:

```
root@kind-control-plane:/var/lib/kubelet# crictl pull registry.svc.ci.openshift.org/openshift/hivev1:hive
FATA[0000] pulling image failed: rpc error: code = Unknown desc = failed to pull and unpack image "registry.svc.ci.openshift.org/openshift/hivev1:hive": failed to resolve reference "registry.svc.ci.openshift.org/openshift/hivev1:hive": no scope specified for token auth challenge 
```

It's supose to not have any problem pull the image from podman/docker, but seems that there is something there con crictl tool. The related PR is: https://github.com/openshift/hive/pull/834

jparrill:
I think that this could be relevant to know the behaviour of Kind and Hive together: https://github.com/openshift/hive/pull/834#issuecomment-587623422

PR #978, opened 2020-05-07
Kind 0.8.1 devenv

dgoodwin:
New script is based off an example in kind.

Been trying to get this to work but it's stuck communicating with the insecure registry, no route to host despite supposedly being on the network. I've exec'd into the kind control plane container and using curl, neither localhost, nor the container name (kind-registry), nor the ip seem to work. It just can't reach that container.

Any advice welcome, going to keep trying to poke at this.

Still haven't tried it with podman.

Fixes: #833


dgoodwin:
Script seems to work once I did this hack: https://github.com/kubernetes-sigs/kind/issues/1547#issuecomment-623756313

It's getting harder and harder to run Docker on Fedora, looks like there are some complications with firewalld and docker networking that require changing the backend for firewalld to work.

So this script replaces both of our current ones for kind and registry into one much smaller and simpler script.

Next step podman.

dgoodwin:
hive-operator appears to be successful, but hotloops and pins the cpu. Investigating.

dgoodwin:
Hotloop fixed, details in commit, some interesting complications may be coming with Kube 1.18+ due to server side apply.

Updated docs.

hack/create-kind-cluster.sh
```diff
@@ -1,197 +1,42 @@
...
-export KUBECONFIG=\"${KUBECONFIG}\"
-
-Then you can overwrite ${kubeconfig} if you prefer using:
-
-kubectl config view --flatten > ${kubeconfig}
-unset KUBECONFIG
-"
-EOF
-fi
+# tell https://tilt.dev to use the registry
```
staebler, line 38: Is this section required? If so, we should remove the references to tilt.

docs/developing.md
```diff
@@ -67,17 +67,11 @@ git clone https://github.com/openshift/hive.git
...
 This approach requires [Docker](https://docs.docker.com/install). At present we do not have kind working with podman.
 
-Deploy a local insecure registry container, and configure your host docker daemon to be able to use it:
-
-```bash
-./hack/create-insecure-registry.sh
-```
-
-Create a kind cluster named 'hive' to deploy to. You can create as many kind clusters as you need.:
+Create a local insecure registry (if one does not already exist) and then a kind cluster named 'hive' to deploy to. You can create as many kind clusters as you need and you have sufficient RAM.
```
staebler, line 74: The part about sufficient RAM could use some grammatical work. I do not think that it is necessary to call out specifically RAM.
```suggestion
Create a local insecure registry (if one does not already exist) and then a kind cluster named 'hive' to deploy to. You can create multiple kind clusters. Each kind cluster needs a unique name, which is specified by the argument passed to the script.
```

hack/create-kind-cluster.sh
dgoodwin: Didn't even realize, no it does not seem necessary, removing.

dgoodwin:
Updates pushed and WIP removed. Matthew and Joel have confirmed this works for them so I think we're ok to proceed.

Added some additional notes on the Fedora hacks that may be required.

dgoodwin:
Needs approval, should be ready to go.

docs/developing.md
```diff
@@ -67,37 +67,36 @@ git clone https://github.com/openshift/hive.git
...
-1dc8a3c59d84        registry:2             "/entrypoint.sh /etc…"   2 weeks ago         Up 8 days           0.0.0.0:5000->5000/tcp                 registry
+```
+Error response from daemon: container fcca36a26da5601d6453c0c53ab5909eb6ca8ffe42de0f8634dd7213f107cef0 is not connected to network kind
 ```
 
-Configure kubectl/oc to talk to your new cluster:
+`docker ps` should now show you a "registry" and a "hive" container running.
 
 ```bash
-export KUBECONFIG="$(kind get kubeconfig-path --name="hive")"
```
suhanime, line 98: The new kind does not have `get kubeconfig-path` so also need to change the docs in Adopting clusterdeployments section

dgoodwin: Thanks, I replaced it with /path/to/admin/kubeconfig.

dgoodwin:
PR updated, would love to get merged before freeze!

docs/developing.md
```diff
@@ -130,7 +129,7 @@ To create a kind cluster and adopt:
 
 ```bash
 ./hack/create-kind-cluster.sh cluster1
-bin/hiveutil create-cluster --base-domain=new-installer.openshift.com kind-cluster1 --adopt --adopt-admin-kubeconfig=$(kind get kubeconfig-path --name="cluster1") --adopt-infra-id=fakeinfra --adopt-cluster-id=fakeid
+bin/hiveutil create-cluster --base-domain=new-installer.openshift.com kind-cluster1 --adopt --adopt-admin-kubeconfig=/path/to/cluster/admin/kubeconfig get kubeconfig-path --name="cluster1") --adopt-infra-id=fakeinfra --adopt-cluster-id=fakeid
```
staebler, line 132: Remove the following.
```
get kubeconfig-path --name="cluster1")
```

dgoodwin: Arghhh sorry, updated again.

PR #982, opened 2020-05-11
Experimental kind podman support for hive development.

dgoodwin:


hack/create-kind-podman-cluster.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# Adapted from: https://github.com/kubernetes-sigs/kind/blob/master/site/static/examples/kind-with-registry.sh
+
+set -o errexit
+
+cluster_name="${1:-hive}"
+
+# create registry container unless it already exists
+reg_name='kind-podman-registry'
+reg_port='5001'
```
dgoodwin, line 16: Should be 5000 I'm just testing along side docker at the moment.

hack/create-kind-podman-cluster.sh
```diff
@@ -0,0 +1,47 @@
...
+    endpoint = ["http://${reg_ip}:5000"]
+EOF
+
+# connect the registry to the cluster network
+#podman network disconnect "kind" "${reg_name}" || true
+#podman network connect "kind" "${reg_name}"
+
+sudo cp /root/.kube/config ~/.kube/${cluster_name}.kubeconfig
+sudo chown $USER ~/.kube/${cluster_name}.kubeconfig
+echo "Kubeconfig written to $HOME/.kube/${cluster_name}.kubeconfig"
```
dgoodwin, line 46: This could probably be better...

dgoodwin:
Current status, this gets hive running but we cannot launch an install, podman will not let us exec/fork to run openshift-install in our install pods, because the binary is copied into /output by the sidecar container and this is mounted with noexec.

Could we perhaps copy to /bin in installmanager? cc @joelddiaz

joelddiaz:
> Current status, this gets hive running but we cannot launch an install, podman will not let us exec/fork to run openshift-install in our install pods, because the binary is copied into /output by the sidecar container and this is mounted with noexec.
> 
> Could we perhaps copy to /bin in installmanager? cc @joelddiaz

I'd like to spend some time with the kind/docker setup to compare with how things are being brought up there compared to the kind/podman setup. But certainly moving the binary to mountpoint that isn't noexec will work around the immedate issue.

dgoodwin:
Commit added to copy the binaries to /bin. This is working for me locally, I have an install running now.

csrwng:
Wonder if we should have a copy retry:
```
time="2020-05-12T18:51:51Z" level=debug msg="checking for SSH private key" installID=2mqhtwnz
time="2020-05-12T18:51:51Z" level=info msg="unable to initialize host ssh key" error="cannot configure SSH agent as SSH_PRIV_KEY_PATH is unset or empty" installID=2mqhtwnz
time="2020-05-12T18:51:51Z" level=info msg="waiting for files to be available: [/output/openshift-install /output/oc]" installID=2mqhtwnz
time="2020-05-12T18:51:51Z" level=info msg="found file" installID=2mqhtwnz path=/output/openshift-install
time="2020-05-12T18:51:51Z" level=info msg="found file" installID=2mqhtwnz path=/output/oc
time="2020-05-12T18:51:51Z" level=info msg="all files found, ready to proceed" installID=2mqhtwnz
time="2020-05-12T18:51:51Z" level=error msg="error copying file /output/openshift-install to /bin/openshift-install" error="exit status 1"
time="2020-05-12T18:51:51Z" level=error msg="error waiting for binaries" error="exit status 1" installID=2mqhtwnz
time="2020-05-12T18:51:51Z" level=fatal msg="runtime error" error="exit status 1"
```

dgoodwin:
I have added a commit which copies the binaries to /home/ where we have a homedir and permissions to write/run. This seems to be working fine.

However it looks like kind and or podman will be fixing the noexec in a future release. I would still propose we just merge this change to run from /home as it works fine and gets us kind+podman in the meantime.

pkg/installmanager/installmanager.go
```diff
@@ -994,49 +1019,6 @@ func (m *InstallManager) gatherBootstrapNodeLogs(cd *hivev1.ClusterDeployment, n
...
-	_, err = os.Stat(destTarball)
-	if err != nil {
-		m.log.WithError(err).Error("error while stat-ing log tarball")
-		return "", err
-	}
-	m.log.Infof("cluster logs gathered: %s", destTarball)
-
-	return destTarball, nil
-}
-
```
dgoodwin, line 1039: This was dead code.

dgoodwin:
Reverted the 5001 port back to 5000.

staebler:
> However it looks like kind and or podman will be fixing the noexec in a future release. I would still propose we just merge this change to run from /home as it works fine and gets us kind+podman in the meantime.

+1
I always thought it a bit strange to run the binary from a folder called output anyway.

dgoodwin:
Multiple testers, I think we're good to go with this one.

hack/create-kind-podman-cluster.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+#
+# This script creates a kind cluster and configures it to use an insecure registry
+# running on the host OS.
+#
+# USAGE: sudo DEV_USER=myusername ./hack/create-kind-cluster-podman.sh [cluster_name]
```
staebler, line 6: Is the advice about using `sudo` and `DEV_USER` left over from a previous iteration?

dgoodwin: Yes good catch thanks. Switched to in-script sudo so it could more easily get you a kubeconfig you could use as a non-root user.

pkg/installmanager/installmanager.go
```diff
@@ -466,12 +471,32 @@ func (m *InstallManager) waitForFiles(files []string) {
...
-func (m *InstallManager) waitForInstallerBinaries() {
+func (m *InstallManager) waitForInstallerBinaries() error {
 	fileList := []string{
 		filepath.Join(m.WorkDir, "openshift-install"),
 		filepath.Join(m.WorkDir, "oc"),
 	}
 	m.waitForFiles(fileList)
+
+	// copy each binary to /bin to avoid situations where the workdir may be mounted with noexec.
+	// (i.e. kind + podman)
```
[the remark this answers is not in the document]
dgoodwin: Was originally meant to be an example. Clarifying a little bit.

dgoodwin:
Updated!

Issue #714, opened 2019-11-27
Any intent to provide vsphere support?


stobias123:
see question

dgoodwin:
Nothing planned, but if there's customer demand I think we could, if not we could also take as community contribution.

CC @katherinedube

katherinedube:
It wouldn't be until we add installer provisioned infrastructure support for deploying to VMware vSphere in the OpenShift installer before we would be able to consider it. However, adding more provider support for Hive is something we'd like to do.

stobias123:
@katherinedube 
Is there any intent to add vsphere IPI support to the installer?

katherinedube:
Yes. We're planning to see what can be done early next year to come up with an approach to be able to provide an installer provisioned infrastructure deployment method for vSphere. Unlike public cloud providers where we readily have access to standard DNS and Load Balancer services with a programmatic API, those services aren't available in a vSphere environment (typically, a myriad of customer-managed services are used.) However, those services are needed in order to bootstrap an OpenShift cluster. While non-trivial, it should be able to be solved by leveraging on-cluster DNS & LB services to handle internal cluster communication while integrating with external customer-managed services for ingress communication.

reenberg:
@katherinedube, I hope you also account for disconnected installations.  I assume especially image/content location is one of the things that is special in this scenario.  

I have just found this project, and havent had the time to comprehend all the current features.  So sorry if this is a non issue.

Do you have any information about how this would map with Resource Pools?  I would assume that most enterprises won't supply (vSphere) cluster wide credentials, but only to a specific Resource Pool for this purpose.  And I would also assume that it would be quite useful to be able to control the various users/teams resource usage through mapping to child pools, such that some teams may be allowed to provision more clusters/machines than others.   

I know this is not your direct use case, but my primary use case would be to use Hive to internally offer cluster-as-a-service to teams through something similar to RedHat cluster manager website/api.

jaykeshur:
+1

gregsheremeta:
we are beginning to work on vSphere support now.

jaykeshur:
@gregsheremeta thanks are there any expected timescales on when this will be available (even as alpha)?

gregsheremeta:
@jaykeshur I cannot commit to anything, but we hope for completion by early summer. I'll try to remember to ping you when PRs are up, or feel free to watch the repo.

rasheedamir:
[...]

PR #975, opened 2020-05-06
Initial VSphere Provisioning Support

abutcher:
- [x] add vsphere platform details to API/types
- [x] extend webhook validation
- [x] add deprovisioning support
- [x] add provisioning support
- [x] add hiveutil create-cluster support

```
export GOVC_USERNAME="abutcher"
export GOVC_PASSWORD="password"
export GOVC_TLS_CA_CERTS=/path/to/cacert

hiveutil create-cluster abutcher \
    --cloud=vsphere \
    --vsphere-vcenter=test-vcenter.com \
    --vsphere-api-vip=192.168.0.2 \
    --vsphere-ingress-vip=192.168.0.3 \
    --vsphere-cluster=devel \
    --vsphere-datacenter=dc1 \
    --vsphere-default-datastore=nvme-ds1 \
    --vsphere-network="VM Network" \
    --release-image="registry.svc.ci.openshift.org/ocp/release:4.5" \
    --pull-secret-file=/path/to/pull-secret.json \
    --base-domain=vmware.hive.openshift.com
```

Fixes: #714

PR #985, opened 2020-05-13
Add vSphere remotemachineset actuator

twiest:
- [x] Add vSphere remotemachineset actuator
- [x] Add unit tests
- [x] Manually test against real vmware cluster

This PR is dependent on PR https://github.com/openshift/hive/pull/975 and must merge after.

Fixes: #714


Dockerfile
```diff
@@ -31,6 +31,18 @@ RUN mkdir -p /home/hive && \
     chgrp -R 0 /home/hive && \
     chmod -R g=u /home/hive
 
+# This is so that we can write source certificate anchors during container start up.
```
staebler, line 34: Please add these to Dockerfile.dev as well.

pkg/apis/hive/v1/validating-webhooks/machinepool_validating_admission_hook.go
```diff
@@ -284,6 +285,11 @@ func validateMachinePoolSpecInvariants(spec *hivev1.MachinePoolSpec, fldPath *fi
 		allErrs = append(allErrs, validateOpenStackMachinePoolPlatformInvariants(p, platformPath.Child("openstack"))...)
 		numberOfMachineSets = 1 // no zone support on OpenStack
 	}
+	if p := spec.Platform.VSphere; p != nil {
+		platforms = append(platforms, "vsphere")
+		allErrs = append(allErrs, validateVSphereMachinePoolPlatformInvariants(p, platformPath.Child("vsphere"))...)
+		numberOfMachineSets = 1 // no zone support on vSphere
```
staebler, line 291: I would prefer `numberOfMachineSets` to be 0 so that the error when `spec.Autoscaling.MinReplicas` is less than 1 is
```
minimum replicas must at least 1
```
instead of
```
minimum replicas must be at least the number of zones
```
particularly since vSphere does not have zones.

pkg/clusterresource/builder.go
```diff
@@ -158,9 +155,7 @@ func (o *Builder) Build() ([]runtime.Object, error) {
 
 	var allObjects []runtime.Object
 	allObjects = append(allObjects, o.generateClusterDeployment())
-	if !o.SkipMachinePoolGeneration {
-		allObjects = append(allObjects, o.generateMachinePool())
-	}
+	allObjects = append(allObjects, o.generateMachinePool())
```
staebler: Let's put a nil check on the generated machine pool so that we don't have to add back the `SkipMachinePoolGeneration` when another platform comes along for which we don't yet have machine pool support.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1608,6 +1608,8 @@ func (r *ReconcileClusterDeployment) mergePullSecrets(cd *hivev1.ClusterDeployme
 		if err != nil {
 			if !apierrors.IsNotFound(err) {
 				return "", err
+			} else {
```
staebler, line 1611: Get rid of the check for whether it is a NotFound error.

pkg/controller/remotemachineset/vsphereactuator.go
```diff
@@ -0,0 +1,74 @@
+package remotemachineset
+
+import (
+	"fmt"
+
+	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
```
staebler, line 7: Put "github.com/openshift/hive" imports in their own section at the end of the imports.

pkg/controller/remotemachineset/vsphereactuator.go
```diff
@@ -0,0 +1,74 @@
+package remotemachineset
+
+import (
+	"fmt"
+
+	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
```
staebler, line 6: Combine this import with the other "github.com/openshift" imports that are not "github.com/openshift/hive".

pkg/controller/remotemachineset/vsphereactuator.go
```diff
@@ -0,0 +1,74 @@
...
+		NumCoresPerSocket: pool.Spec.Platform.VSphere.NumCoresPerSocket,
+		MemoryMiB:         pool.Spec.Platform.VSphere.MemoryMiB,
+		OSDisk: installertypesvsphere.OSDisk{
+			DiskSizeGB: pool.Spec.Platform.VSphere.DiskSizeGB,
+		},
+	}
+
+	ic := &installertypes.InstallConfig{
+		Platform: installertypes.Platform{
+			VSphere: &installertypesvsphere.Platform{},
```
staebler, line 57: This is missing a lot of fields that are used by the installer to create the MachineSet. For example, it is missing the vCenter. Are those fields somehow obtained in-cluster and not needed?

https://github.com/openshift/installer/blob/8da3ffa23277b55b3ddbf53761063d22a30ce8a2/pkg/asset/machines/vsphere/machines.go#L66.

pkg/apis/hive/v1/validating-webhooks/machinepool_validating_admission_hook.go
twiest: Should this be changed for OpenStack as well?

@dgoodwin thoughts as well?

I'll change both if you guys agree.

pkg/clusterresource/builder.go
twiest: I'm not sure what you man here. `o.generatedMachinePool` doesn't seem like it will ever return a nil.

Can you explain your thoughts further here please?

pkg/controller/remotemachineset/vsphereactuator.go
twiest: For this, I copied the pattern used by the [OpenStack actuator](https://github.com/openshift/hive/blob/master/pkg/controller/remotemachineset/openstackactuator.go#L66-L75), which is basically only fill out the bare minimum used to be able to generate the machinesets.

When I originally looked through the `remotemachineset` controller code, I didn't see any place these values were used. But, that was because I was focusing on the remotemachineset Update code.

During a remotemachineset create call, this information is used, so I think you're correct that we should be filling this out.

I'm running some tests now to confirm.

@dgoodwin do you think that this is a problem with remotemachineset create in the OpenStack actuator as well?

twiest:
Blocking until PR #975 is merged.

/block

pkg/clusterresource/builder.go
staebler: For the platforms that we have now, `generateMachinePool` does not return a nil. If we add support for a new platform for which Hive does not support MachinePools yet, then we need a way to skip machine pool generation. This should not come from a flag on the options: It should come from the builder implementation being used.

I am fine with leaving it as is for this PR and dealing with it when we do add support for the next platform.

pkg/apis/hive/v1/validating-webhooks/machinepool_validating_admission_hook.go
staebler: Yes, it should be changed for OpenStack as well.

twiest:
/hold

Blocking until PR #975 is merged.

pkg/controller/remotemachineset/vsphereactuator.go
twiest: For posterity, the decision here was to fill out the install config with enough information so that the created machinesets had the same critical data that the machinesets created during install had. This was manually verified by comparing the data structures.

pkg/clusterresource/builder.go
twiest: Ok, sounds good. We'll look at this in a future PR.

Dockerfile
twiest: Done.

twiest:

pkg/clusterresource/vsphere.go
```diff
@@ -111,7 +111,14 @@ func (p *VSphereCloudBuilder) addClusterDeploymentPlatform(o *Builder, cd *hivev
 }
 
 func (p *VSphereCloudBuilder) addMachinePoolPlatform(o *Builder, mp *hivev1.MachinePool) {
-	mp.Spec.Platform.VSphere = &hivev1vsphere.MachinePool{}
+	mp.Spec.Platform.VSphere = &hivev1vsphere.MachinePool{
+		NumCPUs:           2,
+		NumCoresPerSocket: 1,
+		MemoryMiB:         8192,
```
dgoodwin, line 114: Seems a little low, but is this an installer default?

twiest: I got those values from the default machinesets created by the installer.

This seems to be where the default value is stored, but I'm not 100% sure on that:
https://github.com/openshift/installer/blob/83213bd69d8a3e3937e4cdd7f7d1e6b734a535bf/upi/vsphere/variables.tf#L159-L162

twiest:
@dgoodwin @staebler I've rebased and I believe I've addressed all of your feedback. PTAL.

/hold cancel

pkg/clusterresource/vsphere.go
staebler: Those are the defaults that I see, too.

https://github.com/openshift/installer/blob/a6cd8414b02ba8cac400b8d6628e11900ca31687/pkg/asset/machines/worker.go#L130-L139

pkg/controller/remotemachineset/vsphereactuator.go
```diff
@@ -10,7 +10,8 @@ import (
 	installvsphere "github.com/openshift/installer/pkg/asset/machines/vsphere"
 	installertypes "github.com/openshift/installer/pkg/types"
 	installertypesvsphere "github.com/openshift/installer/pkg/types/vsphere"
-	"github.com/openshift/machine-api-operator/pkg/apis/vsphereprovider/v1alpha1"
+
+	//"github.com/openshift/machine-api-operator/pkg/apis/vsphereprovider/v1beta1"
```
staebler, line 14: Remove this line.

pkg/controller/remotemachineset/vsphereactuator.go
```diff
@@ -82,8 +83,5 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
 		return nil, false, errors.Wrap(err, "failed to generate machinesets")
 	}
 
-	providerSpec := installerMachineSets[0].Spec.Template.Spec.ProviderSpec.Value.Object.(*v1alpha1.VSphereMachineProviderSpec)
```
staebler: Can you move this change to the other commit?

twiest:
@staebler Thanks for the feedback. I believe I've addressed the feedback. I also manually tested this against vsphere and the functionality worked as expected. PTAL.

PR #986, opened 2020-05-13
standardize on GCP creds flags/env vars

joelddiaz:
Fix up the difference where when using hiveutil to create a cluster you would set GCP_SHARED_CREDENTIALS_FILE, but on deprovision you would set GOOGLE_CREDENTIALS.

Standardize around the installer's use of GOOGLE_CREDENTIALS. Update the docs to more clearly specify that we actually search for ~/.gcp/osServiceAccount.json, GOOGLE_CREDENTIALS and the --creds-file parameter. Also add a --creds-file parameter for deprovision to make it more similar to the create-cluster workflow.

### DIFF
--- a/contrib/pkg/deprovision/gcp.go
+++ b/contrib/pkg/deprovision/gcp.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/installer/pkg/types"
 	typesgcp "github.com/openshift/installer/pkg/types/gcp"
 
+	gcputils "github.com/openshift/hive/contrib/pkg/utils/gcp"
 	"github.com/openshift/hive/pkg/gcpclient"
 )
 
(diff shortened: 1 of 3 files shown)


contrib/pkg/deprovision/gcp.go
```diff
@@ -61,8 +64,12 @@ func (o *gcpOptions) Validate(cmd *cobra.Command) error {
 		log.Info("Region is required")
 		return fmt.Errorf("missing region")
 	}
-	credsFile := os.Getenv("GOOGLE_CREDENTIALS")
-	projectID, err := gcpclient.ProjectIDFromFile(credsFile)
+
+	creds, err := gcputils.GetCreds(o.credsFile)
```
staebler, line 68: This is not sufficient. This will only change what is used for determining the project ID. The uninstaller will not respect the `--creds-file` flag.

contrib/pkg/deprovision/gcp.go
```diff
@@ -45,6 +47,7 @@ func NewDeprovisionGCPCommand() *cobra.Command {
 	flags := cmd.Flags()
 	flags.StringVar(&opt.logLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
 	flags.StringVar(&opt.region, "region", "", "GCP region where the cluster is installed")
+	flags.StringVar(&opt.credsFile, "creds-file", "", "File containing GCP credentials")
```
staebler, line 50: If we are going to add this for GCP deprovision (which I am not convinced that we should), then we should add it to all of the platforms.

joelddiaz:
@staebler i ended up just removing the deprovision parameter since the parameter introduces all kinds of complications b/c the installer doesn't allow you to bring your own creds file

PR #987, opened 2020-05-13
crd: strip leading yaml breaks

staebler:


Makefile
```diff
@@ -75,9 +76,10 @@ manifests: crd
 
 # Generate CRD yaml from our api types:
 .PHONY: crd
-crd: ensure-controller-gen
+crd: ensure-controller-gen ensure-yq
 	rm -rf ./config/crds
 	'$(CONTROLLER_GEN)' crd paths=./pkg/apis/hive/v1 output:dir=./config/crds
+	yq -y -i . ./config/crds/*.yaml
```
gregsheremeta, line 82: how does this work? I don't see a mention of `-y` in the yq docs. Is it quiet mode? And does this just make yq do a reformat?

staebler: The `-y` flag makes the output yaml instead of json.

staebler: yq reads in the file then outputs it as yaml and replaces the file with the output.

gregsheremeta: ok cool. I have to use `r` for read in the version I have. Shrug. lgtm!

csrwng: The issue is there are multiple versions of yq out there that work slightly differently.

staebler:
I added `ensure-yq` as a dependency to the `verify-crd` target.

I added a comment about the use of yq in the Makefile.

Makefile
staebler: Yes, you have a completely different yq.

staebler: Ah, I am using the wrong one in the makefile.

PR #989, opened 2020-05-14
standardize Azure creds handling

joelddiaz:
Document Azure creds format, and make notes about how deprovisioning works for Azure.

Standardize reading in of Azure creds.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -3,11 +3,6 @@ package createcluster
 import (
 	"encoding/json"
 	"fmt"
-	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
-	gcputils "github.com/openshift/hive/contrib/pkg/utils/gcp"
-	openstackutils "github.com/openshift/hive/contrib/pkg/utils/openstack"
-	"github.com/openshift/hive/pkg/clusterresource"
-	"github.com/openshift/hive/pkg/gcpclient"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -28,8 +23,14 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
+	azurecredutil "github.com/openshift/hive/contrib/pkg/utils/azure"
+	gcputils "github.com/openshift/hive/contrib/pkg/utils/gcp"
+	openstackutils "github.com/openshift/hive/contrib/pkg/utils/openstack"
 	"github.com/openshift/hive/pkg/apis"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/clusterresource"
+	"github.com/openshift/hive/pkg/gcpclient"
 	"github.com/openshift/hive/pkg/resource"
 )
 
@@ -88,7 +89,6 @@ metadata:
   namespace: bar
 type: TestFailResource
 `
-	azureCredFile = "osServicePrincipal.json"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


contrib/pkg/utils/azure/azure.go
```diff
@@ -0,0 +1,26 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/hive/pkg/constants"
+)
+
+// GetCreds reads Azure credentials used for install/uninstall from either the default
+// credentials file (~/.azure/osServiceAccount.json), the standard environment variable,
+// or or provided credsFile location (in increasing order of preference).
+func GetCreds(credsFile string) ([]byte, error) {
+	credsFilePath := filepath.Join(os.Getenv("HOME"), ".azure", constants.AzureCredentialsName)
```
csrwng, line 17: Not all OS's expose a HOME variable...  there is a utility:
import `k8s.io/client-go/util/homedir` 
`homedir.HomeDir()`

joelddiaz: nice!

csrwng:
Same error in both e2e's:
```
Failed to pull image "registry.svc.ci.openshift.org/ci-op-r3nxcv58/release:latest": unable to parse auth field
```

contrib/pkg/createcluster/create.go
```diff
@@ -3,18 +3,19 @@ package createcluster
...
-	"github.com/openshift/hive/pkg/clusterresource"
-	"github.com/openshift/hive/pkg/gcpclient"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"os/user"
 	"path/filepath"
 	"strings"
 
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
```
staebler, line 13: Move these imports down to the last section of imports, with the other "github.com/openshift/hive" imports.

contrib/pkg/deprovision/azure.go
```diff
@@ -1,11 +1,15 @@
 package deprovision
 
 import (
+	"os"
+
 	azuresession "github.com/openshift/installer/pkg/asset/installconfig/azure"
```
staebler, line 6: While cleaning up imports, the "github.com/openshift" imports that are not hive should be in their own section before the "github.com/openshift/hive" imports.

contrib/pkg/utils/azure/azure.go
```diff
@@ -0,0 +1,28 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"k8s.io/client-go/util/homedir"
+
+	"github.com/openshift/hive/pkg/constants"
+)
+
+// GetCreds reads Azure credentials used for install/uninstall from either the default
+// credentials file (~/.azure/osServiceAccount.json), the standard environment variable,
+// or or provided credsFile location (in increasing order of preference).
```
staebler, line 17:
```suggestion
// or provided credsFile location (in increasing order of preference).
```

docs/hiveutil.md
```diff
@@ -22,12 +22,23 @@ bin/hiveutil create-cluster --base-domain=mydomain.example.com --cloud=aws myclu
 
 #### Create Cluster on Azure
 
-Credentials will be read from `~/.azure/osServicePrincipal.json` typically created via the `az login` command. Alternatively you can specify an Azure credentials file with `--creds-file` or the `AZURE_AUTH_LOCATION` environment variable.
+Credentials will be read from either `~/.azure/osServicePrincipal.json` (or the Windows equivalent of a HOME/HOMEDRIVE/etc), the contents of the `AZURE_AUTH_LOCATION` environment variable, or the value provided with the `--creds-file` parameter (in increasing order of preference). The format for the credentials used for installation/uninstallation follows the same format used by the OpenShift installer:
```
staebler, line 25: We do not call out this Windows equivalent for any other platforms. I don't think it is necessary here, either.

contrib/pkg/utils/azure/azure.go
```diff
@@ -0,0 +1,28 @@
...
+func GetCreds(credsFile string) ([]byte, error) {
+	credsFilePath := filepath.Join(homedir.HomeDir(), ".azure", constants.AzureCredentialsName)
+	if l := os.Getenv(constants.AzureCredentialsEnvVar); l != "" {
+		credsFilePath = l
+	}
+	if credsFile != "" {
+		credsFilePath = credsFile
+	}
+	log.Infof("Loading azure creds from: %s", credsFilePath)
```
staebler, line 26:
```suggestion
	log.WithFieild("credsFile", credsFilePath).Info("Loading azure creds")
```

docs/hiveutil.md
joelddiaz: Azure is the only cloud where we use the homedir.HomeDir() call which introduces this new behavior of not relying on $HOME. This behavior seemed a little surprising to me, so I thought I'd document it somewhere, but perhaps these alternative HOME locations are just how things regularly happen on Windows and wouldn't surprise anyone running binaries on that OS.

joelddiaz: Also, should we also move the GCP equivalent to use homedir.Homedir()?

staebler: Yes, we probably should do the same for GCP, and OpenStack.

staebler: Considering that hiveutil is a development tool, is it safe to assume that someone running hiveutil on Windows is using Powershell? In that case `~` is valid for them as well.

PR #995, opened 2020-05-18
Add MachinePool support for bring your own VPC

### LINKED ISSUE #886: Problem with MachinePools in cluster installed in existing VPC
We are using hive to create and manage ocp clusters in AWS. In some cases we are using the option to install the cluster in an already existent VPC, but we have found a problem with MachinePools in this scenario:

* When a new MachinePool is created, hive creates the proper MachineSets in the ocp cluster, but it does so by specifying the `spec.providerSpec.subnet` field like this:

```yaml
securityGroups:
  - filters:
      - name: tag:Name
        values:
          - stg-saas-ocp4-3-skbkh-worker-sg
```

* This is is problematic because our subnets, that are created before the cluster is, don't have the `Name` tag that this configuration is expectig, thus causing the Machine objects to get into failed state.

As far as I was able to check, there is no way to specify the subnet IDs to be used in the MachinePool, but I migh have overlooked something. Is it possible?
If not, will this be supported at some point?

For the moment, we are going to try to use SyncSets to manage our MachineSets in the case of clusters provisioned in already existent VPCs. Do you think there could be problems with this approach?

This is the MachinePool configuration we were trying to use and that resulted in the problem described:

```yaml
kind: MachinePool
metadata:
  name: stg-saas-ocp4-3-worker-m5-2xlarge
  namespace: hive
spec:
  clusterDeploymentRef:
    name: stg-saas-ocp4-3
  name: worker-m5-2xlarge
  platform:
    aws:
      rootVolume:
        iops: 150
        size: 50
        type: gp2
      type: m5.2xlarge
      zones:
      - us-east-1a
      - us-east-1b
      - us-east-1c
  replicas: 3
```

akhil-rane:
Currently, Hive can install the cluster in an already existent VPC. But when a
new MachinePool is created, Hive creates the MachineSets in the cluster without
specifying the subnet IDs, thus causing the Machine objects to get into failed
state. This PR adds an option to specify the subnet IDs to be used in the
MachinePool.

Fixes #886

### DIFF
--- a/pkg/apis/hive/v1/aws/machinepool.go
+++ b/pkg/apis/hive/v1/aws/machinepool.go
@@ -6,6 +6,10 @@ type MachinePoolPlatform struct {
 	// Zones is list of availability zones that can be used.
 	Zones []string `json:"zones,omitempty"`
 
+	// Subnets is the list of subnets to which to attach the machines.
+	// There must be exactly one subnet for each availability zone used.
+	Subnets []string `json:"subnets,omitempty"`
+
 	// InstanceType defines the ec2 instance type.
 	// eg. m4-large
 	InstanceType string `json:"type"`
--- a/pkg/controller/remotemachineset/awsactuator.go
+++ b/pkg/controller/remotemachineset/awsactuator.go
@@ -1,7 +1,10 @@
 package remotemachineset
 
 import (
+	"context"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -11,7 +14,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/sets"
 	awsprovider "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsproviderconfig/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	installaws "github.com/openshift/installer/pkg/asset/machines/aws"
@@ -19,18 +24,25 @@ import (
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/awsclient"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 )
 
 // AWSActuator encapsulates the pieces necessary to be able to generate
 // a list of MachineSets to sync to the remote cluster.
 type AWSActuator struct {
-	client awsclient.Client
-	logger log.FieldLogger
-	region string
-	amiID  string
+	client    client.Client
+	awsClient awsclient.Client
+	logger    log.FieldLogger
+	region    string
+	amiID     string
 }
 
-var _ Actuator = &AWSActuator{}
+var (
+	_ Actuator = &AWSActuator{}
+
+	// reg is a regex used to fetch condition message from error when subnets specified in the MachinePool are invalid
+	reg = regexp.MustCompile(`^InvalidSubnetID\.NotFound:\s+([^\t]+)\t`)
+)
 
 // NewAWSActuator is the constructor for building a AWSActuator
 func NewAWSActuator(awsCreds *corev1.Secret, region string, pool *hivev1.MachinePool, remoteMachineSets []machineapi.MachineSet, scheme *runtime.Scheme, logger log.FieldLogger) (*AWSActuator, error) {
@@ -50,10 +62,10 @@ func NewAWSActuator(awsCreds *corev1.Secret, region string, pool *hivev1.Machine
 		}
 	}
 	actuator := &AWSActuator{
-		client: awsClient,
-		logger: logger,
-		region: region,
-		amiID:  amiID,
+		awsClient: awsClient,
+		logger:    logger,
+		region:    region,
+		amiID:     amiID,
 	}
 	return actuator, nil
 }
@@ -94,16 +106,55 @@ func (a *AWSActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 		computePool.Platform.AWS.Zones = zones
 	}
 
-	// Private subnets && userTags are settings available in the installconfig
-	// that we are choosing to ignore for the timebeing. These empty settings
-	// should be updated to feed from the machinepool / installconfig in the
-	// future.
 	subnets := map[string]string{}
+	// Fetching private subnets from the machinepool and then mapping availability zones to subnets
+	if len(pool.Spec.Platform.AWS.Subnets) > 0 {
+		subnetsByAvailabilityZone, err := a.getSubnetsByAvailabilityZone(pool)
+		if err != nil {
+			return nil, false, errors.Wrap(err, "describing subnets")
+		}
+		subnets = subnetsByAvailabilityZone
+	}
+	// userTags are settings available in the installconfig that we are choosing
+	// to ignore for the timebeing. These empty settings should be updated to feed
+	// from the machinepool / installconfig in the future.
 	userTags := map[string]string{}
 
 	installerMachineSets, err := installaws.MachineSets(cd.Spec.ClusterMetadata.InfraID, cd.Spec.Platform.AWS.Region, subnets, computePool, pool.Spec.Name, workerUserData, userTags)
 	if err != nil {
+		if strings.Contains(err.Error(), "no subnet for zone") {
+			conds, changed := controllerutils.SetMachinePoolConditionWithChangeCheck(
+				pool.Status.Conditions,
+				hivev1.InvalidSubnetsMachinePoolCondition,
+				corev1.ConditionTrue,
+				"NoSubnetForAvailabilityZone",
+				err.Error(),
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				pool.Status.Conditions = conds
+				if err := a.client.Status().Update(context.Background(), pool); err != nil {
+					return nil, false, err
+				}
+			}
+		}
+
 		return nil, false, errors.Wrap(err, "failed to generate machinesets")
+	}
+
+	conds, changed := controllerutils.SetMachinePoolConditionWithChangeCheck(
+		pool.Status.Conditions,
+		hivev1.InvalidSubnetsMachinePoolCondition,
+		corev1.ConditionFalse,
+		"ValidSubnets",
+		"Subnets are valid",
+		controllerutils.UpdateConditionNever,
+	)
+	if changed {
+		pool.Status.Conditions = conds
+		if err := a.client.Status().Update(context.Background(), pool); err != nil {
+			return nil, false, err
+		}
 	}
 
 	// Re-use existing AWS resources for generated MachineSets.
@@ -150,7 +201,7 @@ func (a *AWSActuator) fetchAvailabilityZones() ([]string, error) {
 	req := &ec2.DescribeAvailabilityZonesInput{
 		Filters: []*ec2.Filter{zoneFilter},
 	}
-	resp, err := a.client.DescribeAvailabilityZones(req)
+	resp, err := a.awsClient.DescribeAvailabilityZones(req)
 	if err != nil {
 		return nil, err
 	}
@@ -188,11 +239,14 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
 	// broken on us once via renames in the installer. We need to start querying for what exists
 	// here.
 	providerConfig.IAMInstanceProfile = &awsprovider.AWSResourceReference{ID: aws.String(fmt.Sprintf("%s-worker-profile", infraID))}
-	providerConfig.Subnet = awsprovider.AWSResourceReference{
-		Filters: []awsprovider.Filter{{
-			Name:   "tag:Name",
-			Values: []string{fmt.Sprintf("%s-private-%s", infraID, providerConfig.Placement.AvailabilityZone)},
-		}},
+	// Update the subnet filter only if subnet id is absent
+	if providerConfig.Subnet.ID == nil {
+		providerConfig.Subnet = awsprovider.AWSResourceReference{
+			Filters: []awsprovider.Filter{{
+				Name:   "tag:Name",
+				Values: []string{fmt.Sprintf("%s-private-%s", infraID, providerConfig.Placement.AvailabilityZone)},
+			}},
+		}
 	}
 	providerConfig.SecurityGroups = []awsprovider.AWSResourceReference{{
 		Filters: []awsprovider.Filter{{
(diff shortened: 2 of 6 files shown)


pkg/apis/hive/v1/aws/machinepool.go
```diff
@@ -6,6 +6,9 @@ type MachinePoolPlatform struct {
 	// Zones is list of availability zones that can be used.
 	Zones []string `json:"zones,omitempty"`
 
+	// Subnets is the list of subnets for each availability zone.
```
staebler, line 9:
```suggestion
	// Subnets is the list of subnets to which to attach the machines.
	// There must be exactly one subnet for each availability zone used.
```

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +213,22 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+		idPointers = append(idPointers, aws.String(id))
+	}
+
+	results, err := a.client.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, subnet := range results.Subnets {
+		subnetsByAvailabilityZone[*subnet.AvailabilityZone] = *subnet.SubnetId
```
staebler, line 231: It should be an error if there has already been a subnet found for the availability zone. This, too, is something that would be good to show in a condition.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +213,22 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+
+// getSubnetsByAvailabilityZones maps availability zones to subnet
+func (a *AWSActuator) getSubnetsByAvailabilityZone(subnets []string) (map[string]string, error) {
+	subnetsByAvailabilityZone := map[string]string{}
+	idPointers := make([]*string, len(subnets))
+	for _, id := range subnets {
+		idPointers = append(idPointers, aws.String(id))
+	}
+
+	results, err := a.client.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
```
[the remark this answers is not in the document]
akhil-rane: I did not understand this part `this is something that would be good to expose as a condition on the MachinePool`. Can you please elaborate

pkg/controller/remotemachineset/awsactuator.go
akhil-rane: Addressed

pkg/controller/remotemachineset/awsactuator.go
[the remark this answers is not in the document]
staebler: What I mean is that this error is most likely a user error. If all we do is log the error in the controller logs, we are not giving any meaningful feedback to the user about (1) the presence of a problem and (2) how they may fix the problem. If we put a condition in the status of the `MachinePool` that tells the user that there is a problem with the subnet IDs that they provided, then the user can act on that information.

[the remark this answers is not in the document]
staebler: For example,
```
status:
  conditions:
  - type: InvalidSubnets
    reason: NoSubnetForAvailabilityZone
    message: There is no subnet specified for some availability zones: useast-1a
```

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -2,7 +2,6 @@ package remotemachineset
 
 import (
 	"fmt"
-
```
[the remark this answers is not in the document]
akhil-rane: Added some more imports here

pkg/controller/remotemachineset/awsactuator.go
akhil-rane: Added a condition in the machinepool for this error

pkg/controller/remotemachineset/awsactuator.go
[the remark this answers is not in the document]
akhil-rane: Added conditions in the machinepool to capture errors

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -216,24 +253,90 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
 
-	results, err := a.client.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
+	results, err := a.awsClient.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
 	if err != nil {
+		conds, changed := controllerutils.SetMachinePoolConditionWithChangeCheck(
+			pool.Status.Conditions,
+			hivev1.InvalidSubnetsMachinePoolCondition,
+			corev1.ConditionTrue,
+			"SubnetsNotFound",
+			err.Error(),
```
akhil-rane, line 270: Sample message for this condition -> `InvalidSubnetID.NotFound: The subnet ID 'subnet-1' does not exist status code: 400, request id: f506c630-ab03-4d81-b116-745e0b3e36ce`. Do we need to format it?

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -94,18 +99,57 @@ func (a *AWSActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
...
+				pool.Status.Conditions,
+				hivev1.InvalidSubnetsMachinePoolCondition,
+				corev1.ConditionTrue,
+				"NoSubnetForAvailabilityZone",
+				err.Error(),
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				pool.Status.Conditions = conds
+				if err := a.client.Status().Update(context.Background(), pool); err != nil {
```
staebler, line 129: Updating the MachinePool in the actuator is going to cause conflicts in the controller if the controller then tries to update the MachinePool.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +251,92 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
 		Value: &runtime.RawExtension{Object: providerConfig},
 	}
 }
+
+// getSubnetsByAvailabilityZones maps availability zones to subnet
+func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (map[string]string, error) {
+	subnetsByAvailabilityZone := map[string]string{}
```
staebler, line 257: Move this declaration down closer to where the variable is used.

staebler, line 257: Do a `make` with the expected size.

pkg/controller/remotemachineset/awsactuator.go
staebler: We have to do some formatting. If we leave it as is, then the condition will change every time since the request id will change.

Also, what do you think about only setting the condition when the error code is "InvalidSubnetID.NotFound"?

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +251,92 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+		if changed {
+			pool.Status.Conditions = conds
+			if err := a.client.Status().Update(context.Background(), pool); err != nil {
+				return nil, err
+			}
+		}
+		return nil, err
+	}
+
+	conds, changed := controllerutils.SetMachinePoolConditionWithChangeCheck(
```
staebler, line 282: The condition should not be cleared here. The condition should only be cleared once the call to `installaws.MachineSets` is successful.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +251,92 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+			pool.Status.Conditions = conds
+			if err := a.client.Status().Update(context.Background(), pool); err != nil {
+				return nil, err
+			}
+		}
+
+		return nil, errors.Errorf("more than one subnet found for some availability zones: %s", strings.Join(zonesWithMultipleSubnets, ", "))
+	}
+
+	conds, changed = controllerutils.SetMachinePoolConditionWithChangeCheck(
```
staebler, line 326: Likewise, the condition should not be cleared here.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +251,92 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+		subnetsByAvailabilityZone[*subnet.AvailabilityZone] = *subnet.SubnetId
+	}
+
+	if len(zonesWithMultipleSubnets) > 0 {
+		conds, changed := controllerutils.SetMachinePoolConditionWithChangeCheck(
+			pool.Status.Conditions,
+			hivev1.InvalidSubnetsMachinePoolCondition,
+			corev1.ConditionTrue,
+			"MoreThanOneSubnetForZone",
+			fmt.Sprintf("more than one subnet found for some availability zones: %s", strings.Join(zonesWithMultipleSubnets, ", ")),
```
staebler, line 313: I think it may be more helpful to list the subnets that conflict rather than the zones with conflicting subnets.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +251,92 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+		if err := a.client.Status().Update(context.Background(), pool); err != nil {
+			return nil, err
+		}
+	}
+
+	var zonesWithMultipleSubnets []string
+	for _, subnet := range results.Subnets {
+		_, ok := subnetsByAvailabilityZone[*subnet.AvailabilityZone]
+		if ok {
+			zonesWithMultipleSubnets = append(zonesWithMultipleSubnets, *subnet.AvailabilityZone)
```
staebler, line 301: Use `sets.String` to accumulate the conflicts. Otherwise, if there are multiple conflicts, then you could end up with the same zone listed multiple times.

https://github.com/kubernetes/apimachinery/blob/bdcc9f4ab675d837965d1bd5babf8299afead78a/pkg/util/sets/string.go#L27.

pkg/controller/remotemachineset/awsactuator.go
akhil-rane: I am not able to understand the reasoning behind it. I believe the Machinepool object is passed as a reference to the actuator. We are updating the Machinepool in the GCP actuator as well. Will that also need to change?

staebler: You are right. Disregard. :)

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +253,73 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+		idPointers[i] = aws.String(id)
+	}
+
+	results, err := a.awsClient.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
+	if err != nil {
+		if strings.Contains(err.Error(), "InvalidSubnetID.NotFound") {
+			// formatting error message before adding it to condition
+			// sample error message: InvalidSubnetID.NotFound: The subnet ID 'subnet-1,subnet-2' does not exist status code: 400, request id: f506c630-ab03-4d81-b116-745e0b3e36ce
+			// message after formatting: The subnet ID 'subnet-1,subnet-2' does not exist
+			reg := regexp.MustCompile("\tstatus code:(.)*")
```
staebler, line 270: Move this into a package-scoped variable. That way if there is a problem compiling the regular expression, then it will kill the process at startup rather than at some random time in the future.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +253,73 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+			conditonMessage := strings.TrimPrefix(reg.ReplaceAllString(err.Error(), ""), "InvalidSubnetID.NotFound: ")
```
staebler, line 271: I think this is a bit clearer without the replaces and trims. It just picks out the submatch that we are interested in.
```suggestion
		reg := regexp.MustCompile(`^InvalidSubnetID\.NotFound:\s+([^\t]+)\t`)
		if submatches := reg.FindStringSubmatches(err.Error()); submatches != nil {
			// formatting error message before adding it to condition
			// sample error message: InvalidSubnetID.NotFound: The subnet ID 'subnet-1,subnet-2' does not exist status code: 400, request id: f506c630-ab03-4d81-b116-745e0b3e36ce
			// message after formatting: The subnet ID 'subnet-1,subnet-2' does not exist
			conditionMessage = submatches[1]
		}
```

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +253,73 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+	idPointers := make([]*string, len(pool.Spec.Platform.AWS.Subnets))
+	for i, id := range pool.Spec.Platform.AWS.Subnets {
+		idPointers[i] = aws.String(id)
+	}
+
+	results, err := a.awsClient.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
+	if err != nil {
+		if strings.Contains(err.Error(), "InvalidSubnetID.NotFound") {
+			// formatting error message before adding it to condition
+			// sample error message: InvalidSubnetID.NotFound: The subnet ID 'subnet-1,subnet-2' does not exist status code: 400, request id: f506c630-ab03-4d81-b116-745e0b3e36ce
```
staebler, line 268: Can we add into this comment `\t`s where the tabs are?

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +253,73 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+				if err := a.client.Status().Update(context.Background(), pool); err != nil {
+					return nil, err
+				}
+			}
+		}
+		return nil, err
+	}
+
+	conflictingSubnets := sets.NewString()
+	subnetsByAvailabilityZone := make(map[string]string, len(pool.Spec.Platform.AWS.Zones))
```
staebler, line 292: Use the length of the subnets rather than the length of the availability zones. The availability zones slice may be empty. The happy scenario has the length of the subnets equal to the number of availability zones used.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +253,73 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+	for _, subnet := range results.Subnets {
+		subnetID, ok := subnetsByAvailabilityZone[*subnet.AvailabilityZone]
+		if ok {
+			conflictingSubnets.Insert(*subnet.SubnetId)
+			if !conflictingSubnets.Has(subnetID) {
```
staebler, line 297: No need to do this check. Just do the insert. The set is guaranteed to not have duplicates.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -204,3 +253,73 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+			}
+		}
+		return nil, err
+	}
+
+	conflictingSubnets := sets.NewString()
+	subnetsByAvailabilityZone := make(map[string]string, len(pool.Spec.Platform.AWS.Zones))
+	for _, subnet := range results.Subnets {
+		subnetID, ok := subnetsByAvailabilityZone[*subnet.AvailabilityZone]
+		if ok {
```
staebler, line 295: Prefer keeping variables in the scope where they are needed.
```suggestion
		if subnetID, ok := subnetsByAvailabilityZone[*subnet.AvailabilityZone]; ok {
```

dgoodwin:
Looks good, could you squash commits @akhil-rane and we can merge.

akhil-rane:
@dgoodwin squashed and ready to merge :)

PR #996, opened 2020-05-18

staebler:
The admission webhook for CluterDeployments has been modified to inspect DELETE requests as well. The webhook will reject any DELETE request for a ClusterDeployment that has the "hive.openshift.io/protected-delete" annotation set to a true value. In order to delete such a ClusterDeployment, a user is required to delete the annotation prior to making the DELETE request.

The "deleteProtection" field has been added to the HiveConfig. When the field is set to "enabled", the clusterdeployment controller will add the "hive.openshift.io/protected-delete" annotation to the ClusterDeployment when transitioning the ClusterDeployment to installed.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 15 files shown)


staebler:
> Please add a backup check where we create ClusterDeprovisions, we have seen situations where hiveadmission was not configured correctly and because it's not in the picture, API server was just accepting requests.

That's a good idea. I will do that. Of course, once it gets to that point, there is no going back on the delete of the ClusterDeployment.

dgoodwin:
Yup absolutely not good, but at least their data is there and we could figure out something to preserve I think.

staebler:
> > Please add a backup check where we create ClusterDeprovisions, we have seen situations where hiveadmission was not configured correctly and because it's not in the picture, API server was just accepting requests.
> 
> That's a good idea. I will do that. Of course, once it gets to that point, there is no going back on the delete of the ClusterDeployment.

Done.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1144,6 +1165,10 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.Clus
 }
 
 func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (reconcile.Result, error) {
+	if controllerutils.IsDeleteProtected(cd) {
```
dgoodwin, line 1168: Should we include r.protectedDelete in the if? Maybe you had it enabled but turned it off later?

staebler: The only thing that enabling protected delete in the HiveConfig does is determine whether Hive will automatically add the annotation on the ClusterDeployment. The user can still add the annotation manually.

PR #998, opened 2020-05-21

staebler:
The admission webhook for CluterDeployments has been modified to inspect DELETE requests as well. The webhook will reject any DELETE request for a ClusterDeployment that has the "hive.openshift.io/protected-delete" annotation set to a true value. In order to delete such a ClusterDeployment, a user is required to delete the annotation prior to making the DELETE request.

The "deleteProtection" field has been added to the HiveConfig. When the field is set to "enabled", the clusterdeployment controller will add the "hive.openshift.io/protected-delete" annotation to the ClusterDeployment when transitioning the ClusterDeployment to installed.

### DIFF
--- a/config/crds/hive.openshift.io_hiveconfigs.yaml
+++ b/config/crds/hive.openshift.io_hiveconfigs.yaml
@@ -70,6 +70,15 @@ spec:
                       type: boolean
                   type: object
               type: object
+            deleteProtection:
+              description: DeleteProtection can be set to "enabled" to turn on automatic
+                delete protection for ClusterDeployments. When enabled, Hive will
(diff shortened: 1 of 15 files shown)


staebler:
```
time="2020-05-21T12:05:11Z" level=info msg="Validating request" group=hive.openshift.io method=Validate operation=DELETE resource=clusterdeployments version=v1
time="2020-05-21T12:05:11Z" level=info msg="Successful validation" group=hive.openshift.io method=validateDelete object.Name=mstaeble operation=DELETE resource=clusterdeployments version=v1
```

staebler:
Redux of #996 with the fix for the validation webhook.

PR #999, opened 2020-05-22
allow selectively enabling/disabling controllers

joelddiaz:
Introduce `--controllers` to allow specifying a comma-separated list of controllers to run.
Introduce `--disabled-controllers` to allow specifying a comma-separated list of controllers to not run (disabled takes precedence over anything in `--controllers`).

Modify how controllers are registered so that we can use the information from these new parameters when initializing controllers on startup.

This should allow a development workflow where you disable the controller(s) you are working on in your "deployed" hive instance, and then only enable the controller(s) you are working on for you local run/debug session. (You do need to remember to disable leader election for your local run.)

Note: the current controller names have been left unchanged, and there may be a desire to standardize on a naming scheme.


cmd/manager/main.go
```diff
@@ -150,12 +202,28 @@ func newRootCommand() *cobra.Command {
...
 }
 
+func newControllerManagerOptions() *controllerManagerOptions {
+	// By default we have all of the controllers enabled
+	controllers := make([]string, len(controllerFuncs))
+	i := 0
+	for name := range controllerFuncs {
+		controllers[i] = name
+		i++
+	}
```
staebler, line 221:
```suggestion
	controllers := make([]string, 0, len(controllerFuncs))
	for name := range controllerFuncs {
		controllers = append(controllers, name)
	}
```

joelddiaz:
incorporated @staebler 's recommendation. i can just go ahead and squash if this is ready.

staebler:
> incorporated @staebler 's recommendation. i can just go ahead and squash if this is ready.

Squash away, as far as I am concerned.

staebler:
There are merge conflicts.

/lgtm cancel

joelddiaz:
@staebler yes, i had to update for the removal of the clusterreolcate bits. should be good again.

PR #1000, opened 2020-05-24
Revert "validation: decode correct object on clusterdeployment delete"

gregsheremeta:
This reverts commit c7bd4f82be829456846e875909eac9ceda0685b5.

There are reports that it is still preventing deletes.

/cc @staebler


gregsheremeta:
```
# time="2020-05-24T12:49:23Z" level=error msg="Failed unmarshaling Object: there is no content to decode" group=hive.openshift.io method=validateDelete operation=DELETE resource=clusterdeployments version=v1
 
$ grep "Failed unmarshaling Object: there is no content to decode" admission.log | wc
  58220  931520 11993320
```
I'm going to merge this revert to unblock people.

PR #1004, opened 2020-05-25
Add counter metric for completed cluster provisions.

dgoodwin:
Labelled by result which will allow us to monitor and possibly alert if
the percentage of failing cluster provisions over a window of time
exceeds some threshold.

### DIFF
--- a/pkg/controller/clusterprovision/clusterprovision_controller.go
+++ b/pkg/controller/clusterprovision/clusterprovision_controller.go
@@ -38,12 +38,21 @@ const (
 	// clusterProvisionLabelKey is the label that is used to identify
 	// resources descendant from a cluster provision.
 	clusterProvisionLabelKey = "hive.openshift.io/cluster-provision"
+
+	resultSuccess = "success"
+	resultFailure = "failure"
 )
 
 var (
 	// controllerKind contains the schema.GroupVersionKind for this controller type.
 	controllerKind = hivev1.SchemeGroupVersion.WithKind("ClusterProvision")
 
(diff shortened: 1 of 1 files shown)


dgoodwin:
@suhanime could you review this one?

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -290,6 +300,7 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 
 func (r *ReconcileClusterProvision) reconcileSuccessfulJob(instance *hivev1.ClusterProvision, job *batchv1.Job, pLog log.FieldLogger) (reconcile.Result, error) {
 	pLog.Info("install job succeeded")
+	metricClusterProvisionsTotal.WithLabelValues(hivemetrics.GetClusterDeploymentType(instance), resultSuccess)
```
staebler, line 303: I assume that you intended to add a `.Inc()` to the end so that the metric is incremented.

staebler, line 303: Move the metric increment to after the stage transition so that the metric is not incremented multiple times when there is an error transitioning the stage.

dgoodwin: Ughhhh yes.

dgoodwin:
Updated.

dgoodwin:
/test e2e

This should be ready for review again.

PR #1006, opened 2020-05-26
disabled cluster relocation when using managed DNS

staebler:
Since the dnszone and dnsendpoint controllers have had their logic for handling relocated clusters neutered, then cluster relocation must be disabled when using managed DNS.


staebler:
> Are they neutered? After attempt #2 my understanding was that the remaining change just lets reconcile continue if ClusterDeployment is gone.

Yes, it lets the reconcile continue, which it should not in the case where the DNSZone has been copied to the destination cluster but the ClusterDeployment has not yet.

PR #1007, opened 2020-05-26
revert cluster relocate

staebler:
Revert #971.


dgoodwin:
Conflicts after I merged that other PR. I have no idea what's up with e2e, looks like it's not us.

dgoodwin:
/lgtm

Will wait for e2e and force.

PR #1008, opened 2020-05-26
Block SyncSets with secret ref in another namespace.

dgoodwin:
Blocked during admission, but as a secondary check we also make sure not
to reconcile such a SyncSet and log a warning.

This change is not relevant for SelectorSyncSets which are cluster
scoped.

### DIFF
--- a/pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
+++ b/pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
@@ -193,6 +193,7 @@ func (a *SyncSetValidatingAdmissionHook) validateCreate(admissionSpec *admission
 	allErrs = append(allErrs, validateResources(newObject.Spec.Resources, field.NewPath("spec").Child("resources"))...)
 	allErrs = append(allErrs, validatePatches(newObject.Spec.Patches, field.NewPath("spec").Child("patches"))...)
 	allErrs = append(allErrs, validateSecrets(newObject.Spec.Secrets, field.NewPath("spec").Child("secretMappings"))...)
+	allErrs = append(allErrs, validateSourceSecretInSyncSetNamespace(newObject.Spec.Secrets, newObject.Namespace, field.NewPath("spec", "secretMappings"))...)
 	allErrs = append(allErrs, validateResourceApplyMode(newObject.Spec.ResourceApplyMode, field.NewPath("spec", "resourceApplyMode"))...)
 
 	if len(allErrs) > 0 {
@@ -240,6 +241,7 @@ func (a *SyncSetValidatingAdmissionHook) validateUpdate(admissionSpec *admission
 	allErrs = append(allErrs, validateResources(newObject.Spec.Resources, field.NewPath("spec", "resources"))...)
 	allErrs = append(allErrs, validatePatches(newObject.Spec.Patches, field.NewPath("spec", "patches"))...)
 	allErrs = append(allErrs, validateSecrets(newObject.Spec.Secrets, field.NewPath("spec", "secretMappings"))...)
+	allErrs = append(allErrs, validateSourceSecretInSyncSetNamespace(newObject.Spec.Secrets, newObject.Namespace, field.NewPath("spec", "secretMappings"))...)
 	allErrs = append(allErrs, validateResourceApplyMode(newObject.Spec.ResourceApplyMode, field.NewPath("spec", "resourceApplyMode"))...)
 
 	if len(allErrs) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
```diff
@@ -311,6 +313,19 @@ func validateSecrets(secrets []hivev1.SecretMapping, fldPath *field.Path) field.
 	return allErrs
 }
 
+func validateSourceSecretInSyncSetNamespace(secrets []hivev1.SecretMapping, syncSetNS string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for i, secret := range secrets {
+		if secret.SourceRef.Namespace != syncSetNS {
```
staebler, line 319: This should also allow an empty namespace, which implies the same namespace as the syncset.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -264,6 +264,15 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, err
 	}
 
+	// Prevent users with permissions to create a SyncSet from referencing Secrets in another namespace
+	// and copying them into their clusters. This should be caught in admission, but just incase we have
```
staebler, line 268: s/incase/in case/

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -303,6 +312,19 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 	return reconcile.Result{RequeueAfter: reapplyDuration}, nil
 }
 
+func (r *ReconcileSyncSetInstance) hasSourceSecretsInAnotherNamespace(ssi *hivev1.SyncSetInstance, spec *hivev1.SyncSetCommonSpec) bool {
+	// This check is not relevant for SelectorSyncSets
+	if ssi.Spec.SyncSetRef == nil {
```
staebler, line 317: Do we need this check? This was already checked by the caller. Can we either remove it here or remove it at the caller location?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -303,6 +312,19 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
...
+		return false
+	}
+	for _, secret := range spec.Secrets {
+		if secret.SourceRef.Namespace != ssi.Namespace {
```
staebler, line 321: Allow empty namespace.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -264,6 +264,15 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, err
 	}
 
+	// Prevent users with permissions to create a SyncSet from referencing Secrets in another namespace
+	// and copying them into their clusters. This should be caught in admission, but just incase we have
+	// a check here to refuse to reconcile. Error is logged but not returned as there's no point re-reconciling
+	// this unless the SyncSet is edited. This check is not relevant for SelectorSyncSets which have no namespace.
+	if ssi.Spec.SyncSetRef != nil && r.hasSourceSecretsInAnotherNamespace(ssi, spec) {
+		ssiLog.Warn("syncsetinstance has a source secret in another namespace")
```
staebler, line 272: In case this does get through admission validation, this should bubble up as a condition on the SyncSetInstance. Otherwise, the user will not know why the SyncSet is not applied.

dgoodwin: Because this is supposed to be handled elsewhere and is just a backup check, it felt like we could get away without carrying code to set a condition similar to some of the other checks above this. Would you agree?

staebler: I can live with that. Do we know if there are existing syncsets that violate this new validation?

pkg/apis/hive/v1/validating-webhooks/syncset_validating_admission_hook.go
dgoodwin: Done and added unit test.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
dgoodwin: Scanned stage, doesn't look like there are any violating.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
dgoodwin: Done, didn't add a unit test in this spot because the harness doesn't seem to like an empty namespace and I don't really want to get into that right now.

dgoodwin:
PR Updated.

PR #1009, opened 2020-05-26
Add e2e test to destroy cluster deployment

csrwng:
Adds e2e test to destroy clusterdeployment and ensure that no DNSZones remain after deletion.


hack/e2e-test.sh
```diff
@@ -104,6 +113,12 @@ function teardown() {
 }
 trap 'teardown' EXIT
 
+function save_hive_logs() {
+  oc logs -n hive deployment/hive-operator &> "${ARTIFACT_DIR}/hive-operator.log"
+  oc logs -n hive deployment/hive-controllers &> "${ARTIFACT_DIR}/hive-controllers.log"
+  oc logs -n hive deployment/hiveadmission &> "${ARTIFACT_DIR}/hiveadmission.log"
```
dgoodwin, line 119: Oh man oc logs on a deployment!? What a world.

dgoodwin:
Looks great but something appears to have gone wrong, no uninstall job found.

test/e2e/destroycluster/destroy_test.go
```diff
@@ -0,0 +1,100 @@
...
+			constants.DNSZoneTypeLabel:           constants.DNSZoneTypeChild,
+			constants.ClusterDeploymentNameLabel: cd.Name,
+		})
+	if len(dnsZoneList.Items) > 0 {
+		fail("At least one DNSZone still exists after clusterdeployment deletion")
+	}
+	logger.Info("No DNS Zones related to the cluster exist")
+}
+
+func waitForClusterDeploymentToGoAway(cd *hivev1.ClusterDeployment, fail func(string, ...interface{})) {
```
dgoodwin, line 66: Would it be possible to confirm we saw an uninstall job here, if we expected there to be one? InfraID does transfer to the ClusterDeployment even when the install is underway, so in theory if that is set, I think we *should* see an uninstall job at some point.

csrwng:
> Looks great but something appears to have gone wrong, no uninstall job found.

Yup, the uninstall job completed and then everything was cleaned up, so there's no uninstall job to output the log for... reworking that part.

test/e2e/destroycluster/destroy_test.go
csrwng: Yup, will add code to monitor the uninstall job and output the log

hack/e2e-test.sh
```diff
@@ -86,16 +86,25 @@ DEPLOY_IMAGE="${HIVE_IMAGE}" make deploy
 function teardown() {
 	echo ""
 	echo ""
+        # Skip tear down if the clusterdeployment is no longer there
+        if ! oc get clusterdeployment | grep ${CLUSTER_NAME}; then
```
staebler, line 90: Can this just search for the specific clusterdeployment instead of doing a grep?
```suggestion
        if ! oc get clusterdeployment ${CLUSTER_NAME}; then
```

hack/e2e-test.sh
```diff
@@ -104,6 +113,12 @@ function teardown() {
 }
 trap 'teardown' EXIT
 
+function save_hive_logs() {
+  oc logs -n hive deployment/hive-operator &> "${ARTIFACT_DIR}/hive-operator.log"
```
staebler, line 117: We don't need to capture this. The hive-operator will still be running after the HiveConfig CR is deleted. The normal e2e capture will already capture this.

csrwng:
Addressed comments, added code to write out the uninstall log. Should hopefully pass e2e now.

test/e2e/destroycluster/destroy_test.go
```diff
@@ -33,63 +42,179 @@ func TestDestroyCluster(t *testing.T) {
...
+		fail("Failed to get dns zone list: %v", err)
+	}
 	if len(dnsZoneList.Items) > 0 {
 		fail("At least one DNSZone still exists after clusterdeployment deletion")
 	}
 	logger.Info("No DNS Zones related to the cluster exist")
 }
 
-func waitForClusterDeploymentToGoAway(cd *hivev1.ClusterDeployment, fail func(string, ...interface{})) {
+var writeLogOnce sync.Once
```
staebler, line 88: Can we put this variable inside the scope of `monitorDeprovisionJob` so that we *could* run multiple tests in parallel?

test/e2e/destroycluster/destroy_test.go
```diff
@@ -33,63 +42,179 @@ func TestDestroyCluster(t *testing.T) {
...
+	gvk, err := apiutil.GVKForObject(&batchv1.Job{}, scheme.Scheme)
+	if err != nil {
+		logger.WithError(err).Error("Cannot get GVK for Job")
+		return
+	}
+	restClient, err := apiutil.RESTClientForGVK(gvk, cfg, serializer.NewCodecFactory(scheme.Scheme))
+	if err != nil {
+		logger.WithError(err).Error("Cannot get RESTClient for Job")
+		return
+	}
```
staebler, line 107: Is there any benefit to getting an untyped rest client rather than the typed client?
```
import batchv1client "k8s.io/client-go/kubernetes/typed/batch/v1"
```

```suggestion
	batchClient, err := batchv1client.NewForConfig(cfg)
	if err != nil {
		logger.WithError(err).Error("cannot create batch client")
		return
	}
	restClient := batchClient.RESTClient() 
```

test/e2e/destroycluster/destroy_test.go
```diff
@@ -33,63 +42,179 @@ func TestDestroyCluster(t *testing.T) {
...
+		options.FieldSelector = fmt.Sprintf("metadata.name=%s", deprovisionJobName)
+	}
+	listWatcher := cache.NewFilteredListWatchFromClient(restClient, "jobs", cd.Namespace, nameFilter)
+	monitor := func(event watch.Event) (bool, error) {
+		switch event.Type {
+		case watch.Added:
+			job := event.Object.(*batchv1.Job)
+			logger.Info("Deprovision job created")
+			go writeLogOnce.Do(func() { writeJobLog(job, logger) })
+		case watch.Modified:
```
staebler, line 120: Should the monitor stop if the job is deleted, too? In case the job does not finish. Or do we want to just timeout in that case?

test/e2e/destroycluster/destroy_test.go
```diff
@@ -33,63 +42,179 @@ func TestDestroyCluster(t *testing.T) {
...
+			pod = event.Object.(*corev1.Pod)
+			if pod.Status.Phase == corev1.PodRunning {
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+	listWatcher := cache.NewFilteredListWatchFromClient(client.CoreV1().RESTClient(), "pods", job.Namespace, podFilter)
+	_, err = clientwatch.UntilWithSync(context.TODO(), listWatcher, &corev1.Pod{}, nil, waitForPodRunning)
+	request := client.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{Follow: true})
```
staebler, line 159: This is great!

test/e2e/destroycluster/destroy_test.go
```diff
@@ -0,0 +1,225 @@
...
+	fail := failTestFunc(t, logger)
+
+	c := common.MustGetHiveClient()
+	logger.Info("Deleting cluster deployment")
+	err := c.HiveV1().ClusterDeployments(cd.Namespace).Delete(context.TODO(), cd.Name, metav1.DeleteOptions{})
+	if err != nil {
+		fail("Failed to delete cluster deployment: %v", err)
+	}
+
+	go monitorDeprovisionJob(cd, logger)
```
staebler, line 57: Can we add a parameter that will get set to true when the uninstall job is seen? The destroy test can then fail if the variable is false at the end, indicating that the uninstall job did not actually run.

test/e2e/destroycluster/destroy_test.go
csrwng: no, I started with the generic client, but eventually changed, will fix

test/e2e/destroycluster/destroy_test.go
csrwng: it would still fail if the clusterdeployment never goes away, no?

staebler: Yes, but the check is to ensure that Hive is actually running the uninstall job and not just removing the finalizer and leaking resources.

csrwng: Ack, will fix

csrwng:
Addressed latest comments

csrwng:
@staebler this should be ready, it even passed the e2e :)

csrwng:
well, still need to squash if evereything looks good

csrwng:
squashed, if there are no more comments, this should be ready to merge

PR #1010, opened 2020-05-27

staebler:
The admission webhook for CluterDeployments has been modified to inspect
DELETE requests as well. The webhook will reject any DELETE request for
a ClusterDeployment that has the "hive.openshift.io/protected-delete" annotation
set to a true value. In order to delete such a ClusterDeployment, a user
is required to delete the annotation prior to making the DELETE request.

The "deleteProtection" field has been added to the HiveConfig. When the
field is set to "enabled", the clusterdeployment controller will add the
"hive.openshift.io/protected-delete" annotation to the ClusterDeployment
when transitioning the ClusterDeployment to installed.

### DIFF
--- a/config/crds/hive.openshift.io_hiveconfigs.yaml
+++ b/config/crds/hive.openshift.io_hiveconfigs.yaml
@@ -70,6 +70,15 @@ spec:
                       type: boolean
                   type: object
               type: object
+            deleteProtection:
+              description: DeleteProtection can be set to "enabled" to turn on automatic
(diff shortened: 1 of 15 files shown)


staebler:
One more go at this. This time DELETE requests will be blindly accepted when running on OpenShift 3.11, since the request does not have enough information in it for the validator to make an informed decision about whether to accept the request.

See #996

gregsheremeta:
/hold
#1009 should go first

PR #1011, opened 2020-05-28
ClusterRelocate to move ClusterDeployments between Hive instances

staebler:
When a ClusterRelocate has a label selector that matches with a ClusterDeployment, the clusterrelocate controller will relocate the matching ClusterDeployment.

1. Set the hive.openshift.io/relocate annotation to outgoing on the ClusterDeployment and the DNSZone.
2. Copy secrets, configmaps, machinesets, syncsets, and syncidentityproviders.
3. Copy the DNSZone for the ClusterDeployment, with the relocate annotation set to incoming.
4. Copy the ClusterDeployment, with the relocate annotation set to incoming.
5. Set the relocate annotation on the DNSZone and ClusterDeployemnt to complete.

When a ClusterDeployment with a relocate annotation set to incoming is reconciled by the clusterRelocate controller, the relocate annotation is removed from the DNSZone and the ClusterDeployment. This indicates that the relocate has completed on the destination side.

When a ClusterDeployment has the relocated annotation, no controllers will do any mutation of the remote cluster (such as syncing syncsets). When the ClusterDeployment has the relocated annotation, then controllers will not run finalizer code when the ClusterDeployment is deleted.

Add a hive_cluster_relocations counter vec that tracks the total number of successful cluster relocations. The metric has a single
cluster_relocate label that is the name of the ClusterRelocate directing the relocation.

Add a hive_aborted_cluster_relocations counter vec that tracks the total number of aborted cluster relocations. The metrics has two labels: cluster_relocate and reason. The cluster_relocate label is the name of the ClusterRelocate directing the aborted relocation. The reason label is the reason why the relocate was aborted. Possible values for the reason label are "no_match", "multiple_matches", and "new_match".

The number of failing cluster relocates can be seen by looking at the hive_cluster_deployments_conditions metric and filtering on
a value of RelocationFailed for the condition label.

https://issues.redhat.com/browse/CO-653

### DIFF
--- a/pkg/apis/hive/v1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1/clusterdeployment_types.go
@@ -431,6 +431,19 @@ type CertificateBundleStatus struct {
 	Generated bool `json:"generated"`
 }
 
+// RelocateStatus is the status of a cluster relocate.
+// This is used in the value of the "hive.openshift.io/relocate" annotation.
+type RelocateStatus string
+
+const (
+	// RelocateOutgoing indicates that a resource is on the source side of an in-progress relocate
+	RelocateOutgoing RelocateStatus = "outgoing"
+	// RelocateComplete indicates that a resource is on the source side of a completed relocate
+	RelocateComplete RelocateStatus = "complete"
+	// RelocateIncoming indicates that a resource is on the destination side of an in-progress relocate
+	RelocateIncoming RelocateStatus = "incoming"
+)
+
 func init() {
 	SchemeBuilder.Register(&ClusterDeployment{}, &ClusterDeploymentList{})
 }
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -140,6 +140,16 @@ const (
 	// protected delete is enabled.
 	ProtectedDeleteEnvVar = "PROTECTED_DELETE"
 
+	// RelocateAnnotation is an annotation used on ClusterDeployments and DNSZones to indicate that the resource
+	// is involved in a relocation between Hive instances.
+	// The value of the annotation has the format "{ClusterRelocate}/{Status}", where
+	// {ClusterRelocate} is the name of the ClusterRelocate that is driving the relocation and
+	// {Status} is the status of the relocate. The status is outgoing, completed, or incoming.
+	// An outgoing status indicates that the resource is on the source side of an in-progress relocate.
+	// A completed status indicates that the resource is on the source side of a completed relocate.
+	// An incoming status indicates that the resource is on the destination side of an in-progress relocate.
+	RelocateAnnotation = "hive.openshift.io/relocate"
+
 	// ManagedDomainsFileEnvVar if present, points to a simple text
 	// file that includes a valid managed domain per line. Cluster deployments
 	// requesting that their domains be managed must have a base domain
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -531,6 +531,19 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, nil
 	}
 
+	// If the ClusterDeployment is being relocated to another Hive instance, stop any current provisioning and do not
+	// do any more reconciling.
+	switch _, relocateStatus, err := controllerutils.IsRelocating(cd); {
+	case err != nil:
+		return reconcile.Result{}, errors.Wrap(err, "could not determine relocate status")
+	case relocateStatus == hivev1.RelocateOutgoing:
+		result, err := r.stopProvisioning(cd, cdLog)
+		if result == nil {
+			result = &reconcile.Result{}
+		}
+		return *result, err
+	}
+
 	// Indicate that the cluster is still installing:
 	hivemetrics.MetricClusterDeploymentProvisionUnderwaySeconds.WithLabelValues(
 		cd.Name,
@@ -1173,39 +1186,40 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.Clus
 }
 
 func (r *ReconcileClusterDeployment) syncDeletedClusterDeployment(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (reconcile.Result, error) {
+	switch _, relocateStatus, err := controllerutils.IsRelocating(cd); {
+	case err != nil:
+		cdLog.WithError(err).Error("could not determine relocate status")
+		return reconcile.Result{}, errors.Wrap(err, "could not determine relocate status")
+	case relocateStatus == hivev1.RelocateComplete:
+		cdLog.Info("clusterdeployment relocated, removing finalizer")
+		err := r.removeClusterDeploymentFinalizer(cd, cdLog)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error removing finalizer")
+		}
+		return reconcile.Result{}, err
+	case relocateStatus != "":
+		cdLog.Debug("ClusterDeployment is in the middle of a relocate. Wait until relocate has been completed or aborted before doing finalization.")
+		return reconcile.Result{}, nil
+	}
+
 	if controllerutils.IsDeleteProtected(cd) {
 		cdLog.Error("deprovision blocked for ClusterDeployment with protected delete on")
 		return reconcile.Result{}, nil
 	}
 
-	result, err := r.ensureManagedDNSZoneDeleted(cd, cdLog)
-	if result != nil {
+	switch result, err := r.ensureManagedDNSZoneDeleted(cd, cdLog); {
+	case result != nil:
 		return *result, err
-	}
-	if err != nil {
+	case err != nil:
 		return reconcile.Result{}, err
 	}
 
 	// Wait for outstanding provision to be removed before creating deprovision request
-	if cd.Status.ProvisionRef != nil {
-		provision := &hivev1.ClusterProvision{}
-		switch err := r.Get(context.TODO(), types.NamespacedName{Name: cd.Status.ProvisionRef.Name, Namespace: cd.Namespace}, provision); {
-		case apierrors.IsNotFound(err):
-			cdLog.Debug("linked provision removed")
-		case err != nil:
-			cdLog.WithError(err).Error("could not get provision")
-			return reconcile.Result{}, err
-		case provision.DeletionTimestamp == nil:
-			if err := r.Delete(context.TODO(), provision); err != nil {
-				cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not delete provision")
-				return reconcile.Result{}, err
-			}
-			cdLog.Info("deleted outstanding provision")
-			return reconcile.Result{RequeueAfter: defaultRequeueTime}, nil
-		default:
-			cdLog.Debug("still waiting for outstanding provision to be removed")
-			return reconcile.Result{RequeueAfter: defaultRequeueTime}, nil
-		}
+	switch result, err := r.stopProvisioning(cd, cdLog); {
+	case result != nil:
+		return *result, err
+	case err != nil:
+		return reconcile.Result{}, err
 	}
 
 	// Skips creation of deprovision request if PreserveOnDelete is true and cluster is installed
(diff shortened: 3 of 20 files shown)


pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,797 @@
...
+	}
+	return nil
+}
+
+func (r *ReconcileClusterRelocate) finishRelocateCompletion(cd *hivev1.ClusterDeployment, relocateName string, logger log.FieldLogger) (reconcile.Result, error) {
+	if err := r.setRelocateAnnotation(cd, relocateName, hivev1.RelocateComplete, logger); err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "could not set relocate status to complete")
+	}
+
+	metricSuccessfulClusterRelocations.WithLabelValues(relocateName).Inc()
```
joelddiaz, line 356: are we risking multiple counting if errors are encountered below?

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,621 @@
...
+	clusterDeployments := &hivev1.ClusterDeploymentList{}
+	if err := r.List(context.Background(), clusterDeployments); err != nil {
+		r.logger.WithError(err).
+			WithField("clusterRelocate", clusterRelocate.Name).
+			Log(controllerutils.LogLevel(err), "failed to list clusterdeployments")
+		return
+	}
+
+	for _, cd := range clusterDeployments.Items {
+		if cd.Annotations[constants.RelocatingAnnotation] != clusterRelocate.Name &&
```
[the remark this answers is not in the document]
staebler: This check is meant to filter out the ClusterDeployments that are not related to the ClusterRelocate. If the ClusterDeployment does not have a relocating annotation matching the ClusterRelocate **AND** does not have a label matching the ClusterRelocate, then the ClusterDeployment is not queued.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,621 @@
...
+	}
+
+	return
+}
+
+// ReconcileClusterRelocate is the reconciler for ClusterRelocate. It will sync on ClusterDeployment resources
+// and relocate those that match with a ClusterRelocate.
+type ReconcileClusterRelocate struct {
+	client.Client
+	scheme *runtime.Scheme
```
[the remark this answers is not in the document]
staebler: It is not. I have removed it.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,621 @@
...
+			); changed {
+				cd.Status.Conditions = conds
+				if err := r.Status().Update(context.Background(), cd); err != nil {
+					logger.WithError(err).Log(controllerutils.LogLevel(err), "could not update condition")
+					// return the move error rather than the update error
+				}
+			}
+			return err
+		} else if !completed {
+			return nil
```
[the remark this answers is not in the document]
staebler: The `completed` return has been removed.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
staebler: Yes. This was a carry-over from when the controller would not reconcile the ClusterDeployment any more once the relocated annotation was set. That has been fixed now so that the controller will still attempt to delete the ClusterDeployment again if there was an error. In light of that, I have moved the metrics increment to after the delete.

staebler:
This has been tested now and is ready for further review.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -431,6 +431,18 @@ type CertificateBundleStatus struct {
 	Generated bool `json:"generated"`
 }
 
+// RelocateStatus is the status of a cluster relocate
+type RelocateStatus string
```
dgoodwin, line 435: Noting the annotation it's used in would be useful context.

pkg/constants/constants.go
```diff
@@ -136,6 +136,16 @@ const (
...
 	ProtectedDeleteEnvVar = "PROTECTED_DELETE"
 
+	// RelocateAnnotation is an annotation used on ClusterDeployments and DNSZones to indicate that the resource
+	// is involved in a relocation between Hive instances.
+	// The value of the annotation has the format "{ClusterRelocate}/{Status}", where
+	// {ClusterRelocate} is the name of the ClusterRelocate that is driving the relocation and
+	// {Status} is the status of the relocate. The status is outgoing, completed, or incoming.
+	// An outgoing status indicates that the resource is on the source side of an in-progress relocate.
+	// A completed status indicates that the resource is on the source side of a completed relocate.
+	// An incoming status indicates that the resource os on the destination side of an in-progress relocate.
```
dgoodwin, line 146: s/os/is

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1173,47 +1186,45 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.Clus
...
+		cdLog.WithError(err).Error("could not determine relocate status")
+		return reconcile.Result{}, errors.Wrap(err, "could not determine relocate status")
+	case relocateStatus == hivev1.RelocateComplete:
+		cdLog.Info("clusterdeployment relocated, removing finalizer")
+		err := r.removeClusterDeploymentFinalizer(cd, cdLog)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error removing finalizer")
+		}
+		return reconcile.Result{}, err
+	}
```
dgoodwin, line 1200: Can you clarify why we only case for relocation complete? It feels really dangerous to move past this block if we're in any other relocating state.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,812 @@
...
+
+	metricSuccessfulClusterRelocations = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_cluster_relocations",
+		Help: "Total number of successful cluster relocations.",
+	}, []string{"cluster_relocate"})
+
+	metricAbortedClusterRelocations = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_aborted_cluster_relocations",
+		Help: "Total number of aborted cluster relocations.",
+	}, []string{"cluster_relocate", "reason"})
```
dgoodwin, line 60: I believe both of these metrics should be counters, not gauges. Gauge would be for a value reflecting a current state, in this case it looks like you'd get the number of relocations since the pod was restarted. Counter will give us an ever incrementing value and allow the use of rate.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,812 @@
...
+		return reconcile.Result{}, err
+	}
+
+	currentRelocateName, relocateStatus, err := controllerutils.IsRelocating(cd)
+	if err != nil {
+		logger.WithError(err).Error("could not determine relocate status")
+		return reconcile.Result{}, errors.Wrap(err, "could not determine relocate status")
+	}
+
+	// Complete the relocation on the destination side
```
dgoodwin, line 183: Please clarify here the logic, because the cluster deployment is the last piece, and we're recocniling on a ClusterDeployment marked incoming, thus the destination side, and thus must be complete.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,812 @@
...
+
+	return r.reconcileSingleMatch(cd, relocateStatus, currentRelocateName, desiredRelocates[0], logger)
+}
+
+func (r *ReconcileClusterRelocate) reconcileSingleMatch(cd *hivev1.ClusterDeployment, oldRelocateStatus hivev1.RelocateStatus, oldRelocateName string, desiredRelocate *hivev1.ClusterRelocate, logger log.FieldLogger) (reconcile.Result, error) {
+	// Abort an in-progress relocate when the ClusterDeployment matches a different ClusterRelocate
+	if oldRelocateStatus == hivev1.RelocateOutgoing && oldRelocateName != desiredRelocate.Name {
+		logger.WithField("currentClusterRelocate", oldRelocateName).
+			WithField("desiredClusterRelocate", desiredRelocate.Name).
+			Info("switching relocation")
```
dgoodwin, line 215: I'd go Warn here, this is pretty unnerving if it were to happen.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: I should have cases for the other relocating states, too. If it is outgoing or incoming, then the clusterdeployment controller should wait for the clusterrelocate controller to clear the state prior to deprovisioning. Good catch.

pkg/installmanager/installmanager.go
```diff
@@ -44,6 +43,8 @@ import (
 	clientwatch "k8s.io/client-go/tools/watch"
 	"k8s.io/utils/pointer"
 
+	k8slabels "github.com/openshift/hive/pkg/util/labels"
```
dgoodwin, line 46: Should be with the other hive imports.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -521,14 +522,18 @@ func (r *ReconcileClusterRelocate) copy(cd *hivev1.ClusterDeployment, destClient
...
+		logger = logger.WithField("type", reflect.TypeOf(dnsZone)).WithField("resource", dnsZone.Name)
 		if err := r.copyResource(dnsZone, destClient, false, logger); err != nil {
 			return errors.Wrap(err, "failed to copy dnszone")
 		}
 	}
 
 	// copy clusterdeployment
-	if err := r.copyResource(cd, destClient, true, logger); err != nil {
-		return errors.Wrap(err, "failed to copy clusterdeployment")
+	{
```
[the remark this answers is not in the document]
staebler: To keep the `logger` variable isolate to inside the scope, in case there is other code added after the code block later that may use `logger` and not want the type and resource field. But, now I see that I replaced the function-scoped `logger` variable instead of creating a new `logger` variable.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
staebler: Definitely!

pkg/controller/utils/dnszone.go
```diff
@@ -0,0 +1,101 @@
+package utils
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
```
joelddiaz, line 8: the grouping of the imports seems inconsistent. only one controller-runtime import is by itself.

pkg/apis/hive/v1/clusterdeployment_types.go
dgoodwin: I'm still not seeing the annotation it's used in mentioned here, something missed?

staebler: Oh, I added it to `ClusterRelocateStatus` but not here. That brings to light that I have a `ClusterRelocateStatus` and a `RelocateStatus`. That is too confusing. I need to rename one or both of those.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -0,0 +1,826 @@
...
+		owner := &corev1.Secret{}
+		if err := r.Get(context.Background(), client.ObjectKey{Namespace: secret.Namespace, Name: ownerRef.Name}, owner); err != nil {
+			logger.WithField("owner", ownerRef.Name).WithError(err).Log(controllerutils.LogLevel(err), "could not get owner secret")
+			return false, err
+		}
+		if owner.Type == corev1.SecretTypeServiceAccountToken {
+			logger.Info("ignoring secret owned by a service account token secret")
+			return true, nil
+		}
+	}
```
dgoodwin, line 770: In testing, I did something dumb and migrated a cluster in my "hive" namespace. I ended up getting my hiveadmission-serving-cert copied over the one in the other cluster which it believed was a diff. 

I feel like we're wading into waters where hive doesn't really support multiple clusters in one namespace. 

Would it be possible to modify this to pass in the cluster deployment and ignore secrets that are not in one of our known secret fields?

staebler: That is a TODO, but something that I would like to hold off on.

staebler: I have not prioritized it since I do not believe that any of our users do multiple clusterdeployments in a single namespace.

staebler: But even if they did, it would still work. The problem is when you have secrets with the same name in both clusters, which would not happen from just a ClusterDeployment.

dgoodwin: :+1:

dgoodwin:
Despite the minor annotation thing I'm going to lgtm and see if we can get this to merge today.

staebler:
Let me fix the annotation comment and address Joel's comment and squash everything down.

PR #1012, opened 2020-06-05
Handle applying resources too large to include in an annotation

csrwng:
If a resource is too large to include in an annotation, we currently fail to apply it via syncsets. This PR adds an applyBehavior field that can be set to either Apply (default), CreateOrUpdate, or CreateOnly
https://issues.redhat.com/browse/CO-379

CreateOrUpdate allows large resources to be synced, and its behavior is almost like Apply. However, we lose some functionality by not having the original resource in an annotation.

### What we lose with CreateOrUpdate
Because the original configuration is not present in the object, we lose the ability to delete previously-created keys in structs. (Lists should be ok afaik)
An example:
My original syncset: { "foo": "1", "bar": "2", "baz": "3"} 
That gets created on the target cluster...
then I change my syncset to {"foo": "1", "bar": "2"}
After applying the new change, the resource will remain unchanged on the target cluster because apply doesn't know that my intent was to delete "baz".

### One alternative:
With every update, we delete and recreate the resource - drawback: the resource will get re-created every time we resync syncsetinstances.

### Another alternative:
We replace the content of the resource with whatever is in the syncset, only keeping cluster-created metadata (resourceVersion, selfLink, etc). drawback: we don't know if there is server-side defaulting or some other change that gets applied every time that we create/update the resource, causing churn with every re-apply.

### DIFF
--- a/pkg/apis/hive/v1/syncset_types.go
+++ b/pkg/apis/hive/v1/syncset_types.go
@@ -22,25 +22,26 @@ const (
 )
 
 // SyncSetApplyBehavior is a string representing the behavior to use when
-// aplying a syncset to t arget cluster.
+// aplying a syncset to target cluster.
+// +kubebuilder:validation:Enum="";Apply;CreateOnly;CreateOrUpdate
 type SyncSetApplyBehavior string
 
 const (
 	// ApplySyncSetApplyBehavior is the default apply behavior. It will result
 	// in resources getting applied using the 'oc apply' command to the target
 	// cluster.
-	ApplySyncSetApplyBehavior = "Apply"
+	ApplySyncSetApplyBehavior SyncSetApplyBehavior = "Apply"
 
 	// CreateOnlySyncSetApplyBehavior results in resources only getting created
 	// if they do not exist, otherwise they are left alone.
-	CreateOnlySyncSetApplyBehavior = "CreateOnly"
+	CreateOnlySyncSetApplyBehavior SyncSetApplyBehavior = "CreateOnly"
 
 	// CreateOrUpdateSyncSetApplyBehavior results in resources getting created if
 	// they do not exist, otherwise they are updated with the contents of the
 	// syncset resource. This is different from Apply behavior in that an annotation
 	// is not added to the target resource with the "lastApplied" value. It allows
 	// for syncing larger resources, but loses the ability to sync map entry deletes.
-	CreateOrUpdateSyncSetApplyBehavior = "CreateOrUpdate"
+	CreateOrUpdateSyncSetApplyBehavior SyncSetApplyBehavior = "CreateOrUpdate"
 )
 
 // SyncSetPatchApplyMode is a string representing the mode with which to apply
--- a/pkg/apis/hive/v1/syncset_types.go
+++ b/pkg/apis/hive/v1/syncset_types.go
@@ -21,6 +21,28 @@ const (
 	SyncResourceApplyMode SyncSetResourceApplyMode = "Sync"
 )
 
+// SyncSetApplyBehavior is a string representing the behavior to use when
+// aplying a syncset to t arget cluster.
+type SyncSetApplyBehavior string
+
+const (
+	// ApplySyncSetApplyBehavior is the default apply behavior. It will result
+	// in resources getting applied using the 'oc apply' command to the target
+	// cluster.
+	ApplySyncSetApplyBehavior = "Apply"
+
+	// CreateOnlySyncSetApplyBehavior results in resources only getting created
+	// if they do not exist, otherwise they are left alone.
+	CreateOnlySyncSetApplyBehavior = "CreateOnly"
+
+	// CreateOrUpdateSyncSetApplyBehavior results in resources getting created if
+	// they do not exist, otherwise they are updated with the contents of the
+	// syncset resource. This is different from Apply behavior in that an annotation
+	// is not added to the target resource with the "lastApplied" value. It allows
+	// for syncing larger resources, but loses the ability to sync map entry deletes.
+	CreateOrUpdateSyncSetApplyBehavior = "CreateOrUpdate"
+)
+
 // SyncSetPatchApplyMode is a string representing the mode with which to apply
 // SyncSet Patches.
 type SyncSetPatchApplyMode string
@@ -199,12 +221,17 @@ type SyncSetCommonSpec struct {
 	// +optional
 	Secrets []SecretMapping `json:"secretMappings,omitempty"`
 
-	// OmitAnnotation if true indicates that the usual annotation used by the apply command
-	// will be omitted from the target resources. This results in some loss of functionality
-	// when syncing removal of map entries, but it makes it possible to sync large resources
-	// that otherwise could not be synced.
+	// ApplyBehavior indicates how resources in this syncset will be applied to the target
+	// cluster. The default value of "Apply" indicates that resources should be applied
+	// using the 'oc apply' command. If no value is set, "Apply" is assumed.
+	// A value of "CreateOnly" indicates that the resource will only be created if it does
+	// not already exist in the target cluster. Otherwise, it will be left alone.
+	// A value of "CreateOrUpdate" indicates that the resource will be created/updated without
+	// the use of the 'oc apply' command, allowing larger resources to be synced, but losing
+	// some functionality of the 'oc apply' command such as the ability to remove annotations,
+	// labels, and other map entries in general.
 	// +optional
-	OmitAnnotation bool `json:"omitAnnotation,omitempty"`
+	ApplyBehavior SyncSetApplyBehavior `json:"applyBehavior,omitempty"`
 }
 
 // SelectorSyncSetSpec defines the SyncSetCommonSpec resources and patches to sync along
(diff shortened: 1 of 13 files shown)


staebler:
I am confused about this fix. As I understand the code, the Apply does not record the change in an annotation. The `setupApplyCommand` sets record to false and a noop recorder.
https://github.com/openshift/hive/blob/a27a3ae210239c5c6475f09dac65ba22b4fd0106/pkg/resource/apply.go#L76

staebler:
> I am confused about this fix. As I understand the code, the Apply does not record the change in an annotation. The `setupApplyCommand` sets record to false and a noop recorder.
> 
> https://github.com/openshift/hive/blob/a27a3ae210239c5c6475f09dac65ba22b4fd0106/pkg/resource/apply.go#L76

And I ask this because I thought that it would have been easier to turn recording off for the Apply rather than switch to a create or update. But I see that recording is already off.

staebler:
> > I am confused about this fix. As I understand the code, the Apply does not record the change in an annotation. The `setupApplyCommand` sets record to false and a noop recorder.
> > https://github.com/openshift/hive/blob/a27a3ae210239c5c6475f09dac65ba22b4fd0106/pkg/resource/apply.go#L76
> 
> And I ask this because I thought that it would have been easier to turn recording off for the Apply rather than switch to a create or update. But I see that recording is already off.

Nevermind. I see it is done with the `CreateApplyAnnotation` instead.
https://github.com/kubernetes/kubectl/blob/f9e4fa6b9cff11b6e2949b76680b8cd5b8192eab/pkg/util/apply.go#L130

pkg/resource/apply.go
```diff
@@ -55,6 +64,9 @@ func (r *Helper) Apply(obj []byte) (ApplyResult, error) {
 	}
 	err = applyOptions.Run()
 	if err != nil {
+		if annotationTooLong(err) {
```
staebler, line 67: I think that I would rather this be something directed by the user rather than fallen back to. In other words, the user should have to explicitly specify in the syncset that they want to use creates and updates rather than applies. That way the user is not surprised by behavior that they would not really have insight into.

pkg/resource/apply.go
```diff
@@ -152,3 +208,20 @@ func (t *changeTracker) ToPrinter(name string) (printers.ResourcePrinter, error)
...
+	if !errors.IsInvalid(err) {
+		return false
+	}
+	apiStatusErr, ok := err.(errors.APIStatus)
+	if !ok {
+		return false
+	}
+	causes := apiStatusErr.Status().Details.Causes
+	for _, c := range causes {
+		if c.Type == fieldTooLong && c.Field == "metadata.annotations" {
```
staebler, line 222: Do we want to return true if and only if *all* of the errors are annotations too long? If there are other errors, then the create and update will presumably still fail.

csrwng:
Updated to use a boolean in the syncset spec to determine which apply method is used. 
I decided to use this instead of a different mode as discussed earlier because the mode (sync/upsert) is orthogonal to whether the resource is synced with annotations or not.

One thing that you don't get with this approach is that you will necessarily have to create a separate syncset for large resources and lose the ordering benefit of including that large resource with other resources in the same syncset.

An alternative to this approach, would be to watch for an annotation in the resource itself that indicates that you want to omit annotations. The annotation wouldn't need to be synced but would indicate to us which method you want to use. This would allow you to include a large resource along with other resources in the same syncset. Please let me know if that sounds preferable.

@abutcher ptal

abutcher:
> One thing that you don't get with this approach is that you will necessarily have to create a separate syncset for large resources and lose the ordering benefit of including that large resource with other resources in the same syncset.
> 
> An alternative to this approach, would be to watch for an annotation in the resource itself that indicates that you want to omit annotations. The annotation wouldn't need to be synced but would indicate to us which method you want to use. This would allow you to include a large resource along with other resources in the same syncset. Please let me know if that sounds preferable.

I think I prefer the latter approach but pinging @jewzaam to see if ordering is important in his case.

jewzaam:
In this case order is not important, the large resource we're hitting issues with is the only resource in the SSS.  +1 to no surprise and having to turn this on explicitly on the SSS.  Am I reading the PR correctly that it will do a patch if this option is enabled and it's an update to an existing object?  This will be a surprise if I don't read how this works.  The use case I have right now would be OK, we only have one data property in a secret to sync.  But patches would be surprising for the reasons Cesar noted in the PR's description.

csrwng:
> Am I reading the PR correctly that it will do a patch if this option is enabled and it's an update to an existing object?

Yes it will do a patch, just like Apply does on updates. The difference in this case is we don't have the resource that was previously applied as we do with normal apply and the annotation.

csrwng:
As noted in the description, the main thing you lose is the ability to delete keys in maps ... (Annotations, Labels, and Data entries in configmaps/secrets). Just thinking out loud...  one way to preserve the regular apply functionality would be to somehow keep the previously applied value in Hive (maybe in the spec of the syncset instance?)
@abutcher @staebler thoughts?

jewzaam:
Any reason why you're not able to use the current data in the resource directly for a diff?

csrwng:
> Any reason why you're not able to use the current data in the resource directly for a diff?

The current data in the resource has potentially other changes applied on the server side... at the very least metadata like uuid, selfLink, etc. as well as any defaulting applied to it by the apiserver. This is the reason that apply stores the value it's applying in an annotation ... so that it has something to compare to without the server-side changes.

pkg/apis/hive/v1/syncset_types.go
```diff
@@ -198,6 +198,13 @@ type SyncSetCommonSpec struct {
 	// Secrets is the list of secrets to sync along with their respective destinations.
 	// +optional
 	Secrets []SecretMapping `json:"secretMappings,omitempty"`
+
+	// OmitAnnotation if true indicates that the usual annotation used by the apply command
+	// will be omitted from the target resources. This results in some loss of functionality
+	// when syncing removal of map entries, but it makes it possible to sync large resources
+	// that otherwise could not be synced.
+	// +optional
+	OmitAnnotation bool `json:"omitAnnotation,omitempty"`
```
staebler, line 207: Can we use an enum instead of a bool? Using a bool locks us in only supporting 2 values. Maybe this field could be something like `ApplyBehavior` with values of "Apply" or "CreateUpdate".

jewzaam: If following `oc` commands would you be open to "Create", "Replace", and "Apply"?  Asking to see if we can sneak in some behavior I've wanted and we had some discussion on: one time sync.  If "Create", only do it if the resource doesn't exist.  Yes, it may get deleted and I think then the correct behavior is to re-create it if we do this.  "Replace" would be without any annotation.  "Apply" would be the current (pre PR) behavior.

csrwng: Hmm, definitely worth considering. Let me check with the team. However, in the case of this code, we're not really doing a Replace but an Update (in case it exists) or Create if it doesn't. My understanding is that Replace involves deleting and recreating the resource. Not sure we want to do that periodically for anything.

pkg/resource/apply.go
```diff
@@ -30,6 +37,8 @@ var (
 	UnknownApplyResult ApplyResult = "unknown"
 )
 
+const fieldTooLong metav1.CauseType = "FieldValueTooLong"
```
akhil-rane, line 40: I did not understand why only this const is declared outside const() block above

pkg/apis/hive/v1/syncset_types.go
jewzaam: Hmm.. yea we don't need a delete+create, wasn't my intent to ask for that.  Update sound good.

pkg/apis/hive/v1/syncset_types.go
```diff
@@ -21,6 +21,28 @@ const (
...
 
+// SyncSetApplyBehavior is a string representing the behavior to use when
+// aplying a syncset to t arget cluster.
+type SyncSetApplyBehavior string
+
+const (
+	// ApplySyncSetApplyBehavior is the default apply behavior. It will result
+	// in resources getting applied using the 'oc apply' command to the target
+	// cluster.
+	ApplySyncSetApplyBehavior = "Apply"
```
staebler, line 32:
```suggestion
	ApplySyncSetApplyBehavior SyncSetApplyBehavior = "Apply"
```

pkg/apis/hive/v1/syncset_types.go
```diff
@@ -21,6 +21,28 @@ const (
 	SyncResourceApplyMode SyncSetResourceApplyMode = "Sync"
 )
 
+// SyncSetApplyBehavior is a string representing the behavior to use when
+// aplying a syncset to t arget cluster.
+type SyncSetApplyBehavior string
```
staebler, line 26:
```suggestion
// +kubebuilder:validation:Enum="";Apply;CreateOnly;CreateOrUpdate
type SyncSetApplyBehavior string
```

dgoodwin: Now that is cool.

csrwng:
Comments addressed

PR #1013, opened 2020-06-08
Add docs on how to vendor the installer.

twiest:
/cc @csrwng

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -23,6 +23,7 @@
   - [Dependency management](#dependency-management)
     - [Updating Dependencies](#updating-dependencies)
     - [Re-creating vendor Directory](#re-creating-vendor-directory)
+    - [Vendoring the OpenShift Installer](#vendoring-the-openshift-installer)
   - [Running the e2e test locally](#running-the-e2e-test-locally)
   - [Viewing Metrics with Prometheus](#viewing-metrics-with-prometheus)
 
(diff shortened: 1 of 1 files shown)


docs/developing.md
```diff
@@ -266,6 +267,74 @@ To recreate *_vendor_* directory, you can run the following command:
...
+* The `go.mod` file contains a section called `replace`. The purpose of this section is to force go modules to use a specific version of that dependency.
+* `replace` directives may need to be copied from the OpenShift Installer or possibly other Hive dependencies. In other words, any dependency may need to be pinned to a specific version.
+* `go mod tidy` is used to add (download) missing dependencies and to remove unused modules.
+* `go mod vendor` is used to copy dependent modules to the vendor directory.
+
+
+The following is a basic flow for vendoring the latest OpenShift Installer. Updating Go modules can sometimes be complex, so it is likely that the flow below will not encompass everything needed to vendor the latest OpenShift Installer. If more steps are needed, please document them here so that the Hive team will know other possible things that need to be done.
+
+Basic flow:
+* Vendor the latest installer. In order to vendor the installer, you must use `go get` with a specific source package. For instance:
```
staebler, line 285: This is not usually the step that I take. I usually change the installer line in the go.mod file directly by replacing the current version with `master`. Then I run `make vendor`, which does `go mod tidy and go mod vendor`.

docs/developing.md
```diff
@@ -266,6 +267,74 @@ To recreate *_vendor_* directory, you can run the following command:
...
+```
+go get github.com/openshift/installer/pkg/asset/machines/vsphere@master
```
staebler, line 287: Why do you get a source package instead of `go get github.com/openshift/installer@master`?

docs/developing.md
```diff
@@ -266,6 +267,74 @@ To recreate *_vendor_* directory, you can run the following command:
 make vendor
 ```
 
+### Vendoring the OpenShift Installer
+
+For various reasons, Hive vendors the OpenShift Installer. The OpenShift installer brings in quite a few dependencies, so it is important to know the general flow of how to vendor the latest version of the OpenShift Installer.
+
+Things to note:
+* Hive vendors the installer from `@master`, NOT from `@latest`. For go modules, `@latest` means the latest github release, which for the installer is not up to date.
```
csrwng, line 275: @latest = latest git tag ... not sure if that is the same as github release

docs/developing.md
```diff
@@ -266,6 +267,74 @@ To recreate *_vendor_* directory, you can run the following command:
...
+
+The following is a basic flow for vendoring the latest OpenShift Installer. Updating Go modules can sometimes be complex, so it is likely that the flow below will not encompass everything needed to vendor the latest OpenShift Installer. If more steps are needed, please document them here so that the Hive team will know other possible things that need to be done.
+
+Basic flow:
+* Vendor the latest installer. In order to vendor the installer, you must use `go get` with a specific source package. For instance:
+```
+go get github.com/openshift/installer/pkg/asset/machines/vsphere@master
+```
+
+* If `go get` errors with a message like the following, then check the installer's replace directives for that go module. This will ensure that  Hive is pulling in the same version. Re-run the `go get` once the replace directive has been added or updated. This process may need to be followed several times to clean up all of the errors.
```
csrwng, line 290: It may be good to preemptively look at the installer's `go.mod` and update any replaces that we have copied from it on our own `go.mod`.

docs/developing.md
```diff
@@ -266,6 +267,74 @@ To recreate *_vendor_* directory, you can run the following command:
...
+	sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1beta1: module sigs.k8s.io/cluster-api-provider-aws@latest found (v0.5.3, replaced by github.com/openshift/cluster-api-provider-aws@v0.2.1-0.20200316201703-923caeb1d0d8), but does not contain package sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1beta1
+```
+
+* If `go mod tidy` errors with a message like the following, then check the installer's replace directives for that go module so that Hive is pulling in the same version. Re-run the `go mod tidy` once the replace directive has been added or updated. This process may need to be followed several times to clean up all of the errors.
+```
+go: sigs.k8s.io/cluster-api-provider-azure@v0.0.0: reading sigs.k8s.io/cluster-api-provider-azure/go.mod at revision v0.0.0: unknown revision v0.0.0
+```
+
+* If `go mod tidy` errors with a message like the following, then check the installer's replace directives to see if the replace needs to be update. In this specific case, the replace was correct, but Hive is referring to `awsproviderconfig/v1beta1`, but the module has renamed that directory to `awsprovider/v1beta1`. Fix the Hive code and re-run `go mod tidy`
```
csrwng, line 319: s/needs to be update/needs to be updated/
s/but the module has renamed that directory/and the module has renamed that directory/

docs/developing.md
twiest: Done.

csrwng:
lgtm

docs/developing.md
staebler: Ah, yes. That is because you should be using the `-d` flag since you do not want to actually install the module. You should also probably use the `-u` flag as well.

docs/developing.md
```diff
@@ -266,6 +267,66 @@ To recreate *_vendor_* directory, you can run the following command:
...
+* `go mod vendor` is used to copy dependent modules to the vendor directory.
+
+
+The following is a basic flow for vendoring the latest OpenShift Installer. Updating Go modules can sometimes be complex, so it is likely that the flow below will not encompass everything needed to vendor the latest OpenShift Installer. If more steps are needed, please document them here so that the Hive team will know other possible things that need to be done.
+
+Basic flow for vendoring the latest OpenShift Installer:
+
+* Compare the `replace` section of the OpenShift Installer `go.mod` with the `replace` section of the Hive `go.mod`. If the Hive `replace` section has the same module listed as the OpenShift Installer `replace` section, ensure that the Hive version matches the installer version. If it doesn't, change the Hive version to match.
+
+* Edit `go.mod` and change the OpenShift Installer `require` to reference `master` instead of the last version. Go will change it to the version that master points to, so this is a temporary change.
```
staebler, line 288:
```suggestion
* Edit `go.mod` and change the OpenShift Installer `require` to reference `master` (or a specific commit hash) instead of the last version. Go will change it to the version that master points to, so this is a temporary change.
```

twiest: Fixed, thanks!

twiest:
@staebler I believe I've addressed your feedback, PTAL.

twiest:
@staebler @csrwng not sure why it didn't apply the label, but would one of you mind LGTM-ing again please?

PR #1014, opened 2020-06-08
Fix panic while adding machinepool in existing vpc

akhil-rane:
Currently, adding machinepool to the cluster installed in existing vpc panics
when subnets are invalid. This is because kubernetes client used to update
machinepool object is not initialized. This PR fixes it.

jira: https://issues.redhat.com/browse/CO-992


dgoodwin:
Looks good just has a legit formatting failure.

PR #1017, opened 2020-06-17
provide a means to abandon deprovisioning

staebler:
When there are AWS resources that cannot be destroyed, the list of blocked ARNs is added to the ClusterDeprovision status in the `blockedResources` field. The user can then take action on those blocked resources.

The AWS destroyer will back off destroy attempts. Each attempt is limited to 5 minutes. The backoff starts at 5 minutes, doubles after each failed attempt, and caps at 24 hours.

If the user wants to abandon a deprovision, the user can add the "hive.openshift.io/abandon-deprovision" annotation to the ClusterDeployment. When this annotation is present with a true value, the clusterdeployment controller will remove the deprovison finalizer from the ClusterDeployment without waiting for the ClusterDeprovision to complete.

https://issues.redhat.com/browse/CO-943

### DIFF
--- a/contrib/pkg/deprovision/awstagdeprovision.go
+++ b/contrib/pkg/deprovision/awstagdeprovision.go
@@ -1,6 +1,7 @@
 package deprovision
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -10,24 +11,36 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/installer/pkg/destroy/aws"
 	"github.com/openshift/library-go/pkg/controller/fileobserver"
 
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
 )
+
+type awsTagDeprovisionOpts struct {
+	logLevel           string
+	region             string
+	filters            []aws.Filter
+	clusterDeprovision string
+	logger             log.FieldLogger
+}
 
 // NewDeprovisionAWSWithTagsCommand is the entrypoint to create the 'aws-tag-deprovision' subcommand
 // TODO: Port to a sub-command of deprovision.
 func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
-	opt := &aws.ClusterUninstaller{}
-	var logLevel string
+	opt := &awsTagDeprovisionOpts{}
 	cmd := &cobra.Command{
 		Use:   "aws-tag-deprovision KEY=VALUE ...",
 		Short: "Deprovision AWS assets (as created by openshift-installer) with the given tag(s)",
 		Long:  "Deprovision AWS assets (as created by openshift-installer) with the given tag(s).  A resource matches the filter if any of the key/value pairs are in its tags.",
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := completeAWSUninstaller(opt, logLevel, args); err != nil {
+			if err := opt.complete(args); err != nil {
 				log.WithError(err).Error("Cannot complete command")
 				return
 			}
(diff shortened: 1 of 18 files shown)


staebler:
/hold

This is using a fork of the installer until https://github.com/openshift/installer/pull/3765 merges.

staebler:
This is an example of the status of a ClusterDeprovision when there is an untagged instance connected to the VPC.

```
  status:
    blockedResources:
    - arn:aws:ec2:us-east-1:125931421481:dhcp-options/dopt-0b5ec6b5149cd3f58
    - arn:aws:ec2:us-east-1:125931421481:network-interface/eni-044b8a17b7c24a0fa
    - arn:aws:ec2:us-east-1:125931421481:subnet/subnet-02afe98c8ac60c7a7
    - arn:aws:ec2:us-east-1:125931421481:vpc/vpc-0f19f125f3d24bfc9
```

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -65,6 +81,66 @@ func completeAWSUninstaller(o *aws.ClusterUninstaller, logLevel string, args []s
...
+				Filters: o.filters,
+				Logger:  o.logger,
+				Region:  o.region,
+			}
+			blockedResources, err := uninstaller.RunWithContext(ctx)
+			if len(blockedResources) == 0 {
+				return true, err
+			}
+			if o.clusterDeprovision == "" {
+				return
```
joelddiaz, line 106: Do we not also return `err` here even if we will not be posting status to a ClusterDeprovision object?

staebler: If there are any blocked resources, then we want to keep trying to uninstall in the backoff loop. If we return `err` here, then the uninstall will stop and the pod will fail.

However, maybe it would be good to distinguish between an error because the context expired and an error for other reasons. In the former case, we want to keep trying. In the latter case, we want to abort. I'll look into that.

staebler:
Now blocked on https://github.com/openshift/installer/pull/3772.

pkg/clusterresource/vsphere.go
```diff
@@ -46,9 +46,6 @@ type VSphereCloudBuilder struct {
 	// IngressVIP is the virtual IP address for ingress
 	IngressVIP string
 
-	// DNSVIP is the virtual IP address for DNS
-	DNSVIP string
```
dgoodwin, line 50: A little strange, is this meant to be in this PR?

pkg/apis/hive/v1/clusterdeprovision_types.go
```diff
@@ -21,6 +21,9 @@ type ClusterDeprovisionSpec struct {
 type ClusterDeprovisionStatus struct {
 	// Completed is true when the uninstall has completed successfully
 	Completed bool `json:"completed,omitempty"`
+
+	// BlockedResources is a list of cloud resources that the deprovision has not been able to delete
+	BlockedResources []string `json:"blockedResources,omitempty"`
```
dgoodwin, line 26: How are things looking for the other cloud providers, will a flat list of strings be sufficient as far as we can see? Wondering if we should go with something that lets us store a little more data or not.

pkg/clusterresource/vsphere.go
staebler: Yes. The `DNSVIP` was removed from the installer types in the version of the installer that I am vendoring.

pkg/apis/hive/v1/clusterdeprovision_types.go
staebler: That's a good idea. I can change `BlockedResources` to a struct that for now has nothing but a name field.

dgoodwin:
Installer PR merged but we've got major conflicts in here now.

staebler:
> Installer PR merged but we've got major conflicts in here now.

Only one of the two installer PRs has merged. The openshift/installer#3772 PR still has not.

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -120,6 +133,66 @@ func completeAWSUninstaller(o *aws.ClusterUninstaller, logLevel string, args []s
...
+func (o *awsTagDeprovisionOpts) run() error {
+	return wait.ExponentialBackoff(
+		// Start the backoff at 5 minutes, double it each time, to a cap of 24 hours.
+		wait.Backoff{
+			Duration: 5 * time.Minute,
+			Factor:   2,
+			Steps:    1 << 8, // large enough to make cap the effective bound
+			Cap:      24 * time.Hour,
+		},
+		func() (done bool, returnErr error) {
```
abhinavdahiya, line 145: nit: named returns are not great as they allow empty `returrn` in the function definition which makes it difficult to read what is being returned and when was it set..

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -120,6 +133,66 @@ func completeAWSUninstaller(o *aws.ClusterUninstaller, logLevel string, args []s
...
+			}
+			kubeconfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+				clientcmd.NewDefaultClientConfigLoadingRules(),
+				&clientcmd.ConfigOverrides{},
+			)
+			namespace, _, err := kubeconfig.Namespace()
+			if err != nil {
+				o.logger.WithError(err).Error("could not get the namespace")
+				return
+			}
```
abhinavdahiya, line 168: does this use the context setup using the kubeconfig to get the namespace? How many users would use something like that? _I personally haven't_ So should we also allow for setting namespace directly?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1296,6 +1296,20 @@ func (r *ReconcileClusterDeployment) ensureClusterDeprovisioned(cd *hivev1.Clust
 		cdLog.Info("PreserveOnDelete=true but creating deprovisioning request as cluster was never successfully provisioned")
 	}
 
+	// Stop waiting for deprovision if the abandon-deprovision annotation is true
+	if value, ok := cd.Annotations[constants.AbandonDeprovisionAnnotation]; ok {
+		logger := cdLog.WithField(constants.AbandonDeprovisionAnnotation, value)
+		if abandon, err := strconv.ParseBool(value); abandon && err == nil {
```
abhinavdahiya, line 1302:
```suggestion
		if abandon, err := strconv.ParseBool(value); err == nil && abandon {
```
wouldn't that be more appropriate?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1296,6 +1296,20 @@ func (r *ReconcileClusterDeployment) ensureClusterDeprovisioned(cd *hivev1.Clust
...
+			logger.Warn("adandoning deprovision")
+			err = r.removeClusterDeploymentFinalizer(cd, cdLog)
+			if err != nil {
+				cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error removing finalizer")
```
abhinavdahiya, line 1306: we will return `deprovisioned = true` even when there was an error to remove the finalizer.. shouldn't we return false here?

PR #1018, opened 2020-06-17
restore behavior that disables velero by default

joelddiaz:
...unless explicitly enabled

### DIFF
--- a/pkg/controller/velerobackup/velerobackup_controller.go
+++ b/pkg/controller/velerobackup/velerobackup_controller.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


joelddiaz:
the question here is whether the Hive operator should use the ability to disable controllers (introduced in the PR that broke velero originally)
/hold

dgoodwin:
I don't see any reason why the operator can't configure the running controllers by that method (the CLI args I assume?), however we do have established precedent of configuring via env vars like this, so for consistency I'd probably stick with what you have here and we do everywhere else.

This looks good to me, over to @twiest  for approval though.

joelddiaz:
@twiest PTAL

PR #1019, opened 2020-06-18
Add hive.openshift.io/managed annotation to resources synced to cluster.

dgoodwin:
This affects only two areas, the remotemachineset controller, and
syncsetinstance.

SyncSet now decodes to unstructured data as we cannot decode to
runtime.Object if the type is not registered with our scheme, which is
not guaranteed with RawExtensions in SyncSets.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -114,6 +114,10 @@ const (
 	// SyncsetPauseAnnotation is a annotation used by clusterDeployment, if it's true, then we will disable syncing to a specific cluster
 	SyncsetPauseAnnotation = "hive.openshift.io/syncset-pause"
 
+	// HiveManagedAnnotation is an annotation added to any resources we sync to the remote cluster to help identify that they are
+	// managed by Hive, and any manual changes may be undone the next time the resource is reconciled.
+	HiveManagedAnnotation = "hive.openshift.io/managed"
+
 	// DisableInstallLogPasswordRedactionAnnotation is an annotation used on ClusterDeployments to disable the installmanager
 	// functionality which refuses to print output if it appears to contain a password or sensitive info. This can be
 	// useful in scenarios where debugging is needed and important info is being redacted. Set to "true".
--- a/pkg/controller/remotemachineset/remotemachineset_controller.go
+++ b/pkg/controller/remotemachineset/remotemachineset_controller.go
@@ -340,6 +340,12 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 
 		// Apply hive MachinePool taints to MachineSet MachineSpec.
 		ms.Spec.Template.Spec.Taints = pool.Spec.Taints
+
+		// Add the managed-by-Hive annotation:
+		if ms.Annotations == nil {
+			ms.Annotations = make(map[string]string, 1)
+		}
+		ms.Annotations[constants.HiveManagedAnnotation] = "true"
 	}
 
 	logger.Infof("generated %v worker machine sets", len(generatedMachineSets))
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -45,6 +46,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+	"sigs.k8s.io/yaml"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
@@ -588,7 +590,16 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 
 	// determine if we can gather info for all resources
 	infos := []hiveresource.Info{}
+	unstructureds := make([]*unstructured.Unstructured, len(resources))
 	for i, resource := range resources {
+		u := &unstructured.Unstructured{}
+		if err := yaml.Unmarshal(resource.Raw, u); err != nil {
+			ssiLog.WithError(err).Warn("error decoding unstructured object")
+			return err
+		}
+		unstructureds[i] = u
+
+		// TODO: read required info from unstructured
 		info, err := h.Info(resource.Raw)
 		if err != nil {
 			ssi.Status.Conditions = r.setUnknownObjectSyncCondition(ssi.Status.Conditions, err, i)
@@ -596,6 +607,7 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 			return err
 		}
 		infos = append(infos, *info)
+
 	}
 
 	ssi.Status.Conditions = r.clearUnknownObjectSyncCondition(ssi.Status.Conditions)
@@ -628,8 +640,8 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 		if rss := findSyncStatus(resourceSyncStatus, ssi.Status.Resources); rss == nil || r.needToReapply(resourceSyncStatus, *rss, ssiLog) {
 			// Apply resource
 			ssiLog.Debug("applying resource")
-			var applyResult hiveresource.ApplyResult
-			applyResult, applyErr = applyFn(resource.Raw)
+
+			applyErr = applyResource(unstructureds[i], ssiLog, applyFn)
 
 			var resourceSyncConditions []hivev1.SyncCondition
 			if rss != nil {
@@ -637,11 +649,6 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 			}
 			resourceSyncStatus.Conditions = r.setApplySyncConditions(resourceSyncConditions, applyErr)
 
-			if applyErr != nil {
-				ssiLog.WithError(applyErr).Warn("error applying resource")
-			} else {
-				ssiLog.WithField("applyResult", applyResult).Debug("resource applied")
-			}
 		} else {
 			// Do not apply resource
 			ssiLog.Debug("resource has not changed, will not apply")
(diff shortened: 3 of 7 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -655,6 +649,36 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 	return nil
 }
 
+func applyResource(resource runtime.RawExtension, logger log.FieldLogger, h Applier) error {
```
dgoodwin, line 652: I extracted this function as I did not like the way the error handling was looking otherwise, we went from one possible error to three, all of which need to result in some common handling to store the result in the condition.

gregsheremeta:
cc @jewzaam

test/e2e/postinstall/syncsets/syncsets_suite_test.go
```diff
@@ -125,7 +126,10 @@ var _ = Describe("Test Syncset and SelectorSyncSet func", func() {
 				resultConfigMap = &corev1.ConfigMap{}
 				err = targetClusterClient.Get(ctx, client.ObjectKey{Name: "foo", Namespace: "default"}, resultConfigMap)
 				Ω(err).ShouldNot(HaveOccurred())
-				Ω(resultConfigMap.Data).Should(Equal(map[string]string{"foo": "bar"}))
+				Ω(resultConfigMap.Data["foo"]).Should(Equal("bar"))
```
staebler, line 129: Why change this? Don't we want to ensure that the configmap synced has only the one entry? With this change, the test would still pass if there were other entries besides the "foo" entry.

test/e2e/postinstall/syncsets/syncsets_suite_test.go
```diff
@@ -125,7 +126,10 @@ var _ = Describe("Test Syncset and SelectorSyncSet func", func() {
...
+
+				By("Verify the managed by Hive annotation was injected automatically")
```
staebler, line 131:
```suggestion
				By("Verify the managed-by-Hive annotation was injected automatically")
```

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -428,6 +434,14 @@ func (r *ReconcileRemoteMachineSet) syncMachineSets(
 				resourcemerge.EnsureObjectMeta(&objectMetaModified, &rMS.ObjectMeta, ms.ObjectMeta)
 				msLog := logger.WithField("machineset", rMS.Name)
 
+				// Ensure any annotations we want our our generated machine sets are present. (additive only)
```
joelddiaz, line 437: comment says "our our"

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -616,20 +615,15 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 		if rss := findSyncStatus(resourceSyncStatus, ssi.Status.Resources); rss == nil || r.needToReapply(resourceSyncStatus, *rss, ssiLog) {
 			// Apply resource
 			ssiLog.Debug("applying resource")
-			var applyResult hiveresource.ApplyResult
-			applyResult, applyErr = h.Apply(resource.Raw)
+
+			applyErr = applyResource(resource, ssiLog, h)
```
staebler, line 619: I don't think it is going to be helpful long-term to embed the unmarshaling into the `applyResource` function. We need to do the same unmarshaling right before in order to get the info to fill out the `SyncStatus`.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-
```
joelddiaz, line 31: i thought we wanted to keep our kube imports grouped together

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -428,6 +434,14 @@ func (r *ReconcileRemoteMachineSet) syncMachineSets(
...
+				for k, v := range ms.Annotations {
+					if rMS.Annotations[k] != v {
+						rMS.Annotations[k] = v
+						objectMetaModified = true
+					}
+				}
+
```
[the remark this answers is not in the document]
dgoodwin: Looks like it, missed that, thanks.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
dgoodwin: I have moved this up with the info parsing, this helped simplify the error handling a little and presumably will help if there's a move to get the info from the unstructured in the future.

pkg/controller/utils/utils.go
```diff
@@ -159,3 +161,10 @@ func GetHiveNamespace() string {
 	// Returning a default here, mostly for unit test simplicity and to avoid having to pass this around to all controllers and libraries..
 	return constants.DefaultHiveNamespace
 }
+
+// DecodeUnstructured will decode json bytes and return an Unstructured object.
+func DecodeUnstructured(objBytes []byte) (*unstructured.Unstructured, error) {
+	u := &unstructured.Unstructured{}
+	err := json.Unmarshal(objBytes, &u)
```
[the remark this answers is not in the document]
dgoodwin: In my testing I was definitely using yaml in the raw extension and yet this worked, not sure why. I will change over to yaml unmarshal.

test/e2e/postinstall/syncsets/syncsets_suite_test.go
dgoodwin: It feels overly fragile with too high a cost to uncover. If something were to start adding annotations then this would only fail after an extremely time consuming e2e run. It doesn't seem like a great idea to assume we're the only thing possibly setting annotations even if that's the case for the most part. I'd rather we test what we know we want to be there.

pkg/controller/utils/utils.go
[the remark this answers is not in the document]
staebler: If it works with yaml, that's cool. You can leave it as is.

test/e2e/postinstall/syncsets/syncsets_suite_test.go
staebler: This is not checking the annotations, though. This is checking the `data` field.

dgoodwin: I am just straight up confused, my bad, will address.

pkg/controller/utils/utils.go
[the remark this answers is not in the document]
dgoodwin: I'm a little nervous and decoding as yaml is probably the better option so it's flipped over with next push.

dgoodwin:
PR updated.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -588,14 +590,27 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 
 	// determine if we can gather info for all resources
 	infos := []hiveresource.Info{}
+	unstructureds := []*unstructured.Unstructured{}
 	for i, resource := range resources {
+		unstructureds = append(unstructureds)
```
staebler, line 595: What is this doing?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -588,14 +590,27 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 
 	// determine if we can gather info for all resources
 	infos := []hiveresource.Info{}
+	unstructureds := []*unstructured.Unstructured{}
```
staebler, line 593:
```suggestion
	unstructureds := make([]*unstructured.Unstructured, len(resources))
```

And later, instead of the append,
```
unstructureds[i] = u
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -588,14 +590,27 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
 	for i, resource := range resources {
+		unstructureds = append(unstructureds)
+		// Inject the hive managed annotation to help end-users see that a resource is managed by hive:
+		u := &unstructured.Unstructured{}
+		err := yaml.Unmarshal(resource.Raw, &u)
+		if err != nil {
```
staebler, line 599: You do not need to get the reference to `u`, it is already a reference.
```suggestion
		if err := yaml.Unmarshal(resource.Raw, u); err != nil {
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -667,6 +677,29 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
 
+func applyResource(u *unstructured.Unstructured, logger log.FieldLogger, applyFn func(obj []byte) (hiveresource.ApplyResult, error)) error {
+	annotations := u.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[constants.HiveManagedAnnotation] = "true"
+	u.SetAnnotations(annotations)
+
+	bytes, err := u.MarshalJSON()
```
staebler, line 688:
```suggestion
	bytes, err := json.Marshal(u)
```

pkg/constants/constants.go
```diff
@@ -114,6 +114,10 @@ const (
 	// SyncsetPauseAnnotation is a annotation used by clusterDeployment, if it's true, then we will disable syncing to a specific cluster
 	SyncsetPauseAnnotation = "hive.openshift.io/syncset-pause"
 
+	// HiveManagedAnnotation is added to any resources we sync to the remote cluster to help identify that they are
```
akhil-rane, line 117: nit: I feel 'HiveManagedAnnotation is an annotation added' will be more consistent

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -340,6 +340,12 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 
 		// Apply hive MachinePool taints to MachineSet MachineSpec.
 		ms.Spec.Template.Spec.Taints = pool.Spec.Taints
+
+		// Add the managed by Hive annotation:
```
akhil-rane, line 344: May be change to 'managed-by-Hive' here?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -588,14 +590,27 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 
 	// determine if we can gather info for all resources
 	infos := []hiveresource.Info{}
+	unstructureds := []*unstructured.Unstructured{}
 	for i, resource := range resources {
+		unstructureds = append(unstructureds)
+		// Inject the hive managed annotation to help end-users see that a resource is managed by hive:
```
akhil-rane, line 596: This comment looks out of place here. Is it more appropriate to put it at line 685 in `applyResource` method?

dgoodwin: :+1: thanks

dgoodwin:
PR updated again.

Apologies I missed the secrets.

Do we think patched resources should be flagged as Hive managed? It's a good question. It may be technically quite a bit more complicated though.

akhil-rane:
> PR updated again.
> 
> Apologies I missed the secrets.
> 
> Do we think patched resources should be flagged as Hive managed? It's a good question. It may be technically quite a bit more complicated though.

I think SRE has requested to add annotation to a patched resource too.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -340,6 +340,12 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 
 		// Apply hive MachinePool taints to MachineSet MachineSpec.
 		ms.Spec.Template.Spec.Taints = pool.Spec.Taints
+
+		// Add the managed-by-Hive annotation:
+		if ms.Annotations == nil {
+			ms.Annotations = map[string]string{}
```
staebler, line 346:
```suggestion
			ms.Annotations = make(map[string]string, 1)
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -667,6 +674,30 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 	return nil
 }
 
+func applyResource(u *unstructured.Unstructured, logger log.FieldLogger, applyFn func(obj []byte) (hiveresource.ApplyResult, error)) error {
+	annotations := u.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
```
staebler, line 680:
```suggestion
		annotations = make(map[string]string, 1)
```

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -830,6 +861,12 @@ func (r *ReconcileSyncSetInstance) applySyncSetSecretMappings(ssi *hivev1.SyncSe
 		secret.UID = ""
 		secret.OwnerReferences = nil
 
+		// Inject our managed-by-Hive annotation:
+		if secret.Annotations == nil {
+			secret.Annotations = map[string]string{}
```
staebler, line 866:
```suggestion
			secret.Annotations = make(map[string]string, 1)
```

dgoodwin:
Updated, patching definitely a good candidate for another PR, I have no idea how we're going to do that one yet.

dgoodwin:
Dumb mistake in the e2e test, checking for the annotations in configmap data instad of configmap annotations.

Feedback addressed, new card filed for patched resources.

PR #1022, opened 2020-06-22
Fix unable to install CRD and CR in same syncset

akhil-rane:
Currently, we first gather info about all the resources before creating them.This way 
we cannot create a CRD and CR in the same syncet, because to fetch the info about
the CR we need to first create the CRD. This PR fixes it by fetching info and creating 
one resource at a time.

Jira: https://issues.redhat.com/browse/CO-540

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -48,7 +48,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-	"sigs.k8s.io/yaml"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
@@ -635,29 +634,6 @@ func (r *ReconcileSyncSetInstance) reapplyTime(status hivev1.SyncStatus) time.Ti
 func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInstance, resources []runtime.RawExtension, applyBehavior hivev1.SyncSetApplyBehavior, dynamicClient dynamic.Interface, h Applier, ssiLog log.FieldLogger) error {
 	ssiLog = ssiLog.WithField("applyTerm", "resource")
 
-	// determine if we can gather info for all resources
-	infos := []hiveresource.Info{}
-	unstructureds := make([]*unstructured.Unstructured, len(resources))
-	for i, resource := range resources {
-		u := &unstructured.Unstructured{}
-		if err := yaml.Unmarshal(resource.Raw, u); err != nil {
-			ssiLog.WithError(err).Warn("error decoding unstructured object")
-			return err
-		}
-		unstructureds[i] = u
-
-		// TODO: read required info from unstructured
-		info, err := h.Info(resource.Raw)
-		if err != nil {
-			ssi.Status.Conditions = r.setUnknownObjectSyncCondition(ssi.Status.Conditions, err, i)
-			ssiLog.WithError(err).Warn("unable to parse resource")
-			return err
-		}
-		infos = append(infos, *info)
-
-	}
-
-	ssi.Status.Conditions = r.clearUnknownObjectSyncCondition(ssi.Status.Conditions)
 	syncStatusList := []hivev1.SyncStatus{}
 
 	applyFn := h.Apply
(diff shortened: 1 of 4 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -656,6 +650,7 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 		}
 	}
 
+	ssi.Status.Conditions = r.clearUnknownObjectSyncCondition(ssi.Status.Conditions)
```
akhil-rane, line 653: There can be a situation wherein we are able to get information about the resource but are not able to apply it. In this case UnknownObjectCondition won't be cleared immediately. I do not know how to handle that yet.

staebler: I think that the UnknownObjectCondition should be moved from a condition on the entire SyncSetInstance to a condition on each of the individual resources in the SyncSetInstance. In other words, the condition should be set on `.status.resources[i].conditions` rather than on `.status.conditions`.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -586,20 +586,8 @@ func (r *ReconcileSyncSetInstance) reapplyTime(status hivev1.SyncStatus) time.Ti
...
-			ssi.Status.Conditions = r.setUnknownObjectSyncCondition(ssi.Status.Conditions, err, i)
-			ssiLog.WithError(err).Warn("unable to parse resource")
-			return err
-		}
-		infos = append(infos, *info)
-	}
-
-	ssi.Status.Conditions = r.clearUnknownObjectSyncCondition(ssi.Status.Conditions)
 	syncStatusList := []hivev1.SyncStatus{}
+	var applyErr error
```
staebler, line 590: Why was this moved?

akhil-rane: I have just moved it outside the loop

akhil-rane:
> I still think that we should switch from using `h.Info` to unmarshaling directly into an `Unstructured`. The discovery is unnecessary. You could build on top of #1019, which is already doing the direct unmarshaling.

We cannot fetch Resource type from the unstructured. Do we need it SyncStatus?

staebler:
> > I still think that we should switch from using `h.Info` to unmarshaling directly into an `Unstructured`. The discovery is unnecessary. You could build on top of #1019, which is already doing the direct unmarshaling.
> 
> We cannot fetch Resource type from the unstructured. Do we need it SyncStatus?

That makes sense. We don't **need** it in the status, but we do need it perform a Delete.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
staebler: It was not inside the loop before. It was declared right before the loop.

akhil-rane:
Updated the PR

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -608,15 +597,22 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
 	}
-
+	// determine if we can gather info for the resource and apply it
 	var applyErr error
-	for i, resource := range resources {
+	var discoveryErr error
+	info := &hiveresource.Info{}
+	for _, resource := range resources {
+		u := &unstructured.Unstructured{}
+		if err := yaml.Unmarshal(resource.Raw, u); err != nil {
```
staebler, line 606: I think we had a misundestanding about where we landed on this. Since we need to do the `h.Info` call anyway in order to get the Resource of the object, there is no point in doing the yaml unmarshaling.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -625,23 +621,30 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 			WithField("syncAPIVersion", resourceSyncStatus.APIVersion).
 			WithField("syncKind", resourceSyncStatus.Kind)
 
+		info, discoveryErr = h.Info(resource.Raw)
+		// If an error occurred, stop processing right here
+		if discoveryErr != nil {
+			resourceSyncStatus.Conditions = r.setUnknownObjectSyncCondition(resourceSyncStatus.Conditions, discoveryErr)
```
staebler, line 627: We should check with our users to make sure that nobody is reading the UnknownObjectSync condition from its current location.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
akhil-rane: yaml unmarshalling was picked up from Devan's PR. I think we are good to use h.Info object here, @dgoodwin Is there any specific reason you want to have unmarshalling? Can we go ahead and remove it?

staebler: We can probably forego the explicit yaml unmarshaling and instead return in the `Info` the `Unstructured` object built when the `Info` object is created.

dgoodwin: Sounds good to me.

akhil-rane: @staebler I do not see any `Unstructured` object getting built when the `Info` object gets created in Info.go. Am I missing something here?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
akhil-rane: @igoihman We are now putting the `UnknownObjectSync` condition on each individual resource instead of the syncset. Are we ok with it from your end?

pkg/controller/syncsetinstance/syncsetinstance_controller.go
staebler: I was thinking something along the lines of the following.
https://github.com/staebler/hive/commit/600fd9a568191d6fd8643496108f60857c7aaffb

akhil-rane: Since we have decided to move UnknownObjectSync condition to individual resource now, doesn't it make sense to have resourceSyncStatus object populated with minimum info before we set any condition on it. I think unmarshaling before calling info will help in here. Does it affect performance wise?

staebler: I don't think that is necessary. Resource *i* in the status will correspond to resource *i* in the spec. The API version, kind, name, and namespace in the resource in the status are there for correlating applied resources to resources in the spec on subsequent reconciles in case the resources in the spec have changed position in the list. In this case, that is not an issue since the resource will not be applied.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
igoihman: we are not checking this condition - it's ok from my end.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -667,28 +643,39 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
-	for i, resource := range resources {
-		resourceSyncStatus := hivev1.SyncStatus{
-			APIVersion: infos[i].APIVersion,
-			Kind:       infos[i].Kind,
-			Resource:   infos[i].Resource,
-			Name:       infos[i].Name,
-			Namespace:  infos[i].Namespace,
-			Hash:       r.hash(resource.Raw),
-		}
+	var discoveryErr error
```
staebler, line 648: We do not need a separate variable for `discoveryErr`. We can use `applyErr` in its place.

PR #1025, opened 2020-06-24
log conflict errors at Info level

joelddiaz:


pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -191,8 +191,10 @@ func (r *ReconcileDNSZone) Reconcile(request reconcile.Request) (reconcile.Resul
 		"lastSyncGeneration": desiredState.Status.LastSyncGeneration,
 	}).Info("Syncing DNS Zone")
 	result, err := r.reconcileDNSProvider(actuator, desiredState)
-	if err != nil {
-		dnsLog.WithError(err).Error("Encountered error while attempting to reconcile")
+	if apierrors.IsConflict(err) {
+		dnsLog.WithError(err).Info("Encountered conflict while attempting to reconcile")
+	} else if err != nil {
+		dnsLog.WithError(err).Info("Encountered error while attempting to reconcile")
 	}
```
staebler, line 198: Assuming that your intention was to log non-confict errors are Error level instead of the Info level that you have,
```suggestion
	if err != nil {
		dnsLog.WithError(err).Log(controllerutils.LogLevel(err), "Encountered error while attempting to reconcile")
	}
```

joelddiaz: Wasn't aware of the controllerutils helper, but the general idea was definitely so that they would be handled differently :)

PR #1026, opened 2020-06-25
allow disabling controllers via HiveConfig

joelddiaz:
Introduce spec.disabledControllers as a list of strings that - when provided - will be used to set the `--disabled-controllers` param when the Hive operator is creating/updating the Hive controllers Deployment.

Also, move the log-level param to also live under spec.containers[].args (instead of spec.containers[].command).

### DIFF
--- a/pkg/apis/hive/v1/hiveconfig_types.go
+++ b/pkg/apis/hive/v1/hiveconfig_types.go
@@ -74,6 +74,10 @@ type HiveConfigSpec struct {
 	// +kubebuilder:validation:Enum=enabled
 	// +optional
 	DeleteProtection DeleteProtectionType `json:"deleteProtection,omitempty"`
+
+	// DisabledControllers allows selectively disabling Hive controllers by name.
+	// The name of an individual controller matches the name of the controller as seen in the Hive logging output.
+	DisabledControllers []string `json:"disabledControllers,omitempty"`
 }
 
 // HiveConfigStatus defines the observed state of Hive
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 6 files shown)


staebler:
> /test verify

Need to run `make update-bindata`.

staebler:
> > /test verify
> 
> Need to run `make update-bindata`.

Sorry. This is the deepcopy not the bindata.

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -74,6 +74,10 @@ type HiveConfigSpec struct {
 	// +kubebuilder:validation:Enum=enabled
 	// +optional
 	DeleteProtection DeleteProtectionType `json:"deleteProtection,omitempty"`
+
+	// DisabledControllers allows selectively disabling Hive controllers by name.
+	// The names of the individual controllers matches the name of the controller as seen in the Hive logging output.
```
staebler, line 79:
```suggestion
	// The name of an individual controller matches the name of the controller as seen in the Hive logging output.
```

pkg/operator/hive/hive.go
```diff
@@ -70,12 +71,13 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
 		)
 	}
 
+	disabledControllersFlag := assembleDisabledControllersFlag(instance.Spec.DisabledControllers)
+	if disabledControllersFlag != "" {
+		hiveContainer.Args = append(hiveContainer.Args, disabledControllersFlag)
+	}
```
staebler, line 77: Does this need a specific function?
```suggestion
	if dc := instance.Spec.DisabledControllers; len(dc) != 0 {
		hiveContainer.Args = append(hiveContainer.Args, "--disabled-controllers", strings.Join(dc, ","))
	}
```

joelddiaz: Nope. It was more complex when I took a first pass at it, and then the function just stuck around. I'll condense it down.

PR #1027, opened 2020-06-29
Change hive.openshift.io/managed from annotation to label.

dgoodwin:
This will allow using a selector in admission webhooks interested in
protecting these resources.

### DIFF
--- a/pkg/controller/remotemachineset/remotemachineset_controller.go
+++ b/pkg/controller/remotemachineset/remotemachineset_controller.go
@@ -327,9 +327,11 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 		}
 
 		if ms.Labels == nil {
-			ms.Labels = map[string]string{}
+			ms.Labels = make(map[string]string, 2)
 		}
 		ms.Labels[machinePoolNameLabel] = pool.Spec.Name
+		// Add the managed-by-Hive label:
+		ms.Labels[constants.HiveManagedLabel] = "true"
 
 		// Apply hive MachinePool labels to MachineSet MachineSpec.
 		ms.Spec.Template.Spec.ObjectMeta.Labels = make(map[string]string, len(pool.Spec.Labels))
@@ -339,12 +341,6 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 
 		// Apply hive MachinePool taints to MachineSet MachineSpec.
 		ms.Spec.Template.Spec.Taints = pool.Spec.Taints
-
-		// Add the managed-by-Hive annotation:
-		if ms.Annotations == nil {
-			ms.Annotations = make(map[string]string, 1)
-		}
-		ms.Annotations[constants.HiveManagedAnnotation] = "true"
 	}
 
 	logger.Infof("generated %v worker machine sets", len(generatedMachineSets))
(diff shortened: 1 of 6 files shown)


pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -341,10 +341,10 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 		ms.Spec.Template.Spec.Taints = pool.Spec.Taints
 
 		// Add the managed-by-Hive annotation:
-		if ms.Annotations == nil {
-			ms.Annotations = make(map[string]string, 1)
+		if ms.Labels == nil {
```
staebler, line 344: This check was already done a few lines back. It does not need to be done again.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -841,7 +839,7 @@ func testMachineSet(name string, machineType string, unstompedAnnotation bool, r
 	}
 	// Add a pre-existing annotation which we will ensure remains in updated machinesets.
 	if unstompedAnnotation {
```
staebler, line 841: This variable name needs to be adjusted. But this verification is probably not need for machinesets since the tests are already adding other labels.

dgoodwin: I actually screwed this up, this was supposed to stay as an Annotation test, it predates my work I just changed how it was initialized. Will fix up.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
```diff
@@ -682,7 +682,7 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 								t.Errorf("machineset %v has unexpected labels:\nexpected: %v\nactual: %v", eMS.Name, eMS.Labels, rMS.Labels)
 							}
 							if !reflect.DeepEqual(eMS.ObjectMeta.Annotations, rMS.ObjectMeta.Annotations) {
-								t.Errorf("machineset %v has unexpected annotations:\nexpected: %v\nactual: %v", eMS.Name, eMS.Labels, rMS.Labels)
+								t.Errorf("machineset %v has unexpected annotations:\nexpected: %v\nactual: %v", eMS.Name, eMS.Annotations, rMS.Annotations)
```
dgoodwin, line 685: Looks like the test had a bug prior to this.

pkg/controller/remotemachineset/remotemachineset_controller_test.go
dgoodwin: Restored to it's original form.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -327,9 +327,11 @@ func (r *ReconcileRemoteMachineSet) generateMachineSets(
 		}
 
 		if ms.Labels == nil {
-			ms.Labels = map[string]string{}
+			ms.Labels = make(map[string]string, 1)
```
staebler, line 330:
```suggestion
			ms.Labels = make(map[string]string, 2)
```

dgoodwin:
All set.

PR #1029, opened 2020-06-29
Add disabledControllers to hack template

jewzaam:


gregsheremeta:
looks good to me. Since Joel wrote this, I prefer he looks too.
/cc @joelddiaz

PR #1030, opened 2020-06-30
Add install log regex for unable to find matching route table.

abutcher:
https://bugzilla.redhat.com/show_bug.cgi?id=1844320

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -5,7 +5,13 @@ metadata:
   namespace: hive
 data:
   regexes: |
-    # AWS Specific:
+    # AWS Specific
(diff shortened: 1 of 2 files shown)


config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -5,7 +5,13 @@ metadata:
   namespace: hive
 data:
   regexes: |
-    # AWS Specific:
+    # AWS Specific
+    # https://bugzilla.redhat.com/show_bug.cgi?id=1844320
+    - name: AWSUnableToFindMatchingRouteTable
+      searchRegexStrings:
+      - "Error: Unable to find matching route for Route Table.*CIDR block (0.0.0.0/0)"
```
dgoodwin, line 12: Those braces around subnet mask look like they would cause problems, should we just go with "Error: Unable to find matching route for Route Table"

PR #1031, opened 2020-07-01
add uninstall job labels to the uninstall pod as well

gregsheremeta:


gregsheremeta:
I didn't see a test for the install pod labels, so I'm thinking this doesn't rise to the level of "needs a test." Let me know if that's incorrect.

PR #1032, opened 2020-07-02
Set ClusterDeployment status when DNS Zone cannot be created due to missing permissions

akhil-rane:
This PR sets the ClusterDeployment status with a more informative reason and
message when a DNS zone cannot be created due to AWS credentials lacking correct 
permissions.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1101,20 +1101,14 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 	}
 }
 
-func (r *ReconcileClusterDeployment) setDNSNotReadyCondition(cd *hivev1.ClusterDeployment, isReady bool, message string, cdLog log.FieldLogger) error {
-	status := corev1.ConditionFalse
-	reason := dnsReadyReason
-	if !isReady {
-		status = corev1.ConditionTrue
-		reason = dnsNotReadyReason
-	}
+func (r *ReconcileClusterDeployment) setDNSNotReadyCondition(cd *hivev1.ClusterDeployment, status corev1.ConditionStatus, reason string, message string, cdLog log.FieldLogger) error {
 	conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
 		cd.Status.Conditions,
 		hivev1.DNSNotReadyCondition,
 		status,
 		reason,
 		message,
-		controllerutils.UpdateConditionNever)
+		controllerutils.UpdateConditionIfReasonOrMessageChange)
 	if !changed {
 		return nil
 	}
@@ -1447,7 +1441,7 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 	case p.Azure != nil:
 	default:
 		cdLog.Error("cluster deployment platform does not support managed DNS")
-		if err := r.setDNSNotReadyCondition(cd, false, "Managed DNS is not supported for platform", cdLog); err != nil {
+		if err := r.setDNSNotReadyCondition(cd, corev1.ConditionTrue, dnsNotReadyReason, "Managed DNS is not supported for platform", cdLog); err != nil {
 			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not update DNSNotReadyCondition")
 			return nil, err
 		}
@@ -1469,7 +1463,7 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 
 	if !metav1.IsControlledBy(dnsZone, cd) {
 		cdLog.Error("DNS zone already exists but is not owned by cluster deployment")
-		if err := r.setDNSNotReadyCondition(cd, false, "Existing DNS zone not owned by cluster deployment", cdLog); err != nil {
+		if err := r.setDNSNotReadyCondition(cd, corev1.ConditionTrue, dnsNotReadyReason, "Existing DNS zone not owned by cluster deployment", cdLog); err != nil {
 			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not update DNSNotReadyCondition")
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -50,6 +51,9 @@ var (
 	},
 		[]string{"force"},
 	)
+
+	// reg is a regex used to fetch condition message from error when credentials are missing permission to create DNS zone
+	reg = regexp.MustCompile(`^AccessDeniedException:\s+([^\t]+)\n\t`)
```
joelddiaz, line 56: For a package-wide constant/var can we give it a more descriptive variable name? Maybe `accessDeniedMessageRegexp`?

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -215,10 +219,48 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
...
+				// formatting error message before adding it to condition
+				// sample error message: AccessDeniedException: User: arn:aws:iam::0123456789:user/testAdmin is not authorized to perform: tag:GetResources with an explicit deny\n\tstatus code: 400, request id: d3b1e18b-52af-4836-9610-45e017476a9c
+				// message after formatting: User: arn:aws:iam::0123456789:user/testAdmin is not authorized to perform: tag:GetResources with an explicit deny
+				conditionMessage = submatches[1]
+			}
+			conds, changed := controllerutils.SetDNSZoneConditionWithChangeCheck(
+				dnsZone.Status.Conditions,
+				hivev1.InvalidCredentialsCondition,
+				corev1.ConditionTrue,
+				"AccessDenied",
```
joelddiaz, line 234: Maybe constify this string.

akhil-rane:
@joelddiaz Updated the PR

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -41,6 +42,7 @@ const (
 	dnsClientTimeout                = 30 * time.Second
 	resolverConfigFile              = "/etc/resolv.conf"
 	zoneCheckDNSServersEnvVar       = "ZONE_CHECK_DNS_SERVERS"
+	accessDeniedMessage             = "AccessDenied"
```
joelddiaz, line 45:
```suggestion
	accessDeniedReason             = "AccessDenied"
```

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -215,10 +220,48 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
...
+		}
 		r.logger.WithError(err).Error("Failed to retrieve hosted zone and corresponding tags")
 		return reconcile.Result{}, err
 	}
 
+	conds, changed := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		dnsZone.Status.Conditions,
+		hivev1.InvalidCredentialsCondition,
+		corev1.ConditionFalse,
+		"AccessDenied",
```
joelddiaz, line 254: use the accessDenied constant here

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1744,6 +1776,17 @@ func assertConditionStatus(t *testing.T, cd *hivev1.ClusterDeployment, condType
 	assert.True(t, found, "did not find expected condition type: %v", condType)
 }
 
+func assertConditionReason(t *testing.T, cd *hivev1.ClusterDeployment, condType hivev1.ClusterDeploymentConditionType, reason string) {
+	found := false
+	for _, cond := range cd.Status.Conditions {
+		if cond.Type == condType {
+			found = true
```
staebler, line 1783: Return when the condition is found rather than having the `found` variable.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -215,10 +220,48 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
 	r.logger.Debug("Retrieving current state")
 	err := actuator.Refresh()
 	if err != nil {
+		if strings.Contains(err.Error(), "AccessDeniedException") {
```
staebler, line 223: Can we check the error code instead of looking through the error string for "AccessDeniedException"? Cast the error to an `awserr.Error` and see what is in there. You may also be able to get the message that you want directly from the `awserr.Error` rather than doing regular expressions, too.

pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -141,6 +141,18 @@ func TestReconcileDNSProviderForAWS(t *testing.T) {
...
 			},
 		},
+		{
+			name:    "Cannot create a hosted zone as credentials are missing permissions",
+			dnsZone: validDNSZoneWithoutID(),
+			setupAWSMock: func(expect *mock.MockClientMockRecorder) {
+				mockAccessDeniedException(expect, validDNSZoneWithoutID())
+			},
+			validateZone: func(t *testing.T, zone *hivev1.DNSZone) {
+				condition := controllerutils.FindDNSZoneCondition(zone.Status.Conditions, hivev1.InvalidCredentialsCondition)
```
staebler, line 151: Can you test also that the condition has the message that you expect?

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -215,10 +220,48 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
...
+				dnsZone.Status.Conditions,
+				hivev1.InvalidCredentialsCondition,
+				corev1.ConditionTrue,
+				accessDeniedReason,
+				conditionMessage,
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				dnsZone.Status.Conditions = conds
+				if err := r.Status().Update(context.Background(), dnsZone); err != nil {
```
staebler, line 241: Should return here when the update is successful, too. Otherwise, the function will add the "Failed to retrieve hosted zone and corresponding tags" error log entry. It would also be good to add a warning log entry when the condition is set.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -50,6 +52,9 @@ var (
 	},
 		[]string{"force"},
 	)
+
+	// accessDeniedMessageRegexp is a regex used to fetch condition message from error when credentials are missing permission to create DNS zone
+	accessDeniedMessageRegexp = regexp.MustCompile(`^AccessDeniedException:\s+([^\t]+)\n\t`)
```
staebler, line 57: What do we gain from restricting the substring to non-tabs? I would expect one of the two following.
```suggestion
	accessDeniedMessageRegexp = regexp.MustCompile(`^AccessDeniedException:\s+(.+)\n\t`)
```
```suggestion
	accessDeniedMessageRegexp = regexp.MustCompile(`^AccessDeniedException:\s+([^\t]+)`)
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1469,30 +1463,42 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
...
+	if err := r.setDNSNotReadyCondition(cd, status, reason, message, cdLog); err != nil {
 		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not update DNSNotReadyCondition")
 		return nil, err
 	}
-	return dnsZone, nil
+
+	if reason == dnsReadyReason {
+		return dnsZone, nil
+	}
+	return nil, nil
```
staebler, line 1501: This is a nit, but I would rather the return at the end of the function be the happy scenario.
```suggestion
	if reason != dnsReadyReason {
		return nil, nil
	}
	return dnsZone, nil
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1744,6 +1776,16 @@ func assertConditionStatus(t *testing.T, cd *hivev1.ClusterDeployment, condType
...
 }
 
+func assertConditionReason(t *testing.T, cd *hivev1.ClusterDeployment, condType hivev1.ClusterDeploymentConditionType, reason string) {
+	for _, cond := range cd.Status.Conditions {
+		if cond.Type == condType {
+			assert.Equal(t, reason, cond.Reason, "condition found with unexpected reason")
+			return
+		}
+	}
+	assert.True(t, false, "did not find expected condition type: %v", condType)
```
staebler, line 1786:
```suggestion
	t.Errorf("did not find expected condition type: %v", condType)
```

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -215,10 +217,43 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
 	r.logger.Debug("Retrieving current state")
 	err := actuator.Refresh()
 	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
```
staebler, line 220: We also need to check the code on the error. Not every `awserr.Error` is going to be caused by insufficient creds.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -215,10 +217,43 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
...
+				awsErr.Message(),
+				controllerutils.UpdateConditionIfReasonOrMessageChange,
+			)
+			if changed {
+				dnsZone.Status.Conditions = conds
+				if err := r.Status().Update(context.Background(), dnsZone); err != nil {
+					return reconcile.Result{}, err
+				}
+				r.logger.WithError(err).Warn("Credentials are lacking permissions")
+				return reconcile.Result{}, err
```
staebler, line 235: This return should be outside of the `if changed` block.

PR #1035, opened 2020-07-08
Design for hibernating clusters with Hive

csrwng:


docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
+# Hibernating Clusters
+
+## Overview
+
+Starting with OpenShift version 4.4.8, clusters can be stopped and started by simply shutting
+down their machines and starting them back up. The only additional requirement to bring a cluster
+back up from a stopped state is that CSRs be approved for the cluster's nodes when certificates
+have expired since the cluster was active.
```
dgoodwin, line 8: "This most commonly happens if the cluster is hibernated in the first 24 hours before initial cert rotation takes place. After this phase, certs will be valid for 30 days."

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+const (
+  // ... other constants
+
+  // ClusterStartingConditionType is set when the ClusterDeployment is transitioning
+  // from a hibernating state to a running state.
+  ClusterStartingConditionType ClusterDeploymentConditionType = "ClusterStarting"
+
+  // ClusterRunningConditionType is set when the ClusterDeployment is in a running and
+  // ready state.
+  ClusterRunningConditionType ClusterDeploymentConditionType = "ClusterRunning"
```
dgoodwin, line 40: Will this be true for clusters that have never hibernated?

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Selecting Cluster Machines
+
+Option 1 (Preferred):
+The hibernation controller relies on the actuator to select machines used by the cluster.
+The actuator, given a ClusterDeployment's InfraID selects machines using a method appropriate to the cloud provider (tags/name prefix/resource group).
+
+Option 2:
+The hibernation controller uses the machine API on the target cluster to determine which machines belong to the cluster. It then stores the machine IDs
+in the clusterdeployment (or a separate CR), then uses those machine IDs to start the cluster again.
```
dgoodwin, line 78: Would a configmap be preferable here? We just want to store some data, I don't know as we necessary need API level commitment here.

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+  // from a running state to a hibernating state.
+  ClusterStoppingConditionType ClusterDeploymentConditionType = "ClusterStopping"
+
+  // ClusterHibernatingConditionType is set when the ClusterDeployment has transitioned
+  // to a hibernating state.
+  ClusterHibernatingConditionType ClusterDeploymentConditionType = "ClusterHibernating"
+
+  // HibernationNotSupportedConditionType is set when a ClusterDeployment's hibernating
+  // property is set to true, but the cluster has an OpenShift version that doesn't support hibernating
+  // (< 4.4.8).
```
dgoodwin, line 52: Should we use this condition if you try to hibernate on a provider we don't have an actuator for yet?

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+continue to check the version in case the cluster is upgraded and eventually is able to be hibernated.
+
+#### Approving CSRs
+In the case that CSRs must be approved for a cluster that has had its certificates expired while hibernating,
+we should follow similar checks as the [cluster machine approver](https://github.com/openshift/cluster-machine-approver/blob/0f50c7bfe9b309ce01937274598f5a807d9545df/csr_check.go)
+to ensure we are not introducing an additional security exposure.
+
+### Changes to Other Controllers
+
+The following controllers need to be modified to skip processing a cluster in a hibernating state:
```
dgoodwin, line 115: The unreachable condition should already disable all of these (if not that's a bug), I think we can safely just let unreachable do it's job here.

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+#### Handling Incompatible OpenShift Versions
+OpenShift versions earlier than 4.4.8 do not support stopping and starting a cluster without additional work
+to restore etcd. In the case that the cluster deployment's `status.clusterVersionStatus.desired.version` is 
+less than 4.4.8, the hibernation controller will add a condition of `HibernationNotSupported` to the cluster
+deployment. When later reconciling cluster deployments with this condition set, the hibernation controller will
+continue to check the version in case the cluster is upgraded and eventually is able to be hibernated.
+
+#### Approving CSRs
+In the case that CSRs must be approved for a cluster that has had its certificates expired while hibernating,
+we should follow similar checks as the [cluster machine approver](https://github.com/openshift/cluster-machine-approver/blob/0f50c7bfe9b309ce01937274598f5a807d9545df/csr_check.go)
```
dgoodwin, line 110: Looks like a fairly small isolated bit of code, would we vendor or copy?

dgoodwin:
Something else to consider, we do the "delete-after" annotation, something similar would be great to have Hive hibernate it at a certain time. "after" doesn't work in this case it would be more like "at 5pm every day"? Or should this just be a separate tool?

I want to link this up with ClusterPool once it's ready so you can specify to hibernate all clusters in the pool until they're requested.

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+
+The ClusterDeploymentSpec should allow setting whether machines are in a running state or in
+a hibernating state.
+
+```go
+type ClusterDeploymentSpec struct {
+  // ... other fields
+
+  // Hibernating indicates whether a cluster should be hibernating. Default is false.
+  Hibernating  `json:"hibernating,omitempty"`
```
staebler, line 24: I assume that this is missing a `bool`. I think we should make this an enum string so that we have the ability to support more modes in the future.

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+
+On the status side, a cluster's state is reflected by the following condition types:
+
+```go
+const (
+  // ... other constants
+
+  // ClusterStartingConditionType is set when the ClusterDeployment is transitioning
+  // from a hibernating state to a running state.
+  ClusterStartingConditionType ClusterDeploymentConditionType = "ClusterStarting"
```
staebler, line 36: "Starting" seems a bit ambiguous. When I first read this, it made me think of starting initially when the cluster is created. How about something like "Resuming" or "Waking"?

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+
+  // ClusterRunningConditionType is set when the ClusterDeployment is in a running and
+  // ready state.
+  ClusterRunningConditionType ClusterDeploymentConditionType = "ClusterRunning"
+
+  // ClusterStoppingConditionType is set when the ClusterDeployment is transitioning
+  // from a running state to a hibernating state.
+  ClusterStoppingConditionType ClusterDeploymentConditionType = "ClusterStopping"
```
staebler, line 44: This seems like a lot of conditions to add to the ClusterDeployment. Can we accomplish the same thing with reasons on a single "ClusterHibernating" condition.

Original | Status | Reason       
--- | --- | ---
ClusterStarting | True | Resuming
ClusterRunning | False | Running
ClusterStopping | True | Stopping
ClusterHibernating | True | Hibernating
HibernationNotSupported | False | NotSupported

docs/hibernating-clusters.md
staebler: :+1:

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Controller Changes
+
+### Cluster Hibernation Controller
+The cluster hibernation controller is a new controller that will watch ClusterDeployments and ensure that the target cluster's
+machines reflect the hibernating state specified in the ClusterDeployment spec.
+
+High level flow of controller logic:
+
+![hibernation_controller](hibernation_controller.png)
```
staebler, line 65: Once a cluster is hibernating, should the controller repeatedly verify that the machines are still stopped? In other words, if a user manually starts a machine while the cluster is hibernating, should Hive override that and stop the machine again?

docs/hibernating-clusters.md
csrwng: Yes, but if not present, we could assume that the cluster is running since that's the default state.

docs/hibernating-clusters.md
csrwng: Ack a configmap would work. However, would you prefer this option to using tags/name ?

docs/hibernating-clusters.md
csrwng: Definitely

docs/hibernating-clusters.md
csrwng: Cool, so no change to existing controllers. Is there anything we should do when the cluster is back up running?

docs/hibernating-clusters.md
csrwng: We need to copy what we can, I believe we will need to modify it, because this code is running on the cluster when the CSRs come up, but yet you still have to manually approve them. So something in the validation must be different.

csrwng:
> Something else to consider, we do the "delete-after" annotation, something similar would be great to have Hive hibernate it at a certain time. "after" doesn't work in this case it would be more like "at 5pm every day"? Or should this just be a separate tool?

I would propose we handle that in a separate tool/controller that is separate from the scope of this controller.

docs/hibernating-clusters.md
csrwng: yes, it is a missing bool. I thought about the enum, but struggled to think about other modes that would make sense. Not opposed to changing it though.

docs/hibernating-clusters.md
csrwng: I like Resuming

docs/hibernating-clusters.md
csrwng: I like it. I was torn with the multiple conditions, but still liked the benefit of a condition such as knowing when it last transitioned. This looks better.

dgoodwin: ClusterRunning could be used for a "HibernateAfter" feature so we could always power down after 8h which might be useful for a lot of use cases. If it's just one condition though it's still possible as that would be the current condition.

docs/hibernating-clusters.md
dgoodwin: Hmm yeah it could take a few hours before unreachable clears, maybe that is a step that needs to happen on resume.

docs/hibernating-clusters.md
dgoodwin: Option 1 seems more resilient to me.

docs/hibernating-clusters.md
staebler: The unreachable controller will check on every change to the ClusterDeployment. So if there was a condition change to indicate that the cluster is now running, then the unreachable controller will check soon after that.

docs/hibernating-clusters.md
staebler: I think the general guidance is to avoid bool because it locks you in to binary state. However, I cannot think of any other modes either.

csrwng:
@dgoodwin @staebler thanks for your comments so far, I've now updated the doc.
Naming is hard, so if you have better suggestions for API variable names/types, they're very welcome :)

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,146 @@
...
+type ClusterRunningState string
+
+const (
+  // RunningClusterRunningState is the default state of a cluster after it has
+  // been installed. All of its machines should be running.
+  RunningClusterRunningState string = "Running"
+
+  // HibernatingClusterRunningState is used to stop the machines belonging to a cluster
+  // and move it to a running state.
+  HibernatingClusterRunningState string = "Hibernating"
```
akhil-rane, line 31: Can we name var `ClusterRunningState` as `ClusterState` and status strings as `ClusterRunningState` & `ClusterHibernatingState`?

docs/hibernating-clusters.md
```diff
@@ -0,0 +1,146 @@
...
+we should follow similar checks as the [cluster machine approver](https://github.com/openshift/cluster-machine-approver/blob/0f50c7bfe9b309ce01937274598f5a807d9545df/csr_check.go)
+to ensure we are not introducing an additional security exposure.
+
+#### Resuming from a Hibernating State
+When a cluster is hibernated, the unreachable controller should properly set the unreachable condition on
+the cluster once it stops responding. This will cause other controllers like the remotemachineset controller to
+stop trying to reconcile the cluster. However, when having resumed a cluster, the hibernation controller should
+remove the unreachable condition if set so that other controllers can start processing the cluster as well. The
+reason is that the unreachable controller could take up to 2 hours to remove the unreachable condition from the
+cluster deployment on its own.
```
akhil-rane, line 146: I think unreachable controller has a connectivity recheck delay of 2 hours when cluster is reachable. When cluster is unreachable we do connectivity recheck more frequently. So may be hibernation controller need not delete the unreachable condition, unreachable controller will do it.

csrwng: Thanks, now updated to indicate that the unreachable controller should handle everything.

csrwng:
Updated to use `PowerState` as the field in the ClusterDeployment that says whether a cluster is running or hibernating.

gregsheremeta:
good to merge this, since we started coding?

csrwng:
@gregsheremeta updated the actuator interface to reflect reality. Other than that, it should be good to go.

PR #1036, opened 2020-07-08
Add time to apply metrics for syncsets and selectorsyncsets

suhanime:
Adds metrics `hive_syncset_apply_duration_seconds` and `hive_selectorsyncset_apply_duration_seconds`

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -27,6 +27,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
@@ -44,6 +45,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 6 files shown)


pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -302,6 +325,25 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
...
+			startTime = cd.Status.InstalledTimestamp.Time
+		}
+		if existingCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions,
+			hivev1.UnreachableCondition); existingCondition != nil {
+			if existingCondition.LastTransitionTime.Time.After(startTime) {
+				startTime = existingCondition.LastTransitionTime.Time
+			}
+		}
+		applyTime := ssi.Status.FirstSuccessTimestamp.Sub(startTime).Seconds()
+		metricTimeToApplySyncset.Observe(float64(applyTime))
```
staebler, line 342: I think that we want to observe the metric **after** updating the status on the SyncSetInstance. We do not want to observe multiple times for the same SyncSetInstance if there was an error doing the status update.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -71,8 +73,29 @@ const (
...
+	metricTimeToApplySyncset = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "hive_syncset_apply_duration_seconds",
+			Help:    "Time to apply syncset",
+			Buckets: []float64{5, 10, 30, 60, 300, 600, 1800, 3600},
+		},
+	)
+	metricTimeToApplySelectorSyncset = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_selector_syncset_apply_duration_seconds",
```
dgoodwin, line 86: Suggest hive_selectorsyncset_apply_duration_seconds to better match the name of the resource.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
dgoodwin: +1 after would be good, the update can easily fail, the metric reporting should never but even if it did, a missed result is much better than dupe results.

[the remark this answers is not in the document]
dgoodwin: :+1:

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -302,6 +325,25 @@ func (r *ReconcileSyncSetInstance) Reconcile(request reconcile.Request) (reconci
...
+			if existingCondition.LastTransitionTime.Time.After(startTime) {
+				startTime = existingCondition.LastTransitionTime.Time
+			}
+		}
+		applyTime := ssi.Status.FirstSuccessTimestamp.Sub(startTime).Seconds()
+		metricTimeToApplySyncset.Observe(float64(applyTime))
+		if sss := ssi.Spec.SelectorSyncSetRef; sss != nil {
+			metricTimeToApplySelectorSyncset.WithLabelValues(sss.Name).Observe(float64(applyTime))
+		}
+	}
```
dgoodwin, line 346: I'm just realizing what will happen when this change rolls out for all existing clusters, we'll get a bunch of first success showing weeks and months. I think they keep a pretty limited window of time, 2 weeks iirc, so in theory this will trend down quickly as new clusters are added and then be fine after a couple weeks. This seems ok to me I think, I can't think of an option other than dont't report metrics over 24h or something.

staebler: Could we check for `ssi.Status.Applied != original.Status.Applied` as well?

staebler: Or just `!original.Status.Applied`?

staebler: And if `original.Status.Applied` is true, then set `ssi.Status.FirstSuccessTimestamp` but do not observe the metric.

dgoodwin: This sounds pretty good to me, what do you think @suhanime

suhanime: Yes it seems to cover all scenarios.

PR #1037, opened 2020-07-10
Add ClusterPool API

dgoodwin:
This is a new construct for Hive which will maintain a pool of X clusters provisioned and ready to be claimed. Once claimed or deleted, the clusterpool controller will replace them. Each cluster is created in it's own namespace.

This feature is intended for organizations that want to have clusters on-hand for quick use rather than waiting 40 minutes for an install.

Future work will add:
- Reference to a ClusterImageSet instead of the hardcoded release image.
- The ClusterClaim API for claiming a cluster and receiving login credentials.
- Namespace cleanup for namespaces which no longer contain a cluster.
- Size override windows for times of the week/day where you'd like more or less clusters kept ready in the pool.


pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,389 @@
...
+			return err
+		}
+		builder.PullSecret = string(pullSec.Data[".dockerconfigjson"])
+	}
+
+	// TODO: regions being ignored throughout, and not exposed in create-cluster cmd either
+	var credsSecretName string
+	if clp.Spec.Platform.AWS != nil {
+		credsSecretName = clp.Spec.Platform.AWS.CredentialsSecretRef.Name
+		// Lookup the platform creds secret for this pool which we are currently assuming
```
staebler, line 277: Why are we assuming that these are in the hive namespace instead of the namespace of the ClusterPool?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,389 @@
...
+		builder.CloudBuilder = azureProvider
+	}
+	// TODO: OpenStack, VMware, and Ovirt.
+
+	objs, err := builder.Build()
+	if err != nil {
+		return errors.Wrap(err, "error building resources")
+	}
+	for _, obj := range objs {
+		if err := r.Client.Create(context.Background(), obj); err != nil {
```
staebler, line 340: I remember from our discussion that we wanted to have the ClusterDeployment be the last thing created to ensure the ClusterDeployment does not attempt to install when there were errors laying down other dependent resources.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,389 @@
...
+		}
+	}
+	logger.WithFields(log.Fields{
+		"installing": installing,
+		"deleting":   deleting,
+		"total":      len(poolCDs),
+		"ready":      len(poolCDs) - installing - deleting,
+	}).Info("found clusters for ClusterPool")
+
+	// If too many, delete some.
```
staebler, line 201: We also need to handle replacing unclaimed ClusterDeployments when the spec of the ClusterPool changes. For example, if the base domain is changed, then all of the existing, unclaimed ClusterDeployments need to be deleted and replaced with new ones.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+			var deletingCDs int
+			for _, cd := range cds.Items {
+				if cd.DeletionTimestamp != nil {
+					deletingCDs++
+				}
+			}
+
+			assert.Equal(t, test.expectedClusters, len(cds.Items))
+			assert.Equal(t, test.expectedDeletedClusters, deletingCDs)
```
staebler, line 126: This is not testing anything. No tests set `expectedDeletedClusters`, and `deletingCDs` will always be 0.

cmd/manager/main.go
```diff
@@ -126,9 +128,11 @@ func newRootCommand() *cobra.Command {
 			hiveNSName := utils.GetHiveNamespace()
 			log.Infof("hive namespace: %s", hiveNSName)
 
+			leaderElection := os.Getenv("DISABLE_LEADER_ELECTION") == ""
```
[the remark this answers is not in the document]
dgoodwin: Broken out to https://github.com/openshift/hive/pull/1042

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -145,6 +149,16 @@ type ClusterImageSetReference struct {
...
 }
 
+// ClusterPoolReference is a reference to a ClusterPool
+type ClusterPoolReference struct {
+	// Name is the name of the ClusterPool that this refers to.
+	Name string `json:"name"`
+	// Namespace is the namespace where the ClusterPool resides.
+	Namespace string `json:"namespace"`
+	// Claimed is true if the ClusterDeployment has been claimed and is not longer actively in the pool.
+	Claimed bool `json:"claimed"`
```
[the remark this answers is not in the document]
dgoodwin: Renamed to State=Unclaimed/Claimed

pkg/apis/hive/v1/clusterpool_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// +required
+	Size int `json:"size"`
+
+	// BaseDomain is the base domain to use for all clusters created in this pool.
+	// +required
+	BaseDomain string `json:"baseDomain"`
+
+	// TODO: implement windows of time in which the Size may be bigger or smaller.
+
+	// DeleteAfter is a duration of time after the ClusterDeployment's creationTimestamp when we should delete the
```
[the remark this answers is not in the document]
dgoodwin: Yeah I'm gonna kill this for now, I think a HibernateAfter is going to prove much more useful.

pkg/constants/constants.go
```diff
@@ -43,6 +43,14 @@ const (
 	// ClusterProvisionNameLabel is the label that is used to identify a relationship to a given cluster provision object.
 	ClusterProvisionNameLabel = "hive.openshift.io/cluster-provision-name"
 
+	// OriginClusterPoolNameLabel is the label that is used to signal that a cluster was originally created for a ClusterPool.
+	// It should always remain set even after a cluster is assigned to fill a request and thus removed from the pool.
+	OriginClusterPoolNameLabel = "hive.openshift.io/origin-cluster-lease-pool-name"
```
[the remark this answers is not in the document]
dgoodwin: I should probably just remove this actually given we now have the explicit reference.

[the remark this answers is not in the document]
dgoodwin: Maybe not, it was used on namespaces and plays into that cleanup card. Will fix. Constant below is gone though.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	// Watch for changes to ClusterPool
+	err = c.Watch(&source.Kind{Type: &hivev1.ClusterPool{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to ClusterDeployments originating from a pool:
+	clusterPredicate := predicate.Funcs{
```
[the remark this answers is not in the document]
dgoodwin: Removed.

pkg/controller/clusterpool/clusterpool_controller.go
dgoodwin: ClusterPool was originally non-Namespaced. I will address.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	// If the pool is deleted, do not reconcile.
+	if clp.DeletionTimestamp != nil {
+		return reconcile.Result{}, nil
+	}
+
+	// Find all ClusterDeployments from this pool:
+	allCDs := &hivev1.ClusterDeploymentList{}
+	if err := r.Client.List(context.Background(), allCDs); err != nil {
```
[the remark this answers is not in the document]
dgoodwin: Hmm I thought we were removing that in favor of the explicit ref, are we sure we want two methods for this? My thought was we've already got all the CDs cached in memory as well so we could live without the label selector.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,389 @@
...
+		logger.WithError(err).Error("error listing ClusterDeployments")
+		return reconcile.Result{}, err
+	}
+	var poolCDs []hivev1.ClusterDeployment
```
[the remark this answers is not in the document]
dgoodwin: Ouch, my bad.

pkg/controller/clusterpool/clusterpool_controller.go
dgoodwin: This is something I was hoping to avoid, changing the release image was the most common case I could think of but I was hoping we could ignore it and let it only affect new clusters being spawned. This would match what we see with the machine API for example, but we hate that, so you're probably right.

I assume we're talking about hashing the spec, storing in status, and if we see a change wipe out all clusters.

I will add a Jira, I'd like to defer this though and get this PR in.

staebler: I think if a user is changing the pool, they don't want to continue getting clusters with the old specs until the pre-existing clusters are all claimed. In practice, I would expect a user that discovers the behavior to always reduce the pool size to 0 and increase it back to what they want to cycle the pool. It seems onerous to put that burden on the user rather than having Hive do it itself.

I am fine with deferring.

pkg/controller/clusterpool/clusterpool_controller.go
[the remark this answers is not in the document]
staebler: I think it will be helpful to have a label to be able to quickly see all of the ClusterDeployments for a ClusterPool. I don't expect ClusterPools to share names that frequently. The fields in the spec are for source of truth: The label is for user-experience. With that being said, I am fine with omitting the label, too.

pkg/controller/clusterpool/clusterpool_controller.go
dgoodwin: Can you refresh my memory on this, I can't recall the details of that convo.

Should we just have the builder return that resource last?

staebler: Let's say that the creds secret is created after the ClusterDeployment is created. Now, let's say that there is an error creating the creds secret. The ClusterDeployment will be stuck in the installing state, counting against the pool's count but never available to be used. If the ClusterDeployment is created last, then it will only be created if all its dependent resources were also created. If there is an error along the way, everything will be cleaned up by the namespace reaper since the namespace will not contain a ClusterDeployment.

dgoodwin: Ok making it created last sounds cleaner than refactoring all this to return the CD separately, I'd hoped to keep that builder object just returning an arbitrary list of resources to create. I will move the CD return to last.

If anyone is still using int we should probably address that underlying rbac problem.

pkg/controller/clusterpool/clusterpool_controller.go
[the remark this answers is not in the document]
dgoodwin: Ok deferring for now on the label.

dgoodwin:
Feedback addressed.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,363 @@
...
+			logger.WithError(err).Error("error looking up credentials secret for pool in hive namespace")
+			return err
+		}
+
+		accessKeyID := credsSecret.Data[constants.AWSAccessKeyIDSecretKey]
+		secretAccessKey := credsSecret.Data[constants.AWSSecretAccessKeySecretKey]
+		awsProvider := &clusterresource.AWSCloudBuilder{
+			AccessKeyID:     string(accessKeyID),
+			SecretAccessKey: string(secretAccessKey),
+		}
```
staebler, line 266: It would be nice to extract this into a `NewAWSProviderFromCredsSecret` function in pkg/clusterresource package so that it is easier to maintain when there are changes to the provider.

pkg/controller/clusterpool/clusterpool_controller.go
staebler: You do not need to change the builder at all. Just have this controller move the CD to the end of the slice.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,363 @@
...
+	for _, obj := range objs {
+		if err := r.Client.Create(context.Background(), obj); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *ReconcileClusterPool) obtainRandomNamespace(clp *hivev1.ClusterPool) (*corev1.Namespace, error) {
```
staebler, line 322:
```suggestion
func (r *ReconcileClusterPool) createRandomNamespace(clp *hivev1.ClusterPool) (*corev1.Namespace, error) {
```

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,363 @@
...
+	return ns, err
+}
+
+func (r *ReconcileClusterPool) deleteExcessClusters(
+	clp *hivev1.ClusterPool,
+	poolCDs []*hivev1.ClusterDeployment,
+	deletionsNeeded int,
+	logger log.FieldLogger) error {
+
+	logger.Infof("too many clusters, searching for %d to delete", deletionsNeeded)
```
staebler, line 343:
```suggestion
	logger.WithField("deletionsNeeded", deletionsNeeded).Info("deleting excess clusters")
```

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,363 @@
...
+	counter := 0
+	for _, cd := range poolCDs {
+		cdLog := logger.WithField("cluster", fmt.Sprintf("%s/%s", cd.Namespace, cd.Name))
+		if cd.DeletionTimestamp != nil {
+			cdLog.WithFields(log.Fields{"cdName": cd.Name, "cdNamespace": cd.Namespace}).Debug("cluster already deleting")
```
staebler, line 348: Do we really need the "cdName" and "cdNamespace" fields when there is already the "cluster" field that includes the cd namespace and name?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -0,0 +1,363 @@
...
+func (r *ReconcileClusterPool) deleteExcessClusters(
+	clp *hivev1.ClusterPool,
+	poolCDs []*hivev1.ClusterDeployment,
+	deletionsNeeded int,
+	logger log.FieldLogger) error {
+
+	logger.Infof("too many clusters, searching for %d to delete", deletionsNeeded)
+	counter := 0
+	for _, cd := range poolCDs {
+		cdLog := logger.WithField("cluster", fmt.Sprintf("%s/%s", cd.Namespace, cd.Name))
```
staebler, line 346: Is the namespace necessary when the name and namespace are always the same?

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+				testClusterDeployment("c6"),
+			},
+			expectedClusters: 3,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			//logger := log.WithField("controller", "clusterProvision")
+			fakeClient := fake.NewFakeClient(test.existing...)
```
staebler, line 87: `NewFakeClient` is deprecated. Use `NewFakeClientWithScheme`.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+			require.NoError(t, err)
+
+			var deletingCDs int
+			for _, cd := range cds.Items {
+				if cd.DeletionTimestamp != nil {
+					deletingCDs++
+				}
+			}
+
+			assert.Equal(t, test.expectedClusters, len(cds.Items))
```
staebler, line 119:
```suggestion
			assert.Len(t, cd.Items, test.expectedClusters)
```

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/openshift/hive/pkg/apis"
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/apis/hive/v1/aws"
+	hivev1aws "github.com/openshift/hive/pkg/apis/hive/v1/aws"
+)
+
+const (
+	testNamespace     = "test-namespace"
+	hiveNamespace     = "hive"
```
staebler, line 30: unused

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+	log.SetLevel(log.DebugLevel)
+}
+
+func TestReconcileClusterPool(t *testing.T) {
+	apis.AddToScheme(scheme.Scheme)
+
+	tests := []struct {
+		name      string
+		existing  []runtime.Object
+		expectErr bool
```
staebler, line 45: unused

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+	hiveNamespace     = "hive"
+	testLeasePoolName = "aws-us-east-1"
+	credsSecretName   = "aws-creds"
+)
+
+func init() {
+	log.SetLevel(log.DebugLevel)
+}
+
+func TestReconcileClusterPool(t *testing.T) {
```
staebler, line 39: Add test cases for the following.
1) Clusters that are not part of a pool do not count against pool size.
2) Clusters that have been claimed from the pool do not count against pool size.
3) Clusters that are part of a different pool do not count against pool size.
4) Clusters that are deleting do not count against pool size.

PR #1039, opened 2020-07-13
wait for dnszone to finish delete before clearing cd finalizer

staebler:
Do not remove the deprovision finalizer from the ClusterDeployment until the DNSZone created for managed DNS has been removed from storage. This helps users that rely on the ClusterDeployment being removed from storage as a trigger for when it is safe to delete the namespace.

https://issues.redhat.com/browse/CO-966

### DIFF
--- a/pkg/test/clusterdeprovision/clusterdeprovision.go
+++ b/pkg/test/clusterdeprovision/clusterdeprovision.go
@@ -1,0 +1,88 @@
+package clusterdeprovision
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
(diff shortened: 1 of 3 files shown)


pkg/test/clusterdeprovision/clusterdeprovision.go
```diff
@@ -0,0 +1,88 @@
...
+	Options(opts ...Option) Builder
+
+	GenericOptions(opts ...generic.Option) Builder
+}
+
+func BasicBuilder() Builder {
+	return &builder{}
+}
+
+func FullBuilder(namespace, name string, typer runtime.ObjectTyper) Builder {
```
twiest, line 35: Ooh, I like this `BasicBuilder` and `FullBuilder` pattern

PR #1042, opened 2020-07-14
Add env var for devs to disable leader election.

dgoodwin:
Allows developers using make run to bypass the delay on startup waiting
for leader election timeouts.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -5,6 +5,7 @@ import (
 	golog "log"
 	"math/rand"
 	"os"
+	"strconv"
 	"time"
 
 	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -128,9 +128,12 @@ func newRootCommand() *cobra.Command {
 			hiveNSName := utils.GetHiveNamespace()
 			log.Infof("hive namespace: %s", hiveNSName)
 
+			// Allow an env var to disable leader election, useful when testing from source with make run to eliminate the wait time.
+			leaderElection := os.Getenv("DISABLE_LEADER_ELECTION") == ""
```
staebler, line 132: Should this check that the value of the environment variable can be interpreted as true? I would find it unintuitive if I ran the following and leader election was disabled.
```
DISABLE_LEADER_ELECTION="false" make run
```

joelddiaz: omg i love this idea. I do agree we should run the value of the env var through some turn-this-into-a-boolean logic.

dgoodwin: Yeah good point, updated.

PR #1024, opened 2020-06-24
oVirt Provisioning Support

abutcher:

PR #1043, opened 2020-07-14
Add OVirt remotemachineset actuator

twiest:
Add OVirt remotemachineset actuator
- [x] Add OVirt remotemachineset actuator
- [x] Add unit tests
- [x] Manually test against real OVirt cluster


twiest:
Holding because this PR relies on #1024.

/hold

twiest:
@joelddiaz Thanks for reviewing. Only reviewing the last commit is correct. The other commits are part of PR #1024. 

I believe I've addressed your feedback, PTAL.

Also, please ignore the "this branch has conflicts" message as this PR will be rebased when PR #1024 is merged.

pkg/controller/remotemachineset/ovirt_test.go
```diff
@@ -0,0 +1,111 @@
...
+	p := testMachinePool()
+	p.Spec.Platform = hivev1.MachinePoolPlatform{
+		Ovirt: &hivev1ovirt.MachinePool{
+			CPU: &hivev1ovirt.CPU{
+				Cores:   4,
+				Sockets: 1,
+			},
+			MemoryMB: 16384,
+			OSDisk: &hivev1ovirt.Disk{
+				SizeGB: 120,
```
joelddiaz, line 93: Sorry for not clarifying. My original feedback was about constifying all of these (cores, sockets, memoryMB, sizeGB).

twiest:
@joelddiaz Ok, I believe I understood this time. PTAL.

pkg/clusterresource/builder.go
```diff
@@ -158,9 +155,8 @@ func (o *Builder) Build() ([]runtime.Object, error) {
 
 	var allObjects []runtime.Object
 	allObjects = append(allObjects, o.generateClusterDeployment())
-	if !o.SkipMachinePoolGeneration {
-		allObjects = append(allObjects, o.generateMachinePool())
-	}
+	allObjects = append(allObjects, o.generateMachinePool())
```
staebler, line 158: Can we please change this to check if the generated machine pool is nil so that we do not have to keep adding and removing `SkipMachinePoolGeneration`?
```suggestion
	if mp := o.generateMachinePool(); mp != nil {
		allObjects = append(allObjects, mp)
	}
```

twiest:

staebler:
@twiest This still has a hold on it. I am assuming that the hold can be lifted now that #1024 has merged, but I will leave that to you.

twiest:
@staebler yep, thanks!

/hold cancel

twiest:
@staebler could I get another lgtm please?

I missed that your example code has used the `mp` variable as part of the append. So I just fixed that.

PR #1048, opened 2020-07-16
update OWNERS

gregsheremeta:
/cc @dgoodwin


gregsheremeta:
this does not need e2e, right?

PR #1049, opened 2020-07-16
Add new metrics to evaluate syncsetinstance performance.

dgoodwin:
hive_syncsetinstance_apply_duration_seconds - Histogram for the amount
of time to apply individual resources within all SyncSetInstances.

hive_syncsetinstance_resources_applied_total - Counter for the rate at
which we're applying individual resources.

hive_kube_client_request_seconds - Histogram for the amount of time to
make remote cluster API requests, similar to existing one for the rate
at which we're making them. Can be combined with the apply duration to
determine how much time we're spending CPU bound vs network bound.

### DIFF
--- a/pkg/controller/utils/clientwrapper.go
+++ b/pkg/controller/utils/clientwrapper.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"k8s.io/client-go/rest"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/clientwrapper.go
```diff
@@ -89,15 +98,18 @@ type ControllerMetricsTripper struct {
...
 	if cmt.Remote {
 		remoteStr = "true"
 	}
 	path, pathErr := parsePath(req.URL.Path)
 	// Call the nested RoundTripper.
 	resp, err := cmt.RoundTripper.RoundTrip(req)
+	applyTime := metav1.Now().Sub(startTime).Seconds()
 	if err == nil && pathErr == nil {
 		metricKubeClientRequests.WithLabelValues(cmt.Controller, req.Method, path, remoteStr, resp.Status).Inc()
+		metricKubeClientRequestSeconds.WithLabelValues(cmt.Controller, req.Method, path, remoteStr, resp.Status).Observe(applyTime)
```
dgoodwin, line 112: This is showing up weirdly fast, like 0.012s for a remote GET, 0.009 for a POST, is this in the right place? @staebler

meowfaceman:
Just wanted to sync on this -- Service Delivery scale needs this PR, and looks like it would merge but the pool is blocked? Is this a Origin CI issue?

staebler:
> Just wanted to sync on this -- Service Delivery scale needs this PR, and looks like it would merge but the pool is blocked? Is this a Origin CI issue?

The merge is blocked because we close the master branch for a few days every week before we do a push to prod to give the version that we push to prod some time to soak in stage. The master branch should open again some time today, at which point this PR should merge.

meowfaceman:
Ack, thanks for the heads up. Sounds good!

PR #1052, opened 2020-07-17
WIP: Server side apply use in resource applier.

dgoodwin:
This needs to be modified to only take this path if the cluster is
OpenShift 4.5+ which has Kube 1.18, and thus server side apply. It also
needs testing to ensure we're getting the apply results we expect.

### DIFF
--- a/pkg/resource/apply.go
+++ b/pkg/resource/apply.go
@@ -3,9 +3,13 @@ package resource
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
+	"net/http"
 
 	"github.com/jonboulle/clockwork"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/resource/apply.go
```diff
@@ -47,11 +52,99 @@ func (r *Helper) Apply(obj []byte) (ApplyResult, error) {
...
 	defer r.deleteTempFile(fileName)
+
 	factory, err := r.getFactory("")
 	if err != nil {
 		r.logger.WithError(err).Error("failed to obtain factory for apply")
 		return "", err
 	}
+
+	// TODO: Replace with a cluster version check, if this cluster is 4.5+ and thus Kube 1.18+, we should have
+	// server side apply available, which is drastically faster than the local approach which currently involves a file
```
staebler, line 63: It must not be the file write that is slow since the file write has already been done at this point, even when doing a server-side apply.

dgoodwin: I think I just botched this PR slightly, I am near certain I commented out the file write, but I did not carry that through to this commit somehow.

dgoodwin:
This is still a beta feature and doesn't work at all for OpenShift types yet. Closing for now, we may want to look at this again later.

/close

PR #1053, opened 2020-07-17
syncset controller: Restore reconcile duration metric

2uasimojo:
This was (accidentally, we think?) removed via 70c7ae2e6e52ef493d6b6a487ca1d8c1aac6f927 in [#437](https://github.com/openshift/hive/pull/437/files#diff-6f5e4dec9c039eb5be5eed0036fc3604L170-L182)

### DIFF
--- a/pkg/controller/syncset/syncset_controller.go
+++ b/pkg/controller/syncset/syncset_controller.go
@@ -3,6 +3,7 @@ package syncset
 import (
 	"context"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


2uasimojo:
@dgoodwin might want :eyes: since you set up the original metrics via #446.

staebler:
The syncset controller should almost certainly be adding metrics.

/hold cancel

dgoodwin:
Nice catch, thanks @2uasimojo

PR #1054, opened 2020-07-17
hive controller: Add reconcile duration metric

2uasimojo:
This was missed in #446.

### DIFF
--- a/pkg/operator/hive/hive_controller.go
+++ b/pkg/operator/hive/hive_controller.go
@@ -12,6 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
 	"github.com/openshift/hive/pkg/resource"
 
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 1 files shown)


2uasimojo:
> This was missed in #446.

If that was deliberate, close me!

2uasimojo:
@dgoodwin PTAL since you authored #446

pkg/operator/hive/hive_controller.go
```diff
@@ -12,6 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
```
staebler, line 15: This will cause the hive-operator pod to push out metrics a bunch of empty metrics that are used by the hive-controller pod.

2uasimojo: Mm, true story. I didn't realize there were separate executables in separate pods. That probably also means the one added here wouldn't show up at the same endpoint as the others, so it would need a whole new thing to scrape it. True?

staebler: Yes, it would be a new thing to scrape. Currently, the hive-operator pod does not serve any metrics.

2uasimojo: Okay, so this PR is certainly no good as written. I guess the next question is: is there value in exposing a reconcile duration metric for the hive controller? At a glance, I don't see anything there that should be impacted by scaling, which is what's driving my current interest here. So I'll go ahead and close this.

Thanks for the feedback @staebler!

2uasimojo:
Closing. See [this discussion](https://github.com/openshift/hive/pull/1054#discussion_r456667884).

PR #1055, opened 2020-07-20
Add AuthenticationFailure condition to dnszone controller

twiest:
Add AuthenticationFailure condition to dnszone controller
- [x] Add AuthenticationFailure condition to dnszone controller
- [x] Add unit tests
- [x] Manually test

### DIFF
--- a/pkg/controller/dnszone/awsactuator.go
+++ b/pkg/controller/dnszone/awsactuator.go
@@ -472,6 +472,103 @@ func (a *AWSActuator) Exists() (bool, error) {
 	return a.zoneID != nil, nil
 }
 
+func (a *AWSActuator) setInsufficientCredentialsConditionToFalse() bool {
+	accessDeniedConds, accessDeniedCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.InsufficientCredentialsCondition,
+		corev1.ConditionFalse,
+		accessGrantedReason,
+		"credentials are valid",
+		controllerutils.UpdateConditionNever,
+	)
+	if accessDeniedCondsChanged {
+		a.dnsZone.Status.Conditions = accessDeniedConds
+	}
+
+	return accessDeniedCondsChanged
+}
+
+func (a *AWSActuator) setInsufficientCredentialsConditionToTrue(message string) bool {
+	accessDeniedConds, accessDeniedCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.InsufficientCredentialsCondition,
+		corev1.ConditionTrue,
+		accessDeniedReason,
+		message,
+		controllerutils.UpdateConditionIfReasonOrMessageChange,
+	)
+
+	if accessDeniedCondsChanged {
+		// Conditions have changed. Update them in the object.
+		a.dnsZone.Status.Conditions = accessDeniedConds
+	}
+	return accessDeniedCondsChanged
+}
+
+func (a *AWSActuator) setAuthenticationFailureConditionToFalse() bool {
+	authenticationFailureConds, authenticationFailureCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.AuthenticationFailureCondition,
+		corev1.ConditionFalse,
+		authenticationSucceededReason,
+		"credentials authenticated",
+		controllerutils.UpdateConditionNever,
+	)
+	if authenticationFailureCondsChanged {
+		a.dnsZone.Status.Conditions = authenticationFailureConds
+	}
+
+	return authenticationFailureCondsChanged
+}
+
+func (a *AWSActuator) setAuthenticationFailureConditionToTrue(message string) bool {
+	var authenticationFailureConds []hivev1.DNSZoneCondition
+	authenticationFailureConds, authenticationFailureCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.AuthenticationFailureCondition,
+		corev1.ConditionTrue,
+		authenticationFailedReason,
+		message,
+		controllerutils.UpdateConditionIfReasonOrMessageChange,
+	)
+
+	if authenticationFailureCondsChanged {
+		// Conditions have changed. Update them in the object.
+		a.dnsZone.Status.Conditions = authenticationFailureConds
+	}
+
+	return authenticationFailureCondsChanged
+}
+
+// SetConditionsForError sets conditions on the dnszone given a specific error. Returns true if conditions changed.
+func (a *AWSActuator) SetConditionsForError(err error) bool {
+	awsErr, ok := err.(awserr.Error)
+	if !ok {
+		accessDeniedCondsChanged := a.setInsufficientCredentialsConditionToFalse()
+		authenticationFailureCondsChanged := a.setAuthenticationFailureConditionToFalse()
+
+		return accessDeniedCondsChanged || authenticationFailureCondsChanged
+	}
+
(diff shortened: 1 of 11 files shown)


pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -217,38 +218,76 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
...
 		corev1.ConditionFalse,
 		accessDeniedReason,
 		"credentials are valid",
 		controllerutils.UpdateConditionNever,
 	)
-	if changed {
-		dnsZone.Status.Conditions = conds
+	if accessDeniedCondsChanged {
+		dnsZone.Status.Conditions = accessDeniedConds
+		if err := r.Status().Update(context.Background(), dnsZone); err != nil {
```
staebler, line 276: Only call `r.Status().Update` once if either `accessDeniedCondsChanged` or `authFailureCondsChanged`.

twiest:
@staebler I believe I've addressed your feedback and I implemented the changes we discussed in standup.

I also finished the manual testing. I believe this PR is ready to go. PTAL.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -472,6 +472,53 @@ func (a *AWSActuator) Exists() (bool, error) {
...
 }
 
+// SetConditionsForError sets conditions on the dnszone given a specific error. Returns true if conditions changed.
+func (a *AWSActuator) SetConditionsForError(err error) bool {
+	awsErr, ok := err.(awserr.Error)
+	if !ok {
+		return false
+	}
+
+	conds := []hivev1.DNSZoneCondition{}
```
staebler, line 482:
```suggestion
	var conds []hivev1.DNSZoneCondition
```

pkg/controller/dnszone/awsactuator.go
```diff
@@ -472,6 +472,53 @@ func (a *AWSActuator) Exists() (bool, error) {
...
+			corev1.ConditionTrue,
+			accessDeniedReason,
+			awsErr.Message(),
+			controllerutils.UpdateConditionIfReasonOrMessageChange,
+		)
+		failureMessage = "Credentials are lacking permissions"
+
+	case "InvalidSignatureException":
+		fallthrough
+	case "UnrecognizedClientException":
```
staebler, line 500: I think the following conveys the logic a bit better. The fallthough, at least to me, implies that the first case did something and should fall through to do more.
```suggestion
	case "InvalidSignatureException", "UnrecognizedClientException":
```

pkg/controller/dnszone/awsactuator.go
```diff
@@ -472,6 +472,53 @@ func (a *AWSActuator) Exists() (bool, error) {
...
+		)
+		failureMessage = "Credentials failed authentication"
+	}
+
+	if changed {
+		// Conditions have changed. Update them in the object.
+		a.dnsZone.Status.Conditions = conds
+	}
+
+	a.logger.WithError(err).Warn(failureMessage)
```
staebler, line 517: This seems out of place here.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -217,38 +217,43 @@ func (r *ReconcileDNSZone) reconcileDNSProvider(actuator Actuator, dnsZone *hive
...
-			r.logger.WithError(err).Warn("Credentials are lacking permissions")
-			return reconcile.Result{}, err
 		}
+
 		r.logger.WithError(err).Error("Failed to retrieve hosted zone and corresponding tags")
 		return reconcile.Result{}, err
 	}
 
-	conds, changed := controllerutils.SetDNSZoneConditionWithChangeCheck(
+	accessDeniedConds, accessDeniedCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
```
staebler, line 232: These conditions will only get set to false when there are no longer **any** errors. For example, let's say that there was initially an authentication error. The user fixes that but then there is an access denied error. The DNSZone will show that there is still an authentication error and that there is now **also** an access denied error.

twiest: In your example, wouldn't fixing the authentication error cause authenticationFailureCondsChanged to become true?

I'm not seeing how the code works the way you describe.

twiest: Ah, I get it. You're saying because we'd go into the ```if err != nil``` above and so this code wouldn't be reached. Good catch.

twiest:
@staebler I believe I've addressed your concerns. PTAL.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -472,6 +472,107 @@ func (a *AWSActuator) Exists() (bool, error) {
 	return a.zoneID != nil, nil
 }
 
+func (a *AWSActuator) setInsufficientCredentialsConditionToFalse() bool {
+	accessDeniedConds, accessDeniedCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.InsufficientCredentialsCondition,
+		corev1.ConditionFalse,
+		accessDeniedReason,
```
staebler, line 480: This is not the correct reason.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -472,6 +472,107 @@ func (a *AWSActuator) Exists() (bool, error) {
...
+// SetConditionsForError sets conditions on the dnszone given a specific error. Returns true if conditions changed.
+func (a *AWSActuator) SetConditionsForError(err error) bool {
+	if err == nil {
+		accessDeniedCondsChanged := a.setInsufficientCredentialsConditionToFalse()
+		authenticationFailureCondsChanged := a.setAuthenticationFailureConditionToFalse()
+
+		return accessDeniedCondsChanged || authenticationFailureCondsChanged
+	}
+
+	awsErr, ok := err.(awserr.Error)
```
staebler, line 552: The InsufficientCredentials and AuthenticationFailure conditions should be set to false when the error is not an `awserr.Error`.

pkg/controller/dnszone/awsactuator.go
twiest: Are you sure?

That's not my code, I just refactored it here.

staebler: I am sure that it is the wrong reason. I do see that it is an existing problem and not something introduce in this PR.

twiest: What should the reason be then?

@akhil-rane git blame says that you added this. Do you have input on what this should be?

pkg/controller/dnszone/awsactuator.go
twiest: Fixed.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -472,6 +472,110 @@ func (a *AWSActuator) Exists() (bool, error) {
...
+		// Conditions have changed. Update them in the object.
+		a.dnsZone.Status.Conditions = authenticationFailureConds
+	}
+
+	return authenticationFailureCondsChanged
+}
+
+// SetConditionsForError sets conditions on the dnszone given a specific error. Returns true if conditions changed.
+func (a *AWSActuator) SetConditionsForError(err error) bool {
+	if err == nil {
```
staebler, line 545: This check is not necessary. It will be covered by the check for whether the error is a `awserr.Error`.

pkg/controller/dnszone/awsactuator.go
staebler: In this case, the condition is being set to False. The reason should not be "AccessDenied" because access was not denied.

twiest:
@staebler @akhil-rane I addressed your comments, PTAL.

Specifically are the new `access*Reason` and `authentication*Reason` acceptable to you?

PR #1057, opened 2020-07-21
reap namespaces created for ClusterPool after clusters deleted

staebler:
When the ClusterPool creates a new cluster, it creates a namespace to house the ClusterDeployment. When the ClusterDeployment is deleted, the namespace then needs to be deleted to avoid leaking namespaces. To this end, these changes add a clusterpoolnamespace controller which will sync namespaces and delete them when appropriate. All namespaces
created for ClusterPools will have a hive.openshift.io/origin-cluster-pool-name label. The new controller will delete a namespace with that label if the namespace does not contain any ClusterDeployments. The controller will wait 5 minutes after the creation of the namespace before deleting the namespace to give time for the ClusterDeployment to be created.

https://issues.redhat.com/browse/CO-1014

### DIFF
--- a/pkg/controller/clusterpoolnamespace/clusterpoolnamespace_controller.go
+++ b/pkg/controller/clusterpoolnamespace/clusterpoolnamespace_controller.go
@@ -1,0 +1,157 @@
+package clusterpoolnamespace
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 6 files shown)


pkg/controller/clusterpoolnamespace/clusterpoolnamespace_controller.go
```diff
@@ -0,0 +1,157 @@
...
+			Reconciler:              r,
+			MaxConcurrentReconciles: controllerutils.GetConcurrentReconciles(),
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to Namespaces
+	if err := c.Watch(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
dgoodwin, line 63: Should we limit to just those with our label?

staebler: We **can**. It is quite a bit more code to do it in a predicate rather than doing it in `Reconcile`. Do we have a strong reason for wanting to use a predicate?

dgoodwin:
Nah that's fine, thought it was fairly straightforward but not sure what I'm remembering.

/lgtm

Issue #847, opened 2020-02-24
Question: When using Hive to provision a cluster in AWS, how does one influence the region the cluster lands in?


benemon:
If I generate a cluster deployment using the following hiveutil command:

`hiveutil create-cluster --base-domain=my.domain --cloud=aws --creds-file=/home/bholmes/.aws/credentials --ssh-private-key-file=/home/bholmes/.ssh/hive_id_rsa --ssh-public-key-file=/home/bholmes/.ssh/hive_id_rsa.pub --pull-secret-file=pull-secret.json feb -o yaml > my-domain.yaml`

Then I get a list of items which - among other things - contains a base64 encoded inall-config, as well as a ClusterDeployment. The ClusterDeployment contains:

```
- apiVersion: hive.openshift.io/v1
  kind: ClusterDeployment
  ...
  spec:
     ...
    platform:
      aws:
       ...
        region: us-east-1
```

If I change the region here to eu-west-1, then it is mostly ignored by the installer and can lead to some weird errors and inconsistencies at deploy time - mostly because the install-config.yaml contains hardcoded references to us-east-1.

Is there any way currently to influence this from hiveutil? Or do we currently need to decode-update-then recode the install-config before applying it to the hive namespace?

Thanks!

dgoodwin:
Could have sworn we had --region but it looks like it's gone. The install config pools and hive pools should generally match, especially with big things like region, otherwise you will get pretty dramatic cycling of workers post-install. In this case I think you need to decode that install config like you said to ensure it has the right region. (or patch hiveutil)

benemon:
Thanks @dgoodwin, that clears up that little mystery. As I know sweet FA about Go, I'll resign myself to decoding the install config for now :+1:

dgoodwin:
I'll file a Jira and see if we can get it readded at some point.

PR #1060, opened 2020-07-24
Add ClusterPool API and controller.

staebler:
Allows defining a pool of clusters of a certain size to be kept ready for use. Uses the builder we use for hiveutil create-cluster to generate all artifacts, including an install-config. As such options for customization are a bit restricted, but should be helpful for anyone looking to keep clusters up and running.

Future work will allow clusters to be claimed by users, and immediately hibernating clusters as soon as they're ready and waking them up when claimed.


staebler:
This builds on the work started in https://github.com/openshift/hive/pull/1037.

staebler:
Resolves #847.

dgoodwin:
I've gone through each commit and everything looks good, but the reap namespace commit appears to be both in this PR and #1057 with a different SHA, can you advise what you'd like to do there? Feels like it should be dropped here and we merge it over in #1057.

dgoodwin:
Maybe the diff sha is just how git works if it's applied on another branch, if they are identical we can merge it all here.

staebler:
> I've gone through each commit and everything looks good, but the reap namespace commit appears to be both in this PR and #1057 with a different SHA, can you advise what you'd like to do there? Feels like it should be dropped here and we merge it over in #1057.

I created the reap namespaces PR first. Then, after doing more ClusterPool work, I added it to this PR. I don't care whether we merge the reap namespaces PR first or not. If it does go in first, then I will rebase this PR to exclude the reap namespaces work.

PR #1061, opened 2020-07-24
do not use temp file for resource apply

staebler:
Stop writing to a temp file when doing a resource apply.


dgoodwin:
This PR is delightful.

/lgtm

PR #1062, opened 2020-07-25
Fix insufficient credentials condition on ClusterDeployment

akhil-rane:
jira: https://issues.redhat.com/browse/CO-1064

### DIFF
--- a/pkg/controller/dnszone/dnszone_controller_test.go
+++ b/pkg/controller/dnszone/dnszone_controller_test.go
@@ -2,12 +2,15 @@ package dnszone
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/golang/mock/gomock"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -141,20 +141,6 @@ func TestReconcileDNSProviderForAWS(t *testing.T) {
...
-				mockAccessDeniedException(expect, validDNSZoneWithoutID())
-			},
-			validateZone: func(t *testing.T, zone *hivev1.DNSZone) {
-				condition := controllerutils.FindDNSZoneCondition(zone.Status.Conditions, hivev1.InsufficientCredentialsCondition)
-				assert.NotNil(t, condition, "invalid credentials condition should be set on dnszone")
-				assert.Equalf(t, "User: arn:aws:iam::0123456789:user/testAdmin is not authorized to perform: tag:GetResources with an explicit deny",
-					condition.Message, "condition has an unexpected message")
-			},
-			errorExpected: true,
-		},
```
akhil-rane, line 157: For unit tests I might have to mock `getActuator` method of dnszone controller and to do that  I will have to make it a function variable in `ReconcileDNSZone` struct (similar to [soaLookup](https://github.com/openshift/hive/blob/3261cec3ba098e1ca4019db3d1b91a2db86f95c9/pkg/controller/dnszone/dnszone_controller.go#L111) method).

staebler: That is a fine way if you need it. Will it still be necessary once https://github.com/openshift/hive/pull/1055 lands, which moves the code for setting conditions based on errors into the individual actuators?

staebler: Actually, I wonder if this entire PR may not be necessary given the changes in #1055.

akhil-rane: Other actuator methods can also throw access denied errors, so I thought of moving this check outside `reconcileDNSProvider`. May be I can use `SetConditionsForError` for each actuator method as done in #1055.

staebler: Ah, yes. I got confused at though that #1055 was doing that, too.

PR #1063, opened 2020-07-27
Detect install pod stuck in pending phase

akhil-rane:
This PR adds a ClusterDeployment condition to detect install pod stuck in the pending phase. The install pod will be in the pending phase for sometime after creating a job. To handle this the pod status check is performed only if 60 seconds have elapsed after creating the job.

jira: https://issues.redhat.com/browse/CO-1037

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -789,11 +789,9 @@ func (r *ReconcileClusterDeployment) reconcileExistingProvision(cd *hivev1.Clust
 
 	switch provision.Spec.Stage {
 	case hivev1.ClusterProvisionStageInitializing:
-		cdLog.Debug("still initializing provision")
-		return reconcile.Result{}, nil
+		return r.reconcileInitializingProvision(cd, provision, cdLog)
 	case hivev1.ClusterProvisionStageProvisioning:
-		cdLog.Debug("still provisioning")
-		return reconcile.Result{}, nil
+		return r.reconcileProvisioningProvision(cd, provision, cdLog)
 	case hivev1.ClusterProvisionStageFailed:
 		return r.reconcileFailedProvision(cd, provision, cdLog)
 	case hivev1.ClusterProvisionStageComplete:
@@ -802,6 +800,28 @@ func (r *ReconcileClusterDeployment) reconcileExistingProvision(cd *hivev1.Clust
 		cdLog.WithField("stage", provision.Spec.Stage).Error("unknown provision stage")
 		return reconcile.Result{}, errors.New("unknown provision stage")
 	}
+}
+
+func (r *ReconcileClusterDeployment) reconcileInitializingProvision(cd *hivev1.ClusterDeployment, provision *hivev1.ClusterProvision, cdLog log.FieldLogger) (reconcile.Result, error) {
+	cdLog.Debug("still initializing provision")
+	// Set condition on ClusterDeployment when install pod is stuck in pending phase
+	installPodStuckCondition := controllerutils.FindClusterProvisionCondition(provision.Status.Conditions, hivev1.InstallPodStuckCondition)
+	if installPodStuckCondition != nil && installPodStuckCondition.Status == corev1.ConditionTrue {
+		if err := r.setInstallLaunchErrorCondition(cd, corev1.ConditionTrue, installPodStuckCondition.Reason, installPodStuckCondition.Message, cdLog); err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not update InstallLaunchErrorCondition")
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileClusterDeployment) reconcileProvisioningProvision(cd *hivev1.ClusterDeployment, provision *hivev1.ClusterProvision, cdLog log.FieldLogger) (reconcile.Result, error) {
+	cdLog.Debug("still provisioning")
+	if err := r.setInstallLaunchErrorCondition(cd, corev1.ConditionFalse, "InstallLaunchSuccessful", "Successfully launched install pod", cdLog); err != nil {
+		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not update InstallLaunchErrorCondition")
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
 }
 
 func (r *ReconcileClusterDeployment) reconcileFailedProvision(cd *hivev1.ClusterDeployment, provision *hivev1.ClusterProvision, cdLog log.FieldLogger) (reconcile.Result, error) {
@@ -1118,6 +1138,22 @@ func (r *ReconcileClusterDeployment) setDNSNotReadyCondition(cd *hivev1.ClusterD
 	}
 	cd.Status.Conditions = conditions
 	cdLog.Debugf("setting DNSNotReadyCondition to %v", status)
+	return r.Status().Update(context.TODO(), cd)
+}
+
+func (r *ReconcileClusterDeployment) setInstallLaunchErrorCondition(cd *hivev1.ClusterDeployment, status corev1.ConditionStatus, reason string, message string, cdLog log.FieldLogger) error {
+	conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+		cd.Status.Conditions,
+		hivev1.InstallLaunchErrorCondition,
+		status,
+		reason,
+		message,
+		controllerutils.UpdateConditionIfReasonOrMessageChange)
+	if !changed {
+		return nil
+	}
+	cd.Status.Conditions = conditions
+	cdLog.WithField("status", status).Debug("setting InstallLaunchErrorCondition")
 	return r.Status().Update(context.TODO(), cd)
 }
 
--- a/pkg/controller/clusterprovision/clusterprovision_controller.go
+++ b/pkg/controller/clusterprovision/clusterprovision_controller.go
@@ -2,6 +2,7 @@ package clusterprovision
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/pkg/errors"
@@ -41,6 +42,8 @@ const (
 
 	resultSuccess = "success"
 	resultFailure = "failure"
+
+	podStatusCheckDelay = 60 * time.Second
 )
 
 var (
@@ -246,7 +249,7 @@ func (r *ReconcileClusterProvision) createJob(instance *hivev1.ClusterProvision,
 
 func (r *ReconcileClusterProvision) adoptJob(instance *hivev1.ClusterProvision, job *batchv1.Job, pLog log.FieldLogger) (reconcile.Result, error) {
 	instance.Status.JobRef = &corev1.LocalObjectReference{Name: job.Name}
-	return reconcile.Result{}, r.setCondition(instance, hivev1.ClusterProvisionJobCreated, "JobCreated", "Install job has been created", pLog)
+	return reconcile.Result{}, r.setCondition(instance, hivev1.ClusterProvisionJobCreated, corev1.ConditionTrue, "JobCreated", "Install job has been created", controllerutils.UpdateConditionAlways, pLog)
 }
 
 // check if the job has completed
@@ -258,7 +261,7 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 	case apierrors.IsNotFound(err):
 		if cond := controllerutils.FindClusterProvisionCondition(instance.Status.Conditions, hivev1.ClusterProvisionFailedCondition); cond == nil {
 			pLog.Error("install job lost")
-			if err := r.setCondition(instance, hivev1.ClusterProvisionFailedCondition, "JobNotFound", "install job not found", pLog); err != nil {
+			if err := r.setCondition(instance, hivev1.ClusterProvisionFailedCondition, corev1.ConditionTrue, "JobNotFound", "install job not found", controllerutils.UpdateConditionAlways, pLog); err != nil {
 				return reconcile.Result{}, err
 			}
 		} else {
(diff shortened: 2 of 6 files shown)


pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -241,12 +245,12 @@ func (r *ReconcileClusterProvision) createJob(instance *hivev1.ClusterProvision,
 		return reconcile.Result{}, err
 	}
 
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: delayAfterInstallJobCreation}, nil
```
staebler, line 248: This is not going to do quite what I suspect you want it to. This will requeue the ClusterProvision in 60 seconds. However, changes to the Job will cause the ClusterProvision to be requeued earlier.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -272,6 +276,20 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 
 	pLog = pLog.WithField("job", job.Name)
 
+	installPod, err := r.getInstallPod(job, pLog)
+	if err != nil {
+		pLog.WithError(err).Error("could not get install pod")
+		return r.transitionStage(instance, hivev1.ClusterProvisionStageFailed, "InstallPodMissing", err.Error(), pLog)
```
staebler, line 282: I am guessing that you are expecting that this will not be hit until the 60 seconds has elapsed, which is not guaranteed.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -272,6 +276,20 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
+	}
+
+	if installPod.Status.Phase == "Pending" {
+		pLog.WithField("pod", installPod.Name).Error("install pod is stuck")
+		if err := r.setCondition(instance, hivev1.ClusterProvisionInitializedCondition, corev1.ConditionFalse, "InstallPodStuck", "install pod is stuck in pending phase", controllerutils.UpdateConditionAlways, pLog); err != nil {
```
staebler, line 287:
```suggestion
		if err := r.setCondition(instance, hivev1.ClusterProvisionInitializedCondition, corev1.ConditionFalse, "InstallPodStuck", "install pod is stuck in pending phase", controllerutils.UpdateConditionIfReasonOrMessageChanged, pLog); err != nil {
```

staebler, line 287: This will not set the condition to False if the condition was not already present.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -272,6 +276,20 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 
 	pLog = pLog.WithField("job", job.Name)
 
+	installPod, err := r.getInstallPod(job, pLog)
```
staebler, line 279: This should be done **after** the check to see if the job was successful or failed. If the job is successful or failed, then the pod may not exist, but we want to rely on what the job is telling us happened.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -306,6 +324,28 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileClusterProvision) getInstallPod(job *batchv1.Job, pLog log.FieldLogger) (*corev1.Pod, error) {
+	installPod := &corev1.Pod{}
+	podList := &corev1.PodList{}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(map[string]string{"job-name": job.Name})}
```
staebler, line 330: Use the selector from the job (`.spec.selector`) to identify the pod.

pkg/controller/clusterprovision/clusterprovision_controller.go
staebler, line 287: I think that I would rather have a separate PodStuck condition that can be set True rather than setting the Initialized condition to False.

pkg/controller/clusterprovision/clusterprovision_controller.go
akhil-rane: Could not find any labels in` .spec.selector` that can identify pod
`spec:
  backoffLimit: 0
  completions: 1
  parallelism: 1
  selector:
    matchLabels:
      controller-uid: 0f5d9e8a-e57e-4d0c-ae86-a43ac3c0c637
`

staebler: The `controller-uid: 0f5d9e8a-e57e-4d0c-ae86-a43ac3c0c637` labels is what you want. But more generically, the pod must match the selector, whatever it is. I am suggesting that you get the job, then get the pod based on the selector specified on the job.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1118,6 +1138,22 @@ func (r *ReconcileClusterDeployment) setDNSNotReadyCondition(cd *hivev1.ClusterD
...
+		hivev1.InstallLaunchErrorCondition,
+		status,
+		reason,
+		message,
+		controllerutils.UpdateConditionIfReasonOrMessageChange)
+	if !changed {
+		return nil
+	}
+	cd.Status.Conditions = conditions
+	cdLog.Debugf("setting InstallLaunchErrorCondition to %v", status)
```
staebler, line 1153:
```suggestion
	cdLog.WithField("status", status).Debug("setting InstallLaunchErrorCondition")
```

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -306,6 +329,28 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 	return reconcile.Result{}, nil
```
staebler, line 329: If it has been less than `podStatusCheckDelay`, need to requeue with a delay equal to the time remaining from `podStatusCheckDelay`. Otherwise, the ClusterProvision will not be reconciled again in a reasonable time to set the InstallPodStuckCondition.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -306,6 +329,28 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
 }
 
+func (r *ReconcileClusterProvision) getInstallPod(job *batchv1.Job, pLog log.FieldLogger) (*corev1.Pod, error) {
+	installPod := &corev1.Pod{}
```
staebler, line 333: Remove this variable. Return nil when there is an error rather than returning a zero `Pod`.

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -374,6 +415,12 @@ func failedJob() jobOption {
 	}
 }
 
+func withCreationTimestamp(time time.Time) jobOption {
+	return func(job *batchv1.Job) {
+		job.CreationTimestamp.Time = time
+	}
```
staebler, line 421:
```suggestion
	return testjob.Generic(testgeneric.WithCreationTimestamp(time))
```

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -285,6 +288,25 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 
 	pLog.Debug("install job still running")
 
+	if time.Since(job.CreationTimestamp.Time) > podStatusCheckDelay {
+		installPod, err := r.getInstallPod(job, pLog)
+		if err != nil {
+			pLog.WithError(err).Error("could not get install pod")
+			return r.transitionStage(instance, hivev1.ClusterProvisionStageFailed, "InstallPodMissing", err.Error(), pLog)
```
staebler, line 295: We cannot transition if the pod cannot be found. We may reconcile the ClusterProvision after the Job has been created but before the Pod has been created.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -303,9 +325,39 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
+	case numberOfPods == 0:
+		pLog.Error("install pod not found")
+		return nil, fmt.Errorf("install pod not found")
+	case numberOfPods > 1:
+		pLog.Error("more than one install pod exists")
+		return nil, fmt.Errorf("more than one install pod exists")
+	}
+
+	installPod := &podList.Items[0]
+	return installPod, nil
```
staebler, line 358:
```suggestion
	return &podList.Items[0], nil
```

staebler, line 358: Take it or leave it, but I find this to be a bit cleaner.
```suggestion
	switch len(podList.Items) {
	case 0:
		pLog.Error("install pod not found")
		return nil, fmt.Errorf("install pod not found")
	case 1:
		return &podList.Items[0], nil
	default:
		pLog.Error("more than one install pod exists")
		return nil, fmt.Errorf("more than one install pod exists")
	}
```

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -285,6 +288,25 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
+		}
+
+		if installPod.Status.Phase == "Pending" {
+			pLog.WithField("pod", installPod.Name).Error("install pod is stuck")
+			if err := r.setCondition(instance, hivev1.InstallPodStuckCondition, corev1.ConditionTrue, "PodInPendingPhase", "pod is in pending phase", controllerutils.UpdateConditionIfReasonOrMessageChange, pLog); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
+		}
+		if err := r.setCondition(instance, hivev1.InstallPodStuckCondition, corev1.ConditionFalse, "PodInRunningPhase", "pod is in running phase", controllerutils.UpdateConditionAlways, pLog); err != nil {
```
staebler, line 305: Since `setCondition` is not doing any change checks, this will do a status update on the ClusterProvision even if nothing has changed.

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -390,3 +439,62 @@ func getProvision(c client.Client) *hivev1.ClusterProvision {
...
+	for _, cond := range provision.Status.Conditions {
+		if cond.Type == condType {
+			found = true
+			assert.Equal(t, string(status), string(cond.Status), "condition found with unexpected status")
+		}
+	}
+	assert.True(t, found, "did not find expected condition type: %v", provision.Status.Conditions)
+}
+
+func assertConditionReason(t *testing.T, cd *hivev1.ClusterProvision, condType hivev1.ClusterProvisionConditionType, reason string) {
```
staebler, line 492: Use the same logic in `assertConditionStatus` and `assertConditionReason`. Either return when the condition is found (my preference) or use a `found` variable.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -285,6 +288,25 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
+			pLog.WithError(err).Error("could not get install pod")
+			return r.transitionStage(instance, hivev1.ClusterProvisionStageFailed, "InstallPodMissing", err.Error(), pLog)
+		}
+
+		if installPod.Status.Phase == "Pending" {
+			pLog.WithField("pod", installPod.Name).Error("install pod is stuck")
+			if err := r.setCondition(instance, hivev1.InstallPodStuckCondition, corev1.ConditionTrue, "PodInPendingPhase", "pod is in pending phase", controllerutils.UpdateConditionIfReasonOrMessageChange, pLog); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
akhil-rane: I did not understand this `Can you something like the following comment here?`502346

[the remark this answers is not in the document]
staebler: Sorry, bad editing. I meant `Can you add a comment like the following here`? The second paragraph was my proposal for the comment.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -285,6 +288,33 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
...
 	pLog.Debug("install job still running")
 
+	if time.Since(job.CreationTimestamp.Time) > podStatusCheckDelay {
+		installPod, err := r.getInstallPod(job, pLog)
+		if err != nil {
+			pLog.WithError(err).Error("could not get install pod")
+			if err := r.setCondition(instance, hivev1.InstallPodStuckCondition, corev1.ConditionTrue, "InstallPodMissing", err.Error(), controllerutils.UpdateConditionIfReasonOrMessageChange, pLog); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, err
```
csrwng, line 298: @akhil-rane @staebler does it matter that the provision instance won't be requeued after this?

staebler: Why would it not be requeued? The reconcile is returning a non-nil error.

csrwng: Ahh you're right, for whatever reason I was thinking it'd be nil if the previous stmt was successful, but yeah, it's only scoped to that stmt

pkg/controller/clusterprovision/clusterprovision_controller_test.go
akhil-rane: Done

akhil-rane:
@staebler addressed your feedback.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -303,9 +333,42 @@ func (r *ReconcileClusterProvision) reconcileRunningJob(instance *hivev1.Cluster
 		}
 	}
 
+	if time.Since(job.CreationTimestamp.Time) < podStatusCheckDelay {
+		timeUntilNextPodStatusCheck := job.CreationTimestamp.Time.Add(podStatusCheckDelay)
+		return reconcile.Result{RequeueAfter: time.Until(timeUntilNextPodStatusCheck)}, nil
+	}
```
staebler, line 339: I would prefer if the values from the if check are used in determining the requeue after, to give it all a bit more cohesion. But it is acceptable as you have it, too.

```suggestion
	if timeUntilNextPodStatusCheck := podStatusCheckDelay - time.Since(job.CreationTimestamp.Time); timeUntilNextPodStatusCheck > 0 {
		return reconcile.Result{RequeueAfter: time.Until(timeUntilNextPodStatusCheck)}, nil
	}
```

```suggestion
	if timeSinceJobCreation := time.Since(job.CreationTimestamp.Time); timeSinceJobCreation < podStatusCheckDelay {
		return reconcile.Result{RequeueAfter: podStatusCheckDelay - timeSinceJobCreation}, nil
	}
```

akhil-rane:
> Changes look good. Squash it down and I will give it a lgtm.

Squashed and ready to merge :)

PR #1066, opened 2020-07-28
delete recordsets in cloud zones when deleting DNSZone

staebler:
When deleting a DNSZone, the cloud zone cannot be deleted while there are still recordsets in the zone. Since the zone was created for the DNSZone, Hive can safely clear the recordsets.

https://issues.redhat.com/browse/CO-1071

### DIFF
--- a/pkg/controller/dnszone/azureactuator.go
+++ b/pkg/controller/dnszone/azureactuator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2018-05-01/dns"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


pkg/controller/dnszone/azureactuator.go
```diff
@@ -82,7 +83,12 @@ func (a *AzureActuator) Delete() error {
 	}
 
 	resourceGroupName := a.dnsZone.Spec.Azure.ResourceGroupName
-	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone).WithField("zoneName", a.managedZone.Name)
+	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone)
```
csrwng, line 86: Just curious, was the `.managedZone.Name` field redundant?

staebler: Yes, mostly. The value of the "zoneName" field was actually a pointer and not a name. I assume the intention was for it to be the dereferenced name, which is the same as the value of the "zone" field.

PR #1067, opened 2020-07-29
Add a metric for time to fully apply all SyncSets for a cluster

abutcher:


pkg/controller/syncset/syncset_controller.go
```diff
@@ -460,6 +483,41 @@ func (r *ReconcileSyncSet) syncSetInstanceForSelectorSyncSet(cd *hivev1.ClusterD
...
+	return nil
+}
+
+func (r *ReconcileSyncSet) updateClusterDeploymentStatus(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	cdLog.Info("updating clusterdeployment status")
+	err := r.Status().Update(context.TODO(), cd)
+	if err != nil {
+		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "cannot update clusterdeployment status")
+	}
+	return nil
```
staebler, line 518:
```suggestion
	return err
```

pkg/controller/syncset/syncset_controller.go
```diff
@@ -460,6 +483,41 @@ func (r *ReconcileSyncSet) syncSetInstanceForSelectorSyncSet(cd *hivev1.ClusterD
...
+		// existing adopted clusterdeployments may not have an installedTimestamp
+		if cd.Status.InstalledTimestamp != nil {
+			allSyncSetsAppliedDuration := lastSuccessTimestamp.Time.Sub(cd.Status.InstalledTimestamp.Time)
+			metricTimeToApplySyncSets.Observe(float64(allSyncSetsAppliedDuration.Seconds()))
+		}
+	}
+	return nil
+}
+
+func (r *ReconcileSyncSet) updateClusterDeploymentStatus(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
```
staebler, line 512: This probably does not need to be a extracted as a function.

pkg/controller/syncset/syncset_controller_test.go
```diff
@@ -304,3 +306,61 @@ func testHashCompute(obj interface{}) (string, error) {
...
+		return
+	}
+
+	uCD := &hivev1.ClusterDeployment{}
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: testName, Namespace: testNamespace}, uCD)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+		return
+	}
+	assert.Equal(t, *uCD.Status.FirstSyncSetsSuccessTimestamp, *syncSetInstances[2].(*hivev1.SyncSetInstance).Status.FirstSuccessTimestamp)
```
staebler, line 365: The expected value should be first. Also, it will read better if you store the expected value in a variable rather than reading from a field in a slice.
```suggestion
	assert.Equal(t, expectedFirstSuccessTime, *uCD.Status.FirstSyncSetsSuccessTimestamp, "unexpected first success timestamp")
```

pkg/controller/syncset/syncset_controller_test.go
```diff
@@ -304,3 +306,61 @@ func testHashCompute(obj interface{}) (string, error) {
...
+	_, err := rss.Reconcile(reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      testName,
+			Namespace: testNamespace,
+		},
+	})
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+		return
+	}
```
staebler, line 357:
```suggestion
	require.NoError(t, err, "unexpected error from Reconcile")
```

pkg/controller/syncset/syncset_controller_test.go
```diff
@@ -304,3 +306,61 @@ func testHashCompute(obj interface{}) (string, error) {
...
+
+	uCD := &hivev1.ClusterDeployment{}
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: testName, Namespace: testNamespace}, uCD)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+		return
+	}
```
staebler, line 364:
```suggestion
	require.NoError(t, err, "unexpected error getting ClusterDeployment")
```

pkg/controller/syncset/syncset_controller_test.go
staebler, line 365: Need to check whether the timestamp in the ClusterDeployment is nil first, to avoid panics in the test when the code is bad.
```
if actualFirstSuccessTime := uCD.Status.FirstSyncSetsSuccessTimestamp;
  assert.NotNil(t, actualFirstSuccessTime, "expected first success timestamp to be set") {
  ...
}
```

PR #1068, opened 2020-07-29
create RBAC for accessing claimed cluster

staebler:
When a ClusterClaim is assigned a cluster, the clusterclaim controller will create a Role and RoleBinding in the namespace
of the ClusterDeployment. The Role will grant full permissions for Hive resources and GET permissions for the admin kubeconfig and password secrets. The RoleBinding will assign the role to the subjects specified in the ClusterClaim.

https://issues.redhat.com/browse/CO-1070

PR #1047, opened 2020-07-15
Enable cluster hibernation

csrwng:
Adds API field to ClusterDeployment (powerState) to either be in
Hibernating or Running state.
Includes initial AWS actuator.

TODO:

- [x] CSR validation code
- [x] Unit tests
- [ ] Additional actuators (GCP, Azure)

PR #1069, opened 2020-07-30
hibernate clusters created for ClusterPool

staebler:
When a cluster is created for a ClusterPool, hibernate the cluster, as the cluster will not be used until it is assigned to a claim. As part of assigning a cluster to a claim, power the cluster back on.

https://issues.redhat.com/browse/CO-1073

### DIFF
--- a/pkg/controller/clusterclaim/clusterclaim_controller.go
+++ b/pkg/controller/clusterclaim/clusterclaim_controller.go
@@ -308,6 +308,7 @@ func (r *ReconcileClusterClaim) reconcileForDeletedCluster(claim *hivev1.Cluster
 func (r *ReconcileClusterClaim) reconcileForNewAssignment(claim *hivev1.ClusterClaim, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (reconcile.Result, error) {
 	logger.Info("cluster assigned to claim")
 	cd.Spec.ClusterPoolRef.ClaimName = claim.Name
+	cd.Spec.PowerState = hivev1.RunningClusterPowerState
 	if err := r.Update(context.Background(), cd); err != nil {
 		logger.WithError(err).Log(controllerutils.LogLevel(err), "could not set claim for ClusterDeployment")
 		return reconcile.Result{}, err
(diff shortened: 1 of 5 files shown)


staebler:
This builds on #1065, #1068, and #1047.

/hold

staebler:
/hold cancel

Dependent PRs have all merged.

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -308,6 +308,7 @@ func (r *ReconcileClusterClaim) reconcileForDeletedCluster(claim *hivev1.Cluster
 func (r *ReconcileClusterClaim) reconcileForNewAssignment(claim *hivev1.ClusterClaim, cd *hivev1.ClusterDeployment, logger log.FieldLogger) (reconcile.Result, error) {
 	logger.Info("cluster assigned to claim")
 	cd.Spec.ClusterPoolRef.ClaimName = claim.Name
+	cd.Spec.PowerState = hivev1.RunningClusterPowerState
```
dgoodwin, line 311: Feels like this should be an option and not the default. Leaving them running would give you immediate response to claims, which might be what some people would want. Hibernation is AWS only right now and probably other clouds will just disregard the setting but arguably validation should be rejecting attempts to create / update powerstate on gcp/azure/etc.

I'd pitch we put this onto the pool spec and make you opt in.

staebler: I could see making this opt out, but I think the default behavior should be to hibernate. In my view, the cluster pools shine brightest with hibernation. If we add validation to block setting the power state on unsupported platforms, then the clusterpool controller can selectively set the power state based on the platform.

Another option would be to have a desired pool size and a desired hot ready size. The hot ready size would be the number of clusters that are ready and powered on. Let's say a pool has a desired size of 10 and a hot ready size of 3. In steady state, the pool would have 7 hibernating clusters and 3 powered-on clusters. When a claim comes in, the cluster pool will create a new cluster and power on one of the hibernating clusters.

PR #1070, opened 2020-07-30
Scale testing improvements

dgoodwin:
Various improvements for scale testing, updated instructions and observations. Details in individual commits.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -107,6 +107,13 @@ func newRootCommand() *cobra.Command {
 				log.WithError(err).Fatal("Cannot parse log level")
 			}
 			log.SetLevel(level)
+
+			// Add some millisecond precision to log timestamps, useful for debugging performance.
+			formatter := new(log.TextFormatter)
+			formatter.TimestampFormat = "2006-01-02T15:04:05.999Z07:00"
+			formatter.FullTimestamp = true
+			log.SetFormatter(formatter)
+
 			log.Infof("Version: %s", version.String())
 			log.Debug("debug logging enabled")
 
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -72,12 +72,12 @@ AWS credentials. These are only relevant for creating a cluster on AWS. If
 --creds-file is used it will take precedence over these environment
 variables.
 
-GOVC_USERNAME and GOVC_PASSWORD - Are used to determine your vSphere 
+GOVC_USERNAME and GOVC_PASSWORD - Are used to determine your vSphere
 credentials.
 GOVC_TLS_CA_CERTS - Is used to provide CA certificates for communicating
 with the vSphere API.
 GOVC_NETWORK, GOVC_DATACENTER, GOVC_DATASTORE and GOVC_HOST (vCenter host)
-can be used as alternatives to the associated commandline argument. 
+can be used as alternatives to the associated commandline argument.
 These are only relevant for creating a cluster on vSphere.
 
 RELEASE_IMAGE - Release image to use to install the cluster. If not specified,
@@ -150,6 +150,8 @@ type Options struct {
 	AdoptAdminPassword       string
 	MachineNetwork           string
 	Region                   string
+	Labels                   []string
+	SkipMachinePools         bool
 
 	// AWS
 	AWSUserTags []string
@@ -226,8 +228,7 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=ovirt --ovirt-api-vip 192.168.1.2
 			}
 			err := opt.Run()
 			if err != nil {
-				log.WithError(err).Error("Error")
-				os.Exit(1)
+				log.WithError(err).Fatal("Error")
 			}
 		},
 	}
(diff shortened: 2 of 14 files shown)


contrib/pkg/createcluster/create.go
```diff
@@ -260,6 +264,8 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=ovirt --ovirt-api-vip 192.168.1.2
 	flags.StringVar(&opt.ManifestsDir, "manifests", "", "Directory containing manifests to add during installation")
 	flags.StringVar(&opt.MachineNetwork, "machine-network", "10.0.0.0/16", "Cluster's MachineNetwork to pass to the installer")
 	flags.StringVar(&opt.Region, "region", "", "Region to which to install the cluster. This is only relevant to AWS, Azure, and GCP.")
+	flags.StringSliceVarP(&opt.Labels, "labels", "l", []string{}, "Label to apply to the ClusterDeployment (key=val)")
```
staebler, line 267:
```suggestion
	flags.StringSliceVarP(&opt.Labels, "labels", "l", nil, "Label to apply to the ClusterDeployment (key=val)")
```

contrib/pkg/createcluster/create.go
```diff
@@ -383,6 +389,14 @@ func (o *Options) Validate(cmd *cobra.Command) error {
 		}
 	}
 
+	if len(o.Labels) > 0 {
```
staebler, line 392: This check is not needed.

hack/scaletest/syncset-template.yaml
```diff
@@ -15,15 +15,139 @@ objects:
...
+        a: asdddaasdjksdhlkajshdkljasdhlaksjdhlaksjdhlaksdhlaksjdhlaksjdhlkasjdhlkajsdhlaksjdhalksdjhalkdsjhlakshdhlkajssdhlkajsdhlaksjdhlkasjdhalksjdhalksjh
+        b: asdhakjsdhlaksdhlaksjsdhlkasjdhlkashdlkasdhlaksjdhlaksdhalskdjhalkdhaksljdhalksjdhaklsjdh
+        c: asjkdhaskjdhalksjdhaklsjdhlkjhllljkkkkkkkkkkkkkkkkkkkkkkhddhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh
+        d: asdiuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu
+    resourceApplyMode: Sync
     clusterDeploymentSelector:
       matchLabels:
-        scaletest: "true"
+        # Added automatically by hiveutil create-cluster, so will match what we create in the adoption script:
+        hive.openshift.io/hiveutil-created: "true"
```
staebler, line 153: I don't understand the point of this change.

pkg/clusterresource/builder.go
```diff
@@ -106,6 +106,9 @@ type Builder struct {
 
 	// MachineNetwork is the subnet to use for the cluster's machine network.
 	MachineNetwork string
+
+	// SkipMachinePools should be true if you do not want to manage MachineSets in-cluster once it is installed.
```
staebler, line 110: The term "in-cluster" is confusing here.
```suggestion
	// SkipMachinePools should be true if you do not want to Hive to manage MachineSets once the cluster is installed.
```

pkg/clusterresource/builder.go
```diff
@@ -157,8 +160,8 @@ func (o *Builder) Build() ([]runtime.Object, error) {
 	var allObjects []runtime.Object
 	allObjects = append(allObjects, o.generateClusterDeployment())
 
-	if mp := o.generateMachinePool(); mp != nil {
-		allObjects = append(allObjects, mp)
+	if !o.SkipMachinePools {
+		allObjects = append(allObjects, o.generateMachinePool())
```
staebler, line 164: Please keep the nil check to be able to support platforms that do not support MachinePools.

pkg/clusterresource/builder.go
```diff
@@ -341,8 +344,8 @@ func (o *Builder) generateInstallConfigSecret() (*corev1.Secret, error) {
 			Namespace: o.Namespace,
 		},
 		Type: corev1.SecretTypeOpaque,
-		Data: map[string][]byte{
-			"install-config.yaml": d,
+		StringData: map[string]string{
```
staebler, line 347: What is the purpose of this change?

cmd/manager/main.go
```diff
@@ -103,6 +103,13 @@ func newRootCommand() *cobra.Command {
 				log.WithError(err).Fatal("Cannot parse log level")
 			}
 			log.SetLevel(level)
+
+			// Add some millisecond precision to log timestamps, useful for debugging performance.
```
staebler, line 107: Do we want the millisecond precision all the time, or just when we are debugging performance?

dgoodwin: It's not critical but I felt it was probably useful and harmless to keep.

pkg/clusterresource/builder.go
dgoodwin: This allows us to create-cluster with -o yaml and the edit the install config without having to base64 decode the embedded secret data.

dgoodwin:
Updated.

cmd/manager/main.go
staebler: Sounds good.

pkg/clusterresource/builder.go
staebler: Wonderful!

PR #1075, opened 2020-08-03
add deleteProtection parameter to hiveconfig in olm template

gregsheremeta:
/cc @staebler

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -18,6 +18,8 @@ parameters:
   value: "[]"
 - name: DISABLED_CONTROLLERS
   value: "[]"
+- name: DELETE_PROTECTION
+  value: ""
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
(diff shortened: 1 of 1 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -59,3 +61,4 @@ objects:
     disabledControllers: "${{DISABLED_CONTROLLERS}}"
     globalPullSecretRef:
       name: ${GLOBAL_PULL_SECRET}
+    deleteProtection: ${ENABLE_DELETION_PROTECTION}
```
staebler, line 64: This will fail validation, as "false" is not a valid value. The only valid non-empty value is "enabled".

gregsheremeta: Ok, can I use empty string? If not, seems like there's no way to use the current pipeline design, then.

staebler: Yes, empty string should be the same as omitted. If it is not, then the kubebuilder validation on the field should be adjusted to allow it.

gregsheremeta: done

PR #1081, opened 2020-08-07
Hibernation: Fix setting unsupported condition when no condition is present

csrwng:
Fixes https://issues.redhat.com/browse/CO-1095


csrwng:
/close 
superseded by #1082

PR #1083, opened 2020-08-10
Enable configuring controllers through HiveConfig

akhil-rane:
This PR enables configuring max concurrent reconciles, client rate
limiter and workqueue rate limiter for each controller through HiveConfig.
Once HiveConfig is edited, an update is made to hive-controllers-config
configmap and hive-controllers pod is restarted. A controller picks up its
latest configurations from hive-controllers-config configmap.

jira: https://issues.redhat.com/browse/CO-1069


pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -35,22 +38,48 @@ const (
 // Add creates a new ClusterClaim Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	return AddToManager(mgr, NewReconciler(mgr))
+	logger := log.WithField("controller", ControllerName)
+	hiveControllersConfig := &corev1.ConfigMap{}
+	if err := mgr.GetAPIReader().Get(context.TODO(), types.NamespacedName{Namespace: controllerutils.GetHiveNamespace(), Name: hive.HiveControllersConfigMapName}, hiveControllersConfig); err != nil {
```
dgoodwin, line 43: Had never seen this GetAPIReader before, looked it up and the comments indicate:

```
	// GetAPIReader returns a reader that will be configured to use the API server.
	// This should be used sparingly and only when the client does not fit your
	// use case.
```

Is there a reason we need this here vs just using the mgr.GetClient? Is it not initialized yet? If so we should add a comment.

Was mounting the configmap into the pod considered?

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -194,6 +214,67 @@ type ManageDNSAzureConfig struct {
...
+
+// ControllerConfig contains the configuration for a particular controller
+type ControllerConfig struct {
+	// name specifies the name of the controller
+	// +kubebuilder:validation:Enum=clusterdeployment;clusterrelocate;clusterstate;clusterversion;controlPlaneCerts;dnsendpoint;dnszone;remoteingress;remotemachineset;syncidentityprovider;syncset;syncsetinstance;unreachable;velerobackup;clusterprovision;clusterDeprovision;clusterpool;clusterpoolnamespace
+	// +optional
+	Name string `json:"name,omitempty"`
+	// value specifies the value of the configuration
+	// +optional
+	Value *int `json:"value,omitempty"`
```
dgoodwin, line 275: I think both name and value here should not be optional. If you're using this struct, both fields seem to be required.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -145,14 +148,35 @@ func init() {
 
 // Add creates a new ClusterDeployment controller and adds it to the manager with default RBAC.
 func Add(mgr manager.Manager) error {
-	return AddToManager(mgr, NewReconciler(mgr))
+	logger := log.WithField("controller", ControllerName)
+	hiveControllersConfig := &corev1.ConfigMap{}
+	if err := mgr.GetAPIReader().Get(context.TODO(), types.NamespacedName{Namespace: controllerutils.GetHiveNamespace(), Name: hive.HiveControllersConfigMapName}, hiveControllersConfig); err != nil {
```
dgoodwin, line 153: Because this repeats in each controller, I think it might still be good to push more down into the controllerutils function. Something like controllerutils.GetControllerConfig taking an apireder and the controller name, and returning all fields and an error. Would reduce a good 20 lines of boilerplate in each.

pkg/controller/utils/utils.go
```diff
@@ -14,13 +17,23 @@ import (
...
 	apihelpers "github.com/openshift/hive/pkg/apis/helpers"
 	"github.com/openshift/hive/pkg/constants"
 )
 
+const (
+	// defaultQueueQPS is the default workqueue qps same as used in DefaultControllerRateLimiter
+	defaultQueueQPS = 10
+	// defaultQueueBurst is the default workqueue burst same as used in DefaultControllerRateLimiter
+	defaultQueueBurst = 100
+)
```
dgoodwin, line 35: :+1:

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -194,6 +214,67 @@ type ManageDNSAzureConfig struct {
 	ResourceGroupName string `json:"resourceGroupName"`
 }
 
+// ControllerGoroutinesConfig contains number of goroutines for each controller
+type ControllerGoroutinesConfig struct {
+	// Default specifies default number of goroutines for each Controller
+	// +optional
+	Default *int `json:"default,omitempty"`
```
staebler, line 221: All `int` fields in the API should use `int32` instead.
```suggestion
	Default *int32 `json:"default,omitempty"`
```

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -194,6 +214,67 @@ type ManageDNSAzureConfig struct {
 	ResourceGroupName string `json:"resourceGroupName"`
 }
 
+// ControllerGoroutinesConfig contains number of goroutines for each controller
+type ControllerGoroutinesConfig struct {
```
staebler, line 218: Why are there different types for each config (e.g., `ControllerGoroutinesConfig` and `ControllerClientQPSConfig`) but not different types for each controller config (i.e., only one `ControllerConfig`)? Either everything can use the same type, or everything should use a different type.

Would it be any easier for a code standpoint and from a UX standpoint if these controller configurations were structured like the following?
```
type ControllerConfig struct {
  Gorotines *int32
  ClientQPS *int32
  ClientBurst *int32
  QueueQPS *int32
  QueueBurst *int32
}

type SpecificControllerConfig struct {
  Name string
  ControllerConfig
}

type ControllersConfig struct {
  Default *ControllerConfig
  Controllers []ControllerConfig
}

type HiveConfigSpec struct {
  ...
  ControllersConfig *ControllerConfig
}
```

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -78,6 +78,26 @@ type HiveConfigSpec struct {
 	// DisabledControllers allows selectively disabling Hive controllers by name.
 	// The name of an individual controller matches the name of the controller as seen in the Hive logging output.
 	DisabledControllers []string `json:"disabledControllers,omitempty"`
+
+	// controllerGoroutines is used to configure the number of goroutines for each controller
+	// +optional
+	ControllerGoroutines *ControllerGoroutinesConfig `json:"controllerGoroutines,omitempty"`
```
staebler, line 84: Can we use "ConcurrentReconciles" here instead of "Goroutines"? The former conveys a concept that the user may care about whereas the latter is an implementation detail.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -49,8 +50,29 @@ var controllerKind = hivev1.SchemeGroupVersion.WithKind("MachinePool")
...
 func Add(mgr manager.Manager) error {
 	logger := log.WithField("controller", ControllerName)
+	hiveControllersConfig := &corev1.ConfigMap{}
+	if err := mgr.GetAPIReader().Get(context.TODO(), types.NamespacedName{Namespace: controllerutils.GetHiveNamespace(), Name: hive.HiveControllersConfigMapName}, hiveControllersConfig); err != nil {
+		log.WithError(err).Errorf("could not get %s configmap", hive.HiveControllersConfigMapName)
+		return err
+	}
+	concurrentReconciles, err := controllerutils.GetConcurrentReconciles(ControllerName, hiveControllersConfig.Data)
+	if err != nil {
+		logger.Error("could not get concurrent reconciles")
```
staebler, line 60:
```suggestion
		logger.WithError(err).Error("could not get concurrent reconciles")
```

pkg/controller/utils/utils.go
```diff
@@ -50,10 +63,92 @@ const (
 )
 
 // GetConcurrentReconciles returns the number of goroutines each controller should
-// use for parallel processing of their queue. For now this is a static value of 5.
-// In future this may be read from an env var set by the operator, and driven by HiveConfig.
-func GetConcurrentReconciles() int {
-	return concurrentControllerReconciles
+// use for parallel processing of their queue. Default value, if not set in
+// hive-conntrollers-config, will be 5.
```
staebler, line 67:
```suggestion
// hive-controllers-config, will be 5.
```

pkg/operator/hive/hive.go
```diff
@@ -4,22 +4,23 @@ import (
...
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
 	hiveconstants "github.com/openshift/hive/pkg/constants"
 	"github.com/openshift/hive/pkg/controller/images"
+	"github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/operator/assets"
 	"github.com/openshift/hive/pkg/operator/util"
 	"github.com/openshift/hive/pkg/resource"
-
-	oappsv1 "github.com/openshift/api/apps/v1"
```
staebler, line 22: Don't move this. The non-hive "github.com/openshift" imports should be in their own section. The hive imports should be last, but the imports in this file are out-of-order.

pkg/operator/hive/hive.go
```diff
@@ -152,6 +161,96 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+	hiveControllersConfigMap := &corev1.ConfigMap{}
+	switch err := r.Get(context.TODO(), types.NamespacedName{Namespace: utils.GetHiveNamespace(), Name: HiveControllersConfigMapName}, hiveControllersConfigMap); {
+	case apierrors.IsNotFound(err):
+		hLog.Infof("configmap for hive controllers not found, creating a new one")
+		newHiveControllersConfigMap := &corev1.ConfigMap{}
+		newHiveControllersConfigMap.Name = HiveControllersConfigMapName
+		newHiveControllersConfigMap.Namespace = utils.GetHiveNamespace()
+		newHiveControllersConfigMap.Data = make(map[string]string)
+		hiveControllersConfigMap = newHiveControllersConfigMap
+		if err := r.Create(context.TODO(), hiveControllersConfigMap); err != nil {
```
staebler, line 173: Do not create it here only to update it later. Do the create after the configmap has been filled out.

Consider using `util.ApplyRuntimeObjectWithGC`, since we also want an owner reference on the configmap.

pkg/operator/hive/hive.go
```diff
@@ -152,6 +161,96 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h *resource.Helpe
...
+	}
+	hLog.Infof("updating the %s configmap", HiveControllersConfigMapName)
+	if err := r.Update(context.TODO(), hiveControllersConfigMap); err != nil {
+		hLog.WithError(err).Error("error updating configmap for hive controllers")
+		return err
+	}
+	hLog.Info("Hashing hive-controllers-config data onto a hive deployment annotation")
+	hiveControllersConfigHash := computeHiveControllersConfigHash(hiveControllersConfigMap.Data)
+	if hiveDeployment.Spec.Template.Annotations == nil {
+		hiveDeployment.Spec.Template.Annotations = make(map[string]string)
```
staebler, line 250:
```suggestion
		hiveDeployment.Spec.Template.Annotations = make(map[string]string, 1)
```

pkg/controller/utils/utils.go
```diff
@@ -50,10 +63,92 @@ const (
...
+	}
+	return concurrentControllerReconciles, nil
+}
+
+// GetClientRateLimiter returns the client rate limiter for the controller
+func GetClientRateLimiter(controllerName string, configData map[string]string) (flowcontrol.RateLimiter, error) {
+	qps := rest.DefaultQPS
+	var err error
+	if qpsStr, ok := configData[fmt.Sprintf("%s-client-qps", controllerName)]; ok {
+		qpsFloat64, err := strconv.ParseFloat(qpsStr, 32)
```
staebler, line 90: This is parsed as a float, but it was written as an int. Should this be a float in HiveConfig?

pkg/apis/hive/v1/hiveconfig_types.go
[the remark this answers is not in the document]
akhil-rane: I restructured the code here as per @staebler's suggestions

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -194,6 +214,67 @@ type ManageDNSAzureConfig struct {
...
+	// +optional
+	Default *int `json:"default,omitempty"`
+	// ControllerConfigs is a list of goroutines configurations for each controller
+	// +optional
+	ControllerConfigs []ControllerConfig `json:"configs,omitempty"`
+}
+
+// ControllerClientQPSConfig contains client rate limiter QPS for each controller
+type ControllerClientQPSConfig struct {
+	// ControllerClientQPSConfig contains allowed client rate limiter QPS for each controller
```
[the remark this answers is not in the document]
akhil-rane: Done

pkg/controller/clusterclaim/clusterclaim_controller.go
akhil-rane: My bad I thought GetAPIReader is the only way to interact with the API server. Using GetClient now.

pkg/apis/hive/v1/hiveconfig_types.go
akhil-rane: This looks way better than what I had implemented. Thanx!

pkg/controller/utils/utils.go
akhil-rane: float is not supported. check [this](https://github.com/kubernetes-sigs/controller-tools/issues/245). Besides, is there any specific use case where we would like to set qps as float?

staebler: Yeah, right, you cannot do a float in the API. But this should still not be trying to parse as a float.

akhil-rane: Agreed, but I still need to convert it into a float as `rest.DefaultQPS ` is a float.

staebler: Yep.

pkg/operator/hive/hive.go
```diff
@@ -333,3 +373,27 @@ func (r *ReconcileHiveConfig) runningOnOpenShift(hLog log.FieldLogger) (bool, er
...
+
+func computeHiveControllersConfigHash(data map[string]string) string {
+	hasher := md5.New()
+	hasher.Write([]byte(fmt.Sprintf("%v", data)))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
+
+func setHiveControllersConfig(config *hivev1.ControllerConfig, hiveControllersConfigMap *corev1.ConfigMap, controllerName string) {
+	if config.ConcurrentReconciles != nil {
+		hiveControllersConfigMap.Data[fmt.Sprintf("%s-concurrent-reconciles", controllerName)] = strconv.Itoa(int(*config.ConcurrentReconciles))
```
staebler, line 385: Move the formats into constants so that they can be shared by the controllers code when reading the data.

pkg/operator/hive/hive.go
```diff
@@ -333,3 +373,27 @@ func (r *ReconcileHiveConfig) runningOnOpenShift(hLog log.FieldLogger) (bool, er
...
 	return len(list.APIResources) > 0, nil
 }
+
+func computeHiveControllersConfigHash(data map[string]string) string {
+	hasher := md5.New()
+	hasher.Write([]byte(fmt.Sprintf("%v", data)))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
+
+func setHiveControllersConfig(config *hivev1.ControllerConfig, hiveControllersConfigMap *corev1.ConfigMap, controllerName string) {
```
staebler, line 383: Let's be consistent between `computeHiveControllersConfigHash` and `setHiveControllersConfig` with whether we are passing in just the data map or the full ConfigMap.

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -194,6 +198,49 @@ type ManageDNSAzureConfig struct {
...
+	QueueBurst *int32 `json:"queueBurst,omitempty"`
+}
+
+// SpecificControllerConfig contains the configuration for a specific controller
+type SpecificControllerConfig struct {
+	// Name specifies the name of the controller
+	// +kubebuilder:validation:Enum=clusterdeployment;clusterrelocate;clusterstate;clusterversion;controlPlaneCerts;dnsendpoint;dnszone;remoteingress;remotemachineset;syncidentityprovider;syncset;syncsetinstance;unreachable;velerobackup;clusterprovision;clusterDeprovision;clusterpool;clusterpoolnamespace;hibernation;clusterclaim
+	Name string `json:"name,omitempty"`
+	// ControllerConfig contains the configuration for the controller specified by Name field
+	ControllerConfig *ControllerConfig `json:"config,omitempty"`
```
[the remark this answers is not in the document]
akhil-rane: I believe I have not set this as optional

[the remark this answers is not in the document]
staebler: The omitepty implies optional. Remove the omitepty, and don't make the field a pointer.

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -194,6 +198,49 @@ type ManageDNSAzureConfig struct {
...
+	QueueQPS *int32 `json:"queueQPS,omitempty"`
+	// QueueBurst specifies workqueue rate limiter burst for a controller
+	// +optional
+	QueueBurst *int32 `json:"queueBurst,omitempty"`
+}
+
+// SpecificControllerConfig contains the configuration for a specific controller
+type SpecificControllerConfig struct {
+	// Name specifies the name of the controller
+	// +kubebuilder:validation:Enum=clusterdeployment;clusterrelocate;clusterstate;clusterversion;controlPlaneCerts;dnsendpoint;dnszone;remoteingress;remotemachineset;syncidentityprovider;syncset;syncsetinstance;unreachable;velerobackup;clusterprovision;clusterDeprovision;clusterpool;clusterpoolnamespace;hibernation;clusterclaim
```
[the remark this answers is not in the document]
akhil-rane: I spent sometime on this but looks like enums are not supported in Go. Can we move all the controller name constants in one place and put a comment there to add any changes to this validation?

pkg/controller/utils/utils.go
```diff
@@ -49,11 +67,114 @@ const (
...
+
+	return workqueue.NewMaxOfRateLimiter(
+		workqueue.NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
+		&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(qps), burst)},
+	), nil
+}
+
+func GetControllerConfig(client client.Client, controllerName string) (int, flowcontrol.RateLimiter, workqueue.RateLimiter, error) {
+	hiveControllersConfig := &corev1.ConfigMap{}
+	if err := client.Get(context.TODO(), types.NamespacedName{Namespace: GetHiveNamespace(), Name: hiveControllersConfigMapName}, hiveControllersConfig); err != nil {
```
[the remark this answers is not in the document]
akhil-rane: Can we instead go for container environment variables similar to [this](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables)?

pkg/apis/hive/v1/hiveconfig_types.go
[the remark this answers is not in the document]
staebler: Sorry, by a go enum I mean something like the following.
```
type ControllerNames string

const (
    ClusterDeploymentControllerName ControllerNames = "clusterdeployment"
    DNSZoneControllerName ControllerNames = "dnszone"
)
```

pkg/controller/utils/utils.go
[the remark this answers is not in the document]
staebler: Yep, that's fine.

pkg/apis/hive/v1/hiveconfig_types.go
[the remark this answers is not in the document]
akhil-rane: What is the advantage of having` type ControllerNames string` here? Why can't we have something like following
```
WARNING: Add following values to validation in .. file
const (
    ClusterDeploymentControllerName = "clusterdeployment"
    DNSZoneControllerName = "dnszone"
)
```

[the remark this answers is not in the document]
staebler: If it is a go enum, then you can use that type directly in the `SpecificControllerConfig` type. Then, you can put the validation on the go enum type itself. That keeps things the validation located very close to the values.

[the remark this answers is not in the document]
staebler:
```
type SpecificControllerConfig struct {
	Name ControllerName `json:"name,omitempty"`
```
```
// +kubebuilder:validation:Enum=clusterdeployment;dnszone
type ControllerName string

const (
    ClusterDeploymentControllerName ControllerName = "clusterdeployment"
    DNSZoneControllerName ControllerName = "dnszone"
)
```

akhil-rane:
@staebler this is ready for another round of review

pkg/controller/utils/utils_test.go
```diff
@@ -111,3 +123,213 @@ func TestLogLevel(t *testing.T) {
...
+			if tc.expectedError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equalf(t, tc.expectedConcurrentReconciles, concurrentReconciles, "unexpected concurrent reconciles")
+			// unset environment variables
+			for k := range tc.environmentVariables {
+				os.Unsetenv(k)
+			}
```
staebler, line 186: This should be in a defer after setting the environment variables needed for the test case.

cmd/manager/main.go
```diff
@@ -221,7 +222,8 @@ func newRootCommand() *cobra.Command {
 
 	cmd.PersistentFlags().StringVar(&opts.LogLevel, "log-level", defaultLogLevel, "Log level (debug,info,warn,error,fatal)")
 	cmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
-	cmd.PersistentFlags().StringSliceVar(&opts.Controllers, "controllers", opts.Controllers, "Comma-separated list of controllers to run")
+	controllersAsStringSliceVar := getControllersAsStringSliceVar(opts.Controllers)
+	cmd.PersistentFlags().StringSliceVar(&controllersAsStringSliceVar, "controllers", controllersAsStringSliceVar, "Comma-separated list of controllers to run")
```
staebler, line 226: The `controllersAsStringSliceVar` variable is set by the command-line flag, but it is never used.

cmd/manager/main.go
```diff
@@ -268,3 +270,11 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+func getControllersAsStringSliceVar(controllerNames []hivev1.ControllerName) []string {
+	controllersAsStringSliceVar := make([]string, 0, len(controllerNames))
+	for _, controllerName := range controllerNames {
+		controllersAsStringSliceVar = append(controllersAsStringSliceVar, controllerName.String())
+	}
```
staebler, line 278:
```suggestion
	controllersAsStringSliceVar := make([]string, len(controllerNames))
	for i, controllerName := range controllerNames {
		controllersAsStringSliceVar[i] = controllerName.String()
	}
```

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -195,6 +199,79 @@ type ManageDNSAzureConfig struct {
...
+type ControllerName string
+
+func (controllerName ControllerName) String() string {
+	return string(controllerName)
+}
+
+// WARNING: All the controller names below should also be added to the kubebuilder validation of the type ControllerName
+const (
+	ClusterClaimControllerName         ControllerName = "clusterclaim"
+	ClusterDeploymentControllerName    ControllerName = "clusterdeployment"
```
staebler, line 231: The name of this controller was "clusterDeployment".

pkg/controller/utils/utils.go
```diff
@@ -45,15 +80,82 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
...
 
-// GetConcurrentReconciles returns the number of goroutines each controller should
-// use for parallel processing of their queue. For now this is a static value of 5.
-// In future this may be read from an env var set by the operator, and driven by HiveConfig.
-func GetConcurrentReconciles() int {
-	return concurrentControllerReconciles
+	burst := rest.DefaultBurst
+	var err error
+	if value, ok := getValueFromEnvVariable(controllerName, ClientBurstEnvVariableFormat); ok {
+		burst, err = strconv.Atoi(value)
```
staebler, line 111:
```suggestion
	if value, ok := getValueFromEnvVariable(controllerName, ClientBurstEnvVariableFormat); ok {
		var err error
		burst, err = strconv.Atoi(value)
```

pkg/operator/hive/hive.go
```diff
@@ -162,6 +173,34 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
...
+			setHiveControllersConfig(&controller.Config, hiveControllersConfigMap, controller.Name)
+		}
+	}
+
+	result, err := util.ApplyRuntimeObjectWithGC(h, hiveControllersConfigMap, instance)
+	if err != nil {
+		hLog.WithError(err).Error("error applying hive-controllers-config configmap")
+		return err
+	}
+	hLog.Infof("hive-controllers-config configmap applied (%s)", result)
```
staebler, line 195:
```suggestion
	hLog.WithField("result", result).Info("hive-controllers-config configmap applied")
```

pkg/client/clientset/versioned/fake/register.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var scheme = runtime.NewScheme()
 var codecs = serializer.NewCodecFactory(scheme)
-var parameterCodec = runtime.NewParameterCodec(scheme)
```
staebler, line 17: These changes were missed as part of upgrading dependencies. Let's put them in a separate PR, or at least a separate commit in this PR. I'll put up a PR for them.

staebler: https://github.com/openshift/hive/pull/1125

cmd/manager/main.go
akhil-rane: Should it be removed?

staebler: The `--controllers` flag needs to be set somewhere. The rest of the code assumes that it will be set in `opts.Controllers`. The easiest thing is probably to keep `opts.Controllers` as a `[]string` rather than changing it to a `[]ControllerName`. Then, the `--controllers` flag can still drive that variable directly. Otherwise, there will need to be some other code to take the slice filled out by the `--controllers` flag and set `opts.Controllers` appropriately.

pkg/client/clientset/versioned/fake/register.go
akhil-rane: Removed these changes, thanks

akhil-rane:
PR updated.

staebler (approved):
/lgtm

dgoodwin:
Looks like an actual e2e test failure here, hive-operator pod getting errors on the hive namespace not existing?

pkg/operator/hive/hive.go
```diff
@@ -162,6 +173,34 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 		hiveDeployment.Spec.Replicas = &replicas
 	}
 
+	hiveControllersConfigMap := &corev1.ConfigMap{}
+	hiveControllersConfigMap.Name = hiveControllersConfigMapName
+	hiveControllersConfigMap.Namespace = utils.GetHiveNamespace()
```
staebler, line 178:
```suggestion
	hiveControllersConfigMap.Namespace = hiveNSName
```

staebler (commented):
/lgtm cancel

As alluded to by @dgoodwin, the new configmap is getting placed in the wrong namespace.

staebler (approved):
/lgtm

PR #1084, opened 2020-08-10
Implement hibernation for GCP

csrwng:
Adds a GCP actuator to the hibernation controller

### DIFF
--- a/pkg/controller/hibernation/gcp_actuator.go
+++ b/pkg/controller/hibernation/gcp_actuator.go
@@ -1,0 +1,162 @@
+package hibernation
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	compute "google.golang.org/api/compute/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	"github.com/openshift/hive/pkg/gcpclient"
+)
+
+const (
+	instanceFields = "items/*/instances(name,zone,status),nextPageToken"
+)
+
+var (
+	gcpRunningStatuses           = sets.NewString("RUNNING")
+	gcpStoppedStatuses           = sets.NewString("STOPPED", "TERMINATED")
+	gcpPendingStatuses           = sets.NewString("PROVISIONING", "STAGING")
+	gcpStoppingStatuses          = sets.NewString("STOPPING")
+	gcpRunningOrPendingStatuses  = gcpRunningStatuses.Union(gcpPendingStatuses)
+	gcpStoppedOrStoppingStatuses = gcpStoppedStatuses.Union(gcpStoppingStatuses)
+	gcpNotRunningStatuses        = gcpStoppedOrStoppingStatuses.Union(gcpPendingStatuses)
+	gcpNotStoppedStatuses        = gcpRunningOrPendingStatuses.Union(gcpStoppingStatuses)
+)
+
+func init() {
+	RegisterActuator(&gcpActuator{getGCPClientFn: getGCPClient})
+}
+
+type gcpActuator struct {
+	getGCPClientFn func(*hivev1.ClusterDeployment, client.Client, log.FieldLogger) (gcpclient.Client, error)
+}
+
+// CanHandle returns true if the actuator can handle a particular ClusterDeployment
+func (a *gcpActuator) CanHandle(cd *hivev1.ClusterDeployment) bool {
+	return cd.Spec.Platform.GCP != nil
+}
+
+// StopMachines will start machines belonging to the given ClusterDeployment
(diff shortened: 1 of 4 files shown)


pkg/gcpclient/client.go
```diff
@@ -217,6 +231,46 @@ func (c *gcpClient) ListComputeImages(opts ListComputeImagesOptions) (*compute.I
 	}
 	return call.Do()
 }
+func (c *gcpClient) ListComputeInstances(opts ListComputeInstancesOptions) ([]*compute.Instance, error) {
```
staebler, line 234:
```suggestion

func (c *gcpClient) ListComputeInstances(opts ListComputeInstancesOptions) ([]*compute.Instance, error) {
```

pkg/gcpclient/client.go
```diff
@@ -217,6 +231,46 @@ func (c *gcpClient) ListComputeImages(opts ListComputeImagesOptions) (*compute.I
...
+		req = req.Filter(opts.Filter)
+	}
+	err := req.Pages(context.TODO(), func(list *compute.InstanceAggregatedList) error {
+		for _, scopedList := range list.Items {
+			for _, item := range scopedList.Instances {
+				result = append(result, item)
+			}
+		}
+		return nil
+	})
```
staebler, line 250: I am uncomfortable with loading all of the instances into memory at once. Can we deal with each page separately instead? I don't think the size is a problem for the current use where there is a filter and only a few fields returned. However, the `ListComputeInstances` method is generic enough that I could see it used somewhere else without giving thought to the memory impact.

pkg/controller/hibernation/gcp_actuator.go
```diff
@@ -0,0 +1,168 @@
...
+	instances, err := gcpClient.ListComputeInstances(gcpclient.ListComputeInstancesOptions{
+		Filter: instanceFilter(cd),
+		Fields: instanceFields,
+	})
+	if err != nil {
+		logger.WithError(err).Log(controllerutils.LogLevel(err), "Failed to fetch instances")
+		return errors.Wrap(err, "failed to fetch instances")
+	}
+	instances = filterByStatus(instances, gcpRunningOrPendingStatuses)
+	errs := []error{}
```
staebler, line 65: To save an allocation when there are no errors.
```suggestion
	var errs []error
```

pkg/controller/hibernation/gcp_actuator.go
```diff
@@ -0,0 +1,168 @@
...
+func (a *gcpActuator) StopMachines(cd *hivev1.ClusterDeployment, hiveClient client.Client, logger log.FieldLogger) error {
+	logger = logger.WithField("cloud", "GCP")
+	gcpClient, err := a.getGCPClientFn(cd, hiveClient, logger)
+	if err != nil {
+		return err
+	}
+	instances, err := gcpClient.ListComputeInstances(gcpclient.ListComputeInstancesOptions{
+		Filter: instanceFilter(cd),
+		Fields: instanceFields,
+	})
```
staebler, line 59: This is duplicated elsewhere. Can we extract this into a function?

```
instances, err := getComputeInstances(gcpClient, cd)
```

pkg/controller/hibernation/gcp_actuator.go
```diff
@@ -0,0 +1,168 @@
...
+
+// StopMachines will start machines belonging to the given ClusterDeployment
+func (a *gcpActuator) StopMachines(cd *hivev1.ClusterDeployment, hiveClient client.Client, logger log.FieldLogger) error {
+	logger = logger.WithField("cloud", "GCP")
+	gcpClient, err := a.getGCPClientFn(cd, hiveClient, logger)
+	if err != nil {
+		return err
+	}
+	instances, err := gcpClient.ListComputeInstances(gcpclient.ListComputeInstancesOptions{
+		Filter: instanceFilter(cd),
```
staebler, line 57: Is it possible to filter on the status on the server side instead?
`(status = "RUNNING") OR (status = "PROVISIONING") OR (status = "STAGING")`

csrwng: It isn't, GCP is real finicky about the filter you give it for instances... which is why in the uninstaller code we ended up doing a client-side filter as well: https://github.com/openshift/installer/blob/077850081b131c098a65b067a063fe8feb8e5d25/pkg/destroy/gcp/instance.go#L44

pkg/gcpclient/client.go
csrwng: Sure, we can also pass a function to handle each  compute.InstanceAggregatedList result instead of returning anything

pkg/controller/hibernation/gcp_actuator.go
csrwng: 👍

csrwng:
@staebler thx for the review, comments addressed

pkg/controller/hibernation/gcp_actuator.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, errors.Wrap(err, "failed to fetch GCP credentials secret")
+	}
+	return gcpclient.NewClientFromSecret(secret)
+}
+
+func instanceFilter(cd *hivev1.ClusterDeployment) string {
+	return fmt.Sprintf("name eq \"%s-.*\"", cd.Spec.ClusterMetadata.InfraID)
+}
+
+func filterByStatus(instances []*compute.Instance, statuses sets.String) []*compute.Instance {
```
staebler, line 140: This function is not used.

csrwng: thx! addressed

csrwng:
had to re-run generate

staebler (commented):
/lgtm

PR #1086, opened 2020-08-12
Add velero namespace config variable

twiest:
- [x] Add velero namespace config variable
- [x] Add unit tests
- [x] Manually test new option


pkg/test/manager/fakemanager.go
```diff
@@ -0,0 +1,102 @@
+package manager
```
csrwng, line 1: Could this have been generated by gomock?

twiest:
@csrwng I believe I've addressed your comments. PTAL.

pkg/controller/velerobackup/velerobackup_controller.go
```diff
@@ -223,14 +237,18 @@ func (r *ReconcileBackup) createVeleroBackupObject(namespace string, t metav1.Ti
...
+			Namespace: r.veleroNamespace,
 		},
 		Spec: velerov1.BackupSpec{
+			VolumeSnapshotLocations: []string{},
 			IncludedNamespaces: []string{
 				namespace,
 			},
 			ExcludedResources: defaultExcludedBackupResources,
 		},
+		Status: velerov1.BackupStatus{
```
staebler, line 249: What is the purpose of including the `Status`? For most resource types, the create of the resource will clear out the status.

twiest: Yeah, this was something new. So apparently the most recent version of Velero has a validation hook that errors without this. I discovered this during my manual tests and adding this line made it work.

PR #1088, opened 2020-08-16
expose hiveAPIEnabled in hiveconfig template

gregsheremeta:
Expose hiveAPIEnabled in hiveconfig template so it can be selectively
enabled and disabled.

/cc @staebler

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -20,6 +20,8 @@ parameters:
   value: "[]"
 - name: DELETE_PROTECTION
   value: ""
+- name: HIVE_API_ENABLED
+  value: true
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
@@ -62,3 +64,4 @@ objects:
     globalPullSecretRef:
       name: ${GLOBAL_PULL_SECRET}
     deleteProtection: ${DELETE_PROTECTION}
+    hiveAPIEnabled: ${HIVE_API_ENABLED}


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -62,3 +64,4 @@ objects:
     globalPullSecretRef:
       name: ${GLOBAL_PULL_SECRET}
     deleteProtection: ${DELETE_PROTECTION}
+    hiveAPIEnabled: ${HIVE_API_ENABLED}
```
staebler, line 67: How is the `hiveAPIEnabled` flag getting set currently?

gregsheremeta: As it's deployed to OSD currently, I honestly have no idea. I think someone added it manually.

See my gitlab patch that will set it to false in stage/int.

staebler: I saw the gitlab patch, which is part of what prompted my question. If there is already some mechanism to set `hiveAPIEnabled` to true, then (1) why do we need these changes and (2) will these changes override or be suppressed by the existing mechanism?

gregsheremeta: that gitlab patch to the saas file basically runs `oc template` on this here file. So in order for us to control this config value via that saas file, it needs to be exposed here. So yes the values for the variable in that file override the default here.

gregsheremeta: (I'm at the point where one repo chasing the other is pretty annoying. I kind of want to have the entire desired hiveconfig exists without these parameters in saas. But until then, we have to do this)

gregsheremeta:
/hold
i don't want this to merge overnight, sigh

gregsheremeta:
manually merging after discussion with @staebler

PR #1089, opened 2020-08-17
Hibernation: azure actuator

csrwng:
Adds Azure actuator to the hibernation controller


csrwng:
Updated to use the ListAll call with statusOnly=true for machine status

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+	azureClient, err := a.azureClientFn(cd, c, logger)
+	if err != nil {
+		return err
+	}
+	machines, err := listAzureMachines(cd, azureClient, azureRunningOrPendingStates, logger)
+	if err != nil {
+		return err
+	}
+	if len(machines) == 0 {
+		logger.Warning("No machines were found to stop")
```
staebler, line 65: Does this deserve to be a warning instead of an info? This could happen if there are machines that are in the Stopping state.

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+	azureClient, err := a.azureClientFn(cd, c, logger)
+	if err != nil {
+		return err
+	}
+	machines, err := listAzureMachines(cd, azureClient, azureStoppedOrStoppingStates, logger)
+	if err != nil {
+		return err
+	}
+	if len(machines) == 0 {
+		logger.Warning("No machines were found to start")
```
staebler, line 92: Same question here about whether this should be info level instead of warning. This would happen if there are machines in the Pending state.

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil
+	}
+	errs := []error{}
```
staebler, line 95:
```suggestion
	var errs []error
```

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	machines, err := listAzureMachines(cd, azureClient, azureRunningOrPendingStates, logger)
+	if err != nil {
+		return err
+	}
+	if len(machines) == 0 {
+		logger.Warning("No machines were found to stop")
+		return nil
+	}
+	errs := []error{}
```
staebler, line 68:
```suggestion
	var errs []error
```

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+		return false, err
+	}
+	machines, err := listAzureMachines(cd, azureClient, azureNotStoppedStates, logger)
+	if err != nil {
+		return false, err
+	}
+	return len(machines) == 0, nil
+}
+
+type azureMachineLister struct {
```
staebler, line 137: This is not used.

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+		result = append(result, filterByResourceGroupAndState(page.Values(), clusterDeploymentResourceGroup(cd), states, logger)...)
+		if err = page.Next(); err != nil {
+			return nil, err
+		}
+	}
+	return result, nil
+}
+
+func filterByResourceGroupAndState(machines []compute.VirtualMachine, resourceGroup string, states sets.String, logger log.FieldLogger) []compute.VirtualMachine {
+	result := []compute.VirtualMachine{}
```
staebler, line 162:
```suggestion
	var result []compute.VirtualMachine
```

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+		if vm.ID == nil {
+			continue
+		}
+		resource, err := azure.ParseResourceID(*vm.ID)
+		if err != nil {
+			logger.WithError(err).Warningf("Failed to parse resource ID")
+			continue
+		}
+		logger.Infof("Parsed resource: %#v", resource)
+		if strings.ToUpper(resource.ResourceGroup) != strings.ToUpper(resourceGroup) {
```
staebler, line 173:
```suggestion
		if strings.EqualFold(resource.ResourceGroup, resourceGroup) {
```

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+	for _, vm := range machines {
+		if vm.ID == nil {
+			continue
+		}
+		resource, err := azure.ParseResourceID(*vm.ID)
+		if err != nil {
+			logger.WithError(err).Warningf("Failed to parse resource ID")
+			continue
+		}
+		logger.Infof("Parsed resource: %#v", resource)
```
staebler, line 172: This should be debug level (or removed entirely). It outputs a log entry for every machine in the account every time that the machines are listed.

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+		if strings.ToUpper(resource.ResourceGroup) != strings.ToUpper(resourceGroup) {
+			continue
+		}
+		state := azureMachinePowerState(vm)
+		logger.Infof("Machine power state: %s", state)
```
staebler, line 177: debug level

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+		if !states.Has(state) {
+			continue
+		}
```
staebler, line 180: Add more context to the log entry.
```suggestion
		logger := logger.WithField("machine", vm.Name).WithField("powerState", state)
		if !states.Has(state) {
			logger.Debug("machine filtered out due to power state"
			continue
		}
		logger.Debug("machine included")
```

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return azureUnknownPowerState
+}
+
+func clusterDeploymentResourceGroup(cd *hivev1.ClusterDeployment) string {
+	if cd.Spec.ClusterMetadata == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s-rg", cd.Spec.ClusterMetadata.InfraID)
```
staebler, line 202: What is your guess on which OpenShift version changes the name of the resource group and breaks us?

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func azureMachineNames(machines []compute.VirtualMachine) []string {
+	result := []string{}
```
staebler, line 206:
```suggestion
	var result []string
```

pkg/azureclient/client.go
```diff
@@ -31,6 +31,13 @@ type Client interface {
 	ListRecordSetsByZone(ctx context.Context, resourceGroupName string, zone string, suffix string) (RecordSetPage, error)
 	CreateOrUpdateRecordSet(ctx context.Context, resourceGroupName string, zone string, recordSetName string, recordType dns.RecordType, recordSet dns.RecordSet) (dns.RecordSet, error)
 	DeleteRecordSet(ctx context.Context, resourceGroupName string, zone string, recordSetName string, recordType dns.RecordType) error
+
+	// Virtual Machines
+	ListAllVirtualMachines(ctx context.Context, statusOnly string) (compute.VirtualMachineListResultPage, error)
+	ListVirtualMachines(ctx context.Context, resourceGroupName string) (compute.VirtualMachineListResultPage, error)
+	VirtualMachineInstanceView(ctx context.Context, resourceGroupName string, machineName string) (compute.VirtualMachineInstanceView, error)
```
staebler, line 38: These two functions are not used.

pkg/controller/hibernation/azure_actuator.go
csrwng: The StopMachines call is currently only done when starting to Hibernate. I would think that if you try to hibernate and there's nothing to stop, that's likely not a good thing.

staebler: Oh, so the controller only ever makes one attempt to stop and start the machines. For some reason, I had it in my head that only ever reconcile while stopping, the controller would attempt to stop any machines that are not in the stopped or stopping state.

pkg/controller/hibernation/azure_actuator.go
staebler: This is already broken. The OpenShift installer allows the user to specify an existing resource group to use now.
https://github.com/openshift/installer/blob/dbf10881b664be15ba525b9be3562dc478910631/pkg/types/azure/platform.go#L69-L77

staebler: Custom resource groups can be dealt with as part of a fix for https://issues.redhat.com/browse/CO-1150.

csrwng:
@staebler thx for the review. Addressed your comments. Calling start/stop as necessary should be addressed with https://issues.redhat.com/browse/CO-1119

pkg/controller/hibernation/azure_actuator.go
staebler: Or, even better,
```
result := make([]string, len(machines)
for i, m := range machines {
  result[i] = to.String(m.Name)
}
return result
```

pkg/controller/hibernation/azure_actuator_test.go
```diff
@@ -0,0 +1,247 @@
...
+			var err error
+			switch test.testFunc {
+			case "StopMachines":
+				err = actuator.StopMachines(testAzureClusterDeployment(), nil, log.New())
+			case "StartMachines":
+				err = actuator.StartMachines(testAzureClusterDeployment(), nil, log.New())
+			default:
+				t.Fatal("Invalid function to test")
+			}
+			assert.Nil(t, err)
```
staebler, line 110:
```suggestion
			assert.NoError(t, err)
```

pkg/controller/hibernation/azure_actuator_test.go
```diff
@@ -0,0 +1,247 @@
...
+			var result bool
+			switch test.testFunc {
+			case "MachinesStopped":
+				result, err = actuator.MachinesStopped(testAzureClusterDeployment(), nil, log.New())
+			case "MachinesRunning":
+				result, err = actuator.MachinesRunning(testAzureClusterDeployment(), nil, log.New())
+			default:
+				t.Fatal("Invalid function to test")
+			}
+			require.Nil(t, err)
```
staebler, line 180:
```suggestion
			require.NoError(t, err)
```

pkg/controller/hibernation/azure_actuator_test.go
```diff
@@ -0,0 +1,247 @@
...
+		})
+	}
+}
+
+func testAzureClusterDeployment() *hivev1.ClusterDeployment {
+	return testcd.BasicBuilder().Options(func(cd *hivev1.ClusterDeployment) {
+		cd.Spec.ClusterMetadata = &hivev1.ClusterMetadata{
+			InfraID: azureTestInfraID,
+		}
+	}).Build()
```
staebler, line 191:
```suggestion
	return testcd.Build(func(cd *hivev1.ClusterDeployment) {
		cd.Spec.ClusterMetadata = &hivev1.ClusterMetadata{
			InfraID: azureTestInfraID,
		}
	})
```
But there is not much benefit in using a builder here.
```suggestion
	return &hivev1.ClusterDeployment{
		Spec: hivev1.ClusterDeploymentSpec{
			ClusterMetdata: &hivev1.ClusterMetadata{
				InfraID: azureTestInfraID,
			},
		},
	}
```

staebler (approved):
/lgtm

PR #1076, opened 2020-08-03
WIP: replace syncsetinstance

staebler:
Preliminary PoC for replacing SyncSetInstance with a CRD for which 1 CR is created for each ClusterDeployment.
This should be stable enough to do scale testing with to see whether it yields significant improvements.

PR #1090, opened 2020-08-19
[Selector]SyncSet label on cluster resources

2uasimojo:
For discoverability by service personnel, add a label to each resource pushed into the cluster at the behest of a [Selector]SyncSet. The label is identical to what's put on the SyncSetInstance, e.g.:

`hive.openshift.io/selector-syncset-name: ccs-dedicated-admins-environment`

Jira: [OSD-4848](https://issues.redhat.com/browse/OSD-4848)

### DIFF
--- a/pkg/controller/syncsetinstance/syncsetinstance_controller.go
+++ b/pkg/controller/syncsetinstance/syncsetinstance_controller.go
@@ -706,7 +706,7 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
 		if rss := findSyncStatus(resourceSyncStatus, ssi.Status.Resources); rss == nil || r.needToReapply(resourceSyncStatus, *rss, ssiLog) {
 			// Apply resource
 			ssiLog.Debug("applying resource")
-			applyErr = applyResource(info.Object, ssiLog, applyFnMetricsLabel, applyFn)
+			applyErr = applyResource(ssi, info.Object, ssiLog, applyFnMetricsLabel, applyFn)
 
 			var resourceSyncConditions []hivev1.SyncCondition
 			if rss != nil {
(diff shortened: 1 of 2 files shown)


2uasimojo:
/hold

- Needs nod from hive team
- If ^, needs test
- This should definitely go behind #1076, whereupon it will almost certainly need a nontrivial rebase.

2uasimojo:
Things discussed offline:

A resource can be managed by more than one [s]ss when using `patches`. So a singular label such as proposed here would need to be restricted to the [s]ss that *creates* the resource.

Otherwise, we would need a richer way of describing the provenance of the resource, such as an annotation containing a JSON string with a list of sources. Each list entry would indicate the kind (sss/ss) and name of the source. When multiple patches come from the same source object, we should collapse to a single entry. (We could provide more details, such as the mechanism (create/patch) and `resourceApplyMode` or `patchType`, respectively. But meh -- as long as we can backtrack to the source, we'll be able to see all of that for ourselves.)

This is appreciably, if not hugely, more complex than what's currently proposed. It's also a bit harder for the consumer to use: now we have to parse (by eye or by tool) JSON rather than a simple kind/name string.

pkg/controller/syncsetinstance/syncsetinstance_controller.go
```diff
@@ -741,16 +741,32 @@ func (r *ReconcileSyncSetInstance) applySyncSetResources(ssi *hivev1.SyncSetInst
...
+	// Resource is managed by hive:
 	labels[constants.HiveManagedLabel] = "true"
-	u.SetLabels(labels)
+	// Which [Selector]SyncSet it originated from:
+	if ssi.Spec.SelectorSyncSetRef != nil {
+		labels[constants.SelectorSyncSetNameLabel] = ssi.Spec.SelectorSyncSetRef.Name
+	} else if ssi.Spec.SyncSetRef != nil {
+		labels[constants.SyncSetNameLabel] = ssi.Spec.SyncSetRef.Name
+	}
+	// TODO(efried): else panic? This is a developer error (forgot to add a condition for new ssi source type).
```
joelddiaz, line 759: I don't think we want to panic on some badly formatted SSI (or b/c we downgraded Hive for some reason), but at minimum we could log the fact that we've encountered an unexpected SSI.

2uasimojo:
Superseded by #1129, which
- takes into account the big switch to ClusterSync
- addresses the whole "you can't just use a label because multiple [S]SSes can touch the resource" thing.

PR #1091, opened 2020-08-19
Improve Hive leader election

akhil-rane:
This PR introduces following changes to hive's leader election
- Lease duration set to 360s 
- Renew deadline set to 270s
- Retry period set to 90s (default of 2s results in excessive etcd writes) 
- Enabled ReleaseOnCancel to immediately release the lock on normal shutdown 
  eliminating the delay before another pod takes over
- Removed env var DISABLE_LEADER_ELECTION used to skip leader election

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -1,26 +1,33 @@
 package main
 
 import (
+	"context"
 	"flag"
 	golog "log"
 	"math/rand"
+	"net/http"
 	"os"
-	"strconv"
+	"os/signal"
+	"syscall"
 	"time"
 
+	"github.com/google/uuid"
 	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	crv1alpha1 "k8s.io/cluster-registry/pkg/apis/clusterregistry/v1alpha1"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 
@@ -57,9 +64,9 @@ import (
 const (
 	defaultLogLevel             = "info"
 	leaderElectionConfigMap     = "hive-controllers-leader"
-	leaderElectionLeaseDuration = "40s"
-	leaderElectionRenewDeadline = "20s"
-	leaderElectionlRetryPeriod  = "4s"
+	leaderElectionLeaseDuration = "360s"
+	leaderElectionRenewDeadline = "270s"
+	leaderElectionRetryPeriod   = "90s"
 )
 
 type controllerSetupFunc func(manager.Manager) error
@@ -124,7 +131,7 @@ func newRootCommand() *cobra.Command {
 			if err != nil {
 				log.WithError(err).Fatal("Cannot parse renew deadline")
 			}
-			retryPeriod, err := time.ParseDuration(leaderElectionlRetryPeriod)
+			retryPeriod, err := time.ParseDuration(leaderElectionRetryPeriod)
 			if err != nil {
 				log.WithError(err).Fatal("Cannot parse retry period")
 			}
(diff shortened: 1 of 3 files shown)


pkg/operator/assets/bindata.go
```diff
@@ -905,10 +905,12 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         readinessProbe:
+          periodSeconds: 360
           httpGet:
             path: /readyz
             port: 8080
         livenessProbe:
+          periodSeconds: 360
```
akhil-rane, line 913: I have set `periodSeconds`(health check interval) equal to lease duration since that is the amount of time a new client will take to acquire leadership and get into ready state

staebler: I don't think this is sufficient. It precludes running multiple pods, where the other pods are waiting to acquire leadership.

staebler: Also, even if we were to go this route, I think that you would want to use `initialDelaySeconds` rather than `periodSeconds`.

akhil-rane: Do you think just `initialDelaySeconds` of 400 should be enough?

staebler: No, that will have the same problem with multiple pods. Let's say there is one pod running that is the leader. Now, let's say that a second pod starts. The second pod should be allowed to run indefinitely waiting for the current leader to relinquish leadership. If the second pod is not serving to satisfy its liveness probe, then the kube controller will kill the pod rather than allowing it to continue waiting.

akhil-rane: So that means we should remove liveness probe then? because as per my understanding the second pod won't become leader (and continuously fail liveness probe) till the leader pod dies.

staebler: No, that is not an acceptable solution either. The liveness probe is needed to ensure that a pod that has stopped working gets killed. The solution, as far as I see it, is to start serving the liveness prior to waiting to become the leader. The pod is alive even when not the leader.

cmd/manager/main.go
```diff
@@ -151,11 +155,23 @@ func newRootCommand() *cobra.Command {
...
+				w.WriteHeader(http.StatusOK)
+			})
+			http.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+				if isLeader {
+					w.WriteHeader(http.StatusOK)
+				}
+				w.WriteHeader(http.StatusServiceUnavailable)
+			})
+			log.Info("Starting /healthz and /readyz endpoints")
+			go http.ListenAndServe(":8080", nil)
```
akhil-rane: I am not sure if this is the right way to do it but I tested it and works perfectly fine

cmd/manager/main.go
```diff
@@ -144,85 +155,143 @@ func newRootCommand() *cobra.Command {
...
-			if val != "" {
-				disableLeaderElection, err = strconv.ParseBool(val)
-				if err != nil {
-					log.WithField("value", val).Fatal("Error parsing DISABLE_LEADER_ELECTION env var")
+			// Create and start liveness and readyness probe endpoints
+			http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+			http.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+				if isLeader {
```
staebler, line 163: Does `http.HandleFunc` provide a happens-before edge that will guarantee that changes to this field made in other goroutines will be visible in this goroutine?

cmd/manager/main.go
```diff
@@ -144,85 +155,143 @@ func newRootCommand() *cobra.Command {
...
 			log.Infof("hive namespace: %s", hiveNSName)
 
-			// Allow an env var to disable leader election, useful when testing from source with make run to eliminate the wait time.
-			var disableLeaderElection bool
-			val := os.Getenv("DISABLE_LEADER_ELECTION")
-			if val != "" {
-				disableLeaderElection, err = strconv.ParseBool(val)
-				if err != nil {
-					log.WithField("value", val).Fatal("Error parsing DISABLE_LEADER_ELECTION env var")
+			// Create and start liveness and readyness probe endpoints
```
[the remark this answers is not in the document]
akhil-rane: The implementation in the above commit only serves liveness probe. We will also need readyness probe right? for which we will need to have a separate endpoint.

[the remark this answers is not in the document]
staebler: I am not sure if we *need* the readiness probe. I added it in the PR including the above commit because it was available for free: We did not have it before. The only service that the hive controllers use is the metrics service. I am not versed well enough in how metrics work to know if it would be problematic or not to have a dormant hive-controllers pod getting requests via that metrics service.

Any insight on this, @dgoodwin?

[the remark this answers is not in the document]
dgoodwin: I'm quite sure prometheus won't have any major issue with it but it might cause some metrics blips where things disappear because there's no data for them. It might also trigger an alert if the endpoint is down for some period of time, I believe that's 5m today.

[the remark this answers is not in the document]
akhil-rane: I will take that as a yes for readiness probe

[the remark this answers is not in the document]
staebler: Yeah, that seems like a yes to me. In a scenario where we have a dormant hive-controllers, it seems as though there is a high likelihood of metrics blips.

cmd/manager/main.go
```diff
@@ -144,85 +155,143 @@ func newRootCommand() *cobra.Command {
...
+			http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+			http.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
akhil-rane: The message in the pod event is something like `Readiness probe failed: HTTP probe failed with statuscode: 503`. Is that what you were looking for?

cmd/manager/main.go
akhil-rane: I think yes. Otherwise readiness probe would have continuously failed during my testing even after the pod gets elected as a leader.

akhil-rane:
> We do leader election in hive-operator as well. That will need similar changes.

I have implemented it for both hive controllers and operator.

cmd/manager/main.go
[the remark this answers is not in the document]
staebler: Yes, that is what I was wondering.

cmd/manager/main.go
staebler: I think no. Just because it happens almost all of the time, does not make it a guarantee. You cannot write a pure int in one goroutine and the value to be seen in another goroutine. You need some mechanism to create a happens-before edge to guarantee visibility.

staebler (commented):
This looks good except for dealing with accessing the `isLeader` variable across goroutines.

akhil-rane:
PR updated

staebler (approved):
/lgtm

staebler:
The managed DNS is not configured properly.
```
time="2020-09-11T19:22:32.529Z" level=error msg="no scraper for domain found, skipping reconcile" controller=dnsendpoint dnszone=hive-5005f3ee-8401-4f9c-b3f3-4f1991621843-zone domain=riodkflv.hive-ci.openshift.com namespace=cluster-test
```

cmd/manager/main.go
```diff
@@ -141,81 +155,150 @@ func newRootCommand() *cobra.Command {
...
-				if !ok {
-					log.WithField("controller", name).Fatal("no entry for controller found")
+				if err := velerov1.AddToScheme(mgr.GetScheme()); err != nil {
+					log.Fatal(err)
 				}
-				if disabledControllersSet.Has(name) {
-					log.WithField("controller", name).Debugf("skipping disabled controller")
-					continue
+
+				if err := awsprovider.AddToScheme(mgr.GetScheme()); err != nil {
```
staebler, line 210: This was removed in d0da9cb7df6f6b927db2a56ae13c26e5646b1d6c.

staebler (commented):
/lgtm cancel

staebler:
> The managed DNS is not configured properly.
> 
> ```
> time="2020-09-11T19:22:32.529Z" level=error msg="no scraper for domain found, skipping reconcile" controller=dnsendpoint dnszone=hive-5005f3ee-8401-4f9c-b3f3-4f1991621843-zone domain=riodkflv.hive-ci.openshift.com namespace=cluster-test
> ```

There are two replicasets for hive-controllers, each referencing a different managed-domains configmap. The hive-controllers pod that won the leader election is using a managed-domains configmap that does not exist.

staebler (approved):
/lgtm

PR #1095, opened 2020-08-24
Revert "expose hiveAPIEnabled in hiveconfig template"

gregsheremeta:
This reverts commit 53c19cc2702ff7276517ec1047abec3937bb5cd8.

The template doesn't like the boolean -- it wants a string.
We don't feel like messing with this anymore because it is
temporary, so we will revert and instead ask SRE-P to change
it manually.

/cc @staebler @dgoodwin


gregsheremeta:
forcing it to unblock the queue

PR #1098, opened 2020-08-26
pick one DNS cleanup codepath and use it everywhere

joelddiaz:
Rather than two similar DNS cleanup functions, unify onto one implementation that can be used by all callers.

This means we now remove more than just A records on a failed install.

### DIFF
--- a/pkg/controller/dnszone/awsactuator.go
+++ b/pkg/controller/dnszone/awsactuator.go
@@ -433,22 +433,30 @@ func (a *AWSActuator) Delete() error {
 
 func (a *AWSActuator) deleteRecordSets(logger log.FieldLogger) error {
 	logger.Info("Deleting route53 recordsets in hostedzone")
+
+	return DeleteAWSRecordSets(a.awsClient, a.dnsZone, logger)
+}
+
+// DeleteAWSRecordSets will clean up a DNS zone down to the minimum required record entries
+func DeleteAWSRecordSets(awsClient awsclient.Client, dnsZone *hivev1.DNSZone, logger log.FieldLogger) error {
+
 	maxItems := "100"
 	listInput := &route53.ListResourceRecordSetsInput{
-		HostedZoneId: a.zoneID,
+		HostedZoneId: dnsZone.Status.AWS.ZoneID,
 		MaxItems:     &maxItems,
 	}
 	for {
-		listOutput, err := a.awsClient.ListResourceRecordSets(listInput)
+		listOutput, err := awsClient.ListResourceRecordSets(listInput)
 		if err != nil {
 			return err
 		}
 		var changes []*route53.Change
 		for _, recordSet := range listOutput.ResourceRecordSets {
 			// Ignore the 2 recordsets that are created with the hosted zone and that cannot be deleted
-			if n, t := aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type); n == controllerutils.Dotted(a.dnsZone.Spec.Zone) && (t == route53.RRTypeNs || t == route53.RRTypeSoa) {
+			if n, t := aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type); n == controllerutils.Dotted(dnsZone.Spec.Zone) && (t == route53.RRTypeNs || t == route53.RRTypeSoa) {
 				continue
 			}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/dnszone/awsactuator.go
```diff
@@ -433,22 +434,39 @@ func (a *AWSActuator) Delete() error {
 
 func (a *AWSActuator) deleteRecordSets(logger log.FieldLogger) error {
 	logger.Info("Deleting route53 recordsets in hostedzone")
+
+	return DeleteAWSRecordSets(a.awsClient, a.dnsZone, []string{}, logger)
+}
+
+// DeleteAWSRecordSets will clean up all non-NS and non-SOA record types from a DNS zone when no
```
staebler, line 441: The function cleans up NS and SOA records too, just not the ones created with the zone.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -433,22 +434,39 @@ func (a *AWSActuator) Delete() error {
 
 func (a *AWSActuator) deleteRecordSets(logger log.FieldLogger) error {
 	logger.Info("Deleting route53 recordsets in hostedzone")
+
+	return DeleteAWSRecordSets(a.awsClient, a.dnsZone, []string{}, logger)
```
staebler, line 438:
```suggestion
	return DeleteAWSRecordSets(a.awsClient, a.dnsZone, nil, logger)
```

pkg/controller/dnszone/awsactuator.go
```diff
@@ -433,22 +434,39 @@ func (a *AWSActuator) Delete() error {
...
 		for _, recordSet := range listOutput.ResourceRecordSets {
 			// Ignore the 2 recordsets that are created with the hosted zone and that cannot be deleted
-			if n, t := aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type); n == controllerutils.Dotted(a.dnsZone.Spec.Zone) && (t == route53.RRTypeNs || t == route53.RRTypeSoa) {
+			if n, t := aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type); n == controllerutils.Dotted(dnsZone.Spec.Zone) && (t == route53.RRTypeNs || t == route53.RRTypeSoa) {
 				continue
 			}
+
+			// If we've been provided a list of record set types to delete
+			// and the current recordSet is not in that list, skip deleting it.
+			if deleteSet.Len() > 0 && !deleteSet.Has(*recordSet.Type) {
```
staebler, line 466: I understand that you are trying to preserve the existing functionality, but I don't think this is necessary. There is no harm in deleting all of the recordsets in between installs the same as we do for a deprovision.

pkg/installmanager/dnscleanup.go
```diff
@@ -1,56 +1,72 @@
...
-			}
-		}
+	if err := dns.DeleteAWSRecordSets(awsClient, dnsZone, recordTypes, zoneLogger); err != nil {
+		logger.WithError(err).Error("failed to clean up DNS Zone")
+	} else {
+		zoneLogger.Info("DNSZone A records deleted")
 	}
-	zoneLogger.Info("DNSZone A records deleted")
-	return nil
+	return err
```
staebler, line 71: I think that you want to return the error returned from `dns.DeleteAWSRecordSets`. This is returning the error returned from `awsclient.NewClient`, which at this point will always be nil.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -433,22 +433,30 @@ func (a *AWSActuator) Delete() error {
...
+	return DeleteAWSRecordSets(a.awsClient, a.dnsZone, logger)
+}
+
+// DeleteAWSRecordSets will clean up a DNS zone down to the minimum required record entries
+func DeleteAWSRecordSets(awsClient awsclient.Client, dnsZone *hivev1.DNSZone, logger log.FieldLogger) error {
+
 	maxItems := "100"
 	listInput := &route53.ListResourceRecordSetsInput{
-		HostedZoneId: a.zoneID,
+		HostedZoneId: dnsZone.Status.AWS.ZoneID,
```
gregsheremeta, line 445: panic here

PR #1099, opened 2020-08-26
Restart deprovision pods if credentials secret changes.

dgoodwin:
Previously the uninstall pod would never restart if it launched with
since deleted or bad creds, and the user updated the credentials secret.

This change switches to mounting the creds as a secret instead of
passing through env vars. The hiveutil wrapper code sets them as env
vars and establishes a file watch on the directory where the secret is
mounted. On any change or watch error we exit to restart the pod.

It seems to usually take about 1 minute for the secret changes to
surface in the pod.

### DIFF
--- a/contrib/pkg/deprovision/awstagdeprovision.go
+++ b/contrib/pkg/deprovision/awstagdeprovision.go
@@ -2,13 +2,19 @@ package deprovision
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
-
-	"github.com/openshift/installer/pkg/destroy/aws"
-	"github.com/spf13/cobra"
+	"time"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/installer/pkg/destroy/aws"
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+
+	"github.com/openshift/hive/pkg/constants"
 )
 
 // NewDeprovisionAWSWithTagsCommand is the entrypoint to create the 'aws-tag-deprovision' subcommand
@@ -25,6 +31,54 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
 				log.WithError(err).Error("Cannot complete command")
 				return
 			}
+
+			// If environment variables are not set, assume we're running in a deprovision pod with an aws credentials secret mounted.
+			// Parse credentials from files mounted in as a secret volume and set as env vars. We use the file
+			// instead of passing env vars directly so we can monitor for changes and restart the pod if an admin
+			// modifies the creds secret after the uninstall job has launched.
+			if os.Getenv("AWS_ACCESS_KEY_ID") == "" && os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {
+				log.WithField("credsMount", constants.AWSCredsMount).Info(
+					"AWS environment variables not set, assume running in pod with cred secret mounted")
+				awsAccessKeyIDFile := filepath.Join(constants.AWSCredsMount, constants.AWSAccessKeyIDSecretKey)
+				awsSecretAccessKeyFile := filepath.Join(constants.AWSCredsMount, constants.AWSSecretAccessKeySecretKey)
+				files := make(map[string][]byte, 2)
+
+				data, err := ioutil.ReadFile(awsAccessKeyIDFile)
+				if err != nil {
+					log.WithError(err).Fatalf("error reading %s file", awsAccessKeyIDFile)
+				}
+				os.Setenv("AWS_ACCESS_KEY_ID", string(data))
+				files[awsAccessKeyIDFile] = data
+
+				data, err = ioutil.ReadFile(awsSecretAccessKeyFile)
+				if err != nil {
+					log.WithError(err).Fatalf("error reading %s file", awsSecretAccessKeyFile)
+				}
+				os.Setenv("AWS_SECRET_ACCESS_KEY", string(data))
+				files[awsSecretAccessKeyFile] = data
+
+				fullFilenames := []string{
+					awsAccessKeyIDFile,
+					awsSecretAccessKeyFile,
+				}
+				obs, err := fileobserver.NewObserver(10 * time.Second)
+				if err != nil {
+					log.WithError(err).Fatal("could not set up file observer to check for credentials changes")
+				}
+				obs.AddReactor(func(filename string, action fileobserver.ActionType) error {
+					log.WithField("filename", filename).Info("exiting because credentials secret mount file changed")
+					os.Exit(1)
+					return nil
+				}, files, fullFilenames...)
+
+				stop := make(chan struct{})
+				go func() {
+					log.WithField("files", fullFilenames).Info("running file observer")
+					obs.Run(stop)
+					log.Fatal("file observer stopped")
+				}()
+			}
+
 			if err := opt.Run(); err != nil {
 				log.WithError(err).Fatal("Runtime error")
 			}
@@ -37,6 +91,7 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
 }
 
 func completeAWSUninstaller(o *aws.ClusterUninstaller, logLevel string, args []string) error {
+
 	for _, arg := range args {
 		filter := aws.Filter{}
 		err := parseFilter(filter, arg)
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -272,6 +272,9 @@ const (
 
 	// OvirtConfigEnvVar is the environment variable specifying the oVirt config path
 	OvirtConfigEnvVar = "OVIRT_CONFIG"
+
+	// AWSCredsMount is the location where the AWS credentials secret is mounted for uninstall pods.
+	AWSCredsMount = "/etc/aws-creds"
 )
 
 // GetMergedPullSecretName returns name for merged pull secret name per cluster deployment
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -19,10 +19,6 @@ import (
 )
 
 const (
-	// DefaultInstallerImage is the image that will be used to install a ClusterDeployment if no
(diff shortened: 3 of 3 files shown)


contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,9 +30,47 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
...
+			}
+			os.Setenv("AWS_SECRET_ACCESS_KEY", string(data))
+
+			// creates a new file watcher
+			watcher, err := fsnotify.NewWatcher()
+			if err != nil {
+				log.WithError(err).Fatal("error creating fsnotify watcher")
+			}
+			defer watcher.Close()
+			done := make(chan bool)
```
dgoodwin, line 53: Not sure if this done chan is needed.

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,9 +30,47 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
 				log.WithError(err).Error("Cannot complete command")
 				return
 			}
+			// Parse credentials from files mounted in as a secret volume and set as env vars. We use the file
+			// instead of passing env vars directly so we can monitor for changes and restart the pod if an admin
+			// modifies the creds secret after the uninstall job has launched.
+			data, err := ioutil.ReadFile(filepath.Join(constants.AWSCredsMount, "aws_access_key_id"))
```
joelddiaz, line 36: use the constants for `aws_access_key_id` and `aws_secret_access_key`

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -2,13 +2,18 @@ package deprovision
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
-	"github.com/openshift/installer/pkg/destroy/aws"
+	"github.com/fsnotify/fsnotify"
```
staebler, line 10: Was there any consideration for using `github.com/openshift/library-go/pkg/controller/fileobserver` instead?

contrib/pkg/deprovision/awstagdeprovision.go
staebler: No, it is not needed.

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,9 +30,47 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
...
+
+			// out of the box fsnotify can watch a single file, or a single directory
+			if err := watcher.Add(constants.AWSCredsMount); err != nil {
+				log.WithError(err).Fatalf("error establishing watch on %s", constants.AWSCredsMount)
+			}
+
 			if err := opt.Run(); err != nil {
 				log.WithError(err).Fatal("Runtime error")
 			}
+			<-done
```
staebler, line 73: This will block indefinitely since nothing is ever sending to the channel.

staebler:
Are we going to do the same thing for the other platforms, as well?

contrib/pkg/deprovision/awstagdeprovision.go
dgoodwin: No was not aware of that, any benefits? The type is uncommented. I'm going to assume yes and replace it.

dgoodwin:
Updated to respect the env vars if they are set, allowing us to keep using aws-tag-deprovision on the command line.

Also now using the library-go file watcher.

As for other clouds, it looks like it would be pretty easy to add for several including Azure and GCP. I am hesitant to get into the others which are harder to test though. In any case will discuss and consider for follow up PRs.

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -2,13 +2,19 @@ package deprovision
...
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
+	"time"
 
-	"github.com/openshift/installer/pkg/destroy/aws"
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
```
staebler, line 11: This import should go down with the other openshift import.

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,6 +31,54 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
 				log.WithError(err).Error("Cannot complete command")
 				return
 			}
+
+			// If environment variables are not set, assume we're running in a deprovision pod with an aws credentials secret mounted.
+			// Parse credentials from files mounted in as a secret volume and set as env vars. We use the file
+			// instead of passing env vars directly so we can monitor for changes and restart the pod if an admin
+			// modifies the creds secret after the uninstall job has launched.
+			if os.Getenv("AWS_ACCESS_KEY_ID") == "" && os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {
```
staebler, line 39: If we are changing the way that we are getting creds, I would **love** to support a creds file to keep parity with the `hiveutil create-cluster` command.

dgoodwin: Our established AWS secret format does not fit the known AWS cred file formats. So to do this we'd have to start generating another secret containing the file contents. That was complexity I'm not super eager to get into.

dgoodwin:
> Alternatively, I will propose again that we can use the RunWithContext function instead of the Run function on the uninstaller. We are going to be doing this at some point anyway. Since that will allow the uninstall to run for a set period of time and restart with a backoff, there is no need to restart the pod.

My concern there is the length of the window. I think QE or anyone doing this is going to expect a fairly quick turnaround when they fix those creds, whereas if we're terminating and stopping uninstall it feels like we need to give that a fairly long window. Best case we're seeing about 1 minute for this to take effect, and the alternative I would have to imagine would be at least 10 minutes and even that is pushing it. The response time here would be ten times better.

contrib/pkg/deprovision/awstagdeprovision.go
staebler: I am not suggesting that we change the AWS secret format (although I **do** think that is a good long-term change). I am suggesting that running `hiveutil aws-tag-deprovision` directly should support using an AWS creds file.

staebler: We made a similar change in the last few months for the GCP deprovision because of users' complaints.

dgoodwin:
/hold 

Pursuing the sidecar file monitor approach instead.

contrib/pkg/deprovision/awstagdeprovision.go
joelddiaz: I think as part of starting up the uninstall, the code can take the contents of the secret and make a minimal AWS creds file that we can point to with `AWS_SHARED_CREDENTIALS_FILE` pointing to a file that looks like

```
[default]
aws_access_key_id = ACCESS_ID_FROM_SECRET
aws_secret_access_key = SECRET_KEY_FROM_SECRET
```

dgoodwin:
I am calling it quits with the sidecar approach, details in #1108. The IPC has been far too complicated and burned too much time on it.

Fixed the import requested above. I will file a jira for additional clouds, which all go through one deprovision command so hopefully we can add some common code there to let them each monitor their own files.

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,6 +31,54 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
...
+				log.WithField("credsMount", constants.AWSCredsMount).Info(
+					"AWS environment variables not set, assume running in pod with cred secret mounted")
+				awsAccessKeyIDFile := filepath.Join(constants.AWSCredsMount, constants.AWSAccessKeyIDSecretKey)
+				awsSecretAccessKeyFile := filepath.Join(constants.AWSCredsMount, constants.AWSSecretAccessKeySecretKey)
+				files := make(map[string][]byte, 2)
+
+				data, err := ioutil.ReadFile(filepath.Join(constants.AWSCredsMount, constants.AWSAccessKeyIDSecretKey))
```
joelddiaz, line 46:
```suggestion
				data, err := ioutil.ReadFile(filepath.Join(awsAccessKeyIDFile)
```

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,6 +31,54 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
...
+				if err != nil {
+					log.WithError(err).Fatalf("error reading %s file", constants.AWSAccessKeyIDSecretKey)
```
joelddiaz, line 48: maybe also use the full path stored in `awsAccessKeyIDFile` here

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,6 +31,54 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
...
+				}
+				os.Setenv("AWS_ACCESS_KEY_ID", string(data))
+				files[awsAccessKeyIDFile] = data
+
+				data, err = ioutil.ReadFile(filepath.Join(constants.AWSCredsMount, constants.AWSSecretAccessKeySecretKey))
```
joelddiaz, line 53:
```suggestion
				data, err = ioutil.ReadFile(awsSecretAccessKeyFile)
```

contrib/pkg/deprovision/awstagdeprovision.go
```diff
@@ -25,6 +31,54 @@ func NewDeprovisionAWSWithTagsCommand() *cobra.Command {
...
+				if err != nil {
+					log.WithError(err).Fatalf("error reading %s file", constants.AWSSecretAccessKeySecretKey)
```
joelddiaz, line 55: use full path name stored in `awsSecretAccessKeyFile` here too

dgoodwin:
Updated, thanks!

joelddiaz (commented):
/lgtm

PR #1100, opened 2020-08-26
perform DNS cleanup after failed install for GCP

joelddiaz:
Similar to how we try to clean up stray DNS records after failed installation (when the ClusterDeployment has ManagedDNS), do the same for GCP.

Share the existing zone cleanup code from the dnszone controller.

### DIFF
--- a/pkg/controller/dnszone/dnszone_controller_test.go
+++ b/pkg/controller/dnszone/dnszone_controller_test.go
@@ -9,8 +9,10 @@ import (
 	"github.com/golang/mock/gomock"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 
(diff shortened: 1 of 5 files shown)


pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -253,7 +253,7 @@ func TestReconcileDNSProviderForGCP(t *testing.T) {
 		},
 		{
 			name:    "Delete managed zone",
-			dnsZone: validDNSZoneBeingDeleted(),
+			dnsZone: validGCPDNSZoneBeingDeleted(),
```
staebler, line 256: We should change all of the dns zones in this test function to use GCP dns zones instead of AWS ones. I am fine with creating a card to do it as a separate PR.

It may also be worth switching to use the builder in pkg/test/dnszone/dnszone.go to build these DNSZones.

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -119,24 +119,28 @@ func (a *GCPActuator) Delete() error {
...
 func (a *GCPActuator) deleteRecordSets(logger log.FieldLogger) error {
 	logger.Info("Deleting recordsets in managedzone")
+	return DeleteGCPRecordSets(a.gcpClient, a.dnsZone, logger)
+}
+
+func DeleteGCPRecordSets(gcpClient gcpclient.Client, dnsZone *hivev1.DNSZone, logger log.FieldLogger) error {
 	listOpts := gcpclient.ListResourceRecordSetsOptions{}
 	for {
-		listOutput, err := a.gcpClient.ListResourceRecordSets(a.managedZone.Name, listOpts)
+		listOutput, err := gcpClient.ListResourceRecordSets(*dnsZone.Status.GCP.ZoneName, listOpts)
```
staebler, line 128: This change looks susceptible to the same panic experienced with the AWS changes.

joelddiaz:
/hold
this requires the PR #1104 (or something similar)

PR #1101, opened 2020-08-27
Revert "pick one DNS cleanup codepath and use it everywhere"

gregsheremeta:
This reverts commit 65fb4af148ca96acae23681bb6865b2724cca417.

Causes a panic.

/cc @joelddiaz @akhil-rane


gregsheremeta:
ok, e2e, I've waited long enough for you.

Forcing...

PR #1102, opened 2020-08-27
[WIP]: HTTP Proxy Support added for hiveutil

klaphi:
Hi Hive-Team,

I need support for a HTTP Proxy in Hive, therefore I added the funtionality to the code and I would be happy if this code can be merged.

Please let me know whats missing.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -34,6 +35,7 @@ import (
 	"github.com/openshift/hive/pkg/constants"
 	"github.com/openshift/hive/pkg/gcpclient"
 	"github.com/openshift/hive/pkg/resource"
+	"github.com/openshift/installer/pkg/validate"
 )
 
 const longDesc = `
@@ -152,6 +154,9 @@ type Options struct {
 	Region                   string
 	Labels                   []string
 	SkipMachinePools         bool
+	HttpProxy                string
+	HttpsProxy               string
+	NoProxy                  string
 
 	// AWS
 	AWSUserTags []string
@@ -263,6 +268,11 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=ovirt --ovirt-api-vip 192.168.1.2
 	flags.StringVar(&opt.Region, "region", "", "Region to which to install the cluster. This is only relevant to AWS, Azure, and GCP.")
 	flags.StringSliceVarP(&opt.Labels, "labels", "l", nil, "Label to apply to the ClusterDeployment (key=val)")
 	flags.BoolVar(&opt.SkipMachinePools, "skip-machine-pools", false, "Skip generation of Hive MachinePools for day 2 MachineSet management")
+
+	// HTTP Proxy flags
+	flags.StringVar(&opt.HttpProxy, "http-proxy", "", "HTTP-Proxy defines the proxy settings for the cluster")
+	flags.StringVar(&opt.HttpsProxy, "https-proxy", "", "HTTPS-Proxy defines the proxy settings for the cluster")
+	flags.StringVar(&opt.NoProxy, "no-proxy", "", "NO-Proxy defines the proxy settings for the cluster")
 
 	// Flags related to adoption.
 	flags.BoolVar(&opt.Adopt, "adopt", false, "Enable adoption mode for importing a pre-existing cluster into Hive. Will require additional flags for adoption info.")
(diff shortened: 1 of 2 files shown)


contrib/pkg/createcluster/create.go
```diff
@@ -392,6 +402,42 @@ func (o *Options) Validate(cmd *cobra.Command) error {
 			return fmt.Errorf("unable to parse key=value label: %s", ls)
 		}
 	}
+
+	if len(o.HttpProxy) > 0 || len(o.HttpsProxy) > 0 {
+		if o.HttpProxy == "" && o.HttpsProxy == "" {
```
dgoodwin, line 407: This doesn't look quite right, I can't see how line 407 could ever be true given line 406.

contrib/pkg/createcluster/create.go
```diff
@@ -392,6 +402,42 @@ func (o *Options) Validate(cmd *cobra.Command) error {
...
+			cmd.Usage()
+			log.Info("must set --http-proxy or --https-proxy")
+			return fmt.Errorf("must set --http-proxy or --https-proxy")			
+		}
+	}
+		
+	if len(o.HttpProxy) > 0 && o.HttpProxy != "" {
```
dgoodwin, line 414: Check for empty string seems to duplicate the length check, I think just len() > 0 is sufficient and correct.

contrib/pkg/createcluster/create.go
```diff
@@ -392,6 +402,42 @@ func (o *Options) Validate(cmd *cobra.Command) error {
...
+		if err := validate.URI(o.HttpProxy); err != nil {
+			cmd.Usage()
+			log.Info("--http-proxy specified in wrong format")
+			return fmt.Errorf("--http-proxy specified in wrong format: %s", err.Error())
+		}
+	}
+	if len(o.HttpsProxy) > 0 && o.HttpsProxy != "" {
```
dgoodwin, line 421: Same here.

contrib/pkg/createcluster/create.go
```diff
@@ -392,6 +402,42 @@ func (o *Options) Validate(cmd *cobra.Command) error {
...
+		if err := validate.URI(o.HttpsProxy); err != nil {
+			cmd.Usage()
+			log.Info("--https-proxy specified in wrong format")
+			return fmt.Errorf("--https-proxy specified in wrong format: %s", err.Error())
+		}
+	}
+	if len(o.NoProxy) > 0 && o.NoProxy != "" {
```
dgoodwin, line 428: Same here.

contrib/pkg/createcluster/create.go
```diff
@@ -392,6 +402,42 @@ func (o *Options) Validate(cmd *cobra.Command) error {
...
+		for _, v := range strings.Split(o.NoProxy, ",") {
+			v = strings.TrimSpace(v)
+			errDomain := validate.NoProxyDomainName(v)
+			_, _, errCIDR := net.ParseCIDR(v)
+			if errDomain != nil && errCIDR != nil {
+				cmd.Usage()
+				log.Info("--no-proxy specified in wrong format - must be a CIDR or domain, without wildcard characters")
```
dgoodwin, line 435: Looks like this should be "must be a comma separated list of CIDR domains without wildcard characters"

pkg/clusterresource/builder.go
```diff
@@ -133,6 +146,32 @@ func (o *Builder) Validate() error {
 		return fmt.Errorf("must set serving cert key to use with serving cert")
 	}
 
+	if len(o.HTTPProxy) > 0  || len(o.HTTPSProxy) > 0  {
+		if o.HTTPProxy == "" && o.HTTPSProxy == "" {
```
dgoodwin, line 150: Looks impossible for line 150 to trigger here as well. Was this meant to be a check that you don't try to set BOTH httpproxy and httpsproxy? If so I'm guessing these should both be != "".

pkg/clusterresource/builder.go
```diff
@@ -133,6 +146,32 @@ func (o *Builder) Validate() error {
...
+			return fmt.Errorf("must set HTTPProxy or HTTPSProxy")
```
dgoodwin, line 151: If above assumptions is correct, this should probably reflect that, something like "cannot set both HTTPProxy and HTTPSProxy".

dgoodwin:
@klaphi curious what you're working on, but some general thoughts, hiveutil is mainly a dev tool, I wouldn't recommend building any major solutions around it, if you're investing in Hive it would be best to focus on the go API rather than hiveutil as it may change underneath you in unexpected ways.

You can use hiveutil to get you started on that path though if desired with -o yaml. Most apps integrating with Hive today are using the controller-runtime Go client with our Go types in pkg/apis/hive/v1.

dgoodwin:
/close

Please reopen if there's interest in pursuing this.

PR #1104, opened 2020-08-27
update DNSZone status after refresh AWS/Azure/GCP

joelddiaz:
Populate DNSZone status immediately after fetching zoneID.
Add status fields for Azure to store the dns zone name like we do for the other clouds.

### DIFF
--- a/pkg/controller/dnszone/awsactuator.go
+++ b/pkg/controller/dnszone/awsactuator.go
@@ -227,6 +227,12 @@ func (a *AWSActuator) Refresh() error {
 		}
 		logger.Debug("Found hosted zone")
 		a.zoneID = &zoneID
+
+		// Update dnsZone status now that we have the zoneID
+		if err := a.ModifyStatus(); err != nil {
+			a.logger.WithError(err).Error("failed to update status after refresh")
+			return err
+		}
 	}
 
 	if a.zoneID == nil {
--- a/config/crds/hive.openshift.io_dnszones.yaml
+++ b/config/crds/hive.openshift.io_dnszones.yaml
@@ -136,6 +136,10 @@ spec:
             azure:
               description: AzureDNSZoneStatus contains status information specific
                 to Azure
+              properties:
+                zoneName:
+                  description: ZoneName is the name of the zone on Azure
+                  type: string
               type: object
             conditions:
               description: Conditions includes more detailed status for the DNSZone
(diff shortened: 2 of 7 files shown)


pkg/controller/dnszone/awsactuator.go
```diff
@@ -227,6 +227,12 @@ func (a *AWSActuator) Refresh() error {
 		}
 		logger.Debug("Found hosted zone")
 		a.zoneID = &zoneID
+
+		// Update dnsZone status now that we have the zoneID
+		if err := a.ModifyStatus(); err != nil {
```
staebler, line 232: How would you feel about a different approach where the `DeleteAWSRecordSets` gets passed the zone ID and zone name rather than the `DNSZone`?

staebler:
The description for this PR does not really match the functional changes made. There *is* a test case added, but the main change of this PR is to set the DNSZone status on refresh.

pkg/controller/dnszone/awsactuator.go
staebler: And if we are going to set the DNSZone status in the `Refresh` function, then we should do that for all of the platforms.

joelddiaz: That works as well (how it previously worked). It just seemed odd that we pulled the zoneID from AWS and didn't update the DNSZone object status to reflect the new data.

Is there any reason we shouldn't be updating the DNSZone status as soon as possible?

staebler: It seems fine to me that the information is stored in the status of the DNSZone rather than the actuator. If we are going to do that, we should just get rid of the `ModifyStatus` function. Maybe @twiest has some insight into this?

joelddiaz: With the currently proposed behavior in this PR, we can make ModifyStatus an internal function for the actuators that runs at the end of a Refresh()

twiest: There is a difference in behavior between when `Refresh()` fails and when `ModifyStatus()` fails (`ModifyStatus()` sets the `RequeueAfter` value).

I'm not sure how significant this is, but just something to look into.

There was a reason that we did `ModifyStatus()` separate, but I don't remember it now.

I looked through the code and it seems like it would act the same, so maybe whatever reason we had has now been refactored out of the code.

I'm a fan of having the actuator interface have as few methods as necessary, so this change seems fine assuming it won't break something. :)

PR #1105, opened 2020-08-28
Small fixes for issues found in the unreachable hotloop investigation

dgoodwin:
Details in individual commits.


dgoodwin:
Verified manually that the ssi controller now detects unreachable and exits early, which should prevent the hotloop.

PR #1106, opened 2020-08-31
Hibernation: increase delay to check for node readiness

csrwng:
It has been observed that certain clusters take longer than 2 minutes
to update their node status to "NotReady" when resuming from
hibernation. This introduces a quickfix of doubling the time that
the controller waits until it starts checking node status. However,
a longer term more deterministic fix is likely needed.

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -43,7 +43,7 @@ const (
 	// ready check after a cluster started resuming. This is to
 	// avoid a false positive when the node status is checked too
 	// soon after the cluster is ready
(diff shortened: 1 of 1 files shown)


dgoodwin:
Unfortunate but seems ok for now.

/lgtm

staebler:
Can the controller approve CSRs during the 4-minute `nodeCheckWaitTime` window rather than waiting until the 4 minutes has elapsed before starting to approve?

csrwng:
> Can the controller approve CSRs during the 4-minute `nodeCheckWaitTime` window rather than waiting until the 4 minutes has elapsed before starting to approve?

Ha, it certainly can, let me attempt a fix

PR #1107, opened 2020-09-01
Aws dns cleanup refactor

joelddiaz:
Try number 2 to use one AWS DNS cleanup implementation.

Rework the DNSZone controller's AWS actuator code for DNS zone cleanup so that it is an exported function that can also be called during a failed cluster installation.
    
Rather than two similar DNS cleanup functions, unify onto one implementation that can be used by all callers.
    
This means we now remove more than just A records on a failed intall.
    
Also, populate DNSZone status immediately after fetching zoneID during Refresh() and Create()
    
Add test case to cover DNSZone being deleted without status.
    
Align with other actuators and just store entire AWS dnszone after receiving it.

### DIFF
--- a/pkg/controller/dnszone/awsactuator.go
+++ b/pkg/controller/dnszone/awsactuator.go
@@ -36,8 +36,8 @@ type AWSActuator struct {
 	// awsClient is a utility for making it easy for controllers to interface with AWS
 	awsClient awsclient.Client
 
-	// zoneID is the ID of the hosted zone in route53
-	zoneID *string
+	// hostedZone is the AWS object representing the hosted zone in route53
+	hostedZone *route53.HostedZone
 
 	// currentTags are the list of tags associated with the currentHostedZone
 	currentHostedZoneTags []*route53.Tag
@@ -76,8 +76,8 @@ func NewAWSActuator(
 
 // UpdateMetadata ensures that the Route53 hosted zone metadata is current with the DNSZone
 func (a *AWSActuator) UpdateMetadata() error {
-	if a.zoneID == nil {
-		return errors.New("zoneID is unpopulated")
+	if a.hostedZone == nil {
+		return errors.New("hostedZone is unpopulated")
 	}
 
 	// For now, tags are the only things we can sync with existing zones.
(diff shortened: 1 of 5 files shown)


pkg/controller/dnszone/awsactuator.go
```diff
@@ -433,22 +443,30 @@ func (a *AWSActuator) Delete() error {
 
 func (a *AWSActuator) deleteRecordSets(logger log.FieldLogger) error {
```
twiest, line 444: Is this function still needed?

Would it be better to just have everything call `DeleteAWSRecordSets` directly?

pkg/controller/dnszone/awsactuator.go
```diff
@@ -360,7 +360,11 @@ func (a *AWSActuator) Create() error {
 		return err
 	}
 
-	a.zoneID = hostedZone.Id
+	a.hostedZone = hostedZone
+	if err := a.ModifyStatus(); err != nil {
+		logger.WithError(err).Error("failed to populate DNSZone status")
+		return err
+	}
```
staebler, line 367: This is snuck in the "just store entire aws dnszone" commit. Should it be in the "populate DNSZone status immediately after fetching zoneID (AWS)" commit instead?

pkg/controller/dnszone/awsactuator.go
```diff
@@ -36,8 +36,8 @@ type AWSActuator struct {
 	// awsClient is a utility for making it easy for controllers to interface with AWS
 	awsClient awsclient.Client
 
-	// zoneID is the ID of the hosted zone in route53
-	zoneID *string
+	// hostedZone is the AWS object representing the hosted zone in route53
```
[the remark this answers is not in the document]
joelddiaz: Just when comparing to the other actuators, it seemed odd that we didn't have access to the full object after we went through the trouble of fetching it. But you're right, it has no immediate effect other that making the aws actuator more similar to the other actuators.

pkg/controller/dnszone/awsactuator.go
joelddiaz: Yes, it came up while testing things and i must have squashed the wrong commits together while cleaning up the commits for submission.

joelddiaz:
@twiest @staebler PTAL, i addressed the feedback (and tried to clean up the history where the populating of DNSZone status is concerned)

twiest:
Looks good to me as well. One of can lgtm after the squash.

joelddiaz:
squashed and updated git commit text (and PR text as well)

joelddiaz:
@twiest i think the bot took an unplanned vacation and missed your lgtm

PR #1108, opened 2020-09-01
WIP: Uninst restart sidecar

dgoodwin:


dgoodwin:
This is a second implementation instead of https://github.com/openshift/hive/pull/1099 but I cannot yet get it to work. The file watcher doesn't seem to see any changes, it seems to be restarting itself, but it never kills the deprovision pod.

staebler:
> This is a second implementation instead of #1099 but I cannot yet get it to work. The file watcher doesn't seem to see any changes, it seems to be restarting itself, but it never kills the deprovision pod.

If it is working properly, I do not think that the sidecar container should ever restart.

staebler:
> > This is a second implementation instead of #1099 but I cannot yet get it to work. The file watcher doesn't seem to see any changes, it seems to be restarting itself, but it never kills the deprovision pod.
> 
> If it is working properly, I do not think that the sidecar container should ever restart.

Oops, ignore that. I think it does restart when the PID changes.

dgoodwin:
Got past problem 1 which was mounting the cred secret into the filewatch container, it actually expects the files to be in the container with the pid we will kill.

Unfortunately problem 2 looks worse, this code appears to be written for processes that never end, it will kill the pid but the pod stays up and that container just restarts. If the deprovision is successful, the file watcher goes into an infinite loop looking for the pid that no longer exists. There's no way to exit this as far as I can see. We're going to need to patch it or write our own filewatcher module. Maybe copy and modify.

dgoodwin:
Giving up on this approach, it's been days of attempts and I still can't get the IPC right even after forking library-go's watcher and trying to modify it for non-permanent processes (and I haven't found a way yet that I think they'll accept and works). Latest progress is in this branch but the code still can't detect when the process is completed, it's probably fixable but it's gone on way too long.

PR #1111, opened 2020-09-02
controller: create clustersync as separate reconcile

staebler:
For a new ClusterDeployment, the ClusterSync is created and then immediately updated with the status of applying any syncsets. If there were errors applying syncsets, then the clustersync controller may do its next reconcile using the created version instead of the updated version. When this happens, the update at the end of the reconcile will fail due to a conflict.

With these changes, when there is no ClusterSync, the clustersync controller will create the ClusterSync as the only thing that it does for the reconciliation. The clustersync controller will requeue the ClusterDeployment when the controller sees the new ClusterSync, and the controller will not get any conflicts when doing updates.

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -274,8 +274,17 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 	clusterSync := &hiveintv1alpha1.ClusterSync{}
 	switch err := r.Get(context.Background(), request.NamespacedName, clusterSync); {
 	case apierrors.IsNotFound(err):
-		logger.Info("ClusterSync does not exist; will need to create")
-		needToCreateClusterSync = true
+		logger.Info("creating ClusterSync as it does not exist")
+		clusterSync.Namespace = cd.Namespace
+		clusterSync.Name = cd.Name
+		ownerRef := metav1.NewControllerRef(cd, cd.GroupVersionKind())
+		ownerRef.Controller = nil
+		clusterSync.OwnerReferences = []metav1.OwnerReference{*ownerRef}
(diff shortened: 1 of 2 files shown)


staebler:
> I like this change -- doing the creation up front rather than deferring it. But based on your explanation in [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1599051849401900) (relevant bit copied below) I don't understand how it actually solves the problem.
> 
> > Let's assume that we have a fresh ClusterDeployment with no ClusterSync. The first reconcile will do the applies of the syncsets, create a ClusterSync (version A), then update the ClusterSync with the status of the applies (version B). Asynchronously, the informers will inform the controller's watches that there has been a ClusterSync created and a ClusterSync updated. In response to both of those, the controller will requeue the ClusterDeployment. When the controller reconciles the ClusterDeployment again, the ClusterSync that is in the local cache may be either version A or version B. If it is version A, then any update of the ClusterSync during that reconciliation will fail with a conflict.
> > Doing an update of a resource in a controller does not directly change that resource in the local cache. The local cache only changes after the API server sends the change out.

With these changes, the clustersync will always do a reconcile using version A of the ClusterSync, but that reconcile will not have an update conflict because version B does not exist yet. Version B is created during the reconcile of version A, rather than during the reconcile where the ClusterSync does not yet exist.

2uasimojo:
Ah, sorry, I just noticed the `return` immediately after the create. ++.

2uasimojo:
I'm like totally leery of adding an actual `/lgtm` now. But this lgtm.

2uasimojo:
/retest

infra blew up

csrwng:
aha, I vaguely remembered there was a reason we only did create (and no update) within a single reconcile :) sorry overlooked in the initial review.

PR #1112, opened 2020-09-02
GCP DNS cleanup after install failure

joelddiaz:
Have GCP dns cleanup be similar to how AWS is now handled (from https://github.com/openshift/hive/pull/1107 ).

Perform DNS cleanup after failed install for GCP.
    
Similar to how we try to clean up stray DNS records after a failed AWS installation (when the ClusterDeployment has ManagedDNS), do the same for GCP.
    
Share the existing zone cleanup code from the dnszone controller.
    
Actually make a GCP DNSZone for the DNSZone test for GCP
    
Consistently populate the DNSZone status after retriving the information from GCP API calls.

### DIFF
--- a/pkg/controller/dnszone/dnszone_controller_test.go
+++ b/pkg/controller/dnszone/dnszone_controller_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/golang/mock/gomock"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -20,6 +21,8 @@ import (
 	azuremock "github.com/openshift/hive/pkg/azureclient/mock"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	gcpmock "github.com/openshift/hive/pkg/gcpclient/mock"
+	testdnszone "github.com/openshift/hive/pkg/test/dnszone"
+	testgeneric "github.com/openshift/hive/pkg/test/generic"
 )
 
 // TestReconcileDNSProviderForAWS tests that ReconcileDNSProvider reacts properly under different reconciliation states on AWS.
@@ -269,8 +272,17 @@ func TestReconcileDNSProviderForGCP(t *testing.T) {
 			},
 		},
 		{
-			name:    "Delete managed zone",
-			dnsZone: validDNSZoneBeingDeleted(),
+			name: "Delete managed zone",
+			dnsZone: testdnszone.BasicBuilder().
+				Options(
+					testdnszone.WithGCPPlatform("testDNSZone"),
+				).
+				GenericOptions(
+					testgeneric.WithNamespace("testNamespace"),
+					testgeneric.WithName("testDNSZone"),
+					testgeneric.Deleted(),
+				).
+				Build(),
 			setupGCPMock: func(expect *gcpmock.MockClientMockRecorder) {
 				mockGCPZoneExists(expect)
 				mockDeleteGCPZone(expect)
@@ -328,14 +340,15 @@ func TestReconcileDNSProviderForGCP(t *testing.T) {
 			// This is necessary for the mocks to report failures like methods not being called an expected number of times.
 			defer mocks.mockCtrl.Finish()
 
-			setFakeDNSZoneInKube(mocks, tc.dnsZone)
+			err := setFakeDNSZoneInKube(mocks, tc.dnsZone)
+			require.NoError(t, err, "failed to create DNSZone into fake client")
 
 			if tc.setupGCPMock != nil {
 				tc.setupGCPMock(mocks.mockGCPClient.EXPECT())
 			}
 
 			// Act
-			_, err := r.reconcileDNSProvider(zr, tc.dnsZone)
+			_, err = r.reconcileDNSProvider(zr, tc.dnsZone)
 
 			// Assert
 			if tc.errorExpected {
--- a/pkg/controller/dnszone/gcpactuator.go
+++ b/pkg/controller/dnszone/gcpactuator.go
@@ -85,23 +85,33 @@ func (a *GCPActuator) Create() error {
 
 	logger.Debug("Managed zone successfully created")
 	a.managedZone = managedZone
+	if err := a.ModifyStatus(); err != nil {
+		logger.WithError(err).Error("failed to sync DNSZone status fields")
+		return err
+	}
+
 	return nil
 }
 
 // Delete implements the Delete call of the actuator interface
 func (a *GCPActuator) Delete() error {
-	if a.managedZone == nil {
-		return errors.New("managedZone is unpopulated")
+	if a.dnsZone.Status.GCP == nil {
+		return errors.New("deleting non-GCP DNSZone with GCP actuator")
 	}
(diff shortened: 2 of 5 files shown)


joelddiaz:
/hold
wait for https://github.com/openshift/hive/pull/1107

pkg/installmanager/dnscleanup.go
```diff
@@ -63,3 +67,35 @@ func cleanupAWSDNSZone(dnsZone *hivev1.DNSZone, region string, logger log.FieldL
...
+func cleanupGCPDNSZone(dnsZone *hivev1.DNSZone, logger log.FieldLogger) error {
+	if dnsZone.Status.GCP == nil {
+		return fmt.Errorf("found non-GCP DNSZone for DNS ClusterDeployment")
+	}
+	if dnsZone.Status.GCP.ZoneName == nil {
+		// Shouldn't happen as we block installs until DNS is ready
+		return fmt.Errorf("DNSZone %s has no ZoneName set", dnsZone.Name)
+	}
+
+	zoneLogger := logger.WithField("zoneName", *dnsZone.Status.GCP.ZoneName)
```
staebler, line 80: Just change `logger`. Otherwise, you need to also change all of the other uses of `logger` in this function to use `zoneLogger` instead.
```suggestion
	logger = logger.WithField("zoneName", *dnsZone.Status.GCP.ZoneName)
```

pkg/installmanager/installmanager.go
```diff
@@ -583,7 +583,12 @@ func cleanupFailedProvision(dynClient client.Client, cd *hivev1.ClusterDeploymen
 		if err != nil {
 			return err
 		}
-		return uninstaller.Run()
+
+		if err := uninstaller.Run(); err != nil {
+			return err
+		}
+
+		return cleanupDNSZone(dynClient, cd, logger)
```
staebler, line 591: Move this out of the switch so that it is called for all platforms.

pkg/test/dnszone/dnszone.go
```diff
@@ -115,3 +116,12 @@ func WithZone(zone string) Option {
 		dnsZone.Spec.Zone = zone
 	}
 }
+
+func WithGCPPlatform() Option {
+	return func(dnsZone *hivev1.DNSZone) {
+		dnsZone.Spec.GCP = &hivev1.GCPDNSZoneSpec{}
+		dnsZone.Status.GCP = &hivev1.GCPDNSZoneStatus{
+			ZoneName: pointer.StringPtr("testZoneName"),
```
staebler, line 124: Pass "testZoneName" in as a parameter to `WithGCPPlatform` rather than assuming that the tests using this want the zone name to the "testZoneName".

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -96,7 +101,8 @@ func (a *GCPActuator) Delete() error {
 
 	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone).WithField("zoneName", a.managedZone.Name)
```
staebler, line 102: The `Delete` function verifies that `a.managedZone` is not nil but does not use `a.managedZone` for anything other than the logger fields. The `Delete` function **does** use `a.dnsZone.Status.GCP.ZoneName`, but never verifies that it is not nil.
```suggestion
	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone).WithField("zoneName", *a.dnsZone.Status.GCP.ZoneName)
```

staebler: Sorry, `a.managedZone` **is** used. But it should be consistent in the `Delete` function. Use either `a.managedZone.Name` or `a.dnsZone.Status.GCP.ZoneName`. I suggest that the `managedZone` field should be removed entirely from the actuator.

pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -269,8 +276,17 @@ func TestReconcileDNSProviderForGCP(t *testing.T) {
...
+			dnsZone: testdnszone.BasicBuilder().
+				Options(
+					testdnszone.WithGCPPlatform(),
+				).
+				GenericOptions(
+					testgeneric.WithNamespace("testNamespace"),
+					testgeneric.WithName("testDNSZone"),
+					testgeneric.Deleted(),
+				).
+				Build(),
```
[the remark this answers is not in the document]
joelddiaz: I believe it was the resourceVersion being set that was causing problems when trying to register it into the fake kubeclient.

pkg/controller/dnszone/gcpactuator.go
joelddiaz: I think the primary utility is that it holds the list of nameservers to satisfy the actuator's GetNameServers() call. We could store it as another struct field after creating/fetching the DNSZone though.

pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -223,6 +227,9 @@ func TestReconcileDNSProviderForGCP(t *testing.T) {
 
 	log.SetLevel(log.DebugLevel)
 
+	hiveScheme := runtime.NewScheme()
```
staebler, line 230: This isn't used.

pkg/controller/dnszone/dnszone_controller_test.go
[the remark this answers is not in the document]
staebler: Ah, I see. It is because the fake kubeclient is not created with the DNSZone. The DNSZone is added later via a `Create` call.

pkg/controller/dnszone/gcpactuator.go
staebler: OK. That makes sense. I am uneasy about there being two different places from which information is obtained in the actuators--either from the `managedZone` in the actuator or from the status of the `DNSZone`. That is going to make it harder to maintain in the future because we are going to need to make sure that both are kept in sync. But I can live with it for now.

pkg/test/dnszone/dnszone.go
```diff
@@ -115,3 +116,12 @@ func WithZone(zone string) Option {
 		dnsZone.Spec.Zone = zone
 	}
 }
+
+func WithGCPPlatform(zoneName string) Option {
+	return func(dnsZone *hivev1.DNSZone) {
+		dnsZone.Spec.GCP = &hivev1.GCPDNSZoneSpec{}
+		dnsZone.Status.GCP = &hivev1.GCPDNSZoneStatus{
+			ZoneName: pointer.StringPtr(zoneName),
```
staebler, line 124:
```suggestion
			ZoneName: &zoneName,
```

staebler (commented):
Just two minor items.

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -85,18 +85,27 @@ func (a *GCPActuator) Create() error {
...
+	if err := a.ModifyStatus(); err != nil {
+		logger.WithError(err).Error("failed to sync DNSZone status fields")
+		return err
+	}
+
 	return nil
 }
 
 // Delete implements the Delete call of the actuator interface
 func (a *GCPActuator) Delete() error {
```
staebler, line 97: This function is still using `a.managedZone.Name` when calling `DeleteManagedZone`. It should use `*a.dnsZone.Status.GCP.ZoneName` instead. It may be helpful to create a `zoneName` variable early in the function and just use that instead.

joelddiaz:
/unhold
#1107 merged; now rebased

staebler (approved):
/lgtm

PR #1113, opened 2020-09-02
Add Velero to OLM Artifact Template

dustman9000:
This adds the velero backup to the olm artifact template and defaults to disabled.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -20,6 +20,8 @@ parameters:
   value: "[]"
 - name: DELETE_PROTECTION
   value: ""
+- name: VELERO_ENABLED
+  value: false
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
@@ -62,3 +64,6 @@ objects:
     globalPullSecretRef:
       name: ${GLOBAL_PULL_SECRET}
     deleteProtection: ${DELETE_PROTECTION}
+    backup:
+      velero:
+        enabled: ${VELERO_ENABLED}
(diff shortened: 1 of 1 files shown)


gregsheremeta:
/cc @twiest 
looks good to me, but would like Thomas to ack as well

jharrington22:
Also lgtm but will wait for @twiest

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -20,6 +20,10 @@ parameters:
   value: "[]"
 - name: DELETE_PROTECTION
   value: ""
+- name: BACKUP_VELERO_ENABLED
+  value: false
```
gregsheremeta, line 24: we might need double quotes here

cc @dustman9000 @jharrington22 

I've seen the pipeline complain about this in the past, on c-i-o stuff that Eric and Candace were working on

gregsheremeta:
not a code change and needed asap, forcing

hack/olm-registry/olm-artifacts-template.yaml
jharrington22: Nice catch @gregsheremeta `oc template` doesn't like the non quoted value.

```
error: failed to read input object (not a Template?): unable to decode "./hack/olm-registry/olm-artifacts-template.yaml": v1.Template.Parameters: []v1.Parameter: v1.Parameter.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"na|..., bigger context ...|lue":""},{"name":"BACKUP_VELERO_ENABLED","value":false},{"name":"BACKUP_VELERO_NAMESPACE","value":"o|...
```

PR #1115, opened 2020-09-02
Azure dns cleanup after install failure and retire ModifyStatus from Actuator

joelddiaz:
Cleanup DNS after install failure (like we do for AWS and GCP).

Update status as soon as we fetch the Azure DNS zone info (a no-op today).

Create new azureclient builder that takes the credentials as a byte array.

Additionally, retire the ModifyStatus function from the DNS actuator. Use it internally for all actuators.

### DIFF
--- a/pkg/controller/dnszone/actuator.go
+++ b/pkg/controller/dnszone/actuator.go
@@ -3,6 +3,7 @@ package dnszone
 // Actuator interface is the interface that is used to add dns provider support to the dnszone controller.
 type Actuator interface {
 	// Create tells the actuator to make a zone in the dns provider.
+	// The platform-specific DNSZone status fields will be populated (eg the zone name).
 	Create() error
 
 	// Delete tells the actuator to remove the zone from the dns provider.
@@ -14,14 +15,12 @@ type Actuator interface {
 	// UpdateMetadata tells the actuator to update the zone's metadata in the dns provider.
 	UpdateMetadata() error
 
-	// ModifyStatus allows the actuator to modify the kube DnsZoneStatus object before it is committed to kube.
-	ModifyStatus() error
-
 	// GetNameServers returns a list of nameservers that service the zone in the dns provider.
 	GetNameServers() ([]string, error)
 
 	// Refresh signals to the actuator that it should get the latest version of the zone from the dns provider.
 	// Refresh MUST be called before any other function is called by the actuator.
+	// Refresh will update the DNSZone object's platform-specific status fields.
 	Refresh() error
 
 	// SetConditionsForError sets conditions on the dnszone given a specific error
--- a/pkg/controller/dnszone/azureactuator.go
+++ b/pkg/controller/dnszone/azureactuator.go
@@ -73,6 +73,10 @@ func (a *AzureActuator) Create() error {
 
 	logger.Debug("Managed zone successfully created")
 	a.managedZone = &managedZone
+	if err := a.ModifyStatus(); err != nil {
+		logger.WithError(err).Error("failed to modify DNSZone status")
+		return err
+	}
 	return nil
 }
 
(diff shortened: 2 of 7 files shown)


joelddiaz:
/hold
wait on #1112 to limit conflicts

pkg/controller/dnszone/azureactuator.go
```diff
@@ -85,7 +89,8 @@ func (a *AzureActuator) Delete() error {
 	resourceGroupName := a.dnsZone.Spec.Azure.ResourceGroupName
 	logger := a.logger.WithField("zone", a.dnsZone.Spec.Zone)
 
-	if err := a.deleteRecordsets(logger); err != nil {
+	logger.Info("Deleting recordsets in managedzone")
+	if err := DeleteAzureRecordSets(a.azureClient, a.dnsZone, logger); err != nil {
```
staebler, line 93: Be consistent between using `a.dnsZone.Status.Azure` and `a.managedZone`.

pkg/installmanager/dnscleanup.go
```diff
@@ -69,6 +72,27 @@ func cleanupAWSDNSZone(dnsZone *hivev1.DNSZone, region string, logger log.FieldL
...
 }
 
+// cleanupAzureDNSZone will return a DNS zone to the minimum set of DNS records
+func cleanupAzureDNSZone(dnsZone *hivev1.DNSZone, logger log.FieldLogger) error {
+	zoneLogger := logger.WithField("dnsZoneID", dnsZone.Spec.Zone)
+	zoneLogger.Info("cleaning up DNSZone")
+
+	creds, err := azureutils.GetCreds("")
+	if err != nil {
+		logger.WithError(err).Error("failed to get Azure creds")
```
staebler, line 82:
```suggestion
		zoneLogger.WithError(err).Error("failed to get Azure creds")
```
Although, I would prefer just overwriting the `logger` variable instead of having a `zoneLogger` variable.

pkg/installmanager/dnscleanup.go
```diff
@@ -69,6 +72,27 @@ func cleanupAWSDNSZone(dnsZone *hivev1.DNSZone, region string, logger log.FieldL
...
+		return err
+	}
+
+	azureClient, err := azureclient.NewClient(creds)
+
+	if err := dns.DeleteAzureRecordSets(azureClient, dnsZone, zoneLogger); err != nil {
+		logger.WithError(err).Error("failed to clean up DNS Zone")
```
staebler, line 89:
```suggestion
		zoneLogger.WithError(err).Error("failed to clean up DNS Zone")
```

pkg/installmanager/installmanager.go
```diff
@@ -563,7 +563,11 @@ func cleanupFailedProvision(dynClient client.Client, cd *hivev1.ClusterDeploymen
 		uninstaller.GraphAuthorizer = session.GraphAuthorizer
 		uninstaller.Authorizer = session.Authorizer
 
-		return uninstaller.Run()
+		if err := uninstaller.Run(); err != nil {
+			return err
+		}
+
+		return cleanupDNSZone(dynClient, cd, logger)
```
staebler, line 570: Move outside switch.

pkg/controller/dnszone/azureactuator.go
```diff
@@ -73,6 +73,10 @@ func (a *AzureActuator) Create() error {
 
 	logger.Debug("Managed zone successfully created")
 	a.managedZone = &managedZone
+	if err := a.ModifyStatus(); err != nil {
```
staebler, line 76: This is the last actuator to do `ModifyStatus` in the `Create` and `Refresh` functions. Can we make `ModifyStatus` private now and remove the call from the dnszone controller?

pkg/controller/dnszone/azureactuator.go
joelddiaz: For AWS and GCP we actually store some information about the DNS Zone (the ZoneID and ZoneName respectively). For Azure, the dnszone.Spec.Name **is** the name of the object up in Azure. I originally wrote up some code to add a status field for Azure DNSZones, but it wasn't storing anything useful. WDYT?

pkg/installmanager/installmanager.go
joelddiaz: Yes. I did that in the GCP PR, so this PR should just inherit that change in flow.

pkg/controller/dnszone/azureactuator.go
staebler: No, don't put anything in the status. My original comment should have said "be consistent between using `a.dnsZone.Spec.Zone` and `a.managedZone`". Like with the GCP PR, I want you to use `a.dnsZone.Spec.Zone` everywhere in the `Delete` function rather than using `*a.managedZone.Name` in some places.

joelddiaz:
/unhold
#1112 merged

pkg/controller/dnszone/actuator.go
```diff
@@ -3,6 +3,7 @@ package dnszone
 // Actuator interface is the interface that is used to add dns provider support to the dnszone controller.
 type Actuator interface {
 	// Create tells the actuator to make a zone in the dns provider.
+	// The platform-specific DNSZone status fields will be populated (ie the zone name).
```
staebler, line 6:
```suggestion
	// The platform-specific DNSZone status fields will be populated (eg the zone name).
```

staebler (commented):
Only a very minor nit. Also, the third word of the last commit should be "an" instead of "and".

staebler (approved):
/lgtm

PR #1092, opened 2020-08-20
CO-1114. Handle different worker-user-data secret name for 4.6.

meowfaceman:
The worker-user-data secret name became worker-user-data-managed in 4.6.
Hive now supports alternate names for this secret.

PR #1116, opened 2020-09-02
CO-1114. Handle different worker-user-data secret name for 4.6

staebler:
The worker-user-data secret name became worker-user-data-managed in 4.6.
Hive now supports alternate names for this secret.

This builds on https://github.com/openshift/hive/pull/1092. I was not able to push my changes into Michael's branch.


staebler:
/lgtm cancel

This does not work for pre-releases of 4.6.

staebler:
b98126cb...333ea957
* Added a step in the `workerUserData` for stripping off the pre-release and build number from the OpenShift version prior to comparing it to 4.6.0 so that 4.6.0 pre-releases use the new name for the worker user data secret.

PR #1117, opened 2020-09-02
fix velero template in pipeline template

gregsheremeta:


gregsheremeta:
tested locally, works, forcing

PR #1119, opened 2020-09-03
controller: fix syncset applies not going immediately after ClusterSync create

staebler:
With commit 0c17526e1ffc6f562c3b83d6d6e57588f0b245e6, the ClusterSync is created as a separate clustersync reconcile. However, the clustersync controller does not watch for changes to ClusterSync. Consequently, the syncsets are not applied to a new cluster until there is some later change to the ClusterDeployment or one of the syncsets. These changes will queue the ClusterDeployment when the ClusterSync is created.

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -22,8 +22,10 @@ import (
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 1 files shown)


csrwng:
@staebler wouldn't it be simpler to simply ignore an `AlreadyExists` error when creating the clustersync?

staebler:
> @staebler wouldn't it be simpler to simply ignore an `AlreadyExists` error when creating the clustersync?

Yep, it would be. Let me write that up.

staebler:
@csrwng I have redone the commit so that the only thing added for this PR is code to requeue the ClusterDeployment when the ClusterSync is created. The controller will not ignore an `AlreadyExists` error, but the controller will abort the reconcile and requeue with an error until the ClusterSync is in the local cache.

PR #1120, opened 2020-09-04
pipeline: lock registry container to 4.5

gregsheremeta:
the :latest container appears broken. Probably a bad idea
to chase :latest in this build anyway.


gregsheremeta:
fyi @jmelis

gregsheremeta:
pipeline broken, forcing

PR #1121, opened 2020-09-04
use image from master machine for new machinesets 

staebler:
Each remotemachineset actuator had its own way of determining which OS image to use when generating machinesets. Some actuators used a hard-coded value, some actuators looked into existing machinesets, and some actuators search the cloud provider for images. With these changes, all actuators will instead get the OS image from an existing master machine on the target cluster.

https://issues.redhat.com/browse/CO-1098


staebler:
This builds on #1116.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -275,10 +304,37 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
+) (*machineapi.Machine, error) {
+	remoteMachines := &machineapi.MachineList{}
+	tm := metav1.TypeMeta{}
+	tm.SetGroupVersionKind(machineapi.SchemeGroupVersion.WithKind("Machine"))
+	if err := remoteClusterAPIClient.List(
+		context.Background(),
+		remoteMachines,
+		&client.ListOptions{
+			Raw: &metav1.ListOptions{
+				TypeMeta:      tm,
```
csrwng, line 320: Just curious, why is the type meta needed here?

pkg/controller/remotemachineset/openstackactuator_test.go
```diff
@@ -18,7 +18,8 @@ import (
 	hivev1osp "github.com/openshift/hive/pkg/apis/hive/v1/openstack"
 )
 
-func TestOpenStackActuator(t *testing.T) {
+// This test is broken! The installer now checks for trunk support by querying the OpenStack service.
+func BROKEN__TestOpenStackActuator(t *testing.T) {
```
abutcher, line 22: Do we need to create a card to fix this?

csrwng:
lgtm, just one question

pkg/controller/remotemachineset/remotemachineset_controller.go
staebler: I don't know. It was done that way with the list of machinesets. I planned on checking whether it was actually needed, but got all my time sucked away by digging out of dependency hell.

pkg/controller/remotemachineset/openstackactuator_test.go
staebler: Yes, although, I am not sure what we are going to be able to do. There is one more unit test that I needed to comment out, too, that I did not have time to figure out how to fix.

PR #1122, opened 2020-09-04
Implement hibernate-after annotation to auto-powerdown.

dgoodwin:
When a cluster has been in a running state (installed with no hibernate
condition, or hibernate condition=false) for more than the
hibernate-after duration, the cluster will be automatically put to
sleep.

Intended for dev clusters where you do not want them to accidentally be
left running beyond the workday.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -10,6 +10,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -131,6 +132,8 @@ type Options struct {
 	ReleaseImage             string
 	ReleaseImageSource       string
 	DeleteAfter              string
+	HibernateAfter           string
+	HibernateAfterDur        *time.Duration
 	ServingCert              string
 	ServingCertKey           string
 	UseClusterImageSet       bool
@@ -243,6 +246,7 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=ovirt --ovirt-api-vip 192.168.1.2
 	flags.StringVar(&opt.BaseDomain, "base-domain", "new-installer.openshift.com", "Base domain for the cluster")
 	flags.StringVar(&opt.PullSecret, "pull-secret", "", "Pull secret for cluster. Takes precedence over pull-secret-file.")
 	flags.StringVar(&opt.DeleteAfter, "delete-after", "", "Delete this cluster after the given duration. (i.e. 8h)")
+	flags.StringVar(&opt.HibernateAfter, "hibernate-after", "", "Automatically hibernate the cluster whenever it has been running for the given duration")
 	flags.StringVar(&opt.PullSecretFile, "pull-secret-file", defaultPullSecretFile, "Pull secret file for cluster")
 	flags.StringVar(&opt.CredsFile, "creds-file", "", "Cloud credentials file (defaults vary depending on cloud)")
 	flags.StringVar(&opt.ClusterImageSet, "image-set", "", "Cluster image set to use for this cluster deployment")
@@ -324,6 +328,14 @@ func (o *Options) Complete(cmd *cobra.Command, args []string) error {
 		case cloudGCP:
 			o.Region = "us-east1"
 		}
+	}
+
+	if o.HibernateAfter != "" {
+		dur, err := time.ParseDuration(o.HibernateAfter)
+		if err != nil {
+			return errors.Wrapf(err, "unable to parse HibernateAfter duration")
+		}
+		o.HibernateAfterDur = &dur
 	}
 
 	return nil
@@ -509,6 +521,7 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
 		BaseDomain:            o.BaseDomain,
 		ManageDNS:             o.ManageDNS,
 		DeleteAfter:           o.DeleteAfter,
+		HibernateAfter:        o.HibernateAfterDur,
 		Labels:                labels,
 		InstallerManifests:    manifestFileData,
 		MachineNetwork:        o.MachineNetwork,
--- a/pkg/apis/hive/v1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1/clusterdeployment_types.go
@@ -123,6 +123,12 @@ type ClusterDeploymentSpec struct {
 	// PowerState defaults to the Running state.
 	// +optional
 	PowerState ClusterPowerState `json:"powerState,omitempty"`
+
+	// HibernateAfter will transition a cluster to hibernating power state after it has been running for the
+	// given duration. The time that a cluster has been running is the time since the cluster was installed or the
+	// time since the cluster last came out of hibernation.
+	// +optional
+	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
 }
 
 // Provisioning contains settings used only for initial cluster provisioning.
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -122,7 +122,7 @@ func AddToManager(mgr manager.Manager, r *hibernationReconciler, concurrentRecon
 }
 
 // Reconcile syncs a single ClusterDeployment
-func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *hibernationReconciler) Reconcile(request reconcile.Request) (result reconcile.Result, returnErr error) {
 	start := time.Now()
 	cdLog := r.logger.WithFields(log.Fields{
 		"controller":        ControllerName,
@@ -164,6 +164,64 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
 	shouldHibernate := cd.Spec.PowerState == hivev1.HibernatingClusterPowerState
 	hibernatingCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
 
+	// Signal a problem if we should be hibernating or have requested hibernate after and the cluster does not support it.
+	if shouldHibernate || cd.Spec.HibernateAfter != nil {
+		if supported, msg := r.canHibernate(cd); !supported {
+			return r.setHibernatingCondition(cd, hivev1.UnsupportedHibernationReason, msg, corev1.ConditionFalse, cdLog)
+		}
+	}
+
+	// Check if HibernateAfter is set, and if the cluster has been in running state for longer than this duration, put it to sleep.
+	if cd.Spec.HibernateAfter != nil && cd.Spec.PowerState != hivev1.HibernatingClusterPowerState {
+		hibernateAfterDur := cd.Spec.HibernateAfter.Duration
+		runningSince := cd.Status.InstalledTimestamp.Time
+		hibLog := cdLog.WithFields(log.Fields{
+			"runningSince":   runningSince,
+			"hibernateAfter": hibernateAfterDur,
+		})
+		var isRunning bool
+
+		cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
+		if cond == nil {
+			hibLog.Debug("cluster has no hibernating condition (never hibernated), using installed time")
+			isRunning = true
+		} else if cond.Status == corev1.ConditionFalse {
+			runningSince = cond.LastTransitionTime.Time
+			hibLog = hibLog.WithField("runningSince", runningSince)
+			hibLog.WithField("reason", cond.Reason).Debug("hibernating condition false")
+			isRunning = true
+		}
+
+		if isRunning {
+			expiry := runningSince.Add(hibernateAfterDur)
+			hibLog.Debugf("cluster should be hibernating after: %s", expiry)
+			if time.Now().After(expiry) {
+				hibLog.WithField("expiry", expiry).Debug("cluster has been running longer than hibernate-after duration, moving to hibernating powerState")
+				cd.Spec.PowerState = hivev1.HibernatingClusterPowerState
+				err := r.Update(context.TODO(), cd)
+				if err != nil {
+					hibLog.WithError(err).Log(controllerutils.LogLevel(err), "error hibernating cluster")
+				}
+				return reconcile.Result{}, err
+			}
+
+			defer func() {
+				requeueNow := result.Requeue && result.RequeueAfter <= 0
+				if returnErr == nil && !requeueNow {
+					// We have an hibernate after time but cluster has not been running that long yet.
+					// Set requeueAfter for just after so that we requeue cluster for hibernation once reconcile has completed
+					requeueAfter := time.Until(expiry)
+					if requeueAfter < result.RequeueAfter || result.RequeueAfter <= 0 {
+						hibLog.Infof("cluster will reconcile due to hibernate-after time in: %v", requeueAfter)
+						result.RequeueAfter = requeueAfter
+						result.Requeue = true
+					}
+				}
+			}()
+		}
+
+	}
+
 	if !shouldHibernate {
 		if hibernatingCondition == nil || hibernatingCondition.Status == corev1.ConditionFalse {
 			return reconcile.Result{}, nil
@@ -177,9 +235,6 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
-	if supported, msg := r.canHibernate(cd); !supported {
-		return r.setHibernatingCondition(cd, hivev1.UnsupportedHibernationReason, msg, corev1.ConditionFalse, cdLog)
-	}
 	if hibernatingCondition == nil || hibernatingCondition.Status == corev1.ConditionFalse || hibernatingCondition.Reason == hivev1.ResumingHibernationReason {
 		return r.stopMachines(cd, cdLog)
 	}
--- a/pkg/controller/hibernation/hibernation_controller_test.go
+++ b/pkg/controller/hibernation/hibernation_controller_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/golang/mock/gomock"
 	log "github.com/sirupsen/logrus"
@@ -13,6 +14,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	certsv1beta1 "k8s.io/api/certificates/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	fakekubeclient "k8s.io/client-go/kubernetes/fake"
@@ -22,6 +24,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
+
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/controller/hibernation/mock"
 	"github.com/openshift/hive/pkg/remoteclient"
@@ -83,7 +86,7 @@ func TestReconcile(t *testing.T) {
 		},
 		{
 			name: "start hibernating, older version",
-			cd:   cdBuilder.Options(o.shouldHibernate, o.olderVersion).Build(),
+			cd:   cdBuilder.Options(o.shouldHibernate, testcd.WithClusterVersion("4.3.11")).Build(),
 			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
 				cond := getHibernatingCondition(cd)
 				require.NotNil(t, cond)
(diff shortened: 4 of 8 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	// Check for the hibernate-after annotation, and if the cluster has been in running state for longer than this duration, put it to sleep.
+	hibernateAfter, ok := cd.Annotations[constants.HibernateAfterAnnotation]
+	if ok && cd.Spec.PowerState != hivev1.HibernatingClusterPowerState {
+		cdLog.Debugf("found hibernate-after annotation: %s", hibernateAfter)
+		dur, err := time.ParseDuration(hibernateAfter)
```
staebler, line 156: Use a more descriptive variable name for `dur`, since it is used later when the reader may have forgotten what the variable is.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+				}
+				return reconcile.Result{}, err
+			}
+
+			defer func() {
+				requeueNow := result.Requeue && result.RequeueAfter <= 0
+				if returnErr == nil && !requeueNow {
+					// We have an hibernate after time but cluster has not been running that long yet.
+					// Set requeueAfter for just after so that we requeue cluster for hibernation once reconcile has completed
+					requeueAfter := time.Until(expiry) + 60*time.Second
```
staebler, line 198: Why add 60 seconds?

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+					if requeueAfter < result.RequeueAfter || result.RequeueAfter <= 0 {
+						hibLog.Infof("cluster will reconcile due to hibernate-after time in: %v", requeueAfter)
+						result.RequeueAfter = requeueAfter
```
staebler, line 201: Set `result.Requeue` to `true` also to cover the scenario (unlikely with the 60-second addition) that `requeueAfter` is not positive.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+
+	cdBuilder := testcd.FullBuilder(namespace, cdName, scheme).Options(
+		func(cd *hivev1.ClusterDeployment) {
+			cd.Spec.Installed = true
+			cd.Status.ClusterVersionStatus = &configv1.ClusterVersionStatus{
+				Desired: configv1.Update{
+					Version: "4.4.9",
+				},
+			}
+		},
```
staebler, line 314:
```suggestion
		testcd.Installed(),
		func(cd *hivev1.ClusterDeployment) {
			cd.Status.ClusterVersionStatus = &configv1.ClusterVersionStatus{
				Desired: configv1.Update{
					Version: "4.4.9",
				},
			}
		},
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+		installedTime      time.Time
+		hibernateAfter     string
+		hibernateCondition hivev1.ClusterDeploymentCondition
+		expectRequeueAfter bool
+		expectedPowerState hivev1.ClusterPowerState
+	}{
+		{
+			name: "cluster due for hibernate no condition", // cluster that has never been hibernated and thus has no running condition
+			cd: cdBuilder.GenericOptions(testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Options(
+				testcd.InstalledTimestamp(time.Now().Add(-10 * time.Hour))).Build(),
```
staebler, line 332:
```suggestion
			cd: cdBuilder.GenericOptions(testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Build(
				testcd.InstalledTimestamp(time.Now().Add(-10 * time.Hour))),
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+			expectRequeueAfter: true,
+			expectedPowerState: "",
+		},
+		{
+			name: "cluster with running condition due for hibernate",
+			cd: cdBuilder.GenericOptions(
+				testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Options(
+				testcd.WithCondition(hibernatingCondition(corev1.ConditionFalse, hivev1.RunningHibernationReason, 9*time.Hour)),
+			).Options(
+				testcd.InstalledTimestamp(time.Now().Add(-10 * time.Hour))).Build(),
```
staebler, line 348:
```suggestion
			cd: cdBuilder.GenericOptions(
				testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Build(
				testcd.WithCondition(hibernatingCondition(corev1.ConditionFalse, hivev1.RunningHibernationReason, 9*time.Hour),
				testcd.InstalledTimestamp(time.Now().Add(-10 * time.Hour)))
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+			expectRequeueAfter: true,
+			expectedPowerState: "",
+		},
+		{
+			name: "cluster stuck resuming condition due for hibernate",
+			cd: cdBuilder.GenericOptions(
+				testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Options(
+				// TODO: check false below
```
staebler, line 365: ?

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+				testcd.WithCondition(hibernatingCondition(corev1.ConditionFalse, hivev1.ResumingHibernationReason, 9*time.Hour)),
+			).Options(
+				testcd.InstalledTimestamp(time.Now().Add(-10 * time.Hour))).Build(),
+			expectedPowerState: hivev1.HibernatingClusterPowerState,
+		},
+		{
+			name: "cluster failed resuming condition due for hibernate",
+			cd: cdBuilder.GenericOptions(
+				testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Options(
+				// TODO:
```
staebler, line 375: ?

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+				testcd.InstalledTimestamp(time.Now().Add(-10 * time.Hour))).Build(),
+			expectRequeueAfter: true,
+			expectedPowerState: "",
+		},
+		{
+			name: "cluster stuck resuming condition due for hibernate",
+			cd: cdBuilder.GenericOptions(
+				testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "8h")).Options(
+				// TODO: check false below
+				testcd.WithCondition(hibernatingCondition(corev1.ConditionFalse, hivev1.ResumingHibernationReason, 9*time.Hour)),
```
staebler, line 366: If the cluster is resuming, then the Hibernating condition will have a status of True.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+				actuator.EXPECT().MachinesRunning(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(false, nil)
+			},
+			cd: cdBuilder.GenericOptions(
+				testgeneric.WithAnnotation(constants.HibernateAfterAnnotation, "4h")).Options(
+				testcd.WithCondition(hibernatingCondition(corev1.ConditionTrue, hivev1.ResumingHibernationReason, 8*time.Hour)),
+			).Options(
+				testcd.InstalledTimestamp(time.Now().Add(-10*time.Hour)),
+				o.shouldRun).Build(),
+			expectedPowerState: hivev1.RunningClusterPowerState,
+			expectRequeueAfter: true,
```
staebler, line 393: This test case highlights a deficiency in the tests. Since the test only checks whether the requeue-after is positive, the test does not distinguish between a requeue-after coming from the hibernate-after annotation or a requeue-after for other reasons. This test case should be verifying that the requeue-after did not get set as a result of the hibernate-after annotation, but it does not verify that. The requeue-after is getting set to `stateCheckInterval`.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+	}
+}
+
+func hibernatingCondition(status corev1.ConditionStatus, reason string, lastTransitionAgo time.Duration) hivev1.ClusterDeploymentCondition {
+	return hivev1.ClusterDeploymentCondition{
+		Type:               hivev1.ClusterHibernatingCondition,
+		Status:             status,
+		Message:            "unused",
+		Reason:             reason,
+		LastTransitionTime: metav1.Time{Time: time.Now().Add(-lastTransitionAgo)},
```
staebler, line 444:
```suggestion
		LastTransitionTime: metav1.NewTime(time.Now().Add(-lastTransitionAgo)),
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+				},
+				csrUtil: mockCSRHelper,
+			}
+			result, err := reconciler.Reconcile(reconcile.Request{
+				NamespacedName: types.NamespacedName{Namespace: namespace, Name: cdName},
+			})
+			if test.expectRequeueAfter {
+				assert.True(t, result.RequeueAfter > 0)
+			} else {
+				assert.Equal(t, 0*time.Second, result.RequeueAfter)
```
staebler, line 424:
```suggestion
				assert.Zero(t, result.RequeueAfter)
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+				remoteClientBuilder: func(cd *hivev1.ClusterDeployment) remoteclient.Builder {
+					return mockBuilder
+				},
+				csrUtil: mockCSRHelper,
+			}
+			result, err := reconciler.Reconcile(reconcile.Request{
+				NamespacedName: types.NamespacedName{Namespace: namespace, Name: cdName},
+			})
+			if test.expectRequeueAfter {
+				assert.True(t, result.RequeueAfter > 0)
```
staebler, line 422: Please add messages to asserts and requires.
```suggestion
				assert.True(t, result.RequeueAfter > 0, "expected positive requeue after")
```

staebler, line 422:
```suggestion
				assert.Greater(t, result.RequeueAfter.Seconds(), 0.)
```
so that the requeue-after value shows up in the test failure.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,158 @@ func TestReconcile(t *testing.T) {
...
+				assert.Equal(t, 0*time.Second, result.RequeueAfter)
+			}
+
+			assert.Nil(t, err)
+			cd := &hivev1.ClusterDeployment{}
+			err = c.Get(context.TODO(), client.ObjectKey{Namespace: namespace, Name: cdName}, cd)
+			require.Nil(t, err)
+			assert.Equal(t, test.expectedPowerState, cd.Spec.PowerState)
+
+			ctrl.Finish()
```
staebler, line 433: Put this in a defer after creating the controller.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	// Check for the hibernate-after annotation, and if the cluster has been in running state for longer than this duration, put it to sleep.
+	hibernateAfter, ok := cd.Annotations[constants.HibernateAfterAnnotation]
+	if ok && cd.Spec.PowerState != hivev1.HibernatingClusterPowerState {
```
staebler, line 154: Also add a check for whether the cluster can handle hibernation. If the cluster cannot handle hibernation, then it is awkward to have Hive setting the power state to Hibernating.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+		cdLog.Debugf("found hibernate-after annotation: %s", hibernateAfter)
+		dur, err := time.ParseDuration(hibernateAfter)
+		if err != nil {
+			cdLog.WithError(err).WithField("hibernateAfter", hibernateAfter).Infof("error parsing %s as a duration", constants.HibernateAfterAnnotation)
+			return reconcile.Result{}, fmt.Errorf("error parsing %s as a duration: %v", constants.HibernateAfterAnnotation, err)
```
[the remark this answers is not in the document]
dgoodwin: Dropped the return and clarified we're ignoring the annotation in the log message. Left it an info, not sure if that should be a warning or not.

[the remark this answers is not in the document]
dgoodwin: Dropped the return and clarified we're ignoring the annotation in the log message. Left it an info, not sure if that should be a warning or not.

pkg/controller/hibernation/hibernation_controller.go
dgoodwin: This was following the precedent with delete-after in the clusterDeployment controller. https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L502

staebler: There is no reason to add 60 seconds in the clusterDeployment controller either. I don't care if we want to keep them: The only effect is waiting 60 seconds more than necessary before taking the desired action.

dgoodwin: Changed to 5, I assume we're just trying to make sure we don't accidentally reconcile before the timer but that does seem pretty unlikely.

pkg/controller/hibernation/hibernation_controller_test.go
dgoodwin: These tests were part of where I got into trouble on the first hack day I worked on this. We can't tell if you're stuck resuming or for how long because the hibernating condition stays true, so the transition time is never updated, it's still set to when you went to hibernating. I'm going to remove these tests, getting stuck in one of those phases is outside the scope of this I think.

pkg/controller/hibernation/hibernation_controller_test.go
dgoodwin: Moved this to fuzzy matching, asserting it's within 10s of expected.

pkg/controller/hibernation/hibernation_controller_test.go
dgoodwin: New output will be:

--- FAIL: TestHibernateAfter/cluster_not_yet_due_for_hibernate_no_condition (0.00s)
        hibernation_controller_test.go:403: 
            	Error Trace:	hibernation_controller_test.go:403
            	Error:      	Should be true
            	Test:       	TestHibernateAfter/cluster_not_yet_due_for_hibernate_no_condition
            	Messages:   	RequeueAfter (2h0m4.366034255s) duration is not within 10s of expected (2h0m0s)

pkg/controller/hibernation/hibernation_controller.go
dgoodwin: Went to implement this but then realized we should probably surface the problem if you set this annotation on a cluster that can't handle it. We can't really log that, nobody will ever see it, it should be a condition. If we leave it, we get that condition for free, when it procs we'll move a cluster to hibernating and it'll get a condition explaining why it won't work. Granted it's delayed until it's time to hibernate but nobody's going to look until something is wrong regardless. I think the current form might be better.

dgoodwin:
PR updated, thanks!

pkg/controller/hibernation/hibernation_controller.go
staebler: This is an example of where the lines getting blurred between the roles of spec and annotation. If something in the annotation is enough to surface a condition on the object, then it is certainly worthy of being in the spec proper. If does nobody any good to continue to put things like this into annotations where it only serves to make it harder for the user to find information about and use.

staebler: If the feature is for Hive developers, then the Hive developer should know enough about the annotation to not need a condition telling them that they did something unsupported. If the feature is for Hive users, then it should be in the spec.

pkg/controller/hibernation/hibernation_controller.go
staebler: 5 seconds or 60 seconds makes little difference. The additional time is unnecessary. If we reconcile before the timer, then the ClusterDeployment will be requeued for a short time in the future where the timer will likely have elapsed.

pkg/controller/hibernation/hibernation_controller_test.go
staebler: My expectation was that this change would be made for all of the test cases and not just the one that I called out as an example.

pkg/controller/hibernation/hibernation_controller_test.go
staebler: Likewise, here, I called out this as an *example* of a more concise way to do it.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -418,19 +398,20 @@ func TestHibernateAfter(t *testing.T) {
...
-				assert.True(t, result.RequeueAfter > 0)
-			} else {
-				assert.Equal(t, 0*time.Second, result.RequeueAfter)
+
+			// Need to do fuzzy requeue after matching
+			if assert.Nil(t, err) {
+				assert.True(t, test.expectRequeueAfter-10*time.Second < result.RequeueAfter &&
+					result.RequeueAfter < test.expectRequeueAfter+10*time.Second,
+					"RequeueAfter (%s) duration is not within 10s of expected (%s)",
+					result.RequeueAfter, test.expectRequeueAfter)
```
staebler, line 407:
```suggestion
				assert.GreaterOrEqual(t, result.RequeueAfter.Seconds(), (test.expectRequeueAfter-10*time.Second).Seconds(), "requeue after too small")
				assert.LessOrEqual(t, result.RequeueAfter.Seconds(), (test.expectRequeueAfter+10*time.Second).Seconds(), "request after too large")

```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -418,19 +398,20 @@ func TestHibernateAfter(t *testing.T) {
...
 			}
 
-			assert.Nil(t, err)
 			cd := &hivev1.ClusterDeployment{}
 			err = c.Get(context.TODO(), client.ObjectKey{Namespace: namespace, Name: cdName}, cd)
 			require.Nil(t, err)
```
staebler, line 412:
```suggestion
			require.NoError(t, err)
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -418,19 +398,20 @@ func TestHibernateAfter(t *testing.T) {
 			result, err := reconciler.Reconcile(reconcile.Request{
 				NamespacedName: types.NamespacedName{Namespace: namespace, Name: cdName},
 			})
-			if test.expectRequeueAfter {
-				assert.True(t, result.RequeueAfter > 0)
-			} else {
-				assert.Equal(t, 0*time.Second, result.RequeueAfter)
+
+			// Need to do fuzzy requeue after matching
+			if assert.Nil(t, err) {
```
staebler, line 403:
```suggestion
			if assert.NoError(t, err) {
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -418,19 +398,20 @@ func TestHibernateAfter(t *testing.T) {
...
+				assert.True(t, test.expectRequeueAfter-10*time.Second < result.RequeueAfter &&
```
staebler, line 404: If `expectedRequeueAfter` is zero, you should check expliictly for zero without the 10-second buffer.

pkg/controller/hibernation/hibernation_controller_test.go
staebler, line 403: Add message to the assert.

pkg/controller/hibernation/hibernation_controller_test.go
staebler, line 412: Add message to the assert.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -418,19 +398,20 @@ func TestHibernateAfter(t *testing.T) {
...
+					result.RequeueAfter < test.expectRequeueAfter+10*time.Second,
+					"RequeueAfter (%s) duration is not within 10s of expected (%s)",
+					result.RequeueAfter, test.expectRequeueAfter)
 			}
 
-			assert.Nil(t, err)
 			cd := &hivev1.ClusterDeployment{}
 			err = c.Get(context.TODO(), client.ObjectKey{Namespace: namespace, Name: cdName}, cd)
 			require.Nil(t, err)
 			assert.Equal(t, test.expectedPowerState, cd.Spec.PowerState)
```
staebler, line 413: Add message to the assert.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+			// Cluster is Hibernating, we have no runningSince time:
+			runningSince = time.Time{}
+		}
+
+		hibLog := cdLog.WithFields(log.Fields{
+			"runningSince":   runningSince,
+			"hibernateAfter": hibernateAfter,
+		})
+
+		if !runningSince.IsZero() {
```
staebler, line 179: Would like to see this done as a separate variable rather than relying on the implicit understanding that a zero value for `runningSince` means that the cluster is hibernating. In other words, create an `isHibernating` variable that is set to true if the cluster is hibernating.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -148,6 +149,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+			}
+
+			defer func() {
+				requeueNow := result.Requeue && result.RequeueAfter <= 0
+				if returnErr == nil && !requeueNow {
+					// We have an hibernate after time but cluster has not been running that long yet.
+					// Set requeueAfter for just after so that we requeue cluster for hibernation once reconcile has completed
+					requeueAfter := time.Until(expiry) + 5*time.Second
+					if requeueAfter < result.RequeueAfter || result.RequeueAfter <= 0 {
+						hibLog.Infof("cluster will reconcile due to hibernate-after time in: %v", requeueAfter)
```
staebler, line 199: Use debug level. This does not need to be output at Info level for every reconcile of a running ClusterDeployment with the annotation.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,139 @@ func TestReconcile(t *testing.T) {
...
 
+func TestHibernateAfter(t *testing.T) {
+	logger := log.New()
+	logger.SetLevel(log.DebugLevel)
+
+	scheme := runtime.NewScheme()
+	corev1.AddToScheme(scheme)
+	batchv1.AddToScheme(scheme)
+	hivev1.AddToScheme(scheme)
+	hivev1.AddToScheme(scheme)
```
staebler, line 304: Duplicate of previous line.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,139 @@ func TestReconcile(t *testing.T) {
 
 }
 
+func TestHibernateAfter(t *testing.T) {
+	logger := log.New()
+	logger.SetLevel(log.DebugLevel)
+
+	scheme := runtime.NewScheme()
+	corev1.AddToScheme(scheme)
+	batchv1.AddToScheme(scheme)
```
staebler, line 302: Unneeded. The batch api is not used by this controller.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,139 @@ func TestReconcile(t *testing.T) {
...
+	o := clusterDeploymentOptions{}
+
+	tests := []struct {
+		name          string
+		setupActuator func(actuator *mock.MockHibernationActuator)
+		cd            *hivev1.ClusterDeployment
+
+		installedTime      time.Time
+		hibernateAfter     string
+		hibernateCondition hivev1.ClusterDeploymentCondition
```
staebler, line 325: All three of these are unused.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -289,6 +293,139 @@ func TestReconcile(t *testing.T) {
...
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			mockActuator := mock.NewMockHibernationActuator(ctrl)
+			mockActuator.EXPECT().CanHandle(gomock.Any()).AnyTimes().Return(true)
+			if test.setupActuator != nil {
+				test.setupActuator(mockActuator)
+			}
+			mockBuilder := remoteclientmock.NewMockBuilder(ctrl)
+			mockCSRHelper := mock.NewMockcsrHelper(ctrl)
+			actuators = []HibernationActuator{mockActuator}
```
staebler, line 387: I know that this is carried over from the existing hibernate tests, and I don't expect you to do anything about it, but this makes it impossible to run the tests in parallel.

pkg/controller/hibernation/hibernation_controller_test.go
dgoodwin: I will update but this should probably be discussed as a team, I am not in favor of requiring text for every assert when the code is self explanatory, it's just overly verbose and more churn in the review process.

dgoodwin:
This is updated for feedback, to move it to Spec.HibernateAfterMinutes, and to add a hiveutil create-cluster option. I am still tracing down why locally creating a cluster with it has no affect and applying the yaml again does.
/hold

dgoodwin:
This needs a squash but ready for review. I have moved it back to a Duration, and ensured that you get the hibernation not supported condition immediately if you try to use this feature as well.

contrib/pkg/createcluster/create.go
```diff
@@ -244,6 +245,7 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=ovirt --ovirt-api-vip 192.168.1.2
 	flags.StringVar(&opt.BaseDomain, "base-domain", "new-installer.openshift.com", "Base domain for the cluster")
 	flags.StringVar(&opt.PullSecret, "pull-secret", "", "Pull secret for cluster. Takes precedence over pull-secret-file.")
 	flags.StringVar(&opt.DeleteAfter, "delete-after", "", "Delete this cluster after the given duration. (i.e. 8h)")
+	flags.StringVar(&opt.HibernateAfter, "hibernate-after", "", "Automatically hibernate the cluster whenever is has been running for the given duration")
```
staebler, line 248:
```suggestion
	flags.StringVar(&opt.HibernateAfter, "hibernate-after", "", "Automatically hibernate the cluster whenever it has been running for the given duration")
```

contrib/pkg/createcluster/create.go
```diff
@@ -511,6 +513,7 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
 		BaseDomain:            o.BaseDomain,
 		ManageDNS:             o.ManageDNS,
 		DeleteAfter:           o.DeleteAfter,
+		HibernateAfter:        o.HibernateAfter,
```
staebler, line 516: It seems like a better interface for the builder if it takes a `*time.Duration` instead of a `string` for `HibernateAfter`. The create-cluster `Complete` function could translate the string to a duration so that the builder can take a duration.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -123,6 +123,13 @@ type ClusterDeploymentSpec struct {
 	// PowerState defaults to the Running state.
 	// +optional
 	PowerState ClusterPowerState `json:"powerState,omitempty"`
+
+	// HibernateAfter will transition a cluster to hibernating power state after it has been running for the
+	// given duration. Running is assumed to be the time since the cluster was installed if no hibernating
+	// condition exists, otherwise the time since the cluster last transitioned that condition to false. If the cluster
+	// predates hibernation functionality (>= 4.4.8) the problem will be reported in the hibernating condition and the
+	// cluster will continue running.
+	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
```
staebler, line 132:
```suggestion
	// +optional
	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
```

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -123,6 +123,13 @@ type ClusterDeploymentSpec struct {
 	// PowerState defaults to the Running state.
 	// +optional
 	PowerState ClusterPowerState `json:"powerState,omitempty"`
+
+	// HibernateAfter will transition a cluster to hibernating power state after it has been running for the
+	// given duration. Running is assumed to be the time since the cluster was installed if no hibernating
```
staebler, line 128:
```suggestion
	// given duration. Running time is assumed to be the time since the cluster was installed if no hibernating
```

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -123,6 +123,13 @@ type ClusterDeploymentSpec struct {
...
+	// condition exists, otherwise the time since the cluster last transitioned that condition to false. If the cluster
```
staebler, line 129: This may be a little too much detail for a user-facing comment. I suggest something like the following.

> The time that a cluster has been running is the time since the cluster was installed or the time since the cluster last came out of hibernation.

pkg/test/clusterdeployment/clusterdeployment.go
```diff
@@ -129,8 +130,24 @@ func InstalledTimestamp(instTime time.Time) Option {
 	}
 }
 
+func ClusterVersion(version string) Option {
```
staebler, line 133:
```suggestion
func WithClusterVersion(version string) Option {
```

pkg/gcpclient/mock/client_generated.go
```diff
@@ -7,8 +7,8 @@ package mock
 import (
 	gomock "github.com/golang/mock/gomock"
 	gcpclient "github.com/openshift/hive/pkg/gcpclient"
-	compute "google.golang.org/api/compute/v1"
-	dns "google.golang.org/api/dns/v1"
+	v1 "google.golang.org/api/compute/v1"
```
staebler, line 10: I don't get any of the changes in this file when I run `make generate`. Is there some difference in the gomock version being used?

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -151,6 +152,63 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+			"hibernateAfter": hibernateAfterDur,
+		})
+		var isRunning bool
+
+		cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
+		if cond == nil {
+			hibLog.Debug("cluster has no hibernating condition (never hibernated), using installed time")
+			isRunning = true
+		} else if cond.Status == corev1.ConditionFalse {
+			runningSince = cond.LastTransitionTime.Time
```
staebler, line 177:
```suggestion
			runningSince = cond.LastTransitionTime.Time
			hibLog = hibLog.WithField("runningSince", runningSince)
```

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/golang/mock/gomock"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
staebler, line 13: Move import to correct section.

staebler (commented):
Looks good overall. Only nits left other than the question about the changes to the mock GCP client.

contrib/pkg/createcluster/create.go
dgoodwin: Agreed, changed to dur.

pkg/gcpclient/mock/client_generated.go
dgoodwin: ❯ /home/dgoodwin/go/bin/mockgen -version
v1.4.4

I tried removing my binary and re-running make generate but it brought the same version back. What version is on your end?

staebler: That is the version that I have to and matches the version from the go.mod file.

dgoodwin: I can't seem to make it revert, tried removing and regenerating as well.

dgoodwin:
Everything else updated but I don't know what's up with the gomock changes.

pkg/gcpclient/mock/client_generated.go
staebler: Can we keep this change out of this PR since it is not related to the rest of the PR? If we determine that the change is needed, then we can add it in its own PR?

dgoodwin: Done.

staebler (approved):
/lgtm

staebler:
@dgoodwin #1134 made it in first, so you lost the rebase lottery.

dgoodwin:
Damn you.

Fixed now, very minor.

staebler (approved):
/lgtm

staebler:
/lgtm cancel
```
Generating deepcopy funcs
F0918 13:31:49.239802   20985 main.go:82] Error: Failed making a parser: unable to add directory "github.com/openshift/hive/pkg/apis/hive/v1": unable to import "github.com/openshift/hive/pkg/apis/hive/v1": go/build: importGo github.com/openshift/hive/pkg/apis/hive/v1: exit status 1
go: github.com/openshift/installer@v0.9.0-master.0.20200817183917-ab4e753ff30b requires
	github.com/openshift/library-go@v0.0.0-20200324092245-db2a8546af81 requires
	bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found
```

staebler:
/lgtm cancel

Remove the lgtm so that we don't retest repeatedly when we know that `test verify` is going to fail.

dgoodwin:
/retest

Needs lgtm again.

PR #1123, opened 2020-09-04
Add hiveutil power on/off command for hibernation.

dgoodwin:


contrib/pkg/power/on.go
```diff
@@ -0,0 +1,62 @@
...
+
+func setClusterPowerState(cdName string, state hivev1.ClusterPowerState) {
+	dynClient, err := utils.GetClient()
+	if err != nil {
+		log.WithError(err).Fatal("error creating kube clients")
+	}
+
+	cdLog := log.WithField("name", cdName)
+	cd := &hivev1.ClusterDeployment{}
+	err = dynClient.Get(context.Background(), types.NamespacedName{Namespace: "hive", Name: cdName}, cd)
```
staebler, line 46: Why only support ClusterDeployments in the hive namespace?

staebler (commented):
This is a lot of code to maintain to avoid a simple `oc patch`.
```
oc patch cd myCluster --type='merge' -p $'spec:\n powerState: Hibernating'
```

PR #1126, opened 2020-09-09
document installing a cluster with hive onto oVirt

joelddiaz:


pkg/install/generate.go
```diff
@@ -237,6 +237,13 @@ func InstallerPodSpec(
 			},
 		)
 		env = append(env, oVirtCredsEnvVars(cd.Spec.Platform.Ovirt.CredentialsSecretRef.Name)...)
+		// TODO: design/implement a system to sync over ClusterImageSet base OS images into OpenStack
```
staebler, line 240: This can already be done via `.spec.provisioning.installerEnv`. Do we need to special-case this?

docs/using-hive.md
```diff
@@ -128,6 +140,26 @@ metadata:
 type: Opaque
 ```
 
+#### oVirt
+Create a `secret` containing your oVirt credentials information:
+
+```yaml
+apiVersion: v1
+data:
```
staebler, line 148: Should this be `stringData`?

docs/using-hive.md
```diff
@@ -224,6 +256,26 @@ and replace the contents of `platform` with:
     region: us-east1
 ```
 
+For oVirt, replace the contents of `compute.platform` and `controlPlane.platform` with empty brackets (`{}`).
+```yaml
+controlPlane:
+  platform: {}
```
staebler, line 262: Can't this just be omitted instead?

docs/using-hive.md
```diff
@@ -276,6 +328,18 @@ gcp:
   region: us-east1
 ```
 
+For oVirt, replace the contents of `spec.platform` with:
+```yaml
+ovirt:
+  certificatesSecretRef:
+    name: mycluster-ovirt-certs
```
staebler, line 335: Missing documentation on how to build this secret.

staebler:
/lgtm cancel

I would like to get a couple questions answered.

pkg/install/generate.go
joelddiaz: I haven't completely thought this through, but what I had in mind was something like having a clusterimageset defined will cause a pre-emptive upload of the RHCOS image to the OpenStack/oVirt/whatever environment, and ClusterDeployments would just use the pre-uploaded images automatically. It gets a little hazy as clusterdeployment A might be targeting oVirt environment A, and clusterdeployment B is targeting oVirt environment B. Hive would needs creds for all possible environments. Hive would need to know that a new clusterdeployment targets an environment that doesn't have a pre-uploaded image. I'm sure there are more things to think about...

But, it is great to know about this .spec.provisioning.installerEnv to simplify any future testing.

docs/using-hive.md
joelddiaz: I was just following along with the install-config.yaml that openshift-install creates.

pkg/install/generate.go
staebler: I guess I don't understand why we are *adding* commented out code.

joelddiaz: I'll remove it now. I had put it in there as an easy-to-find marker for when I wanted to speed up my testing. But with this installerEnv field, it doesn't make sense. I'll remove it and open a Jira card for my longer-term dream of having this "just work".

docs/using-hive.md
```diff
@@ -276,6 +328,30 @@ gcp:
...
+```
+
+And create a Secret that holds the CA certificate data for the oVirt environment:
+```yaml
+- apiVersion: v1
+  data:
+    .cacert: base64-encoded-cacert-data-here
+  kind: Secret
+  metadata:
+    creationTimestamp: null
```
staebler, line 350: Omit this.

docs/using-hive.md
```diff
@@ -276,6 +328,30 @@ gcp:
...
+  ovirt_cluster_id: 00000000-ovirt-uuid
+  ovirt_network_name: ovirt-network-name
+  storage_domain_id: 00000000-storage-domain-uuid
+```
+
+And create a Secret that holds the CA certificate data for the oVirt environment:
+```yaml
+- apiVersion: v1
+  data:
+    .cacert: base64-encoded-cacert-data-here
```
staebler, line 347: Would it be easier to tell the user to use
```
oc create secret generic my-cluster-ovirt-certs -n mynamespace --from-file=.cacert=$CERT_FILENAME
```
rather than telling them to base64-encoded their cacert?

docs/using-hive.md
```diff
@@ -276,6 +328,30 @@ gcp:
...
+  kind: Secret
+  metadata:
+    creationTimestamp: null
+    name: mycluster-ovirt-certs
```
staebler, line 351: Missing namespace.

docs/using-hive.md
staebler: The install-config generated by the installer contains default values that are not required.

staebler (approved):
/lgtm

PR #1129, opened 2020-09-10
Annotate remote resources with hive source(s)

2uasimojo:
For discoverability by service personnel, add an annotation to each resource pushed into the cluster at the behest of a [Selector]SyncSet. The annotation contains enough information to identify the hive [Selector]SyncSet(s) that affected the resource.

Example (not realistic):

`hive.openshift.io/sources: "SelectorSyncSet/ccs-dedicated-admins-environment SyncSet/groups"`

The idea is that you can backtrack from an in-cluster object to its hive
sources:

On the target cluster, get the annotation for the object in question:

```
$ oc get $somekind $somename -o jsonpath="{.metadata.annotations['hive\.openshift\.io/sources']}"
SelectorSyncSet/ccs-dedicated-admins-environment SyncSet/groups
```

And then on the hive, paste that into `oc get` in the ClusterDeployment's namespace (cluster-scoped things show up fine,
ignoring the namespace):
```
$ oc get -n $cd_namespace SelectorSyncSet/ccs-dedicated-admins-environment SyncSet/groups
NAME                                                                 AGE
selectorsyncset.hive.openshift.io/ccs-dedicated-admins-environment   50d

NAME                               AGE
syncset.hive.openshift.io/groups   3h5m

```

Jira: [OSD-4848](https://issues.redhat.com/browse/OSD-4848)

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -582,7 +582,7 @@ func (r *ReconcileClusterSync) applySyncSet(
 
 	// Apply Resources
 	for i, resource := range resources {
-		returnErr, requeue = r.applyResource(i, resource, referencesToResources[i], applyFn, applyFnMetricsLabel, logger)
+		returnErr, requeue = r.applyResource(syncSet, i, resource, referencesToResources[i], applyFn, applyFnMetricsLabel, logger)
 		if returnErr != nil {
 			resourcesApplied = referencesToResources[:i]
 			return
@@ -649,6 +649,7 @@ func referencesToSecrets(syncSet CommonSyncSet) []hiveintv1alpha1.SyncResourceRe
 }
 
 func (r *ReconcileClusterSync) applyResource(
+	syncSet CommonSyncSet,
 	resourceIndex int,
 	resource *unstructured.Unstructured,
 	reference hiveintv1alpha1.SyncResourceReference,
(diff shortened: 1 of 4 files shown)


2uasimojo:
If this looks like something we want to do, I'll fix up the tests tomorrow. (@joelddiaz since you looked at the [previous attempt](https://github.com/openshift/hive/pull/1090); @staebler, @dgoodwin, and @gregsheremeta since you were involved in the discussions there.)

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -740,20 +742,44 @@ func (r *ReconcileClusterSync) applyPatch(
 	return nil, false
 }
 
+// ensureHiveSourceAnnotation makes sure obj has a HiveSourcesAnnotation containing an
+// appropriate value for syncSet.
+// Currently the annotation is a space-separated list of strings of the form Kind/Name,
+// where Kind is either SyncSet or SelectorSyncSet, and Name is the name of the
+// [Selector]SyncSet.
+func ensureHiveSourceAnnotation(obj hivev1.MetaRuntimeObject, syncSet CommonSyncSet) {
```
staebler, line 750: This is not going to work. The `obj` passed into this function will not contain the annotations that exist in the target cluster. If there are multiple syncsets applying the same resource to the target cluster, they will each use a different value for the annotation: The annotation values will not get combined.

2uasimojo: Well dang. I'll have to dig for a way around that.

In general, are you supportive of this as a concept?

staebler: I do not find the concept objectionable, but I am not convinced of the utility or whether we want to expose these details in customer-accessible objects. My biggest concern is that I think it is going to be very difficult to get it right.

2uasimojo: That's fair. It's not *hugely* more convenient than searching through the ClusterSync. If it was easy -- if it would work as written -- then it might be worth it; but if it's going to be hard...

2uasimojo:
Abandoning as not worth the effort. See https://github.com/openshift/hive/pull/1129#discussion_r487221831

PR #1131, opened 2020-09-11
install: fix azure failed provision cleanup

staebler:
As with df024bd7957fc800e363d7f24561ab862b5839cc, the Azure destroy in installmanager needs to include the resource group in the destroyer.

Additionally, the call to the destroyer Run that is repeated for every platform has been brought out of the switch.

https://issues.redhat.com/browse/CO-1142

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -16,18 +16,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openshift/hive/pkg/gcpclient"
-
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/yaml"
-
-	contributils "github.com/openshift/hive/contrib/pkg/utils"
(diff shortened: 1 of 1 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -530,41 +527,30 @@ func (m *InstallManager) cleanupFailedInstall(cd *hivev1.ClusterDeployment, prov
...
-		uninstaller := &azure.ClusterUninstaller{}
-		uninstaller.Logger = logger
-		session, err := azuresession.GetSession(installertypesazure.PublicCloud)
-		if err != nil {
-			return err
+		metadata := &installertypes.ClusterMetadata{
+			InfraID: infraID,
+			ClusterPlatformMetadata: installertypes.ClusterPlatformMetadata{
+				Azure: &installertypesazure.Metadata{
+					CloudName: installertypesazure.PublicCloud,
```
joelddiaz, line 547: will it always be type PublicCloud?

staebler: Yes, until we do https://issues.redhat.com/browse/CO-1005.

PR #1132, opened 2020-09-11
deprovision: kill deprovision when creds change

staebler:
The "--creds-dir" flag has been added to the `hiveutil deprovision` command. When this flag is used, the command will create a file observer watching for any changes to any files in that directory. If a change is detected, the command will be terminated.

The uninstall pods for platforms other than AWS have been updated to include the "--creds-dir" flag in the call to the deprovision
command.

https://issues.redhat.com/browse/CO-1135

### DIFF
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -583,6 +583,8 @@ func completeAzureDeprovisionJob(req *hivev1.ClusterDeprovision, job *batchv1.Jo
 				"azure",
 				"--loglevel",
 				"debug",
+				"--creds-dir",
+				azureAuthDir,
 				req.Spec.InfraID,
 			},
 			VolumeMounts: volumeMounts,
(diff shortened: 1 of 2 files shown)


pkg/install/generate.go
```diff
@@ -700,8 +722,18 @@ func completeVSphereDeprovisionJob(req *hivev1.ClusterDeprovision, job *batchv1.
...
-			Args:            []string{fmt.Sprintf("cp -vr %s/. /etc/pki/ca-trust/source/anchors/ && update-ca-trust && /usr/bin/hiveutil deprovision vsphere --vsphere-vcenter %s --loglevel debug %s", vsphereCloudsDir, req.Spec.Platform.VSphere.VCenter, req.Spec.InfraID)},
-			VolumeMounts:    volumeMounts,
+			Args: []string{
+				fmt.Sprintf(
+					"cp -vr %s/. /etc/pki/ca-trust/source/anchors/ && "+
+						"update-ca-trust && "+
+						"/usr/bin/hiveutil deprovision vsphere --vsphere-vcenter %s --loglevel debug --creds-dir=%s %s",
+					vsphereCloudsDir,
+					req.Spec.Platform.VSphere.VCenter,
+					"/vsphere-creds",
```
joelddiaz, line 732: can we constify this like we've done for the other platforms?

staebler: We could, but it is local to this function. I can create a function-scoped const.

joelddiaz (commented):
/lgtm

PR #1135, opened 2020-09-14
remove clusterVersionStatus field from clusterdeployment

staebler:
Remove the `.status.clusterVersionStatus` field from the ClusterDeployent type. Replaces usages of the field in Hive with usages of the "hive.openshift.io/version-major-minor-patch" label.

https://issues.redhat.com/browse/CO-1123

/assign @joelddiaz 
/cc @twiest


test/e2e/postinstall/machinesets/infra_test.go
```diff
@@ -424,7 +425,8 @@ func machineNamePrefix(cd *hivev1.ClusterDeployment, poolName string) (string, e
 	// GCP clusters running an OpenShift version earlier than 4.4.8 require leases for machine pool names because the
 	// pool name is limited to a single character.
 	if p := cd.Spec.Platform; p.GCP != nil {
-		if cd.Status.ClusterVersionStatus != nil && semver.Compare("v4.4.8", "v"+cd.Status.ClusterVersionStatus.Desired.Version) > 0 {
+		version, versionPresent := cd.Labels[constants.VersionMajorMinorPatchLabel]
```
wking, line 428: Docs for this label aren't particularly clear.  Is this supposed to be the desired version, regardless of whether it has been reconciled or not.  Or the most-recently completed version?  Or...?

joelddiaz (commented):
Seems like a clean conversion of existing functionality. Although, @wking 's comment does make me wonder why we are syncing down the desired version instead of the actual applied/running version.

test/e2e/postinstall/machinesets/infra_test.go
staebler: That is a great question. The version reported in the ClusterDeployment label is the version reported by the cluster in the `.status.desiredUpdate` field of the `ClusterVersion` resource. When Hive is making a decision about how to create, for example, a MachineSet in the cluster that is undergoing an upgrade, should Hive assume that the cluster is going to make it successfully to the new version and use configuration for the desired version? Or should Hive assume that the cluster is still going to need configuration as though it were entirely on the old version? It would seem to me that the cluster is in an indeterminate state at that point. Some operators may have already converted to the new version while some remain on the old version.

joelddiaz: Maybe we should put a card in our todo list to investigate this. At the very least, the code changes in this PR aren't introducing any new behavior.

wking: For the future work, machine-config is one of the last operators to update.  We don't bump bootimages in-cluster today, but if you were looking up a bootimage to use for a new MachineSet, it's probably safer to look through `status.history` for the most recent entry which was actually completed.  Or crib from the control-plane Machines.  Or something...

staebler: We do crib from the control-plane Machines in order to determine the bootimage to use.

wking: So what's this version for?  If it's display-only, the desired target is already reasonable.

staebler: The version in the label is used for a few purposes.
1) The label is used by users for SelectorSyncSets to apply resources to the cluster based on the cluster version.
2) The label is used by the remotemachineset controller when creating MachineSets in the cluster version to determine (a) the MachineSet name to use for GCP clusters, (b) the name of the worker user data secret, and (c) whether the cluster supports spot instances on AWS.
3) The label is used by the hibernate controller to determine whether the cluster supports hibernation.

joelddiaz (commented):
i'm spooked that we have code that takes action based on desired version..
/lgtm

PR #1136, opened 2020-09-14
docs: add scaling documentation

gregsheremeta:
/assign @dgoodwin @akhil-rane 
/cc @twiest


docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+|---|---|---|---|
+|c5.2xlarge|c5.2xlarge|20|700|
+|c5.2xlarge|c5.2xlarge|40|700|
+|c5.4xlarge|c5.4xlarge|20|1200|
+|c5.4xlarge|c5.4xlarge|40|1400|
+|c5.2xlarge|c5.4xlarge|20|1470|
+|c5.2xlarge|c5.9xlarge|20|1750|
+|c5.2xlarge|c5.9xlarge|40|2500|
+|c5.2xlarge|c5.9xlarge|50|2500|
+|c5.4xlarge|c5.24xlarge|100|2400|
```
dgoodwin, line 47: This is a nice summary.

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Threads
+
+Hive supports configuring the number of goroutines per controller by editing values in HiveConfig. See [Using Hive](using-hive.md) for documentation on this. In general, set the number of goroutines to roughly the number of vCPUs on your workers. For c5.2xlarge, we use 10 goroutines. If you run Hive on larger hardware, you need to bump up the goroutines. If you have only 10 goroutines but 100 vCPUs, 90 of the vCPUs will be mostly unused. If Hive manages clusters that are on slow networks or have connectivity issues, you may want to use more goroutines to work around Hive's use of blocking i/o. Don't use too many goroutines, or that will actually cause a slowdown (i.e. don't just set it to 1000 or something).
+
+## SyncSets
+
+Pushing configuation to managed clusters via SyncSets is the most CPU-intensive and network-intensive thing that Hive does. We scale test Hive by mostly looking at how syncsets perform, since post-install this is what Hive spends the majority of its time doing.
+
+The metric we judge syncset performance by is "applies per second". The following table lists some typical applies/sec rates for various hardware combinations we've tested.
```
dgoodwin, line 35: "various AWS instance types"

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+|masters|workers|goroutines|applies/sec|
```
dgoodwin, line 37: Probably should swap masters for control-plane given everything going on.

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
+# Scaling Hive
+
+This document describes the main places users are likely to see scalability issues with Hive.
+
+## Hive uses CRDs!
+
+Most importantly, be aware that Hive uses CRDs to store its state. The amount of data that Hive can store is limited by the cluster's ability to store CRs. In OpenShift, this boils down to "how much data can we stuff into etcd before the cluster itself becomes unstable?" Best to not chance it. We recommend keeping the total number of Hive CRs in a Hive cluster in the thousands. Anything approaching 10,000 is the danger zone. In practice this means that a single Hive cluster can manage about 1000 clusters.
```
dgoodwin, line 7: Suggest "should manage no more than 1000 clusters."

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Horizontal vs. Vertical Scale
+
+With the exception of install pods (used only when clusters are installing), Hive 1.x is not horizontally scalable at the worker level. Most of the work Hive does happens in the hive-controllers pod, which is one single pod on one single worker. This means that when no installs are running, if you have a cluster with 10 workers, 9 of the workers are very bored. Hive clusters are prime candidates for using worker autoscaling. Keep the worker count as low as you can, but allow bursts of concurrent installs to call for temporary workers to spin up.
+
+In AWS, Hive performs best on c instance types. M instances are fine, but c instances are better.
```
dgoodwin, line 13: Lower case "c" vs upper case "M", is this intentional? I'd sync them up. Also might be good to clarify C is for CPU optimized and M is general purpose.

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+In general, the faster your etcd storage, the faster Hive will perform. c5.4xlarge has twice the EBS bandwidth as c5.2xlarge, and so Hive performs a bit better on c5.4xlarge (although in our opinion, not so much faster that c5.4xlarge is worth the premium over c5.2xlarge). In AWS, we recommend using io1 with 1200 iops for your masters.
+
+If you need to scale out the number of clusters you need to manage, use multiple Hive clusters -- approximately 1 Hive per 1000 clusters managed. Or fewer if you prefer to keep the blast radius smaller.
```
dgoodwin, line 17: I'd drop the blast radius comment, it's implied that of course you can manage less per hive, but it makes it sound like we explode a lot. :)

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Install Pods
+
+Hive 1.x request 800 Mib of memory for each install pod. If you use m5.xlarge workers, you can support about (15 Gib / 800 Mib) install pods per worker -- so about 16. If you need to support more concurrent installs, you can use more workers, and/or workers with more memory. Install pods use barely any CPU.
```
dgoodwin, line 21: 1.x eh, I see what you're up too. :rocket: :)

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Blocking I/O
+
+hive-controllers (where the syncset controllers run) uses blocking i/o. We may move to scale-out or non-blocking i/o in the future, but for now you should be aware that communication (mostly syncsets) to managed clusters can block other things.
+
+## Threads
+
+Hive supports configuring the number of goroutines per controller by editing values in HiveConfig. See [Using Hive](using-hive.md) for documentation on this. In general, set the number of goroutines to roughly the number of vCPUs on your workers. For c5.2xlarge, we use 10 goroutines. If you run Hive on larger hardware, you need to bump up the goroutines. If you have only 10 goroutines but 100 vCPUs, 90 of the vCPUs will be mostly unused. If Hive manages clusters that are on slow networks or have connectivity issues, you may want to use more goroutines to work around Hive's use of blocking i/o. Don't use too many goroutines, or that will actually cause a slowdown (i.e. don't just set it to 1000 or something).
+
```
dgoodwin, line 30: Are we going to skip mentions of client throttling presumably because that really only surfaced when we tried to have everything talk to one cluster?

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+## Blocking I/O
+
+hive-controllers (where the syncset controllers run) uses blocking i/o. We may move to scale-out or non-blocking i/o in the future, but for now you should be aware that communication (mostly syncsets) to managed clusters can block other things.
+
+## Threads
+
+Hive supports configuring the number of goroutines per controller by editing values in HiveConfig. See [Using Hive](using-hive.md) for documentation on this. In general, set the number of goroutines to roughly the number of vCPUs on your workers. For c5.2xlarge, we use 10 goroutines. If you run Hive on larger hardware, you need to bump up the goroutines. If you have only 10 goroutines but 100 vCPUs, 90 of the vCPUs will be mostly unused. If Hive manages clusters that are on slow networks or have connectivity issues, you may want to use more goroutines to work around Hive's use of blocking i/o. Don't use too many goroutines, or that will actually cause a slowdown (i.e. don't just set it to 1000 or something).
```
dgoodwin, line 29: Are you recommending setting numvcpus for *each* controller, meaning the total goroutines is numvcpus*numcontrollers? This paragraph needs to be clarified in regards to this in a few places.

Do we still default to 5? If so that should be documented. Are we actively running 10 on c5.2xlarge anywhere? Also we should note that's AWS c5.2xlarge.

I'm really unsure about linking to vcpus so heavily here especially given we just said we're IO blocking. This feels somewhat inaccurate, but maybe you're implying it's just a good general guideline?

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+If you need to scale out the number of clusters you need to manage, use multiple Hive clusters -- approximately 1 Hive per 1000 clusters managed. Or fewer if you prefer to keep the blast radius smaller.
+
+## Install Pods
+
+Hive 1.x request 800 Mib of memory for each install pod. If you use m5.xlarge workers, you can support about (15 Gib / 800 Mib) install pods per worker -- so about 16. If you need to support more concurrent installs, you can use more workers, and/or workers with more memory. Install pods use barely any CPU.
+
+## Blocking I/O
+
+hive-controllers (where the syncset controllers run) uses blocking i/o. We may move to scale-out or non-blocking i/o in the future, but for now you should be aware that communication (mostly syncsets) to managed clusters can block other things.
```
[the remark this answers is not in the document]
gregsheremeta: Not referring to client throttling.

@staebler confirmed we use blocking i/o during a stand up.

If your ss controller has 5 threads, and you send 5 GETs to 5 clusters that are Unavailable, you will do no work in the cc controller for 30 seconds each thread waits to timeout.

Contrast with async i/o like in nodejs where 5 threads send 5 requests, they immediately go do something else, and they are called back when the response comes back. This is much more scalable.

docs/scaling-hive.md
gregsheremeta: M is capitalized because it's the beginning of a sentence. I'll uppercase the C's too.

docs/scaling-hive.md
gregsheremeta: It's a super-common microservices term
e.g. https://www.ibm.com/garage/method/practices/manage/practice_limited_blast_radius/

docs/scaling-hive.md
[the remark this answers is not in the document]
staebler: > you should be aware that communication (mostly syncsets) to managed clusters can block other things

The communication that the *clustersync* controller does with a remote cluster may block the *clustersync* controller from reconciling other clusters. It will not block other controllers from performing work.

docs/scaling-hive.md
staebler: Or re-organize the sentence to avoid putting M first.
```suggestion
In AWS, Hive performs best on c instance types. Hive performs fine on m instances, but c instances are better.
```

docs/scaling-hive.md
gregsheremeta: In a busy hive cluster, over the life of the cluster, the ss controller is the thing that does most of the work. The other controllers are bored compared to ss. So I'm simplifying things and only thinking about ss.

The goal of anyone looking to push hive should be to the keep CPU utilization for the worker as close to 100% as possible.

I found that we get the highest CPU utilization (and thus best overall performance) when we roughly match the number of ss threads to the number of vCPUs on the worker. If we use less, the CPU is bored. If we have 36 CPUs but only 20 threads, CPU utilization is usually 50%. If we match (40 threads for 36 vCPU), CPU utilization is high 90's. If we use too many threads (100 threads for 36 vCPU), we still get high CPU utilization, but the process actually slows down. I assume this is golang overhead of too many threads / thread management.

Generally, blocking i/o is not a problem. It should be rare for hive to manage a lot of clusters that are on slow networks, unreachable, etc. So I simplify things and don't really take that into consideration in this section. If you're in a special scenario where you *do* manage a lot of clusters that are frequently unreachable, you would frequently have several threads blocked on i/o hangs -- doing nothing. CPU utilization would drop. So you could compensate by adding in a couple more threads.

Does this make sense?

docs/scaling-hive.md
gregsheremeta: I'm convinced we shouldn't mess with throttling ourselves, and we shouldn't encourage it. This is the fast lane to cluster death.

dgoodwin: Yeah that seems reasonable, I don't recall it being valuable after we got off one cluster.

docs/scaling-hive.md
dgoodwin: Makes sense but it needs to be clarified here that we're talking about clustersync controller specifically. I'm wouldn't be crazy about this advice for all controllers.

akhil-rane (commented):
This is so well written @gregsheremeta, thanks :)

docs/scaling-hive.md
twiest: Hive 1.x request 800 Mib ... 

Should be:

Hive 1.x requests 800 MiB ...

docs/scaling-hive.md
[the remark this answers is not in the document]
twiest: We may also want to mention that we detect unavailable clusters and we don't attempt to apply syncsets to them. So this is a temporary issue until those clusters are determined to be unavailable.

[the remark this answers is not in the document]
gregsheremeta: I attempted to clarify this. I tried to take into account everyone's suggestions.

docs/scaling-hive.md
gregsheremeta: done

docs/scaling-hive.md
gregsheremeta: done

docs/scaling-hive.md
gregsheremeta: done

docs/scaling-hive.md
gregsheremeta: I like using the letters, but I did add descriptions in parentheses. Hope this works.

docs/scaling-hive.md
gregsheremeta: I reworked it to emphasize architectures that are distributed into multiple regions or are cell-based. Hope this works.

docs/scaling-hive.md
gregsheremeta: done

docs/scaling-hive.md
gregsheremeta: I took large portions of my above comment and added it. Hopefully it is clearer now.

docs/scaling-hive.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Horizontal vs. Vertical Scale
+
+With the exception of install pods (used only when clusters are installing), Hive 1.x is not horizontally scalable at the worker level. Most of the work Hive does happens in the hive-controllers pod, which is one single pod on one single worker. This means that when no installs are running, if you have a cluster with 10 workers, 9 of the workers are very bored. Hive clusters are prime candidates for using worker autoscaling. Keep the worker count as low as you can, but allow bursts of concurrent installs to call for temporary workers to spin up.
+
+In AWS, Hive performs best on c instance types. M instances are fine, but c instances are better.
+
+In general, the faster your etcd storage, the faster Hive will perform. c5.4xlarge has twice the EBS bandwidth as c5.2xlarge, and so Hive performs a bit better on c5.4xlarge (although in our opinion, not so much faster that c5.4xlarge is worth the premium over c5.2xlarge). In AWS, we recommend using io1 with 1200 iops for your masters.
```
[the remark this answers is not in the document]
gregsheremeta: done

gregsheremeta:
@twiest @staebler @dgoodwin @akhil-rane please re-review.

docs/scaling-hive.md
gregsheremeta: thanks

PR #1137, opened 2020-09-15
deprovision: skip symlinks in creds file watch

staebler:
When finding the files to watch for creds changes, skip symlinks. The `..data` symlink links to a directory, causing which causes
the program to error out attempting to read the directory. The other symlinks point to files that are already included in the watch.


joelddiaz (commented):
/lgtm

staebler:
/hold

waiting on an e2e run on gcp

staebler:
/hold cancel

e2e-gcp test passed

PR #1138, opened 2020-09-16
machinepool condition for unsupported spot instances

staebler:
Add the UnsupportedConfiguration condition to the MachinePool type. This condition is used to indicate that the cluster does not support the configuration specified in the MachinePool. If the MachinePool specifies spot instances but the cluster is running an OpenShift version earlier than 4.5, then the UnsupportedConfiguration condition will be set on the MachinePool and not MachineSets will be generated.

https://issues.redhat.com/browse/CO-1042

### DIFF
--- a/pkg/controller/remotemachineset/awsactuator.go
+++ b/pkg/controller/remotemachineset/awsactuator.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/blang/semver/v4"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
@@ -43,6 +44,8 @@ var (
 
 	// reg is a regex used to fetch condition message from error when subnets specified in the MachinePool are invalid
 	reg = regexp.MustCompile(`^InvalidSubnetID\.NotFound:\s+([^\t]+)\t`)
+
+	versionsSupportingSpotInstances = semver.MustParseRange(">=4.5.0")
 )
 
 func addAWSProviderToScheme(scheme *runtime.Scheme) error {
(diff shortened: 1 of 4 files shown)


pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -43,6 +44,8 @@ var (
 
 	// reg is a regex used to fetch condition message from error when subnets specified in the MachinePool are invalid
 	reg = regexp.MustCompile(`^InvalidSubnetID\.NotFound:\s+([^\t]+)\t`)
+
+	versionsSupportingSpotInstances = semver.MustParseRange(">=4.5.0")
```
suhanime, line 48: `MustParseRange` panics when range cannot be parsed... Would it be better to use `ParseRange` instead? We can also just log the error and move on without blocking the flow

staebler: This is done in the init of the package. If the range cannot be parsed, that is a coding error. We want the program to crash in this case.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -349,3 +380,18 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
 }
+
+func isUsingUnsupportedSpotMarketOptions(pool *hivev1.MachinePool, clusterVersion string) bool {
+	if pool.Spec.Platform.AWS.SpotMarketOptions == nil {
+		return false
+	}
+	finalizedVersion, err := semver.FinalizeVersion(clusterVersion)
+	if err != nil {
+		return true
+	}
```
akhil-rane, line 391: Should we log that there was error while parsing here?

akhil-rane, line 391: Should we log that there was an error while parsing here?

pkg/controller/remotemachineset/awsactuator_test.go
```diff
@@ -15,7 +15,7 @@ import (
 
 	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
```
akhil-rane, line 16: I think openshift import can be separated from k8s here

pkg/controller/remotemachineset/awsactuator_test.go
```diff
@@ -187,6 +188,59 @@ func TestAWSActuator(t *testing.T) {
...
+			},
+			expectedCondition: &hivev1.MachinePoolCondition{
+				Type:   hivev1.UnsupportedConfigurationMachinePoolCondition,
+				Status: corev1.ConditionTrue,
+				Reason: "UnsupportedSpotMarketOptions",
+			},
+		},
+		{
+			name:              "unsupported configuration condition cleared",
+			clusterDeployment: withClusterVersion(testClusterDeployment(), "4.4.0"),
```
akhil-rane, line 220: I did not understand how unsupported configuration condition will be cleared for a cluster with version 4.4.0. Aren't we supporting 4.5.0 and above?

staebler: It is cleared because the MachinePool is not using spot market options.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -349,3 +380,18 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
 }
+
+func isUsingUnsupportedSpotMarketOptions(pool *hivev1.MachinePool, clusterVersion string, logger log.FieldLogger) bool {
+	if pool.Spec.Platform.AWS.SpotMarketOptions == nil {
+		return false
+	}
+	finalizedVersion, err := semver.FinalizeVersion(clusterVersion)
+	if err != nil {
+		logger.WithError(err).WithField("clusterVersion", clusterVersion).Warn("could not finalize the cluster version")
+	}
```
akhil-rane, line 391: why not return after logging the error here?

staebler: It should return. I messed it up during a couple different approaches that I took here. Good catch.

staebler: I added a unit test for this. In doing so, I realized that `FinalizeVersion` does a `Parse`, meaning that the subsequent `ParseTolerant` will never return an error. I changed the code so that it does not use `FinalizeVersion` and instead directly removes the parts other than major, minor, and patch from the parsed version.

PR #1139, opened 2020-09-16
sort patches for identityproviders

staebler:
Sort the patches for IdentityProviders by name so that the patch in the created SyncSet is consistent. This keeps the SyncSet from potentially thrashing due to a non-determinstic ordering.

https://issues.redhat.com/browse/CO-1144


abutcher (commented):
Looks good to me, will let @twiest ack

PR #1140, opened 2020-09-17
requeue cd after creating clustersync

staebler:
Instead of watching for the ClusterSync to be added, have the clustersync controller requeue the ClusterDeployment after creating the ClusterSync resource. Then, the clustersync controller will error on getting the ClusterSync on subsequent reconciles until the ClusterSync resource makes it into the local cache.


gregsheremeta:
assign and cc?

PR #1141, opened 2020-09-17
Bug 1880443: fix Openstack machinesets

joelddiaz:
- [x] vendor in new installer with ability to provide our own OpenStack credentials fetching functions
- [x] provide custom credentials fetching functions that use the contents of the ClusterDeployment's secret as the credentials source
- [x] provide an empty (non-nil) list of Zones when creating MachinePools to MachineSets for OpenStack

### DIFF
--- a/pkg/controller/remotemachineset/openstackactuator.go
+++ b/pkg/controller/remotemachineset/openstackactuator.go
@@ -71,6 +71,13 @@ func (a *OpenStackActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, po
 	computePool := baseMachinePool(pool)
 	computePool.Platform.OpenStack = &installertypesosp.MachinePool{
 		FlavorName: pool.Spec.Platform.OpenStack.Flavor,
+		// The installer's MachinePool-to-MachineSet function will distribute the generated
+		// MachineSets across the list of Zones. As we don't presently support defining zones
+		// in Hive MachinePools, make sure we send at least a list of one zone so that we
+		// get back a MachineSet.
+		// Providing the empty string will give back a MachineSet running on the default
+		// OpenStack Nova availability zone.
+		Zones: []string{""},
 	}
 
 	if pool.Spec.Platform.OpenStack.RootVolume != nil {
--- a/pkg/controller/remotemachineset/openstackactuator.go
+++ b/pkg/controller/remotemachineset/openstackactuator.go
@@ -1,15 +1,21 @@
 package remotemachineset
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/gophercloud/utils/openstack/clientconfig"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
 	openstackprovider "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis"
 	openstackproviderv1alpha1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	installosp "github.com/openshift/installer/pkg/asset/machines/openstack"
@@ -17,13 +23,15 @@ import (
 	installertypesosp "github.com/openshift/installer/pkg/types/openstack"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
 )
 
 // OpenStackActuator encapsulates the pieces necessary to be able to generate
 // a list of MachineSets to sync to the remote cluster.
 type OpenStackActuator struct {
-	logger  log.FieldLogger
-	osImage string
+	logger     log.FieldLogger
+	osImage    string
+	kubeClient client.Client
 }
 
 var _ Actuator = &OpenStackActuator{}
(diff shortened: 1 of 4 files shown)


joelddiaz:
this depends on the changes in openshift/installer#4196

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -1,29 +1,38 @@
...
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
 
+	"github.com/gophercloud/utils/openstack/clientconfig"
```
staebler, line 11:
```suggestion
	"github.com/gophercloud/utils/openstack/clientconfig"
```

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -134,3 +156,46 @@ func decodeOpenStackMachineProviderSpec(rawExt *runtime.RawExtension, scheme *ru
...
+// unmarshaled into the format expected by the OpenStack clients.
+type yamlOptsBuilder struct {
+	cloudYaml map[string]clientconfig.Cloud
+}
+
+func newYamlOptsBuilder(kubeClient client.Client, credsSecretKey types.NamespacedName) (*yamlOptsBuilder, error) {
+
+	credsSecret := &corev1.Secret{}
+	if err := kubeClient.Get(context.TODO(), credsSecretKey, credsSecret); err != nil {
+		return nil, err
```
staebler, line 171:
```suggestion
		return nil, errors.Wrap(err, "failed to get OpenStack credentials")
```

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -134,3 +156,46 @@ func decodeOpenStackMachineProviderSpec(rawExt *runtime.RawExtension, scheme *ru
...
+	}
+
+	cloudsYaml, ok := credsSecret.Data[constants.OpenStackCredentialsName]
+	if !ok {
+		return nil, errors.New("did not find credentials in the ClusterDeployment's secret")
```
staebler, line 176:
```suggestion
		return nil, errors.New("did not find credentials in the OpenStack credentials secret")
```

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -134,3 +156,46 @@ func decodeOpenStackMachineProviderSpec(rawExt *runtime.RawExtension, scheme *ru
...
+	}
+
+	var clouds clientconfig.Clouds
+	if err := yaml.Unmarshal(cloudsYaml, &clouds); err != nil {
```
staebler, line 180: How stable is `clientconfig.Clouds`? Do you anticipate any issues with supporting older OpenShift versions that may use an older version of this type?

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -134,3 +156,46 @@ func decodeOpenStackMachineProviderSpec(rawExt *runtime.RawExtension, scheme *ru
...
+}
+
+func (opts *yamlOptsBuilder) LoadCloudsYAML() (map[string]clientconfig.Cloud, error) {
+	return opts.cloudYaml, nil
+}
+
+func (opts *yamlOptsBuilder) LoadSecureCloudsYAML() (map[string]clientconfig.Cloud, error) {
+	// secure.yaml is optional so just pretend it doesn't exist
+	var secureClouds clientconfig.Clouds
+	return secureClouds.Clouds, nil
```
staebler, line 197: Isn't this just an nil map?
```suggestion
	return nil, nil
```

pkg/controller/remotemachineset/openstackactuator.go
```diff
@@ -76,6 +76,7 @@ func (a *OpenStackActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, po
 	computePool := baseMachinePool(pool)
 	computePool.Platform.OpenStack = &installertypesosp.MachinePool{
 		FlavorName: pool.Spec.Platform.OpenStack.Flavor,
+		Zones:      []string{""},
```
staebler, line 79: The commit message is not quite right. It states that Hive is providing "an empty (but non-nil) list of zones." However, Hive is actually providing a list with a single empty entry.

I also think it is worth a comment in the code to explain what is happening here.

staebler, line 79: It looks like Hive needs to start supporting zones in MachinePools for OpenStack.
https://github.com/openshift/installer/blob/96579442a7526367e86cf916ec85089c18dfaf24/pkg/types/openstack/machinepool.go#L26-L29

staebler:
Is there any chance that we can re-enable the unit test for the OpenStack actuator as part of this PR? Or are we still far away from that since we have no way of mocking out the call to OpenStack that is baked into the installer?

pkg/controller/remotemachineset/openstackactuator.go
joelddiaz: This is a good question. Outside of this PR I've never dealt with this package before.
Now that you ask, I would worry not so much about versions of OpenShift but the version of OpenStack that this package will be making calls into. Is there someone more familiar with OpenStack that can perhaps answer this?

As far as OpenShift releases, older versions of the installer made you specify whether this "trunk" support was available, and now it is dynamic. And we mark the field as optional in Hive https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1/openstack/platform.go#L19. This feels like we've just automated detection of what you previously had to specify.

pkg/controller/remotemachineset/openstackactuator.go
joelddiaz: Agreed. Created a Jira card on our backlog.

joelddiaz:
> Is there any chance that we can re-enable the unit test for the OpenStack actuator as part of this PR? Or are we still far away from that since we have no way of mocking out the call to OpenStack that is baked into the installer?

Not yet. We would need more reworking of the installer repo to allow us to stub out the OpenStack client.

joelddiaz:
/assign @staebler 
/cc @akhil-rane 

this PR pre-dates us explicitly assigning and CCing people...

joelddiaz:
staebler's hive watch has ended..
/assign @dgoodwin

PR #1142, opened 2020-09-18
Log first success times when we report the metrics.

dgoodwin:
This will give us some info to dig a little deeper in terms of which
cluster was involved and what to search for in logs.

/assign @suhanime


staebler:
/hold

so that this does not keep retesting with the verify failure

staebler:
/hold cancel

Verify test has been fixed.

PR #1143, opened 2020-09-18
dockerfiles: use go 1.15

staebler:
Change dockerfiles to use go 1.15.

We need to upgrade the dependency on github.com/openshift/library-go. That necessitates upgrading github.com/openshift/build-machinery-go. That, in turn, requires that we build with at least go 1.14.4.


staebler:
This PR will need to be force merged. It is the first step in fixing `test verify` failures that are occurring in all PRs.

staebler:
/hold

Hold this so that it does not continually retest.

dgoodwin:
lgtm as well.

dgoodwin:
We can force this per the comment in #1144 once the hivecop opens the merge window today.

PR #1145, opened 2020-09-21
Add hiveutil commands to create clusterpool and clusterclaim

suhanime:


contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
```
staebler, line 11: Separate the k8s imports.
```suggestion
	log "github.com/sirupsen/logrus"
	"github.com/spf13/cobra"
	
	"k8s.io/apimachinery/pkg/api/meta"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/kubernetes/scheme"
```

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+
+	homeDir string
+}
+
+func NewClaimClusterPoolCommand() *cobra.Command {
+	opt := &ClusterClaimOptions{}
+
+	opt.homeDir = "."
+
+	defaultDuration, _ := time.ParseDuration("8h")
```
staebler, line 31:
```suggestion
	defaultDuration := 8 * time.Hour
```

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "claim CLUSTER_POOL_NAME",
+		Short: "claim a cluster from a clusterpool",
+		Long:  "claims a cluster from the clusterpool in the given namespace",
```
staebler, line 36:
```suggestion
		Short: "claim a cluster from a ClusterPool",
		Long:  "claims a cluster from the ClusterPool in the given namespace",
```

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+			if err != nil {
+				log.WithError(err).Fatal("Error")
+			}
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVarP(&opt.Namespace, "namespace", "n", "",
+		"Namespace to create cluster claim in. Has to be the namespace in which the cluster pool is deployed")
+	flags.DurationVar(&opt.Lifetime, "lifetime", defaultDuration, "Lifetime of the cluster claim. "+
```
staebler, line 51: I would have expected that the default would be to not set a lifetime.

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+			log.Error("Cannot determine default namespace")
+			return err
+		}
+	}
+	accessor, err := meta.Accessor(claim)
+	if err != nil {
+		log.WithError(err).Errorf("Cannot create accessor for object of type %T", claim)
+		return err
+	}
+	accessor.SetNamespace(o.Namespace)
```
staebler, line 79:
```suggestion
	claim.Namespace = o.Namespace
```

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+	flags.StringVarP(&opt.Namespace, "namespace", "n", "",
+		"Namespace to create cluster claim in. Has to be the namespace in which the cluster pool is deployed")
+	flags.DurationVar(&opt.Lifetime, "lifetime", defaultDuration, "Lifetime of the cluster claim. "+
+		"Default is 8 hours")
+
+	return cmd
+}
+
+func (o ClusterClaimOptions) run() error {
+	if err := apis.AddToScheme(scheme.Scheme); err != nil {
```
staebler, line 58: Create a new scheme rather than using the global scheme.

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (o ClusterClaimOptions) generateClaim() *hivev1.ClusterClaim {
+	return &hivev1.ClusterClaim{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ClusterClaim",
+			APIVersion: hivev1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        o.getClusterClaimName(),
+			Annotations: map[string]string{},
```
staebler, line 95: This is not needed.
```suggestion
```

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		Spec: hivev1.ClusterClaimSpec{
+			ClusterPoolName: o.ClusterPoolName,
+			Lifetime:        &metav1.Duration{Duration: o.Lifetime},
+		},
+	}
+}
+
+func (o ClusterClaimOptions) getClusterClaimName() string {
+	return fmt.Sprintf("%s-claim", o.ClusterPoolName)
```
staebler, line 105: The user needs to be able to specify the name that they want for the ClusterClaim. Otherwise, the user will only ever be able to create one per ClusterPool, as the name must be unique.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/user"
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
```
staebler, line 11: Seperate k8s imports.
```suggestion
	"github.com/spf13/cobra"
	
	corev1 "k8s.io/api/core/v1"
```

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+The hiveutil clusterpool create-pool command generates and applies the
+artifacts needed to create a new Hive cluster pool.
+Currently cluster pools are supported in AWS, Azure and GCP.
+
+IMAGES
+An existing ClusterImageSet can be specified with the --image-set
+flag. Otherwise, one will be generated using the images specified for the
+cluster pool. If you don't wish to use a ClusterImageSet, specify
+--use-image-set=false. This will result in images only specified on the
+cluster itself.
```
staebler, line 45: ClusterPool requires the use of a ClusterImageSet.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+	PullSecret         string
+	PullSecretFile     string
+	Cloud              string
+	CredsFile          string
+	CloudSecretName    string
+	createCloudSecret  bool
+	ClusterImageSet    string
+	ReleaseImage       string
+	ReleaseImageSource string
+	UseClusterImageSet bool
```
staebler, line 83: This is not used in generating the ClusterPool and is not supported by ClusterPool.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+	Region             string
+	Size               int32
+
+	awsAccessKeyID     string
+	awsSecretAccessKey string
+
+	azureServicePrincipal            []byte
+	AzureBaseDomainResourceGroupName string
```
staebler, line 91: Why are some options exported and some are not?

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/openshift/hive/pkg/apis"
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+)
+
+type ClusterClaimOptions struct {
+	Namespace       string
+	Lifetime        time.Duration
+	ClusterPoolName string
+
+	homeDir string
```
staebler, line 23: This is not used when generating the ClusterClaim.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+
+	homeDir string
+}
+
+func NewCreateClusterPoolCommand() *cobra.Command {
+	opt := &ClusterPoolOptions{}
+
+	opt.homeDir = "."
+
+	if u, err := user.Current(); err == nil {
```
staebler, line 103: Use `HomeDir` from `k8s.io/client-go/util/homedir` to get the home directory.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+		opt.homeDir = u.HomeDir
+	}
+
+	defaultPullSecretFile := filepath.Join(opt.homeDir, ".pull-secret")
+	if _, err := os.Stat(defaultPullSecretFile); os.IsNotExist(err) {
+		defaultPullSecretFile = ""
+	} else if err != nil {
+		log.WithError(err).Errorf("%v can not be used", defaultPullSecretFile)
```
staebler, line 111: This should not be logged as an error until it is determined that the file is going to be the one actually used. In other words, if the user specifies a pull secret, then it is a bad experience to present the user with an error that the default pull secret cannot be used, given that the user has stated not to use it.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+		log.Infof("Unsupported cloud: %s", o.Cloud)
+		return fmt.Errorf("unsupported cloud: %s", o.Cloud)
+	}
+
+	return nil
+}
+
+// run executes the command
+func (o *ClusterPoolOptions) run() error {
+	if err := apis.AddToScheme(scheme.Scheme); err != nil {
```
staebler, line 195: Use a new scheme.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+	if len(pullSecret) != 0 {
+		result = append(result, o.generatePullSecret(pullSecret))
+	}
+
+	return result, nil
+}
+
+func (o *ClusterPoolOptions) configureImages() (*hivev1.ClusterImageSet, error) {
+	if len(o.ClusterImageSet) > 0 {
```
staebler, line 285: Should it be an error if the user uses `--image-set` in addition to either `--release-image` or `--release-image-source`?

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+
+func (o *ClusterPoolOptions) build() ([]runtime.Object, error) {
+	// validate
+	if len(o.Name) == 0 {
+		return nil, fmt.Errorf("name is required")
+	}
+	if len(o.BaseDomain) == 0 {
+		return nil, fmt.Errorf("BaseDomain is required")
+	}
+	if len(o.ClusterImageSet) == 0 && len(o.ReleaseImage) == 0 {
```
staebler, line 323: This makes it impossible for the user to use the `--release-image` flag since `o.ClusterImageSet` will be set in the `configureImages` function.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+		Spec: hivev1.ClusterImageSetSpec{
+			ReleaseImage: o.ReleaseImage,
+		},
+	}
+	o.ClusterImageSet = imageSet.Name
+	return imageSet, nil
+}
+
+func (o *ClusterPoolOptions) build() ([]runtime.Object, error) {
+	// validate
```
staebler, line 316: As the comment alludes to, the following should be in the `validate` function.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+	if o.ClusterImageSet != "" {
+		cp.Spec.ImageSetRef = hivev1.ClusterImageSetReference{Name: o.ClusterImageSet}
+	}
+	o.addClusterPoolPlatform(cp)
+
+	return cp
+}
+
+func (o *ClusterPoolOptions) generateCredentialsSecret() *corev1.Secret {
```
staebler, line 419: I would like to see the `CloudBuilder` from `pkg/clusterresource` used to generate the creds secret rather than duplicating the code from there. Export the existing `generateCredentialsSecret` function.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+		}
+	case cloudGCP:
+		secret.Data = map[string][]byte{
+			constants.GCPCredentialsName: o.gcpServiceAccount,
+		}
+	}
+	return secret
+}
+
+func (o *ClusterPoolOptions) generatePullSecret(pullSecret string) *corev1.Secret {
```
staebler, line 449: Export the `generatePullSecretSecret` from the `Builder` in `pkg/clusterresource`.

contrib/pkg/utils/generic.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	payload := &releasePayload{}
+	err = json.Unmarshal(data, payload)
+	if err != nil {
+		return "", err
+	}
+	return payload.PullSpec, nil
+}
+
+func GetResourceHelper(command string) (resource.Helper, error) {
```
staebler, line 39: This should take a logger rather than the name of the command.
```suggestion
func GetResourceHelper(logger log.FieldLogger) (resource.Helper, error) {
```

contrib/pkg/utils/generic.go
```diff
@@ -0,0 +1,74 @@
...
+	if len(envPullSecret) > 0 {
+		return envPullSecret, nil
+	}
+	if len(pullSecret) > 0 {
+		return pullSecret, nil
+	}
+	if len(pullSecretFile) > 0 {
+		data, err := ioutil.ReadFile(pullSecretFile)
+		if err != nil {
+			log.Error("Cannot read pull secret file")
```
staebler, line 67: `GetPullSecret` should take a logger as a parameter rather than using the global logger.

contrib/pkg/clusterpool/clusterclaim.go
dgoodwin: Agreed.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+// validate ensures that option values make sense
+func (o *ClusterPoolOptions) validate(cmd *cobra.Command) error {
+	if !o.UseClusterImageSet && len(o.ClusterImageSet) > 0 {
+		cmd.Usage()
+		log.Info("If not using cluster image sets, do not specify the name of one")
+		return fmt.Errorf("invalid option")
+	}
+	if !validClouds[o.Cloud] {
+		cmd.Usage()
+		log.Infof("Unsupported cloud: %s", o.Cloud)
```
dgoodwin, line 186: Feels like this should be Errorf

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	rh, err := utils.GetResourceHelper("create-pool")
+	if err != nil {
+		return err
+	}
+	if len(o.Namespace) == 0 {
+		o.Namespace, err = utils.DefaultNamespace()
+		if err != nil {
+			log.Error("Cannot determine default namespace")
```
dgoodwin, line 211: WithError(err) would be useful

contrib/pkg/utils/generic.go
dgoodwin: You can just create a contextual logger in run(), or I think you could also pass the one from your commands main down into the "Options" struct you have here as a non-exported field, and then use it as o.logger.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,419 @@
...
+// validate ensures that option values make sense
+func (o *ClusterPoolOptions) validate(cmd *cobra.Command) error {
+	if len(o.Name) == 0 {
+		return fmt.Errorf("name is required")
+	}
+	if len(o.BaseDomain) == 0 {
+		return fmt.Errorf("base domain is required")
+	}
+	if (len(o.ClusterImageSet) != 0 && len(o.ReleaseImage) != 0) ||
+		(len(o.ClusterImageSet) != 0 && len(o.ReleaseImageSource) != 0) {
```
akhil-rane, line 175: nit: `(len(o.ClusterImageSet) != 0 && ( len(o.ReleaseImage) != 0) ||  len(o.ReleaseImageSource) != 0))` in the if expression will give same the result?

akhil-rane (commented):
Looks good @suhanime. Few suggestions

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	return allObjects, nil
+}
+
+func (o *ClusterPoolOptions) getPullSecretSecretName() string {
+	return fmt.Sprintf("%s-pull-secret", o.Name)
+}
+
+func (o *ClusterPoolOptions) addClusterPoolPlatform(cp *hivev1.ClusterPool) {
```
staebler, line 347: The code in this function is a duplicate of what is done by the CloudBuilder `addClusterDeploymentPlatform` function. Can you change that function to return a `*hivev1.Platform` instead of setting the platform on a cluster deployment?

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,419 @@
...
+	if o.createCloudSecret {
+		switch o.Cloud {
+		case cloudAWS:
+			return fmt.Sprintf("%s-aws-creds", o.Name)
+		case cloudAzure:
+			return fmt.Sprintf("%s-azure-creds", o.Name)
+		case cloudGCP:
+			return fmt.Sprintf("%s-gcp-creds", o.Name)
+		}
+		return o.CloudSecretName
```
staebler, line 390:
```suggestion
```

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,419 @@
...
+
+func (o *ClusterPoolOptions) credsSecretName() string {
+	if o.createCloudSecret {
+		switch o.Cloud {
+		case cloudAWS:
+			return fmt.Sprintf("%s-aws-creds", o.Name)
+		case cloudAzure:
+			return fmt.Sprintf("%s-azure-creds", o.Name)
+		case cloudGCP:
+			return fmt.Sprintf("%s-gcp-creds", o.Name)
```
staebler, line 388: This knowledge is already known in the CloudBuilder. We should not need to take it out from there. Add a function to the CloudBuilder that will return the creds secret name.

contrib/pkg/utils/generic.go
suhanime: While explicit logger to be passed down from hiveutil mail would be ideal, I wonder if simply passing an empty context is enough for this case.

contrib/pkg/clusterpool/clusterpool.go
staebler: I don't follow what you are saying. I am suggesting that the code in `addClusterDeploymentPlatform` that creates a `hivev1.Platform` object be extracted out of that function so that it can be used by `addClusterDeploymentPlatform` and `addClusterPoolPlatform`. I don't understand what `Spec` has to do with that.

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/openshift/hive/contrib/pkg/utils"
+	"github.com/openshift/hive/pkg/apis"
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+)
+
+type ClusterClaimOptions struct {
+	Name            string
+	Namespace       string
+	Lifetime        time.Duration
+	ClusterPoolName string
```
dgoodwin, line 24: Regarding the confusion around logging: suggest adding a field here: "log log.FieldLogger"

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func NewClaimClusterPoolCommand() *cobra.Command {
+	opt := &ClusterClaimOptions{}
```
dgoodwin, line 28: opt := &ClusterClaimOptions{log: log.WithField("command": "cluster-pool-claim"}

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		Short: "claims a cluster from a ClusterPool",
+		Long:  "claims a cluster from the ClusterPool in the given namespace",
+		Args:  cobra.ExactArgs(2),
+		Run: func(cmd *cobra.Command, args []string) {
+			log.SetLevel(log.InfoLevel)
+			ctx := context.Background()
+			opt.Name = args[0]
+			err := opt.run(ctx)
+			if err != nil {
+				log.WithContext(ctx).WithError(err).Fatal("Error")
```
dgoodwin, line 41: then drop the WithContext(ctx) here, I don't think we typically use WithContext. We may refer to contextual loggers though but this may have gotten confused with the go context. By contextual logger we just mean a logger that's carrying fields we added and passed around.

contrib/pkg/clusterpool/clusterclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		return fmt.Errorf("ClusterPool name is required")
+	}
+
+	scheme := runtime.NewScheme()
+	if err := apis.AddToScheme(scheme); err != nil {
+		return err
+	}
+	claim := o.generateClaim()
+
+	rh, err := utils.GetResourceHelper(ctx, "claim")
```
dgoodwin, line 65: rh, err := utils.GetResourceHelper(o.log)

contrib/pkg/utils/generic.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	payload := &releasePayload{}
+	err = json.Unmarshal(data, payload)
+	if err != nil {
+		return "", err
+	}
+	return payload.PullSpec, nil
+}
+
+func GetResourceHelper(ctx context.Context, command string) (resource.Helper, error) {
```
dgoodwin, line 40: func GetResourceHelper(logger log.FieldLogger) (resource.Helper, error) {

dgoodwin (commented):
Quick suggestions on how I would do the logging here.

contrib/pkg/clusterpool/clusterclaim.go
suhanime: Thank you this was way easier and made sense for the usecase

contrib/pkg/clusterpool/clusterpool.go
suhanime: ... [Ah now I get it](https://media.giphy.com/media/TJawtKM6OCKkvwCIqX/giphy.gif) :disappointed:

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,419 @@
...
+	opt.homeDir = homedir.HomeDir()
+
+	if u, err := user.Current(); err == nil {
+		opt.homeDir = u.HomeDir
+	}
+
+	defaultPullSecretFile := ""
+	if _, err := os.Stat(filepath.Join(opt.homeDir, ".pull-secret")); err == nil {
+		defaultPullSecretFile = filepath.Join(opt.homeDir, ".pull-secret")
+	}
```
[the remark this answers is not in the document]
staebler: This still has the same behavior. If the `~/.pull-secret` file does not exist, the help will use an empty string for the name of the default file.

contrib/pkg/utils/generic.go
```diff
@@ -0,0 +1,78 @@
...
+	rules := clientcmd.NewDefaultClientConfigLoadingRules()
+	kubeconfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, &clientcmd.ConfigOverrides{})
+	ns, _, err := kubeconfig.Namespace()
+	return ns, err
+}
+
+func GetPullSecret(logger log.FieldLogger, pullSecret string, pullSecretFile string) (string, error) {
+	envPullSecret := os.Getenv("PULL_SECRET")
+	if len(envPullSecret) > 0 {
+		if _, err := os.Stat(envPullSecret); os.IsNotExist(err) {
```
staebler, line 59: This is not right. The PULL_SECRET environment variable is the text of a pull secret not the name of a file containing the pull secret.

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,369 @@
...
+	switch {
+	case o.ClusterImageSet != "":
+		releaseImageFlagsUsed++
+	case o.ReleaseImage != "":
+		releaseImageFlagsUsed++
+	case o.ReleaseImageSource != "":
+		releaseImageFlagsUsed++
+	}
+	if releaseImageFlagsUsed == 0 {
+		return fmt.Errorf("must specify one of image set, release image or release image source")
```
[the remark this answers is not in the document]
suhanime: I did find something in the description of `Use` in https://godoc.org/github.com/spf13/cobra#Command
However, neither of these flags here are mandatory since we do have a default value of `release-image-source`
Maybe adding the `cmd.Usage()` here would help?

dgoodwin:
One outstanding item but I think it's non-critical and I'd like to get this merged.

/lgtm

contrib/pkg/clusterpool/clusterpool.go
```diff
@@ -0,0 +1,365 @@
...
+specified, default image is obtained from the following URL:
+https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest
+
+ENVIRONMENT VARIABLES
+The command will use the following environment variables for its output:
+
+PULL_SECRET - If present, it is used as the cluster pool's pull
+secret and will override the --pull-secret flag. If not present, and
+the --pull-secret flag is not specified, then the --pull-secret-file is
+used.
```
staebler, line 51: There is still no way for the user to determine what the name of the default pull secret file is if it does not already exist.

staebler (commented):
> One outstanding item but I think it's non-critical and I'd like to get this merged.
> 
> /lgtm

Fun times.

PR #1147, opened 2020-09-21
CO-1187: Latest hive doesn't honor syncset-pause annotation

abutcher:
/assign @staebler 
/cc @dgoodwin

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -282,6 +282,11 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	// Is syncing paused?
+	if !controllerutils.ShouldSyncCluster(cd, logger) {
+		return reconcile.Result{}, nil
+	}
+
 	if cd.DeletionTimestamp != nil {
 		logger.Debug("cluster is being deleted")
 		return reconcile.Result{}, nil
--- a/pkg/controller/clustersync/clustersync_controller_test.go
+++ b/pkg/controller/clustersync/clustersync_controller_test.go
@@ -278,6 +278,10 @@ func TestReconcileClusterSync_NoWorkToDo(t *testing.T) {
 				}),
 			),
 		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -282,6 +282,11 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	// Is syncing paused?
```
2uasimojo, line 285: Should we consider putting this [after creating the initial ClusterSync and ClusterSyncLease](https://github.com/openshift/hive/pull/1147/files#diff-d64ff608a67db50bd6a15a35465607c1R341)?

pkg/controller/clustersync/clustersync_controller_test.go
```diff
@@ -278,6 +278,10 @@ func TestReconcileClusterSync_NoWorkToDo(t *testing.T) {
 				}),
 			),
 		},
+		{
+			name: "syncset pause",
+			cd:   cdBuilder(scheme).GenericOptions(testgeneric.WithAnnotation(constants.SyncsetPauseAnnotation, "true")).Build(),
+		},
```
2uasimojo, line 284: Confirmed this test fails when the fix is reverted ✓

2uasimojo (commented):
Fix looks good. One question/suggestion inline. I shall avoid official lgtm tag in any case, because that seems to merge things _sometimes_.

pkg/controller/clustersync/clustersync_controller.go
dgoodwin: This looks consistent with what would happen today if the cluster was marked unreachable. I can't think of a reason we'd need a clustersync created for a new cluster that was immediately paused.

2uasimojo:
Is there a way to stop this burning CI resources until the issue is resolved?

staebler:
> Is there a way to stop this burning CI resources until the issue is resolved?

/hold

staebler:
/hold cancel

Verify tests have been fixed.

PR #1148, opened 2020-09-21
add hive-backend role

joelddiaz:
This role/ServiceAccount can be used while investigating issues by allowing viewing ClusterDeployments.
Also, the granting of access to Secrets allows access to the kubeconfig's for the managed clusters for investigating issues with the managed clusters.1
This is related to the request found in https://issues.redhat.com/browse/CO-1181

### DIFF
--- a/config/rbac/hive_backend_role.yaml
+++ b/config/rbac/hive_backend_role.yaml
@@ -1,0 +1,24 @@
+# hive-backend is a role intended to enable groups running hive to have
+# debug access when investigating issues (whether the issue is with
+# hive or with the cluster(s) being managed).
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: hive-backend
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - hive.openshift.io
+  resources:
+  - clusterdeployments
+  verbs:
+  - get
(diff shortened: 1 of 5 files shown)


config/rbac/hive_backend_role.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ClusterRole
+metadata:
+  name: hive-backend
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
```
staebler, line 14: Are we absolutely sure that we want to grant read access to all of the secrets in the cluster to the hive-backend user?

joelddiaz: One of the use cases was to have access to the kubeconfig for the managed clusters 🤷

gregsheremeta: My initial reaction was "this will not be allowed in OSD." Then I saw the ticket was created by SRE-P for backplane.

/cc @jaybeeunix @jewzaam @wanghaoran1988 @jmelis please advise.

gregsheremeta:
/hold

Who or what will have access to hive-backend? I know the ability to read secrets is restricted for compliance reasons.

Need to make sure this is ok with @jaybeeunix @jewzaam @wanghaoran1988 @jmelis

dgoodwin:
Via Jira is sounds like they no longer want this. :(

Sorry @joelddiaz. Closing for now.

/close

PR #1149, opened 2020-09-22
Tag librarygo version

suhanime:
Ignore librarygo from installer that requires missing goautoneg dependencies

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -83,3 +83,7 @@ replace (
 
 // needed because otherwise v12.0.0 is picked up as a more recent version
 replace k8s.io/client-go => k8s.io/client-go v0.19.0
+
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -83,3 +83,5 @@ replace (
 
 // needed because otherwise v12.0.0 is picked up as a more recent version
 replace k8s.io/client-go => k8s.io/client-go v0.19.0
+
+replace github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20200918101923-1e4c94603efe
```
staebler, line 87: Please add a comment above this explaining why it is necessary, so that we know when to update it and when it can be removed.

PR #1151, opened 2020-09-23
Allow mutating ClusterDeployment.Spec.HibernateAfter.

dgoodwin:


dgoodwin:
/assign @abutcher 

Just assigning one person given the size of this.

dgoodwin:
It looks like we don't unit test mutability of individual properties, just that the general mechanism is working and then trust that it will for additional mutable properties.

PR #1152, opened 2020-09-23
Modify 'verify' target to execute 'go mod vendor'

suhanime:
Currently our CI tests do not run 'go mod vendor'
This commit ensures 'vendor' target in Makefile is ran on every PR in
order to check that all the packages vendored in are downloadable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,8 @@ crd: ensure-controller-gen ensure-yq
 	$(foreach p,$(wildcard ./config/crds/*.yaml),$(call strip-yaml-break,$(p)))
 update: crd
 
+verify: vendor
+
 .PHONY: verify-crd
 verify-crd: ensure-controller-gen ensure-yq
 	./hack/verify-crd.sh


staebler:
This is not what verify should do. The CI should be using what is already in `vendor`. We could verify that the `vendor` directory is correct, but the tests should not be relying on updating the contents of `vendor`.

csrwng:
@staebler it's not about updating the contents of vendor, it's really verifying that the vendor directory can be rebuilt with the contents of go.mod.
@suhanime I would probably go about this differently. In a test specific to validating that the vendor directory can be rebuilt, I would rename the existing vendor directory, try to do a go mod vendor and have it succeed or fail, then discard the resulting vendor directory and restore the original one.

staebler:
> @staebler it's not about updating the contents of vendor, it's really verifying that the vendor directory can be rebuilt with the contents of go.mod.
> @suhanime I would probably go about this differently. In a test specific to validating that the vendor directory can be rebuilt, I would rename the existing vendor directory, try to do a go mod vendor and have it succeed or fail, then discard the resulting vendor directory and restore the original one.

Yes, that is what I am suggesting. And we should take it one step further than just verifying that the directory can be rebuilt. We should be verifying that there are no changes to the directory or to go.mod.

PR #1154, opened 2020-09-23
suport clusters installed into manually created Azure resource groups

joelddiaz:
New versions of the installer supports installing an Azure cluster into a pre-existing resource group.
The name of the resource group needs to be used when deprovisioning a cluster that was installed this way.

- [x] Update hiveutil with the new --resource-group param
- [x] Update clusterdeployment and clusterdeprovision types to hold this field
- [x] Update clusterdeployment controller to set the resourceGroupName field for the clusterdeprovision
- [x] Update deprovision job generator to pass the --resource-group param when it is set in the clusterdeprovision
- [x] Add test cases
- [x] Update hibernation controller to use the resourceGroupName to find instances

### DIFF
--- a/contrib/pkg/deprovision/azure.go
+++ b/contrib/pkg/deprovision/azure.go
@@ -15,9 +15,15 @@ import (
 	azureutils "github.com/openshift/hive/contrib/pkg/utils/azure"
 )
 
+type azureParams struct {
+	logLevel      string
+	resourceGroup string
+}
+
 // NewDeprovisionAzureCommand is the entrypoint to create the azure deprovision subcommand
 func NewDeprovisionAzureCommand() *cobra.Command {
-	var logLevel string
+	var params azureParams
+
 	cmd := &cobra.Command{
 		Use:   "azure INFRAID",
 		Short: "Deprovision Azure assets (as created by openshift-installer)",
@@ -26,7 +32,7 @@ func NewDeprovisionAzureCommand() *cobra.Command {
 			if err := validate(); err != nil {
 				log.WithError(err).Fatal("Failed validating Azure credentials")
 			}
-			uninstaller, err := completeAzureUninstaller(logLevel, args)
+			uninstaller, err := completeAzureUninstaller(args, params)
 			if err != nil {
 				log.WithError(err).Error("Cannot complete command")
 				return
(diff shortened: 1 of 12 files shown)


contrib/pkg/deprovision/azure.go
```diff
@@ -15,9 +15,17 @@ import (
 	azureutils "github.com/openshift/hive/contrib/pkg/utils/azure"
 )
 
+type azureParams struct {
+	logLevel      string
+	resourceGroup string
+}
+
+var (
+	params azureParams
```
staebler, line 24: Please do not use a package-level variable for this.

pkg/apis/hive/v1/azure/platform.go
```diff
@@ -18,6 +18,11 @@ type Platform struct {
 
 	// BaseDomainResourceGroupName specifies the resource group where the azure DNS zone for the base domain is found
 	BaseDomainResourceGroupName string `json:"baseDomainResourceGroupName,omitempty"`
+
+	// ResourceGroupName is used to indicate that the cluster was installed into an already-existing
+	// Azure resource group (instead of having the installer create the resource group based on the
+	// cluster name). This will be used to indicate to the uninstall process where it should clean up.
```
staebler, line 24: Let's make this description more user-focused.
```suggestion
	// ResourceGroupName specifies the name of the Azure resource group containing the cluster.
	// This must match the name of the Azure resource group specified in the InstallConfig.
	// If the name of the Azure resource group is omitted from the InstallConfig, then it must be
	// omitted here as well.
```

pkg/controller/hibernation/azure_actuator.go
```diff
@@ -187,6 +187,12 @@ func azureMachinePowerState(vm compute.VirtualMachine) string {
...
 
 func clusterDeploymentResourceGroup(cd *hivev1.ClusterDeployment) string {
+	// For the case where a specific resource group is specified (eg the cluster was installed
+	// into a pre-existing resource group), use it.
+	if cd.Spec.Platform.Azure != nil && cd.Spec.Platform.Azure.ResourceGroupName != "" {
+		return cd.Spec.Platform.Azure.ResourceGroupName
+	}
+
 	// TODO: Fix this to use explicit resource group name when we
 	// collect that from the installer.
```
staebler, line 197: Remove this TODO.

staebler (approved):
/lgtm

PR #1155, opened 2020-09-24
Add support for ClusterPool.Spec.Labels to apply to new ClusterDeploy…

dgoodwin:
…ments.

This will not affect any cluster already created and in the pool, or
claimed clusters from the pool.

### DIFF
--- a/pkg/apis/hive/v1/clusterpool_types.go
+++ b/pkg/apis/hive/v1/clusterpool_types.go
@@ -28,6 +28,11 @@ type ClusterPoolSpec struct {
 	// ImageSetRef is a reference to a ClusterImageSet. The release image specified in the ClusterImageSet will be used
 	// by clusters created for this cluster pool.
 	ImageSetRef ClusterImageSetReference `json:"imageSetRef"`
+
+	// Labels to be applied to new ClusterDeployments created for the pool. ClusterDeployments that have already been
+	// claimed will not be affected when this value is modified.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
 }
 
 // ClusterPoolStatus defines the observed state of ClusterPool
--- a/pkg/controller/clusterpool/clusterpool_controller_test.go
+++ b/pkg/controller/clusterpool/clusterpool_controller_test.go
@@ -76,15 +76,17 @@ func TestReconcileClusterPool(t *testing.T) {
 		expectedMissingDependenciesMessage string
 		expectedAssignedClaims             int
 		expectedUnassignedClaims           int
+		expectedLabels                     map[string]string // Tested on all clusters, so will not work if your test has pre-existing cds in the pool.
 	}{
 		{
 			name: "create all clusters",
 			existing: []runtime.Object{
-				poolBuilder.Build(testcp.WithSize(5)),
+				poolBuilder.Build(testcp.WithSize(5), testcp.WithClusterDeploymentLabels(map[string]string{"foo": "bar"})),
 			},
 			expectedTotalClusters: 5,
 			expectedObservedSize:  0,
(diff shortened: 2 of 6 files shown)


pkg/apis/hive/v1/clusterpool_types.go
```diff
@@ -28,6 +28,10 @@ type ClusterPoolSpec struct {
 	// ImageSetRef is a reference to a ClusterImageSet. The release image specified in the ClusterImageSet will be used
 	// by clusters created for this cluster pool.
 	ImageSetRef ClusterImageSetReference `json:"imageSetRef"`
+
+	// Labels to be applied to new ClusterDeployments created for the pool. ClusterDeployments already in the
+	//pool, or claimed from the pool, will not be affected when this value is modified.
```
staebler, line 33:
```suggestion
	// pool, or claimed from the pool, will not be affected when this value is modified.
```

staebler, line 33: Is the "already in the pool" part long-term or just until we have the clusterpool controller replace ClusterDeployments on changes to the ClusterPool?

pkg/apis/hive/v1/clusterpool_types.go
```diff
@@ -28,6 +28,10 @@ type ClusterPoolSpec struct {
...
+	Labels map[string]string `json:"labels,omitempty"`
```
staebler, line 34:
```suggestion
	// +optional
	Labels map[string]string `json:"labels,omitempty"`
```

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -468,6 +470,11 @@ func TestReconcileClusterPool(t *testing.T) {
 
 			for _, cd := range cds.Items {
 				assert.Equal(t, hivev1.HibernatingClusterPowerState, cd.Spec.PowerState, "expected cluster to be hibernating")
+				if test.expectedLabels != nil {
+					for k, v := range test.expectedLabels {
+						assert.Equal(t, v, cd.Labels[k])
+					}
+				}
```
staebler, line 477:
```
				assert.Equal(t, test.expectedLabels, cd.Labels, "unexpected labels on cluster deployment")
```

pkg/test/clusterpool/clusterpool.go
```diff
@@ -94,6 +94,12 @@ func WithSize(size int) Option {
 	}
 }
 
+func WithLabels(labels map[string]string) Option {
```
staebler, line 97: This is confusing since it is not clear from the name of the function whether it is setting `.metadata.labels` or `.spec.labels`.

pkg/apis/hive/v1/clusterpool_types.go
dgoodwin: If we pursue replacing every current pool on spec changes, this limitation will go away implicitly. I'll remove the note in documentation if we don't think it's permanent.

pkg/controller/clusterpool/clusterpool_controller_test.go
dgoodwin: This does pass today, but I didn't want to encode the assumption that these would be the exact and only labels on the resulting clusterdeployment. We inject labels for region and version, that isn't happening in this test right now but it seems more correct to assert just the labels we expect are there, rather than assume they'll be the only ones.

staebler: :+1:

staebler (commented):
Looks good, once CRD changes are made.

dgoodwin:
Ok should be good now after rebasing off master and replacing controller gen.

PR #1156, opened 2020-09-24
makefile: use earlier version of controller-gen

staebler:
Force build-machinery-go to grab an earlier version of controller-gen. The newer version adds annotations on pod specs that will cause the CRDs to fail validation. The fix for that requires the CRDs to use v1, but we still need to support users that are running kube versions that only have v1beta1.

https://issues.redhat.com/browse/CO-1193

/assign @dgoodwin 
/cc @abutcher

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 .PHONY: all
 all: vendor update test build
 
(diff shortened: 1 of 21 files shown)


gregsheremeta:
/lgtm

@achuzhoy verifies it works

PR #1157, opened 2020-09-24
Refine ClusterDeployment printcolumns.

dgoodwin:
Drop ClusterName, it's virtually always identical to the
ClusterDeployment.Name.

Drop BaseDomain, doesn't seem useful.

Add Platform, Region, OpenShift version, and the observed power state
from the Hibernating condition.


akhil-rane (commented):
/lgtm

dgoodwin:
e2e test script had a sanity check to make sure the printer was working and it was looking for basedomain. Updated to look for infraid which is still there.

PR #1158, opened 2020-09-25
types: add status sub-resource to syncsets

staebler:
Add a status sub-resource to the SyncSet and SelectorSyncSet types. With older versions of kubernetes (1.11 as an example), the CRD strategy will not update the generation of the resource if the CRD does not have a status sub-resource.

See https://github.com/kubernetes/kubernetes/blob/v1.11.0/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L90

https://issues.redhat.com/browse/CO-1191

/assign @abutcher 
/cc @dgoodwin


dgoodwin (approved):
LGTM, over to you @abutcher

PR #1159, opened 2020-09-28
Fix stalled leader election after Deployment updates

akhil-rane:
With the current Deployment strategy of rolling update, a new pod is brought up
before the old is terminated. The new pod leader election has already determined
someone else is leader before the old releases it's lock, and will wait the full
renew interval of 270 seconds before checking again.

To fix we switch to the recreate strategy which will delete the old pod first,
then create the new. This seems the desired strategy for leader election based
operators.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -7,8 +7,6 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"github.com/google/uuid"
@@ -159,6 +157,10 @@ func newRootCommand() *cobra.Command {
 			log.Info("Starting /healthz and /readyz endpoints")
 			go http.ListenAndServe(":8080", nil)
 
+			// use a Go context so we can tell the leaderelection code when we want to step down
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
 			run := func(ctx context.Context) {
 				// Create a new Cmd to provide shared dependencies and start components
 				mgr, err := manager.New(cfg, manager.Options{
@@ -214,27 +216,17 @@ func newRootCommand() *cobra.Command {
 				log.Info("Starting the Cmd.")
 
 				// Start the Cmd
-				log.Fatal(mgr.Start(signals.SetupSignalHandler()))
+				err = mgr.Start(signals.SetupSignalHandler())
+				if err != nil {
+					log.WithError(err).Error("error running manager")
+				}
+				// Canceling the leader election context
+				cancel()
 			}
 
 			// Leader election code based on:
 			// https://github.com/kubernetes/kubernetes/blob/f7e3bcdec2e090b7361a61e21c20b3dbbb41b7f0/staging/src/k8s.io/client-go/examples/leader-election/main.go#L92-L154
 			// This gives us ReleaseOnCancel which is not presently exposed in controller-runtime.
-
-			// use a Go context so we can tell the leaderelection code when we want to step down
-			ctx, cancel := context.WithCancel(context.Background())
-			defer cancel()
-
-			// listen for interrupts or the Linux SIGTERM signal and cancel
-			// our context, which the leader election code will observe and
-			// step down
-			ch := make(chan os.Signal, 1)
-			signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
-			go func() {
-				<-ch
-				log.Info("received termination, signaling shutdown")
-				cancel()
-			}()
 
 			id := uuid.New().String()
 			leLog := log.WithField("id", id)
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -6,8 +6,6 @@ import (
 	golog "log"
 	"net/http"
 	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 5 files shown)


cmd/manager/main.go
```diff
@@ -214,7 +217,12 @@ func newRootCommand() *cobra.Command {
 				log.Info("Starting the Cmd.")
 
 				// Start the Cmd
-				log.Fatal(mgr.Start(signals.SetupSignalHandler()))
+				err = mgr.Start(signals.SetupSignalHandler())
+				if err != nil {
+					log.Fatal(err)
```
staebler, line 222: The leader election context should be cancelled whenever `mgr.Start` returns, regardless of whether it returned an error or not.

cmd/manager/main.go
```diff
@@ -214,7 +217,12 @@ func newRootCommand() *cobra.Command {
 				log.Info("Starting the Cmd.")
 
 				// Start the Cmd
-				log.Fatal(mgr.Start(signals.SetupSignalHandler()))
+				err = mgr.Start(signals.SetupSignalHandler())
```
staebler, line 220: We are missing the part of leader election where we ensure that the controllers are stopped before releasing the leader-election lock. The following order of events should occur.
1) TERM signal is received.
2) Object is sent to managers stop channel.
3) Managers stop.
4) Object is sent to leader-election stop channel.
5) Leader election lock is released.

akhil-rane: I may be wrong about this but looks like stop channel for manager returned by signals.SetupSignalHandler() is a receive only channel. So I think we cannot send anything on it to ensure controllers are stopped.

cmd/manager/main.go
akhil-rane: replacing `log.Fatal` with `log.Error` here

cmd/manager/main.go
staebler: Yes, you cannot use `signals.SetupSignalHandler`. You need to create your own channel to use instead.

cmd/manager/main.go
```diff
@@ -217,11 +217,14 @@ func newRootCommand() *cobra.Command {
...
-				err = mgr.Start(signals.SetupSignalHandler())
+				stopCh := signals.SetupSignalHandler()
+				err = mgr.Start(stopCh)
 				if err != nil {
-					log.Fatal(err)
+					log.Error(err)
 				}
-				// Canceling the leader election context
+				// Stop controllers
+				stopCh <- struct{}{}
```
staebler, line 226: This won't do anything. At this point, the controllers have already stopped. We need to send to the stop channel when the TERM signal is received in order to stop the controllers.

pkg/signals/signal.go
```diff
@@ -0,0 +1,31 @@
...
+
+var (
+	onlyOneSignalHandler = make(chan struct{})
+	shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
+)
+
+// SetupSignalHandler registers for SIGTERM and SIGINT. A stop channel is returned
+// which is closed on one of these signals. If a second signal is caught, the program
+// is terminated with exit code 1.
+func SetupSignalHandler() (stopCh chan struct{}) {
```
akhil-rane, line 17: Created a [similar](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/signals/signal.go) function locally that returns a stop channel that can send and receive.

cmd/manager/main.go
```diff
@@ -228,7 +239,6 @@ func newRootCommand() *cobra.Command {
 			// listen for interrupts or the Linux SIGTERM signal and cancel
 			// our context, which the leader election code will observe and
 			// step down
-			ch := make(chan os.Signal, 1)
 			signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
```
staebler, line 242: You don't want to do this part any more for the leader-election context. The TERM signal should only directly causing the controllers to stop and not also directly causing the release of the leader-election lock.

cmd/manager/main.go
akhil-rane: That was already happening with the previous implementation I suppose. Look at the code [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/signals/signal.go)

akhil-rane: I think this is taken care of by the original implementation of the `signals.SetupSignalHandler`

staebler: Yeah, you are right. I steered you wrong here telling you to replace that call.

staebler: The part that I was missing was that the either (1) the signals should not directly cause the release of the lock or (2) the release of the lock should block on the manager stopping.

cmd/manager/main.go
```diff
@@ -214,7 +215,12 @@ func newRootCommand() *cobra.Command {
 				log.Info("Starting the Cmd.")
 
 				// Start the Cmd
-				log.Fatal(mgr.Start(signals.SetupSignalHandler()))
+				err = mgr.Start(signals.SetupSignalHandler())
+				if err != nil {
+					log.Error(err)
```
staebler, line 220:
```suggestion
					log.WithError(err).Error("error running manager")
```

cmd/manager/main.go
```diff
@@ -214,7 +215,12 @@ func newRootCommand() *cobra.Command {
...
+				}
+				// Canceling the leader election context
+				ch <- os.Interrupt
```
staebler, line 223: We don't need a channel to do this any more. Just cancel the leader-election context directly.

staebler (commented):
Nice! I think we have the right logic now. Thanks for working through this.

cmd/manager/main.go
```diff
@@ -159,7 +157,7 @@ func newRootCommand() *cobra.Command {
 			log.Info("Starting /healthz and /readyz endpoints")
 			go http.ListenAndServe(":8080", nil)
 
-			run := func(ctx context.Context) {
+			run := func(ctx context.Context, cancel context.CancelFunc) {
```
staebler, line 160: The cancel function doesn't need to be passed in as a parameter. Create the context for the leader election prior to creating this function.

akhil-rane: addressed

PR #1160, opened 2020-09-30
Stop reporting SelectorSyncSet metrics after initial sync success.

dgoodwin:
Some SelectorSyncSet time to first apply metrics were abnormally large
as labels can be added to clusters long after they're installed,
resulting in a time to apply metric of install time -> now, which could
be days weeks or months.

Instead use the new "FirstSuccessTimestamp" on ClusterSync.Status. If we
see a first success for a SelectorSyncSet, but the FirstSuccessTimestamp
overall is set, we know this was a new match and thus we don't really
have an accurate start time, and the metric should be skipped.

/assign @abutcher 
/cc @staebler

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -67,7 +67,7 @@ var (
 	metricTimeToApplySelectorSyncSet = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name:    "hive_selectorsyncset_apply_duration_seconds",
-			Help:    "Time to first successfully apply selector syncset to a cluster",
+			Help:    "Time to first successfully apply each SelectorSyncSet to a new cluster. Does not include results when cluster labels change post-install and new syncsets match.",
 			Buckets: []float64{5, 10, 30, 60, 300, 600, 1800, 3600},
 		},
 		[]string{"name"},
@@ -352,6 +352,7 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		syncSets,
 		clusterSync.Status.SyncSets,
 		needToDoFullReapply,
+		false, // no need to report SelectorSyncSet metrics if we're reconciling non-selector SyncSets
 		resourceHelper,
 		logger,
 	)
@@ -364,6 +365,7 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		selectorSyncSets,
 		clusterSync.Status.SelectorSyncSets,
 		needToDoFullReapply,
+		clusterSync.Status.FirstSuccessTime == nil, // only report SelectorSyncSet metrics if we haven't reached first success
 		resourceHelper,
 		logger,
 	)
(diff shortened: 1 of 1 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -517,10 +520,23 @@ func (r *ReconcileClusterSync) applySyncSets(
...
+				if reportSelectorSyncSetMetrics {
+					// Report SelectorSyncSet metric *only if* we have not yet reached our "everything has applied successfully for the first time"
+					// state. This is to handle situations where a clusters labels change long after it was installed, resulting
+					// in a new SelectorSyncSet matching, and a metric showing days/weeks/months time to first apply. In this scenario
+					// we have no idea when the label was added, and thus it is not currently possible to report the time from
+					// label added to SyncSet successfully applied.
+					logger.WithField("applyTime", applyTime).Debug("observed first successful apply of SelectorSyncSet for cluster")
+					metricTimeToApplySelectorSyncSet.WithLabelValues(syncSet.AsMetaObject().GetName()).Observe(applyTime)
+				} else {
+					log.WithField("SelectorSyncSet", syncSet.AsMetaObject().GetName()).
```
staebler, line 534: This logger field is already set [here](https://github.com/openshift/hive/blob/4bd6d19e745bcdea3a59bdf85756a17f6c53ac89/pkg/controller/clustersync/clustersync_controller.go#L435)

staebler (commented):
Looks fine to me.

I would like to see the help of the `hive_selectorsyncset_apply_duration_seconds` metric changed to indicate that it only applies to SelectorSyncSets matched to the cluster at install time.

This metric may become less and less useful over time if there are more SelectorSyncSets that are applied based on cluster version, since the version label is not added to the ClusterDeployment until after installation.

pkg/controller/clustersync/clustersync_controller.go
dgoodwin: Also just noticed I used log instead of logger. Oops.

dgoodwin:
Updated.

staebler (commented):
Looks good to me. I'll leave it up to @abutcher to give it a lgtm after he gets a chance to look over it.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -517,10 +520,22 @@ func (r *ReconcileClusterSync) applySyncSets(
...
+				if reportSelectorSyncSetMetrics {
+					// Report SelectorSyncSet metric *only if* we have not yet reached our "everything has applied successfully for the first time"
+					// state. This is to handle situations where a clusters labels change long after it was installed, resulting
+					// in a new SelectorSyncSet matching, and a metric showing days/weeks/months time to first apply. In this scenario
+					// we have no idea when the label was added, and thus it is not currently possible to report the time from
+					// label added to SyncSet successfully applied.
+					logger.WithField("applyTime", applyTime).Debug("observed first successful apply of SelectorSyncSet for cluster")
+					metricTimeToApplySelectorSyncSet.WithLabelValues(syncSet.AsMetaObject().GetName()).Observe(applyTime)
+				} else {
+					logger.Info("skipped observing first successful apply of SelectorSyncSet metric because ClusterSync has a FirstSuccessTimestamp")
```
abutcher, line 534: Nit for field name `FirstSuccessTime`

```suggestion
					logger.Info("skipped observing first successful apply of SelectorSyncSet metric because ClusterSync has a FirstSuccessTime")
```

dgoodwin:
Updated, thx!

PR #1161, opened 2020-10-05
Add s3 install logs config to hiveconfig

twiest:
- [x] Add s3 install logs config to hiveconfig
- [x] unit tests
- [x] manual tests


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -761,6 +773,73 @@ func (r *ReconcileClusterDeployment) startNewProvision(
...
+	if cloudProvider == constants.InstallLogsUploadProviderAWS {
+		secretName, foundSrc := os.LookupEnv(constants.InstallLogsAWSCredentialsSecretRefEnvVar)
+		if foundSrc {
+			extraEnvVars = append(extraEnvVars, corev1.EnvVar{
+				Name:  constants.InstallLogsAWSCredentialsSecretRefEnvVar,
+				Value: secretPrefix + "-" + secretName,
+			})
+		}
+
+		extraEnvVars = addEnvVarIfFound(constants.InstallLogsAWSRegionEnvVar, extraEnvVars)
```
joelddiaz, line 821: Would it make adding future providers cleaner if we stored the `installLogsUploadProvider` in the `extraEnvVars` (so that we can `switch` on the value of the provider?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2321,6 +2326,75 @@ func TestMergePullSecrets(t *testing.T) {
...
+			mockRemoteClientBuilder := remoteclientmock.NewMockBuilder(mockCtrl)
+			rcd := &ReconcileClusterDeployment{
+				Client:                        fakeClient,
+				scheme:                        scheme.Scheme,
+				logger:                        log.WithField("controller", "clusterDeployment"),
+				remoteClusterAPIClientBuilder: func(*hivev1.ClusterDeployment) remoteclient.Builder { return mockRemoteClientBuilder },
+			}
+
+			for _, envVar := range test.existingEnvVars {
+				os.Setenv(envVar.Name, envVar.Value)
```
joelddiaz, line 2378: Unlikely, but Setenv can return an error.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2321,6 +2326,75 @@ func TestMergePullSecrets(t *testing.T) {
...
+
+				defer os.Unsetenv(envVar.Name)
```
joelddiaz, line 2380: Same here. An error on Unsetenv leaves further test cases in an unknown state.

pkg/install/generate_test.go
```diff
@@ -72,11 +73,22 @@ func TestInstallerPodSpec(t *testing.T) {
...
+				},
+			},
 			validate: func(t *testing.T, actualPodSpec *corev1.PodSpec, actualError error) {
 				expectedPodMemoryRequest := resource.MustParse("800Mi")
 				actualPodMemoryRequest := actualPodSpec.Containers[2].Resources.Requests[corev1.ResourceMemory]
 
 				assert.Equal(t, expectedPodMemoryRequest, actualPodMemoryRequest, "Incorrect pod memory request")
+
+				for _, container := range actualPodSpec.Containers {
+					assert.Equal(t, "TESTVAR", container.Env[1].Name, "Environment variable name not passed to pod spec")
```
joelddiaz, line 89: Why the hardcoded `1` index. Is something like `assert.Contains(t, container.Env, corev1.EnvVar{Name / Value})` usable?

joelddiaz (commented):
I like the new AWS credentials format, but would like a second set of eyes before we commit to this for the long-term.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
twiest: Done.

twiest:
@joelddiaz I believe I've addressed your feedback. PTAL.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2321,6 +2326,81 @@ func TestMergePullSecrets(t *testing.T) {
...
+						}
+					}()
+				} else {
+					t.Error(err)
+				}
+			}
+
+			err := rcd.copyInstallLogSecret(testNamespace, test.existingEnvVars)
+			secretList := &corev1.SecretList{}
+			rcd.List(context.TODO(), secretList, client.InNamespace(testNamespace))
```
joelddiaz, line 2391: List() can return an error.

pkg/constants/constants.go
```diff
@@ -275,6 +275,24 @@ const (
...
 	AWSCredsMount = "/etc/aws-creds"
+
+	// InstallLogsUploadProviderEnvVar is used to specify which object store provider is being used.
+	InstallLogsUploadProviderEnvVar = "HIVE_INSTALL_LOGS_UPLOAD_PROVIDER"
+
+	// InstallLogsUploadProviderAWS is used to specify that AWS is the cloud provider to upload logs to.
+	InstallLogsUploadProviderAWS = "aws"
+
+	// InstallLogsAWSCredentialsSecretRefEnvVar is the environment variable specifying what secret to use for storing logs.
+	InstallLogsAWSCredentialsSecretRefEnvVar = "HIVE_INSTALL_LOGS_AWS_CREDENTIALS_SECRET"
```
joelddiaz, line 286: Does this need to be a cloud-specific env var? Could it be `HIVE_INSTALL_LOGS_CREDENTIALS_SECRET` and knowing that `HIVE_INSTALL_LOGS_UPLOAD_PROVIDER` is `aws` already tells what credentials format will be in the secret?

twiest:
@joelddiaz I believe I've addressed your latest feedback. PTAL.

joelddiaz (commented):
/lgtm

PR #1162, opened 2020-10-05
vendor: replace k8s.io/apiserver with fork that fixes hiveadmission delay

staebler:
Disable the observers of the watermark in the max-in-flight filter. The observations take a long time to process for the first request after a long period with no requests.

This is a temporary hack fix until the issue is resolved upstream.

See https://github.com/kubernetes/kubernetes/issues/95300

https://issues.redhat.com/browse/CO-1200


dgoodwin:
All looks good to me, over to you @twiest

PR #1165, opened 2020-10-07
Add option to limit install attempts

akhil-rane:
This PR adds an option to limit the number of install restarts by setting a
value in the ClusterDeployment field `limitInstallRestarts`. Once this limit has
reached, `ProvisionStopped` error condition will be set in the status.

### DIFF
--- a/pkg/apis/hive/v1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1/clusterdeployment_types.go
@@ -128,6 +128,10 @@ type ClusterDeploymentSpec struct {
 	// time since the cluster last came out of hibernation.
 	// +optional
 	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
+
+	// InstallAttemptsLimit is the maximum number of times Hive will attempt to install the cluster.
+	// +optional
+	InstallAttemptsLimit *int32 `json:"installAttemptsLimit,omitempty"`
 }
 
 // Provisioning contains settings used only for initial cluster provisioning.
@@ -313,6 +317,9 @@ const (
 
 	// DeprovisionLaunchErrorCondition is set when a cluster deprovision fails to launch.
 	DeprovisionLaunchErrorCondition ClusterDeploymentConditionType = "DeprovisionLaunchError"
+
+	// ProvisionStoppedCondition is set when cluster provisioning is stopped
+	ProvisionStoppedCondition ClusterDeploymentConditionType = "ProvisionStopped"
 )
 
 // AllClusterDeploymentConditions is a slice containing all condition types. This can be used for dealing with
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -640,8 +640,60 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 	if cd.Status.ProvisionRef == nil {
 		if cd.Status.InstallRestarts > 0 && cd.Annotations[tryInstallOnceAnnotation] == "true" {
 			cdLog.Debug("not creating new provision since the deployment is set to try install only once")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.ProvisionStoppedCondition,
+				corev1.ConditionTrue,
+				"InstallOnlyOnceSet",
+				"Deployment is set to try install only once",
+				controllerutils.UpdateConditionIfReasonOrMessageChange)
+			if changed {
+				cd.Status.Conditions = conditions
+				cdLog.Debugf("setting ProvisionStoppedCondition to %v", corev1.ConditionTrue)
+				if err := r.Status().Update(context.TODO(), cd); err != nil {
+					cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+					return reconcile.Result{}, err
+				}
+			}
 			return reconcile.Result{}, nil
 		}
+		if cd.Spec.InstallAttemptsLimit != nil && cd.Status.InstallRestarts >= int(*cd.Spec.InstallAttemptsLimit) {
+			cdLog.Debug("not creating new provision since the install attempts limit has been reached")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.ProvisionStoppedCondition,
+				corev1.ConditionTrue,
+				"InstallAttemptsLimitReached",
+				"Install attempts limit reached",
+				controllerutils.UpdateConditionIfReasonOrMessageChange)
+			if changed {
+				cd.Status.Conditions = conditions
+				cdLog.Debugf("setting ProvisionStoppedCondition to %v", corev1.ConditionTrue)
+				if err := r.Status().Update(context.TODO(), cd); err != nil {
+					cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+					return reconcile.Result{}, err
+				}
+			}
+			return reconcile.Result{}, nil
+		}
+
+		conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+			cd.Status.Conditions,
+			hivev1.ProvisionStoppedCondition,
+			corev1.ConditionFalse,
+			"ProvisionNotStopped",
+			"Provision is not stopped",
+			controllerutils.UpdateConditionNever)
+		if changed {
+			cd.Status.Conditions = conditions
+			cdLog.Debugf("setting ProvisionStoppedCondition to %v", corev1.ConditionFalse)
+			if err := r.Status().Update(context.TODO(), cd); err != nil {
+				cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
+		}
+
 		return r.startNewProvision(cd, releaseImage, cdLog)
 	}
 
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -1443,6 +1443,63 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				assertConditionReason(t, cd, hivev1.InstallLaunchErrorCondition, "PodInPendingPhase")
 			},
 		},
+		{
+			name: "install attempts is less than the limit",
+			existing: []runtime.Object{
+				func() runtime.Object {
+					cd := testClusterDeployment()
+					cd.Status.InstallRestarts = 1
+					cd.Spec.InstallAttemptsLimit = pointer.Int32Ptr(2)
+					return cd
+				}(),
+				testSecret(corev1.SecretTypeDockerConfigJson, pullSecretSecret, corev1.DockerConfigJsonKey, "{}"),
+				testSecret(corev1.SecretTypeDockerConfigJson, constants.GetMergedPullSecretName(testClusterDeployment()), corev1.DockerConfigJsonKey, "{}"),
+			},
+			expectPendingCreation: true,
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				require.NotNil(t, cd, "could not get ClusterDeployment")
+				assert.NotContains(t, cd.Status.Conditions, hivev1.ProvisionStoppedCondition, "ClusterDeployment should not have ProvisionStopped condition")
+			},
+		},
+		{
+			name: "install attempts is equal to the limit",
+			existing: []runtime.Object{
+				func() runtime.Object {
+					cd := testClusterDeployment()
+					cd.Status.InstallRestarts = 2
+					cd.Spec.InstallAttemptsLimit = pointer.Int32Ptr(2)
+					return cd
+				}(),
+				testSecret(corev1.SecretTypeDockerConfigJson, pullSecretSecret, corev1.DockerConfigJsonKey, "{}"),
+				testSecret(corev1.SecretTypeDockerConfigJson, constants.GetMergedPullSecretName(testClusterDeployment()), corev1.DockerConfigJsonKey, "{}"),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				require.NotNil(t, cd, "could not get ClusterDeployment")
+				assertConditionStatus(t, cd, hivev1.ProvisionStoppedCondition, corev1.ConditionTrue)
+				assertConditionReason(t, cd, hivev1.ProvisionStoppedCondition, "InstallAttemptsLimitReached")
+			},
+		},
+		{
+			name: "install attempts is greater than the limit",
+			existing: []runtime.Object{
+				func() runtime.Object {
+					cd := testClusterDeployment()
+					cd.Status.InstallRestarts = 3
+					cd.Spec.InstallAttemptsLimit = pointer.Int32Ptr(2)
+					return cd
+				}(),
+				testSecret(corev1.SecretTypeDockerConfigJson, pullSecretSecret, corev1.DockerConfigJsonKey, "{}"),
+				testSecret(corev1.SecretTypeDockerConfigJson, constants.GetMergedPullSecretName(testClusterDeployment()), corev1.DockerConfigJsonKey, "{}"),
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cd := getCD(c)
+				require.NotNil(t, cd, "could not get ClusterDeployment")
+				assertConditionStatus(t, cd, hivev1.ProvisionStoppedCondition, corev1.ConditionTrue)
+				assertConditionReason(t, cd, hivev1.ProvisionStoppedCondition, "InstallAttemptsLimitReached")
+			},
+		},
 	}
 
 	for _, test := range tests {
--- a/config/crds/hive.openshift.io_clusterdeployments.yaml
+++ b/config/crds/hive.openshift.io_clusterdeployments.yaml
@@ -327,6 +327,11 @@ spec:
                 - name
                 type: object
               type: array
+            installAttemptsLimit:
+              description: InstallAttemptsLimit is the maximum number of times Hive
+                will attempt to install the cluster.
+              format: int32
+              type: integer
             installed:
               description: Installed is true if the cluster has been installed
               type: boolean


pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -128,6 +128,10 @@ type ClusterDeploymentSpec struct {
 	// time since the cluster last came out of hibernation.
 	// +optional
 	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
+
+	// LimitInstallRestarts is the number of times Hive attempts to restart an installation. If the last attempt
+	// fails ClusterDeployment will show an error in the status
+	LimitInstallRestarts *int `json:"limitInstallRestarts,omitempty"`
```
staebler, line 134: Make this a noun instead of a verb.
```suggestion
	InstallRestartLimit *int `json:"installRestartLimit,omitempty"`
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -642,6 +642,44 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			cdLog.Debug("not creating new provision since the deployment is set to try install only once")
 			return reconcile.Result{}, nil
 		}
+
+		if cd.Spec.LimitInstallRestarts != nil && cd.Status.InstallRestarts > *cd.Spec.LimitInstallRestarts {
+			cdLog.Debug("not creating new provision since the install restarts limit has reached")
```
staebler, line 647:
```suggestion
			cdLog.Debug("not creating new provision since the install restarts limit has been reached")
```

pkg/apis/hive/v1/clusterdeployment_types.go
staebler, line 134: The size of the integer must be specified.
```suggestion
	LimitInstallRestarts *int32 `json:"limitInstallRestarts,omitempty"`
```

staebler, line 134: Add `+optional`.
```suggestion
	// +optional
	LimitInstallRestarts *int `json:"limitInstallRestarts,omitempty"`
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -642,6 +642,44 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			cdLog.Debug("not creating new provision since the deployment is set to try install only once")
 			return reconcile.Result{}, nil
 		}
+
+		if cd.Spec.LimitInstallRestarts != nil && cd.Status.InstallRestarts > *cd.Spec.LimitInstallRestarts {
```
staebler, line 646: I think that this should be greater than **or equal to**. The `InstallRestarts` field is the number of restarts that have already been done.
```suggestion
		if cd.Spec.LimitInstallRestarts != nil && cd.Status.InstallRestarts >= *cd.Spec.LimitInstallRestarts {
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -642,6 +642,44 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+			cdLog.Debug("not creating new provision since the install restarts limit has reached")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.ProvisionStoppedCondition,
+				corev1.ConditionTrue,
+				"InstallRestartsLimitReached",
+				fmt.Sprintf("Install restarts limit reached. Current limit is %v.", cd.Spec.LimitInstallRestarts),
```
staebler, line 653: This message will include the address of the `LimitInstallRestarts` pointer and not the maximum number of restarts. Also, I don't know that it is necessary to include the limit in the condition message, given that the limit is directly available in the cd spec.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1482,6 +1482,26 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
 			},
 		},
+		{
+			name: "set ProvisionStoppedCondition when install restarts limit has reached",
+			existing: []runtime.Object{
+				func() runtime.Object {
+					cd := testClusterDeployment()
+					cd.Status.InstallRestarts = 3
+					cd.Spec.LimitInstallRestarts = new(int)
+					*cd.Spec.LimitInstallRestarts = 2
```
staebler, line 1492: Assuming that `LimitInstallRestarts` is changed to be an `*int32`,
```suggestion
					cd.Spec.LimitInstallRestarts = pointer.Int32Ptr(2)
```

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1482,6 +1482,26 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				assert.Equal(t, "manual-resourcegroup", deprovision.Spec.Platform.Azure.ResourceGroupName)
 			},
 		},
+		{
+			name: "set ProvisionStoppedCondition when install restarts limit has reached",
```
staebler, line 1486: We should have one test where `InstallRestarts` is less than the limit, one test where `InstallRestarts` is equal to the limit, and one test where `InstallRestarts` is greater than the limit.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
akhil-rane: `The InstallRestarts field is the number of restarts that have already been done.` This statement is not true. We first increment InstallRestarts when previous provision fails and then go ahead and create new Provision.

pkg/apis/hive/v1/clusterdeployment_types.go
akhil-rane: I kept it `*int` because type of InstallRestarts field in the ClusterDeployment is `*int`. If we change that will we have any impact on existing CRs?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -981,6 +939,48 @@ func (r *ReconcileClusterDeployment) reconcileCompletedProvision(cd *hivev1.Clus
 }
 
 func (r *ReconcileClusterDeployment) clearOutCurrentProvision(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (reconcile.Result, error) {
+	// clear out current provision only if install restart is required
+	if cd.Status.InstallRestarts == 0 && cd.Annotations[tryInstallOnceAnnotation] == "true" {
+		cdLog.Debug("not creating new provision since the deployment is set to try install only once")
+		return reconcile.Result{}, nil
+	}
```
akhil-rane, line 946: @staebler Do we need this code now? We can do the same thing by setting `InstallRestartsLimit` to 0

gregsheremeta:
We were planning to use this feature to prevent any install from ever starting, for scale test purposes.
https://issues.redhat.com/browse/CO-1184?focusedCommentId=15266161&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15266161

`InstallRestartLimit` implies the install will run at least once. Can you clarify?

akhil-rane:
> We were planning to use this feature to prevent any install from ever starting, for scale test purposes.
> https://issues.redhat.com/browse/CO-1184?focusedCommentId=15266161&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15266161
> 
> `InstallRestartLimit` implies the install will run at least once. Can you clarify?

Yes you right, `InstallRestartLimit` implies the install will run at least once.

pkg/apis/hive/v1/clusterdeployment_types.go
staebler: You are right that we cannot change the existing `InstallRestarts` type. However, that is not a good enough reason to use the improper `*int` here as well.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -943,6 +939,48 @@ func (r *ReconcileClusterDeployment) reconcileCompletedProvision(cd *hivev1.Clus
...
+	if cd.Spec.InstallRestartLimit != nil && cd.Status.InstallRestarts >= *cd.Spec.InstallRestartLimit {
+		cdLog.Debug("not creating new provision since the install restarts limit has been reached")
+		conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+			cd.Status.Conditions,
+			hivev1.ProvisionStoppedCondition,
+			corev1.ConditionTrue,
+			"InstallRestartsLimitReached",
+			"Install restarts limit reached",
```
staebler, line 954:
```suggestion
			"InstallRestartLimitReached",
			"Install restart limit reached",
```
or change the name of the field to `InstallRestartsLimit`.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
staebler: Yes, we still need the code. We cannot stop supporting the annotation, even though there is an easier way to do it now.

staebler:
> > We were planning to use this feature to prevent any install from ever starting, for scale test purposes.
> > https://issues.redhat.com/browse/CO-1184?focusedCommentId=15266161&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15266161
> > `InstallRestartLimit` implies the install will run at least once. Can you clarify?
> 
> Yes you right, `InstallRestartLimit` implies the install will run at least once.

Can we change `InstallRestartLimit` to `InstallAttemptsLimit` so that we can support 0 attempts?

akhil-rane:
> > > We were planning to use this feature to prevent any install from ever starting, for scale test purposes.
> > > https://issues.redhat.com/browse/CO-1184?focusedCommentId=15266161&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15266161
> > > `InstallRestartLimit` implies the install will run at least once. Can you clarify?
> > 
> > 
> > Yes you right, `InstallRestartLimit` implies the install will run at least once.
> 
> Can we change `InstallRestartLimit` to `InstallAttemptsLimit` so that we can support 0 attempts?

There is one problem here, our attempts start from 0.

akhil-rane:
@gregsheremeta can we go for InstallAttemptsLimit here? or do we need separate InstallRestartsLimit and install-skip annotation features?

staebler:
> > > > We were planning to use this feature to prevent any install from ever starting, for scale test purposes.
> > > > https://issues.redhat.com/browse/CO-1184?focusedCommentId=15266161&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15266161
> > > > `InstallRestartLimit` implies the install will run at least once. Can you clarify?
> > > 
> > > 
> > > Yes you right, `InstallRestartLimit` implies the install will run at least once.
> > 
> > 
> > Can we change `InstallRestartLimit` to `InstallAttemptsLimit` so that we can support 0 attempts?
> 
> There is one problem here, our attempts start from 0.

That is not a problem.
```
if InstallAttempts >= InstallAttemptsLimit {
   // do not start a new provision
}
```

dgoodwin:
+1 for InstallAttemptsLimit and killing two birds with one stone.

akhil-rane:
PR updated

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -128,6 +128,11 @@ type ClusterDeploymentSpec struct {
 	// time since the cluster last came out of hibernation.
 	// +optional
 	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
+
+	// InstallAttemptsLimit is the number of times Hive attempts an installation. If the last attempt
+	// fails ClusterDeployment will show an error in the status
```
staebler, line 133:
```suggestion
	// fails ClusterDeployment will show an error in the status.
```

staebler, line 133: The second sentence of this description does not add very much. I suggest that it either be removed or expanded to describe the type of error and where in the status the error would appear. Is the error different than the normal error that would show up if `InstallAttemptsLimit` were omitted?

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -128,6 +128,11 @@ type ClusterDeploymentSpec struct {
 	// time since the cluster last came out of hibernation.
 	// +optional
 	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
+
+	// InstallAttemptsLimit is the number of times Hive attempts an installation. If the last attempt
```
staebler, line 132:
```suggestion
	// InstallAttemptsLimit is the maximum number of times Hive will attempt to install the cluster. If the last attempt
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -642,6 +642,43 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			cdLog.Debug("not creating new provision since the deployment is set to try install only once")
 			return reconcile.Result{}, nil
 		}
+		if cd.Spec.InstallAttemptsLimit != nil && cd.Status.InstallRestarts >= int(*cd.Spec.InstallAttemptsLimit) {
+			cdLog.Debug("not creating new provision since the install attempts limit has been reached")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.ProvisionStoppedCondition,
```
staebler, line 649: We should set this condition when provisioning is stopped due to the try-once annotation as well.

staebler (commented):
This looks good. I have couple a couple of minor notes.

pkg/apis/hive/v1/clusterdeployment_types.go
akhil-rane: Keeping it short sounds better. Removing the second sentence.

akhil-rane: > Is the error different than the normal error that would show up if `InstallAttemptsLimit` were omitted?

No

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -128,6 +128,10 @@ type ClusterDeploymentSpec struct {
 	// time since the cluster last came out of hibernation.
 	// +optional
 	HibernateAfter *metav1.Duration `json:"hibernateAfter,omitempty"`
+
+	// InstallAttemptsLimit is the maximum number of times Hive will attempt to install the cluster.
+	// +optional
+	InstallAttemptsLimit *int32 `json:"installAttemptsLimit,omitempty"`
```
abhinavdahiya, line 134: is `0` is valid value? I don't think *int32 is required here, using int32 and using >0 for when this is used
- keeps the API backwards compatible in since 0 will be the default value of decoded int32
- keeps the API much simple as users don't have to capture nil

staebler: `0` **is** a valid value. It means that the user does not want to attempt any installs. A nil value means that the user does not want to impose any limit on the number of install attempts.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -313,6 +317,9 @@ const (
 
 	// DeprovisionLaunchErrorCondition is set when a cluster deprovision fails to launch.
 	DeprovisionLaunchErrorCondition ClusterDeploymentConditionType = "DeprovisionLaunchError"
+
+	// ProvisionStoppedCondition is set when cluster provisioning is stopped
+	ProvisionStoppedCondition ClusterDeploymentConditionType = "ProvisionStopped"
```
abhinavdahiya, line 322: personally i do not think `ProvisionStopped` is a very user friendly condition name, because it does very little to explain why the provision was stopped.
- was it because some pre requisites were failed
- was it because we know retrying the install will not succeed, like when we know the creds or limit has caused the failure
- was it just because the retry budget was completed.

I would recommend we maybe start with some thing like `MaxRetryLimitReachedForProvisioning` maybe?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -640,8 +640,60 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+				cdLog.Debugf("setting ProvisionStoppedCondition to %v", corev1.ConditionTrue)
+				if err := r.Status().Update(context.TODO(), cd); err != nil {
+					cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+					return reconcile.Result{}, err
+				}
+			}
+			return reconcile.Result{}, nil
+		}
+		if cd.Spec.InstallAttemptsLimit != nil && cd.Status.InstallRestarts >= int(*cd.Spec.InstallAttemptsLimit) {
+			cdLog.Debug("not creating new provision since the install attempts limit has been reached")
```
abhinavdahiya, line 661: will including the attempt index and the max retry count in the log help in triage?

pkg/apis/hive/v1/clusterdeployment_types.go
abhinavdahiya: based on https://github.com/openshift/hive/pull/1165/files#diff-5fc238b39e2a1fc17bd412fbac2d14eab4e0f205c91886b5671611bec4c665ccR645 and https://github.com/openshift/hive/pull/1165/files#diff-5fc238b39e2a1fc17bd412fbac2d14eab4e0f205c91886b5671611bec4c665ccR664

I can see this condition name being good enough.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -640,8 +640,60 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.ProvisionStoppedCondition,
+				corev1.ConditionTrue,
+				"InstallAttemptsLimitReached",
+				"Install attempts limit reached",
```
abhinavdahiya, line 667: not sure what this value represents, but if this is the message of the condition, i think it makes sense to include for info for the user like `"Install attempts limit of %d reached", cd.Spec.InstallAttemptsLimit` which helps them figure out why the limit was reached.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
akhil-rane: This information is already available in the ClusterDeployment (cd.spec.InstallAttemptsLimit). I do not think it will add anything useful.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
akhil-rane: This was removed after previous review feedback as we thought exact limit is already available in `cd.Spec`.

pkg/apis/hive/v1/clusterdeployment_types.go
abhinavdahiya: The code is comparing this value to cd.Status.InstallRestarts , and restart == 0 doesn't mean there was no attempt (if i go by the name). So the handling should be updated to make sure the controller never starts provisioning..

pkg/controller/clusterdeployment/clusterdeployment_controller.go
abhinavdahiya: So conditions can and are used by clients outside the context of the complete definition of the cluster deployment esp error conditions,

for example, when the cloud providers return errors for quota exceeded, even though the quota is available to be looked at in the console, the error message usually goes something like `requested 8 vCPUs but only 4 available in the region` . So personally I think the messages for these conditions should be useful/actionable for the users.

but i'll defer to previous review.

pkg/apis/hive/v1/clusterdeployment_types.go
akhil-rane: restart == 0 does mean there was no attempt before. We increment installRestarts after previous failed attempt and before starting a new provision.

abhinavdahiya: > restart == 0 does mean there was no attempt before. We increment installRestarts after previous failed attempt and before starting a new provision.

that's not what `InstallRestarts` name means, but as long as it does what you say it does.

akhil-rane:
@staebler changes look good to you?

staebler (approved):
/lgtm

PR #1172, opened 2020-10-12
handle switch from unsupported to supported hibernation capability

joelddiaz:
Previously the hibernation controller would mark a ClusterDeployment with a condition indicating that hibernation was not supported (b/c the clusterversion had not yet become available). It would never clear the Unsupported condition once the ClusterDeployment was deemed to be hibernation-capable.

Detect and update the condition for this case.

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -164,6 +164,13 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (result rec
 		}
 	}
 
+	// Clear any lingering unsupported hibernation condition
+	if hibernatingCondition != nil && hibernatingCondition.Reason == hivev1.UnsupportedHibernationReason {
+		if supported, msg := r.canHibernate(cd); supported {
+			return r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, msg, corev1.ConditionFalse, cdLog)
+		}
+	}
+
 	// Check if HibernateAfter is set, and if the cluster has been in running state for longer than this duration, put it to sleep.
 	if cd.Spec.HibernateAfter != nil && cd.Spec.PowerState != hivev1.HibernatingClusterPowerState {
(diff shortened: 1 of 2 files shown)


csrwng:
@joelddiaz seems fine to me, but wouldn't we just start hibernating if it's now supported? In which case the condition would be set to true and reason == "Stopping" or whatever it is when we start hibernating. Are you observing something different?

joelddiaz:
We don't necessarily need to be hibernating yet. Maybe the cluster shouldn't be hibernating yet, but we still want the condition to reflect that we can hibernate when the hibernateAfter condition is satisfied (or an explicit hibernation is triggered).

What we have before this PR is that the cluster version isn't available so we set Unsupported. Then the cluster version becomes available but it isn't time to hibernate, and it stays Unsupported until a hibernation path is taken.

csrwng:
Got it. So maybe the code to remove the condition shouldn't be inside the block of `if shouldHibernate || hibernate-after is set`. We should check whether the condition is set outside of that condition and remove it if it's no longer true.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -172,6 +172,11 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (result rec
 		}
 	}
 
+	// Clear any lingering unsupported hibernation condition
+	if hibernatingCondition != nil && hibernatingCondition.Reason == hivev1.UnsupportedHibernationReason {
+		return r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, "Hibernation capable", corev1.ConditionFalse, cdLog)
```
csrwng, line 177: we should only remove it if we determine that it is now supported

joelddiaz:
@csrwng are we good with the most recent update?

csrwng:
Sorry, yes
/lgtm

joelddiaz:
/hold
until we figure out where to pull our builds from

joelddiaz:
@csrwng i just rebased on master to get the docker image fixes. no other changes

PR #1173, opened 2020-10-12
allow skipping leader election

joelddiaz:
Allow skiping leader election to add some flexibility when developing on hive where we can simultaneously have hive running in a cluster (potentially with some controllers disabled), and running hive locally.


akhil-rane (commented):
/lgtm

PR #1174, opened 2020-10-13
Revert "Merge pull request #1154 from joelddiaz/azure-resource-groups"

joelddiaz:
This reverts commit 34d28c1d6e001687fae2d0fc3ac722fe173f0c8c, reversing
changes made to 54442c3959748609d63aeacdba010b86a97e7a5e.

The bring-your-own-resource-group introduces a behavior when we hit an error during installation that causes the pre-made resource group to be deleted (that's how the uninstall process works). So any future installation attempt will fail b/c that resource group no longer exists.

This behavior doesn't map to the kind of retry loops that are implemented in hive, so we will not be supporting this BYO-resource-group.

### DIFF
--- a/config/crds/hive.openshift.io_clusterdeployments.yaml
+++ b/config/crds/hive.openshift.io_clusterdeployments.yaml
@@ -385,13 +385,6 @@ spec:
                       description: Region specifies the Azure region where the cluster
                         will be created.
                       type: string
-                    resourceGroupName:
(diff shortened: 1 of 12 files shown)


joelddiaz:
/hold
until we figure out where to pull the builder image from

joelddiaz:
@dgoodwin i rebased to get the docker hub fix

PR #1181, opened 2020-10-20
Add install failure details to provisioning underway metric

suhanime:
For a stuck provision, add the install failure details to the provision underway metric.

Relevant cluster deployment conditions are InstallFailing, ProvisionFailed and InstallLaunchError.
InstallFailing is not used anywhere currently and looks to be replaced by ProvisionFailed, hence it has been dropped.

/assign @dgoodwin

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -77,15 +77,6 @@ var (
 		Help: "Total number of SyncSetsInstances referencing non-selector SyncSets that have not successfully applied all resources/patches/secrets.",
 	})
 
-	// MetricClusterDeploymentProvisionUnderwaySeconds is a prometheus metric for the number of seconds
-	// between when a still provisioning cluster was created and now.
-	MetricClusterDeploymentProvisionUnderwaySeconds = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
-			Name: "hive_cluster_deployment_provision_underway_seconds",
-			Help: "Length of time a cluster has been provisioning.",
-		},
-		[]string{"cluster_deployment", "namespace", "cluster_type"},
-	)
 	// MetricClusterDeploymentDeprovisioningUnderwaySeconds is a prometheus metric for the number of seconds
 	// between when a still deprovisioning cluster was created and now.
 	MetricClusterDeploymentDeprovisioningUnderwaySeconds = prometheus.NewGaugeVec(
@@ -137,7 +128,6 @@ func init() {
 	metrics.Registry.MustRegister(metricSyncSetsUnappliedTotal)
 	metrics.Registry.MustRegister(metricControllerReconcileTime)
 
-	metrics.Registry.MustRegister(MetricClusterDeploymentProvisionUnderwaySeconds)
 	metrics.Registry.MustRegister(MetricClusterDeploymentDeprovisioningUnderwaySeconds)
 }
 
@@ -147,11 +137,11 @@ func Add(mgr manager.Manager) error {
 		Client:   mgr.GetClient(),
 		Interval: 2 * time.Minute,
 	}
+	metrics.Registry.MustRegister(newProvisioningUnderwayCollector(mgr.GetClient()))
 	err := mgr.Add(mc)
 	if err != nil {
 		return err
 	}
-
 	return nil
 }
 
@@ -193,18 +183,7 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 			for _, cd := range clusterDeployments.Items {
 				accumulator.processCluster(&cd)
 
-				if cd.DeletionTimestamp == nil {
-					if !cd.Spec.Installed {
-						// For installing clusters we report the seconds since
-						// cluster was created. clusterdeployment_controller should delete this for any cluster
-						// that is installed or deleted.
-						MetricClusterDeploymentProvisionUnderwaySeconds.WithLabelValues(
-							cd.Name,
-							cd.Namespace,
-							GetClusterDeploymentType(&cd)).Set(
-							time.Since(cd.CreationTimestamp.Time).Seconds())
-					}
-				} else {
+				if cd.DeletionTimestamp != nil {
 
 					// For deprovisioning clusters we report the seconds since
 					// cluster was deleted. clusterdeployment_controller should delete this
--- a/pkg/controller/metrics/provision_underway_collector.go
+++ b/pkg/controller/metrics/provision_underway_collector.go
@@ -1,0 +1,96 @@
+package metrics
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+)
+
+var (
+	// ClusterDeployment conditions that could indicate provisioning problems with a cluster
+	provisioningDelayCondition = [...]hivev1.ClusterDeploymentConditionType{
+		hivev1.ProvisionFailedCondition,
+		hivev1.InstallLaunchErrorCondition,
+	}
+)
+
+// provisioning underway metrics collected through a custom prometheus collector
+type provisioningUnderwayCollector struct {
+	client client.Client
+
+	// metricClusterDeploymentProvisionUnderwaySeconds is a prometheus metric for the number of seconds
+	// between when a still provisioning cluster was created and now.
+	metricClusterDeploymentProvisionUnderwaySeconds *prometheus.Desc
+}
+
+// collects the metrics for provisioningUnderwayCollector
+func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
+	ccLog := log.WithField("controller", "metrics")
+	ccLog.Info("calculating provisioning underway metrics across all ClusterDeployments")
+
+	// Load all ClusterDeployments so we can accumulate facts about them.
+	clusterDeployments := &hivev1.ClusterDeploymentList{}
+	err := cc.client.List(context.Background(), clusterDeployments)
+	if err != nil {
+		log.WithError(err).Error("error listing cluster deployments")
+		return
+	}
+	for _, cd := range clusterDeployments.Items {
+		if cd.DeletionTimestamp != nil {
+			continue
+		}
+		if cd.Spec.Installed {
+			continue
+		}
+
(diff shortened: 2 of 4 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -18,14 +18,90 @@ import (
...
+func (cc customCollector) Collect(ch chan<- prometheus.Metric) {
+	ccLog := log.WithField("controller", "metrics")
+	ccLog.Info("calculating metrics across all ClusterDeployments")
+	// Load all ClusterDeployments so we can accumulate facts about them.
+	clusterDeployments := &hivev1.ClusterDeploymentList{}
+	err := cc.client.List(context.Background(), clusterDeployments)
+	if err != nil {
+		log.WithError(err).Error("error listing cluster deployments")
+	} else {
+		accumulator, err := newClusterAccumulator(infinity, []string{"0h", "1h", "2h", "8h", "24h", "72h"})
```
dgoodwin, line 48: I don't think you need to include or use the accumulator in this collector, this is ok to leave handled by the normal metrics controller loop where it is also being run.

pkg/controller/metrics/metrics.go
```diff
@@ -18,14 +18,90 @@ import (
...
+		if err != nil {
+			ccLog.WithError(err).Error("unable to calculate metrics")
+			return
+		}
+		for _, cd := range clusterDeployments.Items {
```
dgoodwin, line 53: Your implementation looks a little different than in https://stackoverflow.com/a/57308144 where they maintain the data in an external map. Initially I was a little worried that this Collect function is called every time we're scraped, but I'm not sure that's the case, I actually can't tell how often this is called. I'm a little worried about iterating every ClusterDeployment, but otoh they should all be cached in memory anyhow, so I'm not sure how much of an issue it is even if this is called every 15 seconds. (which it doesn't seem to be)

Your implementation has a nice benefit where we don't need something external updating a shared map with the metrics data, locks etc. Metrics disappear as soon as the Collect method runs so delete is no longer an issue.

Can you track down when Collect will be called so we can understand it and make a final call? As it stands I like this and I hope we can stick with it as it is.

dgoodwin:
Can you clarify you worries about relocating clusters? I'm not super familiar what happens if you relocate a provisioning cluster, and also not sure why it's a concern for this implementation.

pkg/controller/metrics/metrics.go
```diff
@@ -18,14 +18,90 @@ import (
...
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/imageset"
 )
 
 const (
 	ControllerName = hivev1.MetricsControllerName
 )
 
+// metrics collected through custom prometheus collector
+type customCollector struct {
```
dgoodwin, line 30: lets rename to provisioningUnderwayCollector and godoc appropriately as well.

pkg/controller/metrics/metrics.go
```diff
@@ -124,16 +191,17 @@ func init() {
...
 }
 
 // Add creates a new metrics Calculator and adds it to the Manager.
 func Add(mgr manager.Manager) error {
 	mc := &Calculator{
-		Client:   mgr.GetClient(),
-		Interval: 2 * time.Minute,
+		Client:        mgr.GetClient(),
+		Interval:      2 * time.Minute,
+		CustomMetrics: newCustomCollector(mgr.GetClient()),
```
dgoodwin, line 202: Lets rename this var as well, and it can be private and not exported it looks like.

It's a little strange we carry this on the metrics controller when it isn't really used there, I would have expected an init block and a global package variable, but that is harder to get a client to.

pkg/controller/metrics/metrics.go
dgoodwin: also we should probably break this out into a separate provisionunderwaycollector.go module in this package.

dgoodwin (commented):
couple more comments.

pkg/controller/metrics/metrics.go
suhanime: I considered adding the collector to the manager as Matthew suggested, however manager can only add [Runnable](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager#Runnable), which would require prometheus [Collector](https://godoc.org/github.com/prometheus/client_golang/prometheus#Collector) to implement Runnable's Start method.
We have to instantiate and register the custom collector when the manager client is available, so Add method seems to be the appropriate place. I did remove it from the Calculator struct though, since we don't use the object itself anywhere else.

pkg/controller/metrics/metrics.go
suhanime: Changed.
FWIW my intention was to keep it abstract - so we can easily adapt it to add more metrics this way.

suhanime (commented):
Apologies for the confusion regarding cluster relocate in cluster deployment controller - it was simply another place we were logging the metric. Removed that code block since it is no longer needed

pkg/controller/metrics/metrics.go
staebler: The `Add` method should get the client and save it in the object that is being added. Then the `Start` method should be registering the metric.

staebler: But I am not strongly opposed to the approach that is being taken now. Registering the metrics in init blocks is already pervasive throughout the code, so adding to that anti-pattern is not a deal-breaker for me.

suhanime: > The `Add` method should get the client and save it in the object that is being added. Then the `Start` method should be registering the metric.

However, the custom collector is instantiating a prometheus.Collector... So the Add and Start methods need to have been implemented by prometheus.Collector.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -0,0 +1,87 @@
...
+	prometheus.DescribeByCollect(cc, ch)
+}
+
+func newProvisioningUnderwayCollector(client client.Client) prometheus.Collector {
+	return provisioningUnderwayCollector{
+		client: client,
+		metricClusterDeploymentProvisionUnderwaySeconds: prometheus.NewDesc(
+			"hive_cluster_deployment_provision_underway_seconds",
+			"Length of time a cluster has been provisioning.",
+			[]string{"cluster_deployment", "namespace", "cluster_type", "reason"},
```
dgoodwin, line 83: Apologies I think we should keep "condition" here, because it seems like useful info as the reason could collide. (imagine Unknown for example)

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -0,0 +1,87 @@
...
+	clusterDeployments := &hivev1.ClusterDeploymentList{}
+	err := cc.client.List(context.Background(), clusterDeployments)
+	if err != nil {
+		log.WithError(err).Error("error listing cluster deployments")
+	} else {
+		for _, cd := range clusterDeployments.Items {
+			if cd.DeletionTimestamp == nil {
+				if !cd.Spec.Installed {
+					// Add install failure reason for stuck provision
+					reason := "Unknown"
```
dgoodwin, line 48: lets do Condition = Unknown as a default value here as well so we can differentiate easily between ProvisionFailed + Unknown and literally no condition Unknown + Unknown.

dgoodwin (commented):
Everything else looks good to me.

dgoodwin:
/lgtm 

Can't wait to see this in prometheus tomorrow, thanks!

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -0,0 +1,91 @@
...
+func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
+	ccLog := log.WithField("controller", "metrics")
+	ccLog.Info("calculating provisioning underway metrics across all ClusterDeployments")
+
+	// Load all ClusterDeployments so we can accumulate facts about them.
+	clusterDeployments := &hivev1.ClusterDeploymentList{}
+	err := cc.client.List(context.Background(), clusterDeployments)
+	if err != nil {
+		log.WithError(err).Error("error listing cluster deployments")
+	} else {
```
abhinavdahiya, line 43: I don't think you need else hear, 
```

// use early return on error
if err != nil {
    do something with error
    return
}

// continue with success path
```

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -0,0 +1,91 @@
...
+		for _, cd := range clusterDeployments.Items {
+			if cd.DeletionTimestamp == nil {
+				if !cd.Spec.Installed {
```
abhinavdahiya, line 46:
```
if cd.DeletionTimestamp == nil {
  continue
}

if cd.Spec.Installed {
   continue
}

// actually what matters
```

Keeping the actual business logic to the left as much possible makes reading the code more easy, and also helps not-remembering what the context was when reading it, which is necessary when you have it in nested conditionals.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -0,0 +1,91 @@
...
+								reason = cdCondition.Reason
+							}
+							break
+						}
+					}
+
+					// For installing clusters we report the seconds since the cluster was created.
+					ch <- prometheus.MustNewConstMetric(
+						cc.metricClusterDeploymentProvisionUnderwaySeconds,
+						prometheus.GaugeValue,
```
abhinavdahiya, line 63: > https://prometheus.io/docs/concepts/metric_types/#gauge

`single numerical value that can arbitrarily go up and down.`

> https://prometheus.io/docs/concepts/metric_types/#counter

represents a single monotonically increasing counter whose value can only increase or be reset to zero on restart

shouldn't the time since start be a counter instead of gauge.

dgoodwin: I believe the increase referenced there is by one, the Inc() function, and there is not a way to increment to another value. I think gauge is correct here despite the fact we'll never go down.

pkg/controller/metrics/provision_underway_collector.go
dgoodwin: These are good advice, up to you if you want to cancel the lgtm and update or let it go @suhanime.

pkg/controller/metrics/provision_underway_collector.go
abhinavdahiya: https://godoc.org/github.com/prometheus/client_golang/prometheus#Counter
has an Add(float64) interface

but i agree the guage has more ergonomic API for this usecase since it allows Set(float64)

suhanime (commented):
Addressed - thanks!

pkg/controller/metrics/provision_underway_collector.go
staebler: Strong +1 to this general advice!

PR #1183, opened 2020-10-20
Multiple changes to improve insight into clustersync performance.

dgoodwin:
Use custom buckets for the kube client request histogram, previously capping at 10, we might be going well beyond that in some cases.

Add a reconcileID random string field to the logger in the cluster sync controller. Searching for "reconcile complete" will help us identify slow reconciles, and we can then walk back easily to search just for the log entries for that loop. Went with a random string 8 chars long instead of a full uuid just for brevity and log space, as a true globally unique ID is not really required here.

Add utility for controllers to log their elapsed time and report the metric. Will now log consistent elapsedMillis instead of a duration with varying units. We probably can't do greater than queries in kibana because of limitations in how openshift uses it, but it's still a good idea.

Add elapsedMillisGT categories to let us search logs for requests greater than one of the fixed limits. 

Log kube requests greater than 5 seconds as a warning. This appears quite rare from current metrics.

/assign @suhanime 
/cc @twiest @gregsheremeta 

Greg added you as well just in case you have thoughts.

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
@@ -250,7 +251,10 @@ type ReconcileClusterSync struct {
 // applied or re-applied.
 func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	start := time.Now()
-	logger := r.logger.WithField("ClusterDeployment", request.NamespacedName)
+	logger := r.logger.WithFields(log.Fields{
+		"clusterDeployment": request.NamespacedName,
+		"reconcileID":       utilrand.String(8), // not a true UUID, just a short random string to help us search logs
+	})
 
 	logger.Infof("reconciling ClusterDeployment")
 	defer func() {
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
-	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
@@ -250,19 +249,12 @@ type ReconcileClusterSync struct {
 // Reconcile reads the state of the ClusterDeployment and applies any SyncSets or SelectorSyncSets that need to be
 // applied or re-applied.
 func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	start := time.Now()
-	logger := r.logger.WithFields(log.Fields{
-		"clusterDeployment": request.NamespacedName,
-		"reconcileID":       utilrand.String(8), // not a true UUID, just a short random string to help us search logs
-	})
-
-	// action will be logged to help identify the type of action taken during a reconcile loop:
-	outcome := hivemetrics.ReconcileOutcomeNoOp
-
+	logger := controllerutils.BuildControllerLogger(ControllerName, "clusterDeployment", request.NamespacedName)
 	logger.Infof("reconciling ClusterDeployment")
-	defer func() {
-		hivemetrics.ObserveControllerReconcileTime(ControllerName.String(), start, outcome, logger)
-	}()
+	recobsrv := hivemetrics.NewReconcileObserver(ControllerName, logger)
+	defer recobsrv.ObserveControllerReconcileTime()
+
+	recobsrv.SetOutcome(hivemetrics.ReconcileOutcomeNoOp)
 
 	// Fetch the ClusterDeployment instance
 	cd := &hivev1.ClusterDeployment{}
@@ -316,8 +308,8 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 			logger.WithError(err).Log(controllerutils.LogLevel(err), "could not create ClusterSync")
 			return reconcile.Result{}, err
 		}
+		recobsrv.SetOutcome(hivemetrics.ReconcileOutcomeClusterSyncCreated)
 		// requeue immediately so that we reconcile soon after the ClusterSync is created
-		outcome = hivemetrics.ReconcileClusterSyncCreated
 		return reconcile.Result{Requeue: true}, nil
 	case err != nil:
 		logger.WithError(err).Log(controllerutils.LogLevel(err), "could not get ClusterSync")
@@ -350,7 +342,7 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 	if needToDoFullReapply {
 		logger.Info("need to reapply all syncsets")
 	}
-	outcome = hivemetrics.ReconcileOutcomeFullSync
+	recobsrv.SetOutcome(hivemetrics.ReconcileOutcomeFullSync)
 
 	// Apply SyncSets
 	syncStatusesForSyncSets, syncSetsNeedRequeue := r.applySyncSets(
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -256,11 +256,12 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		"reconcileID":       utilrand.String(8), // not a true UUID, just a short random string to help us search logs
 	})
 
+	// action will be logged to help identify the type of action taken during a reconcile loop:
+	outcome := hivemetrics.ReconcileOutcomeNoOp
+
 	logger.Infof("reconciling ClusterDeployment")
 	defer func() {
-		dur := time.Since(start)
-		hivemetrics.MetricControllerReconcileTime.WithLabelValues(ControllerName.String()).Observe(dur.Seconds())
-		logger.WithField("elapsed", dur).Info("reconcile complete")
+		hivemetrics.ObserveControllerReconcileTime(ControllerName.String(), start, outcome, logger)
 	}()
 
 	// Fetch the ClusterDeployment instance
@@ -316,6 +317,7 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 			return reconcile.Result{}, err
 		}
 		// requeue immediately so that we reconcile soon after the ClusterSync is created
+		outcome = hivemetrics.ReconcileClusterSyncCreated
 		return reconcile.Result{Requeue: true}, nil
 	case err != nil:
 		logger.WithError(err).Log(controllerutils.LogLevel(err), "could not get ClusterSync")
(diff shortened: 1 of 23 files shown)


pkg/controller/utils/clientwrapper.go
```diff
@@ -28,8 +28,9 @@ var (
 		[]string{"controller", "method", "resource", "remote", "status"},
 	)
 	metricKubeClientRequestSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
-		Name: "hive_kube_client_request_seconds",
-		Help: "Length of time for kubernetes client requests.",
+		Name:    "hive_kube_client_request_seconds",
+		Help:    "Length of time for kubernetes client requests.",
+		Buckets: []float64{0.05, 0.1, 0.5, 1, 5, 10, 30, 60, 120},
```
abhinavdahiya, line 33: what is the effect of changing these buckets in existing data that is available in the monitoring system? if there are any existing dashboards that use previous histogram buckets they are probably going to be wrong ot invalid?

also any reason why we chose these specific values?

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -316,6 +317,7 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 			return reconcile.Result{}, err
 		}
 		// requeue immediately so that we reconcile soon after the ClusterSync is created
+		outcome = hivemetrics.ReconcileClusterSyncCreated
```
abhinavdahiya, line 320: i personally think this is magic and not really something some reading this code will immediately understand that the defer will use this outcome.

i would recommend we create a ReconcileObserver

```
robsrv := hivemetrics.NewOutcomeReconcileObserver(ControllerName.String(), logger) // this starts a timer and defautls the outcome to no-op
defer robsrv.ObserveControllerReconcileTime()

...
   // in some code somewhere.
    robsrv.SetOutcome(hivemetrics.ReconcileClusterSyncCreated)
```

pkg/controller/metrics/metrics.go
```diff
@@ -305,9 +305,6 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 		}
 
 		mc.calculateSelectorSyncSetMetrics(mcLog)
-
-		elapsed := time.Since(start)
-		mcLog.WithField("elapsed", elapsed).Info("metrics calculation complete")
```
abhinavdahiya, line 299: any reason why this was dropped?

pkg/controller/metrics/metrics.go
```diff
@@ -597,6 +594,6 @@ func GetClusterDeploymentType(obj metav1.Object) string {
 // optionally an outcome in a consistent fashion.
 func ObserveControllerReconcileTime(controller string, start time.Time, outcome ReconcileOutcome, logger log.FieldLogger) {
 	dur := time.Since(start)
-	metricControllerReconcileTime.WithLabelValues(controller).Observe(dur.Seconds())
-	logger.WithFields(log.Fields{"elapsed": dur, "outcome": outcome}).Info("reconcile complete")
+	metricControllerReconcileTime.WithLabelValues(controller, string(outcome)).Observe(dur.Seconds())
```
abhinavdahiya, line 597: this change should be in the previous commit

pkg/controller/metrics/metrics.go
```diff
@@ -595,5 +595,17 @@ func GetClusterDeploymentType(obj metav1.Object) string {
...
+	for _, millis := range elapsedDurationBuckets {
+		if dur.Milliseconds() >= millis {
+			fields["elapsedMillisGT"] = millis
+			break
+		}
+	}
+	logger.WithFields(fields).Info("reconcile complete")
 }
+
+var elapsedDurationBuckets = []int64{120000, 60000, 30000, 10000, 5000, 1000, 0}
```
abhinavdahiya, line 611: this doesn't make it clear these values are in milliseconds. also i think we can maybe use something like `120_000` to make reading easy in terms of seconds.

pkg/controller/utils/clientwrapper.go
dgoodwin: Suspect it's dropped completely similar to what happens if you alter labels. This should be fine as we have very little historical data regardless. (a week or two)

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -250,7 +251,10 @@ type ReconcileClusterSync struct {
 // applied or re-applied.
 func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	start := time.Now()
-	logger := r.logger.WithField("ClusterDeployment", request.NamespacedName)
+	logger := r.logger.WithFields(log.Fields{
+		"ClusterDeployment": request.NamespacedName,
+		"reconcileID":       utilrand.String(8), // not a true UUID, just a short random string to help us search logs
```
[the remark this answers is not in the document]
dgoodwin: yeah that ClusterDeployment is a bit uncommon in our code, I think reconcileID is probably the more consistent of the two. I will drop the case on that one.

pkg/controller/metrics/metrics.go
dgoodwin: Duplicated with another identical calculation.

pkg/controller/metrics/metrics.go
```diff
@@ -595,5 +595,17 @@ func GetClusterDeploymentType(obj metav1.Object) string {
...
+	fields := log.Fields{"elapsedMillis": dur.Milliseconds(), "outcome": outcome}
+	// Add a log field to categorize request duration into buckets. We can't easily query > in Kibana in
+	// OpenShift deployments due to logging limitations, so these constant buckets give us a small number of
+	// constant search strings we can use to identify slow reconcile loops.
+	for _, millis := range elapsedDurationBuckets {
+		if dur.Milliseconds() >= millis {
+			fields["elapsedMillisGT"] = millis
+			break
+		}
+	}
```
abhinavdahiya, line 607: I think we can use https://golang.org/pkg/sort/#SearchInts maybe because that uses binary search instead..
for this sort array i know this won't be a lot better just saying..

pkg/controller/utils/clientwrapper.go
dgoodwin: As for the values they're just a rough feel for what I'm interested in seeing and our upper range.

pkg/controller/utils/clientwrapper.go
```diff
@@ -117,6 +117,15 @@ func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Respons
 	if err == nil && pathErr == nil {
 		metricKubeClientRequests.WithLabelValues(cmt.Controller.String(), req.Method, path, remoteStr, resp.Status).Inc()
 		metricKubeClientRequestSeconds.WithLabelValues(cmt.Controller.String(), req.Method, path, remoteStr, resp.Status).Observe(applyTime)
+		if applyTime >= 5.0 {
```
abhinavdahiya, line 120: hmm.. isn't applyTime int64 ?

dgoodwin: No the .Seconds() returns float64

pkg/controller/utils/clientwrapper.go
```diff
@@ -117,6 +117,15 @@ func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Respons
...
+			log.WithFields(log.Fields{
+				"controller": cmt.Controller.String(),
+				"method":     req.Method,
+				"path":       path,
+				"remote":     remoteStr,
+				"status":     resp.Status,
+			}).Warn("slow client request")
```
abhinavdahiya, line 127: is the log message really more useful than tracking this in a separate timeseries(metrics)?

pkg/controller/metrics/metrics.go
dgoodwin: Interesting, never saw that syntax before. I will adjust the values and indicate they're millis.

pkg/controller/utils/clientwrapper.go
dgoodwin: Technically it is represented in the timeseries we have now, with Suhani's help today I found 

```
sum without(job, endpoint, pod, service, instance)(hive_kube_client_request_seconds_bucket{remote="true",le="+Inf"} - ignoring(le) hive_kube_client_request_seconds_bucket{remote="true",le="5"} > 0)
```

Which gives us the counter for between 5s and infinity, which we could easily use with rate as well. However I would still like this in place for Thursday's test because it can help if we get into debugging the logs of a slow clustersync reconcile loop, sadly this logger won't have the reconcileID but it still should be possible to pinpoint via the controller and may help us determine if we're slow locally or remotely.

pkg/controller/clustersync/clustersync_controller.go
dgoodwin: This looks marginally more clear, but what is the purpose of the timer you allude to?

pkg/controller/metrics/metrics.go
dgoodwin: I am cursing you for noticing that. :)

pkg/controller/clustersync/clustersync_controller.go
abhinavdahiya: > This looks marginally more clear

to me the use if far more clearer, but :shrug: 

> but what is the purpose of the timer you allude to?

we have a observer struct to maintain the outcome, we could also just stop calling `start := time.Now()` at the top of each function and the creating new observer would call that already, less to work about.

dgoodwin:
Some updates pushed, the observe suggestion will have to wait until tomorrow.

pkg/controller/clustersync/clustersync_controller.go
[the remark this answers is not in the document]
twiest: I like the idea of having a reconcileID. Easier for us to track through the logs.

twiest:
@dgoodwin This looks good to me so far. I like the suggestions and approach. LMK if / when you need me to re-review.

pkg/controller/metrics/metrics.go
```diff
@@ -581,3 +589,23 @@ func GetClusterDeploymentType(obj metav1.Object) string {
 	}
 	return hivev1.DefaultClusterType
 }
+
+// ObserveControllerReconcileTime is used by all controllers in a defer block to log and report their elapsed time and
+// optionally an outcome in a consistent fashion.
+func ObserveControllerReconcileTime(controller string, start time.Time, outcome ReconcileOutcome, logger log.FieldLogger) {
```
staebler, line 595:
```suggestion
func ObserveControllerReconcileTime(controller ControllerName, start time.Time, outcome ReconcileOutcome, logger log.FieldLogger) {
```
So that all of the callers do not have to convert to a string.

pkg/controller/metrics/metrics.go
```diff
@@ -581,3 +589,23 @@ func GetClusterDeploymentType(obj metav1.Object) string {
...
+	for _, millis := range elapsedDurationMillisBuckets {
+		if dur.Milliseconds() >= millis {
+			fields["elapsedMillisGT"] = millis
+			break
+		}
+	}
+	logger.WithFields(fields).Info("reconcile complete")
+}
+
+var elapsedDurationMillisBuckets = []int64{120_000, 60_000, 30_000, 10_000, 5_000, 1_000, 0}
```
staebler, line 611: I personally find it a little more readable if the buckets are defined using durations rather than milliseconds.
```suggestion
	for _, bucket := range elapsedDurationBuckets {
		if dur >= bucket {
			fields["elapsedMillisGT"] = bucket.Milliseconds()
			break
		}
	}
	logger.WithFields(fields).Info("reconcile complete")
}

var elapsedDurationBuckets = []time.Duration{2 * time.Minute, time.Minute, 30 * time.Second, 10 * time.Second, 5 * time.Second, time.Second, 0}
```

pkg/controller/metrics/metrics.go
twiest: This is a really good suggestion. Noted for future PRs.

dgoodwin:
Added the reconcile observer to all controllers and in going though them all again, I decided to add the reconcileID across the board, which also revealed some inconsistencies in how controllers were adding their initial logger fields, so this is all done with a controller util now.

pkg/controller/metrics/metrics.go
dgoodwin: Addressed

dgoodwin:
As before need priority rush on this so it's in and ready for tomorrow's scale testing.

twiest:
@dgoodwin These changes look good. I like the standardization. I'll +1 this whenever you'd like, but I'd prefer to let the others weigh in first.

LMK if you need me to +1 it before they weigh in.

dgoodwin:
Tests are clear, at this point I'd be strongly opposed to another round of e2e for anything other than a major problem in the code, I'd say go for it and I can followup for any more minor issues.

abhinavdahiya (approved):

PR #1185, opened 2020-10-22
First draft of parallel syncset enhancement.

dgoodwin:


docs/enhancements/scalable-syncset-processing.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+
+Even after implementing dramatic changes to improve SyncSet performance, we are struggling to get past a max of about 20 applies/s. Full clustersync controller reconcile time can range from 2 to 5 minutes to process everything for one cluster under load. At 20 goroutines on an AWS c5.4xlarge, the workqueue is struggling to process.
+
+Individual apply time appears to be approximately 0.2s - 1.6s, although 99% of all applys complete in under 0.5s. 0.5s * 600 = 300 seconds to process an entire cluster.
+
+Attempts to increase the goroutines to 100 lead to pod instability, at present problems with allocating OS threads.
+
+Moving to a design where we can horizontally scale the number of pods processing syncsets should let us scale out to meet the requirements of heavier SyncSet usage. In doing so we may be able to push far beyond the 1k ClusterDeployment limit and possibly save substantial cost for extremely large Hive deployments now requiring fewer shards.
```
staebler, line 15: My recollection is that the 1k limit was based on the size of the ClusterSync resources. If that is correct, then horizontally scaling will still not allow Hive to move beyond the 1k limit.

docs/enhancements/scalable-syncset-processing.md
```diff
@@ -0,0 +1,91 @@
...
+## Implementation Details / Notes / Constraints
+
+Deploy a RabbitMQ message bus along-side Hive.
+
+Hive's ClusterSync controller (or equivalent) will serve as the publisher of messages on the bus to a work queue. Each message will contain an identifier for the ClusterDeployment and possibly the SyncSet to process.
+
+Subscriber pods will process these messages, each message will be delivered to only one of the subscriber pods. (a supported configuration with RabbitMQ) Subscriber pods will initially leverage the normal use of Kubernetes clients and caching such that the actual data for the SyncSet resources and cluster kubeconfig Secrets are ready in memory when the message arrives. 
+
+Messages will initially be a request to sync *everything* for a cluster, exactly as the clustersync controller operates today. 
+* Doing this will allow us to have our subscriber pods write to the ClusterSync CR status as the controller does today, which means the API functions as before for users.
```
staebler, line 45: The piece that I see missing in this is something that ensures that only one subscriber is processing a given ClusterDeployment at once. There would need to be come coordination such that the clustersync controller knows that a request has already been sent for the ClusterDeployment and that a new one should not be sent until processing is complete on the first one.

twiest (commented):
/hold

We discussed this in our meeting today and there's possibly a way to achieve this by using built in kubernetes facilities like StatefulSets.

It was decided that we are going to try an implementation of the StatefulSet backed horizontal scaling and come back to this enhancement proposal if that fails.

PR #1186, opened 2020-10-22
vsphere creds checking

joelddiaz:
Fix the immediate problem where passing bad credentials for a vSphere installation means that deleting the ClusterDeployment doesn't get deleted (b/c the deprovision fails (b/c of the bad creds)).

Test the creds, and set a condition so that we can avoid ever even creating a ClusterProvision when we detect bad credentials. This means deprovision works b/c the provision was never attempted. Right now we only check vSphere credentials, but more platforms can (and should) be added.

Add test cases to validate this.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -57,8 +57,8 @@ const (
 	defaultRequeueTime = 10 * time.Second
 	maxProvisions      = 3
 
-	platformAuthFailureResason = "PlatformAuthError"
-	platformAuthSuccessReason  = "PlatformAuthWorking"
+	platformAuthFailureReason = "PlatformAuthError"
+	platformAuthSuccessReason = "PlatformAuthSuccess"
 
 	clusterImageSetNotFoundReason = "ClusterImageSetNotFound"
 	clusterImageSetFoundReason    = "ClusterImageSetFound"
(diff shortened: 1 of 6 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1228,6 +1249,36 @@ func (r *ReconcileClusterDeployment) setDNSNotReadyCondition(cd *hivev1.ClusterD
...
 
+func (r *ReconcileClusterDeployment) setAuthenticationFailure(cd *hivev1.ClusterDeployment, authWorking bool, cdLog log.FieldLogger) (bool, error) {
+
+	var status corev1.ConditionStatus
+	var reason, message string
+
+	if authWorking {
+		status = corev1.ConditionFalse
+		reason = platformAuthSuccessReason
+		message = "Platform credentails passed authentication check"
```
dgoodwin, line 1260: credentails/credentials

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -583,6 +584,26 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+	changed, err := r.setAuthenticationFailure(cd, validCreds, cdLog)
+	if changed || err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// If the platform credentials are no good, do not bother with ClusterProvision objects
+	authCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.AuthenticationFailureClusterDeploymentCondition)
+	if authCondition != nil && authCondition.Status == corev1.ConditionTrue {
+		cdLog.Info("Skipping provision while platform credentials authentication is failing.")
+		// Periodically retry???
```
dgoodwin, line 603: It would be really nice to just requeue every 5 min and be done with it, but there's no reason to think that would help. It probably needs to be done right, watch secrets for changes to a platform secret, requeue the CD if we spot one.

pkg/controller/utils/credentials.go
```diff
@@ -0,0 +1,67 @@
...
+		// If we have no platform-specific credentials verification
+		// assume the creds are valid.
+		return true, nil
+
+	}
+}
+
+func validateVSphereCredentials(vcenter, username, password string) (bool, error) {
+
+	_, _, err := vsphere.CreateVSphereClients(context.TODO(), vcenter, username, password)
```
dgoodwin, line 46: need a log.WithError().Warn here, user may not be able to see what's wrong easily. I'd be tempted to return the error for the message on the condition but that's too risky for hotloops, logging is probably the best we can do.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1443,6 +1444,51 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+				cd := getCD(c)
+				require.NotNil(t, cd, "could not get ClusterDeployment")
+
+				provisionList := &hivev1.ClusterProvisionList{}
+				err := c.List(context.TODO(), provisionList, client.InNamespace(cd.Namespace))
+				require.NoError(t, err, "unexpected error listing ClusterProvisions")
+
+				assert.Zero(t, len(provisionList.Items), "expected no ClusterProvision objects when platform creds are bad")
+			},
+		},
```
dgoodwin, line 1491: I'd like to explore breaking these out into a separate unit test struct if it doesn't result in massive code duplication. If so then this is fine.

dgoodwin (commented):
Looks great.

Given what's going on don't worry about these changes, I or someone on the team will get them.

dgoodwin:
Also please add this new condition to the array in provision_underway_collector.go so it's reported in the alerts if this happens.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
abutcher: We now return an error when credentials are not valid.

PR #1188, opened 2020-10-23
WIP: fake apply

gregsheremeta:
WIP / temp hack to do fake applies for scale testing

to use, set
HIVE_SYNCSETS_FAKE_APPLY_URL
to something like
HIVE_SYNCSETS_FAKE_APPLY_URL=http://httpwait-default.apps.gshereme-spoke-1.new-installer.openshift.com/?wait=

the time to wait will be appended to that

/cc @dgoodwin

### DIFF
--- a/pkg/resource/apply.go
+++ b/pkg/resource/apply.go
@@ -3,7 +3,11 @@ package resource
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
+	"math/rand"
+	"net/http"
+	"os"
 
 	"github.com/jonboulle/clockwork"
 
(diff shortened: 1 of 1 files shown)


gregsheremeta:
If you want to stand up your own fake server:

`server.js`:

```
const http = require('http');
const url = require('url');

function parseWait(wait) {
    const parsed = parseInt(wait);
    if (isNaN(parsed)) {
        return 0;
    }
    return parsed;
}

http.createServer(function (req, res) {
    const queryObject = url.parse(req.url, true).query;
    var wait = parseWait(queryObject['wait']);
    console.log(wait);

    setTimeout(function()
    {
        res.writeHead(200, {'Content-Type': 'text/html'});
        res.end('I waited ' + wait + "ms to reply to you");

    }, wait);

}).listen(8080);
```

`Dockerfile`:

```
FROM node:12

# Create app directory
WORKDIR /usr/src/app

# Bundle app source
COPY server.js .

EXPOSE 8080
CMD [ "node", "server.js" ]
```

PR #1189, opened 2020-10-26
Store install logs in s3 compatible storage

twiest:
- [x] Implement sending install logs to s3 storage on failure
- [x] Remove provisioning PV previously used to store install logs.
- [x] Add unit tests
- [x] Manually test bootstrap node failure

/assign @joelddiaz 
/cc @csrwng

### DIFF
--- a/pkg/apis/hive/v1/hiveconfig_types.go
+++ b/pkg/apis/hive/v1/hiveconfig_types.go
@@ -125,8 +125,9 @@ type VeleroBackupConfig struct {
 // FailedProvisionConfig contains settings to control behavior undertaken by Hive when an installation attempt fails.
 type FailedProvisionConfig struct {
 
-	// SkipGatherLogs disables functionality that attempts to gather full logs from the cluster if an installation
-	// fails for any reason. The logs will be stored in a persistent volume for up to 7 days.
+	// TODO: Figure out how to mark SkipGatherLogs as deprecated (more than just a comment)
+
+	// DEPRECATED: This flag is no longer respected and will be removed in the future.
 	SkipGatherLogs bool                      `json:"skipGatherLogs,omitempty"`
 	AWS            *FailedProvisionAWSConfig `json:"aws,omitempty"`
 }
@@ -159,11 +160,11 @@ type ManageDNSConfig struct {
 type FailedProvisionAWSConfig struct {
 	// CredentialsSecretRef references a secret in the TargetNamespace that will be used to authenticate with
 	// AWS S3. It will need permission to upload logs to S3.
-	// Secret should have key named "cloud" that contains an AWS credentials formatted file.
-	// Example:
-	//   [default]
-	//   aws_access_key_id = minio
-	//   aws_secret_access_key = minio123
+	// Secret should have keys named aws_access_key_id and aws_secret_access_key that contain the AWS credentials.
+	// Example Secret:
+	//   data:
+	//     aws_access_key_id: minio
+	//     aws_secret_access_key: minio123
 	CredentialsSecretRef corev1.LocalObjectReference `json:"credentialsSecretRef"`
 
 	// Region is the AWS region to use for S3 operations.
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -51,8 +51,6 @@ func InstallerPodSpec(
 	provisionName string,
 	releaseImage string,
 	serviceAccountName string,
-	pvcName string,
-	skipGatherLogs bool,
 	extraEnvVars []corev1.EnvVar,
 ) (*corev1.PodSpec, error) {
 
@@ -86,6 +84,12 @@ func InstallerPodSpec(
 			},
 		},
 		{
+			Name: "logs",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+		{
 			Name: "installconfig",
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
@@ -106,6 +110,10 @@ func InstallerPodSpec(
 		{
 			Name:      "output",
 			MountPath: "/output",
+		},
+		{
+			Name:      "logs",
+			MountPath: "/logs",
 		},
 		{
 			Name:      "installconfig",
@@ -278,28 +286,6 @@ func InstallerPodSpec(
 				MountPath: "/manifests",
 			},
 		)
-	}
-
-	if !skipGatherLogs {
-		// Add a volume where we will store full logs from both the installer, and the
-		// cluster itself (assuming we made it far enough).
-		volumes = append(volumes, corev1.Volume{
-			Name: "logs",
-			VolumeSource: corev1.VolumeSource{
-				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-					ClaimName: pvcName,
-				},
-			},
-		})
-		volumeMounts = append(volumeMounts, corev1.VolumeMount{
-			Name:      "logs",
-			MountPath: "/logs",
-		})
-	} else {
-		env = append(env, corev1.EnvVar{
-			Name:  constants.SkipGatherLogsEnvVar,
-			Value: "true",
-		})
 	}
 
 	if cd.Spec.Provisioning.SSHPrivateKeySecretRef != nil {
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -108,14 +108,16 @@ type InstallManager struct {
 	uploadAdminPassword              func(*hivev1.ClusterProvision, *InstallManager) (*corev1.Secret, error)
 	readInstallerLog                 func(*hivev1.ClusterProvision, *InstallManager, bool) (string, error)
 	waitForProvisioningStage         func(*hivev1.ClusterProvision, *InstallManager) error
-	isGatherLogsEnabled              func() bool
 	waitForInstallCompleteExecutions int
 	binaryDir                        string
+	actuator                         LogUploaderActuator
 }
 
 // NewInstallManagerCommand is the entrypoint to create the 'install-manager' subcommand
 func NewInstallManagerCommand() *cobra.Command {
-	im := &InstallManager{}
+	im := &InstallManager{
+		actuator: getActuator(),
+	}
 	cmd := &cobra.Command{
 		Use:   "install-manager NAMESPACE CLUSTER_PROVISION_NAME",
 		Short: "Executes and oversees the openshift-installer.",
@@ -168,7 +170,6 @@ func (m *InstallManager) Complete(args []string) error {
 	m.uploadAdminKubeconfig = uploadAdminKubeconfig
 	m.uploadAdminPassword = uploadAdminPassword
 	m.readInstallerLog = readInstallerLog
-	m.isGatherLogsEnabled = isGatherLogsEnabled
 	m.cleanupFailedProvision = cleanupFailedProvision
 	m.waitForProvisioningStage = waitForProvisioningStage
 
@@ -328,6 +329,7 @@ func (m *InstallManager) Run() error {
 			m.log.WithError(readErr).Error("error reading asset generation log")
 			return err
 		}
+
 		m.log.Info("updating clusterprovision")
 		if err := m.updateClusterProvision(
 			provision,
@@ -415,8 +417,10 @@ func (m *InstallManager) Run() error {
 		}
 
 		// Fetch logs from all cluster machines:
-		if m.isGatherLogsEnabled() {
-			m.gatherLogs(cd, sshKeyPath, sshAgentSetupErr)
+		if m.actuator == nil {
+			m.log.Debug("Unable to find log storage actuator. Disabling gathering logs.")
+		} else {
+			m.gatherLogs(provision, cd, sshKeyPath, sshAgentSetupErr)
 		}
 
 		// TODO: should we timebox this deprovision attempt in the event it gets stuck?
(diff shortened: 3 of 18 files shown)


twiest:
/hold

SRE has asked that we hold merging this until they have the configs in place. This is ready to be reviewed, however.

pkg/apis/hive/v1/hiveconfig_types.go
```diff
@@ -125,7 +125,9 @@ type VeleroBackupConfig struct {
 // FailedProvisionConfig contains settings to control behavior undertaken by Hive when an installation attempt fails.
 type FailedProvisionConfig struct {
 
-	// SkipGatherLogs disables functionality that attempts to gather full logs from the cluster if an installation
+	// TODO: Figure out how to mark SkipGatherLogs as deprecated (more than just a comment)
```
twiest, line 128: I can't seem to find a way to mark SkipGatherLogs as deprecated, other than what I did below with the comment.

I checked kubebuilder markers and didn't find anything related to deprecation. Did I miss one?

Does anyone know how to mark these as deprecated?

csrwng: Looks like the deprecated attribute was introduced in upstream in version 1.19 (ocp 4.6)
https://github.com/kubernetes/kubernetes/pull/92329
I doubt that there is tooling support for it yet

twiest: ok, thanks!

I guess the comment will have to be good enough for now. :)

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/pkg/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+func init() {
+	registerActuator(&awsActuator{awsClientFn: getAWSClient})
+}
+
+// Ensure AWSActuator implements the Actuator interface. This will fail at compile time when false.
+var _ Actuator = &awsActuator{}
```
dgoodwin, line 26: Suggest s3LogUploader, as it may not necessarily be specific to AWS.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+
+// AWSActuator manages getting the desired state, getting the current state and reconciling the two.
```
dgoodwin, line 28: We're not reconciling state here, we're just uploading a file to s3.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+type awsActuator struct {
+	// awsClientFn is the function to build an AWS client, here for testing
+	awsClientFn func(client.Client, string, string, string, log.FieldLogger) (awsclient.Client, error)
+}
+
+// CanHandle returns true if the actuator can handle a particular ClusterDeprovision
```
dgoodwin, line 34: Guessing these are just copied from another actuator, no relevance to ClusterDeprovision here afaik.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return provider == constants.InstallLogsUploadProviderAWS
+}
+
+// UploadLogs uploads installer logs to the provider's storage mechanism.
+func (a *awsActuator) UploadLogs(clusterName string, clusterprovision *hivev1.ClusterProvision, c client.Client, log log.FieldLogger, filenames ...string) error {
+	secretName, foundSecretName := os.LookupEnv(constants.InstallLogsCredentialsSecretRefEnvVar)
+	if !foundSecretName {
+		return errors.New("couldn't find secret name in environment variable. Skipping upload")
+	}
```
dgoodwin, line 50: I would move all these checks into CanHandle/IsConfigured so we can differentiate between are we configured, vs an error uploading a little more clearly.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+	awsc, err := a.awsClientFn(c, secretName, clusterprovision.Namespace, region, log)
+	if err != nil {
+		return err
+	}
+
+	retvalErrs := []error{}
+
+	folder := fmt.Sprintf("%v-%v", clusterName, clusterprovision.Namespace)
+
+	log.Infof("Uploading log(s) to S3: s3://%v/%v/", bucket, folder)
```
dgoodwin, line 71: Suggest "to S3 compatible storage"

pkg/installmanager/installmanager.go
```diff
@@ -415,8 +425,11 @@ func (m *InstallManager) Run() error {
 		}
 
 		// Fetch logs from all cluster machines:
-		if m.isGatherLogsEnabled() {
-			m.gatherLogs(cd, sshKeyPath, sshAgentSetupErr)
+		actuator := m.getActuator()
+		if actuator == nil {
+			m.log.Debug("Unable to find log storage actuator. Disabling gathering logs.")
```
dgoodwin, line 430: Suggest: m.log.Warn("No log storage configured, disabling gathering logs.")

pkg/installmanager/installmanager_test.go
```diff
@@ -237,16 +240,25 @@ func TestInstallManager(t *testing.T) {
 			// We don't want to run the uninstaller, so stub it out
 			im.cleanupFailedProvision = alwaysSucceedCleanupFailedProvision
 
+			// Save the list of actuators so that it can be restored at the end of this test
+			actuatorsSaved := actuators
+			actuators = []Actuator{&awsActuator{awsClientFn: func(c client.Client, secretName, namespace, region string, logger log.FieldLogger) (awsclient.Client, error) {
+				return mocks.mockAWSClient, nil
+			}}}
```
dgoodwin, line 247: This feels off, any time we have to do this in a test carries risk state leaks from one test to another, and in this case I can't see any major reason why we need to track the actuators we know about in a global variable.

How about dropping the global array, add an InstallManager.LogGatherer or Actuator field, and initialize it in NewInstallManagerCommand. Your tests here have direct access to the InstallManager type and are already manipulating fields on it to control what is tested, so you can just im.actuator = awsActuator or equivalent vars if they get renamed.

vendor/modules.txt
```diff
@@ -145,6 +145,7 @@ github.com/dimchansky/utfbom
 github.com/docker/spdystream
 github.com/docker/spdystream/spdy
 # github.com/emicklei/go-restful v2.9.5+incompatible
+## explicit
```
dgoodwin, line 148: Was this a comment you added or something go did?

pkg/installmanager/actuator.go
```diff
@@ -0,0 +1,16 @@
+package installmanager
+
+import (
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Actuator interface is the interface that is used to add cloud provider support to the deprovision controller.
```
dgoodwin, line 9: deprovision controller copy paste.

pkg/installmanager/actuator.go
```diff
@@ -0,0 +1,16 @@
...
+type Actuator interface {
```
dgoodwin, line 10: WDYT about LogUploader here? I don't love the idea of using the actuator terminology because we're doing something so different here. We use actuators where we fork behavior for a cluster depending on the cloud it's on, and I think typically when we're doing a desired/current reconciliation.

In this case we're checking if a log upload storage is configured for multiple implementations and telling it to upload something, we might use s3 upload for gcp/azure/any cloud provider clusters provisioned in a Hive that's running on AWS, or even on any platform to minio.

pkg/installmanager/actuator.go
```diff
@@ -0,0 +1,16 @@
...
+	// CanHandle returns true if the actuator can handle a particular case
+	CanHandle() bool
```
dgoodwin, line 12: Suggest: IsConfigured(), CanHandle sounds like we'd provide it an argument to see if it can handle that, but that is not our case here.

dgoodwin (commented):
Functionally looking great, just some concerns about organization and naming.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Ensure AWSActuator implements the Actuator interface. This will fail at compile time when false.
+var _ Actuator = &awsActuator{}
+
+// AWSActuator manages getting the desired state, getting the current state and reconciling the two.
+type awsActuator struct {
+	// awsClientFn is the function to build an AWS client, here for testing
+	awsClientFn func(client.Client, string, string, string, log.FieldLogger) (awsclient.Client, error)
+}
```
abhinavdahiya, line 32: hmm.. i don't really see the benefit of this,

- globally registered list of actuators that use init function to configure them. Can we just explicitly have a list locally where it matters. Is there a benefit to this dependency injection here?
- why not just have a awsclient.Client interface directly embed in the struct and use a newAWSActuator function to create that when we know we can configure it. This style of dependency injection imo is only useful that it abstracts an external API/function not what is required right here like the client to AWS.

pkg/installmanager/installmanager.go
```diff
@@ -849,14 +866,36 @@ func (m *InstallManager) gatherLogs(cd *hivev1.ClusterDeployment, sshPrivKeyPath
...
+		m.log.WithError(err).WithField("clusterprovision", types.NamespacedName{Name: provision.Name, Namespace: provision.Namespace}).Error("error reading Logsdir")
+		return
+	}
+
+	filepaths := []string{}
+	for _, file := range files {
+		if !file.IsDir() {
+			filepaths = append(filepaths, filepath.Join(m.LogsDir, file.Name()))
+		}
+	}
```
abhinavdahiya, line 893: this seems like it will only include the first level of files in the LogsDir, is that what is expected? shouldn't we upload all the files from all the children dirs too? or atleast log that we are skipping those dirs?

pkg/install/generate.go
```diff
@@ -83,6 +81,12 @@ func InstallerPodSpec(
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
 			},
 		},
+		{
+			Name: "logs",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
```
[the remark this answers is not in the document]
twiest: @dgoodwin I believe it's a better approach because there are limits on the size of the container FS, which limits usually fairly small (at least they are in OSD). I prefer the EmptyDir approach, but I'll remove it if you'd like.

@abhinavdahiya My understanding is that emptydir is not a ramdisk. They're scratch volumes.

vendor/modules.txt
twiest: Nope. It was added by tooling.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/aws/aws-sdk-go/service/s3"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	awsclient "github.com/openshift/hive/pkg/awsclient"
+	"github.com/openshift/hive/pkg/constants"
+
+	restful "github.com/emicklei/go-restful"
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
twiest: I'm not sure what you're saying here. Where should the errors import go?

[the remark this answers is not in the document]
dgoodwin: Sections now: aws+logrus+sigs.k8s.io, hive, go-restful+errors+k8s.io

Typically that would be something like: logrus+go-restful+errors (random things), aws, hive, k8s.io+sigs.k8s.io

pkg/installmanager/installmanager.go
twiest: When I tested this, the only directory was `lost+found`. That's why this check is in there. I'll add a log message when skipping.

pkg/installmanager/awsactuator.go
```diff
@@ -0,0 +1,113 @@
...
+			Body:               file,
+			ContentLength:      aws.Int64(stat.Size()),
+			ContentType:        aws.String(restful.MIME_OCTET),
+			ContentDisposition: aws.String("attachment"),
+		})
+
+		if err != nil {
+			retvalErrs = append(retvalErrs, errors.Wrapf(err, "Failed uploading log file: %v", filename))
+		}
+	}
```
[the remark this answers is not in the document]
twiest: The only problem with that is that we're renaming the files as we upload them and this doesn't seem like it supports that ability.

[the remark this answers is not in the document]
abhinavdahiya: in the example the key and filepath seem separate https://github.com/aws/aws-sdk-go/blob/d0f43158ad3bcbe89ab04b7fdef545cf7a591154/example/service/s3/sync/sync.go#L27-L28 , so i think re-named should be possible

twiest:
@dgoodwin @abhinavdahiya I believe I've addressed your feedback. PTAL.

abhinavdahiya (commented):
/lgtm

twiest:
Just remembered that this is on hold waiting for SRE to add configuration to their clusters. Will touch base with them in the next sync up meeting.

twiest:
Here is the SRE card we're waiting on in order to merge this:
https://issues.redhat.com/browse/OSD-5712

twiest:
@abhinavdahiya @dgoodwin I rebased and changed a code comment to more clearly represent what we're expecting in the secret. The code in this PR hasn't changed.

Please re-lgtm :)

twiest:
@dgoodwin hey, I rebased this PR and added a commit for updating logextractor.sh and the docs.

The previous commit didn't change, so the latest commit is the only one that needs to be reviewed.

PTAL.

twiest:
Today I verified the configs with SRE. This PR can now merge.

/hold cancel

PR #1190, opened 2020-10-28
Major improvement for syncset apply performance and simulated scale testing.

dgoodwin:
Adds a new developer annotation that can be applied to HiveConfig, hive.openshift.io/fake-applies: true, which will configure hive to fake out all Apply operations (SyncSets) to remote clusters. This is for simulated scale testing and allows us to use one spoke cluster and adopt it thousands of times. 

Commit still included for Greg's original work in this area in case we ever want to reference or use it, his solution uses a nodejs server which we tell to wait a specified ms before returning. In my testing a Sleep performs fairly similar, a little faster but I believe this is because we are double simulating requests in the nodejs approach, if we lookup real world data for the sleep array, then do an http request telling it to wait this time, it will be slower than the data we looked up. I believe Sleep is a reasonable simulation and much easier to commit in this fashion for re-use, however I wanted to keep the http request sim in the commit history.

Major fix in this PR is for performance, I found that we surprisingly still are doing OpenAPI schema builds on every single resource apply. We found this previously and it was the driver for moving to ClusterSync instead of SyncSetInstance controllers, however somewhere we missed this along the way. This operation took 3s when we first discovered it in the wild IIRC, and in my testing against one cluster it was taking 11s. Done 600 times, we can't complete a reconcile loop in less than an hour.

This fix moves the schema build up into the constructor for the resource apply Helper and then stores it as a field.

### DIFF
--- a/pkg/controller/utils/clientwrapper.go
+++ b/pkg/controller/utils/clientwrapper.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -34,11 +35,18 @@ var (
 	},
 		[]string{"controller", "method", "resource", "remote", "status"},
 	)
+	metricKubeClientRequestsCancelled = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "hive_kube_client_requests_cancelled_total",
+		Help: "Counter incremented for each kube client request cancelled.",
+	},
+		[]string{"controller", "method", "resource", "remote"},
+	)
 )
 
 func init() {
 	metrics.Registry.MustRegister(metricKubeClientRequests)
 	metrics.Registry.MustRegister(metricKubeClientRequestSeconds)
+	metrics.Registry.MustRegister(metricKubeClientRequestsCancelled)
 }
 
 // NewClientWithMetricsOrDie creates a new controller-runtime client with a wrapper which increments
@@ -103,6 +111,21 @@ type ControllerMetricsTripper struct {
 	Remote     bool
 }
 
+func (cmt *ControllerMetricsTripper) CancelRequest(req *http.Request) {
+	// "CancelRequest not implemented by *utils.ControllerMetricsTripper" was seen during simulated scale
+	// testing as the cluster was falling over. We will now implement this and track a metric for requests
+	// being cancelled as this could prove valuable in tracking when we have a performance issue.
+	remoteStr := strconv.FormatBool(cmt.Remote)
+	path, _ := parsePath(req.URL.Path)
+	metricKubeClientRequestsCancelled.WithLabelValues(cmt.Controller.String(), req.Method, path, remoteStr).Inc()
+	log.WithFields(log.Fields{
+		"controller": cmt.Controller.String(),
+		"method":     req.Method,
+		"path":       path,
+		"remote":     remoteStr,
+	}).Warn("cancelled request")
+}
+
 // RoundTrip implements the http RoundTripper interface.
 func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	startTime := time.Now()
@@ -111,6 +134,7 @@ func (cmt *ControllerMetricsTripper) RoundTrip(req *http.Request) (*http.Respons
 		remoteStr = "true"
 	}
 	path, pathErr := parsePath(req.URL.Path)
+
 	// Call the nested RoundTripper.
 	resp, err := cmt.RoundTripper.RoundTrip(req)
 	applyTime := metav1.Now().Sub(startTime)
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -50,6 +50,12 @@ const (
 	// hiveConfigHashAnnotation is annotation on hivedeployment that contains
 	// the hash of the contents of the hive-controllers-config configmap
 	hiveConfigHashAnnotation = "hive.openshift.io/hiveconfig-hash"
+
+	// fakeAppliesAnnotation is an annotation which can be added to HiveConfig indicating that
+	// Hive should be configured to completely fake all "apply" operations to target clusters.
+	// This is used for scale testing simulations where we want to adopt one cluster thousands
+	// of times, but not overwhelm it with requests.
+	fakeAppliesAnnotation = "hive.openshift.io/fake-applies"
 )
 
 func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper, instance *hivev1.HiveConfig, recorder events.Recorder, mdConfigMap *corev1.ConfigMap) error {
@@ -162,6 +168,21 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 				Value: instance.Spec.Backup.Velero.Namespace,
 			}
 			hiveContainer.Env = append(hiveContainer.Env, tmpEnvVar)
+		}
+	}
+
+	if _, ok := instance.Annotations[fakeAppliesAnnotation]; ok {
+		fakeApplies, err := strconv.ParseBool(instance.Annotations[fakeAppliesAnnotation])
+		if err != nil {
+			hLog.WithError(err).Errorf("unable to parse %s as a boolean", fakeAppliesAnnotation)
+			return err
+		}
+		if fakeApplies {
+			hLog.Warnf("%s specified, configuring hive to fake all resource helper apply/delete operations", fakeAppliesAnnotation)
+			hiveContainer.Env = append(hiveContainer.Env, corev1.EnvVar{
+				Name:  hiveconstants.HiveSyncSetsFakeApplyEnvVar,
+				Value: "true",
+			})
 		}
 	}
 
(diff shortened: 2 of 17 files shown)


dgoodwin:
/assign @twiest 
/cc @csrwng 

Cesar could you verify the resource helper changes look ok to you?

pkg/resource/apply.go
```diff
@@ -41,21 +47,24 @@ const fieldTooLong metav1.CauseType = "FieldValueTooLong"
...
+	applyOptions, changeTracker, err := r.setupApplyCommand(obj, ioStreams)
 	if err != nil {
 		r.logger.WithError(err).Error("failed to setup apply command")
 		return "", err
 	}
+
+	// fakeApply mode (used in scale testing only) doesn't run the apply. Instead it will Sleep for a period of time
+	// that matches our request time metrics in real hive environments to simulate i/o wait.
+	fakeAppliesEnvVar, _ := os.LookupEnv(constants.HiveSyncSetsFakeApplyEnvVar)
+	if fakeApplies, err := strconv.ParseBool(fakeAppliesEnvVar); fakeApplies && err == nil {
```
csrwng, line 64: should the env var and and ParseBool be called when constructing the resource helper instead of every apply?

pkg/resource/apply.go
```diff
@@ -125,11 +123,29 @@ func (r *helper) CreateRuntimeObject(obj runtime.Object, scheme *runtime.Scheme)
...
+	time.Sleep(wait)
+}
+
 func (r *helper) createOnly(f cmdutil.Factory, obj []byte) (ApplyResult, error) {
 	info, err := r.getResourceInternalInfo(f, obj)
 	if err != nil {
 		return "", err
 	}
+	if info == nil {
+		r.logger.Debug("err getting info")
```
csrwng, line 147: Should likely return an error here. Is this something you've observed? (where err == nil and info == nil?)

pkg/resource/info.go
```diff
@@ -21,11 +21,7 @@ type Info struct {
 
 // Info determines the name/namespace and type of the passed in resource bytes
 func (r *helper) Info(obj []byte) (*Info, error) {
-	factory, err := r.getFactory("")
-	if err != nil {
-		return nil, err
-	}
-	resourceInfo, err := r.getResourceInfo(factory, obj)
+	resourceInfo, err := r.getResourceInfo(r.factory, obj)
```
csrwng, line 24: just `return r.getResourceInfo(r.factory, obj)`

pkg/resource/helper.go
```diff
@@ -44,59 +47,70 @@ type helper struct {
...
-func NewHelperWithMetrics(kubeconfig []byte, controllerName hivev1.ControllerName, remote bool, logger log.FieldLogger) Helper {
-	r := &helper{
-		logger:         logger,
-		controllerName: controllerName,
-		metricsEnabled: true,
-		cacheDir:       getCacheDir(logger),
-		kubeconfig:     kubeconfig,
-		remote:         true,
+	// OpenAPISchema lookup is an extremely expensive operation, taking upwards of 3s often.
+	// We want to be very sure to only due this once per interaction with a cluster and as such,
```
csrwng, line 78: s/due/do/

csrwng (commented):
Hey @dgoodwin, looks good. Some initial comments.
Just thinking about the fakeApply, I wonder if it gives you a realistic simulation of load given that regardless of the network communication, cpu cycles are normally spent on serializing/deserializing resources.

dgoodwin:
> Just thinking about the fakeApply, I wonder if it gives you a realistic simulation of load given that regardless of the network communication, cpu cycles are normally spent on serializing/deserializing resources.

Ideally we'd plug the fake apply in at a lower level and hit more serialization, but this appears to be as low as we can get without getting into modifying vendored code. I'm wondering if we're hitting most of the serialization though:

From prod cluster data, we can see that the average apply time is about 0.4s, average GET from clustersync controller is about 0.046, PATCH 0.047, POST 0.085. I'm not entirely sure what combo of those would get done for a normal apply, probably just a GET and a no-op? So network maybe 46ms out of 400ms total for an apply.

In our test cluster we're seeing applys take 70ms (although with beefier nodes, c5.4xlarge as opposed to m5.4xlarge), and simulating an avg around 46ms. This may indicate you are right. I'm going to try to get some data for pods running on m5.xlarge today and see how close we are to the apply time in prod. I'm not sure how we could get it more accurate, maybe try to replicate some of the expensive function calls up here in our code? Or maybe we just can't reliably commit this hack and need to do it by editing vendored code *when* we're doing scale testing.

abhinavdahiya:
qq: does fetching openAPI data once and re using affect list of manifests that install a CRD and the it's corresponding CR ? The clients usually use the openAPI data to figure out the api endpoint for a GroupVersionKind and old openAPI data will not have the new API created by the CRD ...

dgoodwin:
> qq: does fetching openAPI data once and re using affect list of manifests that install a CRD and the it's corresponding CR ? The clients usually use the openAPI data to figure out the api endpoint for a GroupVersionKind and old openAPI data will not have the new API created by the CRD ...

We've seen this fail even when we rebuilt the openapi schema every request as well, it seems sometimes in the past kube can be too slow to fully register a CRD and make it available, and if you try to use it on the next request it might fail saying no such resource.

This will likely be possible to surface but the clustersync should just requeue, which rebuilds the openapi schema and the next attempt should work. If you were doing a lot of CRDs + CRs, it could be a number of syncs for a successful reconcile. Definitely something to think about, but hopefully will be ok, and probably our only option, I don't think we can leave it the way it is. 

Perhaps a rebuild if we see we sync a CRD?

pkg/resource/apply.go
dgoodwin: I honestly don't even remember adding this. I'll remove, probably something from when I was stuck trying to figure out what was wrong.

pkg/resource/helper.go
```diff
@@ -44,59 +47,70 @@ type helper struct {
...
+// NewHelperWithMetricsFromRESTConfig returns a new object that allows apply and patch operations, with metrics tracking enabled.
+func NewHelperWithMetricsFromRESTConfig(restConfig *rest.Config, controllerName hivev1.ControllerName, logger log.FieldLogger) (Helper, error) {
+	r, err := NewHelperFromRESTConfig(restConfig, logger)
+	if err != nil {
+		return nil, err
 	}
-	r.getFactory = r.getKubeconfigFactory
-	return r
+	rh, _ := r.(*helper)
+	rh.metricsEnabled = true
```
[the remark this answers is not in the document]
dgoodwin: Superb catch, this probably broke metrics.

dgoodwin:
PR is updated for feedback. I don't have a good idea yet on how to better simulate the serialization we're missing by bypassing some of the apply code. I think we can defer this for the time being and the current form as our baseline for now, with horizontal scaling pods we can mitigate any differences in time spent serializing if necessary.

dgoodwin:
/hold
Legit failure, looks like some of the dropped kubeconfig factory code for the NewHelper(kubeconfig) variant was ruined in my refactor, I am trying to figure out how to restore.

pkg/controller/utils/clientwrapper.go
```diff
@@ -103,6 +110,24 @@ type ControllerMetricsTripper struct {
 	Remote     bool
 }
 
+func (cmt *ControllerMetricsTripper) CancelRequest(req *http.Request) {
+	// "CancelRequest not implemented by *utils.ControllerMetricsTripper" was seen during simulated scale
+	// testing as the cluster was falling over. We will now implement this and track a metric for requests
+	// being cancelled as this could prove valuable in tracking when we have a performance issue.
+	remoteStr := "false"
```
csrwng, line 117: nit: `remoteStr := strconv.FormatBool(cmt.Remote)`

pkg/operator/hive/hive.go
```diff
@@ -165,6 +171,14 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 		}
 	}
 
+	if fakeApplies, err := strconv.ParseBool(instance.Annotations[fakeAppliesAnnotation]); fakeApplies && err == nil {
+		hLog.Warnf("%s specified, configuring hive to fake all SyncSet apply operations", fakeAppliesAnnotation)
+		hiveContainer.Env = append(hiveContainer.Env, corev1.EnvVar{
+			Name:  hiveconstants.HiveSyncSetsFakeApplyEnvVar,
+			Value: "true",
+		})
+	}
```
csrwng, line 180: should at least log an error if the value can't be parsed (or even better, just exit/crash)

csrwng (commented):
Current code looks good to me, a couple of nits, feel free to ignore.

pkg/resource/patch.go
```diff
@@ -27,11 +27,7 @@ func (r *helper) Patch(name types.NamespacedName, kind, apiVersion string, patch
 		Out:    &bytes.Buffer{},
 		ErrOut: &bytes.Buffer{},
 	}
-	factory, err := r.getFactory(name.Namespace)
```
dgoodwin, line 30: This is where I got into trouble, I didn't notice that some of these getFactory commands were specifying a namespace, so my attempt to store the factory/dynClient/etc on the helper and re-use them was misguided. I'm reworking it to only cache the openapi schema.

pkg/operator/hive/hive.go
dgoodwin: This is an operator controller and we don't seem to have a precedent for crashing the pod if something is misconfigured for the hive controllers deployment, so maybe log and return an error here is best as that seems to be consistent.

dgoodwin:
PR is now updated, I had to rework the resource handler patch quite a bit, it's closer to what it was before now, with just a temporary factory used during resource helper constuction to get the openapi schema, and then that is all we cache on the helper. 

Also updated with your last suggestions @csrwng

PR #1191, opened 2020-10-30
Add dev prometheus config for scraping multiple pods.

dgoodwin:
This is in preparation for the hive-clustersync pods that we expect to
soon be running, we no longer will have one pod to scrape by going to a
service, rather we'll want to scrape all pods by endpoints behind the
service.

Breaks out hive-controllers (name matches app-sre configuration,
although it doesn't matter as this is just dev prometheus setup for
scale testing), and adds hive-clustersync job. Both are capable of
scraping multiple pods, though this is not strictly relevant for
hive-controllers, it should work fine.

### DIFF
--- a/config/prometheus/prometheus-configmap.yaml
+++ b/config/prometheus/prometheus-configmap.yaml
@@ -14,7 +14,24 @@ data:
         static_configs:
           - targets: ['localhost:9090']
 
-      - job_name: 'hive'
-        static_configs:
-          - targets: ['hive-controllers:2112']
+      - job_name: 'hive-controllers'
+        kubernetes_sd_configs:
+          - role: endpoints
+        relabel_configs:
+          - source_labels:
+              - __meta_kubernetes_namespace
+              - __meta_kubernetes_endpoints_name
+              - __meta_kubernetes_endpoint_port_name
+            action: keep
+            regex: hive;hive-controllers;metrics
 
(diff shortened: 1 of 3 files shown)


dgoodwin:
/assign @twiest 
/cc @suhanime 

I think this can go in separately but it is in preparation for Thomas' work. This has no impact on any living prometheus it's just what we document for developers.

config/prometheus/prometheus-configmap.yaml
```diff
@@ -14,7 +14,24 @@ data:
         static_configs:
           - targets: ['localhost:9090']
 
-      - job_name: 'hive'
-        static_configs:
-          - targets: ['hive-controllers:2112']
+      - job_name: 'hive-controllers'
+        kubernetes_sd_configs:
+          - role: endpoints
```
abhinavdahiya, line 19: just for next time https://prometheus.io/docs/prometheus/latest/configuration/configuration/#endpoints

dgoodwin: Could you clarify what you mean above? I'm not following.

dgoodwin:
Bump @twiest @suhanime

suhanime:
> Bump @twiest @suhanime

Seems okay to me, but I will let someone else LGTM

dgoodwin:
[test]

PR #1192, opened 2020-11-03
Configure DNS before stopping install when attempts limit is set

akhil-rane:
This PR ensures that DNS is configured before hive halts installation 
when install attempts limit is set in the ClusterDeployment

jira: https://issues.redhat.com/browse/CO-1251

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -64,6 +64,10 @@ const (
 	dnsReadyReason     = "DNSReady"
 	dnsReadyAnnotation = "hive.openshift.io/dnsready"
 
+	installAttemptsLimitReachedReason = "InstallAttemptsLimitReached"
+	installOnlyOnceSetReason          = "InstallOnlyOnceSet"
+	provisionNotStoppedReason         = "ProvisionNotStopped"
+
 	deleteAfterAnnotation    = "hive.openshift.io/delete-after" // contains a duration after which the cluster should be cleaned up.
 	tryInstallOnceAnnotation = "hive.openshift.io/try-install-once"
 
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -711,6 +719,43 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		}
 	}
 
+	if cd.Annotations[configureDNSSkipInstallAnnotation] == "true" {
```
dgoodwin, line 722: Is there a reason why the InstallAttemptsLimit from recent PR check couldn't move right here, it would still allow you to limit provision attempts but we assume DNS is still fair game to configure. I can't think of a reason why limit install attempts would mean DON'T setup DNS, our one request for this asked specifically for DNS to be setup. And if it ever arose that someone did want that, we could tell them to just not set manageDNS up, or add it when they're ready to proceed. Thoughts? 

I'm not crazy about having more annotations blocking install attempts, this is the third.

(technically the install-once annotation that was first in this area should be removed I think at some point)

akhil-rane: yes you are right we can do this. About removing "install-once" annotation, do we want to go ahead and remove it now?

dgoodwin: Awesome thanks!

As for install-once annotation I added a jira and while writing it I realized that ACM is probably using it for bare metal, so we can't remove without coordinating with them. We can worry about that later.

akhil-rane:
@dgoodwin addressed your feedback.

dgoodwin:
Terrified, thanks!

/lgtm

twiest:
This looks good to me as well.

PR #1193, opened 2020-11-03
Patch APIServer config for certs instead of assuming full control.

dgoodwin:
To allow other callers to manage parts of the APIServer configuration
in-cluster, switch to patching in the certs we need instead of fully
managing the resource.

Refactors tests a bit to not rely on validation functions and instead
test most things consistently across every set of test inputs.

### DIFF
--- a/pkg/controller/controlplanecerts/controlplanecerts_controller.go
+++ b/pkg/controller/controlplanecerts/controlplanecerts_controller.go
@@ -7,12 +7,12 @@ import (
 	"io"
 	"net/url"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
-	k8slabels "github.com/openshift/hive/pkg/util/labels"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -30,7 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
-	configv1 "github.com/openshift/api/config/v1"
 	apihelpers "github.com/openshift/hive/pkg/apis/helpers"
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 	"github.com/openshift/hive/pkg/constants"
@@ -38,6 +37,7 @@ import (
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/remoteclient"
 	"github.com/openshift/hive/pkg/resource"
+	k8slabels "github.com/openshift/hive/pkg/util/labels"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


dgoodwin:
/hold 

Once this is lgtm'd I want to grab all the cp-certs syncsets in stage and save them for comparison to after this rolls out. I don't have creds to check cluster apiserver configs but maybe I can get some help with that.

pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
```diff
@@ -203,8 +174,44 @@ func TestReconcileControlPlaneCerts(t *testing.T) {
...
 
-			if test.validate != nil {
-				test.validate(t, fakeClient, applier.appliedObjects)
+			cd := getFakeClusterDeployment(t, fakeClient)
+
+			if test.expectNoSyncSet {
+				assert.Len(t, applier.appliedObjects, 0, "unexpected syncset apply")
+			} else {
+				assert.Len(t, applier.appliedObjects, 1, "single apply expected")
+				assert.IsType(t, &hivev1.SyncSet{}, applier.appliedObjects[0], "syncset apply expected")
```
joelddiaz, line 183: should this be a `require.IsType` so that we don't bother with the next tests if this is not true?

dgoodwin: Made this and the line above requires.

abutcher (commented):
Changes look good to me, defer to @joelddiaz for tag

joelddiaz (commented):
/lgtm

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -357,6 +330,41 @@ func (r *ReconcileControlPlaneCerts) generateControlPlaneCertsSyncSet(cd *hivev1
...
+		if i > 0 {
+			buf.WriteString(",")
+		}
+		bundle := certificateBundle(cd, additional.Name)
+		cdLog.WithField("name", additional.Name).Debug("adding named certificate to control plane config")
+		buf.WriteString(fmt.Sprintf(` { "names": [ "%s" ], "servingCertificate": { "name": "%s" } }`,
+			additional.Domain, remoteSecretName(bundle.CertificateSecretRef.Name, cd)))
+	}
+
+	var kubeAPIServerNamedCertsTemplate = `[ {"op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [ %s ] } ]`
```
abhinavdahiya, line 362:
```suggestion
	var kubeAPIServerNamedCertsTemplate = `[ { "op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [ %s ] } ]`
```
this was in consistent with other json patch template being added

abhinavdahiya (commented):
/lgtm

pkg/resource/patch.go
```diff
@@ -43,6 +43,9 @@ func (r *helper) Patch(name types.NamespacedName, kind, apiVersion string, patch
 			WithField("stderr", ioStreams.ErrOut.(*bytes.Buffer).String()).Warn("running the patch command failed")
 		return err
 	}
+	r.logger.
+		WithField("stdout", ioStreams.Out.(*bytes.Buffer).String()).
+		WithField("stderr", ioStreams.ErrOut.(*bytes.Buffer).String()).Info("patch command successful")
```
dgoodwin, line 48: Apologies reviewers, I did end up pushing another change. I was starting to request SRE-P help to get the apiserver configs from some clusters before and after when I realized I can just log the stdout from the patch command to see if anything was changed or not. Much simpler to verify if my work did or did not make changes to the config during the transition.

Also got the spacing issue you mentioned @abhinavdahiya 

Thanks all.

joelddiaz (commented):
/lgtm

PR #1198, opened 2020-11-05
Add some ClusterPool documentation

dgoodwin:


docs/clusterpools.md
```diff
@@ -0,0 +1,83 @@
...
+claimed, a cluster is removed from the pool and a new one will be created to
+replace it. Claimed clusters never return to the pool, they are intended to be
+destroyed when no longer needed. The `ClusterClaim.Spec.Namespace` will be
+populated once the claim has been filled, and the `ClusterDeployment` will be
+present in that namespace with an identical name.
+
+`ClusterPools` are a namespaced resource, and can be used to centralize billing
+by using a namespace limited to a team via Kubernetes RBAC. All clusters in the
+pool will use the same set of cloud credentials specified in the platform for
+the pool. `ClusterClaims` are created in the same namespace as their
```
suhanime, line 22:
```suggestion
the pool. `ClusterClaims` must be created in the same namespace as their
```

docs/clusterpools.md
```diff
@@ -0,0 +1,83 @@
...
+`ClusterPool`, but each actual `ClusterDeployment` is given it's own namespace.
```
suhanime, line 23:
```suggestion
`ClusterPool`, but each actual `ClusterDeployment` is given its own namespace.
```

docs/clusterpools.md
```diff
@@ -0,0 +1,83 @@
...
+prevent anyone else from being able to access it.
+
+Presently once a `ClusterDeployment` is ready, it will be
+[hibernated](./hibernating-clusters.md) automatically. Once claimed it will be
+automatically resumed, meaning that the typical time to claim a cluster and be
+ready to go is in the 2-5 minute range while the cluster starts up. An option
+to keep the clusters in the pool always running so they can instantly be used
+may be added in the future.
+
+When done with a cluster, users can just delete their `ClusterClaim` and the
```
suhanime, line 34: It would be a good idea to mention the `lifetime` of a ClusterClaim in this section

suhanime (commented):
Looks good @dgoodwin. Just some nits here

We might want to add a tiny section for clusterpools and clusterclaim in hiveutil docs.

docs/clusterpools.md
dgoodwin: Very good idea, also added to the sample yaml below.

dgoodwin:
Feedback addressed including a couple examples for the hiveutil docs as well. Thanks!

docs/clusterpools.md
```diff
@@ -0,0 +1,86 @@
...
+`ClusterPool`, but each actual `ClusterDeployment` is given its own namespace.
+The user who claims a cluster can be given RBAC to their clusters namespace to
+prevent anyone else from being able to access it.
+
+Presently once a `ClusterDeployment` is ready, it will be
+[hibernated](./hibernating-clusters.md) automatically. Once claimed it will be
+automatically resumed, meaning that the typical time to claim a cluster and be
+ready to go is in the 2-5 minute range while the cluster starts up. An option
+to keep the clusters in the pool always running so they can instantly be used
+may be added in the future.
```
abhinavdahiya, line 32: > An option
to keep the clusters in the pool always running so they can instantly be used
may be added in the future.

hmm.. is it okay to mix docs with roadmap ?

abhinavdahiya: rewording it as a Note would be better imo

docs/clusterpools.md
```diff
@@ -0,0 +1,86 @@
...
+clusters.
+
+When a user needs a cluster they create a `ClusterClaim` resource which will be
+filled immediately with details on where to find their cluster.  (or as soon as
+a cluster is available in the pool if none were presently available). Once
+claimed, a cluster is removed from the pool and a new one will be created to
+replace it. Claimed clusters never return to the pool, they are intended to be
+destroyed when no longer needed. The `ClusterClaim.Spec.Namespace` will be
+populated once the claim has been filled, and the `ClusterDeployment` will be
+present in that namespace with an identical name.
```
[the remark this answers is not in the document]
dgoodwin: Yes I am not happy with it when I realized how confusing it looked writing this, but I fear this ship has sailed as it's already published and in use.  It was probably done that way because we try to do that for anything we want to survive a backup and can't easily discover, however in this case we probably could discover it by looking for a cluster matching the label.

PR #1203, opened 2020-11-09
Fix control plane cert patching for new clusters.

dgoodwin:
Recent change was not tested with new clusters as all eyes were on
making sure we rolled out for existing clusters smoothly.

JSON patches will not apply if the json sections referenced do not yet
exist.

This change adds two new directions to the json patch to ensure that the
two parent sections for what we need to patch exist. Because these are
all within the same json patch, they do not modify the resource. (which
they would if we applied each individually)

Because we need to ensure the .spec.serviceCerts node exists (by inserting
an empty node followed by then adding the list of certs), we still
effictively own everything underneath .spec.serviceCerts and will stomp
any non-hive changes when reconciling the control plane certs.

### DIFF
--- a/pkg/controller/controlplanecerts/controlplanecerts_controller.go
+++ b/pkg/controller/controlplanecerts/controlplanecerts_controller.go
@@ -363,7 +363,7 @@ func (r *ReconcileControlPlaneCerts) getServingCertificatesJSONPatch(cd *hivev1.
 			additional.Domain, remoteSecretName(bundle.CertificateSecretRef.Name, cd)))
 	}
 
-	var kubeAPIServerNamedCertsTemplate = `[ { "op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [ %s ] } ]`
+	var kubeAPIServerNamedCertsTemplate = `[ { "op": "add", "path": "/spec/servingCerts", "value": {} }, { "op": "add", "path": "/spec/servingCerts/namedCertificates", "value": [  ] }, { "op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [ %s ] } ]`
 	namedCerts := buf.String()
 	return fmt.Sprintf(kubeAPIServerNamedCertsTemplate, namedCerts), nil
 
--- a/pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
+++ b/pkg/controller/controlplanecerts/controlplanecerts_controller_test.go
@@ -81,7 +81,7 @@ func TestReconcileControlPlaneCerts(t *testing.T) {
 				fakeClusterDeployment().defaultCert("default-cert", "default-secret").obj(),
 				fakeCertSecret("default-secret"),
 			},
-			expectedPatch:   `[ { "op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [  { "names": [ "test-api-url" ], "servingCertificate": { "name": "fake-cluster-default-secret" } } ] } ]`,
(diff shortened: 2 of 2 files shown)


cblecker:
/lgtm
/hold

Because this is impacting staging clusters, I tested this patch manually on my cluster and it worked successfully. I also reapplied the patch and there was no change.

Placing a hold for further review by hive team, but this seems safe to merge whenever.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -363,7 +363,7 @@ func (r *ReconcileControlPlaneCerts) getServingCertificatesJSONPatch(cd *hivev1.
 			additional.Domain, remoteSecretName(bundle.CertificateSecretRef.Name, cd)))
 	}
 
-	var kubeAPIServerNamedCertsTemplate = `[ { "op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [ %s ] } ]`
+	var kubeAPIServerNamedCertsTemplate = `[ { "op": "add", "path": "/spec/servingCerts", "value": {} }, { "op": "add", "path": "/spec/servingCerts/namedCertificates", "value": [  ] }, { "op": "replace", "path": "/spec/servingCerts/namedCertificates", "value": [ %s ] } ]`
```
abhinavdahiya, line 366: nit: why can't we combine the two add operations into one so that we add `{"nameCertificates": []}` to `/spec/servingCerts` ?

joelddiaz (commented):
looks like there is a question about simplifying the patch itself, but otherwise :+1:

pkg/controller/controlplanecerts/controlplanecerts_controller.go
abhinavdahiya: doing some testing,
```json
{
	"spec": {
		"servingCerts": {
			"another key": "some value"
		}
	}
}
```

Applying this patch will,

```
[{
	"op": "add",
	"path": "/spec/servingCerts",
	"value": {}
}, {
	"op": "add",
	"path": "/spec/servingCerts/namedCertificates",
	"value": []
}, {
	"op": "replace",
	"path": "/spec/servingCerts/namedCertificates",
	"value": [{
		"domain": "test-api.com",
		"secret": "some-secret-test-api-com"
	}]
}]
```

will create new object without the `another key` ??
```
{
  "spec" : {
    "servingCertificates" : {
      "another key" : "some value"
    },
    "servingCerts" : {
      "namedCertificates" : [ {
        "domain" : "test-api.com",
        "secret" : "some-secret-test-api-com"
      } ]
    }
  }
}
```

I tried it out here, https://json-schema-validator.herokuapp.com/jsonpatch.jsp

I don't think that is correct, right?

dgoodwin: I tried this and got a failure as well, it doesn't like sections not existing and it turns out spec.servingCerts is not set either in this scenario, so it seemed like I needed to add it explicitly first.

dgoodwin: @abhinavdahiya previously this assumed full control over the entire APIServer resource, and this transitions us to assuming control over just the servingCerts and below. It is true that any user attempts to edit other serving cert fields would get overwritten, but I think this is acceptable.

I'm open to work around it if there's a reasonable way to do so but I don't see how to conditionally add spec.servingCerts if it's missing in a json patch. Any ideas?

abhinavdahiya: >  this transitions us to assuming control over just the servingCerts and below. It is true that any user attempts to edit other serving cert fields would get overwritten, but I think this is acceptable

Let's move this to the PR descriptions/ commit message because the current description imo does not make that clear. And then we can move ahead and merge it.

jessesarn:
@dgoodwin / @abhinavdahiya / @cblecker / @joelddiaz We have a pretty significant need for this to fix QE testing of a release that's well overdue due to broken clusters in stage.  Is there anything we can do to help get this moving along more quickly?

blrm:
Now that the hive team has reviewed this, I'm going to drop the hold.

/unhold

cblecker:
/retest
Failed to complete before 4h timeout.

PR #1204, opened 2020-11-09
WIP: Add e2e for control plane certificates.

dgoodwin:
/assign @csrwng 
/cc @joelddiaz

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -175,10 +175,16 @@ else
 	CLUSTER_DOMAIN="${BASE_DOMAIN}"
 fi
 
-echo "Generating letsencrypt certificates for cluster"
-go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" certificate create "${CLUSTER_NAME}" --base-domain "${CLUSTER_DOMAIN}" --output-dir "${SRC_DIR}"
-# This will cause a hive-controllers pod restart:
-${SRC_ROOT}/hack/set-additional-ca.sh "${SRC_DIR}/${CLUSTER_NAME}.ca"
+# Our certificates command only works against AWS today, but this gives us the coverage we need to ensure
+# control plane certs get deployed.
+if [ "$CLOUD" = "aws" ]; then
+	echo "Generating letsencrypt certificates for AWS cluster"
+	go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" certificate create "${CLUSTER_NAME}" \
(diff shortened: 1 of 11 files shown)


dgoodwin:
Temporarily includes a change from #1203 as this wouldn't pass otherwise.

csrwng:
@dgoodwin should we wait to ensure that the certs were applied? My thinking is that this can introduce some flakiness in the post-install tests. The post-install test will create a client to access the target cluster. If at the time the test starts, certs have not been applied, but they take effect while the post-install test is running, the test will stop communicating with the target cluster and fail.

csrwng:
looks like you also need to have the test script install certbot

dgoodwin:
Oof yeah I forgot to verify they got applied. :) Will work on it, thanks.

joelddiaz (commented):
This looks reasonable, but how do we know that the certs have been applied before these postinstall e2e tests run?

dgoodwin:
> This looks reasonable, but how do we know that the certs have been applied before these postinstall e2e tests run?

I was looking at the SyncSet tests and there it hasn't seemed necessary to do much polling to wait, it's not too hard to add some retries but it also didn't seem necessary, though I'll have to see if we start seeing failures. I can add some retries though if you think it's worth it.

joelddiaz:
> > This looks reasonable, but how do we know that the certs have been applied before these postinstall e2e tests run?
> 
> I was looking at the SyncSet tests and there it hasn't seemed necessary to do much polling to wait, it's not too hard to add some retries but it also didn't seem necessary, though I'll have to see if we start seeing failures. I can add some retries though if you think it's worth it.

Seems fine as is. I was just wondering since I'm not so familiar with our e2e.

hack/e2e-test.sh
```diff
@@ -33,6 +33,20 @@ if ! which kustomize > /dev/null; then
   popd
 fi
 
+if ! which certbot > /dev/null; then
+  certbot_dir="$(mktemp -d)"
+  export PATH="$PATH:${certbot_dir}"
+  pushd "${certbot_dir}"
+  wget https://dl.eff.org/certbot-auto
```
joelddiaz, line 40: Do we really want to pull unsigned code down like this?

dgoodwin: We do the same just above for kustomize, I think it's probably ok in e2e test pods. 

It's in epel but I didn't really want to play the "Am I running in EPEL / Fedora / OSX" game in the e2e test script.

dgoodwin:
Was not able to get this to happen, mainly because I could not get cert generation working in the CI images. Not a priority at this time.

/close

PR #1205, opened 2020-11-10
CO-1107: Wait to Hibernate ClusterPool Clusters until SyncSets Applied

abutcher:
*  Set a `ClusterSync.Status.FirstSuccessTime` of `time.Now()` when there are no `SyncStatuses` found for `ClusterSync`. This means there are no `SyncSets` to apply.
* Determine if `SyncSets` have been applied in hibernation controller
  * If no `ClusterSync.Status.FirstSuccessTime` has been set for 10 minutes since `ClusterDeployment.Status.InstalledTimestamp`, hibernate anyway.
  * If 10 minutes have not passed since install and `ClusterSync.Status.FirstSuccessTime` is unset, requeue `ClusterDeployment` for time remaining to meet 10 minutes since cluster was installed.

/assign @dgoodwin 
/cc @akhil-rane

### DIFF
--- a/pkg/apis/hive/v1/clusterdeployment_types.go
+++ b/pkg/apis/hive/v1/clusterdeployment_types.go
@@ -360,6 +360,9 @@ const (
 	// FailedToStartHibernationReason is used when there was an error starting machines
 	// to leave hibernation
 	FailedToStartHibernationReason = "FailedToStart"
+	// SyncSetsNotAppliedReason is used as the reason when SyncSets have not yet been applied
+	// for the cluster based on ClusterSync.Status.FirstSucessTime
+	SyncSetsNotAppliedReason = "SyncSetsNotApplied"
 )
 
 // +genclient
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -169,10 +169,19 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (result rec
 		if supported, msg := r.hibernationSupported(cd); !supported {
 			return r.setHibernatingCondition(cd, hivev1.UnsupportedHibernationReason, msg, corev1.ConditionFalse, cdLog)
 		}
-		if syncSetsApplied, msg := r.syncSetsApplied(cd); !syncSetsApplied {
-			// Allow hibernation if hibernateAfterSyncSetsNotApplied duration has passed since cluster installed and syncsets still not applied
+
+		// Determine if SyncSets have been applied
+		clusterSync := &hiveintv1alpha1.ClusterSync{}
+		err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("could not get ClusterSync: %v", err)
+		}
+		// SyncSets have not been applied
+		if clusterSync.Status.FirstSuccessTime == nil {
+			// Allow hibernation (do not set condition) if hibernateAfterSyncSetsNotApplied duration has passed since cluster
+			// installed and syncsets still not applied
 			if cd.Status.InstalledTimestamp != nil && time.Now().Sub(cd.Status.InstalledTimestamp.Time) < hibernateAfterSyncSetsNotApplied {
-				return r.setHibernatingCondition(cd, hivev1.SyncSetsNotAppliedReason, msg, corev1.ConditionFalse, cdLog)
+				return r.setHibernatingCondition(cd, hivev1.SyncSetsNotAppliedReason, "Cluster SyncSets have not been applied", corev1.ConditionFalse, cdLog)
 			}
 		}
 	}
@@ -367,6 +376,7 @@ func (r *hibernationReconciler) setHibernatingCondition(cd *hivev1.ClusterDeploy
 			logger.Infof("cluster will reconcile due to syncsets not applied in: %v", requeueAfter)
 			result.RequeueAfter = requeueAfter
 			result.Requeue = true
+			returnErr = errors.New(message)
 		}()
 	}
 
@@ -405,18 +415,6 @@ func (r *hibernationReconciler) hibernationSupported(cd *hivev1.ClusterDeploymen
 		return false, fmt.Sprintf("Unsupported version, need version %s or greater", minimumClusterVersion.String())
 	}
 	return true, "Hibernation capable"
-}
-
-func (r *hibernationReconciler) syncSetsApplied(cd *hivev1.ClusterDeployment) (bool, string) {
-	clusterSync := &hiveintv1alpha1.ClusterSync{}
-	err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync)
-	if err != nil {
-		return false, fmt.Sprintf("could not get ClusterSync: %v", err)
-	}
-	if clusterSync.Status.FirstSuccessTime == nil {
-		return false, "Cluster SyncSets have not been applied"
-	}
-	return true, ""
 }
 
 func (r *hibernationReconciler) nodesReady(cd *hivev1.ClusterDeployment, remoteClient client.Client, logger log.FieldLogger) (bool, error) {
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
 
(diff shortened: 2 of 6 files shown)


dgoodwin:
Would this make sense on the hibernation controller instead of clusterpool? I'm thinking the cluster pool doesn't really care about this, it's job is just to create the clusters. The hibernation controller when reconciling can just check, are all syncsets applied, and if not are you installed more than 10 minutes ago, and requeue for that 10 minute threshold if not. We then would not need to reconcile a pool every time a clustersync changes for a clusterdeployment from that pool.

We'd essentially be changing from saying it's the job of the clusterpool controller to set the desired state when it knows syncsets are applied, to saying hibernation only kicks in when syncsets are applied or the cluster is over 10 minutes since installed across the board.

WDYT?

abutcher:
I think that’s fine if we’re okay with SyncSets driving all hibernation. I
implemented the change in the hibernation controller originally and moved
it all to pool and it was pretty simple on that side.

dgoodwin:
Ok cool, lets get some team feedback before making any changes.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -234,6 +278,37 @@ func (r *ReconcileClusterPool) Reconcile(request reconcile.Request) (reconcile.R
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileClusterPool) hibernateReadyClusters(cds []*hivev1.ClusterDeployment) ([]*hivev1.ClusterDeployment, error) {
```
akhil-rane, line 281: This function returns ClusterDeployments for which syncsets have not been applied. Should we name it to suggest that?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -26,18 +27,20 @@ import (
...
-	finalizer                  = "hive.openshift.io/clusters"
-	imageSetDependent          = "cluster image set"
-	pullSecretDependent        = "pull secret"
-	credentialsSecretDependent = "credentials secret"
+	ControllerName                   = hivev1.ClusterpoolControllerName
+	finalizer                        = "hive.openshift.io/clusters"
+	imageSetDependent                = "cluster image set"
+	pullSecretDependent              = "pull secret"
+	credentialsSecretDependent       = "credentials secret"
+	hibernateAfterSyncSetsNotApplied = time.Minute * 10
```
akhil-rane, line 43: Some documentation to indicate what this variable is doing will be helpful

akhil-rane, line 43: Some documentation to indicate what this variable is doing will be really helpful

pkg/controller/clusterpool/clusterpool_controller.go
akhil-rane, line 281: This function returns ClusterDeployments for which syncsets have not been applied. Should we name it to suggest that?

akhil-rane (commented):
I agree with @dgoodwin 's suggestion here. Its the job of the hibernation controller to ensure that all conditions are met before hibernating a cluster.

akhil-rane:
I loved the way you have structured the commit message :)

pkg/controller/clusterpool/clusterpool_controller.go
abutcher: This has moved to the hibernation controller but I added a comment for it.

pkg/controller/clusterpool/clusterpool_controller.go
abutcher: This function has gone away in the refactor.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -360,6 +360,9 @@ const (
 	// FailedToStartHibernationReason is used when there was an error starting machines
 	// to leave hibernation
 	FailedToStartHibernationReason = "FailedToStart"
+	// SyncSetsNotAppliedReason is used as the reason when SyncSets have not yet been applied
+	// for the cluster based on ClusterSync.Status.FirstSucessTime
+	SyncSetsNotAppliedReason = "SyncSetsNotApplied"
```
dgoodwin, line 365: Used on the hibernating condition, good call.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -157,16 +163,23 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (result rec
...
 	if shouldHibernate || cd.Spec.HibernateAfter != nil {
-		if supported, msg := r.canHibernate(cd); !supported {
+		if supported, msg := r.hibernationSupported(cd); !supported {
 			return r.setHibernatingCondition(cd, hivev1.UnsupportedHibernationReason, msg, corev1.ConditionFalse, cdLog)
 		}
+		if syncSetsApplied, msg := r.syncSetsApplied(cd); !syncSetsApplied {
+			// Allow hibernation if hibernateAfterSyncSetsNotApplied duration has passed since cluster installed and syncsets still not applied
+			if cd.Status.InstalledTimestamp != nil && time.Now().Sub(cd.Status.InstalledTimestamp.Time) < hibernateAfterSyncSetsNotApplied {
+				return r.setHibernatingCondition(cd, hivev1.SyncSetsNotAppliedReason, msg, corev1.ConditionFalse, cdLog)
+			}
```
dgoodwin, line 176: When this is not the case and we've past expiry, is there any indication on the cluster deployment that we were not able to apply syncsets even though hibernating is now true? I suspect there probably is, via our pre-existing syncsets failing condition. If so this is great.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -346,12 +359,23 @@ func (r *hibernationReconciler) setHibernatingCondition(cd *hivev1.ClusterDeploy
...
 	}
+
+	if reason == hivev1.SyncSetsNotAppliedReason {
+		defer func() {
+			expiry := cd.Status.InstalledTimestamp.Time.Add(hibernateAfterSyncSetsNotApplied)
+			requeueAfter := time.Until(expiry)
+			logger.Infof("cluster will reconcile due to syncsets not applied in: %v", requeueAfter)
+			result.RequeueAfter = requeueAfter
+			result.Requeue = true
+		}()
```
dgoodwin, line 370: Looked like this might run after the if, but I looked it up and it appears it probably runs after the function we're in, so :+1:

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -383,6 +407,18 @@ func (r *hibernationReconciler) canHibernate(cd *hivev1.ClusterDeployment) (bool
 	return true, "Hibernation capable"
 }
 
+func (r *hibernationReconciler) syncSetsApplied(cd *hivev1.ClusterDeployment) (bool, string) {
+	clusterSync := &hiveintv1alpha1.ClusterSync{}
+	err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync)
+	if err != nil {
+		return false, fmt.Sprintf("could not get ClusterSync: %v", err)
```
dgoodwin, line 414: feels like we may need to return an error here and force a requeue rather than following the path for normal execution when we're waiting for syncsets to reach first success.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -73,23 +81,51 @@ func TestReconcile(t *testing.T) {
...
+			setupActuator: func(actuator *mock.MockHibernationActuator) {
+				actuator.EXPECT().StopMachines(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(nil)
+			},
+			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
+				cond := getHibernatingCondition(cd)
+				require.NotNil(t, cond)
+				assert.Equal(t, corev1.ConditionTrue, cond.Status)
+				assert.Equal(t, hivev1.StoppingHibernationReason, cond.Reason)
+			},
+		},
```
dgoodwin, line 124: Would it be feasible to add a test for when there are no syncsets to apply or is that covered elsewhere. I guess it's every other test isn't it.

dgoodwin: Also are we missing a test for when SyncSets have been applied successfully?

pkg/controller/hibernation/hibernation_controller.go
abutcher: We'll have a `SyncSetApplyFailure` failing condition, right.

```  - lastProbeTime: "2020-11-18T15:40:04Z"
    lastTransitionTime: "2020-11-18T15:40:04Z"
    message: One of the SyncSet applies has failed
    reason: SyncSetApplyFailure
    status: "True"
    type: SyncSetFailed
  - lastProbeTime: "2020-11-18T15:40:04Z"
    lastTransitionTime: "2020-11-18T15:40:02Z"
    message: Cluster SyncSets have not been applied
    reason: SyncSetsNotApplied
    status: "False"```

pkg/controller/hibernation/hibernation_controller_test.go
abutcher: Added a test to ensure `cs.Status.FirstSuccessTime` is set w/ no syncsets and another in hibernation for syncsets applied successfully.

pkg/controller/hibernation/hibernation_controller.go
abutcher: Thinking about this I'm not sure that we need to error and go into back off if we're requeuing for the time to force hibernate. I could do this though by squashing the returns from setting the condition and returning an err from reconcile.

abutcher: Suppose it would hibernate sooner in the event that the syncsets did succeed.

PR #1207, opened 2020-11-11
Fix e2e to adapt to multi-steps

suhanime:
Hive CI config was recently changed to use "multi-steps" workflow for
IPI installs of e2e tests.
This commit changes the e2e hack scripts to adapt to the environment
variables made available from the CI change.

PR #1208, opened 2020-11-12
Report DNSNotReady in the cluster provisioning underway metric.

dgoodwin:
Update the metric to only use conditions if their status is true
indicating an unexpected state.

Add DNSNotReady at the start of the list, so that if this is true it is
the reason we report without checking the others.

Similarly reorder provision launch error as this is likely to surface
before an install failure.

/assign @suhanime 
/cc @twiest


suhanime:
/hold

To hold the retesting until [1207](https://github.com/openshift/hive/pull/1207) merges

PR #1209, opened 2020-11-13
Add note on what platforms are supported for ClusterPool.

dgoodwin:
/assign @suhanime


suhanime (commented):
/lgtm

Can't believe we forgot this

PR #1210, opened 2020-11-13
Implement horizontal scaling of clustersync

twiest:
- [x] Implement horizontal scaling of clustersync
- [x] Implement deployment of horizontal scaling in hive-operator
- [x] Add unit tests
- [x] Manually test
- [x] Run performance tests to ensure performance gains

/assign @dgoodwin 
/cc @abutcher


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -268,6 +338,18 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if me, err := r.isSyncAssignedToMe(cd, logger); !me || err != nil {
+		if err != nil {
+			logger.WithError(err).Error("failed determining which instance is assigned to sync this cluster")
+			recobsrv.SetOutcome(hivemetrics.ReconcileOutcomeSyncErrored)
```
dgoodwin, line 344: This outcome is making me a little nervous, it looks like the metrics today do not differentiate success / failure, and yet we don't set outcomes for errors anywhere else. I don't want individual reconcile errors needing an outcome each as that will bloat cardinality. I'd like a reconcile error to generically inject success/failure across all controllers when we report it but I'm not quite sure how to do that. We could perhaps dynamically combine the given outcome with "Success/Failure" someday.

I'd propose we drop this outcome for now and defer reporting error outcomes until a future PR. For now outcome = the action we tried to take, not an indicator of success/failure.

pkg/controller/metrics/metrics.go
```diff
@@ -109,6 +109,8 @@ type ReconcileOutcome string
 const (
 	ReconcileOutcomeUnspecified        ReconcileOutcome = "unspecified"
 	ReconcileOutcomeNoOp               ReconcileOutcome = "no-op"
+	ReconcileOutcomeSkippedSync        ReconcileOutcome = "skipped-sync"
+	ReconcileOutcomeSyncErrored        ReconcileOutcome = "sync-errored"
```
dgoodwin, line 113: Suggested dropping this one above.

pkg/operator/hive/clustersync.go
```diff
@@ -0,0 +1,137 @@
...
+	hiveconstants "github.com/openshift/hive/pkg/constants"
+	"github.com/openshift/hive/pkg/controller/images"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	"github.com/openshift/hive/pkg/operator/assets"
+	"github.com/openshift/hive/pkg/operator/util"
+	"github.com/openshift/hive/pkg/resource"
+)
+
+func (r *ReconcileHiveConfig) deployClusterSync(hLog log.FieldLogger, h resource.Helper, instance *hivev1.HiveConfig) error {
+	asset := assets.MustAsset("config/clustersync/statefulset.yaml")
```
dgoodwin, line 20: You're well positioned here, lets just lump in that other card we talked about, add a setting in HiveConfig for configuring the number of replicas. https://issues.redhat.com/browse/CO-1290 can be deleted once done.

pkg/operator/hive/clustersync.go
```diff
@@ -0,0 +1,137 @@
...
+			hiveContainer.Env,
+			corev1.EnvVar{
+				Name:  images.HiveImagePullPolicyEnvVar,
+				Value: string(r.hiveImagePullPolicy),
+			},
+		)
+	}
+
+	// clustersync is the only controller that should run.
+	hiveContainer.Args = append(hiveContainer.Args, "--controllers", "clustersync")
```
dgoodwin, line 49: Any reason this is done here and not hard coded in the yaml?

pkg/operator/hive/clustersync.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	result, err := util.ApplyRuntimeObjectWithGC(h, hiveControllersConfigMap, instance)
+	if err != nil {
+		hLog.WithError(err).Error("error applying hive-controllers-config configmap")
+		return err
+	}
+	hLog.WithField("result", result).Info("hive-controllers-config configmap applied")
+
+	hLog.Info("Hashing hive-controllers-config data onto a hive deployment annotation")
```
dgoodwin, line 120: s/hive deployment/hive-clustersync statefulset

pkg/operator/hive/clustersync.go
```diff
@@ -0,0 +1,137 @@
...
+	hiveControllersConfigHash := computeHiveControllersConfigHash(hiveControllersConfigMap)
+	if clusterSyncStatefulSet.Spec.Template.Annotations == nil {
+		clusterSyncStatefulSet.Spec.Template.Annotations = make(map[string]string, 1)
+	}
+	clusterSyncStatefulSet.Spec.Template.Annotations[hiveConfigHashAnnotation] = hiveControllersConfigHash
+
+	clusterSyncStatefulSet.Namespace = hiveNSName
+	result, err = util.ApplyRuntimeObjectWithGC(h, clusterSyncStatefulSet, instance)
+	if err != nil {
+		hLog.WithError(err).Error("error applying deployment")
```
dgoodwin, line 130: statefulset

pkg/operator/hive/hive.go
```diff
@@ -88,9 +88,9 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 		)
 	}
 
-	if dc := instance.Spec.DisabledControllers; len(dc) != 0 {
-		hiveContainer.Args = append(hiveContainer.Args, "--disabled-controllers", strings.Join(dc, ","))
-	}
+	// Always add clustersync to the list of disabled controllers since clustersync is running in a statefulset now.
+	disabledControllers := append(instance.Spec.DisabledControllers, "clustersync")
+	hiveContainer.Args = append(hiveContainer.Args, "--disabled-controllers", strings.Join(disabledControllers, ","))
```
dgoodwin, line 93: Same question here, feels like this could be hard coded in the yaml to avoid one avenue of confusion for people.

dgoodwin (commented):
Looks great, mostly small stuff, but I do think we may as well do the HiveConfig to update the replicas to call this done and dusted.

pkg/controller/clustersync/clustersync_controller.go
twiest: Should I just remove this SetOutcome then?

From the code, I believe that would mean we'd get ReconcileOutcomeNoOp for this error path.

I'm fine with that, just making sure that's what you're intending here, or if there's a different outcome you'd prefer.

pkg/operator/hive/clustersync.go
twiest: Just to be clear, you're asking for that card to be done as part of this PR, correct?

pkg/operator/hive/hive.go
twiest: I agree on the statefulset one, but on this one, I need to be able to add to the list of disabled controllers that's in hiveconfig.

If I add it to the yaml, then the hiveconfig disabled controllers will stomp on this and if clustersync isn't disabled in the hiveconfig disabled controllers list, then it would run.

I'm willing to change, it just seems wrong to do on this one.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -246,6 +250,72 @@ type ReconcileClusterSync struct {
...
+		// This ensures that we don't have partial syncing which may make it seem like things are working.
+		// TODO: Remove this once https://issues.redhat.com/browse/CO-1268 is completed as this should no longer be needed.
+		return false, fmt.Errorf("statefulset replica count is off. current: %v  expected: %v", sts.Status.CurrentReplicas, *sts.Spec.Replicas)
+	}
+
+	logger.Debug("Getting HIVE_CLUSTERSYNC_POD_NAME")
+	podname, found := os.LookupEnv("HIVE_CLUSTERSYNC_POD_NAME")
+	if !found {
+		return false, errors.New("environment variable HIVE_CLUSTERSYNC_POD_NAME not set")
+	}
```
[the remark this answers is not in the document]
twiest: Fixed.

pkg/operator/hive/hive.go
dgoodwin: Ah that makes sense, sounds good to me.

pkg/operator/hive/clustersync.go
dgoodwin: Yes please.

pkg/controller/clustersync/clustersync_controller.go
dgoodwin: Yeah no-op for now. Thanks.

twiest:
Holding until manual testing is completed.

/hold

twiest:
@dgoodwin I believe I've addressed your feedback, PTAL.

I'm still doing some manual testing of the feedback changes, so there may be minor changes if I find any bugs, but the overall functionality should now be there.

Thanks!

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -244,6 +252,73 @@ type ReconcileClusterSync struct {
...
+	logger.Debug("Ensuring replicas is set")
+	if sts.Spec.Replicas == nil {
+		return false, errors.New("sts.Spec.Replicas not set")
+	}
+
+	if sts.Status.CurrentReplicas != *sts.Spec.Replicas {
+		// This ensures that we don't have partial syncing which may make it seem like things are working.
+		// TODO: Remove this once https://issues.redhat.com/browse/CO-1268 is completed as this should no longer be needed.
+		return false, fmt.Errorf("statefulset replica count is off. current: %v  expected: %v", sts.Status.CurrentReplicas, *sts.Spec.Replicas)
+	}
```
abhinavdahiya, line 281: This feels out of place in `isSyncAssignedToMe` function. The task of if any reconciliation should be happening should be out of scope for this function. I would probably check before calling this function. Also the unit tests will be more clearer.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -244,6 +252,73 @@ type ReconcileClusterSync struct {
...
+
+	logger.Debug("Getting ordinalID")
+	parts := strings.Split(r.podname, "-")
+	ordinalIDStr := parts[len(parts)-1]
+	ordinalID32, err := strconv.Atoi(ordinalIDStr)
+	if err != nil {
+		return false, errors.Wrap(err, "error converting ordinalID to int")
+	}
+
+	ordinalID := int64(ordinalID32)
```
abhinavdahiya, line 305: This should be happening once per controller start/initialize right? Why are we calculating our ordinal id, every sync cycle?

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -244,6 +252,73 @@ type ReconcileClusterSync struct {
...
+
+	if cd == nil {
+		return false, errors.New("clusterdeployment is nil")
+	}
+
+	logger.Debug("Getting uid for hashing")
+	var uidAsBigInt big.Int
+	hexUID := strings.Replace(string(cd.UID), "-", "", 4)
+	logger.Debugf("hexUID: %+v", hexUID)
+	uidAsBigInt.SetString(hexUID, 16)
```
abhinavdahiya, line 295: this seems like quite a bit of assumption here, 
- one that there are 4 sections
- and each one is a hex number.

Since these UIDs are all UUIDv4 can we link in the commit/here why these assumptions hold true?

config/crds/hive.openshift.io_hiveconfigs.yaml
```diff
@@ -114,6 +114,11 @@ spec:
                               QPS for a controller
                             format: int32
                             type: integer
+                          replicas:
+                            description: Replicas specifies the number of replicas
+                              the specific controller pod should use.
```
dgoodwin, line 119: Could you elaborate here that this is only for controllers which have been split out to run in their own pods. For others it will be ignored.

pkg/controller/utils/clusterdeployment_test.go
```diff
@@ -66,39 +66,40 @@ func TestShouldSyncCluster(t *testing.T) {
 		{
 			name:     "no annotation",
 			cd:       clusterdeployment.Build(),
-			expected: true,
```
dgoodwin, line 69: TestShouldSyncCluster function could use a rename now.

dgoodwin (commented):
Looking great, couple very minor comments and then Abhinav's feedback and I think we're good to go.

pkg/operator/hive/clustersync.go
```diff
@@ -0,0 +1,130 @@
...
+			clusterSyncStatefulSet.Spec.Replicas = pointer.Int32Ptr(defaultClustersyncReplicas)
+		} else {
+			clusterSyncControllerConfig, found := getHiveControllerConfig(hivev1.ClustersyncControllerName, hiveconfig.Spec.ControllersConfig.Controllers)
+
+			if !found || clusterSyncControllerConfig.Replicas == nil {
+				clusterSyncStatefulSet.Spec.Replicas = pointer.Int32Ptr(defaultClustersyncReplicas)
+			} else {
+				clusterSyncStatefulSet.Spec.Replicas = clusterSyncControllerConfig.Replicas
+			}
+		}
```
abhinavdahiya, line 117:
```suggestion
		clusterSyncStatefulSet.Spec.Replicas = pointer.Int32Ptr(defaultClustersyncReplicas)
		// Set the number of replicas that was given in the hiveconfig
		if hiveconfig.Spec.ControllersConfig != nil {
			clusterSyncControllerConfig, found := getHiveControllerConfig(hivev1.ClustersyncControllerName, hiveconfig.Spec.ControllersConfig.Controllers)
			if found && clusterSyncControllerConfig.Replicas != nil {
				clusterSyncStatefulSet.Spec.Replicas = clusterSyncControllerConfig.Replicas
			}
		}
```

dgoodwin:
One suggestion from @abhinavdahiya but otherwise lgtm.

/hold

until scale test is over today.

twiest:
level=error msg="Error: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached."

/test e2e

twiest:
level=fatal msg="failed to fetch Master Machines: failed to fetch dependency of \"Master Machines\": failed to generate asset \"Platform Credentials Check\": validate AWS credentials: mint credentials check: error simulating policy: Throttling: Rate exceeded\n\tstatus code: 400, request id: 78b93ffa-d098-4430-b46f-c1923b5955f6"

/test e2e

dgoodwin:
Not sure why the bot isn't retesting this one.

PR #1212, opened 2020-11-17
Add two new GCP specific install failure search strings.

dgoodwin:
Will now set a provision failed condition reason for bad project IDs and
non-existent instance types on GCP. Both should flow out through the
alert as well.

/assign @abutcher 
/cc @twiest


gregsheremeta:
This is great.

I wonder if @staebler can help us push these into the installer.

staebler:
> This is great.
> 
> I wonder if @staebler can help us push these into the installer.

@gregsheremeta I'm not sure what you mean. Both of these failure messages are coming from the installer.

twiest:
This looks good to me. I was actually wondering today where we specify these regexes. :)

PR #1213, opened 2020-11-19
Add support for editing Labels and Taints on MachinePools

akhil-rane:
This PR ensures that editing 'labels' and 'taints' on MachinePools gets
reflected in the underlying MachineSet created. Currently, both 'labels' and
'taints' fields are validated as immutable fields by hive admission webhook
and hence cannot be updated. Changing it to treat them as mutable.

jira: https://issues.redhat.com/browse/CO-1265


nimrodshn:
Thanks @akhil-rane !

PR #1214, opened 2020-11-19
initial support for installconfigtemplate in clusterpool

hassenius:
Please consider this pull request to enhance clusterpool to take a install-config.yaml as a secret in a similar fashion to the ClusterDeployment.

This will allow Clusterpool to accept an object like this

```yaml
kind: ClusterPool
metadata:
  name: hktest
  namespace: hive
spec:
  baseDomain: hive.mytests.io
  imageSetRef:
    name: openshift-v4.5.13
  installConfigSecretRef: 
    name: my-install-config-template
  platform:
    aws:
      credentialsSecretRef:
        name: global-aws-creds
      region: eu-west-1
  size: 1
```

Right now it works similar to supplying a installconfig directly to clusterdeployment, with the only exception that `baseDomain` and `metadata.name` is merged with info from the clusterpool (and pull-secret which is merged downstream already)
This should mean that hive _should_ be quite agnostic to the target cloud and install-config.yaml content.

### A couple of implementation details and questions

- Initially I'm disabling machinepools when installconfigtemplate is provided. An alternative would be to get the worker spec from the install-config.yaml template when unmarshaling, but this might throw up a question of long term support and version alignment as discussed in #1182 

- Initially I've placed InstallConfigTemplate struct and handling in pkg/clusterresource/installconfigtemplate.go. I'm not sure if this is the best location for this, but unsure what location would be most consistent?

- In this implementation I am still calling the cloudbuilder when a InstallConfigTemplate is provided, not sure if this is actually required for the 
functioning of the deployment controller or any downstream processes. I'm assuming so, as ClusterDeployment seems to need the platform section which cloudbuilder seems to provide.


hassenius:
@dgoodwin as discussed in #1182

dgoodwin (commented):
This looks pretty great to me. Couple thoughts, and I think @abhinavdahiya is planning to take a look as well.

rayashworth:
My team is in need of this PR as well.

hassenius:
@rayashworth it's encouraging to see that we're not the only ones needing this feature. I'm planning to finish implementing the input to the PR this week.

dgoodwin:
Let us know when ready and we'll jump back in on it, thanks so much for contributing this!

pkg/clusterresource/installconfigtemplate.go
```diff
@@ -0,0 +1,54 @@
+package clusterresource
+
+import (
+	"encoding/json"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// InstallConfigTemplate allows for overlaying generic InstallConfig with
```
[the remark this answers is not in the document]
hassenius: I've been looking into this a bit @abhinavdahiya , and it appears to me (if I understand things correctly) that the install-config.yaml is not quite valid for the unstructured workflow (At least I can't seem to get around the `while decoding JSON: Object 'Kind' is missing in...` problem). So with that  I've started looking at dissecting yqlib (which is not completely trivial at my go skill level), but if there's a consensus on the team that this is the route to go I'll continue down that route.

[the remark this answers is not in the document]
dgoodwin: I think we can defer this to future if it's proving non-trivial, and what is here now isn't inherently broken. It's an internal impl detail we can always swap later if needed.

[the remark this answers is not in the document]
hassenius: Thanks @dgoodwin, in that case I'll tidy things up and implement the unit test, shouldn't take too long....

hassenius:
I've added test cases mergeInstallConfigTemplate, some unit tests for json marshal/unmarshal (maybe not the most elegant, but easy for folks at my go-level to reason with).
Also added some doc to describe the new feature.

Note, `make test` fail at the moment, but doesn't seem to be associated with code added in this PR, as the same tests fail on main branch.

Should be good for review now.

pkg/clusterresource/installconfigtemplate_test.go
```diff
@@ -0,0 +1,69 @@
...
+	err := json.Unmarshal([]byte(fakeInstallConfigTemplateJSON), d)
+	require.NoError(t, err)
+
+	assert.Equal(t, d.BaseDomain, string("fake.domain"))
+	assert.Equal(t, d.MetaData.Name, string("my_own_cluster"))
+}
+
+func TestInstallConfigTemplateMarshalling(t *testing.T) {
+
+	{
```
dgoodwin, line 28: Use of blocks like this is new to me so it's possible I just need to be educated, but haven't seen it used in our team yet. Ideally I'd rather this stick with the standard sub-tests pattern as used up in builder_test.go.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -472,6 +473,13 @@ func (r *ReconcileClusterPool) addClusters(
 		errs = append(errs, fmt.Errorf("%s: %w", pullSecretDependent, err))
 	}
 
+	// Load install config templates if specified
+	installConfigTemplate, err := r.getInstallConfigTemplate(clp, logger)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("%s: %w", icSecretDependent, err))
+	}
+
+	// Not sure if we still need a cloudbuilder if we're getting a fully spec'ed install-config.yaml
```
dgoodwin, line 482: I think we do, for various other objects we generate in those builders, we're really just skipping one. Safe to drop this comment.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -509,17 +518,24 @@ func (r *ReconcileClusterPool) createCluster(
 	}
 	logger.WithField("cluster", ns.Name).Info("Creating new cluster")
 
+	// Skip machinepools if we're provided a complete installConfigTemplate
+	// as the machinepool will replace the workers specified in the
+	// installconfigtemplate
+	skipMachinePools := installConfigTemplate != ""
```
dgoodwin, line 524: I think this assumption should be removed, we accepted that a user might give us a worker pool def in install config, and then change that work pool def day 2, obviously it's a rare situation where someone would want to do that but we considered it a non-goal to prevent it. 

However I think users with install-configs still might want to manage their workers from Hive day 2, so there's no reason to assume to skip machine pools just because you used your own template. 

This *might* be boiling down to a need for worker instance type control on the hiveutil CLI at some point.

hassenius: If I remember correctly, when I was looking through this section of the code, skipMachinePools isn't really exposed anywhere. 
So if a user specifies a clusterpool with installconfigsecrettemplate, there is no way to include in the spec that machinepools should be disabled. The implication of this is that the machinepool will use the hard-coded worker spec, which will overwrite the definition in the templated install-config.yaml

I think there are three potential approaches:
1. Disable the machinepool when using install config template (current implementation)
2. Expose skipMachinePools in clusterpool spec
3. Unmarshal install-config template sections to Hive MachinePoolPlatform and populate the relevant node info from the install-config template (probably requires completion and merging of #1182 + some additional work)

For our particular use case we specifically want the owner of the cluster (i.e. the user who claims it) to manage the workers, not Hive, but maybe that's an edge-case in the big picture.

dgoodwin: That is a perfectly reasonable use case. I don't think use of install config implies it, so I would rule out 1, and I think option 2 is my preference. Could be a followup PR.

hassenius: Exposing this would work fine for me. I'm happy to do it as a follow up PR, so that I can start with a clean base.

I'd be doing it fairly swiftly as the lack of ability to disable machinepools will render this feature noop for us, as hive will reconcile the workers down to a size that is smaller than what we need in our clusters.

What's the best way to proceed?

dgoodwin: I'd rather hold this PR and lump that change in, than put this assumption in, and someone start using it only to get surprised because it changes shortly thereafter. If it's disruptive to merge this without the assumption, lets lump it into this PR.

hassenius: It was a very straight forward change. Updated it now.

dgoodwin:
/lgtm

Thanks for all this effort, sounds like it will be a popular feature.

hassenius:
Thanks for the patience and the guidance @dgoodwin

PR #1216, opened 2020-11-24
Fix regexes

dgoodwin:
The escape chars in the install log regex configmap led to it being unparseable, so all attempts to parse install logs are presently failing in stage and prod.

Also added the cluster platform and imageset to the metric labels to help us identify when it's a failing 4.7 nightly on gcp, which is where all our failures seem to be right now.

/assign @suhanime 
/cc @akhil-rane

### DIFF
--- a/pkg/controller/metrics/provision_underway_collector.go
+++ b/pkg/controller/metrics/provision_underway_collector.go
@@ -66,6 +66,12 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			}
 		}
 
+		platform := cd.Labels[hivev1.HiveClusterPlatformLabel]
+		imageSet := "none"
+		if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.ImageSetRef != nil {
+			imageSet = cd.Spec.Provisioning.ImageSetRef.Name
+		}
+
 		// For installing clusters we report the seconds since the cluster was created.
 		ch <- prometheus.MustNewConstMetric(
 			cc.metricClusterDeploymentProvisionUnderwaySeconds,
@@ -76,6 +82,8 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			GetClusterDeploymentType(&cd),
 			condition,
 			reason,
+			platform,
+			imageSet,
 		)
 
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -76,6 +82,8 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			GetClusterDeploymentType(&cd),
 			condition,
 			reason,
+			platform,
```
suhanime, line 85: We are hitting the [number 10](https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels) here - just a caution to keep in mind

pkg/operator/assets/bindata.go
```diff
@@ -1363,7 +1363,7 @@ data:
     # GCP Specific
     - name: GCPInvalidProjectID
       searchRegexStrings:
-      - "platform\.gcp\.project.* invalid project ID"
+      - "platform.gcp.project.* invalid project ID"
```
suhanime, line 1366: I'm curious - is there a way to test these?

suhanime: Also, this file belongs in the previous commit

pkg/controller/metrics/provision_underway_collector.go
dgoodwin: Fair point. This is probably abusing prometheus because we're not well positioned to run some kind of script on an alert that would gather this info for us, or a link to a page we could view for info. Platform probably is fine here as someone might want to route alerts depending on that. Image set probably is not great here... 

Do you think it's safe for now? 

We should definitely think about how we could improve in future.

pkg/operator/assets/bindata.go
dgoodwin: I try to test anything I'm not sure will work in the installlogmonitor tests, however in this case I included slashes here and not in the unit test. :(

I also realized this configmap is kinda misguided right now and filed this jira this morning: https://issues.redhat.com/browse/CO-1301 If we implement that change, and hardcode these regexes in our code, we could unit test them all.

pkg/controller/metrics/provision_underway_collector.go
suhanime: It is just a caution and/or a good practice. I think the alert for a failing nightly would help - we can also try and not log a metric if not available - since we don't have to bother cleaning up (for ex, say imageset is none, then we don't need prometheus to scrape the none, just don't report it)
Regardless, I think we are okay for now, we can revisit alternatives if we want to add more labels in the future

dgoodwin:
Fixed the commit containing something for the previous one. Thanks!

Issue #1206, opened 2020-11-11
How to recover a cluster "FailedToStart"?


tojosephyang:
There is a clusterclaim marked as "FailedToStart".
```
  conditions:
  - lastProbeTime: "2020-11-11T00:02:39Z"
    lastTransitionTime: "2020-11-11T00:00:05Z"
    message: "Failed to start machines: IncorrectInstanceState: The instance 'i-0af38ae20ac9715a7' is not in a state from which it can be started.\n\tstatus code: 400, request id: ada05e60-b2fe-4dbd-a7d1-824c19861d95"
    reason: FailedToStart
    status: "True"
    type: Hibernating
```
I tried to change its .spec.powerState to Hibernate/Running by "oc edit". Nothing changed. I can bring the cluster back by starting the cluster node EC2 in AWS console; however, changing .spec.powerState does not work any more to switch it to hibernate. 
Is there a way to recover this cluster back to normal?

gregsheremeta:
Sounds like the instance got into a bad state, and now hive can't recover?

Can you share any relevant logs?

cc @dgoodwin

tojosephyang:
I have removed that cluster already. 

To reproduce that is, hibernate a running cluster by following:
oc patch cd $1 --type='merge' -p $'spec:\n powerState: Hibernating' 

Before it hibnerate completely, manually resume it again. 
oc patch cd $1 --type='merge' -p $'spec:\n powerState: Running'                                                                                                                 

My work around is stop taking resume requests until the previous hibernation completed.

PR #1220, opened 2020-11-25
Fix resuming clusters that hit a FailedToStart issue.

dgoodwin:
In the event clusters fail to start (i.e. AWS instance capacity issues
we've seen the last few days), Hive will mark the Hibernation condition
reason FailedToStart, but then would stop all processing, there was no
way to get out of this state, or resume the cluster again.

This change adds the FailedToStart reason to the case where we continue
trying to start machines. If an error is encountered starting the
machines, we also now return this error from the controller, which gives
us the normal reconcile with backoff loop we want in a situation like
this.

In practice with the AWS capacity issues this week, this situation will
actually resolve itself fairly quickly if you keep trying.

/assign @twiest 
/cc @csrwng

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -249,7 +249,7 @@ func (r *hibernationReconciler) Reconcile(request reconcile.Request) (result rec
 			return reconcile.Result{}, nil
 		}
 		switch hibernatingCondition.Reason {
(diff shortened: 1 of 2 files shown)


dgoodwin:
Fixes: #1206

csrwng:
lgtm as well

dgoodwin:
Pushed an update which fixes the unit test and adds a fix for allowing you to re-hibernate if you are stuck in failedtostart.

PR #1222, opened 2020-12-03
Improve Scale Testing Simulation

dgoodwin:
This PR removes the global HiveConfig "fake all applies" annotation and instead moves to a per-ClusterDeployment annotation `hive.openshift.io/fake-cluster`. We no longer "adopt" one cluster thousands of times, no real cluster is required and the fake clusters will never be communicated with. With this new approach we do actually launch ClusterProvisions and installation Job/Pods, to better simulate the churn and load on etcd. We also get to see Kubernetes queue up and work through more install pods than we have capacity to run at once.

The new fakeHelper does bypass some CPU cycles for serializing resources that would be applied, however our primary concern lately is etcd load now that we have parallel processing of SyncSets.

/assign @abhinavdahiya 
/cc @gregsheremeta

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -295,13 +295,9 @@ const (
 	// InstallLogsAWSS3BucketEnvVar is the environment variable specifying the S3 bucket to use.
 	InstallLogsAWSS3BucketEnvVar = "HIVE_INSTALL_LOGS_AWS_S3_BUCKET"
 
-	// HiveFakeInstallKubeconfigSecretAnnotation is set to the name of a Secret in the cluster namespace which
-	// contains an admin kubeconfig.
-	// When set, this will enable a number of changes that are useful for simulating scale testing in a Hive cluster.
-	// Hive will launch an install pod that does not actually run the openshift-installer. Instead if will fake out
-	// some data and re-use this admin kubeconfig. Additionally, this annotation will disable all resource applies
-	// to that cluster to avoid overloading the one cluster.
-	HiveFakeInstallKubeconfigSecretAnnotation = "hive.openshift.io/fake-install-kubeconfig-secret"
+	// HiveFakeClusterAnnotation can be set to true on a cluster deployment to create a fake cluster that never
+	// provisions resources, and all communication with the cluster will be faked.
+	HiveFakeClusterAnnotation = "hive.openshift.io/fake-cluster"
 
 	// ReconcileIDLen is the length of the random strings we generate for contextual loggers in controller
 	// Reconcile functions.
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -299,6 +299,14 @@ const (
 	// fake out all apply operations for simulated scale testing.
 	HiveSyncSetsFakeApplyEnvVar = "HIVE_SYNCSETS_FAKE_APPLY"
 
+	// HiveFakeInstallKubeconfigSecretAnnotation is set to the name of a Secret in the cluster namespace which
+	// contains an admin kubeconfig.
+	// When set, this will enable a number of changes that are useful for simulating scale testing in a Hive cluster.
+	// Hive will launch an install pod that does not actually run the openshift-installer. Instead if will fake out
+	// some data and re-use this admin kubeconfig. Additionally, this annotation will disable all resource applies
+	// to that cluster to avoid overloading the one cluster.
+	HiveFakeInstallKubeconfigSecretAnnotation = "hive.openshift.io/fake-install-kubeconfig-secret"
+
 	// ReconcileIDLen is the length of the random strings we generate for contextual loggers in controller
 	// Reconcile functions.
 	ReconcileIDLen = 8
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -190,6 +190,10 @@ const (
 	// path where we mount in the SSH key for connecting to the bare metal libvirt provisioning host.
 	LibvirtSSHPrivKeyPathEnvVar = "LIBVIRT_SSH_PRIV_KEY_PATH"
 
+	// FakeClusterInstallEnvVar is the environment variable Hive will set for the installmanager pod to request
+	// a fake install.
+	FakeClusterInstallEnvVar = "FAKE_INSTALL"
+
 	// ControlPlaneCertificateSuffix is the suffix used when naming objects having to do control plane certificates.
 	ControlPlaneCertificateSuffix = "cp-certs"
 
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -295,10 +295,6 @@ const (
 	// InstallLogsAWSS3BucketEnvVar is the environment variable specifying the S3 bucket to use.
 	InstallLogsAWSS3BucketEnvVar = "HIVE_INSTALL_LOGS_AWS_S3_BUCKET"
 
-	// HiveSyncSetsFakeApplyEnvVar is set to true by the hive-operator when a developer has configured HiveConfig to
-	// fake out all apply operations for simulated scale testing.
-	HiveSyncSetsFakeApplyEnvVar = "HIVE_SYNCSETS_FAKE_APPLY"
-
 	// HiveFakeInstallKubeconfigSecretAnnotation is set to the name of a Secret in the cluster namespace which
 	// contains an admin kubeconfig.
 	// When set, this will enable a number of changes that are useful for simulating scale testing in a Hive cluster.
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -562,11 +562,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			}
 
 			if cd.Status.WebConsoleURL == "" || cd.Status.APIURL == "" {
-				if !controllerutils.IsFakeCluster(cd) {
-					return r.setClusterStatusURLs(cd, cdLog)
-				} else {
-					return r.setFakeClusterStatusURLs(cd, cdLog)
-				}
+				return r.setClusterStatusURLs(cd, cdLog)
 			}
 
 		}
@@ -1347,19 +1343,6 @@ func (r *ReconcileClusterDeployment) setClusterStatusURLs(cd *hivev1.ClusterDepl
 	}
 	cdLog.Debugf("read remote route object: %s", routeObject)
 	cd.Status.WebConsoleURL = "https://" + routeObject.Spec.Host
-
-	if err := r.Status().Update(context.TODO(), cd); err != nil {
-		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not set cluster status URLs")
-		return reconcile.Result{Requeue: true}, nil
-	}
-
-	return reconcile.Result{}, nil
-}
-
-// setFakeClusterStatusURLs skips cluster communication and sets fake cluster URLs in status.
-func (r *ReconcileClusterDeployment) setFakeClusterStatusURLs(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (reconcile.Result, error) {
-	cd.Status.APIURL = "https://example.com/veryfakeapi"
-	cd.Status.WebConsoleURL = "https://example.com/veryfakeconsole"
 
 	if err := r.Status().Update(context.TODO(), cd); err != nil {
 		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not set cluster status URLs")
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -562,7 +562,11 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			}
 
 			if cd.Status.WebConsoleURL == "" || cd.Status.APIURL == "" {
-				return r.setClusterStatusURLs(cd, cdLog)
+				if !controllerutils.IsFakeCluster(cd) {
+					return r.setClusterStatusURLs(cd, cdLog)
+				} else {
+					return r.setFakeClusterStatusURLs(cd, cdLog)
+				}
 			}
 
 		}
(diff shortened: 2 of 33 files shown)


pkg/resource/helper.go
```diff
@@ -68,11 +69,12 @@ func (r *helper) cacheOpenAPISchema() error {
 }
 
 // NewHelperFromRESTConfig returns a new object that allows apply and patch operations
-func NewHelperFromRESTConfig(restConfig *rest.Config, logger log.FieldLogger) (Helper, error) {
+func NewHelperFromRESTConfig(restConfig *rest.Config, fake bool, logger log.FieldLogger) (Helper, error) {
```
abhinavdahiya, line 72: imo, we should create a separate function `NewFakeHelperFromRESTConfig` . fake argument in new helper seems out of place, a separate function will make the intent and calling easier to understand.

pkg/resource/apply.go
```diff
@@ -66,7 +66,15 @@ func (r *helper) Apply(obj []byte) (ApplyResult, error) {
 	// fakeApply mode (used in scale testing only) doesn't run the apply. Instead it will Sleep for a period of time
 	// that matches our request time metrics in real hive environments to simulate i/o wait.
 	fakeAppliesEnvVar, _ := os.LookupEnv(constants.HiveSyncSetsFakeApplyEnvVar)
-	if fakeApplies, err := strconv.ParseBool(fakeAppliesEnvVar); fakeApplies && err == nil {
+	if fakeApplies, err := strconv.ParseBool(fakeAppliesEnvVar); r.fake || (fakeApplies && err == nil) {
```
abhinavdahiya: > (fakeApplies && err == nil)

it seems like we are accessing fakeApplies output before checking for err here, imo other way around would be better?

pkg/resource/apply.go
```diff
@@ -66,7 +66,15 @@ func (r *helper) Apply(obj []byte) (ApplyResult, error) {
...
+		r.fakeApply()
+		return ConfiguredApplyResult, nil
+	}
+	// When scale testing we create many clusters all using the same kubeconfig. We do not want to overwhelm it with
+	// PATCH requests, so instead we fake that operation here and sleep for a period of time that roughly matches
+	// request times we see in the real world.
+	// This links up to the ClusterDeployment annotation hive.openshift.io/fake-install-kubeconfig-secret
+	if r.fake {
```
abhinavdahiya, line 66: isn't this already handled in the if condition above at ln-69

pkg/installmanager/installmanager.go
```diff
@@ -368,8 +371,16 @@ func (m *InstallManager) Run() error {
...
+			if fakeCluster {
+				// If we're faking this install, it's safer to put an ID in that will never match any
+				// resource tags in the real world.
+				fakeClusterString := "fake cluster"
+				provision.Spec.InfraID = pointer.StringPtr(fakeClusterString)
+				provision.Spec.ClusterID = pointer.StringPtr(fakeClusterString)
+			} else {
+				provision.Spec.InfraID = pointer.StringPtr(metadata.InfraID)
+				provision.Spec.ClusterID = pointer.StringPtr(metadata.ClusterID)
+			}
```
abhinavdahiya, line 383: super not important:
A:
// set the most used path
// if fake, set the less taken code-path
B:
// if fake do this
// if not do the actual work

I personally like A .

pkg/installmanager/installmanager.go
```diff
@@ -398,54 +409,59 @@ func (m *InstallManager) Run() error {
 		}
 	}
 
-	installErr := m.provisionCluster()
-	if installErr != nil {
-		m.log.WithError(installErr).Error("error running openshift-install, running deprovision to clean up")
+	if fakeCluster {
+		m.log.Warnf("skipping openshift-install create cluster due to %s annotation on ClusterDeployment",
+			constants.HiveFakeInstallKubeconfigSecretAnnotation)
+	} else {
```
abhinavdahiya, line 415: super not important:
most of the actual work is in else block now, can we switch so that actual work is in first block.. just for readability.

abhinavdahiya (commented):
did first pass.

not loving the conditionals :P moving the actual business logic to the right, just makes things difficult to read for me.. but i know we can't really get away from that.

secondly the resource helper taking on fake seems like would be better to just write a new fake helper that does nothing and let the caller choose the right one to initialize. this might be easier to read when we move to faking out more things.

dgoodwin:
This just got a little more complicated. A fully fake helper means we're now skipping the CPU hit for some of the local apply logic and serialization, it may not be insignificant. This impacts the followup card I had to fake out all comms with the server, if we can't get a real openapi schema we can't run that serialization.

With the situation now where we can horizontally scale the syncset pods, I'm more worried about the load on etcd.  However this gets weird, if we're instantly returning, clustersyncs happen a lot faster, and thus we're hitting etcd a lot more often writing clustersync status. It could cause a chain reaction we don't necessarily want.

I guess the fake can do the sleeps, we're skipping some cpu hit for serializing, that wouldn't be so bad.

pkg/resource/helper.go
dgoodwin: That does sound better.

pkg/resource/apply.go
dgoodwin: Code now gone due to the fakeHelper refactor.

pkg/resource/apply.go
dgoodwin: Also now gone.

pkg/installmanager/installmanager.go
dgoodwin: Changed to if !fakeCluster do the normal, else do the fake.

pkg/installmanager/installmanager.go
dgoodwin: Done, although I'm worried I misunderstood your previous comment.

dgoodwin:
PR updated for feedback.

abhinavdahiya (commented):
/lgtm

dgoodwin:
Apologies I have the work ready for the other card, so decided to lump into this PR rather than push one annotation and then change it to another.

dgoodwin:
Updated again for fixes.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1352,6 +1356,19 @@ func (r *ReconcileClusterDeployment) setClusterStatusURLs(cd *hivev1.ClusterDepl
 	return reconcile.Result{}, nil
 }
 
+// setFakeClusterStatusURLs skips cluster communication and sets fake cluster URLs in status.
```
abhinavdahiya, line 1359: Do you think keeping these fake functions in the same file as actual code is good/ acceptable.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -388,7 +391,11 @@ func (r *ReconcileClusterSync) Reconcile(request reconcile.Request) (reconcile.R
 		logger.WithError(err).Error("unable to get REST config")
 		return reconcile.Result{}, err
 	}
-	resourceHelper, err := r.resourceHelperBuilder(restConfig, logger)
+
+	// If this cluster carries the fake annotation we will fake out all helper communication with it.
```
abhinavdahiya, line 395: Replace this with util function that was added in controllerutils?

pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -169,29 +169,35 @@ func (r *ReconcileRemoteMachineSet) Reconcile(request reconcile.Request) (reconc
...
-			// to this reconciliation, then attempt to connect to the remote cluster using the fallback API URL.
-			// Even when the controller continues to reconcile a ClusterDeployment waiting for the preferred API URL to
-			// become accessible, the controller should not recheck connectivity via the fallback API URL more often
-			// than once every 2 hours.
-			if connectivityRecheckNeeded || wasPrimaryActive {
-				if _, secondaryErr := remoteClientBuilder.UseSecondaryAPIURL().Build(); secondaryErr != nil {
-					cdLog.WithError(secondaryErr).Warn("unable to create remote API client with either the initial API URL or the API URL override, marking cluster unreachable")
-					unreachableError = utilerrors.NewAggregate([]error{primaryErr, secondaryErr})
+	var primaryErr error
+	fakeCluster := controllerutils.IsFakeCluster(cd)
```
abhinavdahiya, line 173: Couldn't see this being used outside of the if condition

abhinavdahiya:
Looking at all the places that now have conditional blocks I was thinking if we could replace these with some interface and its implementations?

InstallPodManager => actual impl , faked out ...

I’m just thinking because there is quite a bit of actual code is now in this conditions.. which feels very weird to me..

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: I will move to a fake.go.

pkg/controller/unreachable/unreachable_controller.go
dgoodwin: > Looking at all the places that now have conditional blocks I was thinking if we could replace these with some interface and its implementations?
> 
> InstallPodManager => actual impl , faked out ...
> I’m just thinking because there is quite a bit of actual code is now in this conditions.. which feels very weird to me..

This would be nice, I've toyed with this idea for places we fork based on cloud provider in the past as well though it wasn't super popular at the time.

The install manager part is tough, I want a huge portion of the normal install manager to run, and just skip on some very small pieces. That one in particular would be tough to push out to a fake interface. We skip/fake upload of admin pw, cluster ID, infra ID, and the execution of the create cluster command as well as log upload afterwards. I don't want to fake the entire installmanager for that. I guess we could use a little composition, one struct for uploading info, another for executing the create cluster command? WDYT, worth it?

Throughout the controllers we could potentially setup some kind of GetClusterAdapter(cd), and that adapter could expose something like SetClusterStatusURLs, GetClusterOperators, GetHelperForRestConfig, GetClusterVersion, GetMachineSets, Unreachable. Actually all of those boil down close to that remoteclient.ConnectToRemoteCluster function, perhaps that could check the annotation and return a fake client like we use in unit tests, with the various hard coded data we'd want to return for that cluster. That might work... 

> Also, if we do go about this then our fake and real implementation are 2 different things, being maintained separately... soare we really faking the real thing?

We definitely need to be very careful with that. In this PR alone we're losing that serialization hit, though we could probably restore it. 

> 
> Secondly, why do we want a per cd annotation to fake clusters?
> 
>     * I’m not sure what workflow would benefit from having real cds and faked cds together being acted upon
> 
>     * also this makes it so actual users can do this for their cds which I don’t think is appropriate..
>       Is is too bad to keep hive configuration that you in fake mode and fake out all you can?

The stage test workflow would benefit from this and is the primary driver for the work. We don't have fully dedicated clusters for stage testing and instead we run those tests in existing clusters, which do contain some active clusters. It makes it simpler for a test suite driving that test to be able to control the behavior for it's own clusterdeployments, without requiring an admin or CI change to flip global hiveconfig. The latter was ok for us on the dev side but there's a lot of interest in this on SD side to be able to try to replicate the load on etcd, and api, with OCM and SRE operators in the mix, without running up massive bills and having to rush to finish and clean up.

With regard to users being able to use this, this is why I go with annotations rather than anything documented and commited to in the actual API Spec. We do this in a couple places to let users simulate things that we don't necessarily want to draw attention to or commit to.

dgoodwin:
Commits pushed to help isolate the fake code more.

pkg/constants/constants.go
```diff
@@ -295,9 +299,9 @@ const (
 	// InstallLogsAWSS3BucketEnvVar is the environment variable specifying the S3 bucket to use.
 	InstallLogsAWSS3BucketEnvVar = "HIVE_INSTALL_LOGS_AWS_S3_BUCKET"
 
-	// HiveSyncSetsFakeApplyEnvVar is set to true by the hive-operator when a developer has configured HiveConfig to
-	// fake out all apply operations for simulated scale testing.
-	HiveSyncSetsFakeApplyEnvVar = "HIVE_SYNCSETS_FAKE_APPLY"
+	// HiveFakeClusterAnnotation can be set to true on a cluster deployment to create a fake cluster that will
+	// launch an install job that never provisions resources, or tries to communicate with the cluster.
```
abhinavdahiya, line 303: > that never provisions resources

it's at least clear to that this means that the results will be positive i.e. the object will say installed but actuall no cluster was created..

dgoodwin: Updated.

pkg/installmanager/fake.go
```diff
@@ -0,0 +1,43 @@
...
+	"encoding/json"
+	"fmt"
+
+	installertypes "github.com/openshift/installer/pkg/types"
+	"github.com/pkg/errors"
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+)
+
+const fakeMetadata = `{"aws":{"identifier":[{"kubernetes.io/cluster/fake-infraid":"owned"},{"openshiftClusterID":"fake-cluster-id"}],"region":"us-east-1"},"clusterID":"fake-cluster-id","clusterName":"%s","infraID":"fake-infra-id"}`
```
abhinavdahiya, line 13: since this is supposed to be used as format string, i would name is accordingly.

pkg/installmanager/fake.go
```diff
@@ -0,0 +1,43 @@
...
+	// resources.
+	md := &installertypes.ClusterMetadata{}
+	if err := json.Unmarshal(metadataBytes, md); err != nil {
+		m.log.WithError(err).Error("error unmarshalling cluster metadata")
+		return nil, nil, err
+	}
+	if md.InfraID == "" {
+		m.log.Error("cluster metadata did not contain infraID")
+		return nil, nil, errors.New("cluster metadata did not contain infraID")
+	}
```
abhinavdahiya, line 35: is this required here? the fake metadata string is a constant..?

pkg/remoteclient/fake.go
```diff
@@ -0,0 +1,142 @@
...
+// Build returns a fake controller-runtime test client populated with the resources we expect to query for a
+// fake cluster.
+func (b *fakeBuilder) Build() (client.Client, error) {
+	scheme, err := machineapi.SchemeBuilder.Build()
+	if err != nil {
+		return nil, err
+	}
+
+	autoscalingv1.SchemeBuilder.AddToScheme(scheme)
+	autoscalingv1beta1.SchemeBuilder.AddToScheme(scheme)
```
abhinavdahiya, line 44: This vs
```
	scheme := runtime.NewScheme()
	hivev1.AddToScheme(scheme)
	rbacv1.AddToScheme(scheme)
```

what's the difference? the latter has no error requirement..

pkg/remoteclient/fake.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if err := openshiftapiv1.Install(scheme); err != nil {
+		return nil, err
+	}
+
+	if err := routev1.Install(scheme); err != nil {
+		return nil, err
+	}
```
abhinavdahiya, line 52: some AddToScheme has error handling, while the previous one's don't any reason why in particular?

pkg/remoteclient/fake.go
```diff
@@ -0,0 +1,142 @@
...
+
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+)
+
+// fakeBuilder builds fake clients for fake clusters. Used to simulate communication with a cluster
+// that doesn't actually exist in scale testing.
+type fakeBuilder struct {
+	c              client.Client
+	cd             *hivev1.ClusterDeployment
+	controllerName hivev1.ControllerName
```
abhinavdahiya, line 31: didn't see these fields being used in this file? Why are these required?

pkg/remoteclient/fake.go
dgoodwin: Old code coming over from the regular builder creation, I'll see if I can clean them up.

pkg/remoteclient/fake.go
dgoodwin: Those AddToSchemes still return errors. Seems like we should always check those errors as opposed to some of these that were not. I'm going to move them to common func with the regular builder, and add error checking, otherwise probably leave them be.

pkg/remoteclient/fake.go
dgoodwin: New commit added for this, a couple don't appear to have an Install function only AddToScheme. Added error handling for all.

dgoodwin:
@abhinavdahiya ready for another look here.

abhinavdahiya (approved):
/lgtm

PR #1223, opened 2020-12-04
clusterpool: allow setting max size and max concurrent actions

abhinavdahiya:
xref: https://issues.redhat.com/browse/CO-1278

Today the cluster pool keeps creating clusters to match `.spec.size` as ready clusters
are claimed by users. This is usually not possible as the secret used by the cluster pool usually
maps to an account in cloud that has limits/quotas and therefore there is need to allow admin
to set a limit on the maximum number of clusters created by a pool (already claimed + unclaimed).
The new field `.spec.maxSize` allows the admin to set this limit on maximum number of clusters
that are created for a pool. This includes the already claimed clusters, clusters ready to be claimed and
clusters that being provisioned, and does not include clusters are being deleted.
NOTE: when unset, this implies Inf capacity.

Similar to quotas in accounts, most cloud providers have API quotas that limit number of cloud API interactions
over a period of time. An close approximation of this can be made in terms of number of clusters being brought
up/down as these are periods of high cloud API interactions. So if the admin can define a limit on the
number of UP/DOWN activity that can happen for a pool, it allows the admin to prevent rate limiting.
The new field `.spec.maxConcurrent` specifies the maximum number of cluster create/delete can be in progress at
a given instant.
NOTE: when unset, this implies Inf concurrent limit.

### DIFF
--- a/pkg/apis/hive/v1/clusterpool_types.go
+++ b/pkg/apis/hive/v1/clusterpool_types.go
@@ -21,6 +21,17 @@ type ClusterPoolSpec struct {
 	// +required
 	Size int32 `json:"size"`
 
+	// MaxSize is the maximum number of clusters that will be provisioned including clusters that have been claimed
+	// and ones waiting to be used.
+	// By default there is no limit.
+	// +optional
+	MaxSize *int32 `json:"maxSize,omitempty"`
+
+	// MaxConcurrent is the maximum number of clusters that will be provisioned or deprovisioned at an time.
+	// By default there is no limit.
+	// +optional
+	MaxConcurrent *int32 `json:"maxConcurrent,omitempty"`
+
 	// BaseDomain is the base domain to use for all clusters created in this pool.
 	// +required
 	BaseDomain string `json:"baseDomain"`
@@ -72,9 +83,12 @@ type ClusterPoolCondition struct {
 type ClusterPoolConditionType string
 
 const (
-	// ClusterPoolMissingDependentsCondition is set when a cluster pool is missing dependencies required to create a
+	// ClusterPoolMissingDependenciesCondition is set when a cluster pool is missing dependencies required to create a
 	// cluster. Dependencies include resources such as the ClusterImageSet and the credentials Secret.
 	ClusterPoolMissingDependenciesCondition ClusterPoolConditionType = "MissingDependencies"
+	// ClusterPoolCapacityAvailableCondition is set to provide information on whether the cluster pool has capacity
+	// available to create more clusters for the pool.
+	ClusterPoolCapacityAvailableCondition ClusterPoolConditionType = "CapacityAvailable"
 )
 
 // +genclient
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -237,7 +237,7 @@ func (r *ReconcileClusterPool) Reconcile(request reconcile.Request) (reconcile.R
 	}
 
 	// Find all ClusterDeployments from this pool:
-	poolCDs, err := r.getAllUnclaimedClusterDeployments(clp, logger)
+	claimedCDs, unClaminedCDs, err := r.getAllClusterDeploymentsForPool(clp, logger)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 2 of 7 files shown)


pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -650,6 +698,35 @@ func (r *ReconcileClusterPool) setMissingDependenciesCondition(pool *hivev1.Clus
...
 
+func (r *ReconcileClusterPool) setAvailableCapacityCondition(pool *hivev1.ClusterPool, available bool, logger log.FieldLogger) error {
+	status := corev1.ConditionFalse
+	reason := "Available"
+	message := "There is capacity to add more clusters to the pool."
+	updateConditionCheck := controllerutils.UpdateConditionNever
+	if !available {
+		status = corev1.ConditionTrue
+		reason = "NoMoreCapacity"
+		message = fmt.Sprintf("There is no more capacity available for the pool. Additional capacity will be required to fulfil the requirements.")
```
dgoodwin, line 709: s/fulfil/fulfill but would suggest maybe a little more concise: "Pool is at maximum capacity of %d waiting and claimed clusters."

pkg/apis/hive/v1/clusterpool_types.go
```diff
@@ -72,9 +83,12 @@ type ClusterPoolCondition struct {
 type ClusterPoolConditionType string
 
 const (
-	// ClusterPoolMissingDependentsCondition is set when a cluster pool is missing dependencies required to create a
+	// ClusterPoolMissingDependenciesCondition is set when a cluster pool is missing dependencies required to create a
 	// cluster. Dependencies include resources such as the ClusterImageSet and the credentials Secret.
 	ClusterPoolMissingDependenciesCondition ClusterPoolConditionType = "MissingDependencies"
+	// ClusterPoolCapacityUnAvailableCondition is set when cluster pool does not have capacity
+	// available to create more clusters for the pool.
+	ClusterPoolCapacityUnAvailableCondition ClusterPoolConditionType = "CapacityUnAvailable"
```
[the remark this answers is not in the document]
dgoodwin: Actually I can't imagine anyone would ever oc wait for CapcityUnAvailable = true, why don't we invert this and use CapacityAvailable?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -650,6 +698,35 @@ func (r *ReconcileClusterPool) setMissingDependenciesCondition(pool *hivev1.Clus
...
 }
 
+func (r *ReconcileClusterPool) setAvailableCapacityCondition(pool *hivev1.ClusterPool, available bool, logger log.FieldLogger) error {
+	status := corev1.ConditionFalse
+	reason := "Available"
+	message := "There is capacity to add more clusters to the pool."
+	updateConditionCheck := controllerutils.UpdateConditionNever
+	if !available {
+		status = corev1.ConditionTrue
+		reason = "NoMoreCapacity"
```
dgoodwin, line 708: Suggest "MaxCapacity"

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -650,6 +698,35 @@ func (r *ReconcileClusterPool) setMissingDependenciesCondition(pool *hivev1.Clus
...
+		status,
+		reason,
+		message,
+		updateConditionCheck,
+	)
+	if changed {
+		pool.Status.Conditions = conds
+		if err := r.Status().Update(context.Background(), pool); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "could not update ClusterPool conditions")
+			return fmt.Errorf("could not update ClusterPool conditions: %w", err)
```
dgoodwin, line 724: We're inconsistent with this but maybe good to use errors.Wrap here?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -288,14 +297,48 @@ func (r *ReconcileClusterPool) Reconcile(request reconcile.Request) (reconcile.R
...
+		if err := r.deleteExcessClusters(installingCDs, readyCDs, toDel, logger); err != nil {
 			return reconcile.Result{}, err
 		}
 	// If too few, create new InstallConfig and ClusterDeployment.
 	case drift < 0:
-		if err := r.addClusters(clp, -drift, logger); err != nil {
+		toAdd := -drift
+		if !hasAvailableCapacity {
+			logger.WithFields(log.Fields{
+				"InUse":    clp.Status.Size,
```
dgoodwin, line 323: InUse seems misleading here, my understanding was Status.Size is the number of clusters currently in the pool, and it would exclude claimed clusters. Suggest CurrentPoolSize?

It might be good to move this to line 278 and include all the details, capacity, poolSize, claimedSize.

dgoodwin (commented):
Looks great!

PR #1229, opened 2020-12-10
clusterpool: include deleting claimed clusters for max concurrent

abhinavdahiya:
The cluster pool 's max concurrent will include the deleting claimed clsuters. This is
to make sure cluster pool actions are safe and under limits as specified by the admin.

previously the claimed clusters from a pool, were deleted out band directly by the clusterclaim
controller when the claim was deleted. This made controlling the max concurrent provision/
deprovision difficult because 2 controllers are making the same decision making independently.
To enforce the max concurrent for pool, all the actions of cluster provision/deprovision had to
be moved to clusterpool controller so that it can ensure the max concurrent limit.

Thefore, the clusterclaim controller now annotates the clusterdeployment that it should be removed
when the claim is deleted, it still removes all the other claim specific resources from the the
namespace like rolebindings etc, and removes the finalizer. The clusterpool controller removes
this previously claimed cluster whenever there is budget to do so.

Since clusters in a pool that have been claimed can live after the cluster pool has been deleted,
we need to make sure these clusters are also cleaned up when the claims are deleted. To fulfil that
the clusterpoolnamespace controller will cleanup up all clusters that belonged to a clusterpool that
no longer exists and have been annotated by the claim as safe to remove.

xref: https://issues.redhat.com/browse/CO-1278

/assign @dgoodwin 
/cc @abutcher

### DIFF
--- a/pkg/controller/clusterclaim/clusterclaim_controller.go
+++ b/pkg/controller/clusterclaim/clusterclaim_controller.go
@@ -22,6 +22,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
 	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/resource"
@@ -306,10 +307,15 @@ func (r *ReconcileClusterClaim) cleanupResources(claim *hivev1.ClusterClaim, log
 	}
 
 	// Delete ClusterDeployment
-	if cd.DeletionTimestamp == nil {
+	_, toRemove := cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation]
+	if cd.DeletionTimestamp == nil && !toRemove {
 		logger.Info("deleting clusterDeployment")
-		if err := r.Delete(context.Background(), cd); err != nil {
-			logger.WithError(err).Log(controllerutils.LogLevel(err), "error deleting ClusterDeployment")
+		if cd.Annotations == nil {
+			cd.Annotations = map[string]string{}
+		}
+		cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation] = ""
+		if err := r.Update(context.Background(), cd); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "error updating ClusterDeployment to mark it for deletion")
 			return err
 		}
 	}
--- a/pkg/controller/clusterclaim/clusterclaim_controller.go
+++ b/pkg/controller/clusterclaim/clusterclaim_controller.go
@@ -307,13 +307,13 @@ func (r *ReconcileClusterClaim) cleanupResources(claim *hivev1.ClusterClaim, log
 	}
 
 	// Delete ClusterDeployment
-	_, toRemove := cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation]
+	toRemove := controllerutils.IsClaimedClusterMarkedForRemoval(cd)
 	if cd.DeletionTimestamp == nil && !toRemove {
 		logger.Info("deleting clusterDeployment")
 		if cd.Annotations == nil {
 			cd.Annotations = map[string]string{}
 		}
-		cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation] = ""
+		cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation] = "true"
 		if err := r.Update(context.Background(), cd); err != nil {
 			logger.WithError(err).Log(controllerutils.LogLevel(err), "error updating ClusterDeployment to mark it for deletion")
 			return err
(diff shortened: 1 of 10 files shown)


abhinavdahiya:
/hold
I'm not sure convinced this is the best path forward.

abhinavdahiya:
/hold cancel

@dgoodwin this implements the proposed solution that moves deletes of previously claimed clusters to clusterpool controller. This should allow us to enforce the max concurrent for unclaimed + claimed clusters.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -303,44 +316,45 @@ func (r *ReconcileClusterPool) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	maxConcurrent := math.MaxInt32
```
dgoodwin, line 319: Var needs a rename, concurrentCapacity?

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -608,6 +641,233 @@ func TestReconcileClusterPool(t *testing.T) {
...
+				cdBuilder("c4").Build(
+					testcd.WithClusterPoolReference(testNamespace, testLeasePoolName, "test-claim"),
+				),
+			},
+			expectedTotalClusters: 4,
+			expectedObservedSize:  3,
+			expectedObservedReady: 2,
+		},
+		{
+			name: "do not delete previously claimed clusters 2",
```
dgoodwin, line 660: Rather than appending a number I would prefer using the test name to note the specifics of that test vs the others.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -608,6 +641,233 @@ func TestReconcileClusterPool(t *testing.T) {
...
+						testcd.WithClusterPoolReference(testNamespace, testLeasePoolName, "test-claim"),
+					),
+			},
+			expectedTotalClusters:   2,
+			expectedObservedSize:    2,
+			expectedObservedReady:   1,
+			expectedDeletedClusters: []string{"c4"},
+		},
+		{
+			name: "scale up should include previouly deleted in max concurrent 2",
```
dgoodwin, line 764: previously (couple locations in here)

pkg/controller/clusterpoolnamespace/clusterpoolnamespace_controller.go
```diff
@@ -147,6 +147,18 @@ func (r *ReconcileClusterPoolNamespace) Reconcile(request reconcile.Request) (re
...
+	switch {
+	case someCleanupDone && err != nil:
+		logger.WithError(err).Log(controllerutils.LogLevel(err), "error cleaning up some clusters")
+		return reconcile.Result{RequeueAfter: durationBetweenDeletingClusterDeploymentChecks}, err
+	case err != nil:
+		logger.WithError(err).Log(controllerutils.LogLevel(err), "error cleaning up some clusters")
+		return reconcile.Result{}, err
+	case someCleanupDone:
+		return reconcile.Result{RequeueAfter: durationBetweenDeletingClusterDeploymentChecks}, nil
+	}
```
dgoodwin, line 160: Nice catch.

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -306,10 +307,15 @@ func (r *ReconcileClusterClaim) cleanupResources(claim *hivev1.ClusterClaim, log
...
-	if cd.DeletionTimestamp == nil {
+	_, toRemove := cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation]
+	if cd.DeletionTimestamp == nil && !toRemove {
 		logger.Info("deleting clusterDeployment")
-		if err := r.Delete(context.Background(), cd); err != nil {
-			logger.WithError(err).Log(controllerutils.LogLevel(err), "error deleting ClusterDeployment")
+		if cd.Annotations == nil {
+			cd.Annotations = map[string]string{}
+		}
+		cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation] = ""
```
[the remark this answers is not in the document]
abhinavdahiya: I don't think a true/false value is valid here for this annotation.. since the presence provides the necessary information. true/false only adds additional error handling and possible break when someone sets some weird value here.

i'll change it to match the existing code.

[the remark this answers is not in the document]
dgoodwin: I agree but I've been nacked for it in the past. :)

abhinavdahiya:
pushed fixes for the previous comments.

pkg/controller/utils/clusterclaim.go
```diff
@@ -0,0 +1,20 @@
...
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
+)
+
+// IsClaimedClusterMarkedForRemoval returns true when hive.openshift.io/remove-claimed-cluster-from-pool
+// annotation is set to true value in the clusterdeployment
+func IsClaimedClusterMarkedForRemoval(cd *hivev1.ClusterDeployment) bool {
+	toRemove := false
+	if v, ok := cd.Annotations[constants.ClusterClaimRemoveClusterAnnotation]; ok && v != "" {
+		if b, err := strconv.ParseBool(v); err == nil {
```
dgoodwin, line 15: Can probably be a little simpler if you need to do this in future, ParseBool returns false, error on "", so it can be just one line I think, but not necessary to change.

abhinavdahiya:
@dgoodwin udpated to include the crd generation. PTAL

/hold cancel

PR #1231, opened 2020-12-14
[DOC] Replace SyncsetInstance with ClusterSync in troubleshooting

abutcher:
/assign @akhil-rane 
/cc @dgoodwin


akhil-rane (commented):
/lgtm

PR #1233, opened 2020-12-15
Add documentation for time-based scaling of clusterpool

akhil-rane:
jira: https://issues.redhat.com/browse/CO-1280

### DIFF
--- a/docs/clusterpools.md
+++ b/docs/clusterpools.md
@@ -147,3 +147,105 @@ spec:
 ```
 
 **Note** When you use installConfigSecretTemplate you will most likely want to disable MachinePools, so that Hive does not reconcile away from the machine config specified in install-config.yaml
+
+## Time-based scaling of Cluster Pool
+
+You can use kubenetes cron jobs to scale clusterpools as per a defined schedule.
+
+The following are the yaml configurations for setting up the permissions: Role, RoleBinding and ServiceAccount. It sets up a role with permissions to get a clusterpool and patch clusterpool’s scale subresource.
+
+```yaml
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  namespace: hive
+  name: scale-clusterpool
+rules:
+- apiGroups:
+  - hive.openshift.io
+  resources:
+  - clusterpools
+  verbs:
+  - get
+- apiGroups:
+  - hive.openshift.io
+  resources:
+  - clusterpools/scale
+  verbs:
+  - patch
+
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: scale-clusterpool
+  namespace: hive
+subjects:
+- kind: ServiceAccount
+  name: sa-scale-clusterpool
+  namespace: hive
+roleRef:
+  kind: Role
+  name: scale-clusterpool
+  apiGroup: rbac.authorization.k8s.io
+
+---
+apiVersion: v1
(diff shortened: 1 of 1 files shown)


twiest:
Looks great!

/lgtm

docs/clusterpools.md
```diff
@@ -147,3 +147,81 @@ spec:
...
+  namespace: hive
+
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: scale-clusterpools
+  namespace: hive
+spec:
+  schedule: "45 23 * * 5"
```
abhinavdahiya, line 205: add a comment that describes this schedule. people can easily look up but i think having that is more readable.

docs/clusterpools.md
```diff
@@ -147,3 +147,81 @@ spec:
...
+      template:
+        spec:
+          serviceAccountName: sa-scale-clusterpool
+          containers:
+          - name: scale-clusterpools-size-0
+            image: quay.io/openshift/origin-cli:latest
+            command:
+            - /bin/sh 
+            - -c 
+            - oc scale cp pool-1 -n hive --replicas=0
```
abhinavdahiya, line 217: `oc scale cp` i personally would recommend we use full forms for documentation. `cp` is too generic for very body to immediately understand.

docs/clusterpools.md
```diff
@@ -147,3 +147,81 @@ spec:
 ```
 
 **Note** When you use installConfigSecretTemplate you will most likely want to disable MachinePools, so that Hive does not reconcile away from the machine config specified in install-config.yaml
+
+## Time-based scaling of Cluster Pool
+
+You can use kubenetes cron jobs to scale clusterpools as per a defined schedule. Following are the sample yaml configurations to scale a clusterpool to size 0 at 23:45 (11:45 PM) every Friday. 
+
+```yaml
```
abhinavdahiya, line 155: split these up into 2 sections, 

one set of yaml is for setting up the rbac,
and the next set is examples of how to scale up and down based on time. <== for this one let's pick up a real world example, like 6 am (some timezone) everyday run scale up to X and 8 pm same day scale down to Y. I think an example like that conveys more...

akhil-rane:
@abhinavdahiya Updated the PR to address your feedback.

docs/clusterpools.md
```diff
@@ -147,3 +147,105 @@ spec:
 ```
 
 **Note** When you use installConfigSecretTemplate you will most likely want to disable MachinePools, so that Hive does not reconcile away from the machine config specified in install-config.yaml
+
+## Time-based scaling of Cluster Pool
+
+You can use kubenetes cron jobs to scale clusterpools as per a defined schedule.
```
staebler, line 153:
```suggestion
You can use kubernetes cron jobs to scale clusterpools as per a defined schedule.
```

docs/clusterpools.md
```diff
@@ -147,3 +147,105 @@ spec:
...
+      template:
+        spec:
+          serviceAccountName: sa-scale-clusterpool
+          containers:
+          - name: scale-clusterpool-size-10
+            image: quay.io/openshift/origin-cli:latest
+            command:
+            - /bin/sh 
+            - -c 
+            - oc scale clusterpool pool-1 -n hive --replicas=10
```
staebler, line 221: I think that it is easier to follow along with examples if we keep names consistent. I see three different names used for example ClusterPools--`openshift-46-aws-us-east-1`, `testpool`, and now `pool-1`.

PR #1234, opened 2020-12-17
Allow public subnets in AWS MachinePool

suhanime:
To be consistent with install config, allow public subnets to be
specified in MachinePool even if the MachineSets created would use only
private subnets.

Account for success in all of the following cases:
- No subnets specified
- Only private subnets specified
- Both public and private subnets specified

Private subnets are required to be exactly one per availability zone if
subnets are specified. If public subnets are also specified, then there
should be exactly 1 public and 1 private subnet for each availability
zone.

### DIFF
--- a/config/crds/hive.openshift.io_machinepools.yaml
+++ b/config/crds/hive.openshift.io_machinepools.yaml
@@ -117,8 +117,10 @@ spec:
                       type: object
                     subnets:
                       description: Subnets is the list of subnets to which to attach
-                        the machines. There must be exactly one subnet for each availability
-                        zone used.
+                        the machines. There must be exactly one private subnet for
+                        each availability zone used. If public subnets are specified,
+                        there must be exactly one private and one public subnet specified
+                        for each availability zone.
                       items:
                         type: string
                       type: array
--- a/pkg/controller/remotemachineset/awsactuator.go
+++ b/pkg/controller/remotemachineset/awsactuator.go
@@ -158,7 +158,7 @@ func (a *AWSActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 	subnets := map[string]string{}
 	// Fetching private subnets from the machinepool and then mapping availability zones to subnets
 	if len(pool.Spec.Platform.AWS.Subnets) > 0 {
-		subnetsByAvailabilityZone, err := a.getSubnetsByAvailabilityZone(pool)
+		subnetsByAvailabilityZone, err := a.getPrivateSubnetsByAvailabilityZone(pool)
 		if err != nil {
 			return nil, false, errors.Wrap(err, "describing subnets")
 		}
@@ -313,15 +313,15 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
 
 }
 
-// getSubnetsByAvailabilityZones maps availability zones to subnet
-func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (map[string]string, error) {
+// getPrivateSubnetsByAvailabilityZones maps availability zones to private subnet
+func (a *AWSActuator) getPrivateSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (map[string]string, error) {
 	idPointers := make([]*string, len(pool.Spec.Platform.AWS.Subnets))
 	for i, id := range pool.Spec.Platform.AWS.Subnets {
 		idPointers[i] = aws.String(id)
 	}
 
 	results, err := a.awsClient.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
-	if err != nil {
+	if err != nil || len(results.Subnets) == 0 {
 		if strings.Contains(err.Error(), "InvalidSubnetID.NotFound") {
 			var conditionMessage string
 			if submatches := reg.FindStringSubmatch(err.Error()); submatches != nil {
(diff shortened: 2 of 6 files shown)


pkg/controller/remotemachineset/awsactuator_test.go
```diff
@@ -158,7 +159,9 @@ func TestAWSActuator(t *testing.T) {
 				}(),
 			},
 			mockAWSClient: func(client *mockaws.MockClient) {
-				mockDescribeSubnets(client, []string{"zone1", "zone1", "zone2"}, []string{"subnet-zone1", "subnet-zone2", "subnet-zone3"})
+				mockDescribeSubnets(client, []string{"zone1", "zone1", "zone2"},
+					[]string{"subnet-zone1", "subnet-zone2", "subnet-zone3"}, "vpc-1")
+			// What's a neat way of creating a mock client with both DescribeSubnets and DescribeRouteTablesPages here?
```
dgoodwin, line 164: If I'm understanding the issue, it should simple be creating a mockRouteTables function that accepts the client reference as well, and calling it here exactly as you do for mockDescribeSubnets. This mockAWSClient is a function reference the test will call to build the mock client, but you can have additional statements to configure that client as you wish. See dnszone_controller_test.go for an example.

suhanime: Yes - writing up the mock function is my real problem :/
https://github.com/openshift/hive/pull/1234/files#diff-3686c203cc4ff7e788327e6a9261edccd61dd438114c0e389a975d567f760143R426-R439

dgoodwin: Apologies I missed that other comment.

One thought, do you need the Pages variant or would DescribeRouteTables be ok?

Or, perhaps even better, we do have one other client call using pages, and it is mocked here: https://github.com/openshift/hive/blob/de7dd1d6a81e0c48360aa4e52bb29c64d1d9f8a0/pkg/controller/dnszone/awsactuator_test.go#L178

Do either of those help?

suhanime: Hmmm I suppose DescribeRouteTables should be okay. Installer uses the Pages variant for some reason even though the code they imported from upstream Kube uses the simpler describe (https://github.com/kubernetes/kubernetes/blob/9f036cd43d35a9c41d7ac4ca82398a6d0bef957b/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3302)
I'm guessing the pages variant will be better in the case where there are many routetables? Perhaps I shouldn't be needlessly complicating things... Thanks!

config/crds/hive.openshift.io_machinepools.yaml
```diff
@@ -117,8 +117,8 @@ spec:
                       type: object
                     subnets:
                       description: Subnets is the list of subnets to which to attach
```
abhinavdahiya, line 119: additionally we should enforce that there is also only one public subnet per private subnet 
- this makes sure there is one pair of public+private subnet per AZ.
- no additional public subnets in some AZs that cluster does not already have a private subnet.
- there is also a requirement for the subnet to have some type of tags to make sure kube sees these subnets correctly.

see https://github.com/openshift/enhancements/blob/master/enhancements/installer/aws-customer-provided-subnets.md#subnets

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,9 +348,36 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
 		return nil, err
 	}
 
-	conflictingSubnets := sets.NewString()
-	subnetsByAvailabilityZone := make(map[string]string, len(pool.Spec.Platform.AWS.Subnets))
+	// Fetch private subnets from the provided list
+	vpc := *results.Subnets[0].VpcId
```
abhinavdahiya, line 352: hmm.. i don't see a check before that makes sure the slice has atleast one item. we should probably add a check to be sure.

abhinavdahiya (commented):
i think we should add one more constraint on the public subnets.

pkg/controller/remotemachineset/awsactuator.go
suhanime: This is method (getPrivateSubnetsByAvailabilityZone) is called only when Subnets are specified - https://github.com/openshift/hive/pull/1234/files/4f53e3540d76d08435e5852d52dad266607db9dc#diff-3d3ba26bc232b893654e5bc0fb8bacb63a11a8c559cf415413d9b837c974d8d5L160-L162

suhanime: ... On second thought, an additional check wouldn't hurt - especially if kube can't read the subnets  as you described above

config/crds/hive.openshift.io_machinepools.yaml
suhanime: Added. Haven't added the check for the expected tag - kube controller does that, however have included a check if subnets cannot be fetched by describe subnets.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,34 +351,62 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
+		if isPublic {
+			publicSubnets[*subnet.SubnetId] = *subnet
+		} else {
+			privateSubnets[*subnet.SubnetId] = *subnet
 		}
-		subnetsByAvailabilityZone[*subnet.AvailabilityZone] = *subnet.SubnetId
 	}
 
-	if len(conflictingSubnets) > 0 {
+	if len(publicSubnets) > 0 {
```
akhil-rane, line 382: What happens if we don't have this condition check?

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -398,3 +429,82 @@ func isUsingUnsupportedSpotMarketOptions(pool *hivev1.MachinePool, clusterVersio
...
+		// or vpc peering connections (starting with pcx).
+		if strings.HasPrefix(aws.StringValue(route.GatewayId), "igw") {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (a *AWSActuator) validateSubnets(subnets map[string]ec2.Subnet, pool *hivev1.MachinePool) (map[string]string, error) {
```
akhil-rane, line 479: Doc that describes what this function does will be helpful. From my understanding, it does not check if a subnet is valid or not (`DescribeSubnets` function does that) but rather ensures that we have only one subnet for an availability zone and returns an error if not so.

akhil-rane (commented):
Looks good. Some minor suggestions.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,34 +351,62 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
+
+	var privateSubnets, publicSubnets = map[string]ec2.Subnet{}, map[string]ec2.Subnet{}
 	for _, subnet := range results.Subnets {
-		if subnetID, ok := subnetsByAvailabilityZone[*subnet.AvailabilityZone]; ok {
-			conflictingSubnets.Insert(*subnet.SubnetId)
-			conflictingSubnets.Insert(subnetID)
-			continue
+		isPublic, err := isSubnetPublic(routeTables.RouteTables, *subnet.SubnetId, a.logger)
+		if err != nil {
+			return nil, errors.Wrap(err, "describing route tables")
```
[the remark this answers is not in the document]
suhanime: The method itself fails when the route tables are not appropriate (or the gateway id cannot be fetched) - so in my opinion, route tables error will help more

pkg/controller/remotemachineset/awsactuator.go
suhanime: Public subnets are not mandatory to be mentioned, so it proceeds to fetch private subnets (which must be mentioned if using subnets) if public subnets are not present

pkg/controller/remotemachineset/awsactuator.go
suhanime: I briefly considered naming this function the getSubnetsByAvailabilityZone since it returns the list, however we also use that to validate the subnets (aka ensure no conflicts). So I added a small doc - let me know if it is enough

akhil-rane: Doc should be good.

pkg/controller/remotemachineset/awsactuator.go
akhil-rane: should we have same check for private subnets?

suhanime: Private subnets are mandatory (if subnets are mentioned)... And since the total subnets are Public+Private, we don't need a check for that

pkg/controller/remotemachineset/awsactuator.go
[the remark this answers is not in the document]
akhil-rane: You are right about the cause of the error here. but do you think having `error checking if subnet is public` in the error message will help in pinpointing exactly where the error has occurred? Also, the subnet which failed the check should be included in the message. I feel the cause of the actual error will be incorporated when you do the wrap. Can you paste the sample error message here?
 
 @abhinavdahiya what is your opinion?

akhil-rane:
This looks good to me. @abhinavdahiya over to you :)

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,34 +348,62 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
+		Filters: []*ec2.Filter{{
+			Name:   aws.String("vpc-id"),
+			Values: []*string{aws.String(vpc)},
+		}},
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "error describing route tables")
+	}
+
+	var privateSubnets, publicSubnets = map[string]ec2.Subnet{}, map[string]ec2.Subnet{}
```
abhinavdahiya, line 366: nit: storing pointers would probably be more memory optimistic ...

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,34 +348,62 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
 	for _, subnet := range results.Subnets {
-		if subnetID, ok := subnetsByAvailabilityZone[*subnet.AvailabilityZone]; ok {
-			conflictingSubnets.Insert(*subnet.SubnetId)
-			conflictingSubnets.Insert(subnetID)
-			continue
+		isPublic, err := isSubnetPublic(routeTables.RouteTables, *subnet.SubnetId, a.logger)
+		if err != nil {
+			return nil, errors.Wrap(err, "error describing route tables")
```
abhinavdahiya, line 370: the error descriptions seems off from the function generating the error.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,34 +348,62 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
+			privateSubnets[*subnet.SubnetId] = *subnet
 		}
-		subnetsByAvailabilityZone[*subnet.AvailabilityZone] = *subnet.SubnetId
 	}
 
-	if len(conflictingSubnets) > 0 {
+	if len(publicSubnets) > 0 {
+		_, err := a.validateSubnets(publicSubnets, pool)
+		if err != nil {
+			return nil, err
```
abhinavdahiya, line 382: let's wrap the error with context.

pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -348,34 +348,62 @@ func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (ma
...
+		}
+	}
+
+	subnetsByAvailabilityZone, err := a.validateSubnets(privateSubnets, pool)
+	if err != nil {
+		return nil, err
```
abhinavdahiya, line 388: let's wrap the error with context.

abhinavdahiya:
mostly cosmetic suggestions.

/lgtm

PR #1236, opened 2020-12-17
Add regex definitions for this week's alerts

twiest:
- [x] Add regex definitions for this week's alerts
- [x] Test regex against messages seen

/assign @suhanime 
/cc @dgoodwin

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -85,6 +85,21 @@ data:
       - "failed to initialize the cluster: Cluster operator monitoring is still updating"
       installFailingReason: MonitoringOperatorStillUpdating
       installFailingMessage: Timeout waiting for the monitoring operator to become ready
+    - name: AuthenticationOperatorDegraded
+      searchRegexStrings:
+      - "Cluster operator authentication Degraded is True"
+      installFailingReason: AuthenticationOperatorDegraded
+      installFailingMessage: Timeout waiting for the authentication operator to become ready
+    - name: GeneralOperatorDegraded
+      searchRegexStrings:
+      - "Cluster operator.*Degraded is True"
+      installFailingReason: GeneralOperatorDegraded
+      installFailingMessage: Timeout waiting for an operator to become ready
+    - name: GeneralClusterOperatorsStillUpdating
+      searchRegexStrings:
+      - "failed to initialize the cluster: Some cluster operators are still updating:"
+      installFailingReason: GeneralClusterOperatorsStillUpdating
+      installFailingMessage: Timeout waiting for all cluster operators to become ready
     # Processing stops at the first match, so this more generic
     # message about the connection failure must always come after the
     # more specific message for LibvirtSSHKeyPermissionDenied.
@@ -93,3 +108,8 @@ data:
       - "could not connect to libvirt"
       installFailingReason: LibvirtConnectionFailed
       installFailingMessage: "Could not connect to libvirt host"
+    - name: GeneralQuota
+      searchRegexStrings:
+      - "Quota '[A-Z_]*' exceeded"
+      installFailingReason: GeneralQuotaExceeded
+      installFailingMessage: Quota exceeded
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -1612,6 +1612,21 @@ data:
       - "failed to initialize the cluster: Cluster operator monitoring is still updating"
       installFailingReason: MonitoringOperatorStillUpdating
       installFailingMessage: Timeout waiting for the monitoring operator to become ready
+    - name: AuthenticationOperatorDegraded
+      searchRegexStrings:
+      - "Cluster operator authentication Degraded is True"
+      installFailingReason: AuthenticationOperatorDegraded
+      installFailingMessage: Timeout waiting for the authentication operator to become ready
+    - name: GeneralOperatorDegraded
+      searchRegexStrings:
+      - "Cluster operator.*Degraded is True"
+      installFailingReason: GeneralOperatorDegraded
+      installFailingMessage: Timeout waiting for an operator to become ready
+    - name: GeneralClusterOperatorsStillUpdating
+      searchRegexStrings:
+      - "failed to initialize the cluster: Some cluster operators are still updating:"
+      installFailingReason: GeneralClusterOperatorsStillUpdating
+      installFailingMessage: Timeout waiting for all cluster operators to become ready
     # Processing stops at the first match, so this more generic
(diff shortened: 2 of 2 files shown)


twiest:
@suhanime @dgoodwin This is the first time I've submitted these. I think I followed the pattern, but please double check that I did it correctly. I'm positive the regexes match the log messages (I used an online regex checker).

Thanks!

suhanime (commented):
Thank you! These will be very helpful...

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -42,6 +42,11 @@ data:
       - "Throttling: Rate exceeded"
       installFailingReason: AWSAPIRateLimitExceeded
       installFailingMessage: AWS API rate limit exceeded
+    - name: GeneralQuota
+      searchRegexStrings:
+      - "Quota '[A-Z_]*' exceeded\."
```
suhanime, line 47:
```suggestion
     - "Quota '[A-Z_]*' exceeded"
```

pkg/operator/assets/bindata.go
```diff
@@ -1594,6 +1599,16 @@ data:
       - "failed to initialize the cluster: Cluster operator monitoring is still updating"
       installFailingReason: MonitoringOperatorStillUpdating
       installFailingMessage: Timeout waiting for the monitoring operator to become ready
+    - name: AuthenticationOperatorDegraded
+      searchRegexStrings:
+      - "Cluster operator authentication Degraded is True"
+      installFailingReason: AuthenticationOperatorDegraded
+      installFailingMessage: Timeout waiting for the authentication operator to become ready
```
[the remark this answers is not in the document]
twiest: That sounds great, as long as we can assume all OperatorDegraded messages are not Hive. Otherwise I'd prefer to have each degraded or group of degradeds have separate entries so that know we know if we have to take action or not.

In other words, I'd be against grouping OperatorDegraded together if we have both ones that we can ignored and ones that need action.

@dgoodwin Can we assume that all OperatorDegraded's are not Hive related (at least until shown otherwise)?

config/configmaps/install-log-regexes-configmap.yaml
twiest: Hmm, doesn't the `-` need to be right under the `s` from the line above?

Also, what's the reason to match on the period?

suhanime: Oh my mistake - I didn't mean to shift the formatting, just wanted to point out we don't need to match the period - fixed the comment

pkg/operator/assets/bindata.go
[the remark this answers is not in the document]
dgoodwin: Feels like it's probably safe to assume it's not us at this point in time, I can't think of any examples off hand where that happened and it was something we could address.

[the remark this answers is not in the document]
gregsheremeta: Let's also think about other hive users who could benefit from these. As an SRE I think it would be nice to look at the ClusterDeployment CR status and see exactly which operator is degraded.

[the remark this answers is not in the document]
twiest: I added a new regex called `GeneralOperatorDegraded` to catch the general operator degraded case. I made sure to add it after the more specific `AuthenticationOperatorDegraded` regex so that the specific one is matched first.

twiest:
@dgoodwin @suhanime @gregsheremeta I believe I've addressed your feedback. PTAL.

pkg/operator/assets/bindata.go
```diff
@@ -1594,6 +1599,21 @@ data:
...
+    - name: GeneralOperatorDegraded
+      searchRegexStrings:
+      - "Cluster operator.*Degraded is True"
+      installFailingReason: GeneralOperatorDegraded
+      installFailingMessage: Timeout waiting for an authentication operator to become ready
```
suhanime, line 1611: This one is a general message for operators other than authentication one, if degraded...

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -42,6 +42,11 @@ data:
       - "Throttling: Rate exceeded"
       installFailingReason: AWSAPIRateLimitExceeded
       installFailingMessage: AWS API rate limit exceeded
+    - name: GeneralQuota
+      searchRegexStrings:
+      - "Quota '[A-Z_]*' exceeded"
+      installFailingReason: AWSQuotaExceeded
+      installFailingMessage: AWS Quota exceeded
```
akhil-rane, line 49: @twiest Is this for AWS only or can happen for any cloud? I got one today for GCP with similar error message.
```
level=info msg="Credentials loaded from environment variable \"GOOGLE_CREDENTIALS\", file \"/.gcp/osServiceAccount.json\""
    level=info msg="Consuming Install Config from target directory"
    level=info msg="Manifests created in: manifests and openshift"
    level=warning msg="Found override for release image. Please be warned, this is not advised"
    level=info msg="Consuming OpenShift Install (Manifests) from target directory"
    level=info msg="Consuming Common Manifests from target directory"
    level=info msg="Consuming Worker Machines from target directory"
    level=info msg="Consuming Master Machines from target directory"
    level=info msg="Consuming Openshift Manifests from target directory"
    level=info msg="Ignition-Configs created in: . and auth"
    level=info msg="Credentials loaded from environment variable \"GOOGLE_CREDENTIALS\", file \"/.gcp/osServiceAccount.json\""
    level=info msg="Consuming Master Ignition Config from target directory"
    level=info msg="Consuming Bootstrap Ignition Config from target directory"
    level=info msg="Consuming Worker Ignition Config from target directory"
    level=warning msg="Failed to find information on quotas compute.googleapis.com/cpus, compute.googleapis.com/regional_static_addresses"
    level=info msg="Creating infrastructure resources..."
    level=error
    level=error msg="Error: Error waiting for instance to create: Quota 'SSD_TOTAL_GB' exceeded.  Limit: 500.0 in region asia-northeast2."
    level=error
    level=error
    level=error msg="  on ../tmp/openshift-install-944165972/bootstrap/main.tf line 44, in resource \"google_compute_instance\" \"bootstrap\":"
    level=error msg="  44: resource \"google_compute_instance\" \"bootstrap\" {"
    level=error
    level=error
    level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply Terraform: failed to complete the change"
    level=fatal msg="failed to get bootstrap and control plane host addresses from \"terraform.tfstate\": no bootstrap instance found"
```

twiest:
@akhil-rane @suhanime I believe I've addressed your feedback. PTAL.

PR #1237, opened 2020-12-17
clusterdeployment: include InstallImagesNotResolved condition when job fails

abhinavdahiya:
The cluster deployment controller creates a job that resolves the installer and cli
images for installing the cluster. this job uses the release image to extract these details.
When the release image being used can't be pulled due to some failure like authentication
the controller provides no update / condition that would provide use details on this failure and
the user sees no update on the clusterdeployment object.
With this change we add InstallImagesNotResolved condition to the clusterdeployment when the job fails.

Previously the job had activeDeadline set to 24 hrs, therefore the job controller would wait for that interval
before marking it complete+failed. To make sure the controller can actually give users report of job failure
the deadline is reduces to 2 minutes. This now matches the deadline used by CVO that uses a very similar job to
interact with the release-image contents [1].

NOTE: currently this depends on job controller's status reporting to provide user the reason for failure, but
we could later extend it to summarize the pods's failure reason as the pods from job stay until the job is deleted.

[1]: https://github.com/openshift/cluster-version-operator/blob/84b3884e422739dbbfa33078e62349752b5afa18/pkg/cvo/updatepayload.go#L146

xref: https://issues.redhat.com/browse/CO-1306

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1125,6 +1125,11 @@ func (r *ReconcileClusterDeployment) statusUpdate(cd *hivev1.ClusterDeployment, 
 	return err
 }
 
+const (
+	imagesResolvedReason = "ImagesResolved"
+	imagesResolvedMsg    = "Images required for cluster deployment installations are resolved"
+)
+
 func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDeployment, imageSet *hivev1.ClusterImageSet, releaseImage string, cdLog log.FieldLogger) (*reconcile.Result, error) {
 	areImagesResolved := cd.Status.InstallerImage != nil && cd.Status.CLIImage != nil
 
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2428,3 +2476,20 @@ func testCompletedImageSetJob() *batchv1.Job {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      imageSetJobName,
+			Namespace: testNamespace,
+		},
+		Status: batchv1.JobStatus{
+			Conditions: []batchv1.JobCondition{{
+				Type:    batchv1.JobFailed,
+				Status:  corev1.ConditionTrue,
+				Reason:  "ImagePullBackoff",
+				Message: "The pod failed to start because on the containers did not start",
```
akhil-rane, line 2491:
```suggestion
				Message: "The pod failed to start because one of the containers did not start",
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1136,7 +1136,7 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 	// The job does not exist. If the images have been resolved, continue reconciling. Otherwise, create the job.
 	case apierrors.IsNotFound(err):
 		if areImagesResolved {
-			return nil, nil
+			return nil, r.setInstallImagesNotResolvedCondition(cd, corev1.ConditionFalse, "ImagesResolved", "Images required for cluster deployment installations are resolved", cdLog)
```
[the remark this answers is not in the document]
abhinavdahiya: @akhil-rane done with 88f1f0f

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
abhinavdahiya: @akhil-rane done with 88f1f0f

Issue #1235, opened 2020-12-17
vsphere hive deployments broken as of v1.0.16


rayashworth:
been test vsphere hive deployments and it would appear that v1.0.16 has a breaking change.

Scenarios:
- Install OCP cluster using ACM UI hive version v1.0.14. works fine
- install OCP cluster using direct call to hive (v1.0.14) instance under ACM, works fine.
- install OCP cluster using hive v1.0.14 public operator. Works fine.
- install OCP clsuter using hive v1.0.16, not working.
- install OCP cluster using hive v1.0.15, not working. does not appear to create all of the hive pods after deploy.

error messages (snippet)
time="2020-12-17T14:49:58.259Z" level=info msg="reconcile complete" controller=metrics elapsedMillis=0 elapsedMillisGT=0 outcome=unspecified
time="2020-12-17T14:51:58.259Z" level=info msg="calculating metrics across all ClusterDeployments" controller=metrics
time="2020-12-17T14:51:58.259Z" level=warning msg="condition type AuthenticationFailure missing from AllClusterDeploymentConditions slice"
time="2020-12-17T14:51:58.26Z" level=warning msg="condition type AuthenticationFailure missing from AllClusterDeploymentConditions slice"
time="2020-12-17T14:51:58.26Z" level=debug msg="calculating metrics across all install jobs" controller=metrics
time="2020-12-17T14:51:58.26Z" level=debug msg="calculating metrics across all uninstall jobs" controller=metrics
time="2020-12-17T14:51:58.26Z" level=debug msg="calculating metrics across all imageset jobs" controller=metrics
time="2020-12-17T14:51:58.26Z" level=debug msg="calculating metrics across all ClusterSyncs" controller=metrics
time="2020-12-17T14:51:58.26Z" level=info msg="reconcile complete" controller=metrics elapsedMillis=0 elapsedMillisGT=0 outcome=unspecified
time="2020-12-17T14:53:58.26Z" level=info msg="calculating metrics across all ClusterDeployments" controller=metrics
time="2020-12-17T14:53:58.26Z" level=warning msg="condition type AuthenticationFailure missing from AllClusterDeploymentConditions slice"
time="2020-12-17T14:53:58.261Z" level=warning msg="condition type AuthenticationFailure missing from AllClusterDeploymentConditions slice"
time="2020-12-17T14:53:58.261Z" level=debug msg="calculating metrics across all install jobs" controller=metrics
[...]

PR #1238, opened 2020-12-21
use ClusterDeployment's CA data when validating vsphere credentials

joelddiaz:
When a vSphere ClusterDeployment defines the CA certificates Secret, use the CA certificate data when validating vSphere credentials.

The high-level client library doesn't allow providing a CA certificate early enough, so we need to implement a lot of the lower-level connection setup so that we can slip in the custom CA data when needed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
+	github.com/vmware/govmomi v0.22.2
 	golang.org/x/lint v0.0.0-20200302205851-738671d3881b
 	golang.org/x/mod v0.3.0
 	golang.org/x/net v0.0.0-20200707034311-ab3426394381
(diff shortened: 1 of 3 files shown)


joelddiaz:
I put this together while the investigation into #1235 was fresh in my head.
This needs to be tested/verified against both a vSphere environment that doesn't need CA Secret data (ie properly signed and trusted certificates) and a vSphere environment using untrusted certificates.

go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
+	github.com/vmware/govmomi v0.22.2
```
suhanime, line 43: Was there any go sum changes with this?

joelddiaz: No. The only thing that has changed (vendoring-wise) is that hive is now directly importing packages from github.com/vmware/govmmomi instead of calling into github.com/openshift/installer to test connectivity to a vsphere endpoint.

dgoodwin:
/cc @akhil-rane 

In case this is relevant for your current work.

chrisahl:
ACM team is hoping this can be reviewed and merged ASAP so we can get a new hive build published.

akhil-rane (commented):
/lgtm

PR #1240, opened 2020-12-21
clusterpool: allow setting lifetime for claims

abhinavdahiya:
claimLifetime allows admins to set the lifetime for all cluster claims created for
a cluster pool. Two new lifetime configurations are available, `default` and `maximum`

clusterpool`.spec.cliamLifetime.default`
This sets the default lifetime for the claim when one is not set by the claim itself.

clusterpool`.spec.claimLifetime.maximum`
This sets the maximum lifetime for all claims in the cluster pool. If any user sets lifetime
on the claim, the actual lifetime will be minimum of the maximum allowed by the pool and what was
set on the claim.
This allows the admins makes sure that the tightest lifetime constraint is always respected for the
cluster pool.

Also a new status field is added to the clusterclaim `.status.lifetime` that provides information of
the exact lifetime including all the pool configurations and the claim spec.

NOTE: the cluster pool controller could also set the lifetime on the claim when it assigns a cluster
to the claim but that would would require setting the intention on the object as part of the UPDATE hence
making it difficult to
- later figure out if the lifetime was set by user or pool which is important if the claim ever had to
move to a different pool.
- also the user could change the claim lifetime to somthing else after the controller updated it and the controller
would have to fiure out if the change should be kept and then revert the change, possibly race with the user.
This current approach with clusterclaim controller computing lifetime allows users to express their intentions at
both levels in a much more clear way and it can be enforced similarly.

/assign @dgoodwin 
/cc @abutcher 

xref: https://issues.redhat.com/browse/HIVE-1283


pkg/apis/hive/v1/clusterpool_types.go
```diff
@@ -62,6 +62,11 @@ type ClusterPoolSpec struct {
 	// SkipMachinePools allows creating clusterpools where the machinepools are not managed by hive after cluster creation
 	// +optional
 	SkipMachinePools bool `json:"skipMachinePools,omitempty"`
+	// ClaimLifetime is the maximum lifetime of the claim after it is assigned a cluster. If the claim still exists
+	// when the lifetime has elapsed, the claim will be deleted by Hive.
+	// The lifetime of a claim is the mimimum of the lifetimes set by the cluster pool and the claim itself.
+	// +optional
```
dgoodwin, line 68: I'm concerned about one use case here, if someone were to hit a serious issue and needed to keep a cluster around as a reproducer until someone could investigate. With this policy I don't think there's any way for them to override that and it's the kind of thing that seems like it almost surely will arise at some point.

We should probably assume some level of trust and let explicit lifetime requests on a claim be respected, even if it's just on updates. We could use this logic on create, but let users explicitly modify it after creation forcing them to take a more explicit step. Tooling to look for clusters around too long would be easy to write.

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -244,6 +250,40 @@ func (r *ReconcileClusterClaim) Reconcile(request reconcile.Request) (result rec
...
+}
+
+func (r *ReconcileClusterClaim) clusterPoolLifetimeForClaim(claim *hivev1.ClusterClaim, logger log.FieldLogger) (*metav1.Duration, error) {
+	// Fetch the ClusterPool instance
+	clp := &hivev1.ClusterPool{}
+	// claims exists in the same namespace as
+	key := client.ObjectKey{Namespace: claim.Namespace, Name: claim.Spec.ClusterPoolName}
+	err := r.Get(context.TODO(), key, clp)
+	if apierrors.IsNotFound(err) {
+		logger.Info("pool not found")
```
dgoodwin, line 276: Suggest: logger.WithField("pool", key).WithField("claim", claim.Name).Info("cluster pool no longer exists")

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -200,8 +200,14 @@ func (r *ReconcileClusterClaim) Reconcile(request reconcile.Request) (result rec
 
 	logger = logger.WithField("cluster", clusterName)
 
+	poolLifetime, err := r.clusterPoolLifetimeForClaim(claim, logger)
+	if err != nil {
+		logger.Log(controllerutils.LogLevel(err), "error getting ClusterPool")
```
dgoodwin, line 205: Nit: error getting cluster claim lifetime

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -200,8 +200,14 @@ func (r *ReconcileClusterClaim) Reconcile(request reconcile.Request) (result rec
...
+		return reconcile.Result{}, err
+	}
+
 	// Delete ClusterClaim after its lifetime elapses
-	if lifetime := claim.Spec.Lifetime; lifetime != nil {
+	if lifetime := getClaimLifetime(poolLifetime, claim.Spec.Lifetime); lifetime != nil {
```
dgoodwin, line 210: Am I correct that a user can view their ClusterClaim, they are able to set a lifetime on it, see the lifetime that they set, but that may not be the actual lifetime in use at all? If so I fear someone is going to get a nasty surprise from this at an inopportune time. 

I think this plays into the above. I think ClusterPool.Spec.ClaimLifetime should be a default, but the Claim can always override. A claim created without a lifetime, when the pool has one, should be updated to include that value as part of reconciliation so the user can always see the actual value in play by looking at the claim.

WDYT?

pkg/apis/hive/v1/clusterpool_types.go
abhinavdahiya: i think goal for setting a lifetime by the clusterpool admin is to make sure no cluster stays up _for long_ because these clusters use up the resources/quota for the clusterpool.  only setting the lifetime at create and allowing **users** to override the lifetime to anything seems incorrect from the admin's perspective as users should not be allowed to override the will the admin.

> 'm concerned about one use case here, if someone were to hit a serious issue and needed to keep a cluster around

i don't think user that created the claim should have the power to decide or lengthen the lifetime of the cluster and it should only be the admins that have the power to make the exception. 

the admins could detach the clusterdeployment from the pool/claim so that it lives for longer.. _not sure yet if that is possible_

pkg/controller/clusterclaim/clusterclaim_controller.go
abhinavdahiya: > Am I correct that a user can view their ClusterClaim, they are able to set a lifetime on it, see the lifetime that they set, but that may not be the actual lifetime in use at all? If so I fear someone is going to get a nasty surprise from this at an inopportune time.

i think we can add a status field to provide the accurate lifetime of the claim which will incorporate both the sources.

pkg/apis/hive/v1/clusterpool_types.go
dgoodwin: My assumption was it was best as a guideline from the admin to make sure nobody forgot to set a lifetime, and they would be willing to exercise a level of trust. I'll ping Steve and some ACM folks and get some feedback and see.

abhinavdahiya:
@dgoodwin updated the lifetime to support both the default and maximum explicitly. PTAL

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -244,6 +260,47 @@ func (r *ReconcileClusterClaim) Reconcile(request reconcile.Request) (result rec
...
+
+	}
+	return lifetime
+}
+
+// clusterPoolLifetimeForClaim returns the default and max lifetimes for the cluster pool the claim belongs to.
+func (r *ReconcileClusterClaim) clusterPoolLifetimeForClaim(claim *hivev1.ClusterClaim, logger log.FieldLogger) (*hivev1.ClusterPoolClaimLifetime, error) {
+	// Fetch the ClusterPool instance
+	clp := &hivev1.ClusterPool{}
+	// claims exists in the same namespace as
```
dgoodwin, line 289: "as the pool"

dgoodwin (commented):
Looks great just a couple nits from last review.

pkg/controller/clusterclaim/clusterclaim_controller.go
abhinavdahiya: fixed in 
c29403e

pkg/controller/clusterclaim/clusterclaim_controller.go
abhinavdahiya: this function is to get the lifetime specified for the claim in the cluster pool. so we know this function relates to cluster pool. changed to `error getting cluster pool lifetime` in c29403e

pkg/controller/clusterclaim/clusterclaim_controller.go
abhinavdahiya: @dgoodwin fixed in c29403e

PR #1242, opened 2021-01-05
Move pullspecs to new openshift CI registry.

dgoodwin:
QE reporting problems with the old registry, which appears to at the very least need auth now.

/assign @suhanime

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 as builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.15 as builder
 RUN mkdir -p /go/src/github.com/openshift/hive
 WORKDIR /go/src/github.com/openshift/hive
 COPY . .
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,5 +1,5 @@
 #FROM registry.access.redhat.com/ubi7/ubi
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15
+FROM registry.ci.openshift.org/openshift/release:golang-1.15
 
 # ssh-agent required for gathering logs in some situations:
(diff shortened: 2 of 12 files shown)


suhanime:
Few questions/requests:

- We should change in the hiveutil files as well - see [1](https://github.com/openshift/hive/blob/master/contrib/pkg/createcluster/create.go#L84), [2
](https://github.com/openshift/hive/blob/master/contrib/pkg/createcluster/create.go#L252), [3](https://github.com/openshift/hive/blob/master/contrib/pkg/clusterpool/clusterpool.go#L46) and [4](https://github.com/openshift/hive/blob/master/contrib/pkg/clusterpool/clusterpool.go#L136)

- Will [this](https://github.com/openshift/hive/blob/master/hack/hiveadmission-dev-cert.sh#L9-L14) be okay?

- Did we test this? Since e2es are passing regardless...

dgoodwin:
> Few questions/requests:
> 
>     * We should change in the hiveutil files as well - see [1](https://github.com/openshift/hive/blob/master/contrib/pkg/createcluster/create.go#L84), [2](https://github.com/openshift/hive/blob/master/contrib/pkg/createcluster/create.go#L252), [3](https://github.com/openshift/hive/blob/master/contrib/pkg/clusterpool/clusterpool.go#L46) and [4](https://github.com/openshift/hive/blob/master/contrib/pkg/clusterpool/clusterpool.go#L136)

Good thinking, the URLs though do not appear to be valid if we change those over. (https://openshift-release.ci.openshift.org/api/v1/releasestream/4-stable/latest) Suspect it's not something they've migrated to the new registry yet.



> 
>     * Will [this](https://github.com/openshift/hive/blob/master/hack/hiveadmission-dev-cert.sh#L9-L14) be okay?

This one is ok, svc in that context is just the kubernetes DNS stuff.

> 
>     * Did we test this? Since e2es are passing regardless...

I tested deploying yes, but mostly leaning on e2e to expose if anything is wrong. Those registry.svc images should not be used in SD environments afaik.

PR #1246, opened 2021-01-07
Make ocm-2.2 branch up to date with master

akhil-rane:
/assign: @dgoodwin 
/cc: @chrisahl


dgoodwin:
It looks like we do not have CI on this branch. That's probably ok for now as this is a fast forward to master, I will add a Jira for future PRs where we might be cherry picking a fork.

PR #1247, opened 2021-01-07
Add agent based install platform

dgoodwin:
Purpose of this PR is to get the discussion going between assisted-service and hive teams to figure out how to best model this currently imperative API, as a declarative CRD based Kube API.

I will start by bringing over the WIP API the assisted-service team, adding comments as a PR review, and then once we've resolved issues, updating the API.


pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Platform stores cluster configuration for an agent based install where hosts are booted and registered, ready
+// to be assigned and used in a cluster.
+type Platform struct {
+	// Cluster name
+	Name string `json:"name"`
```
dgoodwin, line 12: Looks like we can drop this one, this would be ClusterDeployment.Spec.ClusterName.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	// Installation will start once cluster is in known state and approved
+	Approved                 bool   `json:"approved,omitempty"`
```
dgoodwin, line 14: Approval feels imperative to me, if the ClusterDeployment has been created, the request has been made to create a cluster. We might be waiting in the controllers until some minimum number of machines is present. 

Propose we drop this at least for now, and assume it's approved in the controller code.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	OpenshiftVersion         string `json:"openshiftVersion"`
```
dgoodwin, line 15: @avishayt notes that this field would contain something like "4.6" here: https://github.com/openshift/enhancements/pull/574/files#r553131459 and would map to a key in a configmap such as:

```json
{
        "4.6": {
                "display_name": "4.6.8",
                "release_image": "quay.io/openshift-release-dev/ocp-release:4.6.8-x86_64",
                "rhcos_image": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-4.6.8-x86_64-live.x86_64.iso",
                "support_level": "production"
        },
        "4.7": {
                "display_name": "4.7-pre-release",
                "release_image": "quay.io/openshift-release-dev/ocp-release@sha256:2419f9cd3ea9bd114764855653012e305ade2527210d332bfdd6dbdae538bd66",
                "rhcos_image": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-4.6.8-x86_64-live.x86_64.iso",
                "support_level": "beta"
        }
}
```

Hive's closest concept to this is the [ClusterImageSet](https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1/clusterimageset_types.go#L27), which currently has only one field which is release image. ClusterDeployments link to one via 

Can rhcos_image be determined by inspecting the release_image contents? If so, this is a very good parallel to what Hive does today, before launching an install, we extract the installer container image from the release image, so we can run that in our pod that performs the openshift-install. We could use the same path to also extract the rhcosImage, and set it on ClusterDeployment. Status similar to [this](https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1/clusterdeployment_types.go#L223).

I've always wanted to instead store this only once per ClusterImageSet in it's status, you set the release image, and the controllers populate the status for the installer image, cli image, and now rhcos image, but this hasn't happened yet. Would be a nice optimization though and maybe we can start doing it this way here.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	BaseDnsDomain            string `json:"baseDnsDomain,omitempty"`
```
dgoodwin, line 16: Looks like [ClusterDeployment.Spec.BaseDomain](https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1/clusterdeployment_types.go#L75) and could be dropped.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	ClusterNetworkCidr       string `json:"clusterNetworkCidr,omitempty"`
+	ClusterNetworkHostPrefix int64  `json:"clusterNetworkHostPrefix,omitempty"`
+	ServiceNetworkCidr       string `json:"serviceNetworkCidr,omitempty"`
+	ApiVip                   string `json:"apiVip,omitempty"`
+	ApiVipDnsName            string `json:"apiVipDnsName,omitempty"`
+	IngressVip               string `json:"ingresVip,omitempty"`
+	MachineNetworkCidr       string `json:"machineNetworkCidr,omitempty"`
```
dgoodwin, line 23: We do not presently have any of these in the Hive API, we leave them to the install configs. Presumably now we'll need to add them, a question would be if they should be in an agent install platform, or higher level in the clusterdeployment itself, and only used for agent based install for now? For example, will some of these be useful for hypershift?

I'd be quite tempted here to model these to exactly match how the installer's InstallConfig does, move them higher up on clusterdeployment, and only let you specify them if platform = agentbased.

Properties need slight renames to match api conventions. IngressVIP, APIVIP, etc.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	SshPublicKey             string `json:"sshPublicKey,omitempty"`
```
dgoodwin, line 24: Should be SSHPublicKey.

We have CD.Spec.Provisioning.SSHPrivateKeySecretRef. Could we re-use? If we've got the private key, we've got the public key.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	VipDhcpAllocation        bool   `json:"vipDhcpAllocation,omitempty"`
+	HttpProxy                string `json:"httpProxy,omitempty"`
+	HttpsProxy               string `json:"httpsProxy,omitempty"`
+	NoProxy                  string `json:"noProxy,omitempty"`
```
dgoodwin, line 28: Could NoProxy be assumed by the absence of HttpProxy and HttpsProxy?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	UserManagedNetworking    bool   `json:"userManagedNetworking,omitempty"`
+	AdditionalNtpSource      string `json:"additionalNtpSource,omitempty"`
+	InstallConfigOverrides   string `json:"installConfigOverrides,omitempty"`
+	// A reference to the secret containing the pull secret
+	PullSecretRef *corev1.SecretReference `json:"pullSecretRef"`
```
dgoodwin, line 33: Covered by CD.Spec.PullSecretRef and can be dropped.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
+package agent
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Platform stores cluster configuration for an agent based install where hosts are booted and registered, ready
+// to be assigned and used in a cluster.
+type Platform struct {
```
dgoodwin, line 10: What is the bare minimum set of required params for assisted install? Most things are marked omitempty here, implying they are optional, but I suspect a few may actually be mandatory.

dgoodwin (commented):
First pass, haven't broken into status yet, but here are some initial ideas for the API and some things we can hopefully sync up on.

cc @filanov @avishayt @eparis @mhrivnak @abhinavdahiya @markmc

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	// Installation will start once cluster is in known state and approved
+	Approved                 bool   `json:"approved,omitempty"`
+	OpenshiftVersion         string `json:"openshiftVersion"`
+	BaseDnsDomain            string `json:"baseDnsDomain,omitempty"`
+	ClusterNetworkCidr       string `json:"clusterNetworkCidr,omitempty"`
+	ClusterNetworkHostPrefix int64  `json:"clusterNetworkHostPrefix,omitempty"`
+	ServiceNetworkCidr       string `json:"serviceNetworkCidr,omitempty"`
+	ApiVip                   string `json:"apiVip,omitempty"`
+	ApiVipDnsName            string `json:"apiVipDnsName,omitempty"`
+	IngressVip               string `json:"ingresVip,omitempty"`
```
mhrivnak, line 22: s/ingres/ingress/

pkg/apis/hive/v1/agent/platform.go
mhrivnak: This field is based on a commonly-used convention born out of libcurl, where you can list domain names in this field that will not be proxied even if there are proxy settings in place. It's often used to prevent local traffic from being sent through a proxy.

pkg/apis/hive/v1/agent/platform.go
mhrivnak: What assumptions does hive have, if any, about the namespace of this secret as it relates to the ClusterDeployment?

pkg/apis/hive/v1/agent/platform.go
mhrivnak: I think this field delegates responsibility to the user for waiting until all of the expected agents have checked in and completed their inspection and validation. We otherwise do not (yet!) have a way to declare "wait until 3 agents are available with the role 'master' and 5 with the role 'worker'" or similar. Perhaps we should add that now, because it definitely matches the declarative pattern better.

pkg/apis/hive/v1/agent/platform.go
avishayt: Agree

pkg/apis/hive/v1/agent/platform.go
avishayt: This is actually required for the interactive discovery flow, where the user doesn't know details about the hosts that will boot, wants to see validation results, wants to see default values that the service chose and the user might want to override.  There is also a security concern in BIY mode, where we don't automatically want to add every host that boots the ISO without the user's approval.  It's an imperative process, there's no way around it as far as I can tell.

pkg/apis/hive/v1/agent/platform.go
avishayt: Note that this configmap is used by the service, not the user.  The user asks for 4.6.  The service uses the information in the configmap to know how to provide that.

pkg/apis/hive/v1/agent/platform.go
filanov: In order to start installation we need most of the parameters, just to create a cluster you need name, openshift version and pull secret ref.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Must be in the namespace with the ClusterDeployment, so we can mount it into install pods etc.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Ack!

pkg/apis/hive/v1/agent/platform.go
dgoodwin: This is how ClusterImageSet works as well, they are configured centrally but management for the Hive service, people creating clusters can then reference them.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Hmm interesting, ok so moving to declarative we should assume there's no distinction there anymore, and all of these properties are now mandatory. I will drop the omitempties.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Sounds similar to CertificateSigningRequest.

Feels like Clusters are not what is actually getting approved, rather Agents coming up and trying to bind to a cluster. I would pitch that we move this off cluster and instead add the min requirements to start installation. Then the Agent API could optionally require approval to count against those min requirements.

pkg/apis/hive/v1/agent/ipnet.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+var nullString = "null"
+var nullBytes = []byte(nullString)
+var emptyIPNet = net.IPNet{}
+
+// IPNet wraps net.IPNet to get CIDR serialization.
+type IPNet struct {
+	net.IPNet
+}
```
dgoodwin, line 18: This entire file is copied right out of the installer. Suggestions on how best to handle? We could probably just use the vendored copy we have instead.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,131 @@
...
+	//corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Platform stores cluster configuration for an agent based install where hosts are booted and registered, ready
+// to be assigned and used in a cluster.
+type Platform struct {
+	ApiVip            string `json:"apiVip,omitempty"`
+	ApiVipDnsName     string `json:"apiVipDnsName,omitempty"`
+	IngressVip        string `json:"ingresVip,omitempty"`
```
dgoodwin, line 13: These three items above look to be very metal specific to me. Installer has these within the BareMetal platform object. If we keep them here as is, it implies that an Agent based install platform is metal specific. I've heard reports that Agent based installs could be done on any cloud though, is this a feature we wish to leave the door open for in the API? I'm not clear why anyone would do that when there's a cloud API in play. It also waters down the installer's purpose, so I'm really unsure if we should focus on it here.

If we do want to keep that door open, we have an unfortunate platform within a platform requirement... How could we model that sanely?

Suggestion is to leave these where they are and consider Agent based install to be metal specific, but open to alternatives.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,131 @@
...
+	VipDhcpAllocation bool   `json:"vipDhcpAllocation,omitempty"`
+
+	// Networking is the configuration for the pod network provider in
+	// the cluster.
+	*Networking `json:"networking,omitempty"`
```
dgoodwin, line 18: Replaced network settings with structs identical to installer InstallConfig.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Proxy defines the proxy settings for the cluster.
+	// If unset, the cluster will not be configured to use a proxy.
+	// +optional
+	Proxy *Proxy `json:"proxy,omitempty"`
```
dgoodwin, line 23: Similar for proxy, now identical to install config.

pkg/apis/hive/v1/agent/platform.go
mhrivnak: Perfect.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,131 @@
...
+
+	UserManagedNetworking  bool   `json:"userManagedNetworking,omitempty"`
+	AdditionalNtpSource    string `json:"additionalNtpSource,omitempty"`
+	InstallConfigOverrides string `json:"installConfigOverrides,omitempty"`
+
+	SshPublicKey string `json:"sshPublicKey,omitempty"`
```
dgoodwin, line 29: Still a bit unsure what to do with this public key here. 

I assume this is used to place in the generated InstallConfig?

Hive has cd.Spec.Provisioning.SSHPrivateKeySecretRef. THis is *assumed* to match whatever was in the install-config you gave us, and only used to gather logs from hosts in the event of a failure.

In theory, with the private key we can get the public key.

```bash
ssh-keygen -y -f ~/.ssh/id_rsa > ~/.ssh/id_rsa.pub
```

But I'm not super eager to start shelling out. Tempted to not mandate giving us a private key and let assisted callers specify a public key only.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,131 @@
...
+	// the cluster.
+	Networking *Networking `json:"networking,omitempty"`
+
+	// Proxy defines the proxy settings for the cluster.
+	// If unset, the cluster will not be configured to use a proxy.
+	// +optional
+	Proxy *Proxy `json:"proxy,omitempty"`
+
+	UserManagedNetworking  bool   `json:"userManagedNetworking,omitempty"`
+	AdditionalNtpSource    string `json:"additionalNtpSource,omitempty"`
```
dgoodwin, line 26: @avishayt what is this AdditionalNtpSource and how is it used? Is it something we're positive must be exposed on the kube API now? Is it day 2 configuration going to the cluster?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,131 @@
...
+	InstallConfigOverrides string `json:"installConfigOverrides,omitempty"`
```
dgoodwin, line 27: @avishayt this one also looks like a candidate for keeping off the API, at least for now, and analyzing when and why it's needed on a case by case basis.

pkg/apis/hive/v1/agent/platform.go
filanov: It's a day1 configuration, we require to sync clocks between the hosts before we start the installation, so if user for some reason block access to a default ntp server we have a way for him to provide additional ntp sources.

dgoodwin: What's the format? IP address or hostname? Actually if you've got time Michael, if you could go through these in their current state and add godoc for each field, that would be super helpful for me as I don't know the source material very well.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: After discussing with hive team, I've moved these to a BareMetal specific sub-struct. That struct is optional, but the fields within it are mandatory. This keeps the door open for non-metal specific agent based installs in the future.

pkg/apis/hive/v1/agent/platform.go
filanov: I'll try to do it tomorrow, it will a copy from what we have in the swagger file. 
https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L3297-L3300

filanov: In this case it can be both ip or name

dgoodwin: Ah no worries then, I can work off your swagger so I can add it here.

pkg/apis/hive/v1/agent/platform.go
abhinavdahiya: i think requiring an ssh private key will be key for debugging info. so i think we should keep the private key and update the assisted installer code to extract public key before giving it to the installer..

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	//corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Platform stores cluster configuration for an agent based install where hosts are booted and registered, ready
+// to be assigned and used in a cluster.
+type Platform struct {
+	BareMetal *BareMetal `json:"bareMetal"` // mandatory for now, possibly optional in future
```
abhinavdahiya, line 11: let's create a Provider field and then add BareMetal to it. this should allow for extension later if we need different provider for agent based installs.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+
+	SshPublicKey string `json:"sshPublicKey,omitempty"`
+}
+
+// BareMetal contains agent based install configuration specific to bare metal clusters.
+type BareMetal struct {
+	APIVIP            string `json:"apiVIP"`
+	APIVIPDNSName     string `json:"apiVIPDNSName"`
+	IngressVIP        string `json:"ingressVIP"`
+	VIPDHCPAllocation bool   `json:"vipDHCPAllocation"`
```
abhinavdahiya, line 34: `VIPDHCPAllocation` what does this field do? auto assign all the VIPS requested above?

i would recommend using union discriminator like https://github.com/openshift/api/blob/790e0a84e3e01dc3f6218e655b0e92b809ae7fd0/config/v1/types_infrastructure.go#L145-L146 here.

enumA -> auto assign
enumB -> user provided.

pkg/apis/hive/v1/agent/ipnet.go
abhinavdahiya: hmm.. i think we can replace IPNet with just string. For installer this is mainly used to reject install configs with invalid cidr string on unmarshalling, but since this is a CRD this is going to problems in hive code after the object has been accepted.

so the better use here will be to have it as string, and use the validating web hook to enforce the structure.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	BareMetal *BareMetal `json:"bareMetal"` // mandatory for now, possibly optional in future
+
+	// Networking is the configuration for the pod network provider in
+	// the cluster.
+	Networking *Networking `json:"networking,omitempty"`
+
+	// Proxy defines the proxy settings for the cluster.
+	// If unset, the cluster will not be configured to use a proxy.
+	// +optional
+	Proxy *Proxy `json:"proxy,omitempty"`
```
abhinavdahiya, line 20: since these fields are mostly from install config? Can we look at accepting an install-config with predetermined platform like `None` ?
Since these fields seem to be used for openshift-installer to generate the right ignition configs, we can use maybe use install config as input and require that the platform None be used in it.
assisted installer can use this install config + fields in cluster deployment to generate a valid install-config for generating ignition configs. Since we have to generate an install config and also allow users to provide some overrides, this could be used to pass on Networking and Proxy configuration which seems like is optional for most cases??

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+
+	UserManagedNetworking  bool   `json:"userManagedNetworking,omitempty"`
```
abhinavdahiya, line 22: this looks like it could fit into similar discriminator pattern for networking like mentioned here https://github.com/openshift/hive/pull/1247/files#r556741988

pkg/apis/hive/v1/agent/platform.go
abhinavdahiya: why is this named additional ? Is this added to the existing RHCOS embeded one or does it overwrite to make this the only ntp source. If this is latter i think we should just rename to NTP.
also make it a struct since it seems like it will be more than one field like sources, frequency etc..

pkg/apis/hive/v1/agent/platform.go
abhinavdahiya: personally i would say instead of overrides we use something like installconfigtemplate to support something like https://github.com/openshift/hive/pull/1247/files#r556745975

Overrides is a patch semantic and it is difficult to validate imo... template has more similarity with reality that we use this as input and replace whatever we need to get to a valid state.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	Cidr string `json:"cidr,omitempty"`
+	// host ids
+	HostIds []string `json:"hostIDs"`
+}
+
+// ClusterStatus defines the observed state of Cluster
+type ClusterStatus struct {
+	State                        string              `json:"state,omitempty"`
+	StateInfo                    string              `json:"stateInfo,omitempty"`
+	HostNetworks                 []HostNetwork       `json:"hostNetworks,omitempty"`
```
abhinavdahiya, line 122: we allow setting machineNetworking in the spec, this field seems to be in conflict with that..?
since users boot machines in some network, we store that fact here in status, what is the goal of machineNetworks in spec?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+type HostNetwork struct {
+	// cidr
+	Cidr string `json:"cidr,omitempty"`
+	// host ids
+	HostIds []string `json:"hostIDs"`
+}
+
+// ClusterStatus defines the observed state of Cluster
+type ClusterStatus struct {
+	State                        string              `json:"state,omitempty"`
```
abhinavdahiya, line 120: this should be enum

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	StateInfo                    string              `json:"stateInfo,omitempty"`
```
abhinavdahiya, line 121: i think we should replace state info with some conditions. the state enum is good enough to convey what state we are in, and any other info should be in conditions.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	HostNetworks                 []HostNetwork       `json:"hostNetworks,omitempty"`
+	InstallationStartTime        string              `json:"installationStartTime,omitempty"`
+	InstallationCompletionTime   *metav1.Time        `json:"installationCompletionTime,omitempty"`
+	Hosts                        int                 `json:"hosts,omitempty"`
```
abhinavdahiya, line 125: this should probably be a list of objects or list of references to another object that have info on these hosts.

and instead of just a count, we should keep count of useful categories of host, like total, ready, validating, failed etc..

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	Progress                     ClusterProgressInfo `json:"progress,omitempty"`
+	ValidationsInfo              string              `json:"validationsInfo,omitempty"`
```
abhinavdahiya, line 127: i think this should be a list like conditions.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	ConnectivityMajorityGroups   string              `json:"connectivityMajorityGroups,omitempty"`
+	LastUpdateTime               *metav1.Time        `json:"lastUpdateTime,omitempty"`
+	ControllerLogsCollectionTime *metav1.Time        `json:"controllerLogsCollectionTime,omitempty"`
+	// Display api errors
+	Error string `json:"error,omitempty"`
+	// Cluster ID
+	// TODO: probably need to work with labels
+	ID string `json:"id,omitempty"`
```
abhinavdahiya, line 135: not sure what these are for?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,136 @@
...
+	State                        string              `json:"state,omitempty"`
+	StateInfo                    string              `json:"stateInfo,omitempty"`
+	HostNetworks                 []HostNetwork       `json:"hostNetworks,omitempty"`
+	InstallationStartTime        string              `json:"installationStartTime,omitempty"`
+	InstallationCompletionTime   *metav1.Time        `json:"installationCompletionTime,omitempty"`
+	Hosts                        int                 `json:"hosts,omitempty"`
+	Progress                     ClusterProgressInfo `json:"progress,omitempty"`
+	ValidationsInfo              string              `json:"validationsInfo,omitempty"`
+	ConnectivityMajorityGroups   string              `json:"connectivityMajorityGroups,omitempty"`
+	LastUpdateTime               *metav1.Time        `json:"lastUpdateTime,omitempty"`
```
abhinavdahiya, line 129: why is this important?

pkg/apis/hive/v1/agent/platform.go
avishayt: Yes assigns VIPs using DHCP

pkg/apis/hive/v1/agent/platform.go
avishayt: The API guidelines say not to use enum:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Do not use enums. Use aliases for string instead (e.g., NodeConditionType).

pkg/apis/hive/v1/agent/platform.go
avishayt: StateInfo provides error messages that aren't all formalized

pkg/apis/hive/v1/agent/platform.go
avishayt: This is a list of all available networks

pkg/apis/hive/v1/agent/platform.go
avishayt: @mhrivnak Suggested this, the user can list hosts by cluster label

pkg/apis/hive/v1/agent/platform.go
avishayt: It can be useful to clients watching an object to quickly know if a resource was updated

pkg/apis/hive/v1/agent/platform.go
avishayt: In the REST API we refer to clusters by UUID, we need some way to map the cluster resource to UUID

pkg/apis/hive/v1/agent/platform.go
YuviGold: That's correct. Added to any predefined NTP sources (could be received by DHCP, default configuration, or manually added)

pkg/apis/hive/v1/agent/platform.go
dgoodwin: After today's call, I'm going to drop this, it sounds like this may move to a new InstallEnvironment, tied to the data needed to get an ISO created for a cluster. 

@avishayt @filanov when you add this, could you re-use this structure which matches the installconfig in openshift-installer:

```go
// Proxy defines the proxy settings for the cluster.    
// At least one of HTTPProxy or HTTPSProxy is required.    
type Proxy struct {    
    // HTTPProxy is the URL of the proxy for HTTP requests.    
    // +optional    
    HTTPProxy string `json:"httpProxy,omitempty"`    
    
    // HTTPSProxy is the URL of the proxy for HTTPS requests.    
    // +optional    
    HTTPSProxy string `json:"httpsProxy,omitempty"`    
    
    // NoProxy is a comma-separated list of domains and CIDRs for which the proxy should not be used.    
    // +optional    
    NoProxy string `json:"noProxy,omitempty"`    
}    
```

pkg/apis/hive/v1/agent/ipnet.go
dgoodwin: Taking this back to string. Thanks!

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Conditions are typically composed of a Type, a Reason, and a Message. An unformalized message is still quite likely ok there. I would agree with @abhinavdahiya, we really should get these boiled down into a handful of conditions for better compat with Kube APIs.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Will do typed string aliases style enum here.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Suspect we should drop this one, finding the agents/hosts for a cluster should be handled by some other association. We should avoid per host entries within a CRD (size concerns, large updates/etcd writes) if possible.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: :+1:

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Is it meant to literally be the last time some status changed and the object was written to etcd? The last time we attempted to reconcile even if nothing changed?

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Is this the same UUID generated by installer for each cluster? If so Hive has this for us in cd.Spec.ClusterMetadata.ClusterID.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,129 @@
...
+// to be assigned and used in a cluster.
+type Platform struct {
+
+	// BareMetal contains bare metal specific configuration for this cluster.
+	BareMetal *BareMetal `json:"bareMetal"` // mandatory for now, possibly optional in future if other platforms are supported
+
+	// Networking is the configuration for the pod network provider in
+	// the cluster.
+	// TODO: should this be optional and defaulted?
+	Networking *Networking `json:"networking"`
```
dgoodwin, line 18: @avishayt @filanov thoughts on whether this networking is required or optional with defaulting?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,129 @@
...
+	StateInfo                    string              `json:"stateInfo,omitempty"`
+	HostNetworks                 []HostNetwork       `json:"hostNetworks,omitempty"`
+	InstallationStartTime        string              `json:"installationStartTime,omitempty"`
+	InstallationCompletionTime   *metav1.Time        `json:"installationCompletionTime,omitempty"`
+	Hosts                        int                 `json:"hosts,omitempty"`
+	Progress                     ClusterProgressInfo `json:"progress,omitempty"`
+	ValidationsInfo              string              `json:"validationsInfo,omitempty"`
+	ConnectivityMajorityGroups   string              `json:"connectivityMajorityGroups,omitempty"`
+	LastUpdateTime               *metav1.Time        `json:"lastUpdateTime,omitempty"`
+	ControllerLogsCollectionTime *metav1.Time        `json:"controllerLogsCollectionTime,omitempty"`
```
dgoodwin, line 133: What controller logs is this referring to, and where are they collected?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,129 @@
...
+	// Display api errors
+	Error string `json:"error,omitempty"`
```
dgoodwin, line 135: Between State, StateInfo, Progress, ValidationsInfo, and Error, I am not entirely sure how to represent status cleanly. Would it be possible to get a good example of a serialized resource from the REST API that shows how all of these are used? Would help beat this down into something smaller, I feel like there's too much going on here for a caller to check if things are ok or not. 

ValidationsInfo seems like it would belong on Agents not the cluster?

pkg/apis/hive/v1/agent/platform.go
mhrivnak: I agree with choosing the former unless we have a very good reason otherwise.

pkg/apis/hive/v1/agent/platform.go
mhrivnak: Generally there's value in understanding how many agents are expected vs. present. For example, a replica set will show you the number of observed pods in its status, and additionally how many are ready vs. available.

That said, this number becomes less relevant after installation is started, and especially after it is complete. It may be more useful to report `agentsExpected`, `agentsDiscovered`, and `agentsReady` or something like that.

pkg/apis/hive/v1/agent/platform.go
mhrivnak: No, this ID gets created before the installer is ever called.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -234,6 +235,9 @@ type ClusterDeploymentStatus struct {
 	// +optional
 	CertificateBundles []CertificateBundleStatus `json:"certificateBundles,omitempty"`
 
+	// InstallStartedTimestamp is the time when all pre-requisites were met and cluster installation was launched.
+	InstallStartedTimestamp *metav1.Time `json:"installStartedTimestamp,omitempty"`
```
mhrivnak, line 239: s/Timestamp/Time per k8s API conventions.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,149 @@
...
+	VIPDHCPAllocation bool `json:"vipDHCPAllocation"`
+}
+
+// Networking defines the pod network provider in the cluster.
+type Networking struct {
+	// NetworkType is the type of network to install. The default is OpenShiftSDN
+	//
+	// +kubebuilder:default=OpenShiftSDN
+	// +optional
+	//NetworkType string `json:"networkType,omitempty"`
```
mhrivnak, line 60: Is this commented out on purpose?

mhrivnak (changes requested):

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Is it a fully distinct UUID, or is it injected into the installer and the cluster itself for things like telemetry etc?

Can it be restored in the event the value was lost from Status? (i.e. disaster recovery with velero or similar) If not we'll need to move to Spec.

pkg/apis/hive/v1/clusterdeployment_types.go
dgoodwin: We unfortunately already used InstalledTimestamp in Hive before we learned this, though we are using Time elsewhere for other fields. Would we prefer InstalledTimestamp (which we can't change due to v1) + InstallStartedTime, or InstallStartedTimestamp bearing that discrepancy in mind? I'm ok with either.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Yes I was meaning to ask if this needs to be exposed in the first draft.

pkg/apis/hive/v1/agent/platform.go
avishayt: The user can use this information to select a proper machine CIDR (and the service validates that all hosts are connected to the chosen CIDR).

pkg/apis/hive/v1/agent/platform.go
avishayt: Makes sense to me

pkg/apis/hive/v1/agent/platform.go
avishayt: State is just the state in the state machine.
StateInfo is just a string with some extra information regarding that state that could be helpful to the user.
Progress is used only during the actual installation part to report on the installation progress.
ValidationsInfo is used only during the discovery phase (before the actual installation part).
Error is something that we added for the k8s API to display API errors.

pkg/apis/hive/v1/agent/platform.go
avishayt: It is a distinct UUID.  And you're right, it cannot be recovered.

pkg/apis/hive/v1/agent/platform.go
avishayt: This is a bad name.  Once we have the two-node control plane and are waiting for the third master to pivot, we run a k8s job to monitor and report the remainder of the progress (because at this point we have no more agents running).  So here we include the logs of the job itself (e.g. waiting for the 3rd master to join, waiting for console, waiting for CVO), ad well as the output of `oc must-gather`.

pkg/apis/hive/v1/agent/platform.go
markmc: +1

There doesn't appear to be anything specific to agent-based for these fields, and ClusterDeployment already has installConfigSecretRef - that should be the starting point for addressing this use case

pkg/apis/hive/v1/agent/platform.go
markmc: See https://github.com/openshift/enhancements/pull/574#discussion_r551417090

It's a very common requirement for users to need to be able to specify a different NTP source than the default - for example, AIUI we have no access to external NTP servers from within Red Hat and have to configure our installs to use an internal NTP server

So, while for agent-based installs we need to configure the pre-install machines the agent is running on, and also the machines post-install ... there is nothing specific to agent-based installation about needing to configure an NTP server

IMO this should be at the ClusterDeployment level, not platform-specific

avishayt: I agree, it's not platform-specific.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Networking is the configuration for the pod network provider in
+	// the cluster.
+	// TODO: should this be optional and defaulted?
+	Networking Networking `json:"networking"`
+
+	// UserManagedNetworking indicates if the networking is managed by the user.
+	// TODO: more detail needed here.
+	// +optional
+	UserManagedNetworking bool `json:"userManagedNetworking,omitempty"`
```
markmc, line 24: I found https://github.com/openshift/assisted-service/pull/674 and https://issues.redhat.com/browse/MGMT-2752 but there's not a lot of discussion in those about the "why"

As best I can make out, there effect of this is to use the 'none' platform and not require a machine network CIDR to be specified. Also that the API and Ingress VIP is not required?

This seems to be for 2 specific use cases - (a) remote worker nodes deployments where the worker nodes are not on the same L2 network as the control plane nodes, and (b) single-node deployments

Hard to see any of this as being specific to the agent-based platform

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Ok then for now lets drop from API and we'll keep that assumption internal for time being. We're looking at possibly taking an optional install config template elsewhere as an input which would cover this use case if it ever needed to be explicit.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: I will try to sketch something out in the next couple days.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Would like to discuss this one a little in person sometime this week.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: We could combine the first two into a condition, Type = AgentInstallState, Reason = your set of supported states, Message = StateInfo.

Progress also feels like a condition.

ValidationsInfo feels like it belongs on Agents possibly? Or if cluster wide then another condition.

I'm unsure where or how Error fits in, or why it wouldn't fall into one of the previous.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Ok I will move this to Spec somewhere.

pkg/apis/hive/v1/agent/platform.go
avishayt: What are the criteria for condition vs. property in status?  For example, it looks like in BMH I can find the provisioning state in the status.  (I have no problem moving things, just trying to learn)
As for the validations, these are validations at the cluster level (e.g., if required cluster parameters are defined and valid, has enough masters, etc.).  We have other validations on agents.

pkg/apis/hive/v1/agent/platform.go
mhrivnak: Since we need this for the discovery ISO (so that clocks can be synchronized prior to starting installation), and since the discovery ISO will not be cluster-specific in the future, do we need this on the cluster definition at all? NTP will presumably be on the `InstallEnvironment` (or similar name) resource we've been discussing, and used there in the generation of an ISO.

pkg/apis/hive/v1/clusterdeployment_types.go
mhrivnak: Gotcha, either seems reasonable then. It's a hive API, so I'm happy to defer to your team's preference.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Interesting tidbit from https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

*"Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility. Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See #7856 for more details and discussion."*

Based on that, I think we drop the State enum as well.

@avishayt @filanov could one of you add the list of valid states here so we can discuss how to model.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: I have replaced hosts with the above three distinctions.

pkg/apis/hive/v1/clusterdeployment_types.go
dgoodwin: With it's proximity/similarity to the pre-existing InstalledTimestamp, it probably looks a little cleaner to do it wrong again and use Timestamp again within that CRD. I'll leave a TODO to correct them both if we ever go to a V2.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: We will replace with a structure that defines minimum required control plane hosts (1 or 3), and a label selector to associate Agents, and same for workers. Agent CRD will have a role field, so the label selector can be used for both.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Assisted service does it's own log gathering if failed, doesn't require a private key. 

Sounds like we do want to expose just a public ssh key in this scenario. Likely should move higher up in Hive cluster deployment. We do not want to force to give a private key.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Notes from call: Assisted service today does not want to require users to have to provide an install config, this allows for better parameter validation etc. It does support (arbitrary) overrides. 

Install config and ignition config today are generated only when a user confirms to proceed with actual install.

InstallEnvironment could have an install config template as a starting point. Users can either retrieve this, add customizations, and provide directly to hive exactly as it works today. Hive does not know about the existence of InstallEnvironment. If none is provided, we generate one for you.

Today, virtually no one uses the install config and ignition config overrides and customizations, they just use the generated one assisted handles for them. Group on the call proposes that we drop this for now, it will not be possible to customize or provide your own install/ignition config directly with the Kube API, and we can add later if needed.

This means that configuration we have requirements for must get modeled directly in a Kube API. Proxy is not cluster specific and should fit well in the proposed InstallEnvironment CRD.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,140 @@
...
+type Platform struct {
+
+	// Provider contains settings for different providers for which an agent based cluster installation
+	// can be performed.
+	Provider Provider `json:"provider"`
+
+	// Networking is the configuration for the pod network provider in
+	// the cluster.
+	// TODO: should this be optional and defaulted?
+	Networking Networking `json:"networking"`
```
dgoodwin, line 18: We propose to keep this one on ClusterDeployment as it is per cluster, not a good fit for InstallEnvironment and does not go into an installation ISO.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: We agreed on call that ClusterImageSet is effectively authoritative. There are a lot of open questions on how this flows to assisted, whether we can override and pass each setting to assisted installer, or we automatically generate the configmap for them, or if there should be some tighter integration between assisted controllers working directly with ClusterImageSet directly.

However for purposes of API, we will rely on ClusterDeployment.Spec.ClusterImageSetRef and work our the rest of the details beneath the covers over time.

pkg/apis/hive/v1/agent/platform.go
markmc: It's pretty easy to imagine at some point that Hive will need to have an answer for "what CRD should I use to configure my company's proxy so that it is used for all installs run by Hive?"

But InstallEnvironment wouldn't be that API? It's an API purely for configuring the generated discovery ISO?

(Maybe I have missed where the InstallEnvironment idea is written down somewhere? TIA)

dgoodwin: I don't think it's documented yet just an idea we've been discussing a lot. We should probably change that.

Loosely it sounded to me like it was "what is required to generate an ISO". Today it would be cluster specific but hopefully not within a few releases. Things like proxy, MAC address to static IP mappings, etc. Others can probably cover it better than I, but we should sketch something out soon and decide where it would live.

markmc: Ok, well it sounds to me like Hive inevitably should have an API for stuff like proxy and NTP that goes beyond the "what is required to generate an ISO" semantics

avishayt: I would say it can go beyond "what is required to generate an ISO".  It should contain properties of the environment as opposed to properties for a specific cluster installation.
Rule of thumb - I should be able to create different clusters from the InstallEnvironment.  So between installations, or if we have multiple installations from the same InstallEnvironment, we would have common settings for proxy, NTP, static IPs, etc.  We would have different parameters for the cluster's networking, for example, so that goes into the ClusterDeployment .

pkg/apis/hive/v1/agent/platform.go
dgoodwin: My team has flated the idea of this being an "InstallStrategy" instead of a "Platform", leaving "Platform" open to alternate implementations beyond bare metal. Not totally sure how to structure this, we already have a baremetal platform which has some similarities to what Agent metal installs want, but not much, and most of it would be incompatible. This would like mean an "Agent" install strategy and a new "AgentBareMetal" platform.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Keeping this here for now, flagged for possibly moving up to higher in Hive stack someday before we go stable.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: From call today this looks ok, will make the change asap.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Removed as we expect to go to InstallEnvironment CRD.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Property means "use none platform in install config" instead of bare metal. Useful for single-node clusters. 

We think we could infer this by the new provisionRequirements, if control plane count is 1 and workers is 0 or undefined.

Will remove.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: This would be used for a user to select the machine network after letting the assisted installer discover all the available options.

Going to keep this property for now.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: status:
        type: string
        description: Status of the OpenShift cluster.
        enum:
          - insufficient
          - ready
          - error
          - preparing-for-installation
          - pending-for-input
          - installing
          - finalizing
          - installed
          - adding-hosts
          - cancelled
          - installing-pending-user-action

dgoodwin: We will add one condition per each of these states. Assisted controllers will translate their internal state machine into a set of these conditions.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Examples: 

      - 'machine-cidr-defined'
      - 'cluster-cidr-defined'
      - 'service-cidr-defined'
      - 'no-cidrs-overlapping'
      - 'network-prefix-valid'
      - 'machine-cidr-equals-to-calculated-cidr'
      - 'api-vip-defined'
      - 'api-vip-valid'
      - 'ingress-vip-defined'
      - 'ingress-vip-valid'
      - 'all-hosts-are-ready-to-install'
      - 'sufficient-masters-count'
      - 'dns-domain-defined'
      - 'pull-secret-set'
      - 'ntp-server-configured'
      - 'lso-requirements

Condition type: ValidatedNTPServerConfig, True/False/Unknown, Reason: SomethingWrong, Message: "Something is wrong see http://blah.com"

May also have a condition for overall full validation success true/false.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: @filanov  confirms no longer needed assisted is just storing namespace+name key in db to identify.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Kube updates generation on changes. With this we likely do not need last update time for now.

dgoodwin: Some controllers do require setting lastObservedGeneration in status to track if they've caught up to the current generation.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: @abhinavdahiya if we send an install config without network cidrs set, do the defaults kick in and work?

dgoodwin: We're going to likely keep mandatory for now.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: We suspect this would be covered by the LogsCollected condition's LastProbeTime if state is true. This is mostly just used for debugging.

Will drop this property.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: 400-500 errors from the assisted service backend.

Condition: Type: AssistedServiceError, Reason: Error, Message: 504 error received

dgoodwin: We hope to eliminate this condition in the future.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Proxy is removed for now likely moving to InstallEnvironment.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Property was removed.

pkg/apis/hive/v1/agent/status.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// AgentNetwork contains infromation about a detected host network on the agents for this cluster.
+type AgentNetwork struct {
+
+	// CIDR is the agent network CIDR.
+	CIDR string `json:"cidr,omitempty"`
+
+	// AgentRefs contains the namespace and name of each Agent resource associated with the cluster.
+	AgentRefs []AgentRef `json:"agentRefs"`
```
dgoodwin, line 33: Switched these to explicit references rather than a vague ID.

pkg/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -234,12 +240,30 @@ type ClusterDeploymentStatus struct {
...
 	// InstalledTimestamp is the time we first detected that the cluster has been successfully installed.
 	InstalledTimestamp *metav1.Time `json:"installedTimestamp,omitempty"`
 
 	// ProvisionRef is a reference to the last ClusterProvision created for the deployment
 	// +optional
 	ProvisionRef *corev1.LocalObjectReference `json:"provisionRef,omitempty"`
+
+	// InstallStrategy contains observed state from specific install strategies.
+	// +optional
+	InstallStrategy *InstallStrategyStatus `json:"installStrategy,omitempty"`
```
dgoodwin, line 258: Agent install status links into overall ClusterDeployment status here.

dgoodwin (commented):
@filanov @avishayt @mhrivnak @markmc @abhinavdahiya please take a hopefully close to final look, we think we're narrowing in on completion here. I can't merge until Abhinav's work is there to feature gate it, so we'll be on hold until then but assisted team can still pick it up from this branch in the interim.

Next steps for me are to get the hive controllers ignoring agent clusters for now, we may need parts of those controllers later on.

And then to setup webhook validation and start filling it in for what we've got here. That would purely be for static sanity checking incoming data, not anything dynamic.

dgoodwin:
I have updated PR to incorporate feature gating, move the InstallStrategy under Provisioning, and skip processing of agent based clusters.

@abhinavdahiya can you take another pass and help me start working this towards a merge?

pkg/apis/hive/v1/agent/installstrategy.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// InstallStrategy is the install strategy configuration for provisioning a cluster with the
+// Agent based assisted installer.
+type InstallStrategy struct {
+	// Networking is the configuration for the pod network provider in
+	// the cluster.
+	Networking Networking `json:"networking"`
```
abhinavdahiya, line 12: why is networking part of the strategy? to me it looks like a platform configuration?

can you add a new lines describing the difference in strategy and platform somewhere?

pkg/apis/hive/v1/agent/installstrategy.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// TODO: consider moving higher into ClusterDeployment in future?
```
abhinavdahiya, line 14: drop todo comment

pkg/apis/hive/v1/agent/installstrategy.go
```diff
@@ -0,0 +1,84 @@
...
+	AgentSelector metav1.LabelSelector `json:"agentSelector"`
+}
+
+// Networking defines the pod network provider in the cluster.
+type Networking struct {
+	// MachineNetwork is the list of IP address pools for machines.
+	// This field replaces MachineCIDR, and if set MachineCIDR must
+	// be empty or match the first entry in the list.
+	// Default is 10.0.0.0/16 for all platforms other than libvirt.
+	// For libvirt, the default is 192.168.126.0/24.
```
abhinavdahiya, line 47: i think this needs rewording.

pkg/apis/hive/v1/agent/installstrategy.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// WorkerAgents is the minimum number of matching approved and ready Agents with the worker role
+	// required to launch the install.
+	// +optional
+	WorkerAgents int `json:"workerAgents,omitempty"`
+
+	// AgentSelector is a label selector used for associating relevant custom resources with this cluster.
+	// (Agent, BareMetalHost, etc)
+	AgentSelector metav1.LabelSelector `json:"agentSelector"`
+}
```
abhinavdahiya, line 39: defining how to select from space and the constraint of defining provision complete in same struct seems overloaded. I think maybe we should move the selector out to Strategy?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,14 @@
+package agent
+
+// Platform defines agent based install configuration specific to bare metal clusters.
```
abhinavdahiya, line 3: this comment ties the generically named agent.Platform to bare metal. that doesn't sound right..

pkg/apis/hive/v1/agent/status.go
```diff
@@ -0,0 +1,44 @@
...
+	WorkerAgentsReady int `json:"workerAgentsReady,omitempty"`
+
+	ConnectivityMajorityGroups string `json:"connectivityMajorityGroups,omitempty"`
+}
+
+// AgentNetwork contains infromation about a detected host network on the agents for this cluster.
+type AgentNetwork struct {
+
+	// CIDR is the agent network CIDR.
+	CIDR string `json:"cidr,omitempty"`
```
abhinavdahiya, line 30: this should compose the MachineNetworkEntry inline?

pkg/apis/hive/v1/agent/status.go
```diff
@@ -0,0 +1,44 @@
+package agent
+
+// InstallStrategyStatus defines the observed state of the Agent install strategy for this cluster.
+type InstallStrategyStatus struct {
+
+	// AgentNetworks provides an aggregate view of all network CIDRs detected across all Agents.
```
abhinavdahiya, line 6: > aggregate view

what type of aggregation is happening to agent (its ip address and network) and one item in this list? 
Is it just a 
- list of each agent and it ip/network or 
- list of aggregated newtorks and agents (more than one) in that network.

i think we should probably clarify the expectations here.

pkg/apis/hive/v1/agent/status.go
```diff
@@ -0,0 +1,44 @@
...
+	// CIDR is the agent network CIDR.
+	CIDR string `json:"cidr,omitempty"`
+
+	// AgentRefs contains the namespace and name of each Agent resource associated with the cluster.
+	AgentRefs []AgentRef `json:"agentRefs"`
+}
+
+// AgentRef defines the namespace and name of an Agent resource.
+type AgentRef struct {
+
```
abhinavdahiya, line 38: previously i saw that the agents for provision requirement used selector to Agent CRD objects or BareMetalHost CRD objects ... is this Agent CRD object only?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,14 @@
...
+type Platform struct {
+	// APIVIP is the virtual IP used to reach the OpenShift cluster's API.
+	APIVIP string `json:"apiVIP"`
+	// APIVIPDNSName is the domain name used to reach the OpenShift cluster API.
+	APIVIPDNSName string `json:"apiVIPDNSName"`
+	// IngressVIP is the virtual IP used for cluster ingress traffic.
+	IngressVIP string `json:"ingressVIP"`
+	// VIPDHCPAllocation indicates if virtual IP DHCP allocation mode is enabled.
+	// +optional
+	VIPDHCPAllocation bool `json:"vipDHCPAllocation"`
```
abhinavdahiya, line 13: personally i would recommend not using `boolean`, this types of field almost always end up growing in valid states.

make it a string enum,
```
Disabled ""
Enabled "Enabled"
```
to capture the same default of empty type value and expand later on.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -454,6 +456,14 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
 	}
 
+	// Return early and stop processing if Agent install strategy is in play. The controllers that
+	// handle this portion of the API currently live in the assisted service repo, rather than hive.
+	// This will hopefully change in the future.
+	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallStrategy != nil && cd.Spec.Provisioning.InstallStrategy.Agent != nil {
+		cdLog.Info("skipping processing of agent install strategy cluster")
+		return reconcile.Result{}, nil
+	}
+
```
abhinavdahiya, line 466: for now this might be okay, but i think we will have to define a more appropriate boundary because this controller does a lot more than just provisioning...

pkg/apis/hive/v1/agent/installstrategy.go
dgoodwin: My thinking was that Networking is likely to be used for any agent based install strategy, and isn't bare metal specific. I actually really wonder if this shouldn't be in the ClusterDeployment itself, to move towards what we talked about a few weeks ago where we might make provising an install config optional, and generate one for you if you don't. We have the option to do that given this is behind an alpha gate, but it did seem specific to the use of the agent installer rather than the agent metal platform. I'm not sure how to convey that in comments in a way that's relevant to an API user but I'll give it a shot.

pkg/apis/hive/v1/agent/installstrategy.go
dgoodwin: Good point, moving up a level.

pkg/apis/hive/v1/agent/installstrategy.go
dgoodwin: I originally copied out of installer directly, but those libvirt mentions do seem out of place here. Removing.

dgoodwin: Ah and removing the mention of the legacy field we didn't carry.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: I guess in this case it should be bare metal platform, as we're not repeating a package name.

pkg/apis/hive/v1/agent/platform.go
dgoodwin: Good idea.

pkg/apis/hive/v1/agent/status.go
dgoodwin: Need your help here @avishayt @filanov @mhrivnak , could one of you send me some godoc text for this?

pkg/apis/hive/v1/agent/status.go
dgoodwin: "be composed of"? Seems reasonable. Changing it over.

dgoodwin: Switched over, named machineNetwork, and dropped the omitempty.

pkg/apis/hive/v1/agent/status.go
dgoodwin: In this context I believe it's agent only, as this is used in discovered agent machine networks, it would be specific to an agent and not a bare metal host that may have preceded it. Is this right?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Yes I have a card roughly about this for next sprint, you are right, we need to think carefully about this and make sure that hive still does what it should for an agent based cluster.

I also want to discuss a little more about the plans to get the controllers doing this work into hive.

cc @filanov FYI

dgoodwin:
Updated.

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,26 @@
+package agent
+
+// BareMetalPlatform defines agent based install configuration specific to bare metal clusters.
+// Can only be used with spec.installStrategy.agent.
+type BareMetalPlatform struct {
+	// APIVIP is the virtual IP used to reach the OpenShift cluster's API.
+	APIVIP string `json:"apiVIP"`
+
+	// APIVIPDNSName is the domain name used to reach the OpenShift cluster API.
+	APIVIPDNSName string `json:"apiVIPDNSName"`
```
filanov, line 10: Why is that required? in assisted we are setting it only for day2 clusters.

dgoodwin: Updated to optional.

pkg/apis/hive/v1/agent/status.go
```diff
@@ -0,0 +1,44 @@
...
+	ConnectivityMajorityGroups string `json:"connectivityMajorityGroups,omitempty"`
+}
+
+// AgentNetwork contains infromation about a detected host network on the agents for this cluster.
+type AgentNetwork struct {
+
+	// MachineNetwork contains the agent network CIDR.
+	MachineNetwork MachineNetworkEntry `json:"machineNetwork"`
+
+	// AgentRefs contains the namespace and name of each Agent resource associated with the cluster.
```
markmc, line 32: Is this the set of agents which currently match the agent label selector?

And the list becomes empty when the install completes?

And it's an aggregation of data available on the Agent CRs?

And its primary use is to help you complete a required field (machine network) in this CR before the CR can be reconciled?

That all seems pretty strange, no?

markmc:
Something seems odd to me about fields like API VIP, Ingress VIP (in the platform) and the cluster/machine/service networks (in the agent install strategy)

For other platforms, this sort of configuration would be supplied by the required install-config. And the CLI is one way this is made less painful.

But especially for agent-based installs, we are saying install-config is not required, and we're adding API fields for this stuff.

Are we sure we won't add API fields for this stuff in future in a cross-platform way?

config/crds/hive.openshift.io_clusterdeployments.yaml
```diff
@@ -617,6 +642,142 @@ spec:
                         TODO: Add other useful fields. apiVersion, kind, uid?'
                       type: string
                   type: object
+                installStrategy:
+                  description: InstallStrategy provides platform agnostic configuration
+                    for the use of alternate install strategies. Defaults to openshift-install
+                    if none specified.
```
markmc, line 648: How should we conceptualize the openshift-install InstallStrategy?

e.g. it means "clusters created by running `openshift-install create cluster`" ?

So a future installStrategy.openshiftInstall field would contain stuff only relevant to `openshift-install create cluster` type deployments?

dgoodwin: Yes that is what we had in mind I believe, essentially control over the implementation used to deploy the cluster, and settings specific to just one stategy would nest within.

dgoodwin:
> Something seems odd to me about fields like API VIP, Ingress VIP (in the platform) and the cluster/machine/service networks (in the agent install strategy)
> 
> For other platforms, this sort of configuration would be supplied by the required install-config. And the CLI is one way this is made less painful.
> 
> But especially for agent-based installs, we are saying install-config is not required, and we're adding API fields for this stuff.
> 
> Are we sure we won't add API fields for this stuff in future in a cross-platform way?

APIVIP and IngressVIP were not on my radar as a problem, they do mirror similar settings in the pre-existing bare metal platform in hive, but we made an agent bare metal platform so we could avoid the huge amount of other settings tied up in that platform that are not relevant for agent installs.

Networking has been bothering me. I am not entirely happy with the pre-existing convention that Hive requires an InstallConfig, and as a team we discussed and a long term goal we'd like to get to would be making that optional. We'd like hive to be able to generate an install config for you, but if you want you can specify your own,. (possibly with cluster deployment values possibly getting injected into the installconfig you provide)

Networking in particular I suspect should move up into the main hive ClusterDeployment.Spec and the door is open for that given this is behind an alpha feature gate. (even if we block it's usage for openshift-install clusters for now)

Are there other examples?

pkg/apis/hive/v1/agent/status.go
dgoodwin: > Is this the set of agents which currently match the agent label selector?

It would be the set that match the label selector, and the machine network here.

> 
> And the list becomes empty when the install completes?

We should probably wipe it all out once machine network is set, or the cluster is marked installed.

> 
> And it's an aggregation of data available on the Agent CRs?

My understanding was just a cidr and all agents that match it.

> 
> And its primary use is to help you complete a required field (machine network) in this CR before the CR can be reconciled?
> 
> That all seems pretty strange, no?

This is definitely not a pattern I'm familiar with. Is it a use case we need to satisfy with the kube API? Or could we assume users must give us a machine network before we start, and we don't need to provide this tooling to help you find one? 

Is use of a default machine network like openshift-install does out of the question for metal clusters? I'm guessing it is.

mhrivnak: I agree with the sentiments, and we may be able to get by without this pattern at all. The high-level idea is that each agent discovers which subnets it is connected to, then those get reduced on the cluster resource as a list of unique subnets available on all related agents, thus making it easy for the user to pick one. It's useful in the interactive wizard UX, but not so useful for a declarative (or especially gitops) approach to cluster creation.

@avishayt @ronniel1 WDYT about omitting this list of networks for now?

filanov: If it's expected from the user to aggregate this information from the hosts then it's ok.
You are correct that it was added as a UX feature to help the users select the right network

dgoodwin: Decision taken to remove this, it's too imperative and that's not our primary use case right now. We will require the machine cidr at creation time.

dgoodwin:
Two new commits based on today's conversations. Machine network is now required, and the agent network aggregation is gone in status.

pkg/apis/hive/v1/agent/installstrategy.go
```diff
@@ -0,0 +1,77 @@
...
+	// WorkerAgents is the minimum number of matching approved and ready Agents with the worker role
+	// required to launch the install.
+	// +optional
+	WorkerAgents int `json:"workerAgents,omitempty"`
+}
+
+// Networking defines the pod network provider in the cluster.
+type Networking struct {
+	// MachineNetwork is the list of IP address pools for machines.
+	// Default is 10.0.0.0/16.
```
abhinavdahiya, line 42: is default valid for required field ?

pkg/apis/hive/v1/agent/installstrategy.go
```diff
@@ -0,0 +1,77 @@
...
+	MachineNetwork []MachineNetworkEntry `json:"machineNetwork"`
+
+	// ClusterNetwork is the list of IP address pools for pods.
+	// Default is 10.128.0.0/14 and a host prefix of /23.
```
abhinavdahiya, line 46: is default valid for required field ?

pkg/apis/hive/v1/agent/platform.go
```diff
@@ -0,0 +1,27 @@
...
+	// IngressVIP is the virtual IP used for cluster ingress traffic.
+	IngressVIP string `json:"ingressVIP"`
+
+	// VIPDHCPAllocation indicates if virtual IP DHCP allocation mode is enabled.
+	// +optional
+	VIPDHCPAllocation VIPDHCPAllocationType `json:"vipDHCPAllocation"`
+}
+
+// VIPDHCPAllocationType is a valid value for bareMetalPlatform.vipDHCPAllocation.
+type VIPDHCPAllocationType string
```
abhinavdahiya, line 22: should have the kubebuilder enum `kubebuilder:validation:Enum`

abhinavdahiya, line 22:
```suggestion
// +kubebuilder:validation:Enum="";Enabled
type VIPDHCPAllocationType string
```

dgoodwin:
Updated.

PR #1252, opened 2021-01-13
add new prometheus metric to indicate that the cluster provision failed after all the retries

bmeng:
Currently, the alert will be fired after 2 hours, but the cluster provision will retry for 3 times, which will exceed the 2 hours.

Maybe the better solution here is fire the ClusterProvisionDelay as warning, which will not page.
And the use the new metric to alert like ClusterProvisionFailed, which is the real problem people should work on.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -663,7 +663,7 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		}
 	}
 
-	r.deleteStaleProvisions(existingProvisions, cdLog)
+	r.deleteStaleProvisions(existingProvisions, cd, cdLog)
 
 	if cd.Spec.ManageDNS {
 		dnsZone, err := r.ensureManagedDNSZone(cd, cdLog)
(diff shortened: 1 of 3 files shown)


gregsheremeta:
/hold

needs to wait for https://issues.redhat.com/browse/SDA-3037

gregsheremeta:
/unhold

looks like it doesn't *remove* the existing metric, so no need to wait :)

dgoodwin:
Hi @bmeng , big thanks for attempting to fix and submitting.


The location you have added the alert is not actually looking at the attempts limit rather how many provisions we keep in etcd for debugging, so this is not quite the right location.

ClusterDeployments have a condition when the install attempts limit is reached: https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L702 (condition would have type ProvisionStopped and reason InstallAttemptsLimitReached)

A tally of clusters with each condition is reported today in prometheus, but this would only tell you how many clusters were experiencing this problem, not which, and that's probably not good enough for what you want.

I think what you want is to go to pkg/controllers/metrics/ possibly in provision_underway_collector.go, and to check the ClusterDeployments being processed for that condition. If set, you want to add a new metrics indicating a cluster had a provision failed. I would kind of like this to include any detected error messages just as we do for provision underway. @suhanime is the author of this code and may have some ideas on whether or not this is the same situation that required the provision_underway_collector, and if your change would go *into* that collector, or be in another similar collector. 

Bear in mind we need to make sure the metric gets cleared in a similar fashion. (most likely when the cluster is deleted in this case)

suhanime:
If the `InstallAttemptsLimit` is set on ClusterDeployment, then we are already handling that in the code by setting the condition in the ClusterDeployment controller, so we can and should add the condition [here](https://github.com/openshift/hive/blob/master/pkg/controller/metrics/provision_underway_collector.go#L20), this would ensure we have some additional info when the alert fires. I would add it to the bottom of the map because the first condition that is true is set, so for ex, `InstallImagesNotResolvedCondition` should take presence over the `ProvisionStoppedCondition`

However, for clusters that do not have `InstallAttemptsLimit` set, it is a bit tricky. Technically this fits under the bracket of delay in cluster provision, there by it should be in `provision_underway_collector`, but I am worried about the cardinality of `metricClusterDeploymentProvisionUnderwaySeconds`. 
In this situation, I would not want to use the custom collector at all, because the new metric will have all the labels necessary in a predictable fashion to handle deleting it. So I would add it to metrics collector, but I would set it [here](https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1075-L1083), and also ensure it is cleaned up when provision succeeds. This will ensure the metric is scraped only when the value changes.

suhanime:
@bmeng After discussions with the team, we realized adding the metric in this fashion will not make the provision information available, like ClusterDeployment conditions and reasons.

So I've logged a [ticket](https://issues.redhat.com/browse/HIVE-1341) that we could take up in the future sprint. Please provide us additional information and context there to help us work on it.

dgoodwin:
Replaced by #1275 

/close

PR #1253, opened 2021-01-14
CO 1301: Read additional install log regexes from configmap

abutcher:
Add support for reading additional, optional install log regexes from a separate configmap `additional-install-log-regexes`. Additional regexes are tested after regexes defined in the `install-log-regexes` configmap.

This change also allows users with the `hive-reader` role to modify the `additional-install-log-regexes` configmap.

/cc @dgoodwin


config/rbac/hive_reader_role.yaml
```diff
@@ -68,3 +68,13 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - "configmaps"
+  resourceNames:
+  - "install-log-regexes"
+  verbs:
+  - get
+  - patch
+  - update
```
dgoodwin, line 80: Can this limit just to the hive namespace?

pkg/controller/clusterprovision/installlogmonitor.go
```diff
@@ -51,6 +52,16 @@ func (r *ReconcileClusterProvision) parseInstallLog(log *string, pLog log.FieldL
 		return unknownReason, regexBadMessage
 	}
 
+	additionalRegexesRaw, _ := regexCM.Data[additionalRegexDataEntryName]
+
+	additionalRegexes := []installLogRegex{}
+	if additionalRegexesRaw != "" {
+		if err := yaml.Unmarshal([]byte(additionalRegexesRaw), &additionalRegexes); err != nil {
+			pLog.WithError(err).Errorf("cannot unmarshal data from %s configmap", regexConfigMapName)
```
dgoodwin, line 60: Lets clarify in the error message that it's something specific to the additional regexes key.

dgoodwin (commented):
I don't see anything related to the operator ownership of the config map. Is this working as is? i.e. you deploy hive, you get the default regexes, but you can add more additionalRegexes, restart the operator, and it will not overwrite them?

config/rbac/hive_reader_role.yaml
abutcher: I've done this with a separate rolebinding but it may not be necessary with our plans to integrate with config management.

config/rbac/hive_regex_role.yaml
```diff
@@ -0,0 +1,16 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - "configmaps"
+  resourceNames:
+  - "additional-install-log-regexes"
+  verbs:
+  - get
+  - patch
+  - update
```
dgoodwin, line 16: If this is managed in app-interface, I don't think we need patch/update here, and we already have get. Feels safe to drop the role changes, wdyt?

abutcher: I agree. I'll remove them.

config/rbac/hive_reader_role.yaml
dgoodwin: Still a role change in here. :)

PR #1254, opened 2021-01-20
Add logo

gregsheremeta:
Add Hive logo to:
 - CSV template for operatorhub
 -  the docs directory, for posterity

/assign @twiest 
/cc @joelddiaz


gregsheremeta:
yeah it's actually live already

Selection_999(014)

PR #1255, opened 2021-01-21
add issue template to tell people to use issues.redhat.com

gregsheremeta:
add issue template to tell people to use issues.redhat.com. We don't use this tracker, and if people open things here, we usually don't see them.


gregsheremeta:
/close

going to disable issues instead

PR #1259, opened 2021-01-25
HiveConfig allow setting feature gates

abhinavdahiya:
### hivev1: add feature gate selection to hiveconfig

Allow adding feature gates to the hiveconfig. The feature gate selection
allows choosing feature sets, currenty which are default("") or "custom".
A feature set, except custom, allows hive owners to define a set of enabled/disabled
feature gates the users can directly pick.

Default feature set currently has no feature gates enabled.
```yaml
spec:
  featureGates:
    featureSet: ""
```

Custom feature set allows the users to enabled any list of feature gates. This is very
fragile and should be discouraged in favor of using pre defined feature sets.
```yaml
spec:
  featureGates:
    featureSet: "Custom"
    custom:
      enabled:
      - AlphaPlatformA
      - AlphaPlatformB
```

This mimics the OpenShift featuregate api [1]

[1]: https://github.com/openshift/api/blob/master/config/v1/types_feature.go

###  hivev1 validating-webhooks: allow adding checks for fields in clusterdeployment based on feature gates

the webhook looks for feature gates enabled as a comma separate list of feature gates
set as env HIVE_FEATURE_GATES_ENABLED

Adds 2 methods to help validating
- making sure certain field is set only when a feature gate was enabled
- maling sure certain field is might be equal in value to certain value only when
  a feature gate was enabled.
These help the most common use case of making sure certain platforms are only set when
the alpah feature gates are set.

The webhook reads the env for every request so that it can respond to changing feature gates
without reintialization.
If validation fails due to missing feature gate, the object is immediately rejected to prevent
running any code later on that assumes features gates to be set.

###  hive operator: sync featuregate config map and hive admission to use it

the operator now creates and manages a config map hive-feature-gate to allow
operands to fetch which feature gates are enabled. The operands can load the
configmap as env using
```yaml
envFrom:
  configMapRef:
    name: hive-feature-gate
```
Currently the k8s api doesn't support auto updating the env when the config map
changes, see [2], so the operator stores the hash of the config map data on the
deployment as annotation. This ensures that the deployment rolls when the data
changes in the config map.

Currently only the hive-admission deployment is wired to get the feature gates.

[2]: https://kubernetes.io/docs/concepts/configuration/configmap/#mounted-configmaps-are-updated-automatically

xref: https://issues.redhat.com/browse/HIVE-1335

/assign @dgoodwin


abhinavdahiya:
Also https://asciinema.org/a/ousTc1AuyJEZygOyTvn7Qzn7u should help show how to use these feature gates.

pkg/apis/hive/v1/validating-webhooks/clusterdeployment_validating_admission_hook.go
```diff
@@ -277,6 +284,38 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec
...
+		}
+	}
+
+	fs := newFeatureSet()
+	contextLogger.WithField("enabled_feature_gates", fs.Enabled).Info("feature gates enabled")
+
+	errs := field.ErrorList{}
+	// To add validation for feature gates use these examples
+	// 		errs = append(errs, equalOnlyWhenFeatureGate(fs, obj, "spec.platform.type", "AlphaPlatformAEnabled", "platformA")...)
+	// 		errs = append(errs, existsOnlyWhenFeatureGate(fs, obj, "spec.platform.platformA", "AlphaPlatformAEnabled")...)
```
dgoodwin, line 306: Helpful for me to plug in once we merge, thanks.

dgoodwin (approved):
All I had was a nit about a log field name so I'm deleting that and approving. :) Thanks!

PR #1260, opened 2021-01-25
Delete clustersync sts when spec is changed

twiest:
- [x] Delete clustersync sts when spec.Selector is changed
- [x] Manually test fix to ensure it is working as expected
- [x] Fixes bug https://issues.redhat.com/browse/HIVE-1353

/assign @dgoodwin 
/cc @abhinavdahiya

### DIFF
--- a/pkg/operator/hive/clustersync.go
+++ b/pkg/operator/hive/clustersync.go
@@ -1,9 +1,13 @@
 package hive
 
 import (
+	"context"
+
 	log "github.com/sirupsen/logrus"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
@@ -21,8 +25,8 @@ const (
 func (r *ReconcileHiveConfig) deployClusterSync(hLog log.FieldLogger, h resource.Helper, hiveconfig *hivev1.HiveConfig, hiveControllersConfigHash string) error {
 	asset := assets.MustAsset("config/clustersync/statefulset.yaml")
 	hLog.Debug("reading statefulset")
-	clusterSyncStatefulSet := controllerutils.ReadStatefulsetOrDie(asset)
-	hiveContainer := &clusterSyncStatefulSet.Spec.Template.Spec.Containers[0]
+	newClusterSyncStatefulSet := controllerutils.ReadStatefulsetOrDie(asset)
+	hiveContainer := &newClusterSyncStatefulSet.Spec.Template.Spec.Containers[0]
 
 	hLog.Infof("hive image: %s", r.hiveImage)
 	if r.hiveImage != "" {
@@ -62,10 +66,10 @@ func (r *ReconcileHiveConfig) deployClusterSync(hLog log.FieldLogger, h resource
 
 	hiveNSName := getHiveNamespace(hiveconfig)
 
-	if clusterSyncStatefulSet.Spec.Template.Annotations == nil {
-		clusterSyncStatefulSet.Spec.Template.Annotations = make(map[string]string, 1)
+	if newClusterSyncStatefulSet.Spec.Template.Annotations == nil {
+		newClusterSyncStatefulSet.Spec.Template.Annotations = make(map[string]string, 1)
 	}
-	clusterSyncStatefulSet.Spec.Template.Annotations[hiveConfigHashAnnotation] = hiveControllersConfigHash
+	newClusterSyncStatefulSet.Spec.Template.Annotations[hiveConfigHashAnnotation] = hiveControllersConfigHash
 
 	// Load namespaced assets, decode them, set to our target namespace, and apply:
 	namespacedAssets := []string{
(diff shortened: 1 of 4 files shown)


pkg/operator/hive/clustersync.go
```diff
@@ -82,22 +86,48 @@ func (r *ReconcileHiveConfig) deployClusterSync(hLog log.FieldLogger, h resource
...
+
+	existingClusterSyncStatefulSet := &appsv1.StatefulSet{}
+	existingClusterSyncStatefulSetNamespacedName := apitypes.NamespacedName{Name: newClusterSyncStatefulSet.Name, Namespace: newClusterSyncStatefulSet.Namespace}
+	err = r.Get(context.TODO(), existingClusterSyncStatefulSetNamespacedName, existingClusterSyncStatefulSet)
+	if err == nil && hasStatefulSetSpecChanged(existingClusterSyncStatefulSet, newClusterSyncStatefulSet, hLog) {
+		// The statefulset spec.selector changed. Trying to apply this new statefulset will error with:
+		//     invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"
+		// The only fix is to delete the statefulset and have the apply below recreate it.
+		hLog.Info("deleting the existing clustersync statefulset because spec has changed")
+		err := r.Delete(context.TODO(), existingClusterSyncStatefulSet)
```
abhinavdahiya, line 122: what's the default forf delete in terms of waiting for deletion propagation, are we waiting for all pods and the object to be deleted before moving ahead.

what happens if we delete, the deletion timestamp is set and object is not yet deleted and we try to create a new one immediately after.

pkg/operator/hive/clustersync.go
```diff
@@ -107,3 +137,31 @@ func (r *ReconcileHiveConfig) deployClusterSync(hLog log.FieldLogger, h resource
...
+	if !found {
+		hLog.Debug("existing clustersync statefulset Annotations doesn't contain the spec hash, assuming spec changed.")
+		return true
+	}
+
+	newClusterSyncStatefulSetSpecHash := newClusterSyncStatefulSet.Annotations[hiveClusterSyncStatefulSetSpecHashAnnotation]
+	if existingClusterSyncStatefulSetSpecHash != newClusterSyncStatefulSetSpecHash {
+		hLog.Debug("existing clustersync statefulset Spec changed")
+		return true
+	}
```
abhinavdahiya, line 163: what if we stored 2 hashed,
- hash for entire spec
- hash for mutable spec parts like replicas, template, updateStrategy

and we delete+create when hash(spec) changed and hash(mutable spec) remain unchanged?

this would reduce the times where we delete and recreate instead of rolling out the changes...

pkg/operator/hive/clustersync.go
dgoodwin: Would a simple solution to this possible problem be to setup a watch for our StatefulSet that triggers a reconcile, and then in this logic we return immediately after issuing the Delete, and probably add a check for DeletionTimestamp already being set, in which case we also return right away. Basically just keep reconciling if we see the spec changed and we're not fully deleted, once it's gone we know to recreate it.

dgoodwin: Bearing in mind that kind of watch is a long standing goal of hive-operator we haven't tackled yet. We'd love for it to react to changes to the resources it creates so it could repair them.

twiest: @dgoodwin are you proposing those changes for this PR or a future PR when we switch to watching resources we create?

I'm fine doing the work on this, but we should likely create a card as it seems like an arch change, not a bug.

pkg/operator/hive/clustersync.go
twiest: That's an interesting idea.

Is it worth having to manually determine / maintain the list of mutable parts of the statefulset spec?

Since the deletes aren't ultimately destructive (we setup a new sts), and since the spec values don't change all that often, we're basically optimizing for not restarting the hive-clustersync pod as much (which we don't do that often now).

I'm down to do the work, not sure if it's worth it, though. Typically simpler logic is easier to maintain.

@dgoodwin What is your preference?

pkg/operator/hive/clustersync.go
dgoodwin: I'm proposing just the addition of a watch for this StatefulSet, and the return early logic proposed above, I did not consider this an architectural change, rather an improvement to this PR to prevent a potential bug. We already watch more things than I realized in hive_controller.go, but StatefulSet is not one of them.

twiest: Ah, ok. From your description, I thought it was a bigger change.

twiest: @dgoodwin I believe I've implemented your proposed fix for this one. PTAL.

pkg/operator/hive/hive_controller.go
```diff
@@ -155,26 +155,31 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-		OwnerType:    &hivev1.HiveConfig{},
+		OwnerType: &hivev1.HiveConfig{},
 	})
 	if err != nil {
 		return err
 	}
 
 	// Monitor changes to Services:
 	err = c.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
```
dgoodwin, line 176: These IsController fields were incorrect I take it?

dgoodwin (commented):
I think this looks good to me.

 If you needed a way to manually test and didn't already think of it, you could probably add a finalizer to the statefulset to block the delete and make sure the controller was short circuiting and watching for it to actually disappear.

@abhinavdahiya over to you but IMO this looks good to me

pkg/operator/hive/hive_controller.go
twiest: Yeah. The owner references are setup with `IsController: false`. I found that when I copied and pasted for my watch and the watch didn't fire.

I tested this change for each of the watch types except daemonsets as I couldn't find any daemonsets that had the hiveconfig as an owner.

twiest:
> If you needed a way to manually test and didn't already think of it, you could probably add a finalizer to the statefulset to block the delete and make sure the controller was short circuiting and watching for it to actually disappear.


That's exactly how I tested it. Everything worked as expected.

PR #1262, opened 2021-01-27
metrics: Add metricClusterDeploymentSyncsetPaused

mbarnes:
The OpenShift Dedicated SRE team would like to be alerted when a production ClusterDeployment has a `hive.openshift.io/syncset-pause="true"` annotation for some threshold duration.

This adds a new Prometheus metric to allow for such an alert.  It's essentially a boolean for each ClusterDeployment.

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -98,6 +99,17 @@ var (
 		},
 		[]string{"controller", "outcome"},
 	)
+	// metricClusterDeploymentSyncsetPaused tracks ClusterDeployments which
+	// are NOT being synchronized by Hive.  This is achieved by the presence
+	// of a hive.openshift.io/syncset-pause="true" annotation.  The value is
+	// a boolean with "1" indicating the annotation is present.
+	metricClusterDeploymentSyncsetPaused = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "hive_cluster_deployment_syncset_paused",
+			Help: "Whether Hive has paused syncing to the cluster",
+		},
+		[]string{"cluster_deployment", "namespace", "cluster_type"},
+	)
 )
 
 // ReconcileOutcome is used in controller "reconcile complete" log entries, and the metricControllerReconcileTime
@@ -130,6 +142,7 @@ func init() {
 	metrics.Registry.MustRegister(metricControllerReconcileTime)
 
 	metrics.Registry.MustRegister(MetricClusterDeploymentDeprovisioningUnderwaySeconds)
+	metrics.Registry.MustRegister(metricClusterDeploymentSyncsetPaused)
 }
 
 // Add creates a new metrics Calculator and adds it to the Manager.
@@ -182,6 +195,7 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 				return
 			}
 			for _, cd := range clusterDeployments.Items {
+				clusterType := GetClusterDeploymentType(&cd)
 				accumulator.processCluster(&cd)
 
 				if cd.DeletionTimestamp != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -192,9 +206,18 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
 				}
+
+				var value float64
+				if paused, err := strconv.ParseBool(cd.Annotations[constants.SyncsetPauseAnnotation]); err == nil && paused {
+					value = 1.0
+				}
+				metricClusterDeploymentSyncsetPaused.WithLabelValues(
+					cd.Name,
+					cd.Namespace,
+					clusterType).Set(value)
```
dgoodwin, line 220: This is a great idea and thanks for sending us the PR directly!

We want to limit the data we report as much as possible, as implemented it's reporting 0's for every cluster in the system, I think you can just skip the 0's.

Then you would want to make sure to delete any metric that shouldn't be set (for clusters that were paused but no longer are) so it stops getting reported (and eventually cycles out of the prometheus data as it expires a week or two later), which you can do idempotently with this approach:
https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1927

mbarnes: That makes sense, and thanks for the quick review.  I'll rework as you suggest.

mbarnes:
@dgoodwin Reworked as you suggested, though CI tests seem to be wedged at the moment.

This version is a little more intrusive as I had to split a util function into two and tend to its unit tests.

pkg/controller/metrics/metrics.go
```diff
@@ -146,6 +159,19 @@ func Add(mgr manager.Manager) error {
...
+// ClearClusterDeploymentSyncsetPaused deletes the MetricClusterDeploymentSyncsetPaused
+// labeled value for the given ClusterDeployment.
+func ClearClusterDeploymentSyncsetPaused(cd *hivev1.ClusterDeployment, logger log.FieldLogger) {
+	cleared := MetricClusterDeploymentSyncsetPaused.Delete(map[string]string{
+		"cluster_deployment": cd.Name,
+		"namespace":          cd.Namespace,
+		"cluster_type":       GetClusterDeploymentType(cd),
+	})
+	if cleared {
+		logger.Debug("cleared metric: %v", MetricClusterDeploymentSyncsetPaused)
```
dgoodwin, line 171: Probably worth Info level as this should be pretty rare.

pkg/controller/metrics/metrics.go
```diff
@@ -192,9 +219,16 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
...
+						clusterType).Set(
 						time.Since(cd.CreationTimestamp.Time).Seconds())
 				}
+
+				if paused, err := strconv.ParseBool(cd.Annotations[constants.SyncsetPauseAnnotation]); err == nil && paused {
+					MetricClusterDeploymentSyncsetPaused.WithLabelValues(
+						cd.Name,
+						cd.Namespace,
+						clusterType).Set(1.0)
+				}
```
dgoodwin, line 231: I might be mistaken, but I think you can just do the Delete in an else right here, does that work? I fear my link misled you to plug it into a controller, whereas in this case I think it's safe for us to clear it right here. (was intended to demonstrate how to Delete for specific labels)

pkg/controller/utils/clusterdeployment.go
```diff
@@ -23,17 +23,21 @@ func IsFakeCluster(cd *hivev1.ClusterDeployment) bool {
...
+func IsClusterPaused(cd *hivev1.ClusterDeployment, logger log.FieldLogger) bool {
 	if paused, err := strconv.ParseBool(cd.Annotations[constants.SyncsetPauseAnnotation]); err == nil && paused {
 		logger.WithField("annotation", constants.SyncsetPauseAnnotation).Warn("syncing to cluster is disabled by annotation")
 		return true
 	}
+	return false
+}
+
+// IsClusterRelocating checks if the cluster is relocating
+func IsClusterRelocating(cd *hivev1.ClusterDeployment, logger log.FieldLogger) bool {
```
dgoodwin, line 36: If the above is correct, the rest of this refactor can be reverted back to original form.

dgoodwin:
:100: 

Thanks!

/lgtm

PR #1263, opened 2021-01-28
ensure controller runtime uses logrus for logging

abhinavdahiya:
controller runtime currently defaults to NullLogger [1] which doesn't log anything.
This is bad because any error returned from Reconcile function is therefore not logged anywhere
which is same as just eating the error. To make sure the controller runtime errors are also logged
we would need an adapter for logrus as this project uses that for logging.

Therefore, this implements a wrapper around lorgus that satisfies the logr.Logger interface and
confure the hive-controller and hive-operator managers with it.

The wrapper uses logrus.DebugLevel for all info messages and logrus.ErrorLevel for all error messages
from controller-runtime regardless of level (V) selected by caller in controller-runtime. This makes
sure all error messages from controller-runtime are always visible and the rest are only visible when in
debug mode.

[1]: https://github.com/kubernetes-sigs/controller-runtime/blob/9e78e653228851684b6b9cdabe5aae8559fe3722/pkg/log/null.go#L28

/assign @dgoodwin 
/cc @joelddiaz

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -56,6 +56,7 @@ import (
 	"github.com/openshift/hive/pkg/controller/unreachable"
 	"github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/controller/velerobackup"
+	utillogrus "github.com/openshift/hive/pkg/util/logrus"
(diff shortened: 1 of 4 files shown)


dgoodwin:
This is terrific, thanks!

/lgtm

PR #1270, opened 2021-02-01
Change fake-cluster-id to a real uuid

twiest:
- [x] Change fake-cluster-id to a real uuid
- [x] Test the change to ensure it is working correctly

/assign @joelddiaz 
/cc @suhanime

### DIFF
--- a/pkg/installmanager/fake.go
+++ b/pkg/installmanager/fake.go
@@ -4,12 +4,13 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/google/uuid"
 	installertypes "github.com/openshift/installer/pkg/types"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 )
 
-const fakeMetadataFormatStr = `{"aws":{"identifier":[{"kubernetes.io/cluster/fake-infraid":"owned"},{"openshiftClusterID":"fake-cluster-id"}],"region":"us-east-1"},"clusterID":"fake-cluster-id","clusterName":"%s","infraID":"fake-infra-id"}`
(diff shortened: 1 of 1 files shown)


twiest:
Fixes: https://issues.redhat.com/browse/HIVE-1354

joelddiaz (commented):
/lgtm

pkg/installmanager/fake.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"fmt"
 
+	"github.com/google/uuid"
 	installertypes "github.com/openshift/installer/pkg/types"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1"
 )
 
-const fakeMetadataFormatStr = `{"aws":{"identifier":[{"kubernetes.io/cluster/fake-infraid":"owned"},{"openshiftClusterID":"fake-cluster-id"}],"region":"us-east-1"},"clusterID":"fake-cluster-id","clusterName":"%s","infraID":"fake-infra-id"}`
+const fakeMetadataFormatStr = `{"aws":{"identifier":[{"kubernetes.io/cluster/fake-infraid":"owned"},{"openshiftClusterID":"%s"}],"region":"us-east-1"},"clusterID":"%s","clusterName":"%s","infraID":"fake-infra-id"}`
```
abhinavdahiya, line 13: I think the goal of using `fake-cluster-id` was to make it sure clear that this is a fake cluster, replacing this with UUID only now makes these clusters look very similar to real ones which can cause problems.. any reason why we didn't go for the original recommendation from slack thread of `fake-cluster-UUID` @twiest

twiest: For posterity, this pattern was discussed on slack, but didn't make it to the card. Working on a separate PR to change to using this pattern.

twiest: Fixed in PR #1273

PR #1271, opened 2021-02-02
[WIP] Move ClusterSync Status out of Etcd enhancement

suhanime:
/assign @dgoodwin - Need some pointers

### DIFF
--- a/docs/enhancements/clustersync_data_breakout.md
+++ b/docs/enhancements/clustersync_data_breakout.md
@@ -1,0 +1,50 @@
+# Move SyncSet Status out of Etcd
+
+
+## Summary
+The purpose of this enhancement is to break out SyncSet status out of Etcd and move to external storage
+
+## Motivation
+Once a cluster is installed, ClusterSync account for a bulk of Etcd's traffic and storage since it stores per cluster
+status for all matching SyncSets. It can easily be reconstructed if lost, and it is not a resource we reconcile on.
+In order to avoid such regular large writes to Etcd, Hive SyncSet status is the perfect candidate to move out of Etcd.
+This can help Hive clusters scale to the anticipated 1000 per Hive target.
+
+### Goals
+Deploy an alternative storage backend via the Hive Operator and eliminate the ClusterSync Etcd storage.
+
+### Non-Goals
+None
+
+## Open Questions
+1. How to get the details of failing SyncSets for a cluster? Some options to consider:
+    1. Add a small REST API and possibly CLI tool to scrape. However, how to secure the REST API?
+    1. Add a CRD to request SyncSet status, only used when someone is explicitly debugging a problem, and handle cleaning
+    up shortly thereafter.
+
+## Proposal
+- Deploy an official Red Hat postgresql image automatically via the Hive operator. On a later date, we may let the users
+  point to an external database they control.
+- Hive operator managing schema migrations automatically.
+- Hive ClusterSync pods continue reconciling on ClusterDeployment as they do today, but now store and load ClusterSync 
+  state in/from postgres.
+
+## Implementation Details / Notes / Constraints
+1. postgresql is relational unlike Etcd, so we have to set up schema. We can use https://github.com/pressly/goose
+1. We can store the status as XML directly? - will we replace directly if it changes?
+
+## Risks and Mitigations
+- Role based access - postgresql allows fine-grained control
+- Delete protection? Or reconstruction
+
+### Test Plan
+
+
+## Drawbacks
+- End user must maintain external storage, including upgrading it when it is out of space.
+- Default is deploying the external storage - what happens when there's an issue with that storage?
+
+## Alternatives
+- Use aggregated APIServer, backed by either a separate Etcd, or a different layer of database (and maintain that 
+  database), although we have been cautioned this has not been done successfully yet, but some projects like 
+  https://github.com/kubernetes-sigs/apiserver-runtime are attempting it.


docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+This can help Hive clusters scale to the anticipated 1000 per Hive target.
+
+### Goals
+Deploy an alternative storage backend via the Hive Operator and eliminate the ClusterSync Etcd storage.
+
+### Non-Goals
+None
+
+## Open Questions
+1. How to get the details of failing SyncSets for a cluster? Some options to consider:
```
dgoodwin, line 20: These are the key elements in this enhancement as this is what our users will be asking about, we need to get these fleshed out and into the proposal section.

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+    1. Add a small REST API and possibly CLI tool to scrape. However, how to secure the REST API?
```
dgoodwin, line 21: I believe SRE usually gets cluster kubeconfig when there's a problem. How would this theoretical REST API be securely exposed to SRE and not the outside world? Routes would probably be a bad idea. We could do just a Service and require SRE to use kubectl port-forward? Is this secure enough?

How would SRE obtain a CLI tool when they need it? There is some precedent for them checking out hive and running a script. Maybe building a go binary?

Would they be better off with curl/httpie and forget the CLI?

What could the REST API endpoints look like?

What input parameters are needed and how would SRE obtain them?

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+    1. Add a CRD to request SyncSet status, only used when someone is explicitly debugging a problem, and handle cleaning
```
dgoodwin, line 22: What would this CRD look like? Is it Cluster specific or [Selector]SyncSet specific? (probably has to be cluster specific as it is today) When would it be cleaned up? 

What does the flow look like for SRE to know a SyncSet is having problems, and identifying the parameters to make the request?

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+- Hive operator managing schema migrations automatically.
+- Hive ClusterSync pods continue reconciling on ClusterDeployment as they do today, but now store and load ClusterSync 
+  state in/from postgres.
+
+## Implementation Details / Notes / Constraints
+1. postgresql is relational unlike Etcd, so we have to set up schema. We can use https://github.com/pressly/goose
+1. We can store the status as XML directly? - will we replace directly if it changes?
+
+## Risks and Mitigations
+- Role based access - postgresql allows fine-grained control
```
dgoodwin, line 37: I would hope to avoid relying on postgresql rbac for this, the db should hopefully be abstracted from the user (SRE), and exposed via one of the other mechanisms above which would control rbac for us. (hopefully)

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+- Delete protection? Or reconstruction
+
+### Test Plan
+
+
+## Drawbacks
+- End user must maintain external storage, including upgrading it when it is out of space.
+- Default is deploying the external storage - what happens when there's an issue with that storage?
```
dgoodwin, line 45: It may be more awkward for SRE to find the details of why a syncset is failing.

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Risks and Mitigations
+- Role based access - postgresql allows fine-grained control
+- Delete protection? Or reconstruction
+
+### Test Plan
+
+
+## Drawbacks
+- End user must maintain external storage, including upgrading it when it is out of space.
```
dgoodwin, line 44: Cluster must have persistent storage configured and available. This might be an issue for ACM who I believe does not require this today.

I don't have any major concerns about how much storage is available or running out, this should not be a very large amount of data.

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+- Default is deploying the external storage - what happens when there's an issue with that storage?
+
+## Alternatives
+- Use aggregated APIServer, backed by either a separate Etcd, or a different layer of database (and maintain that 
+  database), although we have been cautioned this has not been done successfully yet, but some projects like 
+  https://github.com/kubernetes-sigs/apiserver-runtime are attempting it.
```
dgoodwin, line 50: Lots of projects do aggregated with separate etcd. Aggregated + non-etcd database has not been done in any serious capacity, as far as we're aware. 

Aggregated APIs are cautioned against as they are a large maintenance burden. However Hive ran one recently for the v1alpha1 to v1 migration.

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+- Hive ClusterSync pods continue reconciling on ClusterDeployment as they do today, but now store and load ClusterSync 
+  state in/from postgres.
+
+## Implementation Details / Notes / Constraints
+1. postgresql is relational unlike Etcd, so we have to set up schema. We can use https://github.com/pressly/goose
+1. We can store the status as XML directly? - will we replace directly if it changes?
+
+## Risks and Mitigations
+- Role based access - postgresql allows fine-grained control
+- Delete protection? Or reconstruction
```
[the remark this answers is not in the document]
dgoodwin: This might also mean we don't really need persistent storage and could use ephemeral? But postgresql pod restarts would be expensive as all syncsets would need to reapply.

docs/enhancements/clustersync_data_breakout.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Proposal
+- Deploy an official Red Hat postgresql image automatically via the Hive operator. On a later date, we may let the users
+  point to an external database they control.
+- Hive operator managing schema migrations automatically.
+- Hive ClusterSync pods continue reconciling on ClusterDeployment as they do today, but now store and load ClusterSync 
+  state in/from postgres.
+
+## Implementation Details / Notes / Constraints
+1. postgresql is relational unlike Etcd, so we have to set up schema. We can use https://github.com/pressly/goose
```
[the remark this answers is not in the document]
dgoodwin: Including sample schema wouldn't be a bad idea just to demonstrate this point.

docs/enhancements/clustersync_data_breakout.md
dgoodwin: HIve clusters are not on public internet, is routing actually ok? SRE would then sshuttle and curl something like https://hive-aux-api.hivep01s01.openshift.com/clustersync/ ?

PR #1272, opened 2021-02-02
vendor: bump installer to latest

abhinavdahiya:
bumps the installer vendor to latest, also updates the code to work with new vendor

xref: https://issues.redhat.com/browse/HIVE-1354

/assign @dgoodwin 
/cc @suhanime


go.mod
```diff
@@ -3,92 +3,102 @@ module github.com/openshift/hive
...
+	github.com/Azure/go-autorest/autorest/validation => github.com/Azure/go-autorest/autorest/validation v0.2.1-0.20191028180845-3492b2aff503
+	github.com/apparentlymart/go-cidr => github.com/apparentlymart/go-cidr v1.0.1
+	github.com/go-openapi/spec => github.com/go-openapi/spec v0.19.0
+	github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1
+	github.com/hashicorp/go-plugin => github.com/hashicorp/go-plugin v1.2.2
+	github.com/ulikunitz/xz => github.com/ulikunitz/xz v0.5.7
+	google.golang.org/api => google.golang.org/api v0.25.0
+	google.golang.org/genproto => google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
+	google.golang.org/grpc => google.golang.org/grpc v1.29.1
+)
```
suhanime, line 101: This is way better - I wonder if this warranties some changes [here](https://github.com/openshift/hive/blob/master/docs/developing.md#vendoring-the-openshift-installer)

suhanime:
I wonder if master should use a forked version of installer as well

go.mod
abhinavdahiya: this is one off change we need to replace because of https://github.com/openshift/installer/pull/4350 , i think this should be removed when installer bumps to set of dependencies that don't need this specifically.

the doc provides general guidelines of how to add or update the dependencies in most/usual case, like the cluster-api vendor bumps for aws etc included here too in the original replace block

abhinavdahiya:
> I wonder if master should use a forked version of installer as well

previously in the installer repo we used always tried to use upstream directly as much as possible and use the forks when we need some changeset which not in any supported tips(branches). So personally i would not recommend using the fork unless required.

PR #1273, opened 2021-02-02
Add fake-cluster- prefix to fake cluster uid

twiest:
- [x] Add fake-cluster- prefix to fake cluster uid
- [x] Test that it is generating the uid properly

/assign @abhinavdahiya 
/cc @joelddiaz


joelddiaz (commented):
:+1:

PR #1275, opened 2021-02-03
Add hive_cluster_deployment_provision_underway_install_restarts metric

abhinavdahiya:
###  provision_underway_collector.go: allow setting min duration for provision time, add metric for install restarts

hive_cluster_deployment_provision_underway_seconds metric currently reports the elapsed time for all clusterdeployments
that provisioning. To allow for reduce the metrics reported and still ensure the metric can be used to alert installs taking
too long, the collector now allows setting minDuration. The minDuration ensures the collector only cares about provisioning times
greater than minDuration.

Another metric that's important for tracking failing installs is the number of restarts for installs. This allows catching failures
that fail really quickly, like permissions, compared to underway_seconds where the installs might have to restart for a very large value
to be alerted.

so a new metric hive_cluster_deployment_provision_underway_install_restarts is added that tracks the number of restarts for a cluster that
is still provisioning.

### controller/metrics/metrics.go: report metrics for only at least 1 hour elapsed time and 1 install restarts

track clusters in hive_cluster_deployment_provision_underway_seconds which have been provisioning for atleast
1 hour.
track clusters in hive_cluster_deployment_provision_underway_install_restarts which have atleast 1 restarts.

This makes sure the amount of metrics collected are smaller but still provide most value in alerting failed installs.

xref: https://issues.redhat.com/browse/HIVE-1341

/assign @dgoodwin
/cc @suhanime

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -151,7 +151,8 @@ func Add(mgr manager.Manager) error {
 		Client:   mgr.GetClient(),
 		Interval: 2 * time.Minute,
 	}
-	metrics.Registry.MustRegister(newProvisioningUnderwayCollector(mgr.GetClient()))
+	metrics.Registry.MustRegister(newProvisioningUnderwaySecondsCollector(mgr.GetClient(), 1*time.Hour))
+	metrics.Registry.MustRegister(newProvisioningUnderwayInstallRestartsCollector(mgr.GetClient(), 1))
 	err := mgr.Add(mc)
 	if err != nil {
 		return err
--- a/pkg/controller/metrics/provision_underway_collector.go
+++ b/pkg/controller/metrics/provision_underway_collector.go
@@ -30,6 +30,11 @@ var (
 type provisioningUnderwayCollector struct {
 	client client.Client
 
+	// minDuration, when non-zero, is the minimum duration afer which clusters provisioning
+	// will start becomming part of this metric. When set to zero, all clusters provisioning
+	// will be included in the metric.
+	minDuration time.Duration
+
 	// metricClusterDeploymentProvisionUnderwaySeconds is a prometheus metric for the number of seconds
 	// between when a still provisioning cluster was created and now.
 	metricClusterDeploymentProvisionUnderwaySeconds *prometheus.Desc
@@ -56,17 +61,7 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 		}
 
 		// Add install failure details for stuck provision
-		condition, reason := "Unknown", "Unknown"
-		for _, delayCondition := range provisioningDelayCondition {
-			if cdCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions,
-				delayCondition); cdCondition != nil {
-				if cdCondition.Status == corev1.ConditionTrue && cdCondition.Reason != "" {
-					condition = string(delayCondition)
-					reason = cdCondition.Reason
-				}
-				break
-			}
-		}
+		condition, reason := getKnownConditions(cd.Status.Conditions)
 
 		platform := cd.Labels[hivev1.HiveClusterPlatformLabel]
 		imageSet := "none"
@@ -74,11 +69,16 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			imageSet = cd.Spec.Provisioning.ImageSetRef.Name
 		}
 
+		elapsedDuration := time.Since(cd.CreationTimestamp.Time)
+		if cc.minDuration.Seconds() > 0 && elapsedDuration < cc.minDuration {
+			continue // skip reporting the metric for clusterdeployment until the elapsed time is at least minDuration
+		}
+
 		// For installing clusters we report the seconds since the cluster was created.
 		ch <- prometheus.MustNewConstMetric(
 			cc.metricClusterDeploymentProvisionUnderwaySeconds,
 			prometheus.GaugeValue,
-			time.Since(cd.CreationTimestamp.Time).Seconds(),
+			elapsedDuration.Seconds(),
 			cd.Name,
 			cd.Namespace,
 			GetClusterDeploymentType(&cd),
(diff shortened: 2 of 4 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -151,7 +151,8 @@ func Add(mgr manager.Manager) error {
 		Client:   mgr.GetClient(),
 		Interval: 2 * time.Minute,
 	}
-	metrics.Registry.MustRegister(newProvisioningUnderwayCollector(mgr.GetClient()))
+	metrics.Registry.MustRegister(newProvisioningUnderwayCollectorWithMinDuration(mgr.GetClient(), 1*time.Hour))
+	metrics.Registry.MustRegister(newProvisioningUnderwayInstallRestartsCollectorWithMinRestarts(mgr.GetClient(), 2))
```
dgoodwin, line 155: My vote would be to start reporting after 1 restart, and this would more closely match when the min restarts start showing up. They're limiting at 3 in OSD so 1 feels a little better for the min.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -96,14 +105,132 @@ func (cc provisioningUnderwayCollector) Describe(ch chan<- *prometheus.Desc) {
...
+var (
+	metricClusterDeploymentProvisionUnderwaySecondsDesc = prometheus.NewDesc(
+		"hive_cluster_deployment_provision_underway_seconds",
+		"Length of time a cluster has been provisioning.",
+		[]string{"cluster_deployment", "namespace", "cluster_type", "condition", "reason", "platform", "image_set"},
+		nil,
+	)
+)
+
 func newProvisioningUnderwayCollector(client client.Client) prometheus.Collector {
```
dgoodwin, line 117: Do we need to keep this old function? Seems like we could just transition fully to the new.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -30,6 +30,10 @@ var (
 type provisioningUnderwayCollector struct {
 	client client.Client
 
+	// minDuration is the minimum duration afer which clusters provisioning
+	// will start reporting the metric. Only used when non-zero.
```
dgoodwin, line 34: "Only used when non-zero", could you clarify that if left zero the metric is always reported? When first read I thought it meant we would never report it if it was zero.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -96,14 +105,132 @@ func (cc provisioningUnderwayCollector) Describe(ch chan<- *prometheus.Desc) {
...
+		minDuration: minimum,
+	}
+}
+
+// provisioning underway install restarts metrics collected through a custom prometheus collector
+type provisioningUnderwayInstallRestartsCollector struct {
+	client client.Client
+
+	// minRestarts is the minimum restarts after which clusters provisioning
+	// will start reporting the metric. Only used when non-zero.
```
dgoodwin, line 137: Same here, if zero will always report.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -96,14 +105,132 @@ func (cc provisioningUnderwayCollector) Describe(ch chan<- *prometheus.Desc) {
...
+		for _, delayCondition := range provisioningDelayCondition {
+			if cdCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions,
+				delayCondition); cdCondition != nil {
+				if cdCondition.Status == corev1.ConditionTrue && cdCondition.Reason != "" {
+					condition = string(delayCondition)
+					reason = cdCondition.Reason
+				}
+				break
+			}
+		}
```
dgoodwin, line 176: Probably worth pulling this out to a little function to share between the collectors. Might be subtle fixes in there in future we'd miss in both.

pkg/controller/metrics/provision_underway_collector_test.go
```diff
@@ -0,0 +1,519 @@
...
+			var got []string
+			for sample := range ch {
+				var d dto.Metric
+				require.NoError(t, sample.Write(&d))
+				got = append(got, metricPrettyWithValue(d))
+			}
+			assert.Equal(t, test.expected, got)
+		})
+	}
+}
```
dgoodwin, line 503: Bonus! I've read in some places not to unit test your metrics but I think in this case, with a custom collector, it makes sense.

dgoodwin (changes requested):

pkg/controller/metrics/metrics.go
abhinavdahiya: sure, fixed to 1

pkg/controller/metrics/provision_underway_collector.go
abhinavdahiya: updated to use the previous function to setup the collect instead of a new one, instead of dropping in favor of new one

pkg/controller/metrics/provision_underway_collector.go
abhinavdahiya: updated

pkg/controller/metrics/provision_underway_collector.go
abhinavdahiya: updated

pkg/controller/metrics/provision_underway_collector.go
abhinavdahiya: updated to keep the condition extraction in a separate function.

pkg/controller/metrics/provision_underway_collector_test.go
abhinavdahiya: > I've read in some places not to unit test your metrics

it's code that can and will have assumptions, so unit tests make sense to make sure the assumptions don't change underneath you. So i don't get why we would not unit test collectors... unit tests also just help me understang how things work, _these are better than documentation in my case_ :P

dgoodwin: Yeah maybe their stance is more related to simpler instrumentation just reporting up a value. IIRC the reasoning was to not make it overly difficult to add metrics, so people would be more likely to do it.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -96,14 +96,124 @@ func (cc provisioningUnderwayCollector) Describe(ch chan<- *prometheus.Desc) {
...
+var (
+	metricClusterDeploymentProvisionUnderwaySecondsDesc = prometheus.NewDesc(
+		"hive_cluster_deployment_provision_underway_seconds",
+		"Length of time a cluster has been provisioning.",
+		[]string{"cluster_deployment", "namespace", "cluster_type", "condition", "reason", "platform", "image_set"},
+		nil,
+	)
+)
+
+func newProvisioningUnderwayCollector(client client.Client, minimum time.Duration) prometheus.Collector {
```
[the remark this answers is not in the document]
dgoodwin: :+1:

Otherwise LGTM unless @suhanime has any feedback.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -30,6 +30,11 @@ var (
 type provisioningUnderwayCollector struct {
 	client client.Client
 
+	// minDuration, when non-zero, is the minimum duration afer which clusters provisioning
+	// will start becomming part of thismetric. When set to zero, all clusters provisioning
```
suhanime, line 34: nit - but we tend to ignore doc errors later on
`will start becoming part of this metric`*

pkg/controller/metrics/provision_underway_collector.go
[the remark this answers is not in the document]
suhanime: +1 to this suggestion

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -96,14 +96,124 @@ func (cc provisioningUnderwayCollector) Describe(ch chan<- *prometheus.Desc) {
...
+	// will start becoming part of the metric. When set to zero, all clusters provisioning
+	// will be included in the metric.
+	minRestarts int
+
+	// metricClusterDeploymentProvisionUnderwayInstallRestarts is a prometheus metric for the number of install
+	// restarts for a still provisioning cluster.
+	metricClusterDeploymentProvisionUnderwayInstallRestarts *prometheus.Desc
+}
+
+// collects the metrics for provisioningUnderwayCollector
```
suhanime, line 130: This doc is now confusing. We can even scrape it since the func definition is enough to understand

pkg/controller/metrics/provision_underway_collector_test.go
suhanime: Unit tests are always a plus in my book

suhanime (commented):
The changes look good - just some minor nits

abhinavdahiya:
@dgoodwin @suhanime fixed the last nits , ready for another review..

PR #1283, opened 2021-02-09
 remotemachineset/openstackactuator.go: re enable using clientOpts

abhinavdahiya:
dc89529 removed passing clientOpts to generate the machinesets as it was
dropped as supported option in the installer libraries.

since openshift/installer#4638 has re enabled the option
and reverting the change in openstackactuator to use the configure client opts
for correctly authenticating with openstack apis.

/assign @abutcher 

xref: https://issues.redhat.com/browse/HIVE-1372


abhinavdahiya:
/hold

currently has temp vendor from personal fork to test https://github.com/openshift/installer/pull/4638

abhinavdahiya:
/hold cancel

the vendor for installer uses the correct upstream

/assign @abutcher 
/cc @dgoodwin

PR #1286, opened 2021-02-12
Add some documentation around ClusterRelocate.

dgoodwin:
Built off @staebler's comments in #1011.

Tested and still working. 

Would like to get some feedback from SRE on how they're linking their operators into this process, thus this is still a TODO in the doc.

/assign @abhinavdahiya 
/cc @abutcher 

@gregsheremeta do you know any names of folks working on this SRE side who you could ping here for feedback?

### DIFF
--- a/docs/cluster-relocation.md
+++ b/docs/cluster-relocation.md
@@ -1,0 +1,79 @@
+# Cluster Relocation
+
+Hive supports the ability to migrate a `ClusterDeployment` and its related custom resources to another Hive cluster through the use of the `ClusterRelocate` custom resource definition.
+
+## Usage
+
+In your source Hive cluster, create a Secret containing a kubeconfig for the destination Hive cluster in the namespace of your choosing:
+
+```bash
+$ kubectl create secret generic hub2-migration-kubeconfig -n default --from-file=kubeconfig=./hub2.kubeconfig
+```
+
+In your source Hive cluster, create a `ClusterRelocate` custom resource:
+
+```yaml
+apiVersion: hive.openshift.io/v1
+kind: ClusterRelocate
+metadata:
+  name: migrator
+spec:
+  kubeconfigSecretRef:
+    namespace: default
+    name: hub2-migration-kubeconfig
+  clusterDeploymentSelector:
+    matchLabels:
+      migrateme: hub2
+```
+
+Apply the desired label to your `ClusterDeployment`:
+
+```bash
+$ kubectl label clusterdeployment mycluster migrateme=hub2
+```
+
+The `ClusterDeployment` should appear in the destination hive, and be deleted in the source Hive, without triggering any cleanup of cluster resources.
+
+## Caveats
+
+The relocation process will migrate most of the relevant resources in a source namespace, so if you have multiple `ClusterDeployments` in one namespace, it is possible some of their secrets will be copied to the destination cluster even if only one of the `ClusterDeployments` matched the label selector. Best practice for Hive is to use a namespace per `ClusterDeployment`.
+
+## Implementation Details
+
+[Original Pull Request](https://github.com/openshift/hive/pull/1011)
+
+Relocation was implemented via a `hive.openshift.io/relocate` annotation which is added to the `ClusterDeployment`. This annotation's value will be `{cluster_relocate_name}/{relocation_status}`, where status can be "outgoing", "complete", or "incoming". When this annotation is set, most hive controllers are disabled so as not to make changes during the relocation.
+
+Relocation proceeds with several steps:
+
+When a `ClusterRelocate` has a label selector that matches with a `ClusterDeployment`, the clusterrelocate controller will relocate all matching `ClusterDeployments`.
+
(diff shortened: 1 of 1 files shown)


gregsheremeta:
pick random names from https://issues.redhat.com/browse/SDE-806

It's probably best for us to have a sync-up meeting with a few of them. The work on their side is split across several teams.

docs/cluster-relocation.md
```diff
@@ -0,0 +1,82 @@
+# Cluster Relocation
+
+Hive supports the ability to migrate a `ClusterDeployment` and it's related custom resources to another Hive cluster through the use of the `ClusterRelocate` custom resource definition.
```
staebler, line 3:
```suggestion
Hive supports the ability to migrate a `ClusterDeployment` and its related custom resources to another Hive cluster through the use of the `ClusterRelocate` custom resource definition.
```

docs/cluster-relocation.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Usage
+
+In your source Hive cluster, create a Secret containing a kubeconfig for the destination Hive cluster:
+
+```bash
+$ kubectl create secret generic hub2-migration-kubeconfig --from-file=kubeconfig=./hub2.kubeconfig
```
staebler, line 10: For consistency between this command and the `ClusterRelocate` resource, I recommend specifying the namespace. I also recommend using a namespace other than `hive` so that it does not give the impression that the secret needs to be in the `hive` namespace.
```suggestion
$ kubectl create secret generic hub2-migration-kubeconfig -n hive --from-file=kubeconfig=./hub2.kubeconfig
```

docs/cluster-relocation.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Caveats
+
+The relocation process will migrate most of the relevant resources in a source namespace, so if you have multiple `ClusterDeployments` in one namespace, it is possible some of their secrets will be copied to the destination cluster even if only one of the `ClusterDeployments` matched the label selector. Best practice for Hive is to use a namespace per `ClusterDeployment`.
+
+## Implementation Details
+
+[Original Pull Request](https://github.com/openshift/hive/pull/1011)
+
+Relocation was implemented via a `hive.openshift.io/relocate` annotation which is added to the `ClusterDeployment`. This annotation's value can be "outgoing", "complete", or "incoming". When set, most hive controllers are disabled so as not to make changes during the relocation.
```
staebler, line 45: This is not quite accurate. The annotation value is actually "{relocate_name}/{relocate_status}", where "{relocate_name}" is the name of the `ClusterRelocate` resource and "{relocate_status}" is one of "outgoing", "complete", or "incoming". The "{relocate_name}" is important for aborting an on-going relocation when labels or selectors change.

docs/cluster-relocation.md
dgoodwin: Moved to default and added a note on "in the namespace of your choosing"

docs/cluster-relocation.md
```diff
@@ -0,0 +1,82 @@
...
+```
+hive_aborted_cluster_relocations{cluster_relocate="",reason="no_match"} 5
+```
+
+The number of failing cluster relocates can be found by looking at the `hive_cluster_deployments_conditions` metric and filtering on
+a value of RelocationFailed for the condition label.
+
+## Integrating Other Applications
+
+TODO
```
dgoodwin, line 82: @dofinn could you clarify what the plan is for integrating operators to link into this migration process?

mrbarge: @dgoodwin I think it's probably going to vary from operator to operator as to how they need to specifically handle this situation. We've recently looked at two candidates, the `pagerduty-operator` and `deadmanssnitch-operator`. For both of them, we have updated the operators to specifically check whether the relocation annotation has been applied to the `clusterdeployments` that they watch. This then governs their behaviour during the relocation process on both shards.

For example, if the annotation indicates that it is relocating on the outgoing shard, the operators will not remove their corresponding PagerDuty and DeadMansSnitch services when the clusterdeployment is being deleted on the outgoing shard. If the annotation indicates that it is relocating on the incoming shard, the operators will not attempt to reconcile any of the resources they manage until that annotation is removed and we can infer that all related resources have been relocated.

Is that the sort of information you were looking for here? I have written up a more detailed analysis for those two operators in the document below [0], if that's helpful.

[0] https://docs.google.com/document/d/13J9iiL0wtgBz56B_pGy47hc1uIz5DIC4_K6nw48dEXM/

dgoodwin: Yes thats perfect, thanks. I think it's probably best to skip this section and then merge this PR. Thanks!

dgoodwin:
Ready to review/merge.

abhinavdahiya:
Oops wrong PR

/lgtm cancel

abhinavdahiya:
/lgtm

No it was the right PR 🤦🏽

dgoodwin:
:clinking_glasses:

PR #1288, opened 2021-02-16
document scaling clustersync controller

twiest:
/assign @gregsheremeta 
/cc @dgoodwin

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -29,6 +29,7 @@
   - [Managed DNS](#managed-dns-1)
   - [Configuration Management](#configuration-management)
     - [SyncSet](#syncset)
+    - [Scaling ClusterSync](#scaling-clustersync)
     - [Identity Provider Management](#identity-provider-management)
   - [Cluster Deprovisioning](#cluster-deprovisioning)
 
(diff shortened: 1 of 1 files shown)


docs/using-hive.md
```diff
@@ -780,6 +781,22 @@ Hive offers two CRDs for applying configuration in a cluster once it is installe
 
 For more information please see the [SyncSet](syncset.md) documentation.
 
+### Scaling ClusterSync
+The clustersync controller manages SyncSets. By default there is only a single clustersync controller running. The clustersync controller is designed to scale horizontally, so increasing the number of clustersync controller replicas will scale the number of clusters applying syncsets simultaneously.
```
gregsheremeta, line 785: > scale the number of clusters applying syncsets simultaneously

you mean "pods" applying?

twiest: I actually did mean clusters as each pod is assigned to specific clusters.

But clearly this isn't explained well. Any ideas how to explain this better?

Maybe this is going down into the implementation too much? Should I make it more vague?

gregsheremeta: Ah. Suggest something like:
`The clustersync controller is designed to scale horizontally, so increasing the number of clustersync controller replicas will scale the number of clustersync pods running, thereby increasing the number of simultaneous clusters getting syncsets applied to them`

twiest:
@gregsheremeta I believe I've addressed your feedback. PTAL.

PR #1289, opened 2021-02-16
add controllersConfig section to template hiveconfig

gregsheremeta:
add controllersConfig section to template hiveconfig so the OSD pipeline can clustersync replicas to 3

/assign @twiest 
/cc @dgoodwin


gregsheremeta:
this doesn't need to wait for tests. Forcing.

PR #1292, opened 2021-02-18
Move hive apis to a go submodule

abhinavdahiya:
- [x] move apis to go submodule
- [x] make sure build works
- [x] make sure tests work
- [x] update makefile targets to make sure testing includes submodules
- [x] provide some documentation on how the developer workflow will be affected and gotchas

#### why move apis to /apis vs /pkg/apis

based on the [doc][1] converting a pre-existing packge in a published module to a separate  submodule requires these steps,
1. add dependency in the submodule to root module that doesn't have the package in root module
2. publish the modules
This would defeat the purpose of moving to submodule as we want to reduce the dependencies in apis module for assisted-installer folks.
Therefore, it is better to move apis from pkg/apis to apis and creating a new module.

#### why is there a vendor directory in apis/vendor

OpenShift CI currently forces all go commands to use `-mod=vendor`. To support codegen using the apis submodule we need the submodule to have it's own vendor directory so that commands like `go list` do not try and fetch the dependencies.

#### can the vendor in apis submodule change what is used at build time

NO,  based on vendoring with go modules as documented in [doc][2], 

```
Unlike vendoring in GOPATH, the go command ignores vendor directories in locations other than the main module's root directory.
```

Only the requirements set in root module's go.mod and vendor/ directory control the versions of all the lib used to build the artifacts from the root module.

[1]: https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository
[2]: https://golang.org/ref/mod#vendoring


abhinavdahiya:
> https://golang.org/ref/mod#vendoring
```
Unlike vendoring in GOPATH, the go command ignores vendor directories in locations other than the main module's root directory.
```

the vendor directory in ./apis/ seems to be used only when running unit tests and govet in that sub module, the actual build of hive binaries will use the vendor at the root.

abhinavdahiya:
Another thing to be careful is https://github.com/golang/tools/blob/master/gopls/doc/workspace.md
gopls doesn't support yet multi modules, there is provided workaround where you can open ./apis/ folder separately in VSCode workspace for gopls to work correctly.

dgoodwin:
Where do we stand with your comment about callers still having to fight with the inherited replaces flowing all the way through installer/hive etc? Do they still need those of can this new sub-module be vendored simply and cleanly?

dgoodwin:
Spoke with Abhinav yesterday, this PR should take care of the problem with carrying over replaces from installer and hive etc.

@carbonin @filanov @avishayt can you take a look at this branch and maybe test out the vendoring? We're not 100% sure this is the direction we should go vs breaking out an entirely separate repo, but it's looking more promising than it was.

abhinavdahiya:
/assign @dgoodwin 

Updated the docs for developing to cover any changes to dev workflow with submodules. PTAL.

docs/developing.md
```diff
@@ -309,6 +362,24 @@ To recreate *_vendor_* directory, you can run the following command:
...
+These submodules define their own dependencies, most commonly from the Kubernetes ecosystem. All build artifacts are
+generated from the `root` module and therefore the version of all dependencies used is defined by the `go.mod` in the
+root module, but making sure the dependecies of the all the submodules match the root modules help reduce divergence
+and dependencies pain.
+
+So whenever the root module updates the Kubernetes ecosystem dependencies,
+
+1. Update the submodules to use the desired version of the Kubernetes ecosystem.
+
+2. Update the root mosdule dependencies to the desired version of the Kubernetes ecosystem.
```
dgoodwin, line 381: module

docs/developing.md
```diff
@@ -280,6 +284,55 @@ NOTE: The cluster name and domain used to create the certificate must match the
...
+A separate `github.com/openshift/hive/apis` submodule allows other repositories to easily import the Hive APIs without
+having to deal with the complex vendoring introduced by dependencies of Hive controllers like the OpenShift Installer.
+
+Since Hive module `github.com/openshift/hive` or `root` module uses vendor directory for all it's dependencies, **ANY**
+change in `github.com/openshift/hive/apis` module requires updating the dependency in the `root` module for change to
+take effect.
+
+Therefore, these steps must be followed when updating the Hive APIs,
+
+1. Make necessary modifications to `github.com/openshift/hive/apis`.
```
dgoodwin, line 311: I tested the previous workflow with a make update, and was pleased to see the CRD yaml gets updated as before. would it be feasible to hook up an update of this specific dependency into make update such that it's done automatically if we follow this workflow normally?

Really nice to see that make verify also catches that I did not revendor.

filanov:
> Spoke with Abhinav yesterday, this PR should take care of the problem with carrying over replaces from installer and hive etc.
> 
> @carbonin @filanov @avishayt can you take a look at this branch and maybe test out the vendoring? We're not 100% sure this is the direction we should go vs breaking out an entirely separate repo, but it's looking more promising than it was.

I managed to build with https://github.com/filanov/assisted-service/tree/hive-sub-mod 
```
github.com/openshift/hive/apis => github.com/abhinavdahiya/hive/apis v0.0.0-20210224202622-c595fe32f24e
```
Had to change some paths but compilation passed.

docs/developing.md
abhinavdahiya: > would it be feasible to hook up an update of this specific dependency into make update such that it's done automatically if we follow this workflow normally?

@dgoodwin so when someone runs `make update`, we want the apis submodule updated in vendor ? Or some lite version of the vendor update?

abhinavdahiya: > I tested the previous workflow with a make update, and was pleased to see the CRD yaml gets updated as before. 

@dgoodwin i'm hoping this meant when you updated something in apis/ and ran `make update` the crds were updated like it did before the submodules?

dgoodwin: > > I tested the previous workflow with a make update, and was pleased to see the CRD yaml gets updated as before.
> 
> @dgoodwin i'm hoping this meant when you updated something in apis/ and ran `make update` the crds were updated like it did before the submodules?

Yes exactly, it works just like before which is awesome.

Regarding make update, I think just updating the apis submodule in main vendor dir would be ideal ,if it's possible, less likely to pull in unexpected things.

abhinavdahiya: > > > I tested the previous workflow with a make update, and was pleased to see the CRD yaml gets updated as before.
> > 
> > 
> > @dgoodwin i'm hoping this meant when you updated something in apis/ and ran `make update` the crds were updated like it did before the submodules?
> 
> Yes exactly, it works just like before which is awesome.
> 
> Regarding make update, I think just updating the apis submodule in main vendor dir would be ideal ,if it's possible, less likely to pull in unexpected things.

@dgoodwin overstepping the function of `go mod vendor` to make sure the vendor dir has the right contents seems like a bad idea to me. 

i could try adding a dependency on make update to check if there is diff in go submodule and run make vendor to update..

dgoodwin:
Tests are clear so lets merge once master opens and iterate if needed.

/lgtm

dgoodwin:
cc @igoihman @jewzaam @mwoodson @jharrington22 when this merges, there will now be a hive api submodule you can vendor without picking up all of hive and it's subsequent dependencies.

RiRa12621:
@dgoodwin this seems to have some undesired side effects 

```
$ go mod tidy
go: github.com/openshift/hive@v1.1.2 requires
        github.com/openshift/hive/apis@v0.0.0: reading github.com/openshift/hive/apis/apis/go.mod at revision apis/v0.0.0: unknown revision apis/v0.0.0
```

dgoodwin:
Can you show us what you've got in your go.mod?

PR #1293, opened 2021-02-23
Act on agent clusters

dgoodwin:
This change re-enables most of the clusterdeployment controller for Agent based clusters. Previously we just blanket disabled everything, but this skipped the imageset job which was required for extracting useful info for Agent installs, such as the specific semver openshift version.

Now the controller will mostly run as normal with the exception of creation of a ClusterProvision. Getting Agent clusters to link up with the ClusterProvision flow will be handled in the future.

/assign @abhinavdahiya 
/cc @abutcher @filanov @avishayt @mhrivnak

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -456,14 +456,6 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, err
 	}
 
-	// Return early and stop processing if Agent install strategy is in play. The controllers that
-	// handle this portion of the API currently live in the assisted service repo, rather than hive.
-	// This will hopefully change in the future.
-	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallStrategy != nil && cd.Spec.Provisioning.InstallStrategy.Agent != nil {
-		cdLog.Info("skipping processing of agent install strategy cluster")
-		return reconcile.Result{}, nil
-	}
-
 	if cd.DeletionTimestamp != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -650,6 +642,14 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
 	}
 
+	// Return early and stop processing if Agent install strategy is in play. The controllers that
+	// handle this portion of the API currently live in the assisted service repo, rather than hive.
+	// This will hopefully change in the future.
+	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallStrategy != nil && cd.Spec.Provisioning.InstallStrategy.Agent != nil {
+		cdLog.Info("skipping processing of agent install strategy cluster")
+		return reconcile.Result{}, nil
+	}
+
```
abhinavdahiya, line 652: Hmm.. i think we might have to update the code for installed cluster code-path too.

L539-544
```go
		// delete failed provisions which are more than 7 days old
		existingProvisions, err := r.existingProvisions(cd, cdLog)
		if err != nil {
		}
		r.deleteOldFailedProvisions(existingProvisions, cdLog)
```

```

Comment (abhinavdahiya) [reply pkg/controller/clusterdeployment/clusterdeployment_controller.go:652]: Also clenaup for PVC ?
L547
```go
		r.cleanupInstallLogPVC(cd, cdLog)
```

abhinavdahiya: both these functions seem to be happy in case none exist from quick glance, so i don't think it's required per-se.

PR #1297, opened 2021-03-01
Add support for Subnet, and Network for GCP MachinePools

nimrodshn:
cc: @igoihman  @dgoodwin  @staebler  @gregsheremeta

### DIFF
--- a/pkg/controller/remotemachineset/gcpactuator.go
+++ b/pkg/controller/remotemachineset/gcpactuator.go
@@ -53,6 +53,8 @@ type GCPActuator struct {
 	scheme    *runtime.Scheme
 	projectID string
 	imageID   string
+	network   string
+	subnet    string
 	// expectations is a reference to the reconciler's TTLCache of machinepoolnamelease creates each machinepool
 	// expects to see.
 	expectations   controllerutils.ExpectationsInterface
@@ -94,6 +96,12 @@ func NewGCPActuator(
 		return nil, err
 	}
 
+	network, subnet, err := getNetwork(remoteMachineSets, scheme, logger)
+	if err != nil {
+		logger.WithError(err).Error("error getting network information from remote machines")
+		return nil, err
+	}
+
 	actuator := &GCPActuator{
 		gcpClient:      gcpClient,
 		client:         client,
@@ -102,6 +110,8 @@ func NewGCPActuator(
 		expectations:   expectations,
 		projectID:      projectID,
 		imageID:        imageID,
+		network:        network,
+		subnet:         subnet,
 		leasesRequired: requireLeases(clusterVersion, remoteMachineSets, logger),
 	}
 	return actuator, nil
@@ -163,8 +173,10 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 	ic := &installertypes.InstallConfig{
 		Platform: installertypes.Platform{
 			GCP: &installertypesgcp.Platform{
-				Region:    cd.Spec.Platform.GCP.Region,
-				ProjectID: a.projectID,
+				Region:        cd.Spec.Platform.GCP.Region,
+				ProjectID:     a.projectID,
+				ComputeSubnet: a.subnet,
+				Network:       a.network,
 			},
 		},
 	}
@@ -449,6 +461,33 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
 	imageID := providerSpec.Disks[0].Image
 	logger.WithField("image", imageID).Debug("resolved image to use for new machinesets")
 	return imageID, nil
+}
+
+// getNetwork retrieves the network information (Network name and subnet)
+// from existing machines on the remote cluster.
+func getNetwork(remoteMachineSets []machineapi.MachineSet,
+	scheme *runtime.Scheme, logger log.FieldLogger) (string, string, error) {
+	if len(remoteMachineSets) == 0 {
+		return "", "", nil
+	}
+
+	remoteMachineProviderSpec, err := decodeGCPMachineProviderSpec(
+		remoteMachineSets[0].Spec.Template.Spec.ProviderSpec.Value,
+		scheme,
+	)
+	if err != nil {
+		logger.WithError(err).Warn("cannot decode GCPMachineProviderSpec from remote machinesets")
+		return "", "", errors.Wrap(err, "cannot decode GCPMachineProviderSpec from remote machinesets")
+	}
+
+	if len(remoteMachineProviderSpec.NetworkInterfaces) == 0 {
+		logger.Warn("remote machine do not have any network interfaces")
+		return "", "", errors.Wrap(err, "remote machine do not have any network interfaces")
+	}
+
+	network := remoteMachineProviderSpec.NetworkInterfaces[0].Network
+	subnet := remoteMachineProviderSpec.NetworkInterfaces[0].Subnetwork
+	return network, subnet, nil
 }
 
 func decodeGCPMachineProviderSpec(rawExt *runtime.RawExtension, scheme *runtime.Scheme) (*gcpproviderv1beta1.GCPMachineProviderSpec, error) {
--- a/pkg/controller/remotemachineset/gcpactuator_test.go
+++ b/pkg/controller/remotemachineset/gcpactuator_test.go
@@ -1,6 +1,7 @@
 package remotemachineset
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"testing"
@@ -13,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	jsonserializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -32,6 +34,8 @@ import (
 const (
 	gcpCredsSecretName = "gcp-credentials"
 	testProjectID      = "test-gcp-project-id"
+	testNetworkID      = "test-gcp-network-id"
+	testSubnetID       = "test-subnet-id"
 )
 
 func TestGCPActuator(t *testing.T) {
@@ -204,6 +208,8 @@ func TestGCPActuator(t *testing.T) {
 				expectations:   controllerExpectations,
 				projectID:      testProjectID,
 				leasesRequired: test.requireLeases,
+				network:        testNetworkID,
+				subnet:         testSubnetID,
 			}
 
 			generatedMachineSets, _, err := ga.GenerateMachineSets(clusterDeployment, test.pool, ga.logger)
@@ -223,6 +229,10 @@ func TestGCPActuator(t *testing.T) {
 					assert.True(t, ok, "failed to convert to gcpProviderSpec")
 
 					assert.Equal(t, testInstanceType, gcpProvider.MachineType, "unexpected instance type")
+
+					// Ensure network details are propagated correctly.
+					assert.Equal(t, ga.network, gcpProvider.NetworkInterfaces[0].Network)
+					assert.Equal(t, ga.subnet, gcpProvider.NetworkInterfaces[0].Subnetwork)
 
 					// Ensure GCP disk type and size was correctly set or defaulted and made it to the resulting MachineSets:
 					expectedDiskType := test.pool.Spec.Platform.GCP.OSDisk.DiskType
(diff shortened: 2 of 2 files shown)


pkg/apis/hive/v1/gcp/machinepools.go
```diff
@@ -9,6 +9,20 @@ type MachinePool struct {
...
 
+	// ControlPlaneSubnet defines the subnet where the control plane will be deployed.
+	// The value should be the name of the subnet.
+	// If ommitted one will be created by the installer.
+	ControlPlaneSubnet string `json:"control_plane_subnet,omitempty"`
+
+	// ComputeSubnet defines the subnet where the compute nodes will be deployed.
+	// The value should be the name of the subnet.
+	// If ommitted one will be created by the installer.
+	ComputeSubnet string `json:"compute_subnet,omitempty"`
```
abhinavdahiya, line 20: What is the goal of letting the user define the control plane and compute subnet separately? seems like the user should only care which subnet this machinepool should be deployed..

From what i can see MachinePools today are mostly used to manage the `compute` role in cluster, (infra is still compute with additional stipulation) ... So maybe we treat the subnet as installertypes.platform.gcp.computeSubnet .

pkg/apis/hive/v1/gcp/machinepools.go
```diff
@@ -9,6 +9,20 @@ type MachinePool struct {
 	// eg. n1-standard-4
 	InstanceType string `json:"type"`
 
+	// ControlPlaneSubnet defines the subnet where the control plane will be deployed.
+	// The value should be the name of the subnet.
+	// If ommitted one will be created by the installer.
```
abhinavdahiya, line 14: the doc comment seems not valid in the context of hive.

pkg/apis/hive/v1/gcp/machinepools.go
```diff
@@ -9,6 +9,20 @@ type MachinePool struct {
...
+	ControlPlaneSubnet string `json:"control_plane_subnet,omitempty"`
+
+	// ComputeSubnet defines the subnet where the compute nodes will be deployed.
+	// The value should be the name of the subnet.
+	// If ommitted one will be created by the installer.
```
abhinavdahiya, line 19: the doc comment seems not valid in the context of hive.

pkg/apis/hive/v1/gcp/machinepools.go
```diff
@@ -9,6 +9,20 @@ type MachinePool struct {
...
+	ComputeSubnet string `json:"compute_subnet,omitempty"`
+
+	// Network specifies an existing VPC where the cluster should be created
+	// rather than provisioning a new one.
+	Network string `json:"network,omitempty"`
```
abhinavdahiya, line 24: we must enforce that user set `network` and `subnet` both together or neither.

pkg/controller/remotemachineset/gcpactuator_test.go
```diff
@@ -141,6 +141,48 @@ func TestGCPActuator(t *testing.T) {
...
+				pool := testGCPPool(testPoolName)
+				pool.Spec.Platform.GCP.Network = "some-vpc"
+				return pool
+			}(),
+			mockGCPClient: func(client *mockgcp.MockClient) {
+				mockListComputeZones(client, []string{"zone1"}, testRegion)
+			},
+			expectedMachineSetReplicas: map[string]int64{
+				generateGCPMachineSetName("worker", "zone1"): 3,
+			},
```
abhinavdahiya, line 184: this is not a valid accepted case, setting a network with the subnet means hive can't know where to place the machine.

abhinavdahiya (changes requested):
I tried to explain why allowing users to set the subnet on day-2 machinepools for GCP is prone to user errors here.

https://issues.redhat.com/browse/HIVE-1401?focusedCommentId=15873830&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15873830
```
> https://github.com/kubernetes/kubernetes/blob/f6152d152140f42815765c7d9703b9817943c345/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_internal.go#L99-L104

If you look at the code for the k8s-cloud-provider for gcp that handles the internal_load_balancer, you will notice that it only supports ALL nodes in the

SAME subnet.

We carry a patch in OpenShift https://github.com/kubernetes/kubernetes/pull/84466 that allows OpenShift to run control plane nodes in a separate subnet,

and all the rest in the same subnet. So all nodes, compute/infra/ or other that have different meaning as nodes in cluster must continue to be in the same

subnet in GCP context.

So I'm concerned that allowing users to provide subnet for day-2 MachinePools can create problems where if those nodes join the cluster, reconcilation for

all internal_load_balancers will be broken.

 

Installer allows setting only one pool of compute machines at install time and therefore the users cannot end up in this multi subnet situation, but hive MachinePools with subnet option has high chances of user error.
```

I can think of 2 ways to move forward, detailed in https://issues.redhat.com/browse/HIVE-1401?focusedCommentId=15873853&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15873853
```
Can we get logs from the hive-controllers where BYO VPC was in use for GCP. The logs for remotemachineset controller should help understand what's broken.

Secondly, for now I would suggest that

A.

Hive use the already existing compute MachineSets in the cluster to configure the subnets/vpc for all the next MachinePools.

 

B.
If Hive supports the subnet on MachinePool, we enforce in validation webhooks that ALL MachinePools for a clusterdeployment use the same subnet.
```

pkg/apis/hive/v1/gcp/machinepools.go
nimrodshn: > What is the goal of letting the user define the control plane and compute subnet separately? seems like the user should only care which subnet this machinepool should be deployed..

I was just exposing the API to the user as I saw the installer allowed for it. Alas, as you've mentioned - this _currently_ doesn't have a use case

> From what i can see MachinePools today are mostly used to manage the compute role in cluster, (infra is still compute with additional stipulation) ... So maybe we treat the subnet as installertypes.platform.gcp.computeSubnet .

So leave out `ControlPlaneSubnet` :+1:

pkg/apis/hive/v1/gcp/machinepools.go
nimrodshn: Ack :+1:

pkg/controller/remotemachineset/gcpactuator_test.go
nimrodshn: @abhinavdahiya Did you mean a network _without_ a subnet? Will fix the test case.

pkg/apis/hive/v1/gcp/machinepools.go
nimrodshn: Should I remove the last sentence? what would you suggest?

pkg/apis/hive/v1/gcp/machinepools.go
```diff
@@ -9,6 +9,14 @@ type MachinePool struct {
 	// eg. n1-standard-4
 	InstanceType string `json:"type"`
 
+	// ComputeSubnet defines the subnet where the compute nodes will be deployed.
+	// The value should be the name of the subnet.
+	ComputeSubnet string `json:"compute_subnet,omitempty"`
```
abhinavdahiya, line 14:
```suggestion
	Subnet string `json:"subnet,omitempty"`
```

nimrodshn: Ack, thanks!

nimrodshn:
@abhinavdahiya @dgoodwin With regards to you're comment:
> If Hive supports the subnet on MachinePool, we enforce in validation webhooks that ALL MachinePools for a clusterdeployment use the same subnet.

What if instead of validation at the MachinePool level we instead move these fields (i.e. `Network` and `ComputeSubnet` ) to the ClusterDeployment `gcp.Platform` field with the semantics being:
```
 This is the subnet where ALL compute machinepools will reside in
```
That way the invariant (all compute machinepools reside in the same subnet) is kept by the Actuator instead of the webhook validation; thus eliminating room for user errors

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,32 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+	}
+
+	remoteMachineProviderSpec, err := decodeGCPMachineProviderSpec(
+		remoteMachineSets[0].Spec.Template.Spec.ProviderSpec.Value,
+		scheme,
+	)
+	if err != nil {
+		logger.WithError(err).Warn("cannot decode GCPMachineProviderSpec from master machine")
+		err = errors.Wrap(err, "cannot decode GCPMachineProviderSpec from master machine")
+		return
```
abhinavdahiya, line 480:
```suggestion
		return "", "", errors.Wrap(err, "cannot decode GCPMachineProviderSpec from master machine")
```

using implicit returns creates confusion when reading the code as it doesn't become clear immediately that we are returning values compared to empty return..

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,32 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+	}
+	if len(remoteMachineProviderSpec.NetworkInterfaces) == 0 {
+		logger.Warn("remove machine does not have any netowrk interfaces")
+		return
```
abhinavdahiya, line 484: same return explicit error and empty values here.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,32 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+	}
+
+	network = remoteMachineProviderSpec.NetworkInterfaces[0].Network
+	subnet = remoteMachineProviderSpec.NetworkInterfaces[0].Subnetwork
+	return
```
abhinavdahiya, line 489:
```suggestion
	return remoteMachineProviderSpec.NetworkInterfaces[0].Network, remoteMachineProviderSpec.NetworkInterfaces[0].Subnetwork, nil
```

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,32 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
 	return imageID, nil
 }
 
+// Get the network information from existing machines on the remote cluster.
+func getNetwork(remoteMachineSets []machineapi.MachineSet,
```
abhinavdahiya, line 467: let's make sure we add unit tests for this function.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -163,8 +173,10 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 	ic := &installertypes.InstallConfig{
 		Platform: installertypes.Platform{
 			GCP: &installertypesgcp.Platform{
-				Region:    cd.Spec.Platform.GCP.Region,
-				ProjectID: a.projectID,
+				Region:        cd.Spec.Platform.GCP.Region,
+				ProjectID:     a.projectID,
+				ComputeSubnet: a.subnet,
+				Network:       a.network,
```
abhinavdahiya, line 179: also add unit test that checks the network and subnet match when the actuator is configured with those.

nimrodshn: Will do.

pkg/controller/remotemachineset/gcpactuator.go
nimrodshn: Sure.

pkg/controller/remotemachineset/gcpactuator.go
nimrodshn: Done.

pkg/controller/remotemachineset/gcpactuator.go
nimrodshn: Done.

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,33 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+		remoteMachineSets[0].Spec.Template.Spec.ProviderSpec.Value,
+		scheme,
+	)
+	if err != nil {
+		logger.WithError(err).Warn("cannot decode GCPMachineProviderSpec from master machine")
+		return "", "", errors.Wrap(err, "cannot decode GCPMachineProviderSpec from master machine")
+	}
+
+	if len(remoteMachineProviderSpec.NetworkInterfaces) == 0 {
+		logger.Warn("remove machine does not have any netowrk interfaces")
```
abhinavdahiya, line 484:
```suggestion
		logger.Warn("remote machine do not have any network interfaces")
```

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,33 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+		return "", "", nil
```
abhinavdahiya, line 485: should this be an error? vs currently no op?

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,33 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+	if len(remoteMachineSets) == 0 {
+		return "", "", nil
+	}
+
+	remoteMachineProviderSpec, err := decodeGCPMachineProviderSpec(
+		remoteMachineSets[0].Spec.Template.Spec.ProviderSpec.Value,
+		scheme,
+	)
+	if err != nil {
+		logger.WithError(err).Warn("cannot decode GCPMachineProviderSpec from master machine")
```
abhinavdahiya, line 479:
```suggestion
		logger.WithError(err).Warn("cannot decode GCPMachineProviderSpec from remote machinesets")
```

pkg/controller/remotemachineset/gcpactuator.go
```diff
@@ -451,6 +463,33 @@ func getGCPImageID(masterMachine *machineapi.Machine, scheme *runtime.Scheme, lo
...
+		return "", "", errors.Wrap(err, "cannot decode GCPMachineProviderSpec from master machine")
```
abhinavdahiya, line 480:
```suggestion
		return "", "", errors.Wrap(err, "cannot decode GCPMachineProviderSpec from remote machinesets")
```

abhinavdahiya (commented):
just a few cosmetic changes and one question

pkg/controller/remotemachineset/gcpactuator.go
nimrodshn: Probably an error.

nimrodshn: Fixed.

dgoodwin:
i will force merge as soon as CI is clean

PR #1298, opened 2021-03-02
HIVE-1394: Add generic install failing reason ResourceLimitExceeded.

abutcher:
We're encountering more and more "*LimitExceeded" errors and this would capture all of them generically. Do we prefer this to having a regex for each type of resource limit exceeded to prevent needing to check logs.

Add tests for `"NATGatewayLimitExceeded"` and `"VPCLimitExceeded"`.

/assign @dgoodwin 
/cc @gregsheremeta

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -6,17 +6,27 @@ metadata:
 data:
   regexes: |
     # AWS Specific
+    - name: AWSNATGatewayLimitExceeded
+      searchRegexStrings:
+      - "NatGatewayLimitExceeded"
+      installFailingReason: AWSNATGatewayLimitExceeded
+      installFailingMessage: AWS NAT gateway limit exceeded
+    - name: AWSVPCLimitExceeded
+      searchRegexStrings:
+      - "VpcLimitExceeded"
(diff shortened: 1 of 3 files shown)


abhinavdahiya:
> Do we prefer this to having a regex for each type of resource limit exceeded to prevent needing to check logs.

what is current effect of order of regexes in the list here? Will we use the first match or the last match?
If we use the first match, we could keep the explicit ones above the generic one to catch all that miss..

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -6,17 +6,27 @@ metadata:
...
+      searchRegexStrings:
+      - "NatGatewayLimitExceeded"
+      installFailingReason: AWSNATGatewayLimitExceeded
+      installFailingMessage: AWS NAT gateway limit exceeded
+    - name: AWSVPCLimitExceeded
+      searchRegexStrings:
+      - "VpcLimitExceeded"
+      installFailingReason: AWSVPCLimitExceeded
+      installFailingMessage: AWS VPC limit exceeded
+    - name: LimitExceeded
```
abhinavdahiya, line 19: nit: Can we add an comment here that all limit exceeded regexes must be above this entry to allow specific matches before catch-all.

PR #1300, opened 2021-03-03
Point sigs.k8s.io/controller-runtime to a fork

suhanime:
In an effort to reduce the storage for nonessential memory-intensive
metrics, rest_client_request_latency_seconds was identified as the
metric that reports and stores the most time series. This metric was
vendored in from sigs.k8s.io/controller-runtime.
However, it has since been deprecated in kube 1.14 and turned off
in kube 1.17. Since upstream still registers this metric, it was
consequently decided to point controller-runtime to a fork that carries
the patch to not report this metric

/assign @dgoodwin 
/cc @gregsheremeta

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -110,3 +110,6 @@ replace k8s.io/client-go => k8s.io/client-go v0.20.0
 
 // needed for fixing CVE-2020-29529
 replace github.com/hashicorp/go-slug => github.com/hashicorp/go-slug v0.5.0
+
+// point sigs.k8s.io/controller-runtime to forked version
+replace sigs.k8s.io/controller-runtime => github.com/openshift-hive/controller-runtime v0.6.2-openshift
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -110,3 +110,6 @@ replace k8s.io/client-go => k8s.io/client-go v0.20.0
 
 // needed for fixing CVE-2020-29529
 replace github.com/hashicorp/go-slug => github.com/hashicorp/go-slug v0.5.0
+
+// point sigs.k8s.io/controller-runtime to forked version
+replace sigs.k8s.io/controller-runtime => github.com/openshift-hive/controller-runtime v0.6.2
```
dgoodwin, line 115: Normally here we'd make a branch off the tag we want to base on, something like v0.6.2-patched. Creating our own tags gets confusing as we're not making official releases and could cause a nasty conflict if they were to create the same tag when we pull it in later. I'd suggest trying to delete v0.6.2 and v0.6.2.1 and setting up the branch.

You can just swap a tag here in this line for the branch name, I think it's all refs as far as git is concerned.

Patch in the branch looks good, everything gtg otherwise. Thanks!

abhinavdahiya: my 2 cents,
release-0.6.2 as the branch name if that is still not done.
and v0.6.2-openshift tag on the desired commit, makes things very clear.

PR #1304, opened 2021-03-10
hack/update-codegen.sh: make sure deepcopy functions are updated in apis

abhinavdahiya:
the code-generator seems to generate/update the deepcopy functions in
the vendor directory where github.com/openshift/hive/apis package exists
from the context of the root package.

the code-generator fails to run from the apis submodule, so the next
best thing is to copy the deepcopy functions from the vendor to apis
submodule.


dgoodwin:
/lgtm

I think I was hitting this in my agent apis PR.

PR #1308, opened 2021-03-16
Add support for provisioning clusters to use AWS STS temporary credentials

dgoodwin:
See includes docs, which need a little more work but wanted to get up for review today.

/assign @joelddiaz 
/cc @akhil-rane


abhinavdahiya:
@dgoodwin tthis PR seems to tie the service account signing key to AWS platform, when technically it is not. we indent to use for for supporting AWS STS but it's not really not limited to that platform. So i think we should move to to provisioning section of the clusterdeployment as this is not configured day-1 and not reconciled later using Hive.

dgoodwin:
> @dgoodwin tthis PR seems to tie the service account signing key to AWS platform, when technically it is not. we indent to use for for supporting AWS STS but it's not really not limited to that platform. So i think we should move to to provisioning section of the clusterdeployment as this is not configured day-1 and not reconciled later using Hive.

Ah very good, I will make the adjustments.

joelddiaz:
Do we want to bake in this signed serviceaccount signing key functionality into the hive API? Hive (presently) doesn't need this information (aside from the initial install), and my gut feeling is that the person/code creating the ClusterDeployment should be able to slip in any kind of manifest into the install pod (eg a post-manifests list of files).

vendor/github.com/openshift/hive/apis/hive/v1/aws/platform.go
```diff
@@ -17,4 +17,24 @@ type Platform struct {
...
+	// +optional
+	STS *STS `json:"sts,omitempty"`
+}
+
+// STS specifies configuration for deploying a cluster that uses the AWS Security Token Service, instead of long lived credentials.
+type STS struct {
+	// ServiceAccountIssuerKeySecretRef refers to a secret that contains a 'bound-service-account-signing-key.key' data key pointing to the private key that will be used to sign ServiceAccount objects.
+	ServiceAccountIssuerKeySecretRef corev1.LocalObjectReference `json:"serviceAccountIssuerKeySecretRef"`
+
+	CredentialsRequestRoles []CredentialsRequestRoles `json:"credentialsRequestRoles"`
```
joelddiaz: Similar comment to the one about having the signing key be part of the API. What if the Hive API had a .spec.preInstallManifests[] that was just filename/path with yaml/json that would be inserted into the install pod at the appropriate time before calling 'create cluster'?
Or maybe even just a single Secret where each key/value pair was the file and the content to insert...

contrib/pkg/createcluster/create.go
```diff
@@ -247,6 +249,9 @@ create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=ovirt --ovirt-api-vip 192.168.1.2
 	flags.StringVar(&opt.DeleteAfter, "delete-after", "", "Delete this cluster after the given duration. (i.e. 8h)")
 	flags.StringVar(&opt.HibernateAfter, "hibernate-after", "", "Automatically hibernate the cluster whenever it has been running for the given duration")
 	flags.StringVar(&opt.PullSecretFile, "pull-secret-file", defaultPullSecretFile, "Pull secret file for cluster")
+	flags.StringVar(&opt.BoundServiceAccountSigningKeyFile, "bound-service-account-signing-key-file", "", "Private signing key for AWS STS, often created with ccoutil create key-pair.")
+	flags.BoolVar(&opt.CredentialsModeManual, "credentials-mode-manual", false, "Configure the Cloud Credential Operator in the target cluster to Manual mode. Implies the use of --manifests-dir to inject custom Secrets for all CredentialsRequests in the cluster.")
```
joelddiaz, line 253: Both of these should be non-AWS specific. It is perfectly valid to deploy a GCP cluster with CCO in manual mode (assuming you provide the Secrets with the appropriate creds for the components), and also valid to want a custom serviceaccount signing key on any platform.

contrib/pkg/createcluster/create.go
```diff
@@ -375,6 +380,21 @@ func (o *Options) Validate(cmd *cobra.Command) error {
 		}
 	}
 
+	if o.BoundServiceAccountSigningKeyFile != "" {
+		if o.Cloud != cloudAWS {
```
joelddiaz, line 384: anyone should be allowed to bring a custom signing key.

contrib/pkg/createcluster/create.go
```diff
@@ -563,6 +583,14 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
 			UserTags:        userTags,
 			Region:          o.Region,
 		}
+		if len(o.BoundServiceAccountSigningKeyFile) != 0 {
```
joelddiaz: not just AWS

pkg/clusterresource/aws.go
```diff
@@ -110,6 +140,10 @@ func (p *AWSCloudBuilder) addInstallConfigPlatform(o *Builder, ic *installertype
 	ic.ControlPlane.Platform.AWS = mpp
 	ic.Compute[0].Platform.AWS = mpp
 
+	if len(p.BoundServiceAccountSigningKey) > 0 {
+		ic.CredentialsMode = installertypes.ManualCredentialsMode
+	}
+
```
joelddiaz: custom SA signing is not necessarily tied to CCO cred mode

dgoodwin:
> Do we want to bake in this signed serviceaccount signing key functionality into the hive API? Hive (presently) doesn't need this information (aside from the initial install), and my gut feeling is that the person/code creating the ClusterDeployment should be able to slip in any kind of manifest into the install pod (eg a post-manifests list of files).

This is a very good question/observation. Hadn't really thought about it until now beyond seeing that we can't inject into TLS dir with what we have today.

I've long felt pretty uneasy about the way Hive has this big blob of install yaml required to create a ClusterDeployment. I know why we did it (to let Hive provision new install-config features without needing an update), but I dislike that there's this undefined hole in "the Kube API to provision OpenShift" which Kube doesn't know anything about, and isn't validated until runtime, and even then you're digging through pod logs to see what you did wrong.

 Last time we talked about this we started poking around the idea that maybe we one day could stop requiring it, and generate one for you, but you can always specify your own if you want to override/bypass/use something we don't know about. I personally think we should start moving in this direction. With the agent install efforts they're using CD but no IC, and we are planning on mirroring *some* IC fields on CD to do that for things we expect are needed in the at scale Hive hub cluster use cases. I think STS would meet the bar for this and as such, I'm thinking it's probably ok to expose bound service signing token.

If we're not in agreement on that, we'd need some way to inject to tls/ dir, I can't think how we'd do that with the existing manifests configmap. It feels wrong to add another configmap for tls/ dir. That said we did talk about a way to inject Secrets (instead of a ConfigMap), as this is currently expecting those Role cred Secrets in a ConfigMap in the hub.

But overall I'd like to see us moving towards Hive's API being more of a first class citizen rather than a minimal Kube API with a non-Kube API bolted onto the side.

Open to discussion as always.

docs/aws-sts-provisioning.md
```diff
@@ -0,0 +1,78 @@
+# Provisioning AWS STS Clusters
+
+It is possible to use Hive to provision clusters configured to use Amazon's Security Token Service, where cluster components use short lived credentials that are rotated frequently. This feature was added to the in-cluster Cloud Credential Operator in 4.7, see documentation [here](https://docs.openshift.com/container-platform/4.7/authentication/managing_cloud_provider_credentials/cco-mode-sts.html).
```
joelddiaz: I'd word it like: "This feature was added to the in-cluster OpenShift components in 4.7".

docs/aws-sts-provisioning.md
```diff
@@ -0,0 +1,78 @@
...
+
+```bash
+$ mkdir credrequests/
+$ oc adm release extract quay.io/openshift-release-dev/ocp-release:4.7.1-x86_64 --credentials-requests --cloud=aws > credrequests/credrequests.yaml
+```
+
+## Generate AWS IAM Roles
+
+```bash
+$ create iam-roles --credentials-requests-dir /home/dgoodwin/go/src/github.com/openshift/cloud-credential-operator/sts/credrequests/ --identity-provider-arn arn:aws:iam::125931421481:oidc-provider/dgoodsts2-oidc.s3.us-east-1.amazonaws.com --name-prefix dgoodsts2 --region us-east-1
```
joelddiaz, line 22: `ccoctl create iam-roles`

joelddiaz:
I'm definitely not opposed to having Hive expose more of the tunables as official API objects. But we should have an escape hatch for when someone needs to bring a change that isn't specifically handled by the Hive API. If we had that ability today, this would be one less code-change given our aggressive timelines.

We may end up writing many controllers to handle the day-2 changes of these fields (like the ServiceAccount signing key), but that's probably a good thing long-term for the functionality that Hive brings to the table.

If our current manifests injection methods assume that everything goes into the literal 'manifests' directory, then I see we have a definite need to expose this in the Hive API as this is indeed different from the other manifests we normally deal with.

dgoodwin:
> I'm definitely not opposed to having Hive expose more of the tunables as official API objects. But we should have an escape hatch for when someone needs to bring a change that isn't specifically handled by the Hive API. If we had that ability today, this would be one less code-change given our aggressive timelines.
> 
> We may end up writing many controllers to handle the day-2 changes of these fields (like the ServiceAccount signing key), but that's probably a good thing long-term for the functionality that Hive brings to the table.

Oh I've got this in the provisioning sub-struct, that should probably change if this is going to be a thing.

> 
> If our current manifests injection methods assume that everything goes into the literal 'manifests' directory, then I see we have a definite need to expose this in the Hive API as this is indeed different from the other manifests we normally deal with.

Are there lots of other things users can do to inject files outside of manifests? cc @abhinavdahiya We could do a new sort of replacement for our manifest configmap that would live along side it, allowing you to map a secret or a configmap full of files, to a relative path in the installer workdir.

joelddiaz:
> Are there lots of other things users can do to inject files outside of manifests? cc @abhinavdahiya We could do a new sort of replacement for our manifest configmap that would live along side it, allowing you to map a secret or a configmap full of files, to a relative path in the installer workdir.

+1 This theoretical ability to have a Secret with relative path/filenames (and their contents) would solve any future use cases going forward (without someone having to wait for Hive API support).

contrib/pkg/createcluster/create.go
dgoodwin: probably doesn't require manual mode below either does it?

dgoodwin:
This one probably is reasonable for a future PR and doesn't need to be in this one, provided we're ok with exposing the bound SA token in Hive's API. I'd also like to find out the scope of other dirs you can inject to and how often they'd be used before we commit.

dgoodwin:
Updated for review comments.

vendor/github.com/openshift/hive/apis/hive/v1/clusterdeployment_types.go
```diff
@@ -206,6 +206,10 @@ type Provisioning struct {
 	// Defaults to openshift-install if none specified.
 	// +optional
 	InstallStrategy *InstallStrategy `json:"installStrategy,omitempty"`
+
+	// ServiceAccountIssuerKeySecretRef refers to a Secret that contains a 'bound-service-account-signing-key.key' data key pointing to the private key that will be used to sign ServiceAccount objects. Primarily used to provision AWS clusters to use Amazon's Security Token Service.
+	// +optional
+	ServiceAccountIssuerKeySecretRef *corev1.LocalObjectReference `json:"serviceAccountIssuerKeySecretRef,omitempty"`
```
dgoodwin, line 212: Would this be better named BoundServiceAccountSigningKeySecretRef to match the installer filename?

Also why am I not seeing the change in top level apis/ ? Only the vendored version?

dgoodwin: Lol I've been ctrl-p'ing in Vim, picking up the vendored copy, modifying it, everything works, and missing the actual source code in apis. Gonna have to watch out for that.

dgoodwin:
Renamed the API field to BoundServiceAccountSigningKeySecretRef, and moved to top level Spec instead of Provisioning, as we expect this will be mutable someday when supported by the product. By default our validation considers it immutable now.

Also fixed to edit the correct API files and not the vendored versions. (oops)

docs/aws-sts-provisioning.md
```diff
@@ -0,0 +1,74 @@
...
+$ oc adm release extract quay.io/openshift-release-dev/ocp-release:4.7.1-x86_64 --credentials-requests --cloud=aws > credrequests/credrequests.yaml
+```
+
+
+## Setup STS Infrastructure
+
+```bash
+$ ccoctl create key-pair
+$ ccoctl create identity-provider --name-prefix mystsprefix --public-key-file serviceaccount-signer.public --region us-east-1
+$ ccoctl create iam-roles --credentials-requests-dir /home/dgoodwin/go/src/github.com/openshift/cloud-credential-operator/sts/credrequests/ --identity-provider-arn arn:aws:iam::125931421481:oidc-provider/dgoodsts2-oidc.s3.us-east-1.amazonaws.com --name-prefix dgoodsts2 --region us-east-1
```
abhinavdahiya, line 22: these commands are not consistent wrt to name prefix and the resources that are being provided.
Can we generalize them?

also `125931421481` sounds like a real AWS account id, let' not use that.

docs/aws-sts-provisioning.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+May also soon be automated by ccoctl.
+
+## Create Hive ClusterDeployment
+
+Create a ClusterDeployment normally with the following changes:
+
+  1. Create a Secret for your private service account signing key created with ccoctl key-pair above: `kubectl create secret generic bound-service-account-signing-key --from-file=bound-service-account-signing-key.key=serviceaccount-signer.private`
+  1. Create a ConfigMap with keys for each IAM Role Secret. Each key will be the file name provided to the installer: `kubectl create configmap cluster-manifests --from-file=manifest1.yaml=manifests/secret1.yaml --from-file=manifest2.yaml=manifests/secret2.yaml`
```
abhinavdahiya, line 70:
```console
$ oc create configmap -h
...
--from-file=[]: Key file can be specified using its file path, in which case file basename will be used as
configmap key, or optionally with a key and file path, in which case the given key will be used.  Specifying a directory
will iterate each named file in the directory whose basename is a valid configmap key.
...
```

> Specifying a directory
will iterate each named file in the directory whose basename is a valid configmap key.

why not just point to the directory and let the file names be the keys.. ?

pkg/constants/constants.go
```diff
@@ -191,6 +191,14 @@ const (
 	// path where we mount in the SSH key for connecting to the bare metal libvirt provisioning host.
 	LibvirtSSHPrivKeyPathEnvVar = "LIBVIRT_SSH_PRIV_KEY_PATH"
 
+	// BoundServiceAccountSigningKeyEnvVar contains the path to the bound service account signing key and
+	// is set in the install pod for AWS STS clusters.
+	BoundServiceAccountSigningKeyEnvVar = "AWS_BOUND_SA_SIGNING_KEY"
```
abhinavdahiya, line 196: needs a rename to not include AWS

pkg/install/generate.go
```diff
@@ -21,15 +21,17 @@ import (
...
-	openStackCADir     = "/etc/openstack-ca"
-	vsphereCloudsDir   = "/vsphere"
-	ovirtCloudsDir     = "/.ovirt"
-	ovirtCADir         = "/.ovirt-ca"
+	azureAuthDir             = "/.azure"
+	azureAuthFile            = azureAuthDir + "/osServicePrincipal.json"
+	gcpAuthDir               = "/.gcp"
+	gcpAuthFile              = gcpAuthDir + "/" + constants.GCPCredentialsName
+	awsDir                   = "/.aws"
+	awsBoundSASigningKeyFile = awsDir + "/" + constants.BoundServiceAccountSigningKeyFile
```
abhinavdahiya, line 29: needs a rename to be not restricted to aws

docs/aws-sts-provisioning.md
dgoodwin: Super handy, had no idea.

pkg/install/generate.go
dgoodwin: done, thanks!

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -142,6 +142,10 @@ type ClusterDeploymentSpec struct {
 	// provisioning worker machines.
 	// +optional
 	MachineManagement *MachineManagement `json:"machineManagement,omitempty"`
+
+	// BoundServiceAccountSignkingKeySecretRef refers to a Secret that contains a 'bound-service-account-signing-key.key' data key pointing to the private key that will be used to sign ServiceAccount objects. Primarily used to provision AWS clusters to use Amazon's Security Token Service.
```
abhinavdahiya, line 146: super nit: can you wordwrap the lines at around 80-90, like the other comments. this is currently single line..

docs/aws-sts-provisioning.md
```diff
@@ -0,0 +1,74 @@
...
+  name: installer-cloud-credentials
+  namespace: openshift-image-registry
+type: Opaque
+```
+
+
+### Create Authentication Manifest
+
+In the same directory as your Secret manifets, add another to configure the OpenShift Authentication operator to use the S3 OIDC provider created by `ccoctl create identity-provider`.
+
```
abhinavdahiya, line 53: NOTE:  the file name must be `cluster-authentication-02-config.yaml` in the manifests dir as https://github.com/openshift/installer/blob/82f88102baa6960cf64a59d8ed4a6d8a996839ff/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L169

abhinavdahiya (commented):
One super nit, and one update to documentation for filename of authentications.config.openshift.io object

dgoodwin:
Updated, thanks!

/assign @abhinavdahiya

PR #1310, opened 2021-03-22
Fix clusters stuck resuming due to stopped instance.

dgoodwin:
x-ref: https://issues.redhat.com/browse/HIVE-1455

Encountered a cluster deployment stuck resuming for several days, and
thus was not powered off after the hibernateAfter interval.

Issue appeared to be because one worker VM was still in stopped state.
The code made the assumption that we must have issued the start command
when transitioning to resuming state, and then never tried to again.

To fix we now will issue StartMachines every time we check if we're
resumed. All three providers are implemented to only issue to VMs that
are stopped / stopping so the change should have no impact normally.

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -331,6 +331,11 @@ func (r *hibernationReconciler) checkClusterResumed(cd *hivev1.ClusterDeployment
 		return reconcile.Result{}, err
 	}
 	if !running {
+		// Ensure all machines have been started. Should have been handled already but we've seen VMs left in stopped state.
+		if err := actuator.StartMachines(cd, r.Client, logger); err != nil {
+			logger.WithError(err).Error("error starting machines")
+			return reconcile.Result{}, err
+		}
 		return reconcile.Result{RequeueAfter: stateCheckInterval}, nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -325,6 +325,11 @@ func (r *hibernationReconciler) checkClusterResumed(cd *hivev1.ClusterDeployment
 		logger.Warning("No compatible actuator found to check machine status")
 		return reconcile.Result{}, nil
 	}
+	// Ensure all machines have been started. Should have been handled already but we've seen VMs left in stopped state.
+	if err := actuator.StartMachines(cd, r.Client, logger); err != nil {
+		logger.WithError(err).Error("error starting machines")
+		return reconcile.Result{}, err
+	}
```
akhil-rane, line 332: Why are we not executing this code after the check for `running` machines? If machines are not running issue start command again

dgoodwin: Hmm good point, this would avoid an additional API call if they did happen to all be running. I'll drop it down.

dgoodwin: Updated!

akhil-rane (commented):
/lgtm

PR #1312, opened 2021-03-23
Sync changes to managed secrets in cd namespace to target namespace.

abutcher:
https://issues.redhat.com/browse/HIVE-1412
also addresses https://issues.redhat.com/browse/HIVE-1458

/assign @dgoodwin 
/cc @suhanime

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -41,6 +41,7 @@ import (
 	hiveintv1alpha1 "github.com/openshift/hive/apis/hiveinternal/v1alpha1"
 	"github.com/openshift/hive/pkg/constants"
 	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
+	"github.com/openshift/hive/pkg/controller/utils"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/imageset"
 	"github.com/openshift/hive/pkg/install"
@@ -518,6 +519,21 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, nil
 	}
 
+	// Add labels to secrets referenced by cluster deployment for machine management controller
+	if err := r.addClusterDeploymentNameLabelToSecret(utils.CredentialsSecretName(cd), cd, cdLog); err != nil {
+		return reconcile.Result{}, err
+	}
+	if cd.Spec.PullSecretRef != nil {
+		if err := r.addClusterDeploymentNameLabelToSecret(cd.Spec.PullSecretRef.Name, cd, cdLog); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	if cd.Spec.Provisioning.SSHPrivateKeySecretRef != nil {
+		if err := r.addClusterDeploymentNameLabelToSecret(cd.Spec.Provisioning.SSHPrivateKeySecretRef.Name, cd, cdLog); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
 	if cd.Spec.Installed {
 		// set installedTimestamp for adopted clusters
 		if cd.Status.InstalledTimestamp == nil {
@@ -654,6 +670,21 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 	}
 
 	return r.reconcileExistingProvision(cd, cdLog)
+}
+
+func (r *ReconcileClusterDeployment) addClusterDeploymentNameLabelToSecret(secretName string, cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	secret := &corev1.Secret{}
+	if err := r.Get(context.Background(), types.NamespacedName{Name: secretName, Namespace: cd.Namespace}, secret); err == nil {
+		if !metav1.HasLabel(secret.ObjectMeta, constants.ClusterDeploymentNameLabel) {
+			secret.Labels = k8slabels.AddLabel(secret.Labels, constants.ClusterDeploymentNameLabel, cd.Name)
+			err := r.Update(context.TODO(), secret)
+			if err != nil {
+				cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update secret ", secretName)
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 func isDNSNotReadyConditionSet(cd *hivev1.ClusterDeployment) (bool, *hivev1.ClusterDeploymentCondition) {
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -41,7 +41,6 @@ import (
 	hiveintv1alpha1 "github.com/openshift/hive/apis/hiveinternal/v1alpha1"
 	"github.com/openshift/hive/pkg/constants"
 	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -520,6 +521,21 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, nil
 	}
 
+	// Add labels to secrets referenced by cluster deployment for machine management controller
```
abutcher, line 524: Should this labeling be done in the machine management controller?

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -176,57 +186,64 @@ func (r *ReconcileMachineManagement) reconcile(request reconcile.Request, cd *hi
...
+		return nil
+	}
+
+	origSecret := &corev1.Secret{}
+	err = r.Get(context.Background(), types.NamespacedName{Name: secretName, Namespace: cd.Namespace}, origSecret)
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s: %w", secretName, err)
+	}
+	// Update secret in targetNamespace
+	if !reflect.DeepEqual(origSecret.Data, secret.Data) {
```
abutcher, line 235: Only comparing `secret.Data`, do we care about anything else?

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -254,3 +271,26 @@ func (r *ReconcileMachineManagement) addAnnotationToTargetNamespace(cd *hivev1.C
...
+	secret := a.Object.(*corev1.Secret)
+	if secret == nil {
+		// Wasn't a Secret, bail out. This should not happen.
+		log.Errorf("Error converting MapObject.Object to Secret. Value: %+v", a.Object)
+		return retval
+	}
+	if secret.Labels == nil {
+		return retval
+	}
+	cdName, ok := secret.Labels[constants.ClusterDeploymentNameLabel]
```
dgoodwin, line 287: So it looks like we're watching all Secrets just quickly discarding any without our label.

How can we effeciently do this now? Top of my head, maybe we annotate it with a json list of the target namespaces it should be synced to? This will not be fun to maintain though if someone is doing multiple CDs in one namespace re-using that Secret. Need to think on this a bit. Also horrible things people could do to fire Secrets into namespaces they have no business in...

pkg/controller/machinemanagement/machinemanagement_controller.go
abhinavdahiya: i would like us to reverse this, if all equal return nil.

abhinavdahiya: secrets also have StringData that we should check

dgoodwin: Probably don't need to worry about stringData, these are both loaded from etcd, and StringData is just a convenience and is merged into Data on write.

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -176,57 +186,64 @@ func (r *ReconcileMachineManagement) reconcile(request reconcile.Request, cd *hi
...
+		if err != nil {
+			return fmt.Errorf("failed to get secret %s: %w", secretName, err)
+		}
+		secret.Namespace = targetNamespace
+		secret.ResourceVersion = ""
+		if err := r.Create(context.TODO(), secret); err != nil && !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create secret in target namespace: %v", err)
+		}
+		return nil
+	}
```
abhinavdahiya, line 227: We need err not nil check here too since th previous only responds to not found errors

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -82,6 +85,50 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler, concurrentReconci
...
+		if cd.Spec.PullSecretRef != nil {
+			res = append(res, cd.Spec.PullSecretRef.Name)
+		}
+		if cd.Spec.Provisioning.SSHPrivateKeySecretRef != nil {
+			res = append(res, cd.Spec.Provisioning.SSHPrivateKeySecretRef.Name)
+		}
+		return res
+	})
+	if err != nil {
+		log.WithField("controller", ControllerName).WithError(err).Error("Error indexing cluster deployment secrets")
```
abhinavdahiya, line 103: not error return here?

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -82,6 +85,50 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler, concurrentReconci
...
+	// Watch for changes to Secret referenced by cluster deployment
+	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+		retval := []reconcile.Request{}
+
+		secret := a.(*corev1.Secret)
+		if secret == nil {
+			// Wasn't a Secret, bail out. This should not happen.
+			log.Errorf("Error converting MapObject.Object to Secret. Value: %+v", a)
+			return retval
+		}
```
abhinavdahiya, line 115: this will not prevent a panic when a is not type `*corev1.Secret` i think. we probably need `typed, ok := a.(*corev1.Secret)`

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -176,59 +223,69 @@ func (r *ReconcileMachineManagement) reconcile(request reconcile.Request, cd *hi
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			cdLog.Infof("Creating secret %s in the target namespace %s", secretName, targetNamespace)
+			err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: secretName}, secret)
+			if err != nil {
+				return fmt.Errorf("failed to get secret %s: %w", secretName, err)
+			}
+			secret.Namespace = targetNamespace
+			secret.ResourceVersion = ""
+			if err := r.Create(context.TODO(), secret); err != nil && !apierrors.IsAlreadyExists(err) {
```
abhinavdahiya, line 261: feels odd to skip already exists error when we know that we decided to create the secret after checking it did not exist in the target namesapce.

if the secret ends up already  existing, i think we should fail and rety to update the secret.

pkg/controller/machinemanagement/machinemanagement_controller.go
```diff
@@ -256,3 +313,10 @@ func (r *ReconcileMachineManagement) addAnnotationToTargetNamespace(cd *hivev1.C
 
 	return nil
 }
+
+// initializeAnnotations() initializes the annotations if it is not already
+func initializeAnnotations(cd *hivev1.ClusterDeployment) {
+	if cd.Annotations == nil {
+		cd.Annotations = map[string]string{}
+	}
+}
```
abhinavdahiya, line 322: not needed now it seems

abutcher: Removed.

pkg/controller/machinemanagement/machinemanagement_controller.go
abutcher: Removed the `apierrors.IsAlreadyExists(err)` and will return the error here.

pkg/controller/machinemanagement/machinemanagement_controller.go
abutcher: Whoops, added.

abhinavdahiya (commented):
/lgtm

PR #1316, opened 2021-03-24
Add a temporary Fedora 33 dev docker build/push setup.

dgoodwin:
I've been using this for a long time, but ideally it would be nice to see cleaned up and a recommended dev path in the docs.

For now, just publishing to help out Andrew.

/assign @abutcher

### DIFF
--- a/hack/fedora33dev/docker-dev-push.sh
+++ b/hack/fedora33dev/docker-dev-push.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -e
-IMG=quay.io/dgoodwin/hive:latest
+
+# exit if IMG is not set, should be a ref to a registry you can push to (i.e. quay.io/user/hive:latest)
+set -u
+: "$IMG"
 
 # TODO: update not strictly required here unless dealing with manifest changes
 make update build
--- a/hack/fedora33dev/docker-dev-push.sh
+++ b/hack/fedora33dev/docker-dev-push.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -e
+IMG=quay.io/dgoodwin/hive:latest
+
+# TODO: update not strictly required here unless dealing with manifest changes
+make update build
+docker build -t ${IMG} -f hack/fedora33dev/Dockerfile.dev .
+docker push ${IMG}
+kubectl delete pod -n hive -l control-plane=clustersync --wait=false
+kubectl delete pod -n hive -l control-plane=controller-manager --wait=false
+kubectl delete pod -n hive -l control-plane=hive-operator --wait=false
+kubectl delete pod -n hive -l app=hiveadmission --wait=false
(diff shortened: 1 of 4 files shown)


hack/fedora33dev/docker-dev-push.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
+IMG=quay.io/dgoodwin/hive:latest
```
abutcher, line 3: Works for me, just need to drop this `IMG` override

dgoodwin: Updated, propose we merge and iterate if others start using.

abhinavdahiya:
can some one explain how this is better or what it solves compared to make buildah-dev-push ??

dgoodwin:
There's a glibc conflict if you build binaries on F33 and then run them on a Docker image that isn't F33. Containers instantly crash.

abhinavdahiya:
> There's a glibc conflict if you build binaries on F33 and then run them on a Docker image that isn't F33. Containers instantly crash.

so the binaries are being built on F33 host directly and then an image is created using those binaries by just copying??
why not just build the binaries in the container like buildah-dev-push does ? that allows control over the build env like the go versions etc?

dgoodwin:
Yes building directly on how and adding into a container. The reason is development speed:

make update image-hive  109.26s user 29.10s system 49% cpu 4:39.80 total
IMG="quay.io/dgoodwin/hive:latest" hack/fedora33dev/docker-dev-push.sh  143.45s user 34.15s system 203% cpu 1:27.43 total

For the latter I disabled docker push && deleting pods for a better comparison.

It might be possible to improve the docker container layering because this pulls in and installs dozens of rpms every execution, but even still the compile time along on dev machine vs fresh in a container is massively faster, I assume there's some caching going on.

For devel this saves us a heap of time/bandwidth as everything sits today.

abhinavdahiya:
> For devel this saves us a heap of time/bandwidth as everything sits today.

i think there are better ways to speed up builds for dev workflow than to hack and create parallel paths that imo do not provide consistent builds.

i see https://www.redhat.com/sysadmin/speeding-container-buildah and https://www.docker.com/blog/containerize-your-go-developer-environment-part-2/
that show how we can speed up the builds in our existing setup by re-using host cached for dnf and go-build

hive repo should have supported dev paths for new, existing users. this seems like adding code to repo that maybe only one person will use and that's not good.

abhinavdahiya:
As for fastest dev flow there is already documented paths for building binary locally and running it on host using kubeconfig...
it's not best because hive-controllers require various heterogeneous configurations and setting all these up is tedious locally, maybe we improve that to support running locally or build-push-run in cluster.

dgoodwin:
All reasonable ideas but not solutions we have right now, I'm open to jira's and PRs to improve, but the tooling in this PR is helpful until something better is available. it has worked well for me for months without causing any issues.

PR #1317, opened 2021-03-25
Exports AWS credentials in logextractor to fix auth bug

clcollins:
As written, the `logextractor.sh` script fails to authenticate to AWS
with error:

```
Unable to locate credentials. You can configure credentials by running "aws configure".
```

This PR exports the secret access key and access key ID so subsequent
AWS calls succeed.

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>


akhil-rane (commented):
/lgtm

PR #1318, opened 2021-03-26
Upgrade to controller-runtime 0.7.0

dgoodwin:
A number of invasive changes required here not limited to:

* Reconcile func now requires a context
* Broad use of client.Object rather than runtime.Object
* New isUnstructured boolean arg for RESTClientForGVK
* Events now require Object not an ObjectKey (namespace/name)
* Changes to EnqueueRequestsFromMapFunc

The highest risk areas here are:

1. dnszone and dnsendpoint controllers. We have a GenericEvent being emitted used to trigger reconciles which switched from an object key to full object. Plumbing this through was tricky and wreaked havoc on test assertions.
2. Use of context instead of stop functions. (not an area I'm super familiar with)

x-ref: https://issues.redhat.com/browse/HIVE-1452

/assign @abhinavdahiya @joelddiaz


dgoodwin:
Tests are passing and verify is clear, I am doing manual testing of a manageDNS cluster at the moment and it has made it past DNS and launched the install, so looking good so far.

apis/go.mod
```diff
@@ -6,5 +6,5 @@ require (
 	github.com/openshift/api v0.0.0-20210216211028-bb81baaf35cd
 	k8s.io/api v0.20.0
 	k8s.io/apimachinery v0.20.0
-	sigs.k8s.io/controller-runtime v0.6.2
+	sigs.k8s.io/controller-runtime v0.7.0
```
abhinavdahiya, line 9: should we bump this to 0.7.3 as the PR title suggests?

pkg/controller/clusterpool/clusterdeploymentexpectations.go
```diff
@@ -13,21 +14,19 @@ import (
...
+		handler.EnqueueRequestsFromMapFunc(
+			func(a client.Object) []reconcile.Request {
+				cpKey := clusterPoolKey(a.(*hivev1.ClusterDeployment))
+				if cpKey == nil {
+					return nil
+				}
+				return []reconcile.Request{{NamespacedName: *cpKey}}
+			},
+		),
+		r,
```
abhinavdahiya, line 27: struct is being initialized without Field names https://github.com/uber-go/guide/blob/master/style.md#use-field-names-to-initialize-structs

pkg/controller/dnsendpoint/nameserverscraper.go
```diff
@@ -123,7 +125,8 @@ func (s *nameServerScraper) Start(stop <-chan struct{}) error {
 			}()
 		}
 	}()
-	<-stop
+	stopCh := ctx.Done()
+	<-stopCh
```
abhinavdahiya, line 129:
```suggestion
	<-ctx.Done()
```

pkg/controller/metrics/metrics.go
```diff
@@ -18,6 +17,7 @@ import (
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hiveintv1alpha1 "github.com/openshift/hive/apis/hiveinternal/v1alpha1"
+
```
abhinavdahiya, line 20: super nit: extra line

pkg/controller/metrics/metrics.go
```diff
@@ -314,7 +314,7 @@ func (mc *Calculator) Start(stopCh <-chan struct{}) error {
 		}
 
 		mc.calculateSelectorSyncSetMetrics(mcLog)
-	}, mc.Interval, stopCh)
+	}, mc.Interval, ctx.Done())
```
abhinavdahiya, line 317: nit: replace wait.Until with https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#UntilWithContext

test/e2e/common/machineset.go
```diff
@@ -54,8 +53,9 @@ func WaitForMachineSets(cfg *rest.Config, testFunc func([]*machinev1.MachineSet)
 			DeleteFunc: func(obj interface{}) { onUpdate() },
 		})
 
-	go internalCache.Start(stop)
-	defer func() { stop <- struct{}{} }()
+	ctx, stop := context.WithCancel(context.Background())
+	go internalCache.Start(ctx)
+	defer func() { stop() }()
```
abhinavdahiya, line 58:
```suggestion
	defer stop()
```

test/e2e/common/machine.go
```diff
@@ -55,8 +54,9 @@ func WaitForMachines(cfg *rest.Config, testFunc func([]*machinev1.Machine) bool,
 			DeleteFunc: func(obj interface{}) { onUpdate() },
 		})
 
-	go internalCache.Start(stop)
-	defer func() { stop <- struct{}{} }()
+	ctx, stop := context.WithCancel(context.Background())
+	go internalCache.Start(ctx)
+	defer func() { stop() }()
```
abhinavdahiya, line 59:
```suggestion
	defer stop()
```

test/e2e/common/node.go
```diff
@@ -45,8 +44,9 @@ func WaitForNodes(cfg *rest.Config, testFunc func([]*corev1.Node) bool, timeOut
 			DeleteFunc: func(obj interface{}) { onUpdate() },
 		})
 
-	go internalCache.Start(stop)
-	defer func() { stop <- struct{}{} }()
+	ctx, stop := context.WithCancel(context.Background())
+	go internalCache.Start(ctx)
+	defer func() { stop() }()
```
abhinavdahiya, line 49:
```suggestion
	defer stop()
```

apis/go.mod
abhinavdahiya: hmm.. or maybe rename the PR title as we are bumping to 0.7.0

abhinavdahiya:
I didn't see changes to cover these changes in https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0 @dgoodwin 

1.
```
Change default webhook port to 9443 (https://github.com/kubernetes-sigs/controller-runtime/pull/1076)

Impact: update your deployment configuration to use port 9443, or manually
configure the webhook port in manager Options.
```
I think we have webhooks and making sure the correct port was configured.

2.
```
Remove logs from internal controller (https://github.com/kubernetes-sigs/controller-runtime/pull/1096)

Impact: automatic logs for "succesful reconcile" no longer exist
(it tended to duplicate manual logs that folks had in place).
```
dropping the logs means that we won't get the logs from errors that aren't explicity logged. we tried to capture those here https://github.com/openshift/hive/pull/1263 but i think moving to this version will again drop those error logs.

3.
```
pkg/webhook/admission: upgrade v1beta1 admission types to v1 (https://github.com/kubernetes-sigs/controller-runtime/pull/1284)

Impact: construct v1 Go types instead of v1beta1. Both versions are
supported on the wire.
```
Does this affect Hive running on 3.11 cluster as maybe that version is not supported there??

joelddiaz (commented):
i hope moving to 0.8 involves less changes... :+1:

dgoodwin:
Updated per review. Thanks!

joelddiaz (commented):
/lgtm

dgoodwin:
> I didn't see changes to cover these changes in https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0 @dgoodwin
> 
> 
> ```
> Change default webhook port to 9443 (https://github.com/kubernetes-sigs/controller-runtime/pull/1076)
> 
> Impact: update your deployment configuration to use port 9443, or manually
> configure the webhook port in manager Options.
> ```
> 
> I think we have webhooks and making sure the correct port was configured.


Looks like this is ok, we specify --secure-port 9443 in our Deployment and the Service forwards 443 to 9443. I can see the admission hook getting called if I try to make a bad change.


> 
> 
> ```
> Remove logs from internal controller (https://github.com/kubernetes-sigs/controller-runtime/pull/1096)
> 
> Impact: automatic logs for "succesful reconcile" no longer exist
> (it tended to duplicate manual logs that folks had in place).
> ```
> 
> dropping the logs means that we won't get the logs from errors that aren't explicity logged. we tried to capture those here #1263 but i think moving to this version will again drop those error logs.

Reading the discussion in there, I don't think there's any action for us to take. We don't want to log at V(5), we should be logging our own errors, and it sounds like that is the direction controller-runtime is going. Feels like we'll have to handle any situations where we missed logging as they arise.

> 
> 
> ```
> pkg/webhook/admission: upgrade v1beta1 admission types to v1 (https://github.com/kubernetes-sigs/controller-runtime/pull/1284)
> 
> Impact: construct v1 Go types instead of v1beta1. Both versions are
> supported on the wire.
> ```
> 
> Does this affect Hive running on 3.11 cluster as maybe that version is not supported there??

abhinavdahiya:
> > I didn't see changes to cover these changes in https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0 @dgoodwin
> > ```
> > Change default webhook port to 9443 (https://github.com/kubernetes-sigs/controller-runtime/pull/1076)
> > 
> > Impact: update your deployment configuration to use port 9443, or manually
> > configure the webhook port in manager Options.
> > ```
> > 
> > 
> > I think we have webhooks and making sure the correct port was configured.
> 
> Looks like this is ok, we specify --secure-port 9443 in our Deployment and the Service forwards 443 to 9443. I can see the admission hook getting called if I try to make a bad change.
> 
> > ```
> > Remove logs from internal controller (https://github.com/kubernetes-sigs/controller-runtime/pull/1096)
> > 
> > Impact: automatic logs for "succesful reconcile" no longer exist
> > (it tended to duplicate manual logs that folks had in place).
> > ```
> > 
> > 
> > dropping the logs means that we won't get the logs from errors that aren't explicity logged. we tried to capture those here #1263 but i think moving to this version will again drop those error logs.
> 
> Reading the discussion in there, I don't think there's any action for us to take. We don't want to log at V(5), we should be logging our own errors, 

If Reconcile will not log the errors, we should log any error returned by Reconcile... that does not happen today. And the alternative that you have proposed of logging before returning error when we see a bug/requirement is not sustainable.

logging the error returned by Reconcile is important info that should not be left to each return. If the controller-runtime cannot be our backstop, we should enforce all Reconcile functions look like

```go
func (r *Reconciler) Reconcile(_, _) (reconcile.Result, error) {
    // setup logger

    r, err := r.reconcile(_, _, logger)
    if err != nil {
        logger.WithError(err).Error("failed to reconcile")
    }
    return r, err
}
```

and it sounds like that is the direction controller-runtime is going. Feels like we'll have to handle any situations where we missed logging as they arise.
> 
> EDIT: actually may have been restored: [kubernetes-sigs/controller-runtime#1245](https://github.com/kubernetes-sigs/controller-runtime/pull/1245)
> 
> > ```
> > pkg/webhook/admission: upgrade v1beta1 admission types to v1 (https://github.com/kubernetes-sigs/controller-runtime/pull/1284)
> > 
> > Impact: construct v1 Go types instead of v1beta1. Both versions are
> > supported on the wire.
> > ```
> > 
> > 
> > Does this affect Hive running on 3.11 cluster as maybe that version is not supported there??

dgoodwin:
I have confirmed that we still get reconciler errors logged by controller-runtime by modifying clusterdeployment_controller to just immediately return an error:

```
time="2021-03-31T17:00:40.379Z" level=error msg="Reconciler error" _name=clusterdeployment-controller error="new error!!!!!!!!" name=dgoodwin-gcp namespace=hive
```

Which looks like Alberto's commit restored what we were counting on.

Per slack I'm going to remove the hold, let this roll to stage/int so we can test it on one of the v3 clusters and make sure admission webhooks are working properly.

dgoodwin:
/hold cancel

Updated with a merge.

joelddiaz (commented):
/lgtm

PR #1319, opened 2021-03-29
awsclient: fix creating client from env when no config provided

abhinavdahiya:
the session options always picked default profile for AWS Config. So
when the session was from secret object the AWS Config will be generated
for the default profile. But when the session needed to be created from
environment like in installerpod, the setting of default profile would
skip the environement variables weirdly (this is not how it should be
but tested and it acts like that).

/assign @joelddiaz


joelddiaz (commented):
/lgtm

PR #1320, opened 2021-03-30
docs: add enhancement for unified configuration of hive controllers

abhinavdahiya:
Put together thoughts on how i think we should standardize the flow of configuration from HiveConfig to each controller in Hive.

### DIFF
--- a/docs/enhancements/unified-configuration-of-hive-controllers.md
+++ b/docs/enhancements/unified-configuration-of-hive-controllers.md
@@ -1,0 +1,239 @@
+# Unified Configuration of Hive controllers
+
+## Summary
+
+Each Hive controller loads its configuration from a ConfigDir. The ConfigDir provides
+the controller a YAML serialized HiveConfig and all needed ConfigMaps and Secrets.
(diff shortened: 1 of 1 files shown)


docs/enhancements/unified-configuration-of-hive-controllers.md
```diff
@@ -0,0 +1,239 @@
...
+        <secret-1>/
+            <secret-1-key1>
+            <secret-1-key2>
+        <secret-2>/
+            <secret-2-key1>
+        ...
+    configmaps/
+        <configmap-1>/
+            <configmap-1-key1>
+        ...
```
dgoodwin, line 58: Secrets and ConfigMaps I'm not entirely sure what we gain here, as opposed to just reading what we need on the fly. Restarting the pod can be a pretty expensive operation in a heavily loaded Hive, and reading on the fly makes the local dev case quite a bit simpler, you just need the main hiveconfig file. 

 What benefits do you see to the complexity of setting up the tree of mounts, vs reading the ref of what we need from HiveConfig file on disk and loading that object with a kube client as needed?

PR #1322, opened 2021-03-30
Fix issue with the upload when logs are collected using oc adm must-gather

akhil-rane:
Currently, `oc adm must-gather` stores logs inside `logs/must-gather` dir of the
install pod. While uploading we iterate through `logs` dir and only upload files,
ignoring directories.

This fix creates a tar archive file of the `logs/must-gather` dir in the `logs` dir
so that it does not get ignored while uploading to S3.

x-ref: https://issues.redhat.com/browse/HIVE-1462


gregsheremeta:
looks good. i'll leave it for Abhinav to ack

abhinavdahiya (commented):
/lgtm

PR #1324, opened 2021-03-31
Remove api dependency on controller-runtime.

dgoodwin:
We were relying on controller-runtime only for some very small scheme
registration utilities, which was causing problems for anyone vendoring
our API who happened to be on another version.

Drop the dependency entirely by copying the scheme.go from
controller-runtime over.

/assign @abhinavdahiya 
/cc @abutcher


apis/scheme/scheme.go
```diff
@@ -14,6 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Copied from controller-runtime to avoid a dependency on specific versions
+// causing problems with anyone importing hive's api package.
+
```
abhinavdahiya, line 19: how much value do we get by keeping a copy of scheme builder vs using what non-controller-runtime dependent apis do. like openshift/api https://github.com/openshift/api/blob/master/config/v1/register.go#L11

dgoodwin: I'm not seeing more or less value in either, the code was so tiny and simple it seemed best to just keep doing what we're doing. We could certainly do it the way openshift/api does if that is desirable for any reason.

PR #1327, opened 2021-04-01
Container build improvements

dgoodwin:
Cleanup, formalize and document container development process.
    
Formalizes the very fast containerized development workflow for Fedora
    33+.
    
Remove long deprecated make targets.
    
Update documentation to outline the three paths to run Hive from source.
    
Cleanup the instructions for non-development users to install Hive.

Move main Dockerfile to CentOS 8.

x-ref: https://issues.redhat.com/browse/HIVE-1105
x-ref: https://issues.redhat.com/browse/HIVE-1464

/assign @abhinavdahiya 
/cc @abutcher


Dockerfile
```diff
@@ -4,17 +4,17 @@ WORKDIR /go/src/github.com/openshift/hive
 COPY . .
 RUN make build
 
-FROM quay.io/centos/centos:7
+FROM quay.io/centos/centos:8
```
dgoodwin, line 7: We had a card on backlog to do this requested by ACM.

Makefile
```diff
@@ -31,9 +31,6 @@ LOG_LEVEL ?= debug
 # Image URL to use all building/pushing image targets
 IMG ?= hive-controller:latest
 
-# Image to use when deploying
-DEPLOY_IMAGE ?= registry.ci.openshift.org/openshift/hive-v4.0:hive
```
dgoodwin, line 35: IMG is now used for make deploy as well, but I have removed the default registry CI image. Deploying from source should now be developers only, and should never need to go to the CI image. Normal users should use OLM, otherwise they will need to build their own image.

Makefile
```diff
@@ -236,16 +234,16 @@ docker-build:
...
 
-# Build the dev image using docker
-.PHONY: docker-dev-build
-docker-dev-build: build
-	@echo "*** DEPRECATED: Use the image-hive-dev target instead ***"
-	$(DOCKER_CMD) build -t ${IMG} -f Dockerfile.dev .
+.PHONY: buildah-makecache
+buildah-makecache:
+	mkdir -p /tmp/centos8/dnf
+	podman run --rm -v /tmp/centos8/dnf:/var/cache/dnf:z centos:8 dnf -y makecache
```
dgoodwin, line 240: Removing this, mistaken commit.

Makefile
```diff
@@ -236,16 +234,16 @@ docker-build:
...
 
-# Build the dev image using builah
-.PHONY: buildah-dev-build
-buildah-dev-build:
-	buildah bud -f Dockerfile --tag ${IMG}
+# Build the dev image using buildah
+.PHONY: buildah-build
+buildah-build:
+	# Run buildah build with an overlay
+	buildah bud -v ~/.cache/go-build:/.cache/go-build:O -v /tmp/centos8/dnf:/var/cache/dnf:O -f Dockerfile --tag ${IMG}
```
abhinavdahiya, line 246: you can get go cache location using `go env GOCACHE`

dgoodwin: This was a mistaken commit from attempts to get the go cache mounted in, removed now.

dgoodwin:
Tests are clear.

Dockerfile
joelddiaz, line 7: What are the consequences of this as far as the whole CentOS Stream future is concerned?

dgoodwin: No idea what they're doing with the official centos images, possibly not a lot given that change is mostly about updates.

PR #1330, opened 2021-04-05
config/hiveadmission: set sideEffects for validating webhooks to None

abhinavdahiya:
Since all of our validation webhooks only validate the resource being
sent in with info in the object and any other global configuration,
these pretty much have no side effects.

so setting the side effects to None allows dry-runs.

xref: https://issues.redhat.com/browse/HIVE-1468

/assign @dgoodwin 
/cc @hongkailiu


dgoodwin:
Nice find.

/lgtm

PR #1331, opened 2021-04-07
install: add active deadline seconds on provision/deprovision jobs

abhinavdahiya:
Setting an active deadline seconds for the job limits the time an attempt
can continue to run. This is useful when install/cleanup is stuck for
really long and cannot make progress.

In cases when no progress can be be made it is useful to give up the
attempt as failure and report the failure so that we can reattempt. This
also also reattempt limits to be useful.

/assign @dgoodwin 
/cc @joelddiaz

### DIFF
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -47,6 +48,15 @@ const (
 
 	// LibvirtSSHPrivateKeyDir is the directory where the generated Job will mount the libvirt ssh secret to
 	LibvirtSSHPrivateKeyDir = "/libvirtsshkeys"
+
+	// provisionJobDeadline is the maximum time that provision job will be allowed to run.
+	// when this deadline is reached, the provision attempt will be marked failed.
+	// since provision jobs can include cleanup before attempting installation, this should
+	// include maximum of both these durations to allow for a successfull attempt.
+	provisionJobDeadline = 3 * time.Hour
+	// deprovisionJobDeadline is the maximum time that deprovision job will be allowed to run.
+	// when this deadline is reached, the deprovision attempt will be marked failed.
+	deprovisionJobDeadline = 1 * time.Hour
 )
 
 var (
@@ -575,8 +585,9 @@ func GenerateInstallerJob(provision *hivev1.ClusterProvision) (*batchv1.Job, err
 			Labels:    labels,
 		},
 		Spec: batchv1.JobSpec{
-			BackoffLimit: pointer.Int32Ptr(0),
-			Completions:  pointer.Int32Ptr(1),
+			BackoffLimit:          pointer.Int32Ptr(0),
+			Completions:           pointer.Int32Ptr(1),
+			ActiveDeadlineSeconds: pointer.Int64Ptr(int64(provisionJobDeadline.Seconds())),
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: labels,
@@ -626,8 +637,9 @@ func GenerateUninstallerJobForDeprovision(
 	job.ObjectMeta.Labels = labels
 	job.ObjectMeta.Annotations = map[string]string{}
 	job.Spec = batchv1.JobSpec{
-		Completions:  &completions,
-		BackoffLimit: &backoffLimit,
+		Completions:           &completions,
+		BackoffLimit:          &backoffLimit,
+		ActiveDeadlineSeconds: pointer.Int64Ptr(int64(deprovisionJobDeadline.Seconds())),
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: labels,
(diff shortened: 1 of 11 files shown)


pkg/install/generate.go
```diff
@@ -575,8 +585,9 @@ func GenerateInstallerJob(provision *hivev1.ClusterProvision) (*batchv1.Job, err
 			Labels:    labels,
 		},
 		Spec: batchv1.JobSpec{
-			BackoffLimit: pointer.Int32Ptr(0),
-			Completions:  pointer.Int32Ptr(1),
+			BackoffLimit:          pointer.Int32Ptr(0),
+			Completions:           pointer.Int32Ptr(1),
+			ActiveDeadlineSeconds: pointer.Int64Ptr(int64(provisionJobDeadline.Seconds())),
```
dgoodwin, line 590: When this procs the the ClusterProvision shows the following:

```yaml
  - lastProbeTime: "2021-04-08T17:27:33Z"
    lastTransitionTime: "2021-04-08T17:27:33Z"
    message: Cluster install failed but installer log was not captured
    reason: UnknownError
    status: "True"
    type: ClusterProvisionFailed
```

Would it be possible to get some detail in there to clarify it hit the deadline. The linked job does have this detail available and we could copy it back. 

```yaml
status:
  conditions:
  - lastProbeTime: "2021-04-08T17:27:33Z"
    lastTransitionTime: "2021-04-08T17:27:33Z"
    message: Job was active longer than specified deadline
    reason: DeadlineExceeded
    status: "True"
    type: Failed
```

pkg/install/generate.go
```diff
@@ -626,8 +637,9 @@ func GenerateUninstallerJobForDeprovision(
 	job.ObjectMeta.Labels = labels
 	job.ObjectMeta.Annotations = map[string]string{}
 	job.Spec = batchv1.JobSpec{
-		Completions:  &completions,
-		BackoffLimit: &backoffLimit,
+		Completions:           &completions,
+		BackoffLimit:          &backoffLimit,
+		ActiveDeadlineSeconds: pointer.Int64Ptr(int64(deprovisionJobDeadline.Seconds())),
```
dgoodwin, line 642: When this one procs I get no status on my ClusterDeprovision at all, these actually seem woefully uninformative. To retry I was able to delete the Job and it relaunched. Deleting ClusterDeprovision did not help. At the least I think we need a condition that gets set to inform user that it timed out. (presumably we'd document how to re-try and not include in the condition message?)

abhinavdahiya: The deprovision object now has a condition that provides status of failed jobs.

when the deadline is exceeded.
```
oc get clusterdeprovision test-cluster -ojson | jq .status.conditions
[
  {
    "lastProbeTime": "2021-04-13T21:46:43Z",
    "lastTransitionTime": "2021-04-13T21:46:43Z",
    "message": "Deprovision attempt failed because the deadline was exceeded",
    "reason": "AttemptDeadlineExceeded",
    "status": "True",
    "type": "DeprovisionFailed"
  }
]

```

pkg/install/generate.go
abhinavdahiya: the provision also updates the reason for deadline exceeded.

```
$ oc get clusterprovision test-cluster-0-9gj4q -ojson | jq .status.conditions
[
  {
    "lastProbeTime": "2021-04-13T21:40:45Z",
    "lastTransitionTime": "2021-04-13T21:40:45Z",
    "message": "Install job has been created",
    "reason": "JobCreated",
    "status": "True",
    "type": "ClusterProvisionJobCreated"
  },
  {
    "lastProbeTime": "2021-04-13T21:41:11Z",
    "lastTransitionTime": "2021-04-13T21:41:11Z",
    "message": "Install job has completed its initialization. Provisioning started.",
    "reason": "InitializationComplete",
    "status": "True",
    "type": "ClusterProvisionInitialized"
  },
  {
    "lastProbeTime": "2021-04-13T21:43:45Z",
    "lastTransitionTime": "2021-04-13T21:43:45Z",
    "message": "Install job failed due to deadline being exceeded for the attempt",
    "reason": "AttemptDeadlineExceeded",
    "status": "True",
    "type": "ClusterProvisionFailed"
  }
]

$ oc get cd test-cluster -ojson | jq .status.conditions
[
  {
    "lastProbeTime": "2021-04-13T21:40:37Z",
    "lastTransitionTime": "2021-04-13T21:40:37Z",
    "message": "ClusterImageSet test-cluster-imageset is available",
    "reason": "ClusterImageSetFound",
    "status": "False",
    "type": "ClusterImageSetNotFound"
  },
  {
    "lastProbeTime": "2021-04-13T21:43:45Z",
    "lastTransitionTime": "2021-04-13T21:43:45Z",
    "message": "Provision test-cluster-0-9gj4q failed. Next provision at 2021-04-13T21:44:45Z.",
    "reason": "AttemptDeadlineExceeded",
    "status": "True",
    "type": "ProvisionFailed"
  }
]

```

abhinavdahiya:
@dgoodwin updated to include sync of deadline canceled jobs to conditions on appropriate object.

pkg/install/generate.go
dgoodwin: How does one restart a job in this state? Previously we assumed it ran forever. Ideally I'd like to see that noted in condition message to help people out.

abhinavdahiya: the deprovision controller still restarts the job without user intervention like before. Since we don't have a concept of attempt limits for deprovision we end up with the same behavior of retrying forever, one deadline at a time.

dgoodwin: Ah cool I did not realize.

PR #1333, opened 2021-04-08
Check for no applyMode set (defaults to Upsert) when clearing resourcesToDelete

abutcher:
When `SyncSet` `resourceApplyMode` is changed from `Sync` to `Upsert`, we remove the `resourcesToDelete` from `ClusterSync` status. Since `resourceApplyMode` defaults to `Upsert` we should check for no `resourceApplyMode` when clearing the `resourcesToDelete`.

/assign @dgoodwin

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -568,10 +568,12 @@ func (r *ReconcileClusterSync) applySyncSets(
 			ObservedGeneration: syncSet.AsMetaObject().GetGeneration(),
 			Result:             hiveintv1alpha1.SuccessSyncSetResult,
 		}
-		if syncSet.GetSpec().ResourceApplyMode == hivev1.SyncResourceApplyMode {
+		applyMode := syncSet.GetSpec().ResourceApplyMode
+		if applyMode == hivev1.SyncResourceApplyMode {
 			newSyncStatus.ResourcesToDelete = resourcesApplied
 		}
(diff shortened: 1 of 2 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -568,10 +568,12 @@ func (r *ReconcileClusterSync) applySyncSets(
...
 			Result:             hiveintv1alpha1.SuccessSyncSetResult,
 		}
-		if syncSet.GetSpec().ResourceApplyMode == hivev1.SyncResourceApplyMode {
+		applyMode := syncSet.GetSpec().ResourceApplyMode
+		if applyMode == hivev1.SyncResourceApplyMode {
 			newSyncStatus.ResourcesToDelete = resourcesApplied
 		}
-		if syncSet.GetSpec().ResourceApplyMode == hivev1.UpsertResourceApplyMode && len(oldSyncStatus.ResourcesToDelete) > 0 {
+		// applyMode defaults to UpsertResourceApplyMode
+		if applyMode == hivev1.UpsertResourceApplyMode || applyMode == "" && len(oldSyncStatus.ResourcesToDelete) > 0 {
```
dgoodwin, line 576: This might be working as is, but I'm a bit thrown by the || and then && and how things get evaluated.  Testing real quick on the go playground, true || false && false seems to either evaluate the && first, or start on the right side.

Anyhow could you group with braces, it looks to me like it should be ( a || b) && c in this case?

PR #1336, opened 2021-04-09
Transfer proxy env vars from hive-operator pod to all hive/provision …

dgoodwin:
…pods.

These env vars should be set by OLM if a cluster is configured with a
global proxy. This change ensures they are passed through to all pods
the operator creates, and all pods hive-controllers creates.

x-ref: https://issues.redhat.com/browse/HIVE-1355

/assign @abhinavdahiya 
/cc @abutcher

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -781,6 +781,9 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		provisionName,
 		releaseImage,
 		controllerutils.InstallServiceAccountName,
+		os.Getenv("HTTP_PROXY"),
+		os.Getenv("HTTPS_PROXY"),
+		os.Getenv("NO_PROXY"),
 		extraEnvVars,
 	)
 	if err != nil {
@@ -1212,7 +1215,10 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 			return nil, r.setInstallImagesNotResolvedCondition(cd, corev1.ConditionFalse, imagesResolvedReason, imagesResolvedMsg, cdLog)
 		}
 
-		job := imageset.GenerateImageSetJob(cd, releaseImage, controllerutils.InstallServiceAccountName)
+		job := imageset.GenerateImageSetJob(cd, releaseImage, controllerutils.InstallServiceAccountName,
+			os.Getenv("HTTP_PROXY"),
+			os.Getenv("HTTPS_PROXY"),
+			os.Getenv("NO_PROXY"))
 
 		cdLog.WithField("derivedObject", job.Name).Debug("Setting labels on derived object")
 		job.Labels = k8slabels.AddLabel(job.Labels, constants.ClusterDeploymentNameLabel, cd.Name)
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -351,16 +351,39 @@ func BuildControllerLogger(controller hivev1.ControllerName, resource string, ns
 	})
 }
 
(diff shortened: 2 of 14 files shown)


dgoodwin:
Missed ImageSet jobs, adding now.

dgoodwin:
Added some tests and refactored how and where we read env vars. Ready for review @abhinavdahiya

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -278,6 +281,8 @@ type ReconcileClusterDeployment struct {
 	validateCredentialsForClusterDeployment func(client.Client, *hivev1.ClusterDeployment, log.FieldLogger) (bool, error)
 
 	protectedDelete bool
+
+	httpProxy, httpsProxy, noProxy string
```
abhinavdahiya, line 285: not a big fan of storing the proxy information in the reconciler. Go binaries automatically support reading these env vars and using them for http client requests, so all reconciler calls will be covered.
and only thing we need to do it copy these env vars to other pods we create.

So i would like to see that, us copying env vars to pods we created and not using reconciler fields.

pkg/controller/utils/utils.go
```diff
@@ -350,3 +350,38 @@ func BuildControllerLogger(controller hivev1.ControllerName, resource string, ns
 		"reconcileID": utilrand.String(constants.ReconcileIDLen),
 	})
 }
+
+// SetProxyEnvVars will add the standard proxy environment variables to all containers in the given pod spec. If
+// any of the provided values are empty, the environment variable will not be set.
+func SetProxyEnvVars(podSpec *corev1.PodSpec, httpProxy, httpsProxy, noProxy string) {
+	setEnvVarOnContainers := func(podSpec *corev1.PodSpec, envVar, value string) {
+		if value != "" {
```
abhinavdahiya, line 358: return early

pkg/controller/utils/utils.go
```diff
@@ -350,3 +350,38 @@ func BuildControllerLogger(controller hivev1.ControllerName, resource string, ns
...
+			for i := range podSpec.Containers {
+				// Check if the env var is already set, if so we preserve the original but warn if they differ:
```
abhinavdahiya, line 360: this sounds off from the goal of the SetProxyEnvVars .. i would assume we force our value.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: This was for testability reasons and we'd talked in the past about avoiding reading env vars directly in controllers, but I could live with the controllers doing Getenv, skip testing at that level, and pass the strings into SetProxyEnvVars which can still be unit tested cleanly. I'll make the change.

dgoodwin:
Updates pushed.

pkg/operator/hive/hive_controller.go
```diff
@@ -69,7 +69,15 @@ func Add(mgr manager.Manager) error {
...
-	return &ReconcileHiveConfig{Client: mgr.GetClient(), scheme: mgr.GetScheme(), restConfig: mgr.GetConfig(), mgr: mgr}
+	return &ReconcileHiveConfig{
+		Client:     mgr.GetClient(),
+		scheme:     mgr.GetScheme(),
+		restConfig: mgr.GetConfig(),
+		mgr:        mgr,
+		httpProxy:  os.Getenv("HTTP_PROXY"),
+		httpsProxy: os.Getenv("HTTPS_PROXY"),
+		noProxy:    os.Getenv("NO_PROXY"),
+	}
```
abhinavdahiya, line 80: i think this missed the reconciler to callsite change

dgoodwin: My bad! Forgot that area entirely.

PR #1339, opened 2021-04-12
Add S3 Buckets Limit exceeded to install log regex

suhanime:
x-ref: https://issues.redhat.com/browse/HIVE-1411


dgoodwin:
x-ref: https://issues.redhat.com/browse/HIVE-1411

/lgtm

PR #1342, opened 2021-04-13
WIP: awsprivatelink: retry condition update on clusterdeployments

abhinavdahiya:
since the controller tries to update the conditions when
clusterdeployment and clusterprovision controller might be adding
conditions, retying on conflict should be useful to prevent un necessary
reconcile retries.

/assign @dgoodwin

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -340,7 +340,7 @@ func (r *ReconcileAWSPrivateLink) setErrCondition(cd *hivev1.ClusterDeployment,
 	}
 	curr.Status.Conditions = conditions
 	logger.Debug("setting AWSPrivateLinkFailedClusterDeploymentCondition to true")
-	return r.Status().Update(context.TODO(), curr)
+	return r.updateClusterDeploymentConditions(curr, logger)
 }
 
 func (r *ReconcileAWSPrivateLink) setProgressCondition(cd *hivev1.ClusterDeployment,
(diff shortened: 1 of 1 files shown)


pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -1241,6 +1241,19 @@ func (r *ReconcileAWSPrivateLink) updatePrivateLinkStatus(cd *hivev1.ClusterDepl
...
+		curr := &hivev1.ClusterDeployment{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, curr)
+		if err != nil {
+			return err
+		}
+
+		curr.Status.Conditions = cd.Status.Conditions
+		return r.Client.Status().Update(context.TODO(), curr)
+	})
+}
```
dgoodwin, line 1255: Throughout hive I believe we always just return the error and let reconcile run again, is there any reason not to do the same here?

abhinavdahiya: i think it is common for controllers to use the retry on conflict when updating status. i personally would recommend every status update use it as it prevents noise in reconcile errors.

also the privatelinkcontroller reconcile has a little higher cost and error and reconcile again wouldn't be the best.

dgoodwin: That's reasonable, I agree we should try to do this more then.

cc @openshift/openshift-team-hive

staebler: Won't this cause the controller to lose any changes made to conditions by other controllers?
1. Private Link controller wants to change condition A.
2. Some other controller changes condition B.
3. Private Link controller gets a conflict. Private Link controller overwrites all of the conditions, losing the changes to condition B.

dgoodwin: Yeah looks like you're right, need to surgically re-add the condition after fetching latest state.

abhinavdahiya: Ah, yes 🤦🏽

dgoodwin:
bump @abhinavdahiya

PR #1348, opened 2021-04-14
Re-add buildah build target for dev

suhanime:
Give the option to build hive with buildah

/assign @dgoodwin


suhanime:
@dgoodwin I want to edit the [docs](https://github.com/openshift/hive/blob/master/docs/developing.md#full-container-build) as well, but I think image-hive relies on docker

PR #1352, opened 2021-04-15
Fix docker-dev-push target

dastergon:
Running `IMG=localhost:5000/hive:latest make docker-dev-push` fails with 

      make: *** No rule to make target `image-hive-dev', needed by `docker-dev-push'.  Stop.

Looking at the [docs](https://github.com/openshift/hive/blob/master/docs/developing.md#deploying-from-source) and the [deprecation messages](https://github.com/openshift/hive/blob/master/Makefile#L233) in the Makefile, I guess the sub-target should be `image-hive-fedora-dev` rather than `image-hive-dev`. Consequently I sent this PR to update the target to `image-hive-fedora-dev`.

PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ docker-push:
 
 # Build and push the dev image
 .PHONY: docker-dev-push
-docker-dev-push: build image-hive-dev docker-push
+docker-dev-push: build image-hive-fedora-dev docker-push
 
 # Build the dev image using builah
 .PHONY: buildah-dev-build


Makefile
```diff
@@ -240,7 +240,7 @@ docker-push:
 
 # Build and push the dev image
 .PHONY: docker-dev-push
-docker-dev-push: build image-hive-dev docker-push
+docker-dev-push: build image-hive docker-push
```
dgoodwin, line 243:
```suggestion
docker-dev-push: build image-hive-fedora-dev docker-push
```

My mistake, thanks for catching. It's actually supposed to be the above.

My workflow now is:

```
$ time make build image-hive-fedora-dev docker-push && kubectl delete pod -n hive --all
```

dastergon: What's the difference between `image-hive` and `image-hive-fedora-dev`? I am a macOS user, so I want to make sure this works on both Linux and macOS.

dgoodwin: image-hive is a slow, but full build of a fresh container. fedora-dev is using the locally build binaries and go cache, and injecting those binaries into a pre-made base layer for faster development. 

on macos, i think you want to avoid all the fedora/dev targets and just use make image-hive. (build not required)

dastergon: Thanks, @dgoodwin. Updated the line and rebased. PTAL

dgoodwin:
/lgtm

Thanks!

PR #1356, opened 2021-04-16
Add proposal for breaking out install strategy to separate CRDs.

dgoodwin:
cc @abhinavdahiya @joelddiaz @mhrivnak @avishayt @filanov @romfreiman @jnpacker @csrwng @derekwaynecarr

### DIFF
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -85,24 +85,26 @@ The following ClusterDeployment fields are not carrying over to this interface a
   * ReleaseImage: Legacy Hive feature, use ClusterImageSet instead.
   * ManifestsConfigMapRef: Too specific to openshift-install.
 
-All conditions should initialize to Unknown status as soon as possible, and then be set to True/False once known.
+All conditions should initialize to Unknown status as soon as possible per latest Kubernetes API guidelines, and then be set to True/False once more explicit status is known.
 
 Condition | Description
 --------- | -----------
-Ready | Important overall condition indicating if the cluster is ready or not. Reason and Message provide as much detail as possible for the user. Message could contain attempt x of y if the ClusterInstall support retries.
-RequirementsMet | True once we are ready to launch the install. False if we hit a problem and can't start. Reasons: ClusterImageSetNotFound, etc.
-Failed | True once an install attempt has failed. Remains true if retries are in play even if another attempt is in progress. Move to False once Ready goes to True. Reason/Message should contain details on what happened if it's possible to detect, otherwise Unknown. *TODO*: Unfortunate naming here, this matches current ProvisionFailed condition, which means a failure has been encountered, not necessarily that we've given up. Failing would be a better term, but it's risky to change the ProvisionFailed condition meaning on ClusterDeployment when we transfer. Thoughts?
-Retrying | Unknown initially, True if we've hit an error and are retrying, False if we've given up or don't support retries. Message should indicate attempt x of y.
+RequirementsMet | True once all pre-install requirements have been met. False if we hit a problem and can't start. Reasons: ClusterImageSetNotFound, ClusterDeploymentNotFound, etc.
+Completed | True once the cluster install has successfully completed. Reason and Message provide as much detail as possible for the user as install progresses. Message could contain attempt x of y if the ClusterInstall support retries.
+Failed | True once an install attempt has failed. Does not imply we've stopped trying, if retries are supported by the controller we may be re-trying. Returns to False if we later reach Completed=True.
+Stopped | True once the controller is no longer working to reach desired state. Does not imply success or failure by itself, but can be combined with Completed/Failed.
 
-These interface conditions will be copied back onto ClusterDeployment by Hive controllers with a "Provision" prefix, the ClusterInstall controller does not need to worry about this and in general should never write to ClusterDeployment. example: Failed becomes ProvisionFailed (which exists today on ClusterDeployment for this purpose)
+These interface conditions will be copied back onto ClusterDeployment by Hive controllers with a "ClusterInstall" prefix, the ClusterInstall controller does not need to worry about this and in general should never write to ClusterDeployment. example: Failed becomes ClusterInstallFailed (which exists today on ClusterDeployment for this purpose).
+
+In the case of ClusterInstallFailed, we will also copy this condition to ProvisionFailed to maintain the current contract and not affect callers monitoring for this condition.
 
 ClusterInstall controllers can add their own conditions, but these will not transfer back to the ClusterDeployment. UIs around this process may need to show ClusterInstall conditions explicitly.
 
 The core ClusterInstall CRDs and controllers discussed in this document should live in Hive. However, in theory, it would be possible for an external application to implement it's own ClusterInstall outside of Hive and still interface with ClusterDeployment using this system.
 
-#### ClusterInstall Implementation RBAC
+#### External ClusterInstall RBAC
 
-If we wish to support an external application providing their own ClusterInstall CRD implementation, Hive will need RBAC to be able to access those objects. This can be achieved by using [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).
+To support an external application providing their own ClusterInstall CRD implementation, Hive will need RBAC to be able to access those objects. This can be achieved by using [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).
 
 Hive will be updated to deploy an empty ClusterRole `hive-controllers-clusterinstall` which will contain no rules, but will match the label `hive.openshift.io/aggregate-to-hive-clusterinstall`. Integrating components will need to deploy a ClusterRole with this label and rules that provide Hive all permissions on their custom ClusterInstall CRD.
 
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -114,17 +114,9 @@ This will not be required for ClusterInstall implementations that live within Hi
 
 Today in Hive each ClusterDeployment is given a finalizer automatically, and when the API resource is deleted we create a ClusterDeprovision resource, which launches a pod to run the openshift-install deprovision process until it completes, after which the finalizer is removed and the ClusterDeployment is removed.
 
-There is presently no deprovision implementation for agent based or bare metal clusters.
+There is presently no deprovision implementation for agent based or bare metal clusters but there likely will be in the future.
 
-It stands to reason that deprovision implementations are fully tied to ClusterInstall implementations.
-
-Option 1: Deleting ClusterInstall automatically deprovisions cluster. Clean in that we know exactly what implementation to use. Dangerous in that users might think it should be safe to delete a ClusterInstall to free up space in etcd after the cluster is provisioned, inadvertently destroying their cluster.
-
-Option 2: Deleting ClusterDeployment automatically deprovisions the cluster as we do today. (how do we know what implementation to call?
-
-Option 3: Caller must specify a ClusterDeprovision themselves. Discussed many times for safety reasons, but has never been popular with team due to unusual Kube behavior and a desire to instad mitigate risks of accidental deletion in other ways.
-
-
+It stands to reason that deprovision implementations are fully tied to ClusterInstall implementations, and thus deleting the ClusterInstall should trigger the controllers for that ClusterInstall to complete an uninstall/deprovision. *WARNING: Deleting a ClusterInstall will destroy the cluster.*
 
 
 ### Migration Plan
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Break install strategy out of ClusterDeployment.Spec and allow for multiple ClusterInstall CRD implementations that adhere to an interface, where Hive interacts with these CRDs dynamically via duck typing in the non-install related controllers.
+Break out ClusterDeployment.Spec.Provisioning.InstallStrategy and allow for multiple ClusterInstall CRD implementations that adhere to an interface, where Hive interacts with these CRDs dynamically via duck typing in the non-install related controllers.
 
 
 ## Motivation
@@ -98,6 +98,14 @@ These interface conditions will be copied back onto ClusterDeployment by Hive co
 ClusterInstall controllers can add their own conditions, but these will not transfer back to the ClusterDeployment. UIs around this process may need to show ClusterInstall conditions explicitly.
 
 The core ClusterInstall CRDs and controllers discussed in this document should live in Hive. However, in theory, it would be possible for an external application to implement it's own ClusterInstall outside of Hive and still interface with ClusterDeployment using this system.
+
+#### ClusterInstall Implementation RBAC
+
+If we wish to support an external application providing their own ClusterInstall CRD implementation, Hive will need RBAC to be able to access those objects. This can be achieved by using [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).
+
+Hive will be updated to deploy an empty ClusterRole `hive-controllers-clusterinstall` which will contain no rules, but will match the label `hive.openshift.io/aggregate-to-hive-clusterinstall`. Integrating components will need to deploy a ClusterRole with this label and rules that provide Hive all permissions on their custom ClusterInstall CRD.
+
+This will not be required for ClusterInstall implementations that live within Hive. (including AgentClusterInstall)
 
 ### Cluster Deprovisioning
 
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -104,11 +104,7 @@ The core ClusterInstall CRDs and controllers discussed in this document should l
 
 #### External ClusterInstall RBAC
 
-To support an external application providing their own ClusterInstall CRD implementation, Hive will need RBAC to be able to access those objects. This can be achieved by using [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).
-
-Hive will be updated to deploy an empty ClusterRole `hive-controllers-clusterinstall` which will contain no rules, but will match the label `hive.openshift.io/aggregate-to-hive-clusterinstall`. Integrating components will need to deploy a ClusterRole with this label and rules that provide Hive all permissions on their custom ClusterInstall CRD.
-
-This will not be required for ClusterInstall implementations that live within Hive. (including AgentClusterInstall)
+To support an external application providing their own ClusterInstall CRD implementation, Hive will need RBAC to be able to access those objects. Hive will ship with a role granting full access to the apigroup "extensions.hive.openshift.io". CRDs which implement the interface should use this apigroup regardless if they are in Hive or external.
 
 ### Cluster Deprovisioning
 
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -73,7 +73,6 @@ spec.clusterMetadata.adminKubeconfigSecretRef | LocalObjectReference | Admin kub
 spec.clusterMetadata.adminPasswordSecretRef | LocalObjectReference | Admin password secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
 spec.clusterMetadata.clusterID | string | Cluster UUID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
 spec.clusterMetadata.infraID | string | Cluster infra ID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
-spec.manifestsConfigMapRef | LocalObjectReference | ConfigMap containing manifests to inject into the cluster during installation. *TODO*: Hive technically doesn't need this to be there, do we require it to ensure consistent support or leave it up to install strategies?
 
 The following ClusterDeployment fields are not carrying over to this interface as they are not generic enough to assume for all ClusterInstall implementations. They can of course be used by specific stratgies, but Hive will not need to assume they are there.
 
@@ -82,6 +81,7 @@ The following ClusterDeployment fields are not carrying over to this interface a
   * InstallEnv: Too specific to openshift-install.
   * InstallConfigSecretRef: One of the more contentious parts of the current API.
   * ReleaseImage: Legacy Hive feature, use ClusterImageSet instead.
+  * ManifestsConfigMapRef: Too specific to openshift-install.
 
 All conditions should initialize to Unknown status as soon as possible, and then be set to True/False once known.
 
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -73,6 +73,8 @@ spec.clusterMetadata.adminKubeconfigSecretRef | LocalObjectReference | Admin kub
 spec.clusterMetadata.adminPasswordSecretRef | LocalObjectReference | Admin password secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
 spec.clusterMetadata.clusterID | string | Cluster UUID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
 spec.clusterMetadata.infraID | string | Cluster infra ID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
+status.installRestarts | int | Current number of install retries.
+status.conditions | []ClusterInstallCondition | Standard conditions array. See below for required conditions.
 
 The following ClusterDeployment fields are not carrying over to this interface as they are not generic enough to assume for all ClusterInstall implementations. They can of course be used by specific stratgies, but Hive will not need to assume they are there.
 
--- a/docs/enhancements/cluster-install-apis.md
+++ b/docs/enhancements/cluster-install-apis.md
@@ -69,7 +69,6 @@ Field | Type | Description
 ----- | ---- | -----------
 spec.clusterDeploymentRef | LocalObjectReference | Link to the required ClusterDeployment for this install. Bidirectional reference. Common data should live on ClusterDeployment and never be duplicated on ClusterInstall.
 spec.imageSetRef | ClusterImageSetReference | Core mechanism in Hive for choosing what version of OpenShift to install.
-spec.stage | string | Enum string for stage of installation. (Initializing, RequirementsNotMet, Provisioning, RetryProvisioning, Complete, Failed) More details in conditions. *TODO*: Should we just work with conditions? Loss of state in disaster recovery?
 spec.clusterMetadata.adminKubeconfigSecretRef | LocalObjectReference | Admin kubeconfig secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
 spec.clusterMetadata.adminPasswordSecretRef | LocalObjectReference | Admin password secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
 spec.clusterMetadata.clusterID | string | Cluster UUID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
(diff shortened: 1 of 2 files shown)


docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+ClusterDeployment.Spec.Provisioning remains but will eventually be deprecated. Validate this is not used in conjunction with ClusterDeployment.Spec.InstallStrategyRef, must choose one path or the other.
+
+### InstallStrategy Interface
+
+
+Field | Type | Description
+----- | ---- | -----------
+spec.clusterDeploymentRef | LocalObjectReference | Link to the required ClusterDeployment for this install. Bidirectional reference. Common data should live on ClusterDeployment and never be duplicated on InstallStrategy.
+spec.imageSetRef | ClusterImageSetReference | Core mechanism in Hive for choosing what version of OpenShift to install.
+spec.stage | string | Enum string for stage of installation. (Initializing, RequirementsNotMet, Provisioning, RetryProvisioning, Complete, Failed) More details in conditions. *TODO*: Should we just work with conditions? Loss of state in disaster recovery?
```
joelddiaz, line 70: +1 for conditions

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+Retrying | Unknown initially, True if we've hit an error and are retrying, False if we've given up or don't support retries. Message should indicate attempt x of y.
+
+These interface conditions will be copied back onto ClusterDeployment by Hive controllers with a "Provision" prefix, the install strategy controller does not need to worry about this and in general should never write to ClusterDeployment. example: Failed becomes ProvisionFailed (which exists today on ClusterDeployment for this purpose)
+
+Install strategy controllers can add their own conditions, but these will not transfer back to the ClusterDeployment. UIs around this process may need to show install strategy conditions explicitly.
+
+The core install strategy CRDs and controllers discussed in this document should live in Hive. However, in theory, it would be possible for an external application to implement it's own install strategy outside of Hive and still interface with ClusterDeployment using this system.
```
joelddiaz, line 98: +100

joelddiaz (commented):
I'm not seeing where the duck typing is being used. Can you clarify, or add an example?

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+      reason: AgentReason
+      status: "True"
+      type: AgentSpecificCondition
+```
+
+
+### Risks and Mitigations
+
+  * Can Hive controllers watch a dynamic set of types to reconcile when an install strategy changes? Can we watch unstructured?
+  * Is InstallStrategy the naming we're happy with? Consider ClusterProvision off the table given it's use already for another purpose.
```
abhinavdahiya, line 186: i would like the contract to be called ClusterInstall , installstrategy doesn't should like a good name for this interface

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+Hive controllers would then interact with this strategy dynamically with duck typing, similar to Cluster API, Knative, etc.
+
+ClusterDeployment.Spec.Provisioning remains but will eventually be deprecated. Validate this is not used in conjunction with ClusterDeployment.Spec.InstallStrategyRef, must choose one path or the other.
+
+### InstallStrategy Interface
+
+
+Field | Type | Description
+----- | ---- | -----------
+spec.clusterDeploymentRef | LocalObjectReference | Link to the required ClusterDeployment for this install. Bidirectional reference. Common data should live on ClusterDeployment and never be duplicated on InstallStrategy.
```
abhinavdahiya, line 68: so if i understand the purpose of  this field in the ClusterInstall is to support fetching information that is common?

can provide examples of the class of information that would be on ClusterDeployment that can be use by these ClusterInstall implementations.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+spec.clusterMetadata.adminKubeconfigSecretRef | LocalObjectReference | Admin kubeconfig secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
+spec.clusterMetadata.adminPasswordSecretRef | LocalObjectReference | Admin password secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
+spec.clusterMetadata.clusterID | string | Cluster UUID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
+spec.clusterMetadata.infraID | string | Cluster infra ID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
+spec.manifestsConfigMapRef | LocalObjectReference | ConfigMap containing manifests to inject into the cluster during installation. *TODO*: Hive technically doesn't need this to be there, do we require it to ensure consistent support or leave it up to install strategies?
```
abhinavdahiya, line 75: this seems like leak from openshift-install's implementation of the ClusterInstall . So i would be in favor of not requiring this in the contract.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+
+ClusterDeployment.Spec.Provisioning remains but will eventually be deprecated. Validate this is not used in conjunction with ClusterDeployment.Spec.InstallStrategyRef, must choose one path or the other.
+
+### InstallStrategy Interface
+
+
+Field | Type | Description
+----- | ---- | -----------
+spec.clusterDeploymentRef | LocalObjectReference | Link to the required ClusterDeployment for this install. Bidirectional reference. Common data should live on ClusterDeployment and never be duplicated on InstallStrategy.
+spec.imageSetRef | ClusterImageSetReference | Core mechanism in Hive for choosing what version of OpenShift to install.
```
abhinavdahiya, line 69: can we include some background on how this would be used by ClusterDeployment ? we can easily use the cluster api to figure out the various versions. this again seems no suitable for the contract.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+Hive is presently struggling with how to design the API to meet the needs of provisioning standalone, agent based, and hypershift clusters. The current code base evolved quite specific to running openshift-install pods and has some growing pains to accomodate new paths to provision clusters. Changes for one raise a lot of questions and inconsistencies for the others as everything is lumped into ClusterDeployment. The line between day 1 and day 2 is very unclear and mostly just enforced via increasingly complicated validating webhooks.
+
+### Goals
+
+- Clarify Hive APIs. (day 1 vs day 2)
+- Provide space for implementations to define the spec and status they need without causing uncomfortable complications for other implementations.
+- Breakout AgentInstallStrategy to a separate CRD. This is still guarded by an alpha feature gate and thus we should still be able to move it around.
+- Do not break API compatability for existing standalone OpenShift cluster provisioning. We hope to slowly migrate those installs to this system by offering it in parallel and eventually deprecating the old, but for now the focus is on breaking out the AgentInstallStrategy.
+
+### Non-Goals
```
abhinavdahiya, line 19: how do we implement cluster pools for these new ClusterInstall implementations? Would that be a separate discussion? Or will some of the decisions we make here be influenced by cluster pool implementation...

abhinavdahiya: Also can we include details on the deprovision of the clusters. I would be in favor of implementing the deprovision as the delete of the object that implements the ClusterInstall instead of a new different object.

we shuoldn't need to support a deprovision that was not provisioned by our APIs.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+
+Initial goal would be to breakout ClusterDeployment.Spec.InstallStrategy.Agent. The only implementation here is agent based and this is hidden behind an alpha feature gate, and not yet released to customers. Replace this with a new hive.openshift.io.AgentInstallStrategy CRD, and a ClusterDeployment.Spec.InstallStrategyRef link. Refactor the Agent controllers (currently living in the assisted service repo, coming to hive in future) to watch this resource instead of ClusterDeployment.
+
+In the future, we could migrate the current standalone openshift-install implementation to an OpenShiftInstallStrategy, which continues to use ClusterProvision for it's retry mechanism. To do this we would allow a parallel path to specify where you cannot use both installStrategyRef and provisioning sections together. Provisioning then becomes deprecated and after some period of time, could be removed.
+
+We could add an additional strategy which offers InstallConfig generation, and could be used by users who are ok with a simpler interface and don't want to specify an install config secret.
+
+HypershiftInstallStrategy would be another likely future implementation.
```
abhinavdahiya, line 110: this shuold be part of the user stories.

docs/enhancements/install-strategy.md
mhrivnak: I wonder if an owner reference would suffice.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+
+Install strategy controllers can add their own conditions, but these will not transfer back to the ClusterDeployment. UIs around this process may need to show install strategy conditions explicitly.
+
+The core install strategy CRDs and controllers discussed in this document should live in Hive. However, in theory, it would be possible for an external application to implement it's own install strategy outside of Hive and still interface with ClusterDeployment using this system.
+
+### Migration Plan
+
+No breakage for current installation API is planned. (with exception of as yet unreleased Agent install strategy)
+
+Initial goal would be to breakout ClusterDeployment.Spec.InstallStrategy.Agent. The only implementation here is agent based and this is hidden behind an alpha feature gate, and not yet released to customers. Replace this with a new hive.openshift.io.AgentInstallStrategy CRD, and a ClusterDeployment.Spec.InstallStrategyRef link. Refactor the Agent controllers (currently living in the assisted service repo, coming to hive in future) to watch this resource instead of ClusterDeployment.
```
mhrivnak, line 104: Just stating the obvious: these changes would need to happen real soon for inclusion in the first release to customers of the agent-based install tooling.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+  networking:
+    machineNetwork: ...
+  sshPublicKey: ...
+  provisionRequirements:
+    controlPlaneAgents: 3
+    workerAgents: 5
+  controlPlane:
+    hyperthreading: Disabled
+  compute:
+    hyperthreading: Disabled
```
mhrivnak, line 145: Does putting settings like this here in an agent-specific structure just avoid the fact that openshift needs a k8s-native way to express some items that are currently only in the InstallConfig? There's nothing agent-specific about this setting AFAIK.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+metadata:
+  name: mycluster
+  namespace: hive
+spec:
+  stage: RetryProvisioning
+  clusterDeploymentRef:
+    name: mycluster
+  imageSetRef:
+    openshift-4.6
+  clusterMetadata:
```
mhrivnak, line 126: Why not have this in the Status? It's not user-declared state.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+      message: Some agent specific condition.
+      reason: AgentReason
+      status: "True"
+      type: AgentSpecificCondition
+```
+
+
+### Risks and Mitigations
+
+  * Can Hive controllers watch a dynamic set of types to reconcile when an install strategy changes? Can we watch unstructured?
```
mhrivnak, line 185: Hive will need to know the Group, Version and Kind to watch for each applicable type. During reconcile it can load the watched resources using Unstructured and interact with them that way.

docs/enhancements/install-strategy.md
mhrivnak: The whole time I've been reading this, the word "Strategy" felt like the wrong noun, and I figured I'd wait until I got to the end to bring it up. Having arrived, I'm happy to see that I'm not alone. :)

If I could wave a magic wand, I might call this new resource `ClusterDeployment` and call the other thing `Cluster`. But those ships have sailed.

`ClusterInstall` seems like a reasonable option.

dgoodwin: We've always avoided install in favor of provision/deprovision, but I guess with those terms out of the running, ClusterInstall might work.

docs/enhancements/install-strategy.md
dgoodwin: Examples are primarily things in CD.Spec.Platform.[CloudProvider] which I think the installers will often need and we would not want to duplicate. Felt like it made sense to not-duplicate anything, and assume that an Install must link up with a ClusterDeployment.

docs/enhancements/install-strategy.md
dgoodwin: Will remove.

docs/enhancements/install-strategy.md
dgoodwin: This would not be used by ClusterDeployment, we actually wouldn't care day 2 really as it's an install time only artifact. 

This would be used by ClusterInstall's though, I want to encourage a consistent way to control what versions of OpenShift can be installed in a Hive cluster, and we've got good established precedent for this in OSD and ACM both. I believe it's relevant to Agent installs and Hypershift as well.

docs/enhancements/install-strategy.md
dgoodwin: I have not thought about ClusterPool yet, will give that a look and update!

Historically at one point, OSD was supporting a provision a cluster feature where the CLusterDeployment was then removed, and the user was totally in control of the cluster after that point. Then, they wanted to be able to optionally deprovision those, which led us to an original implementation of ClusterDeprovision that was not tied to any ClusterDeployment as it may not exist. This entire feature eventually died, I'm not aware of anyone needing us to deprovision a cluster we did not provision anymore.

It feels like your suggestion would be safe, move the deprovision finalizer to the ClusterInstall.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+  * ReleaseImage: Legacy Hive feature, use ClusterImageSet instead.
+
+All conditions should initialize to Unknown status as soon as possible, and then be set to True/False once known.
+
+Condition | Description
+--------- | -----------
+Ready | Important overall condition indicating if the cluster is ready or not. Reason and Message provide as much detail as possible for the user. Message could contain attempt x of y if the install strategy support retries.
+RequirementsMet | True once we are ready to launch the install. False if we hit a problem and can't start. Reasons: ClusterImageSetNotFound, etc.
+Failed | True once an install attempt has failed. Remains true if retries are in play even if another attempt is in progress. Move to False once Ready goes to True. Reason/Message should contain details on what happened if it's possible to detect, otherwise Unknown. *TODO*: Unfortunate naming here, this matches current ProvisionFailed condition, which means a failure has been encountered, not necessarily that we've given up. Failing would be a better term, but it's risky to change the ProvisionFailed condition meaning on ClusterDeployment when we transfer. Thoughts?
```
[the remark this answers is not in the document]
dgoodwin: I don't think it would quite work as moving this to Reason we'd lose some granularity we're relying on, normal openshift installs scan the installer logs for strings that indicate specific problems, each of which becomes a reason, and that list changes often. It would be hard to monitor for one of those problems if they appear on the overall Ready condition. Having them on a ProvisionFailed/Failing condition means you can monitor for that to go True to know something is up, and you don't need to know specifics about the Reason to take action.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,196 @@
...
+Retrying | Unknown initially, True if we've hit an error and are retrying, False if we've given up or don't support retries. Message should indicate attempt x of y.
+
+These interface conditions will be copied back onto ClusterDeployment by Hive controllers with a "Provision" prefix, the install strategy controller does not need to worry about this and in general should never write to ClusterDeployment. example: Failed becomes ProvisionFailed (which exists today on ClusterDeployment for this purpose)
```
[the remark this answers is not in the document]
dgoodwin: It's part of the contract we have today with ACM/OSD, and I'm hesitant to break that.

docs/enhancements/install-strategy.md
dgoodwin: I was worried about that. How soon is real soon. :) We can get cracking fast but will need a little time on Hive side. The Agent controller changes should be fairly light.

docs/enhancements/install-strategy.md
dgoodwin: This is one of the hardest parts of the Hive api, we're in between a rock and a hard place all the time with this concept. We require installconfig for normal provisions. Agent provisions explicitly don't want to require an install config. We'd like users to not have to always supply one, but then we're duplicating install config, and boxing ourselves in from using new installer features without a hive update.

This overall proposal is an attempt to deal with that by now leaving it up to install implementations. Because Agent installs want to treat install config as an internal impl detail, they have the flexibility to duplicate the settings they wish to expose without impacting or confusing other parts of the HIve API. In the future, maybe there's a new Install impl that duplicates some of installconfig but gives you a purely CRD defined way to provision with openshift-install and you no longer have to provide the separate secret.

docs/enhancements/install-strategy.md
dgoodwin: We found in the past that if you want it persisted across a disaster recovery scenario (with velero or similar), it should go to spec if it cannot easily be reconstructed. This is why our "installed" flag is in spec as well.

docs/enhancements/install-strategy.md
dgoodwin: Excellent thanks!

dgoodwin:
> I'm not seeing where the duck typing is being used. Can you clarify, or add an example?

In hive core controllers basically, we don't watch or code to specific CRDs anymore for install, we work with unstructured and expect certain fields/conditions to be there.

docs/enhancements/install-strategy.md
dgoodwin: ClusterPool support is an interesting one. Looking at agent install strategy as it sits today, it *looks* like you might be able to get away with saying that an install strategy should be re-usable by multiple clusters? (can networking be shared?) However I really doubt this will be generically true. If it's not true, does it even make sense to support ClusterPool's for that platform/clusterinstall impl? I'm not sure it does. 

If we require specific complex settings per-cluster, ClusterPool is a very tough target to hit, it's basically why we're struggling with that vmware + arbitrary DNS requirement.

dgoodwin: Actually now I'm thinking this might be the key to solving that vmware DNS thing. ClusterPool.Spec.ClusterInstallRef points to a VMWareMyCustomDNS ClusterInstall. We don't care what or how that ClusterInstall defines or configures DNS for each cluster, it's up to whomever implements the controllers for it. Hive will just copy the VMWareMyCustomDNS ClusterInstall, as-is, directly to the ClusterDeployment namespace when we create it and associate it with the ClusterDeployment. 

Assumption is that if you're going to use it with a ClusterPool, it can't contain anything cluster specific, but it could contain ways to obtain something cluster specific. Adding a section with this proposal for your thoughts. This enhancement will list implementation of this as a non-goal.

docs/enhancements/install-strategy.md
dgoodwin: RequirementsMet is probably the one we want, I'll remove Initializing.

Provisioned sounds good as well.

docs/enhancements/install-strategy.md
dgoodwin: If it's ok I've instead moved this to a future considerations section, as I don't think we plan to implement this at this time.

docs/enhancements/install-strategy.md
dgoodwin: > Also can we include details on the deprovision of the clusters. I would be in favor of implementing the deprovision as the delete of the object that implements the ClusterInstall instead of a new different object.
> we shuoldn't need to support a deprovision that was not provisioned by our APIs.


A little unsure here now that I think about it. Wouldn't it feel like I provisioned my cluster, I have my ClusterDeployment now for day 2. I don't need my ClusterInstall anymore, so why not free up some space in etcd and delete it, which could then unexpectedly trash their cluster.

docs/enhancements/install-strategy.md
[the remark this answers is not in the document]
dgoodwin: Still a little stuck here.

At the core I think we need Ready, RequirementsMet, and Failed. But again Failed goes true as soon as "an attempt" failed, we didn't necessarily give up. So how does a caller monitor for whether or not we gave up entirely with just these three conditions.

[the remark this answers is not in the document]
dgoodwin: Provisioned - we successfully provisioned (possibly with multiple attempts)
Failed - an attempt failed but we might still be retrying
Completed - true if we're done trying. combine with Provisioned/Failed to know if success or failure. (Reason will also indicate)

Thoughts?

docs/enhancements/install-strategy.md
mhrivnak: It seems like the discussion is getting into feature overlap with central machine management, especially if openshift rebases to the latest cluster-api. It might make sense to further explore those implications before trying to solve deprovision here. If you already have an idea of what that will look like relative to hive, I'd be very interested to know more.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,233 @@
...
+AgentClusterInstall becomes a day 1 resource and will offer more flexibility for the agent controllers to define settings they need in isolation, and track their relevant status.
+
+
+### User Stories
+
+#### Story 1
+
+#### Story 2
+
+## Design Details
```
abhinavdahiya, line 43: we also need to include details on how these implementations providers can give hive access to these new cluster resources. i think we can use https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles to create an aggregate role for hive controllers.

docs/enhancements/install-strategy.md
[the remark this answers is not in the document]
dgoodwin: Ok Attempt #1125 

Installed - True once we've successfully installed. Reasons can indicate current progress, if we're in retries or not, etc.
Failed - A single attempt failed, includes details on detected problems. We might still be retrying.
Stopped - We've either hit Installed=True, or a limit on install attempts, or we hit a single error and do not support install retries.

Clients can monitor for Installed=True for success, and Stopped=True + Installed=False for a full failure, or Failed=True for any partial failure.

docs/enhancements/install-strategy.md
dgoodwin: So one last issue to resolve, in a DR scenario status will be lost. We need to record somewhere that this was officially completed in spec, and technically, we need to think about an install underway that needs cleanup and a retry. 

So I'm still wondering about a Stage in the Spec, even if it's just "InProgress" and "Installed" and everything else is conditions. Technically though we do have an installed flag on ClusterDeployment.Spec we're committed too, maybe this is enough. If we restored and saw ClusterDeployment.Installed = true, ClusterInstall controllers should not act on that ClusterInstall. 

I'll drop Stage for now but let me know what you think.

docs/enhancements/install-strategy.md
```diff
@@ -0,0 +1,233 @@
...
+spec.clusterMetadata.adminKubeconfigSecretRef | LocalObjectReference | Admin kubeconfig secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
+spec.clusterMetadata.adminPasswordSecretRef | LocalObjectReference | Admin password secret for current install attempt underway, set as soon as known. Copied back to Hive automatically once Ready.
+spec.clusterMetadata.clusterID | string | Cluster UUID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
+spec.clusterMetadata.infraID | string | Cluster infra ID generated during install, set as soon as known. Copied back to Hive automatically once Ready.
+spec.manifestsConfigMapRef | LocalObjectReference | ConfigMap containing manifests to inject into the cluster during installation. *TODO*: Hive technically doesn't need this to be there, do we require it to ensure consistent support or leave it up to install strategies?
```
dgoodwin, line 77: Will add status.installRestarts here.

docs/enhancements/install-strategy.md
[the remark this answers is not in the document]
abhinavdahiya: in the implementation i have found these conditions to be useful.
https://github.com/openshift/hive/pull/1358/files#diff-5b7e8b4c0bcc87a454db41521d292a136a5c26cf3efb29139b37aa9b07e40accR72-R76

RequirementsMet (this becomes ClusterInstallRequirementsMet on clusterdeployment)
Failed (this becomes ClusterInstallFailed on clusterdeployment)
Completed (this becomes ClusterInstallCompleted on clusterdeployment)

I think we can have the Completed include a reason that the install are Completed = False with Reason ReachedMaxAttempts .  Having a Stopped condition does not seem to provide additional value and it only creates more confusion as it's value can be part of the failure (no more attempts and therefore failure) or success (completed is true so we have stopped). Both these can easily be include in Failed and Completed conditions already...

docs/enhancements/install-strategy.md
abhinavdahiya: ^^ @dgoodwin i think we need to include something to solve the permission issue.

docs/enhancements/install-strategy.md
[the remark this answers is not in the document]
dgoodwin: We sorted out in scrum, decision was:

Completed = successfully installed
Failed = an attempt failed, we might still be retrying
Stopped = we are no longer working, combine with Completed/Failed to know result

docs/enhancements/install-strategy.md
dgoodwin: Added a section.

dgoodwin:
Updates pushed.

docs/enhancements/cluster-install-apis.md
```diff
@@ -0,0 +1,235 @@
...
+
+This will not be required for ClusterInstall implementations that live within Hive. (including AgentClusterInstall)
+
+### Cluster Deprovisioning
+
+Today in Hive each ClusterDeployment is given a finalizer automatically, and when the API resource is deleted we create a ClusterDeprovision resource, which launches a pod to run the openshift-install deprovision process until it completes, after which the finalizer is removed and the ClusterDeployment is removed.
+
+There is presently no deprovision implementation for agent based or bare metal clusters but there likely will be in the future.
+
+It stands to reason that deprovision implementations are fully tied to ClusterInstall implementations, and thus deleting the ClusterInstall should trigger the controllers for that ClusterInstall to complete an uninstall/deprovision. *WARNING: Deleting a ClusterInstall will destroy the cluster.*
```
dgoodwin, line 119: Clarified that deleting ClusterInstall will destroy the cluster.

Do we want to assume to add an OwnerRef on the ClusterInstall pointing to ClusterDeployment, so it is automatically garbage collected if the CD is deleted to match current behavior? Typically we do not add owner refs to things that we didn't create for you.

mhrivnak: It doesn't seem like a ClusterInstall makes sense on its own, and in fact `spec.clusterDeploymentRef` is required. So it makes sense to me to add the owner ref and let it get garbage collected if the CD is gone.

docs/enhancements/cluster-install-apis.md
```diff
@@ -0,0 +1,235 @@
...
+
+The core ClusterInstall CRDs and controllers discussed in this document should live in Hive. However, in theory, it would be possible for an external application to implement it's own ClusterInstall outside of Hive and still interface with ClusterDeployment using this system.
+
+#### External ClusterInstall RBAC
+
+To support an external application providing their own ClusterInstall CRD implementation, Hive will need RBAC to be able to access those objects. This can be achieved by using [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).
+
+Hive will be updated to deploy an empty ClusterRole `hive-controllers-clusterinstall` which will contain no rules, but will match the label `hive.openshift.io/aggregate-to-hive-clusterinstall`. Integrating components will need to deploy a ClusterRole with this label and rules that provide Hive all permissions on their custom ClusterInstall CRD.
+
+This will not be required for ClusterInstall implementations that live within Hive. (including AgentClusterInstall)
```
dgoodwin, line 111: This morning we agreed that we should move AgentClusterInstall over to the assisted service to live with it's controllers.

@mhrivnak had a suggestion for RBAC which I think is a little simpler where we agree upon extensions.hive.openshift.io as an APIGroup, and Hive grants itself full rights to everything in that APIGroup. If no one objects I'm going to update the enhancement to cover this instead of the above.

dgoodwin:
Updated with a few new commits, I think this is ready @abhinavdahiya

PR #1357, opened 2021-04-19
Add option to fake out all the clusterdeployments in the clusterpool

akhil-rane:
- Created `clusterpool.spec.annotations` field that can be used to apply fake cluster annotation on the cluster deployments
- Hibernation controller will now skip processing for cluster deployments with fake annotation

x-ref: https://issues.redhat.com/browse/HIVE-1493

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -159,6 +159,10 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	if !cd.Spec.Installed {
 		return reconcile.Result{}, nil
 	}
+	// If cluster is fake, skip any processing
+	if controllerutils.IsFakeCluster(cd) {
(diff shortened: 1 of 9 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -159,6 +159,10 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	if !cd.Spec.Installed {
 		return reconcile.Result{}, nil
 	}
+	// If cluster is fake, skip any processing
+	if controllerutils.IsFakeCluster(cd) {
+		return reconcile.Result{}, nil
```
stevekuznetsov, line 164: What does a consumer end up doing with this reconciliation? If a consumer is creating a claim against a pool, they're likely watching the resulting resource to see it go ready and use the kubeconfig. Will a pool with this fake annotation allow for this flow?

joelddiaz (commented):
/lgtm

PR #1359, opened 2021-04-21
Add regex to detect install failed waiting for kube api

akhil-rane:
install log snippet

```
      level=info msg=Creating infrastructure resources...
      level=info msg=Waiting up to 20m0s for the Kubernetes API at https://api.osde2e-nci3m.xmbp.p2.openshiftapps.com:6443...
      level=error msg=Attempted to gather ClusterOperator status after installation failure: listing ClusterOperator objects: Get "https://api.osde2e-nci3m.xmbp.p2.openshiftapps.com:6443/apis/config.openshift.io/v1/clusteroperators": dial tcp: lookup api.osde2e-nci3m.xmbp.p2.openshiftapps.com on 172.30.0.10:53: no such host
      level=info msg=Pulling debug logs from the bootstrap machine
      level=info msg=Bootstrap gather logs captured here "/output/log-bundle-20210421123336.tar.gz"
      level=error msg=Bootstrap failed to complete: Get "https://api.osde2e-nci3m.xmbp.p2.openshiftapps.com:6443/version?timeout=32s": dial tcp: lookup api.osde2e-nci3m.xmbp.p2.openshiftapps.com on 172.30.0.10:53: no such host
      level=error msg=Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane.
      level=fatal msg=Bootstrap failed to complete
      level=info msg=Pulling debug logs from the bootstrap machine
      level=info msg=Bootstrap gather logs captured here "/output/log-bundle-20210421123524.tar.gz"
```

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -120,6 +120,11 @@ data:
       - "failed to initialize the cluster: Some cluster operators are still updating:"
       installFailingReason: GeneralClusterOperatorsStillUpdating
       installFailingMessage: Timeout waiting for all cluster operators to become ready
+    - name: KubeAPIWaitFailed
+      searchRegexStrings:
+      - "Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane"
+      installFailingReason: KubeAPIWaitFailed
+      installFailingMessage: Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane
     # Processing stops at the first match, so this more generic
     # message about the connection failure must always come after the
     # more specific message for LibvirtSSHKeyPermissionDenied.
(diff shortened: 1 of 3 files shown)


gregsheremeta:
/cc @abhinavdahiya
Please review

I think the most interesting thing you could report here is that bootstrap failed. More interesting than that would be the reason why it failed. The api not coming up is a side effect of those two things, and while I think it's useful to report that sometimes, maybe only as a fallback.

Interested in Abhinav's expertise here.

abhinavdahiya:
> /cc @abhinavdahiya
> Please review
> 
> I think the most interesting thing you could report here is that bootstrap failed. More interesting than that would be the reason why it failed. The api not coming up is a side effect of those two things, and while I think it's useful to report that sometimes, maybe only as a fallback.
> 
> Interested in Abhinav's expertise here.

I think the installed team is working on providing more accurate reasons for why k8s api or bootstrapping is failing.
untill then i don't know there is a good way to tell why this error class happened.

PR #1360, opened 2021-04-21
Sketch out ClusterInstall API contract, and a FakeClusterInstall reference implementation.

dgoodwin:
/assign @abhinavdahiya 
/cc @joelddiaz


dgoodwin:
Switched over to cluster-installs branch for target and rebased on your PR @abhinavdahiya

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// Check if we're already Stopped=True and return if so.
+	// If not, we should be Stopped=False as we're actively working.
+	stoppedCond := controllerutils.FindClusterInstallCondition(fci.Status.Conditions, hivev1.ClusterInstallStopped)
+	if stoppedCond == nil {
+		errMsg := fmt.Sprintf("%s condition is nil but should be set", hivev1.ClusterInstallStopped)
+		return reconcile.Result{}, errors.New(errMsg)
+	}
```
abhinavdahiya, line 153: personally, i would depend on Installed = true, to ensure that stopped is set to true.

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+		logger.Debug("ClusterDeployment has been deleted")
+		return reconcile.Result{}, nil
+	}
+
+	// Simulate 30 second wait for RequirementsMet condition to go True:
+	reqsCond := controllerutils.FindClusterInstallCondition(fci.Status.Conditions, hivev1.ClusterInstallRequirementsMet)
+	if reqsCond == nil {
+		errMsg := fmt.Sprintf("%s condition is nil but should be set", hivev1.ClusterInstallRequirementsMet)
+		return reconcile.Result{}, errors.New(errMsg)
+	}
```
abhinavdahiya, line 193: > https://github.com/openshift/hive/pull/1360/files#diff-646c62f35fdb8ec4d5935d66552a1f31ddded8a4394567746419062ec8782c66R125-R145

that will already ensure that we have required conditions set to unknown, so this checks seems not very useful?

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	if reqsCond.Status == corev1.ConditionUnknown {
```
abhinavdahiya, line 194: maybe do a switch here, that would read cleaner..?

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+			controllerutils.UpdateConditionIfReasonOrMessageChange)
+		if changed {
+			fci.Status.Conditions = newConditions
+			err := updateClusterInstallStatus(r.Client, fci, logger)
+			return reconcile.Result{}, err
+		}
+	} else if reqsCond.Status == corev1.ConditionFalse {
+		// Check if it's been 30 seconds since we set condition to False:
+		delta := time.Now().Sub(reqsCond.LastTransitionTime.Time)
+		if delta >= 30*time.Second {
```
abhinavdahiya, line 211: return early here?

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+			return reconcile.Result{RequeueAfter: 30*time.Second - delta}, nil
+		}
+	}
+
+	// Simulate 30 second wait for Completed condition to go True:
+	completedCond := controllerutils.FindClusterInstallCondition(fci.Status.Conditions, hivev1.ClusterInstallCompleted)
+	if completedCond == nil {
+		errMsg := fmt.Sprintf("%s condition is nil but should be set", hivev1.ClusterInstallCompleted)
+		return reconcile.Result{}, errors.New(errMsg)
+	}
```
abhinavdahiya, line 236: same as https://github.com/openshift/hive/pull/1360/files#r619414016

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	if completedCond.Status == corev1.ConditionUnknown {
```
abhinavdahiya, line 237: same, maybe switch will be easier to read here.

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,313 @@
...
+				AdminKubeconfigSecretRef: corev1.LocalObjectReference{Name: "admin-kubeconfig"},
+				AdminPasswordSecretRef:   corev1.LocalObjectReference{Name: "admin-password"},
+			}
+			logger.Info("setting fake ClusterMetadata")
+			return reconcile.Result{}, r.Client.Update(context.Background(), fci)
+		}
+
+		// Check if it's been 30 seconds since we set condition to False:
+		delta := time.Now().Sub(completedCond.LastTransitionTime.Time)
+		if delta >= 30*time.Second {
```
abhinavdahiya, line 268: same, return early here?

apis/hive/v1/agentclusterinstall_types.go
```diff
@@ -0,0 +1,155 @@
...
+	// +optional
+	SSHPublicKey string `json:"sshPublicKey,omitempty"`
+
+	// ProvisionRequirements defines configuration for when the installation is ready to be launched automatically.
+	ProvisionRequirements ProvisionRequirements `json:"provisionRequirements"`
+
+	// ControlPlane is the configuration for the machines that comprise the
+	// control plane.
+	// +optional
+	ControlPlane *AgentMachinePool `json:"controlPlane,omitempty"`
```
abhinavdahiya, line 65: shouldn't we use hive/v1/agent.MachinePool here?

apis/hive/v1/agentclusterinstall_types.go
```diff
@@ -0,0 +1,155 @@
...
+	ControlPlaneAgents int `json:"controlPlaneAgents"`
+
+	// WorkerAgents is the minimum number of matching approved and ready Agents with the worker role
+	// required to launch the install.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	WorkerAgents int `json:"workerAgents,omitempty"`
+}
+
+// Networking defines the pod network provider in the cluster.
```
abhinavdahiya, line 87: there are already defined in hive/v1/agent/installstrategy.go ?? we should try and not crowd the hive/v1 package with these specific types..

apis/hive/v1/agentclusterinstall_types.go
dgoodwin: I don't think we can, there's too much there and no commitment with assisted team they want to expose and support all of that yet. The concept of MachinePools hasn't even really been discussed. Possible in future but right now I think we can only expose the minimum. 

API wise we should be able to transition, assuming we add hyperthreading on the Hive MachinePool. :)

apis/hive/v1/agentclusterinstall_types.go
dgoodwin: hive/v1/agent/installstrategy.go will die off, I discussed with assisted team that we won't yank the API until we complete transition, but this is meant to fully obsolete that.

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
dgoodwin: That would be technically fine in this controller I just was thinking this might serve as a bit of a reference for anyone implementing one of these. In that case we could still hit Completed=False and Stopped=True. Below I get a little more fakecluster specific and I set Stopped=True when Completed=True. Does that seem reasonable?

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
dgoodwin: Yeah probably safe enough, good point.

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
dgoodwin: Sounds good.

dgoodwin:
Updated.

apis/hive/v1/agentclusterinstall_types.go
dgoodwin: Actually given we're in a branch, we can probably remove the old. I'm going to go do that now, I also have to move some missed InstallStrategyStatus over.

apis/hive/v1/agentclusterinstall_types.go
abhinavdahiya: i still don't like that we have something like hive/v1.AgentMachinePool , this should be name spaced to agent package. whatever name we like we can use that if MachinePool is heavy handed.

apis/hive/v1/agentclusterinstall_types.go
abhinavdahiya: I still don't think we should add a Networking struct to hive/v1 when we currently know this is only meant to be used with agent cluster install. I would recommend we move this to hive/v1/agent so that this Networking doesn't become our fo to for all the other types we might add later on.

apis/hive/v1/agentclusterinstall_types.go
```diff
@@ -0,0 +1,170 @@
...
+	ControlPlane *AgentMachinePool `json:"controlPlane,omitempty"`
+
+	// Compute is the configuration for the machines that comprise the
+	// compute nodes.
+	// +optional
+	Compute []AgentMachinePool `json:"compute,omitempty"`
+}
+
+// ProvisionRequirements defines configuration for when the installation is ready to be launched automatically.
+type ProvisionRequirements struct {
```
abhinavdahiya, line 74:
```suggestion
type AgentProvisionRequirements struct {
```

or move this to hive/v1/agent package

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
abhinavdahiya: not reconciling when Stopped = True is a bad recommendation imo. For example, if the impl supported max retries it should also support user increasing the max retries when it previously stopped. This example provides bad advice.

Decide where there is no work to be done on Completed = true or other computable level state.

apis/hive/v1/agentclusterinstall_types.go
dgoodwin: I actually tried that at first but it caused a circular dependency. Now that we've killed off the old types though it might be better. We ref from ClusterDeployment.Spec.Platform into agent package, I think the only remaining problem is AgentClusterInstall.Spec.ImageSet ref being hivev1.ClusterImageSetReference. We could duplicate it, wdyt?

abhinavdahiya: i don't understand what caused the circular dependency? hive/v1/agent has no import to hive/v1 ...

dgoodwin: Abhinav clarified in scrum that this was referring just to the sub-structs like Networking and others, to keep them out of the main hivev1 package and from being used inadvertently. I will relocate those.

apis/hive/v1/agentclusterinstall_types.go
dgoodwin: Moved AgentMachinePool to agent package, I'm still not sure what to call it, MachineConfig seems taken already. We could just move ControlPlaneHyperthreading and ComputeHyperthreading up to Spec but it seems likely they'll someday want more settings for those.

apis/hive/v1/agentclusterinstall_types.go
dgoodwin: done.

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
dgoodwin: Good point, updating.

dgoodwin:
I've updated with the PR feedback and moved AgentClusterInstall.

However in parallel discussions around the OLM bundle limit, I am worried we cannot add a new CRD just for a FakeClusterInstall implementation. I think I have to break this out into a separate repo just for hive testing or something.

dgoodwin:
I'll drop fake controller in another PR if we decide to. This won't hit master just yet anyhow.

dgoodwin:
Ready for another round @abhinavdahiya

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -624,11 +624,11 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
-	// Return early and stop processing if Agent install strategy is in play. The controllers that
+	// Return early and stop processing if ClusterInstallRef is in play. The controllers that
 	// handle this portion of the API currently live in the assisted service repo, rather than hive.
 	// This will hopefully change in the future.
-	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallStrategy != nil && cd.Spec.Provisioning.InstallStrategy.Agent != nil {
-		logger.Info("skipping processing of agent install strategy cluster")
+	if cd.Spec.ClusterInstallRef != nil {
+		logger.Info("ClusterInstallRef specified, skipping creation of ClusterProvision")
 		return reconcile.Result{}, nil
 	}
```
abhinavdahiya, line 633: i don't think we need to add duplicate check for clusterinstallref here, the function already assumes that it was invoked for clusterprovision

pkg/controller/fakeclusterinstall/fakeclusterinstall_controller.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// ReconcileClusterInstall is the reconciler for FakeClusterInstall.
+type ReconcileClusterInstall struct {
+	client.Client
+	scheme *runtime.Scheme
+	logger log.FieldLogger
+
+	// updateStatus updates a given cluster state's status, exposed for testing
+	updateStatus func(client.Client, *hivev1.FakeClusterInstall, log.FieldLogger) error
```
abhinavdahiya, line 84: there are no users of this?

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -280,25 +280,10 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateCreate(admissionSpec
...
-				allErrs = append(allErrs, validateAgentInstallStrategy(specPath, cd)...)
-			} else if cd.Spec.Platform.AgentBareMetal != nil {
-				// agent bare metal platform can only be used with agent install strategy:
-				allErrs = append(allErrs,
-					field.Forbidden(specPath.Child("platform", "agentBareMetal"),
-						"agent bare metal platform can only be used with agent install strategy"))
-			}
+	if !cd.Spec.Installed && cd.Spec.Provisioning != nil {
+		// InstallConfigSecretRef is not required for anyone using the new ClusterInstall interface:
+		if (cd.Spec.Provisioning.InstallConfigSecretRef == nil || cd.Spec.Provisioning.InstallConfigSecretRef.Name == "") && cd.Spec.ClusterInstallRef == nil {
```
abhinavdahiya, line 285: i don't think we should need to support the case where people are setting the provision without install config but also setting cluster install ref.

with set the provision section correctly or don't add it.

abhinavdahiya: so drop the `&& cd.Spec.ClusterInstallRef == nil` imo

dgoodwin:
Updated per review and moved fakeclusterinstall to hiveinternal.

We'll look at getting internal CRDs out of the bundle in a separate jira.

PR #1361, opened 2021-04-22
validating-webhooks: include diff when immutable fields are changed

abhinavdahiya:
when the spec has changes on fields that are not mutable, return the
list of fields that changed (with values) that should not have been
changed.

for example,
```
oc patch cd test-cluster --type='merge' -p $'spec:\n baseDomain:
devcluster.openshift.co#\n clusterName: test-cluste#\n platform:\n
aws:\n   region: us-east-#'
Error from server (BadRequest): admission webhook
"clusterdeploymentvalidators.admission.hive.openshift.io" denied the
request: Attempted to change ClusterDeployment.Spec which is immutable
except for
CertificateBundles,ClusterMetadata,ControlPlaneConfig,Ingress,Installed,PreserveOnDelete,ClusterPoolRef,PowerState,HibernateAfter,InstallAttemptsLimit,MachineManagement
fields. Unsupported change:
        ClusterName: (test-cluster => test-cluste#)
        BaseDomain: (devcluster.openshift.com => devcluster.openshift.co#)
        Platform.AWS.Region: (us-east-2 => us-east-#)

```

/assign @joelddiaz 
/cc @dgoodwin 

xref: https://issues.redhat.com/browse/HIVE-1496

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/go-logr/logr v0.3.0
 	github.com/golang/mock v1.4.4
 	github.com/golangci/golangci-lint v1.31.0
+	github.com/google/go-cmp v0.5.2
 	github.com/google/uuid v1.1.2
 	github.com/gophercloud/utils v0.0.0-20210113034859-6f548432055a
 	github.com/heptio/velero v1.0.0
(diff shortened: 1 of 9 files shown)


abhinavdahiya:
another side effect of this new check is that it supports defining mutable fields in spec that are not top-level fields, like `Platform.AWS.PrivateLink` where previously we couldn't

joelddiaz (commented):
nice!
/lgtm

PR #1364, opened 2021-04-23
Initialize Hibernating condition

suhanime:

This commit Initializes cluster hibernating condition as soon as possible.
It is to False for Unsupported clouds, and to Unknown for others

xref https://issues.redhat.com/browse/HIVE-1509

/cc @dgoodwin

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -64,6 +64,12 @@ var (
 	// actuators is a list of available actuators for this controller
 	// It is populated via the RegisterActuator function
 	actuators []HibernationActuator
+
+	// clusterDeploymentHibernationConditions are the cluster deployment conditions controlled by
+	// hibernation controller
+	clusterDeploymentHibernationConditions = []hivev1.ClusterDeploymentConditionType{
+		hivev1.ClusterHibernatingCondition,
+	}
 )
 
 // Add creates a new Hibernation controller and adds it to the manager with default RBAC.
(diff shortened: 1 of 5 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -245,7 +268,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	}
 
 	if !shouldHibernate {
-		if hibernatingCondition == nil || hibernatingCondition.Status == corev1.ConditionFalse {
+		if hibernatingCondition.Status == corev1.ConditionFalse {
```
dgoodwin, line 271: Might be the source of some of your bugs here, the nil check should be replaced with an Unknown check.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -257,7 +280,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, nil
 	}
 
-	if hibernatingCondition == nil || hibernatingCondition.Status == corev1.ConditionFalse ||
+	if hibernatingCondition.Status == corev1.ConditionFalse ||
```
dgoodwin, line 283: Again here.

pkg/controller/utils/conditions.go
```diff
@@ -41,6 +41,26 @@ func shouldUpdateCondition(
 	return updateConditionCheck(oldReason, oldMessage, newReason, newMessage)
 }
 
+// InitializeConditions initializes the given set of conditions for the first time, set with Status Unknown
+func InitializeConditions(existingConditions []hivev1.ClusterDeploymentCondition, conditionsToBeAdded []hivev1.ClusterDeploymentConditionType) []hivev1.ClusterDeploymentCondition {
```
abhinavdahiya, line 45: nit: InitializeClusterDeploymentConditions

suhanime: Done!

PR #1365, opened 2021-04-23
Change ClusterProvision image pull policy

jharrington22:
To mitigate ClusterProvision pods not starting during a container image registry outage change the CusterProvision pod `imagePullPolicy` from `Always` to `IfNotPresent`.

This should allow any ClusterProvision pod to start without entering `ImagePullBackoff` provided the `hive` image pod is cached locally.

This is a temporary mitigation until [HIVE-1504](https://issues.redhat.com/browse/HIVE-1504) is implemented and Hive can take advantage of registry mirrors.

[HIVE-1505](https://issues.redhat.com/browse/HIVE-1505)

### DIFF
--- a/pkg/controller/images/controller_images.go
+++ b/pkg/controller/images/controller_images.go
@@ -20,6 +20,12 @@ const (
 	// the pull policy to use for pulling the hive image. Typically this is the same
 	// pull policy used by the the hive operator and is set as an EnvVar on the deployment.
 	HiveImagePullPolicyEnvVar = "HIVE_IMAGE_PULL_POLICY"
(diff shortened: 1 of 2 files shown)


jharrington22:
The only issue I can see here is anyone using `HIVE_IMAGE_PULL_POLICY` ENV will not expect this change. I don't see it documented anywhere and since its just the pull policy its probably ok temporarily as I'd expect this to be set back to `GetHiveImagePullPolicy()` when HIVE-1504 is complete.

mwoodson (approved):
/lgtm

PR #1366, opened 2021-04-23
operator: use the correct cache to get secrets

abhinavdahiya:
When we receive a watch, and we reconcile for that watch and then use
that cache to get the latest version of the changed object, we can
assume this gurantee.

Currently, due to historical reasons, we have a separate cache for
watching secrets in hive namespace, but then use the reconciler's client
to get the secrets. This means that our assumption of having the latest
update secret is no longer valid.

So in the current setup, we should use the lister from the correct
cache to get secrets. This is now achieved by storing the lister for the
secrets and then using that when required.

> https://github.com/openshift/hive/pull/1350

that change allows reconcile on secret changes that are used in hiveconfig esp the client certificates secret.
but when the secret was updated, the reconcile would use the state secret data in reconciler client's cache. To make
sure we have the updated secret content we must use the cache that triggered the watch. So using hiveSecretLister
allows us to get the updated secret content.

/assign @joelddiaz

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -17,7 +17,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


joelddiaz (commented):
/lgtm

PR #1368, opened 2021-04-28
HIVE-1498: Add InvalidInstallConfigSubnet install log regex.

abutcher:
/assign @dgoodwin 

I also added the regex to the additional template which I'm testing in integration.

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -31,11 +31,11 @@ data:
       - "LimitExceeded"
       installFailingReason: ResourceLimitExceeded
       installFailingMessage: Resource limit exceeded
-    - name: InvalidSubnetCIDRConfig
+    - name: InvalidInstallConfigSubnet
       searchRegexStrings:
-      - "subnet\'s CIDR range"
-      installFailingReason: InvalidSubnetCIDRConfig
-      installFailingMessage: Invalid subnet CIDR provided in install config.
+      - "CIDR range start.*is outside of the specified machine networks"
+      installFailingReason: InvalidInstallConfigSubnet
+      installFailingMessage: Invalid subnet in install config. Subnet's CIDR range start is outside of the specified machine networks
     # https://bugzilla.redhat.com/show_bug.cgi?id=1844320
     - name: AWSUnableToFindMatchingRouteTable
       searchRegexStrings:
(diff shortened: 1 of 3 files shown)


abutcher:
Example error we're catching:

```
msg=failed to fetch Master Machines: failed to load asset \"Install Config\": [platform.aws.subnets[0]: Invalid value: \"subnet-123456\": subnet's CIDR range start 10.1.1.0 is outside of the specified machine networks, platform.aws.subnets[1]: Invalid value: \"subnet-123456\": subnet's CIDR range start 10.1.100.0 is outside of the specified machine networks]\n"
```

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -31,6 +31,11 @@ data:
       - "LimitExceeded"
       installFailingReason: ResourceLimitExceeded
       installFailingMessage: Resource limit exceeded
+    - name: InvalidSubnetCIDRConfig
+      searchRegexStrings:
+      - "subnet\'s CIDR range"
+      installFailingReason: InvalidSubnetCIDRConfig
+      installFailingMessage: Invalid subnet CIDR provided in install config.
```
dgoodwin, line 38: Slight nit, the regex looks just slightly lacking in an indication of something negative, could we extend a little further to include "is outside of the specified machine networks".

hack/regexes/additional-hive-regexes.yaml
```diff
@@ -14,3 +14,8 @@ objects:
             - "EIP: AddressLimitExceeded"
           installFailingReason: EIPAddressLimitExceeded
           installFailingMessage: EIP Address limit exceeded
+        - name: InvalidSubnetCIDRConfig
+          searchRegexStrings:
+            - "subnet\'s CIDR range"
+          installFailingReason: InvalidSubnetCIDRConfig
+          installFailingMessage: Invalid subnet CIDR provided in install config.
```
dgoodwin, line 21: What value does having in additional-hive-regexes have? Does this tie into what rolls out immediately?

dgoodwin:
> What about generically checking for `failed to load asset \"Install Config\"` as well?

Seems reasonable for low in the list.

hack/regexes/additional-hive-regexes.yaml
abutcher: Yes.

config/configmaps/install-log-regexes-configmap.yaml
abutcher: Made more specific (assuming we want to go ahead with adding it for now) and renamed to `InvalidInstallConfigSubnet`

dgoodwin:
Business as usual until we sort out that proposal from Abhinav.

Thanks!

/lgtm

PR #1369, opened 2021-04-28
operator: support already known contracts

abhinavdahiya:
There are certain contracts and their implementations that we know
are supported always. So we make sure these are allowed without the
discovery from CRDs.

/assign @dgoodwin

### DIFF
--- a/pkg/operator/hive/hiveadmission.go
+++ b/pkg/operator/hive/hiveadmission.go
@@ -329,18 +329,33 @@ var allowedContracts = sets.NewString(
 	hivecontractsv1alpha1.ClusterInstallContractLabelKey,
 )
 
+// knowContracts is a list of contracts and their implementations that doesn't
+// require discovery using CRDs
+var knowContracts = contracts.SupportedContractImplementationsList{{
+	Name: hivecontractsv1alpha1.ClusterInstallContractName,
+	Supported: []contracts.ContractImplementation{{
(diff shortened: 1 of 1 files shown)


pkg/operator/hive/hiveadmission.go
```diff
@@ -329,18 +329,33 @@ var allowedContracts = sets.NewString(
 	hivecontractsv1alpha1.ClusterInstallContractLabelKey,
 )
 
+// knowContracts is a list of contracts and their implementations that doesn't
+// require discovery using CRDs
+var knowContracts = contracts.SupportedContractImplementationsList{{
+	Name: hivecontractsv1alpha1.ClusterInstallContractName,
+	Supported: []contracts.ContractImplementation{{
+		Group:   "extensions.hive.openshift.io",
+		Version: "v1",
```
dgoodwin, line 338: AgentClusterInstal (and all other assistedservice apis) are v1beta1 in their repo right now.

@rollandf @mhrivnak do we want to keep that or move it to v1 for AgentClusterInstall?

mhrivnak: If we can keep it at v1beta1, that's better for now since it's brand new. We can use whatever flexibility we can get.

abhinavdahiya: updated to v1beta1

PR #1370, opened 2021-04-29
Remove Massive PodSpec in ClusterProvision CRD yaml

dgoodwin:
Hide massive PodSpec definition in ClusterProvision CRD yaml.

    This PodSpec alone is taking up 350K of the hard 1024 limit currently
    imposed on OLM bundles.

    We do not consider PodSpec an integral part of this API and more an
    implementation detail, it is extremely unlikely anyone is relying on
    being documented in the CRD.

    x-ref: https://issues.redhat.com/browse/HIVE-1513

    Because we still use v1beta1 CRDs, preserveUnknownFields is default.
    When we switch to v1 we will need preserveUnknown on the
    ClusterProvisionSpec to keep the PodSpec maintained. This is noted in
    the relevant Jira.


config/crds/hive.openshift.io_clusterprovisions.yaml
```diff
@@ -5905,6 +110,7 @@ spec:
           - podSpec
           - stage
           type: object
+          x-kubernetes-preserve-unknown-fields: true
```
dgoodwin, line 113: Not sure why yet but when this is added, Kube no longer wants to let me apply the CRD:

```shell
~/go/src/github.com/openshift/hive remove-clusterprovision-podspec*
❯ k apply -f config/crds/hive.openshift.io_clusterprovisions.yaml
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
The CustomResourceDefinition "clusterprovisions.hive.openshift.io" is invalid: spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root


```

dgoodwin: It looks like adding x-kubernetes-preserve-unknown-fields to any v1beta1 CRD causes this error, it may not be supported until v1. 

This means we need to move from v1beta1 CRD generation to v1, which means we need to orphan the v3 OpenShift Hive shards from receiving updates, if we're going to fix this this way.

dgoodwin:
Resolved, the preserve unknown is default behavior for v1beta1 CRDs and the explicit flag doesn't seem to be supported, it would be when we switch to v1, but we don't need it for now. Ready for review.

/assign @abhinavdahiya 
/cc @suhanime

Makefile
```diff
@@ -119,6 +119,9 @@ crd: ensure-controller-gen ensure-yq
 	$(foreach p,$(wildcard ./config/crds/*.yaml),$(call strip-yaml-break,$(p)))
 	@echo Patching CRD files for additional static information
 	$(foreach p,$(wildcard ./config/crdspatch/*.yaml),$(call patch-crd-yq,$(subst ./config/crdspatch/,./config/crds/,$(p)),$(p)))
+	# Patch ClusterProvision CRD to remove the massive PodSpec def we consider an internal implementation detail:
+	@echo Patching ClusterProvision CRD yaml to remove overly verbose PodSpec details:
+	yq d -i config/crds/hive.openshift.io_clusterprovisions.yaml "spec.validation.openAPIV3Schema.properties.spec.properties.podSpec"
```
abhinavdahiya, line 124: can we use the existing patch (crdspatch) instead of another yq command ?

also don't delete it entirely maybe?

dgoodwin: The current crdspatch is a merge, I am not aware of any way to do a merge that deletes something.

What purpose would keeping podSpec in there serve?

dgoodwin: I can delete just .properties beneath that and get most of it but I'm not clear why we'd want to keep the podSpec at all.

abhinavdahiya: > I can delete just .properties beneath that and get most of it but I'm not clear why we'd want to keep the podSpec at all. Seems more logical that the podSpec entirely is the unknown field to be preserved, and users don't need to see it in the CRD definition.

because that shows up that there is a podSpec object, just that we don't care/know what's in it. compared to completely ghost field..

dgoodwin: I think I'd prefer ghost field honestly, there's no one I want seeing that field there other than hive internals, so I think the more hidden it is the better. (provided it works, which I think it does in this form)

abhinavdahiya: /lgtm

PR #1371, opened 2021-04-29
awsprivatelink: add Name tag to endpoint and endpoint service

abhinavdahiya:
xref: https://issues.redhat.com/browse/HIVE-1517

/assign @joelddiaz 
/cc @jharrington22


joelddiaz (commented):
/lgtm

PR #1372, opened 2021-04-29
Set ClusterDeploymentCondition for all status

suhanime:
Currently we assume all ClusterDeploymentConditions to have negative polarity, consequently we only set them for the first time if their Status=True. Eliminate this assumption and allow all conditions to be set regardless of their status.

/cc @dgoodwin

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
@@ -57,12 +57,8 @@ func Test_setErrCondition(t *testing.T) {
 		err:    errors.New("failed to do something important"),
 		reason: "FailureToDoSomethingImportant",
 
-		expectedConditions: []hivev1.ClusterDeploymentCondition{{
-			Status:  corev1.ConditionTrue,
-			Type:    hivev1.AWSPrivateLinkFailedClusterDeploymentCondition,
-			Reason:  "FailureToDoSomethingImportant",
-			Message: "failed to do something important",
-		}},
+		expectedConditions: getExpectedConditions(true, "FailureToDoSomethingImportant",
+			"failed to do something important"),
 	}, {
 		name: "previous failure",
 
@@ -75,12 +71,8 @@ func Test_setErrCondition(t *testing.T) {
 		err:    errors.New("failed to do something important"),
 		reason: "FailureToDoSomethingImportant",
 
-		expectedConditions: []hivev1.ClusterDeploymentCondition{{
-			Status:  corev1.ConditionTrue,
-			Type:    hivev1.AWSPrivateLinkFailedClusterDeploymentCondition,
-			Reason:  "FailureToDoSomethingImportant",
-			Message: "failed to do something important",
-		}},
+		expectedConditions: getExpectedConditions(true, "FailureToDoSomethingImportant",
+			"failed to do something important"),
 	}, {
 		name: "previous ready",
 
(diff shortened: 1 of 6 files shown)


dgoodwin:
Looks like the polarity stuff from the metrics pr slipped in here. Main change looks good, good luck with those tests. :)

suhanime:
> Looks like the polarity stuff from the metrics pr slipped in here. Main change looks good, good luck with those tests. :)

Oops! Dropped it...

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -910,7 +973,17 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 		{
 			name: "Do not adopt failed provision",
 			existing: []runtime.Object{
-				testClusterDeployment(),
+				//testClusterDeployment(),
```
dgoodwin, line 976: Missed commented code.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -195,15 +202,45 @@ func TestClusterDeploymentReconcile(t *testing.T) {
...
+							Message: imagesResolvedMsg,
+						})
+						cd.Status.Conditions = append(cd.Status.Conditions, hivev1.ClusterDeploymentCondition{
+							Status:  corev1.ConditionFalse,
+							Type:    hivev1.AuthenticationFailureClusterDeploymentCondition,
+							Reason:  platformAuthSuccessReason,
+							Message: "Platform credentials passed authentication check",
+						})
+						return cd
+					}(), cd; !assert.True(t, apiequality.Semantic.DeepEqual(e, a), "unexpected change in clusterdeployment") {
```
dgoodwin, line 243: It feels like there's a good assumed default state for test setup we could put into testClusterDeployment() function itself, ProvisionStopped=False, InstallImagesNotResolved=False, AuthenticationFailure=False. Does this work if you just assume to always set those "default" conditions in the default test input?

dgoodwin (commented):
Seems good to me, just a couple minor nits.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
suhanime: fixed

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
suhanime: That's a good idea - made a new function to do just that

suhanime: FYI did not change original testClusterDeployment() to allow for testing whether clusterdeployment conditions are initialized later on

pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
```diff
@@ -783,6 +743,12 @@ users:
 			m.EXPECT().DescribeLoadBalancers(gomock.Any()).
 				Return(nil, awserr.New("LoadBalancerNotFound", "Loadbalance could not be found", nil))
 		},
+		expectedConditions: []hivev1.ClusterDeploymentCondition{{
+			Status:  corev1.ConditionFalse,
+			Type:    hivev1.AWSPrivateLinkReadyClusterDeploymentCondition,
+			Reason:  "DiscoveringNLBNotYetFound",
+			Message: "discovering NLB for the cluster, but it does not exists yet",
+		}},
```
akhil-rane, line 751: `getExpectedConditions` is not used here. It will be helpful if you can document when to use this function.

PR #1375, opened 2021-05-03
remotemachinesets: sync ready replicas and errors to machinepool

abhinavdahiya:
Add ReadyReplicas and Error{Reason, Message} to status for MachineSets
in the MachinePool. This allows the machinepool to provider more
accurate status of important information of the machinesets in the remote
cluster without having to probe into the cluster itself.

There is still a caveat the the machinepool doesn't react to the
machinesets in the remote cluster, so this status will not be latest
unless the machinepool changed or other updates caused a re sync.

to work around this caveat at least during known non-steady state like scale up,
When the ready replicas and required replicas do not match, we must make
sure we requeue to update the status to match the state in the remote
cluster. This is required because we cannot yet watch machinesets in
remote cluster.

xref: https://issues.redhat.com/browse/HIVE-1511
xref: https://issues.redhat.com/browse/HIVE-1510


joelddiaz (commented):
/lgtm

PR #1377, opened 2021-05-07
Add condition to display status of deployment on HiveConfig

akhil-rane:


dgoodwin:
/assign
/lgtm

Will take this one as Abhinav is out most of the week.

dgoodwin:
/lgtm cancel

Actually should have let @suhanime have a look as well. Over to you Suhani.

pkg/operator/util/conditions.go
```diff
@@ -0,0 +1,55 @@
...
+				hivev1.HiveConfigCondition{
+					Type:               conditionType,
+					Status:             status,
+					Reason:             reason,
+					Message:            message,
+					LastTransitionTime: now,
+					LastProbeTime:      now,
+				},
+			)
+		}
```
suhanime, line 31: I say we move away from this assumption that we only set a condition for the first time if it is true

suhanime (commented):
@akhil-rane I think following the principles of [1478](https://issues.redhat.com/browse/HIVE-1478) as we move forward would benefit the consistency of the codebase. So always having the Ready condition on the CR, initializing it to unknown first and then set it to false or true, whatever seems right.
Also, could you add a few unit tests to cover these changes?

pkg/operator/util/conditions.go
dgoodwin: Good point.

If there are no pre-existing unit tests then I would defer that and do as a separate Jira, but if there are some then yes would be good to add in this PR. I can't recall.

pkg/operator/hive/hive_controller.go
```diff
@@ -298,10 +300,15 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	// Initialize hive ready condition if not present to Unknown
+	instance.Status.Conditions = util.SetHiveConfigCondition(instance.Status.Conditions, hivev1.HiveReadyCondition, corev1.ConditionUnknown, hivev1.InitializedConditionReason, "Condition Initialized")
```
suhanime, line 304: Technically you're setting this condition on every reconcile before it eventually gets set to True or False further in the code - so particularly for this scenario it would be better if you try to find the condition first before calling this line of code

PR #1378, opened 2021-05-07
Add prometheus target for hive-operator metrics

akhil-rane:


dgoodwin:
Very nice, once everything merges we'll need to go configure appsre's app-interface to get this working in stage/prod, and then update the alerts.

PR #1379, opened 2021-05-10
Add owner ref on FakeClusterInstall.

dgoodwin:
Per enhancement, the fake cluster install controller will now add an owner reference to itself indicating it is owned by it's ClusterDeployment. Used for cleanup.


akhil-rane (commented):
/lgtm

PR #1380, opened 2021-05-11
Fix hotloop in awsprivatelink controller in some situations

dgoodwin:
x-ref: https://issues.redhat.com/browse/HIVE-1536

In this scenario the controllers are trying to teardown a cluster that
was never fully installed, and failing because aws private link cleanup
requires an admin kubeconfig Secret, which does not seem to exist. (this
fix will be pursued separately)

The hotloop was caused by the fact that we were setting the Ready
condition to indicate we're trying to tear down, and then also setting
it after an error occurs. Each write causes another reconcile, so we're
flapping the condition back and forth between these two states. The
reconciles then fan out to all other controllers watching
ClusterDeployment.

Per latest kube api guidelines:

"Condition type names should describe the current observed state of the
resource, rather than describing the current state transitions. This
typically means that the name should be an adjective ("Ready",
"OutOfDisk") or a past-tense verb ("Succeeded", "Failed") rather than a
present-tense verb ("Deploying"). Intermediate states may be indicated
by setting the status of the condition to Unknown."

As such I have changed these conditions to only be set once we've
successfully cleaned up the resources. Users will not see "I am
deprovisioning" status, rather "I have deprovisioned" or "I have failed
to deprovision". This will hopefully correct the hotloop.

Per above the fix for the missing secret will be pursued in another PR.

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -250,14 +250,6 @@ func (r *ReconcileAWSPrivateLink) Reconcile(ctx context.Context, request reconci
 			logger.WithField("prevInfraID", *cp.Spec.PrevInfraID).
 				Info("cleaning up PrivateLink resources from previous attempt")
 
-			if err := r.setProgressCondition(cd, corev1.ConditionFalse,
-				"CleanupForProvisionReattempt",
-				"cleaning up resources from previous provision attempt so that next attempt can start",
-				logger); err != nil {
-				logger.WithError(err).Error("failed to update condition on cluster deployment")
-				return reconcile.Result{}, err
-			}
-
 			if err := r.cleanupPreviousProvisionAttempt(cd, cp, logger); err != nil {
 				logger.WithError(err).Error("error cleaning up PrivateLink resources for ClusterDeployment")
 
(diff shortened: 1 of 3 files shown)


pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -267,6 +259,15 @@ func (r *ReconcileAWSPrivateLink) Reconcile(ctx context.Context, request reconci
 				}
 				return reconcile.Result{}, err
 			}
+
+			if err := r.setProgressCondition(cd, corev1.ConditionFalse,
+				"PreviousAttemptCleanupComplete",
```
akhil-rane, line 264: `PreviousAttemptedCleanupComplete` may be?

dgoodwin: Seems a little off, that makes it sound like a previous attempted cleanup, rather than a cleanup for a previous attempted provision. 

What about ProvisionAttemptCleanupComplete?

akhil-rane: My bad. Lets stick with `PreviousAttemptCleanupComplete`

akhil-rane (commented):
/lgtm

dgoodwin:
Sorry missed a test reference to the function.

dgoodwin:
Bot got confused. Force merging now that CI is clean.

PR #1381, opened 2021-05-11
Initialize all cluster deployment conditions

suhanime:

xref: https://issues.redhat.com/browse/HIVE-1529

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -53,6 +53,13 @@ const (
 	defaultRequeueLater = 1 * time.Minute
 )
 
+// clusterDeploymentAWSPrivateLinkConditions are the cluster deployment conditions controlled by
+// AWS private link controller
+var clusterDeploymentAWSPrivateLinkConditions = []hivev1.ClusterDeploymentConditionType{
+	hivev1.AWSPrivateLinkFailedClusterDeploymentCondition,
+	hivev1.AWSPrivateLinkReadyClusterDeploymentCondition,
+}
+
 // Add creates a new AWSPrivateLink Controller and adds it to the Manager with default RBAC.
 // The Manager will set fields on the Controller and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
@@ -161,6 +168,18 @@ func (r *ReconcileAWSPrivateLink) Reconcile(ctx context.Context, request reconci
 		// Error reading the object - requeue the request.
 		logger.WithError(err).Error("error getting ClusterDeployment")
 		return reconcile.Result{}, err
+	}
+
+	// Initialize cluster deployment conditions if not present
+	newConditions := controllerutils.InitializeClusterDeploymentConditions(cd.Status.Conditions, clusterDeploymentAWSPrivateLinkConditions)
+	if len(newConditions) > len(cd.Status.Conditions) {
+		cd.Status.Conditions = newConditions
+		logger.Info("initializing AWS private link controller conditions")
+		if err := r.Status().Update(context.TODO(), cd); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
 	}
 
 	if cd.Spec.Platform.AWS == nil ||
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
@@ -32,6 +32,8 @@ import (
 	"github.com/openshift/hive/pkg/awsclient"
 	"github.com/openshift/hive/pkg/awsclient/mock"
 	"github.com/openshift/hive/pkg/constants"
+	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	testassert "github.com/openshift/hive/pkg/test/assert"
 	testcd "github.com/openshift/hive/pkg/test/clusterdeployment"
 	"github.com/openshift/hive/pkg/test/generic"
 )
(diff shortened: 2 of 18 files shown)


pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -163,6 +170,17 @@ func (r *ReconcileAWSPrivateLink) Reconcile(ctx context.Context, request reconci
...
 
+	// Initialize cluster deployment conditions if not present
+	newConditions := controllerutils.InitializeClusterDeploymentConditions(cd.Status.Conditions, clusterDeploymentAWSPrivateLinkConditions)
+	if len(newConditions) > len(cd.Status.Conditions) {
+		cd.Status.Conditions = newConditions
+		logger.Info("initializing AWS private link controller conditions")
+		if err := r.Status().Update(context.TODO(), cd); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+			return reconcile.Result{}, err
+		}
```
dgoodwin, line 181: Should we return here?

pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
```diff
@@ -1484,6 +1484,24 @@ users:
 				cd.Status.Conditions[i].LastProbeTime = metav1.Time{}
 				cd.Status.Conditions[i].LastTransitionTime = metav1.Time{}
 			}
+			if len(test.expectedConditions) < len(clusterDeploymentAWSPrivateLinkConditions) {
```
dgoodwin, line 1487: This logic feels like it might catch someone off guard in the future. If I'm reading this right we're saying that if the test has fewer expected conditions than the two we know should be there, make sure the two we know should be there are injected. However if the test author was expecting two different conditions, this wouldn't trigger. It's also just strange that we're passing in a set of expected conditions and then changing it under the covers on the test author. 

A couple options:

  1. Add the expected conditions in every test case.
  2. Change the assumption that we're testing the ENTIRE array of conditions to verifying that the expected conditions are present and set as we expect. Honestly this exercise does show how brittle these tests are, whereas arguably we should unit test that condition A we expect is in state X. Additional conditions could be ignored.

I could go either way, I think for a small simple controller like this I'd probably do option (1) as we're really just checking two conditions here. But in controllers with a large number of possible conditions, we may not want to define every condition for every test, and then (2) would make more sense.

pkg/controller/clusterrelocate/clusterrelocate_controller.go
```diff
@@ -167,6 +173,18 @@ func (r *ReconcileClusterRelocate) Reconcile(ctx context.Context, request reconc
...
+	// Initialize cluster deployment conditions if not present
+	newConditions := controllerutils.InitializeClusterDeploymentConditions(cd.Status.Conditions,
+		clusterDeploymentClusterRelocateConditions)
+	if len(newConditions) > len(cd.Status.Conditions) {
+		cd.Status.Conditions = newConditions
+		logger.Info("initializing cluster relocate controller conditions")
+		if err := r.Status().Update(context.TODO(), cd); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+			return reconcile.Result{}, err
+		}
```
dgoodwin, line 185: Suspect we want return here as well.

pkg/controller/controlplanecerts/controlplanecerts_controller.go
```diff
@@ -137,6 +143,17 @@ func (r *ReconcileControlPlaneCerts) Reconcile(ctx context.Context, request reco
...
 
+	// Initialize cluster deployment conditions if not present
+	newConditions := controllerutils.InitializeClusterDeploymentConditions(cd.Status.Conditions, clusterDeploymentControlPlaneCertsConditions)
+	if len(newConditions) > len(cd.Status.Conditions) {
+		cd.Status.Conditions = newConditions
+		cdLog.Info("initializing control plane certs controller conditions")
+		if err := r.Status().Update(context.TODO(), cd); err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+			return reconcile.Result{}, err
+		}
```
dgoodwin, line 154: return

pkg/controller/unreachable/unreachable_controller.go
```diff
@@ -251,12 +269,12 @@ func setActiveAPIURLOverrideCond(cd *hivev1.ClusterDeployment, connectionError e
...
-		// We want an explicit ActiveAPIURLOverride condition even when the condition is false so that the user
-		// can see the details of the connection error.
-		cd.Status.Conditions = append(cd.Status.Conditions, hivev1.ClusterDeploymentCondition{Type: hivev1.ActiveAPIURLOverrideCondition})
-	}
+	//if existingCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ActiveAPIURLOverrideCondition); existingCond == nil {
+	//	// This adds a dummy ActiveAPIURLOverride condition that will be updated when setting the condition later.
+	//	// We want an explicit ActiveAPIURLOverride condition even when the condition is false so that the user
+	//	// can see the details of the connection error.
+	//	cd.Status.Conditions = append(cd.Status.Conditions, hivev1.ClusterDeploymentCondition{Type: hivev1.ActiveAPIURLOverrideCondition})
+	//}
```
dgoodwin, line 277: Neat, no longer need this, looks like we can remove the comment.

dgoodwin (commented):
Good progress, thanks!

pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
suhanime: I kinda took bits and pieces of both suggestions and implemented them as fit.
Originally I moved the `assertConditionStatus` and my new implementation of `assertConditions` to `pkg\test\clusterdeployment\`, however I ended up creating a cyclic dependency via controller utils. So I have added them individually to the controllers for now. Let me know if you're okay with my current implementation and if there's an another place we could move those methods to without creating a cyclic dependency

dgoodwin (commented):
Let's squash commits as well plz.

suhanime:
@dgoodwin Where do you suggest would be the right place to initialize [InstallerImageResolutionFailedCondition](https://github.com/openshift/hive/blob/master/apis/hive/v1/clusterdeployment_types.go#L347)? I'm inclined to initialize it as a part of ClusterDeploymentController Reconcile...

dgoodwin:
Yeah that sounds fine.

dgoodwin (commented):
Looks pretty close to me, @abhinavdahiya can you take a look as well?

/hold

Would also like to merge this next week, as the last round exposed problems where people were just assuming presence of the condition implied true and not actually checking status. Worried no one will notice in stage over the weekend and merging next week would feel a little better with some soak time.

pkg/controller/remoteingress/remoteingress_controller_test.go
```diff
@@ -686,6 +768,20 @@ func validateSyncSet(t *testing.T, existingSyncSet createdSyncSetInfo, expectedS
 	return
 }
 
+// assertConditions asserts if the expected conditions are present on the cluster deployment.
+// It also asserts if those conditions have the expected status and reason
+func assertConditions(t *testing.T, cd *hivev1.ClusterDeployment, expectedConditions []hivev1.ClusterDeploymentCondition) {
```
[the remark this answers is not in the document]
suhanime: As I mentioned [here](https://github.com/openshift/hive/pull/1381#discussion_r631554727), there's an import cycle when I try to move that func to `pkg/test`. Is there any other location that would be appropriate?
`# github.com/openshift/hive/pkg/controller/utils
package github.com/openshift/hive/pkg/controller/utils (test)
        imports github.com/openshift/hive/pkg/test/assert
        imports github.com/openshift/hive/pkg/controller/utils: import cycle not allowed in test`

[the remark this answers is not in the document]
suhanime: copied the `findclusterdeploymentcondition` code to ensure import cycle doesn't happen - the function is basic enough to be used twice in the code base

PR #1384, opened 2021-05-12
allow to set zero replicas also when zones are set

joelddiaz:
in addition to allowing zero replicas w/o zones, also allow it when the
machinepool has a list of zones defined.

xref: https://issues.redhat.com/browse/HIVE-1486


dgoodwin:
Looks good to me, over to you @suhanime

PR #1385, opened 2021-05-12
Add metrics controller for hive-operator

akhil-rane:
Currently this controller only reports metrics for conditions
on HiveConfig. We can add more metrics in the future.

### DIFF
--- a/pkg/operator/metrics/metrics.go
+++ b/pkg/operator/metrics/metrics.go
@@ -1,0 +1,161 @@
+package metrics
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/hive/pkg/constants"
+)
+
+const (
+	ControllerName = hivev1.MetricsControllerName
+)
+
+var (
+	// Capture the various conditions set on the HiveConfig
+	metricHiveConfigConditions = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_hiveconfig_conditions",
+		Help: "HiveConfig with conditions.",
+	}, []string{"condition", "reason"})
+
+	// metricControllerReconcileTime tracks the length of time our reconcile loops take. controller-runtime
+	// technically tracks this for us, but due to bugs currently also includes time in the queue, which leads to
+	// extremely strange results. For now, track our own metric.
+	metricControllerReconcileTime = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_operator_reconcile_seconds",
+			Help:    "Distribution of the length of time each controllers reconcile loop takes.",
+			Buckets: []float64{0.001, 0.01, 0.1, 1, 10, 30, 60, 120},
+		},
+		[]string{"controller", "outcome"},
+	)
+)
+
+// ReconcileOutcome is used in controller "reconcile complete" log entries, and the metricControllerReconcileTime
+// above for controllers where we would like to monitor performance for different types of Reconcile outcomes. To help with
+// prometheus cardinality this set of outcomes should be kept small and only used for coarse and very high value
+// categories. Controllers must report an outcome but can report unspecified if this categorization is not needed.
+type ReconcileOutcome string
+
+const (
+	ReconcileOutcomeUnspecified        ReconcileOutcome = "unspecified"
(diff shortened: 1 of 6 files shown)


pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	ControllerName = hivev1.MetricsControllerName
+)
+
+var (
+	// Capture the various conditions set on the HiveConfig
+	metricHiveConfigConditions = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_config_conditions",
+		Help: "HiveConfig with conditions.",
+	}, []string{"condition", "reason"})
```
dgoodwin, line 29: I am legit not sure if reason should be here. How does this behave if the reason is changing? 
It doesn't feel like a cardinality problem.
I imagine the graph will just be different lines at value 1 over time, and the time series might change in the middle.

I *think* we can still do an alert that triggers if a condition is True for 10 minutes, even if the Reason changes mid-way right? We'd just be querying only on Condition right?

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	ControllerName = hivev1.MetricsControllerName
+)
+
+var (
+	// Capture the various conditions set on the HiveConfig
+	metricHiveConfigConditions = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_config_conditions",
```
dgoodwin, line 27: Nit: lets do hive_hiveconfig_conditions

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,161 @@
...
+		Help: "HiveConfig with conditions.",
+	}, []string{"condition", "reason"})
+
+	// metricControllerReconcileTime tracks the length of time our reconcile loops take. controller-runtime
+	// technically tracks this for us, but due to bugs currently also includes time in the queue, which leads to
+	// extremely strange results. For now, track our own metric.
+	metricControllerReconcileTime = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_controller_reconcile_seconds",
```
dgoodwin, line 36: Need a better name here, looks like the original one you worked off is pretty bad too, maybe hive_operator_metrics_seconds?

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (mc *Calculator) setHiveConfigMetrics(config *hivev1.HiveConfig) {
+	for _, condition := range config.Status.Conditions {
+		if condition.Status == corev1.ConditionTrue {
+			metricHiveConfigConditions.WithLabelValues(string(condition.Type), condition.Reason).Set(1)
+		} else {
+			metricHiveConfigConditions.WithLabelValues(string(condition.Type), condition.Reason).Set(0)
+		}
+	}
+}
```
dgoodwin, line 124: So we don't ever expect a Condition to disappear now I guess, if it did we'd just keep reporting the previous value until the pod was restarted. This is probably no longer a concern though.

pkg/operator/metrics/metrics.go
suhanime: It is the time for reconcile loop though - so reconcile in the name seems more appropriate to me - whereas metrics is a bit vague.
I do like the `operator` instead of controller though, since that's how we refer to it generally

pkg/operator/metrics/metrics.go
suhanime: I don't think hiveconfig is complicated enough to include the reason - although we could if we want. But yes - the value will change everytime the reason changes even if the status doesn't.

akhil-rane: I was thinking of having this as a table showing condition, reason and status (more on the lines of `up` metric) instead of time series graph. 

For alerting, yes we would check on condition and status, doesn't matter if reason changes midway.

pkg/operator/metrics/metrics.go
akhil-rane: I agree with @suhanime. I would like to go with `hive_operator_reconcile_seconds`

pkg/operator/metrics/metrics.go
akhil-rane: If we want to have this metric displayed as a table then may be we should clear the history and only report latest status per condition WDYT?

dgoodwin: if the conditions are always present, be then unknown true or false, then it should effectively clear itself as it is now. If we had conditions literally going away then we'd have a problem, but it looks like it should be ok as is to me.

pkg/operator/metrics/metrics.go
dgoodwin: Sounds good to me.

pkg/operator/metrics/metrics.go
dgoodwin: It's definitely useful in a table view, and if everything works ok for an alert across a timespan when we're potentially hitting multiple time series, then I'm ok with including reason.

pkg/operator/metrics/metrics.go
dgoodwin: (Clearing gets a little tricky as we'd need to zero out any possible combo of condition + reason to get rid of all the time series.)

akhil-rane:
Updated the PR

PR #1386, opened 2021-05-13
Fix stuck privatelink clusters deleted prior to install.

dgoodwin:
In some scenarios when an AWS privatelink cluster is deleted prior to
being fully installed, it was possible for the code to jam up in a
situation where no hosted zone ID was recorded in status, and we cannot
discover it because the admin kubeconfig secret is already gone.

To fix we just gracefully exit in this scenario as we have nothing to
cleanup.

x-ref: https://issues.redhat.com/browse/HIVE-1537

/assign @abhinavdahiya 
/cc @suhanime

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -1186,7 +1186,7 @@ func initialURL(c client.Client, key client.ObjectKey) (string, error) {
 		key,
 		kubeconfigSecret,
 	); err != nil {
-		return "", errors.Wrap(err, "could not get admin kubeconfig secret")
+		return "", err
 	}
 	cfg, err := restConfigFromSecret(kubeconfigSecret)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


dgoodwin:
Doesn't seem to be test coverage of the deleted cluster path, I could look to add next week.

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -1186,7 +1186,7 @@ func initialURL(c client.Client, key client.ObjectKey) (string, error) {
 		key,
 		kubeconfigSecret,
 	); err != nil {
-		return "", errors.Wrap(err, "could not get admin kubeconfig secret")
+		return "", err
```
abhinavdahiya, line 1189: the apierrors.IsNotFound supports wrapped errors, so i don't think we need to *not* wrap the error here.
https://github.com/kubernetes/apimachinery/blob/e337f44144a6ffeee2758c80dea17dd53fa6cbf4/pkg/api/errors/errors.go#L650
https://github.com/kubernetes/apimachinery/blob/e337f44144a6ffeee2758c80dea17dd53fa6cbf4/pkg/api/errors/errors.go#L480-L482

dgoodwin: Cool that was not what I would have assumed.

PR #1389, opened 2021-05-18
Update to aws-sdk-go 1.38.41.

dgoodwin:


dgoodwin:
/assign @abhinavdahiya 
/cc @suhanime 

This is a first attempt to fix the govcloud DNS issue, I don't think it will work but SRE seemed to think it may so we'll give it a shot while we investigate further.

PR #1390, opened 2021-05-18
verify release images used for cluster deployments

abhinavdahiya:
Currently during upgrades, OpenShift 4 checks that the release image
that is being requested in upgrade is trusted. By default it only trusts
release images that are published by Red Hat. This is very important
since release image is the definition of what runs in the cluster and
therefore must be verified before it is used.

Similarly, currently Hive uses the release image specified for the
cluster to extract binaries like oc, openshift-install and then executes
them in internal environment (Hive clusters), it is important that Hive
enforce that the content is trusted before running it.
The mechanism that is used by OpenShift during upgrades should be the
best way to establish and enforce that trust.

How verification works?

for details see https://github.com/openshift/cluster-update-keys

The definition of trust is defined in an ConfigMap data keys,

store-*:
These define the ways in which signature(a) for specific release image
digest can be fetched.
Multiple stores can be used to provide resilient source of signatures

verifier-public-key-*:
This is the public key that will be used to verify the signature and
release image are indeed trusted.
Multiple keys imply that the digest and signatures must be trusted by
every key.

OpenShift publishes this config map by default in each cluster at
openshift-config-managed/release-verification . So if Hive is running on
OpenSHift 4 and it intends to allow only release images published by Red
Hat, it can use this config map as definitions of trust.

The verification requires digest of the release image, the
repository or source of registry is not used for verification. Therefore
release images with tag cannot be verified using this method.

OpenShift publishes
https://pkg.go.dev/github.com/openshift/library-go/pkg/verify library
for working with trust definitions, stores and verification process. Hive
uses this library to perform the verification.

How Hive uses verification?

When verification is enabled (it is disabled by default),

Hive validation will reject ClusterImageSets that use release images
with tags.

Hive controllers will verify the release image before using the release
image for extracting any information.
In cases when the verification fails, Hive will set the
InstallImagesNotResolved condition to True with
ReleaseImageVerificationFailed reason.

How to configure Hive for verifying release images?

By default no trust for release images is defined and therefore on
verification occurs.

Verification is turned on by defining a trust using HiveConfig.

```
spec:
  releaseImageVerificationConfigMapRef:
    namespace: <> # namespace of the config map that defines the trust
    name: <> # name of the config map that defines the
```

xref: https://issues.redhat.com/browse/HIVE-1512


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1896,3 +1932,48 @@ func getClusterRegion(cd *hivev1.ClusterDeployment) string {
 	}
 	return regionUnknown
 }
+
+func LoadReleaseImageVerifier(config *rest.Config) (verify.Interface, error) {
+	ns := os.Getenv(constants.HiveReleaseImageVerificationConfigMapNamespaceEnvVar)
+	name := os.Getenv(constants.HiveReleaseImageVerificationConfigMapNameEnvVar)
```
dgoodwin, line 1938: There are nighty ClusterImageSets in all dedicated hive shards. Am I correct that these will not pass verification?

If so we may need a more complicated change, as implemented this looks to be a global setting. We might need to let users annotate their ClusterImageSets to disable, and coordinate with SD on how to get that done, and get hive config updated to enable.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -579,6 +597,24 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+	// let's verify the release image before using it here.
+	if r.releaseImageVerifier != nil {
+		var releaseDigest string
+		if index := strings.LastIndex(releaseImage, "@"); index != -1 {
+			releaseDigest = releaseImage[index+1:]
+		}
+		cdLog.WithField("releaseImage", releaseImage).
+			WithField("releaseDigest", releaseDigest).Debug("verifying the release image using %s", r.releaseImageVerifier)
+
+		err := r.releaseImageVerifier.Verify(context.TODO(), releaseDigest)
```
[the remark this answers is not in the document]
abhinavdahiya: The verification involves using fetching the signatures and then using the public key to verify that signature.
these are no compute intensive, most verification steps are not.

also the library caches the verified digests and should reduce repetition for our mostly common release images.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
abhinavdahiya: > There are nighty ClusterImageSets in all dedicated hive shards. Am I correct that these will not pass verification?

i can check, but afaik all nightlies published by ART are signed and published like any other release release-images.

> as implemented this looks to be a global setting. We might need to let users annotate their ClusterImageSets to disable

for verification to be useful, it kind of has to be global and any step to skip it by modifying the object that we are modifying seems counter productive.
if we needed a way to define exclusions, it should probably be in HiveConfig. (hopefully explicit list, because selector based api has same problem of annotation on the object being verified)

abhinavdahiya:
/assign @dgoodwin 
/cc @joelddiaz 

now has test and doc for the verification steps. PTAL

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Ok please check, if the nightlies are ok then we're good.

If not, ClusterImageSet creation is intended to be a way for management cluster admins to control what releases can be installed, so being able to disable at that level should be about the same privilege level as editing HiveConfig, but a lot easier for the appsre job the publishes those to implement.

abhinavdahiya: > If not, ClusterImageSet creation is intended to be a way for management cluster admins to control what releases can be installed, so being able to disable at that level should be about the same privilege level as editing HiveConfig, but a lot easier for the appsre job the publishes those to implement.

for me the people that are allowed to manage image set are separate from Hive (the service) . admins of cluster image sets are still users and not admin of Hive.
the admins of Hive service should be the one caring about not running malicious code.

the goal of release image verification is to verify the cluster image sets are good, and having things on that object that disable verification does not sound good.

abhinavdahiya: > ClusterImageSet was meant for the admins of the hive service to control what releases can be installed, but lets not worry about it until we know nightlies are not signed with the right key.

that's not great if that was the intention. 🤷🏽 

checked that nightly release images are signed with same trust as the released release-images

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+them in internal environment (Hive clusters), it is important that Hive
+enforce that the content is trusted before running it.
+The mechanism that is used by OpenShift during upgrades should be the
+best way to establish and enforce that trust.
+
+## How verification works?
+
+for details see https://github.com/openshift/cluster-update-keys
+
+The definition of trust is defined in an ConfigMap data keys,
```
dgoodwin, line 22:
```suggestion
The definition of trust is defined in ConfigMap data keys:
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+
+store-*:
+These define the ways in which signature(a) for specific release image
```
dgoodwin, line 25:
```suggestion
These define the ways in which signature(s) for specific release image
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+digest can be fetched.
```
dgoodwin, line 26:
```suggestion
digests can be fetched.
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+Multiple stores can be used to provide resilient source of signatures
+
+verifier-public-key-*:
+This is the public key that will be used to verify the signature and
+release image are indeed trusted.
+Multiple keys imply that the digest and signatures must be trusted by
+every key.
+
+OpenShift publishes this config map by default in each cluster at
+openshift-config-managed/release-verification . So if Hive is running on
```
dgoodwin, line 36:
```suggestion
openshift-config-managed/release-verification . If Hive is running on
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+OpenSHift 4 and it intends to allow only release images published by Red
```
dgoodwin, line 37:
```suggestion
OpenShift 4 and it intends to allow only release images published by Red
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+Hat, it can use this config map as definitions of trust.
+
+The verification requires digest of the release image, the
+repository or source of registry is not used for verification. Therefore
+release images with tag cannot be verified using this method.
```
dgoodwin, line 42: Looks like OSD is already using digests so that's a win.

@jnpacker could you confirm if ACM is populating ClusterImageSets with tags or digests?

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+
+OpenShift publishes [library][release-image-verify-lib] for working with trust
```
dgoodwin, line 44: Neat, didn't know you could do this with markdown.

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+the verification.
+
+## How Hive uses verification?
+
+When verification is enabled (it is disabled by default),
+
+Hive validation will reject ClusterImageSets that use release images
+with tags.
+
+Hive controllers will verify the release image before using the release
```
dgoodwin, line 55:
```suggestion
Hive controllers will verify the release image before
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+image for extracting any information.
```
dgoodwin, line 56:
```suggestion
extracting any information.
```

docs/releaseimageverify.md
```diff
@@ -0,0 +1,75 @@
...
+In cases when the verification fails, Hive will set the
+InstallImagesNotResolved condition to True with
+ReleaseImageVerificationFailed reason.
+
+How to configure Hive for verifying release images?
+
+By default no trust for release images is defined and therefore on
```
dgoodwin, line 63:
```suggestion
By default no trust for release images is defined and therefore no
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -225,6 +239,10 @@ type ReconcileClusterDeployment struct {
 	// that the platform creds are good (used for testing)
 	validateCredentialsForClusterDeployment func(client.Client, *hivev1.ClusterDeployment, log.FieldLogger) (bool, error)
 
+	// releaseImageVerifier, if provided, will be used to check an release image before it is executed.
+	// Any error will prevent an release image from being accessed.
```
dgoodwin, line 243:
```suggestion
	// Any error will prevent a release image from being accessed.
```

dgoodwin (commented):
Looks great just some text suggestions really.

PR #1391, opened 2021-05-19
Added to controller check for clusters with noAlerts label

luis-falcon:


dgoodwin:
PR really needs some justification as to the thought process behind this and how it works.

Hive is not a managed openshift project and we'd like to avoid managed openshift specifics when possible.

Feels like there should be a better way to silence alerts in the alerting stack rather than going to the metrics?

luis-falcon:
> PR really needs some justification as to the thought process behind this and how it works.
> 
> Hive is not a managed openshift project and we'd like to avoid managed openshift specifics when possible.
> 
> Feels like there should be a better way to silence alerts in the alerting stack rather than going to the metrics?

This is not to silence alerts. we already doing that. this is to get a metric from which we can check cluster deployments that have had alerts silenced for too long and alert a member of SREP.
https://issues.redhat.com/browse/OSD-6261

gregsheremeta:
Is there some way to accomplish this without adding OSD-specific code?

luis-falcon:
@dofinn  Since you wrote this card. Pinging you. Thoughts on this.

fahlmant:
I think the alternative is for us to write an exporter that we run on hive clusters to get this metric.

luis-falcon:
Ok. Apparently the noalert label is OSD specific. So no point to have this code on hive. Closing this PR.

PR #1393, opened 2021-05-24
controller/dnszone: rate limit UPDATE events that are in error state

abhinavdahiya:
#### controller/utils: add event handler wrapper that force rate limits update event

Event handlers in controller-runtime add Update events to non-rate limited queues.
In many cases it is desireable that the Update event be rate limited for
preventing excessive reconciles.

One of the main examples of the use case is when controller adds Failure
condition to the objects. This causes Update event to be trigerred and
immediately handled for the object and almost always ending up in the
same Failure state. The effect is increased if the error message in the
condition changes every time error is enconterred like request ids in
error messages returned from cloud providers.
To make sure the controller can backoff these error states, a new event
handler wrapper is added that can force the UPDATE event to be rate
limited when the controller configured function returns true. Allowing
controllers to write code that can recognize error transitions/states
and rate limit such events.

#### controller/dnszone: rate limit UPDATE events that are in error state

When the controller receives an UPDATE event when the object in in error
state i.e. conditions like InsufficientCredentialsCondition,
AuthenticationFailureCondition are True, add that reconcile request to
the rate limited queue of the reconciler.

xref: https://issues.redhat.com/browse/HIVE-1542

/cc @dgoodwin @joelddiaz 
/assign @suhanime

### DIFF
--- a/pkg/controller/dnszone/dnszone_controller.go
+++ b/pkg/controller/dnszone/dnszone_controller.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -97,7 +98,7 @@ func add(mgr manager.Manager, r *ReconcileDNSZone, concurrentReconciles int, rat
 	}
 
 	// Watch for changes to DNSZone
-	if err := c.Watch(&source.Kind{Type: &hivev1.DNSZone{}}, &handler.EnqueueRequestForObject{}); err != nil {
+	if err := c.Watch(&source.Kind{Type: &hivev1.DNSZone{}}, controllerutils.NewRateLimitedUpdateEventHandler(&handler.EnqueueRequestForObject{}, IsErrorUpdateEvent)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


abhinavdahiya:
doing some testing to see how this reacts in situation like https://issues.redhat.com/browse/HIVE-1542

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -480,3 +481,44 @@ func lookupSOARecord(zone string, logger log.FieldLogger) (bool, error) {
...
+			}
+			if cn.Message != co.Message ||
+				cn.Reason != co.Reason {
+				return true // already failing but change in error reported
+			}
+		}
+	}
+
+	return false
+}
```
dgoodwin, line 524: This is great. Feels like there's potential here where we could just pass our error conditions to controllerutils.NewRateLimitedUpdateEventHandler maybe in conjunction with https://issues.redhat.com/browse/HIVE-1521 in the future.

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -480,3 +481,44 @@ func lookupSOARecord(zone string, logger log.FieldLogger) (bool, error) {
...
+			if co == nil {
+				return true // newly added failure condition
+			}
+			if co.Status != corev1.ConditionTrue {
+				return true // newly Failed failure condition
+			}
+			if cn.Message != co.Message ||
+				cn.Reason != co.Reason {
+				return true // already failing but change in error reported
+			}
```
dgoodwin, line 519: Any value in adding a check on only the timestamps changing? Pretty common beginner mistake in controllers where people assume it's ok to update the lastProbed timestamp every time they reconcile. On the other hand it might be best to let that hotloop so people can see and fix the problem. :)

abhinavdahiya: > Any value in adding a check on only the timestamps changing?

for now, the intention is to rate limit only when error transitions in conditions, and timestamps don;t seem to fit that.. maybe if we want to later on add more cases where we want to rate limit ..

suhanime (commented):
/lgtm

PR #1394, opened 2021-05-24
Metrics change to accomodate clusterdeployment conditions cleanup

suhanime:
We have always assumed that all cluster deployment conditions have negative polarity, i.e., they are in their undesired state if their status=True.
This PR removes that assumption, and logs `hive_cluster_deployments_conditions` for all conditions in their undesired state.
For cluster deployment conditions in  `PositivePolarityClusterDeploymentConditions`, we assume that their desired state is True, and for all other conditions, the desired state is False..
Conditions with Unknown state are ignored.

xref: https://issues.redhat.com/browse/HIVE-1549

/assign @akhil-rane 
/cc @dgoodwin @abhinavdahiya

### DIFF
--- a/pkg/controller/utils/conditions.go
+++ b/pkg/controller/utils/conditions.go
@@ -703,3 +703,13 @@ func FindClusterInstallCondition(conditions []hivev1.ClusterInstallCondition, co
 	}
 	return nil
 }
+
+// IsConditionWithPositivePolarity checks if cluster deployment condition has positive polarity
+func IsConditionWithPositivePolarity(conditionType hivev1.ClusterDeploymentConditionType) bool {
+	for _, condition := range hivev1.PositivePolarityClusterDeploymentConditions {
+		if condition == conditionType {
+			return true
+		}
+	}
+	return false
+}
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -336,7 +336,6 @@ type ClusterDeploymentCondition struct {
 // ClusterDeploymentConditionType is a valid value for ClusterDeploymentCondition.Type
 type ClusterDeploymentConditionType string
 
-// WARNING: All ClusterDeploymentConditionTypes should be added to the AllClusterDeploymentConditions slice below.
 const (
 	// ClusterImageSetNotFoundCondition is set when the ClusterImageSet referenced by the
 	// ClusterDeployment is not found.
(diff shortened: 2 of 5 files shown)


dgoodwin:
Seems ok to me. I do worry we'll still forget to add conditions to that list, wish there was some way to more closely link the polarity to the actual condition type so it had to be defined, or at least would be more likely to be defined.

suhanime:
> I do worry we'll still forget to add conditions to that list, wish there was some way to more closely link the polarity to the actual condition type so it had to be defined, or at least would be more likely to be defined.

I wish things were that black and white... For ex, we can come up with a regex to exclude conditions with *failed, *error, *notfound in them, but we will exclude conditions like unreachable then.
Even the polarity of all conditions are not clear, for instance, 'hibernating' condition doesn't have its error state always tied to the status - for ex, cluster with hibernating false might just mean it is running fine, or it could mean hibernation is not supported or failed to hibernate - all depends on the reason+message.

dgoodwin:
> > I do worry we'll still forget to add conditions to that list, wish there was some way to more closely link the polarity to the actual condition type so it had to be defined, or at least would be more likely to be defined.
> 
> I wish things were that black and white... For ex, we can come up with a regex to exclude conditions with *failed, *error, *notfound in them, but we will exclude conditions like unreachable or ProvisionStopped then.
> Even the polarity of all conditions are not clear, for instance, 'hibernating' condition doesn't have its error state always tied to the status - for ex, cluster with hibernating false might just mean it is running fine, or it could mean hibernation is not supported or failed to hibernate - all depends on the reason+message.

I was actually wondering about using a struct for ClusterDeploymentConditionType, instead of just having it be a string. The struct would just contain a string and a polarity, and then to define a condition type, you see that there are other things you need to set. We'd need it to serialize the same though (just with the string), which is probably easy.

dgoodwin:
Probably not going to work, still need a list of all conditions to iterate, probably best to ignore me :)

pkg/controller/utils/conditions.go
```diff
@@ -703,3 +703,13 @@ func FindClusterInstallCondition(conditions []hivev1.ClusterInstallCondition, co
 	}
 	return nil
 }
+
+// ConditionWithPositivePolarity checks if cluster deployment condition has positive polarity
+func ConditionWithPositivePolarity(conditionType hivev1.ClusterDeploymentConditionType) bool {
```
abhinavdahiya, line 708: it reads better if the check functions start with `Is` .. so `IsConditionWithPositivePolarity`

suhanime: changed

akhil-rane (commented):
/lgtm

PR #1395, opened 2021-05-25
Move to v1 CRD definitions.

dgoodwin:
As of this commit, Hive will no longer run on OpenShift v3.

Tested myself and pre-existing data appears to be unaffected. ACM team reports success with just flipping these over as well. I've have also spoken to the OLM team and they confirm that OLM should roll this out smoothly.

x-ref: https://issues.redhat.com/browse/HIVE-1514


gregsheremeta:
/hold

need to pin internal stage pipeline

dgoodwin:
stage/int v3 shards should now be pinned.

dgoodwin:
Integration test should pass once #1397 merges, still trying to find out why the e2e is failing on syncset validation.

dgoodwin:
Still not sure what's up with e2e, but I merged a PR in release repo to move us to base on openshift 4.7 instead of 4.6, will see if this helps.

I've taken the code from the test and run it against a regular cluster and the syncset creates fine, still not sure what's going on here.

dgoodwin:
Adding debug logging for another e2e run as I cannot repro this locally yet.

dgoodwin:
In case anyone is watching, for some reason with this change the syncset e2e is now failing to create this syncset: https://github.com/openshift/hive/blob/master/test/e2e/postinstall/syncsets/syncsets_suite_test.go#L60

The error is:

```
admission webhook "syncsetvalidators.admission.hive.openshift.io" denied the request: SyncSet.hive.openshift.io "test-syncresource" is invalid: spec.resources[0]: Invalid value: []byte{0x7b, 0x7d}: Unable to unmarshal resource
```

Those two bytes appear to be {}

I've been able to run this go in a hacked up CLI to create the same resource, and partially run the e2e, both against a 4.8 dev cluster using the v1 CRDs and it was fine.

CRD definition for the resources field has not changed in this PR as far as I can see.

dgoodwin:
I was able to reproduce the problem locally when I switched to a 4.7 dev cluster, the problem did not occur on 4.8. 

Traced it down to an issue with preserve unknown fields, once this was added to the resources (type RawExtension), Kube began serializing correctly. Expecting next e2e run to pass after this CI problem resolves.

dgoodwin:
It appears to have made it through CI but more changes were necessary. Errors started showing up that I think were related to bugs in past Kube versions around the resources field. I had to mark that preserve unknown and hack it out in the crd generation make target as we did with the podspec on clusterprovision.

dgoodwin:
The error we get if we leave the resources in the CRD definition surfaces in the test-integration suite, which is currently using kube 1.19.2. 

```
 INFO[2021-06-02T13:16:50Z] --- FAIL: TestInfo (0.11s)                   
INFO[2021-06-02T13:16:50Z]     info_test.go:25:                         
INFO[2021-06-02T13:16:50Z]         	Error Trace:	info_test.go:25        
INFO[2021-06-02T13:16:50Z]         	Error:      	Received unexpected error: 
INFO[2021-06-02T13:16:50Z]         	            	SchemaError(io.openshift.hive.v1.SelectorSyncSet.spec.resources): array should have exactly one sub-item 
INFO[2021-06-02T13:16:50Z]         	            	error getting OpenAPISchema 
INFO[2021-06-02T13:16:50Z]         	            		/usr/local/go/src/runtime/asm_amd64.s:1374 
INFO[2021-06-02T13:16:50Z]         	Test:       	TestInfo               
```

I could reproduce this locally, but when I updated my kubebuilder tools to use version 1.20.2, the problem goes away.

I'm going to do a PR to switch our build docker image to use that kubebuilder tools version as it more closely matches 4.7 anyhow, and will test manually how this behaves with hive on ocp 4.6.

dgoodwin:
Hive seems ok on latest 4.6 in my testing, thus far this is only surfacing in integration tests using kubebuilder envtest tools at ver 1.19.2. #1403 will move us to 1.20.2 and the integration test should pass here once that PR merges.

dgoodwin:
/hold cancel

Ready for review.

/cc @2uasimojo

dgoodwin:
1.20.2 kubebuilder envtest tools won't run :( I think because our controller-runtime version is too old. None of the options are looking good at this point, 

1. upgrade to controller-runtime 0.8.x, including Suhani's patch we carry, no idea how difficult this will be
1. remove the integration test suite entirely, it's of very limited use, and stick with our e2e
1. patch the syncset resources field out of the CRD definition and go back to the 1.19.2 kubebuilder envtest tools

dgoodwin:
The commit we need to run integration tests against 1.20 is coming in controller-runtime 0.9.x which is not yet released. We have disabled integration test suite in CI and will track restoring it in https://issues.redhat.com/browse/HIVE-1560

Makefile
```diff
@@ -1,11 +1,6 @@
 .PHONY: all
 all: vendor update test build
 
-# Force build-machinery-go to grab an earlier version of controller-gen. The newer version adds annotations on pod
-# specs that will cause the CRDs to fail validation. The fix for that requires the CRDs to use v1, but we still need
-# to support users that are running kube versions that only have v1beta1.
-CONTROLLER_GEN_VERSION ?=v0.2.1-37-ga3cca5d
```
2uasimojo, line 7: Note to self: `make clean-controller-gen` to purge the old version.

akhil-rane (commented):
looks good to me

apis/hive/v1/clusterprovision_types.go
```diff
@@ -124,6 +124,7 @@ type ClusterProvision struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// +kubebuilder:pruning:PreserveUnknownFields
```
2uasimojo, line 127: This is necessary due to a bug in 4.7. Follow HIVE-1561 for getting rid of it.

apis/hive/v1/syncset_types.go
```diff
@@ -206,6 +206,7 @@ type SyncStatus struct {
 type SyncSetCommonSpec struct {
 	// Resources is the list of objects to sync from RawExtension definitions.
 	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
```
2uasimojo, line 209: This is necessary due to a bug in 4.7. Follow HIVE-1561 for getting rid of it.

2uasimojo (commented):
I downloaded this PR and ran the `crd` and `vendor` targets to make sure everything was regenerated. Once I had cleaned controller-gen and yq, I got a couple of extra lines removed from go.sum but otherwise clean. So the substance of this change is just in the `make crd` rule, and that

/lgtm

dgoodwin:
/hold

@sttts pointed out that we need an embedded resource flag set and then we can avoid the preserve on delete in https://bugzilla.redhat.com/show_bug.cgi?id=1969418

2uasimojo:
> avoid the preserve on delete

Wohoo!

/lgtm cancel

dgoodwin:
Still having some issues, adding the controller-runtime flag kubebuilder:validation:EmbeddedResource to the Resource array puts the flag in the wrong place, on the array, not the items within it.

I hacked it in with yq but then:

The CustomResourceDefinition "selectorsyncsets.hive.openshift.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[resources].items.properties: Required value: must not be empty if x-kubernetes-embedded-resource is true without x-kubernetes-preserve-unknown-fields

So we need properties in there, or we still need preserve unknown fields. Trying to figure out how to do that.

dgoodwin:
Updated pushed, I don't think we can get away from the preserve unknown as far as I can tell.

2uasimojo:
Looks like a `make vendor` is needed.

dgoodwin:
CI is clear

config/crds/hive.openshift.io_selectorsyncsets.yaml
```diff
@@ -13,187 +15,192 @@ spec:
...
+                description: ResourceApplyMode indicates if the Resource apply mode
+                  is "Upsert" (default) or "Sync". ApplyMode "Upsert" indicates create
+                  and update. ApplyMode "Sync" indicates create, update and delete.
+                type: string
+              resources:
+                description: Resources is the list of objects to sync from RawExtension
+                  definitions.
+                items:
+                  type: object
+                  x-kubernetes-embedded-resource: true
```
2uasimojo, line 148: nts: here

config/crds/hive.openshift.io_syncsets.yaml
```diff
@@ -13,159 +15,163 @@ spec:
...
+                description: ResourceApplyMode indicates if the Resource apply mode
+                  is "Upsert" (default) or "Sync". ApplyMode "Upsert" indicates create
+                  and update. ApplyMode "Sync" indicates create, update and delete.
+                type: string
+              resources:
+                description: Resources is the list of objects to sync from RawExtension
+                  definitions.
+                items:
+                  type: object
+                  x-kubernetes-embedded-resource: true
```
2uasimojo, line 117: And here

2uasimojo (approved):
I’ve asked to poke at the embedded resource thing. But that wouldn’t result in an API change, and this is working, so no sense holding it up while I do that. 

/lgtm

config/crds/hive.openshift.io_selectorsyncsets.yaml
dgoodwin: what is nts :)

2uasimojo: “Note to self” ;)

dgoodwin:
Thanks for lgtm'ing! I'm nervous about this one and with Israel out tomorrow, I think I'll hold until next week master opens so it has the week to soak and expose any issues.

/hold

chrisahl:
@dgoodwin Do you have an ETA on this getting merged and a new release being available?

dgoodwin:
See comment before yours.

dgoodwin:
Rebased, @2uasimojo could I get another lgtm here.

2uasimojo (commented):
/lgtm

2uasimojo:
Weird, that needs-rebase label isn't coming off. I've verified that you're on top of latest master.

dgoodwin:
Removed manually, what do you think of that robot?

2uasimojo:
Robot: "You're still on `/hold`, so nyah."

PR #1396, opened 2021-05-25
fix support of OpenStack autoscalling.minRep=0

joelddiaz:
The semver library returns an error if you try to parse a version with
only major.minor. Provide the semver package the full major.minor.patch
to allow properly detecting whether a version of OpenStack can support
autocaling.minReplicas=0.

xref: https://issues.redhat.com/browse/HIVE-1539


dgoodwin:
LGTM

PR #1398, opened 2021-05-26
Sort clusterdeployment conditions

suhanime:
Now that we have all possible cluster deployment conditions present on
the CR, it can become very difficult to debug.
This commit tries to sort the list of conditions, such that the
conditions in their undesired state are always at the top of the list

xref: https://issues.redhat.com/browse/HIVE-1552

### DIFF
--- a/pkg/controller/clusterdeployment/clusterinstalls.go
+++ b/pkg/controller/clusterdeployment/clusterinstalls.go
@@ -151,6 +151,7 @@ func (r *ReconcileClusterDeployment) reconcileExistingInstallingClusterInstall(c
 		statusModified = true
 	}
 
+	completed = controllerutils.FindClusterDeploymentCondition(conditions, hivev1.ClusterInstallCompletedClusterDeploymentCondition)
 	if completed.Status == corev1.ConditionTrue { // the cluster install is complete
 		cd.Spec.Installed = true
 		cd.Status.InstalledTimestamp = &completed.LastTransitionTime
--- a/pkg/controller/utils/conditions.go
+++ b/pkg/controller/utils/conditions.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"sort"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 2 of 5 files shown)


dgoodwin:
Because addClusterDeploymentCondition is explicitly for adding new conditions that are not already present, and because we're often now initializing to unknown first, it seems like this may not help a lot. I would think you want to hook in where we update conditions, check the polarity, and then sort the current list such that the "unsual" ones are at the top. I would make a couple slices for high and low priority, sort them both and then combine them.

Last transition sorting is interesting, keep the latest at the top. Might be a little risk there of a hotloop but it seems ok if we do our jobs right, and more useful. An alternative would be the condition type.

pkg/controller/utils/conditions.go
```diff
@@ -124,9 +126,36 @@ func SetClusterDeploymentConditionWithChangeCheck(
...
+			undesired = append(undesired, condition)
+		} else {
+			desired = append(desired, condition)
+		}
+	}
+
+	for _, slice := range [][]hivev1.ClusterDeploymentCondition{unknown, desired, undesired} {
+		sort.SliceStable(slice, func(i, j int) bool {
+			return slice[i].Type < slice[j].Type
+		})
```
dgoodwin, line 154: I believe you can remove sortClusterDeploymentConditions function and the need for three slices to be combined, and instead just use the SliceStable sort where we keep falling back to more specific checks for sorting. In that sort function we can do several checks to determine if I should be sorted before J.

I don't know if my true/falses are right here but hopefully this clarifies the idea:

    if J is Unknown and I isn't, return true
    if I is Unkown and J isn't, return false
    if I is out of polarity state and J isn't, return true
    if J is out of polarity state and I isn't, return false
    lastly do the Type string comparison if none of the above was true

I might be wrong but I'd be surprised if string < otherString is doing what you want? Feels like you need strings.Compare().

Definitely want unit tests on SetClusterDeploymentConditionWithChangeCheck to verify they're being sorted the way you expect.

akhil-rane: As per SliceStable example [here](https://golang.org/pkg/sort/#SliceStable) string < otherString should work if we are sorting alphabetically.

+1 for unit tests here.

pkg/controller/clusterdeployment/clusterinstalls.go
```diff
@@ -151,6 +151,7 @@ func (r *ReconcileClusterDeployment) reconcileExistingInstallingClusterInstall(c
 		statusModified = true
 	}
 
+	completed = controllerutils.FindClusterDeploymentCondition(conditions, hivev1.ClusterInstallCompletedClusterDeploymentCondition)
```
dgoodwin, line 154: I can't quite tell, is there a reason why we're finding this condition again when it was done a few lines above? Or was this a mistaken paste?

pkg/controller/utils/conditions.go
```diff
@@ -124,9 +126,34 @@ func SetClusterDeploymentConditionWithChangeCheck(
...
+		}
+		if !IsConditionInDesiredState(conditions[i]) && IsConditionInDesiredState(conditions[j]) {
+			return true
+		}
+		if IsConditionInDesiredState(conditions[i]) && !IsConditionInDesiredState(conditions[j]) {
+			return false
+		}
+		return conditions[i].Type < conditions[j].Type
+	})
+	return conditions
```
dgoodwin, line 154: Much better! Thanks.

pkg/controller/clusterdeployment/clusterinstalls.go
suhanime: Since `FindClusterDeploymentCondition` returns a pointer to the condition in question from the list of conditions, when we update the list of conditions (the list is updated right above this code - it is setting a condition that in turn also sorts it), the pointer changes - so we need to re-fetch the accurate pointer. Caught by a failing unit-test

dgoodwin: Wow, would never have thought. Thanks!

dgoodwin:
/lgtm

I will check in tomorrow and freeze master.

pkg/controller/clusterdeployment/clusterinstalls.go
abhinavdahiya, line 154: why are we re finding the completed?

abhinavdahiya: can we return a deepcopy or something ?? this is not great to remember(hey after update your previous reads might be invalid...)?

PR #1400, opened 2021-05-27
controller/dnszone: rate limit update events from cd cause by dns error

abhinavdahiya:
When DNS zone ends up in error state, the clusterdeployment controller
copies that error to the clusterdeployment status. This causes update on
the clusterdeployment object, in turn creating an update event for dns
zone created for the clusterdeployment. since watch from
clusterdeployments are not rate limited, this creates exceesive
reconciles.

To make sure the dnszone controller limits updates from dns error, we
must also rate limit dns error from clusterdeployment updates. so now
the controller uses the DNSNotReadyCondition to check if error
transitions is because of dns error and then rate limit it.

xref: https://issues.redhat.com/browse/HIVE-1553

/assign @akhil-rane 
/cc @dgoodwin @staebler

### DIFF
--- a/pkg/controller/dnszone/dnszone_controller.go
+++ b/pkg/controller/dnszone/dnszone_controller.go
@@ -105,7 +105,10 @@ func add(mgr manager.Manager, r *ReconcileDNSZone, concurrentReconciles int, rat
 	// Watch for changes to ClusterDeployment
 	if err := c.Watch(
 		&source.Kind{Type: &hivev1.ClusterDeployment{}},
-		controllerutils.EnqueueDNSZonesOwnedByClusterDeployment(r, r.logger),
+		controllerutils.NewRateLimitedUpdateEventHandler(
+			controllerutils.EnqueueDNSZonesOwnedByClusterDeployment(r, r.logger),
+			IsClusterDeploymentDNSErrorUpdateEvent,
+		),
 	); err != nil {
 		return err
 	}
@@ -482,7 +485,7 @@ func lookupSOARecord(zone string, logger log.FieldLogger) (bool, error) {
 	return false, nil
 }
 
-// IsErrorUpdateEvent returns true when the update event is from
+// IsErrorUpdateEvent returns true when the update event for DNSZone is from
 // error state.
 func IsErrorUpdateEvent(evt event.UpdateEvent) bool {
 	new, ok := evt.ObjectNew.(*hivev1.DNSZone)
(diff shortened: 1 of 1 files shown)


staebler (commented):
This approach seems a bit fragile to me. An update with changes to the DNSNotReadyCondition could also include other updates as well. This can even happen when the updates were made separately to the API server. In this case, the controller will ignore the update, even though it should not ignore it. Can we instead compare the rest of the fields to see if there are other changes as well?

abhinavdahiya:
> In this case, the controller will ignore the update, even though it should not ignore it. Can we instead compare the rest of the fields to see if there are other changes as well?

the updates are not getting ignores, they are just getting put in rate limited queue... compared to immediately being put in queue for reconcile.
as long as the update includes a error transition in dns ready condition.. i think it's safe to say lets make sure we reconcile it only when we have budget for it. the rate limiteing is already setup to make sure we are quick initially but take it slowly if we keep seeing the error with that object.

staebler:
> > In this case, the controller will ignore the update, even though it should not ignore it. Can we instead compare the rest of the fields to see if there are other changes as well?
> 
> the updates are not getting ignored, they are just getting put in rate limited queue... compared to immediately being put in queue for reconcile.
> as long as the update includes a error transition in dns ready condition.. i think it's safe to say lets make sure we reconcile it only when we have budget for it. the rate limiteing is already setup to make sure we are quick initially but take it slowly if we keep seeing the error with that object.

Fair enough. The assumption here is that, if DNSNotReadyCondition is true, then there is likely nothing that would have changed in the DNSZone to warrant checking again until the rate-limited timeout.

abhinavdahiya:
ping @akhil-rane @dgoodwin  for review.

dgoodwin:
LGTM. Over to you @akhil-rane

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -522,3 +525,37 @@ func IsErrorUpdateEvent(evt event.UpdateEvent) bool {
...
+	if !ok {
+		return false
+	}
+
+	cn := controllerutils.FindClusterDeploymentCondition(new.Status.Conditions, hivev1.DNSNotReadyCondition)
+	if cn != nil && cn.Status == corev1.ConditionTrue {
+		co := controllerutils.FindClusterDeploymentCondition(old.Status.Conditions, hivev1.DNSNotReadyCondition)
+		if co == nil {
+			return true // newly added failure condition
+		}
```
akhil-rane, line 550: is this check required since we are initializing the ClusterDeployment with all the conditions now?

abhinavdahiya: update events include the old and new object, and therefore there can be a case where old has no condition set.

PR #1402, opened 2021-06-02
Move to kube 1.20.2 for integration tests.

dgoodwin:


dgoodwin:
I moved this to 1.19 recently, but 1.20 is required for another bug found in the integration tests, and matches what we ship in ocp 4.7.

PR #1403, opened 2021-06-02
Validate region in InstallConfig matches ClusterDeployment.

dgoodwin:
In general we try not to fully validate install configs match what you
set in ClusterDeployment. In this particular case however, a region
mismatch will cause the cluster to keep provisioning machines in the
wrong region until your cloud quota is fully exhausted.

Add a minimal check to ensure the InstallConfig region matches the
cluster deployment for AWS, Azure, and GCP.

In the event of a problem, a ReadyToProvision condition on the
ClusterDeployment will go False. Hoping to use this condition to
consolidate a few others in this controller that are failing pre-flight
checks.

x-ref: https://issues.redhat.com/browse/HIVE-1337

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -67,6 +67,7 @@ var (
 		hivev1.DeprovisionLaunchErrorCondition,
 		hivev1.ProvisionStoppedCondition,
 		hivev1.AuthenticationFailureClusterDeploymentCondition,
+		hivev1.RequirementsMetCondition,
 
 		// ClusterInstall conditions copied over to cluster deployment
 		hivev1.ClusterInstallFailedClusterDeploymentCondition,
@@ -690,6 +691,75 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 
 	switch {
 	case cd.Spec.Provisioning != nil:
+		// Ensure the install config matches the ClusterDeployment:
+		// TODO: Move with the openshift-installer ClusterInstall code when we implement https://issues.redhat.com/browse/HIVE-1522
+		if cd.Spec.Provisioning.InstallConfigSecretRef == nil {
+			cdLog.Info("install config not specified")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.RequirementsMetCondition,
+				corev1.ConditionFalse,
+				"InstallConfigRefNotSet",
+				"Install config reference is not set",
+				controllerutils.UpdateConditionIfReasonOrMessageChange)
+			if changed {
+				cd.Status.Conditions = conditions
+				return reconcile.Result{}, r.Status().Update(context.TODO(), cd)
+			}
+			return reconcile.Result{}, nil
+		}
+
+		icSecret := &corev1.Secret{}
+		err = r.Get(context.Background(),
+			types.NamespacedName{
+				Namespace: cd.Namespace,
+				Name:      cd.Spec.Provisioning.InstallConfigSecretRef.Name,
+			},
+			icSecret)
+		if err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "Error loading install config secret")
+			return reconcile.Result{}, err
+		}
+
+		err = ValidateInstallConfig(cd, icSecret.Data["install-config.yaml"])
+		if err != nil {
+			cdLog.WithError(err).Info("install config validation failed")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.RequirementsMetCondition,
+				corev1.ConditionFalse,
+				"InstallConfigValidationFailed",
+				err.Error(),
+				controllerutils.UpdateConditionIfReasonOrMessageChange)
+			if changed {
+				cd.Status.Conditions = conditions
+				return reconcile.Result{}, r.Status().Update(context.TODO(), cd)
+			}
+			return reconcile.Result{}, nil
+		}
+
+		// If we made it this far, RequirementsMet condition should be True:
+		//
+		// TODO: when https://github.com/openshift/hive/pull/1413 is implemented
+		// we'll want to remove this assumption as ClusterInstall implementations may
+		// later indicate their requirements are not met. Instead, we should explicitly clear
+		// the condition back to Unknown if we see our Reason set, but the problem is no longer
+		// present.
+		conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+			cd.Status.Conditions,
+			hivev1.RequirementsMetCondition,
+			corev1.ConditionTrue,
+			"AllRequirementsMet",
+			"All pre-provision requirements met",
+			controllerutils.UpdateConditionIfReasonOrMessageChange)
+		if changed {
+			cd.Status.Conditions = conditions
+			err = r.Status().Update(context.TODO(), cd)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
 		return r.reconcileInstallingClusterProvision(cd, releaseImage, cdLog)
 	case cd.Spec.ClusterInstallRef != nil:
 		return r.reconcileInstallingClusterInstall(cd, cdLog)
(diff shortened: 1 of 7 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -686,6 +687,48 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+		cd.Status.Conditions,
+		hivev1.ReadyToProvisionCondition,
+		corev1.ConditionTrue,
+		"ProvisionChecksPassed",
+		"Ready to provision",
+		controllerutils.UpdateConditionIfReasonOrMessageChange)
+	if changed {
+		cd.Status.Conditions = conditions
+		return reconcile.Result{}, r.Status().Update(context.TODO(), cd)
+	}
```
dgoodwin, line 730: @suhanime hoping to use this condition to consolidate several of the "pre-flight" conditions we set today prior to launching the install, what do you think? (goal being to trim down the number)

pkg/controller/clusterdeployment/installconfigvalidation.go
```diff
@@ -0,0 +1,50 @@
...
+	noGCPPlatformErr   = "install config did not contain a GCP platform"
+	noAzurePlatformErr = "install config did not contain an Azure platform"
+	regionMismatchErr  = "install config region does not match cluster deployment region"
+)
+
+func ValidateInstallConfig(cd *hivev1.ClusterDeployment, installConfig []byte) error {
+
+	ic := &installertypes.InstallConfig{}
+	if err := yaml.Unmarshal(installConfig, &ic); err != nil {
+		return errors.Wrap(err, "could not unmarshal InstallConfig")
```
dgoodwin, line 23: Tried to do this without parsing to a type given our previous goal of limited interaction with install config as we may not have the correct version. Had a really hard time working with nested json in this manner and eventually went to the type as we had it vendored regardless. Idea here would be that the region structs are stable and should not change, and it doesn't matter if the install config has features we don't have vendored yet, we just want to check some core things.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Actually now that I think of it, we hope to soon create a ClusterInstall implementation for our current OpenShift installer provisioning. When we do, the contract requires that we would be copying back a ClusterInstallRequirementsMet condition. Probably best to just use this same condtion name here for consistency.

suhanime: Is it really necessary to introduce a new condition? Can't we just use the `ProvisionFailed` condition? We can set the reason+message of what you send from `ValidateInstallConfig`
We already do this with `Hibernating` condition - we explicitly set it to false when a cluster cannot be hibernated without even trying to hibernate it. In my opinion, same logic applies here
Additional benefit - we would get the alert via `ProvisionUnderway` metric so it could always be monitored

pkg/controller/clusterdeployment/installconfigvalidation_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	"k8s.io/client-go/kubernetes/scheme"
+
+	"github.com/openshift/hive/apis"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	hivev1aws "github.com/openshift/hive/apis/hive/v1/aws"
+	hivev1azure "github.com/openshift/hive/apis/hive/v1/azure"
+	hivev1gcp "github.com/openshift/hive/apis/hive/v1/gcp"
+	testcd "github.com/openshift/hive/pkg/test/clusterdeployment"
+	"github.com/stretchr/testify/assert"
```
suhanime, line 14: Nit: Move this import above the section of hive imports

pkg/controller/clusterdeployment/installconfigvalidation_test.go
```diff
@@ -0,0 +1,236 @@
...
+		cd            *hivev1.ClusterDeployment
+		expectedError string
+	}{
+		{
+			name: "test aws install config valid",
+			cd: cdBuilder.Build(
+				testcd.WithAWSPlatform(&hivev1aws.Platform{Region: "us-east-1"}),
+			),
+			ic:            testAWSIC,
+			expectedError: "",
```
suhanime, line 156: `expectedError` is of string type, so not mentioning it is equivalent to it being ""

pkg/controller/clusterdeployment/installconfigvalidation_test.go
```diff
@@ -0,0 +1,236 @@
...
+			ic:            testAzureIC,
+			expectedError: regionMismatchErr,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			assert.Equal(t, "foo", "foo")
+			err := ValidateInstallConfig(test.cd, []byte(test.ic))
+			if test.expectedError == "" {
+				assert.NoError(t, err)
```
suhanime, line 228: Maybe a message here?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -686,6 +687,48 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		}
 	}
 
+	// Ensure the install config matches the ClusterDeployment:
+	icSecret := &corev1.Secret{}
+	err = r.Get(context.Background(),
+		types.NamespacedName{
+			Namespace: cd.Namespace,
+			Name:      cd.Spec.Provisioning.InstallConfigSecretRef.Name,
```
suhanime, line 695: Do we always expect this to be present? Maybe we should allow for it not to be present before proceeding to validate the install config (which directly calls the unmarshal code)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
suhanime: fwiw, if you do decide to keep this condition around, it becomes a necessary condition for provision to take place, so for the failing tests, if you add this to [defaultConditions](https://github.com/openshift/hive/blob/4376099acc9c78577b1609347655c98bc98def28/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go#L2330-L2347), it might help with some of the failing tests

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Good point I'll add a check if it's not present to set the condition to indicate there is no install config.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
suhanime: This is the condition we already have [ClusterInstallRequirementsMetClusterDeploymentCondition](https://github.com/openshift/hive/blob/master/apis/hive/v1/clusterdeployment_types.go#L409)
It would be too confusing to have both - one with `ClusterDeployment` in the name and one without...
As far as your idea of using the same one for this purpose... I'm a little bit cautious, but worth looking into

dgoodwin: Thoughts on conditions: https://github.com/openshift/hive/pull/1413

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -690,6 +691,75 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+		// TODO: Move with the openshift-installer ClusterInstall code when we implement https://issues.redhat.com/browse/HIVE-1522
+		if cd.Spec.Provisioning.InstallConfigSecretRef == nil {
+			cdLog.Info("install config not specified")
+			conditions, changed := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+				cd.Status.Conditions,
+				hivev1.RequirementsMetCondition,
+				corev1.ConditionFalse,
+				"InstallConfigRefNotSet",
+				"Install config reference is not set",
+				controllerutils.UpdateConditionIfReasonOrMessageChange)
```
suhanime, line 704: Is the InstallConfigSecretRef not optional? From what I can tell from code - it is passed through to the installer if present

suhanime (commented):
Just 1 question, otherwise looks good pending squash/rebase

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: It's optional at the API level, but we can't go further without one here. If we're doing a normal openshift-install provision, you must supply one, and if we don't have it there's no point in launching the ClusterProvision. Previously we'd just let kube hold the pod up because it's missing a secret to mount.

PR #1404, opened 2021-06-02
controller/clusterdeployment: rate limit updates to clusterdeployment because of error

abhinavdahiya:
/assign @suhanime 
/cc @dgoodwin

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -177,7 +178,8 @@ func AddToManager(mgr manager.Manager, r reconcile.Reconciler, concurrentReconci
 	}
 
 	// Watch for changes to ClusterDeployment
-	err = c.Watch(&source.Kind{Type: &hivev1.ClusterDeployment{}}, &handler.EnqueueRequestForObject{})
+	err = c.Watch(&source.Kind{Type: &hivev1.ClusterDeployment{}},
+		controllerutils.NewRateLimitedUpdateEventHandler(&handler.EnqueueRequestForObject{}, IsErrorUpdateEvent))
 	if err != nil {
 		logger.WithError(err).Error("Error watching cluster deployment")
 		return err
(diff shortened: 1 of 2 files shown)


dgoodwin:
Looks good to me.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2013,3 +2015,43 @@ func LoadReleaseImageVerifier(config *rest.Config) (verify.Interface, error) {
...
+		}
+
+		if controllerutils.IsConditionInDesiredState(*oldcond) {
+			return true // newly Failed condition
+		}
+
+		if cond.Message != oldcond.Message ||
+			cond.Reason != oldcond.Reason {
+			return true // already failing but change in error reported
+		}
```
suhanime, line 2052: We have something like this that we pass on as a criteria when we try to [update conditions](https://github.com/openshift/hive/blob/4376099acc9c78577b1609347655c98bc98def28/pkg/controller/utils/conditions.go#L28-L33)
This takes away the logic from the setting of conditions
Is there a way that we have the `UpdateConditionAlways`, `UpdateConditionNever`, `UpdateConditionIfReasonOrMessageChange` logic in only 1 place?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2013,3 +2015,43 @@ func LoadReleaseImageVerifier(config *rest.Config) (verify.Interface, error) {
...
+	}
+
+	for _, cond := range new.Status.Conditions {
+		if controllerutils.IsConditionInDesiredState(cond) {
+			continue
+		}
+
+		oldcond := controllerutils.FindClusterDeploymentCondition(old.Status.Conditions, cond.Type)
+		if oldcond == nil {
+			return true // newly added condition in failed state
```
suhanime, line 2042: failed state - or conditions were just initialized and set to unknown

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2013,3 +2015,43 @@ func LoadReleaseImageVerifier(config *rest.Config) (verify.Interface, error) {
...
+
+	old, ok := evt.ObjectOld.(*hivev1.ClusterDeployment)
+	if !ok {
+		return false
+	}
+
+	for _, cond := range new.Status.Conditions {
+		if controllerutils.IsConditionInDesiredState(cond) {
+			continue
+		}
```
suhanime, line 2038: Wouldn't we want to update once if it was in undesired state before and now it is in desired state? We would want to give the signal that the error condition is cleared...

abhinavdahiya: the function decides if the the update event was due to error. it doesn't create signals for changes we must take care of.

in this scenario when we go from undesired to desired, the function returns false (unless there was another error change in another condition) and the update event is immediately put in the queue for reconcile.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
abhinavdahiya: too specific for comment to be useful long term. unknown is undesired state for this context.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
abhinavdahiya: > Is there a way that we have the UpdateConditionAlways, UpdateConditionNever, UpdateConditionIfReasonOrMessageChange logic in only 1 place?

that goal for update criteria is different from what we want to do here. updating conditions requires allowing caller to decide when to update whereas here we are calculating if the change in conditions was due to error. Using a function like UpdateConditionIfReasonOrMessageChange here creates more confusion than clarity.

PR #1406, opened 2021-06-04
controller/remotemachineset: summarize failures from machines

abhinavdahiya:
Often the machineset status does not include the errorReason,
errorMessage that would summarize failures in machines belonging to the
set. So the controller now uses the machineset's specified selector to
get machines and then collect the reason and messages for failed
machines.

xref: https://issues.redhat.com/browse/HIVE-1511

/assign @akhil-rane 
/cc @joelddiaz

### DIFF
--- a/pkg/controller/remotemachineset/remotemachineset_controller.go
+++ b/pkg/controller/remotemachineset/remotemachineset_controller.go
@@ -1,6 +1,7 @@
 package remotemachineset
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
I see that on errors while trying to collect any machineSet errors, we just log and continue. Not sure if it is worth erroring the entire remotemachineset.Reconcile() b/c of this (I don't have a strong opinion).

Either way, I look forward to having this visibility going forward :+1:

akhil-rane (commented):
/lgtm

PR #1407, opened 2021-06-07
controller runtime 0.8.3

dgoodwin:
- Update to controller-runtime v0.8.3.
- Test fixes for update to controller-runtime 0.8.3.


dgoodwin:
/hold

This is very busted.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2491,10 +2491,11 @@ func testEmptyClusterDeployment() *hivev1.ClusterDeployment {
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name:       testName,
-			Namespace:  testNamespace,
-			Finalizers: []string{hivev1.FinalizerDeprovision},
-			UID:        types.UID("1234"),
+			Name:            testName,
+			Namespace:       testNamespace,
+			Finalizers:      []string{hivev1.FinalizerDeprovision},
+			UID:             types.UID("1234"),
+			ResourceVersion: "999",
```
2uasimojo, line 2498: Is there a reference somewhere for this `999` thing?

Without context, I'm guessing it must be a magic number telling the server to force a patch/update that would ordinarily bounce 409-ish.

dgoodwin: Listed as a breaking change in https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.8.0

dgoodwin:
/hold cancel

Should be fixed now.

pkg/controller/velerobackup/velerobackup_controller_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	testcheckpoint "github.com/openshift/hive/pkg/test/checkpoint"
+	"github.com/openshift/hive/pkg/test/clusterdeployment"
 	testclusterdeployment "github.com/openshift/hive/pkg/test/clusterdeployment"
```
abhinavdahiya, line 19: double import?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
2uasimojo: I got [clarification](https://coreos.slack.com/archives/CB48XQ4KZ/p1623082745004400) from the author of that change. He confirmed that these `999` changes are not necessary. The "breaking" aspect of that change is that it would cause tests that were previously succeeding (incorrectly) to fail (correctly). So no fixes are necessary unless we run into such failures.

2uasimojo (changes requested):
The good news: We don't need all the ResourceVersion changes.

The bad news: Gotta revert all the ResourceVersion changes :P

2uasimojo:
> The good news: We don't need all the ResourceVersion changes.
> 
> The bad news: Gotta revert all the ResourceVersion changes :P

Aw heck, I didn't even notice you prudently put that stuff into a separate commit. Easy peasy.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
2uasimojo: Update: We *do* run into failures -- but they're not conflict-ish. Rather, they're due to the fact that the tests use DeepEquals to determine expected vs actual objects, so now that the fake server is initializing ResourceVersion to `999` those comparisons are failing.

Long-term we ought to fix the tests not to rely on explicit values of this field, which is supposed to be opaque. For now, this is a reasonable workaround.

2uasimojo:
Per https://github.com/openshift/hive/pull/1407#discussion_r646791099 I'm convinced the `999` changes are necessary (until we make the tests smarter).

/lgtm

/test e2e

:arrow_up: timeout

2uasimojo:
Oh, except I agree with @abhinavdahiya's comment.

/lgtm cancel

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
dgoodwin: Good summary, tests definite are a bit too brittle.

pkg/controller/velerobackup/velerobackup_controller_test.go
dgoodwin: fixed!

PR #1408, opened 2021-06-07
Initialize clusterpool+clusterclaim conditions

suhanime:
xref: https://issues.redhat.com/browse/HIVE-1556

/assign @akhil-rane 
/cc @dgoodwin

### DIFF
--- a/pkg/controller/clusterclaim/clusterclaim_controller.go
+++ b/pkg/controller/clusterclaim/clusterclaim_controller.go
@@ -35,6 +35,13 @@ const (
 	hiveClaimOwnerRoleBindingName = "hive-claim-owner"
 )
 
+// clusterClaimConditions are the cluster claim conditions controlled or initialized by cluster claim controller
+var clusterClaimConditions = []hivev1.ClusterClaimConditionType{
+	hivev1.ClusterClaimPendingCondition,
+	hivev1.ClusterClaimClusterDeletedCondition,
+	hivev1.ClusterRunningCondition,
+}
+
 // Add creates a new ClusterClaim Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
@@ -169,6 +176,18 @@ func (r *ReconcileClusterClaim) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	// Initialize cluster claim conditions if not set
+	newConditions := controllerutils.InitializeClusterClaimConditions(claim.Status.Conditions, clusterClaimConditions)
+	if len(newConditions) > len(claim.Status.Conditions) {
+		claim.Status.Conditions = newConditions
+		logger.Infof("initializing cluster claim conditions")
+		if err := r.Status().Update(context.TODO(), claim); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster claim status")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+
 	if claim.DeletionTimestamp != nil {
 		return r.reconcileDeletedClaim(claim, logger)
 	}
@@ -211,7 +230,7 @@ func (r *ReconcileClusterClaim) Reconcile(ctx context.Context, request reconcile
 	if lifetime != nil {
 		logger.WithField("lifetime", lifetime).Debug("checking whether lifetime of ClusterClaim has elapsed")
 		pendingCond := controllerutils.FindClusterClaimCondition(claim.Status.Conditions, hivev1.ClusterClaimPendingCondition)
-		if pendingCond != nil && pendingCond.Status == corev1.ConditionFalse {
+		if pendingCond.Status == corev1.ConditionFalse {
 			if timeSinceAssigned := time.Since(pendingCond.LastTransitionTime.Time); timeSinceAssigned >= lifetime.Duration {
 				logger.WithField("timeSinceAssigned", timeSinceAssigned).
 					WithField("lifetime", lifetime).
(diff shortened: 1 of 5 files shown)


pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -902,10 +920,6 @@ func (r *ReconcileClusterPool) assignClustersToClaims(claims []*hivev1.ClusterCl
 			claim.Spec.Namespace = cds[0].Namespace
 			cds = cds[1:]
 			logger.WithField("cluster", claim.Spec.Namespace).Info("assigning cluster to claim")
-			if err := r.Update(context.Background(), claim); err != nil {
-				logger.WithError(err).Log(controllerutils.LogLevel(err), "could not assign cluster to claim")
-				return cds, err
-			}
```
abhinavdahiya, line 908: hmm.. it seems like removing this would stop assigning clusters to claim, as this updates the spec ??

suhanime: I removed this update because we are calling another update right below `else` - see https://github.com/openshift/hive/blob/9e0a375372593f2bb7890f7cfee4a615ed2daff9/pkg/controller/clusterpool/clusterpool_controller.go#L943-L949

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -944,26 +960,25 @@ func TestReconcileClusterPool(t *testing.T) {
...
 
-			missingDependentsCondition := controllerutils.FindClusterPoolCondition(pool.Status.Conditions, hivev1.ClusterPoolMissingDependenciesCondition)
-			if test.expectedMissingDependenciesStatus != nil {
-				require.NotNil(t, missingDependentsCondition)
-				expectedStatus := corev1.ConditionFalse
-				if *test.expectedMissingDependenciesStatus {
-					expectedStatus = corev1.ConditionTrue
+			if test.expectedMissingDependenciesStatus != "" {
+				missingDependenciesCondition := controllerutils.FindClusterPoolCondition(pool.Status.Conditions, hivev1.ClusterPoolMissingDependenciesCondition)
+				if assert.NotNil(t, missingDependenciesCondition, "did not find MissingDependencies condition") {
```
akhil-rane, line 965: Since we always expect missingDependenciesCondition in any case, shouldn't this be `require.NotNil` ? I believe require is a precondition here

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -944,26 +960,25 @@ func TestReconcileClusterPool(t *testing.T) {
...
+					assert.Equal(t, test.expectedMissingDependenciesStatus, missingDependenciesCondition.Status,
+						"unexpected MissingDependencies conditon status")
+					if test.expectedMissingDependenciesMessage != "" {
```
akhil-rane, line 968: Is this check required? what happens if we don't have this check?

pkg/controller/clusterpool/clusterpool_controller_test.go
suhanime: assert returns a boolean - so it would compare the status only if it missingDependencies is present. I prefer the assert because it allows to proceed even if the assert here returns false

pkg/controller/clusterpool/clusterpool_controller_test.go
suhanime: Some of the tests have message present for the expected condition - I did not want to remove that additional check so I included it here, and at the same time, allow for just the reason to be compared.

pkg/controller/clusterpool/clusterpool_controller_test.go
dgoodwin: If everything below depended on the missingDependenciesCondition I'd say require, but it looks like we transition to other checks so assert is probably ok here and lets us get more info out of the test failure.

pkg/controller/clusterpool/clusterpool_controller.go
abhinavdahiya: the updates calls are not the same and that's why i was concerned.

the removed code updates the spec.
```
if err := r.Update(context.Background(), claim); err != nil {
```

while the linked code in previous comment only updates the status.
```
if err := r.Status().Update(context.Background(), claim); err != nil { 
```

it's weird that no unit test is failing because of this..

suhanime: Ah - I didn't notice - yes - it is weird, but I will make the changes to not drop the update code here...

suhanime: done - thank you for catching that.
Looks like for the test that checks if claim is assigned is passing regardless of whether I keep or remove that update :shrug: 
Here's the [check](https://github.com/openshift/hive/blob/3c57723f538887a124852bffd9b7dd0d9070c852/pkg/controller/clusterpool/clusterpool_controller_test.go#L986-L995) and here's the [test](https://github.com/openshift/hive/blob/3c57723f538887a124852bffd9b7dd0d9070c852/pkg/controller/clusterpool/clusterpool_controller_test.go#L583L597)

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -211,7 +230,7 @@ func (r *ReconcileClusterClaim) Reconcile(ctx context.Context, request reconcile
 	if lifetime != nil {
 		logger.WithField("lifetime", lifetime).Debug("checking whether lifetime of ClusterClaim has elapsed")
 		pendingCond := controllerutils.FindClusterClaimCondition(claim.Status.Conditions, hivev1.ClusterClaimPendingCondition)
-		if pendingCond != nil && pendingCond.Status == corev1.ConditionFalse {
+		if pendingCond.Status == corev1.ConditionFalse {
```
akhil-rane, line 233: Will this break for existing claims with no pending condition? If yes, we can keep `pendingCond != nil` as extra safety net

suhanime: This will never be nil because as a part of reconcile we are always checking if those conditions are not present then [initialize](https://github.com/openshift/hive/pull/1408/files/3c57723f538887a124852bffd9b7dd0d9070c852..48b81ce9c350521fed4a2872f215284d90af44bd#diff-429aa8363e2e77526949e5194551e15dc6b97dd27ca7f9dbdd9ed9b963f8609eR179-R190) them.

akhil-rane: ah, yes

akhil-rane (commented):
/lgtm

PR #1411, opened 2021-06-08
Remove VeleroBackup

2uasimojo:
It was unused, and started causing test failures with [controller-runtime 0.8.x](https://github.com/openshift/hive/pull/1407).

[HIVE-1563](https://issues.redhat.com/browse/HIVE-1563)


apis/hive/v1/hiveconfig_types.go
```diff
@@ -44,11 +44,6 @@ type HiveConfigSpec struct {
 	// +optional
 	GlobalPullSecretRef *corev1.LocalObjectReference `json:"globalPullSecretRef,omitempty"`
 
-	// Backup specifies configuration for backup integration.
```
2uasimojo, line 47: Hm. Are we allowed to remove a field like this without bumping versions?

2uasimojo:
/test e2e

:up: timeout

dgoodwin:
@mdelder @jnpacker here is the PR removing hive's velero backup support. I thought I wrote down the name of the person you suggested I get to look at how this worked but I cannot find it. 

If configured, it would trigger a velero backup for everything (i think) in a cluster namespace, whenever a cluster object changed. This initially caused some rate problems with the cluster velero in scale tests. We went back and corrected with some rate limiting. It was done for our SRE teams but in the end never used, they used their own global velero backup solution last we heard. 

As far as we know this was never used, and we expect it never will be. Because it's implemented only within Hive, it would never encompass everything in ACM or other related operators and thus it's value appears limited. We expect the trend of preferring broader backup solutions to continue.

cc @jewzaam @mwoodson @jharrington22 



However looking through https://kubernetes.io/docs/reference/using-api/deprecation-policy/, I don't think we can remove this. (apologies @2uasimojo as I set you on this path) We probably could do this without harming anyone, but if we're wrong it could be a catastrophe for that user. And according to the doc, we need to take a much longer path to removing it.

2uasimojo:
/hold

Per above, we probably won't do this. But still discussing.

jnpacker:
I asked Dario and he said he is not using it, so we're good.

2uasimojo:
We've decided to go through the proper deprecation process, so it'll be a year before we can actually do this. If somehow this PR magically rebases cleanly at that point, great, we can reopen.

PR #1412, opened 2021-06-08
Abandon deprovision for any cluster when preserveOnDelete is true

akhil-rane:
x-ref: https://issues.redhat.com/browse/HIVE-1457

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -424,7 +424,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 		},
 		{
-			name: "Test PreserveOnDelete",
+			name: "Test PreserveOnDelete when cluster deployment is installed",
 			existing: []runtime.Object{
 				func() *hivev1.ClusterDeployment {
 					cd := testClusterDeploymentWithInitializedConditions(testDeletedClusterDeployment())
@@ -445,7 +445,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 		},
 		{
-			name: "Test creation of uninstall job when PreserveOnDelete is true but cluster deployment is not installed",
+			name: "Test PreserveOnDelete when cluster deployment is not installed",
 			existing: []runtime.Object{
 				func() *hivev1.ClusterDeployment {
 					cd := testClusterDeploymentWithInitializedConditions(testDeletedClusterDeployment())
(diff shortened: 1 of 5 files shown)


2uasimojo:
This looks fine as far as it goes, but it seems we're [still not sure](https://coreos.slack.com/archives/CCX9DB894/p1623185909273900?thread_ts=1623180157.242300&cid=CCX9DB894) that the existing code does the right thing.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -449,24 +449,6 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				assert.Nil(t, deprovision, "expected no deprovision request")
 			},
 		},
-		{
```
2uasimojo, line 452: Rather than removing this test, can we change the expected result to prove that no deprovision is created?

2uasimojo: ...i.e. it would look like the preceding case, just with `cd.Spec.Installed = false`

akhil-rane: Since `cd.Spec.Installed` does not have any effect on preserveOnDelete functionality anymore, do we really need this test? We can remove `cd.Spec.Installed = true` from previous test and it will still pass.

2uasimojo: IMHO yes. The point of UT is not only to make sure the code does what we think; it's also to catch regressions introduced by future code changes. If in the future someone reintroduces logic around `Installed = true`, we want this test to fail.

Happy to be overridden, though. @dgoodwin was [just saying](https://coreos.slack.com/archives/CE3ETN3J8/p1623327974171300) how bloated this suite is...

dgoodwin: Seems like a good idea to me, lets go for it and keep a test to ensure nobody breaks this in the future.

akhil-rane: sounds good

akhil-rane:
> So one more case I'd like us to consider here:
> 
> If a deprovision request exists when `preserveOnDelete` is set, should we delete it? I feel like we should.

IIUC that will automatically be deleted since deprovision request has owner reference set to ClusterDeployment, unless someone does `oc delete cd <name> --cascade='orphan'`

2uasimojo:
> IIUC that will automatically be deleted

Okay, great. (I checked around a little bit, but didn't think to look at OwnerReference.)

/lgtm

Let's remember to inform SRE-P when this lands in prod so they can update their SOPs.

PR #1414, opened 2021-06-09
Update default release image URL.

abutcher:
Redirect no longer exists for `openshift-release.svc.ci.openshift.org`.

PR #1415, opened 2021-06-09
Fix velerobackup tests for DeepEquals

2uasimojo:
Several tests were using DeepEquals (under testify's `assert.Equal`) to compare a runtime.Object from the (fake) kube server to a locally-constructed expected value. This is brittle, as fields such as ResourceVersion cannot be relied on to have predictable values.

Commit c7d60527612c141f58d748b8bf2e25d72f9499d5 in #1407 worked around the problem by explicitly setting the ResourceVersion to the value set by the fake kube server in controller-runtime 0.8.x. This PR has two commits:
- Revert that workaround;
- Switch to a comparison that disregards ResourceVersion and TypeMeta.

### DIFF
--- a/pkg/controller/velerobackup/helpers_test.go
+++ b/pkg/controller/velerobackup/helpers_test.go
@@ -1,12 +1,14 @@
 package velerobackup
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 	"time"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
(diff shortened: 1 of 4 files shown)


2uasimojo:
Note to reviewers: Ignore the first two commits. Their PRs (#1410, #1414) are approved and pending merge, so I stacked this up to avoid a dirty rebase later.

dgoodwin:
Looks great ty for cleaning this up.

/lgtm

PR #1417, opened 2021-06-11
Initialize MachinePool conditions

suhanime:
This PR initializes MachinePool conditions, and ensures conditions do not have to be "True" to be set for the first time.

xref: https://issues.redhat.com/browse/HIVE-1562

/assign @abutcher 
/cc @akhil-rane

### DIFF
--- a/pkg/controller/remotemachineset/awsactuator.go
+++ b/pkg/controller/remotemachineset/awsactuator.go
@@ -322,10 +322,10 @@ func (a *AWSActuator) getPrivateSubnetsByAvailabilityZone(pool *hivev1.MachinePo
 
 	results, err := a.awsClient.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
 	if err != nil || len(results.Subnets) == 0 {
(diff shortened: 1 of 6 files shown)


pkg/controller/remotemachineset/awsactuator.go
```diff
@@ -322,10 +322,10 @@ func (a *AWSActuator) getPrivateSubnetsByAvailabilityZone(pool *hivev1.MachinePo
 
 	results, err := a.awsClient.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: idPointers})
 	if err != nil || len(results.Subnets) == 0 {
-		if strings.Contains(err.Error(), "InvalidSubnetID.NotFound") {
-			var conditionMessage string
+		if strings.Contains(err.Error(), "InvalidSubnet") {
```
suhanime, line 325: I hit `InvalidSubnets` for `subnets specfied in the machinepool do not exist` test - this seemed like a good compromise.
`err.Error()` is `InvalidSubnets`

PR #1418, opened 2021-06-11
Skeleton enhancement doc for Hive v2

2uasimojo:
Lay down the skeleton for an enhancement doc to collect ideas for v2 of the Hive API.

### DIFF
--- a/docs/enhancements/hivev2.md
+++ b/docs/enhancements/hivev2.md
@@ -1,0 +1,76 @@
+# Hive v2
+
+- [Hive v2](#hive-v2)
+  - [Overview](#overview)
+    - [This is Not](#this-is-not)
+    - [How To Use](#how-to-use)
+  - [APIs](#apis)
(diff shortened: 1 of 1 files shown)


dgoodwin:
Some other items on my mind:

- Is our ClusterInstall duck typing design working out as planned? Will know more once we implement an equivalent for our openshift-install code. I'm nervous that we have broken out the day 0 install settings, but some details still leak into clusterdeployment.spec.platform. 
- Drop the provisioning section on ClusterDeployment, and all other fields that became a part of ClusterInstalls ^^. 
- Are we a good fit for Hypershift? For managed Red Hat installs? Why or why not?

2uasimojo:
@dgoodwin The idea would be to land this skeleton, then propose those items via one or more subsequent PRs (like #1419). That way we don't hold one PR open forever while we flesh out a bunch of disparate ideas.

PR #1420, opened 2021-06-15
Fix operator bundle scripts for v1 CRD schemas.

dgoodwin:


2uasimojo:
Ah, yes.

Do we have good a way to test this pipeline locally?

dgoodwin:
Not that I know of, the appsre job here doesn't seem to hit it.

I also thought we were off jenkins.

PR #1426, opened 2021-06-16
Bump to RHEL8 in Dockerfile for libvirt dependencies.

dgoodwin:
See: https://github.com/open-cluster-management/backlog/issues/13183


2uasimojo:
Syncs the Dockerfile with [2.3](https://github.com/openshift/hive/blob/ocm-2.3/Dockerfile#L7-L17), which seems okay. I'd prefer someone more experienced to lgtm though.

PR #1413, opened 2021-06-09
Rough thoughts on re-organizing clusterdeployment conditions.

dgoodwin:

PR #1430, opened 2021-06-21
Remove ClusterImageSetMissing condition.

dgoodwin:
Move to RequirementsMet with reason ClusterImageSetMissing.

x-ref: https://issues.redhat.com/browse/HIVE-1481

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -58,7 +58,6 @@ var (
 	// clusterDeploymentConditions are the cluster deployment conditions controlled or
 	// initialized by cluster deployment controller
 	clusterDeploymentConditions = []hivev1.ClusterDeploymentConditionType{
-		hivev1.ClusterImageSetNotFoundCondition,
 		hivev1.DNSNotReadyCondition,
 		hivev1.InstallImagesNotResolvedCondition,
 		hivev1.ProvisionFailedCondition,
@@ -310,6 +309,36 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
 	if len(newConditions) > len(cd.Status.Conditions) {
 		cd.Status.Conditions = newConditions
 		cdLog.Info("initializing cluster deployment controller conditions")
+		if err := r.Status().Update(context.TODO(), cd); err != nil {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+
+	// Remove legacy conditions if present:
+	legacyConditions := []string{
+		"ClusterImageSetNotFound",
+	}
+	newConditions = []hivev1.ClusterDeploymentCondition{}
+	var removedLegacyConditions bool
+	for _, c := range cd.Status.Conditions {
+		isLegacy := false
+		for _, lc := range legacyConditions {
+			if string(c.Type) == lc {
+				isLegacy = true
+				break
+			}
+		}
+		if !isLegacy {
+			newConditions = append(newConditions, c)
+		} else {
+			cdLog.Infof("removing legacy condition: %v", c)
+			removedLegacyConditions = true
+		}
+	}
+	if removedLegacyConditions {
+		cd.Status.Conditions = newConditions
 		if err := r.Status().Update(context.TODO(), cd); err != nil {
 			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update cluster deployment status")
 			return reconcile.Result{}, err
@@ -855,7 +884,7 @@ func (r *ReconcileClusterDeployment) getClusterImageSet(cd *hivev1.ClusterDeploy
 		imageSetKey.Name = isName
 	default:
 		cdLog.Warning("clusterdeployment references no clusterimageset")
-		if err := r.setImageSetNotFoundCondition(cd, "unknown", true, cdLog); err != nil {
+		if err := r.setReqsMetConditionImageSetNotFound(cd, "unknown", true, cdLog); err != nil {
 			return nil, err
 		}
 	}
@@ -865,7 +894,7 @@ func (r *ReconcileClusterDeployment) getClusterImageSet(cd *hivev1.ClusterDeploy
 	if apierrors.IsNotFound(err) {
 		cdLog.WithField("clusterimageset", imageSetKey.Name).
 			Warning("clusterdeployment references non-existent clusterimageset")
-		if err := r.setImageSetNotFoundCondition(cd, imageSetKey.Name, true, cdLog); err != nil {
+		if err := r.setReqsMetConditionImageSetNotFound(cd, imageSetKey.Name, true, cdLog); err != nil {
 			return nil, err
 		}
 		return nil, err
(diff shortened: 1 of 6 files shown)


dgoodwin:
/assign @suhanime 
/cc @abhinavdahiya 

This is a first stab at implementing what we talked about in https://github.com/openshift/hive/pull/1413

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -317,6 +316,35 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
...
+				legacy = true
+				break
+			}
+			if !legacy {
+				newConditions = append(newConditions, c)
+			} else {
+				cdLog.Infof("removing legacy condition: %v", c)
+			}
+		}
+	}
```
suhanime, line 337: Can't we use `FindClusterDeploymentCondition`? Looks like it is working [here](https://github.com/openshift/hive/pull/1430/files#diff-b9685aa0bf796d1db13613fe411ed287146777107405bcc935ebd1d45d64ea47R1457)

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2524,8 +2574,6 @@ func testEmptyClusterDeployment() *hivev1.ClusterDeployment {
 }
 
 func testInstallConfigSecret() *corev1.Secret {
-	encodedIC := base64.StdEncoding.EncodeToString([]byte(testAWSIC))
-	log.Info(encodedIC)
```
suhanime, line 2528: Doesn't look like it is related to these changes

suhanime (commented):
Trajectory looks great.
Needs a `make vendor` and changing the tests which mention `hivev1.ClusterImageSetNotFoundCondition` in metrics and controllerutils tests

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: That would let me find if it's there, but we have no delete function and probably shouldn't given we want them always there (except for legacy). Reslicing looked a little clunk, and requires an index, which FindClusterDeploymentCondition doesn't expose.  Given all this it felt more reasonable to just iterate the slice and built up a new with the ones we want to keep. Let me know if I missed something obvious though.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
dgoodwin: It's not it's from a recent PR I did, but doing a separate PR and CI cycle to remove two lines of logging feels worse to me than including it in a diff PR. WDYT?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1086,26 +1115,38 @@ func (r *ReconcileClusterDeployment) setDeprovisionLaunchErrorCondition(cd *hive
 }
 
 func (r *ReconcileClusterDeployment) setImageSetNotFoundCondition(cd *hivev1.ClusterDeployment, name string, isNotFound bool, cdLog log.FieldLogger) error {
```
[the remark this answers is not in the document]
dgoodwin: renamed to setReqsMetConditionImageSetNotFound

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1086,26 +1115,38 @@ func (r *ReconcileClusterDeployment) setDeprovisionLaunchErrorCondition(cd *hive
...
-		message,
-		controllerutils.UpdateConditionNever)
 	if !changed {
 		return nil
 	}
-	cdLog.Infof("setting ClusterImageSetNotFoundCondition to %v", status)
 	cd.Status.Conditions = conds
+	reqsCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions,
+		hivev1.RequirementsMetCondition)
+	cdLog.Info("updating RequirementsMetCondition: %v", reqsCond)
```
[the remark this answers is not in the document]
dgoodwin: Updated

pkg/controller/clusterdeployment/clusterdeployment_controller.go
suhanime: The logic is good - I just meant to see if we could make it not clunky - it is fine this way...

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
suhanime: No worries - just wanted to make sure it was purposeful

suhanime (commented):
metrics and controllerutils tests need to be fixed post vendor changes, everything else lgtm
` vet: pkg/controller/metrics/metrics_test.go:35:12: ClusterImageSetNotFoundCondition not declared by package v1
# github.com/openshift/hive/pkg/controller/utils
vet: pkg/controller/utils/conditions_test.go:15:10: ClusterImageSetNotFoundCondition not declared by package v1
make: *** [verify-govet] Error 2 `

dgoodwin:
Updated, thanks for looking!

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -317,6 +316,35 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
...
+				break
+			}
+			if !legacy {
+				newConditions = append(newConditions, c)
+			} else {
+				cdLog.Infof("removing legacy condition: %v", c)
+			}
+		}
+	}
+	if len(newConditions) < len(cd.Status.Conditions) {
```
2uasimojo, line 338: Is this future-proof? Would it be better to use a bool flag set to `true` in the loop above? (Like `legacy`, but wider scoped.)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -317,6 +316,35 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
...
+			if string(c.Type) == lc {
+				legacy = true
+				break
+			}
+			if !legacy {
+				newConditions = append(newConditions, c)
+			} else {
+				cdLog.Infof("removing legacy condition: %v", c)
+			}
+		}
```
2uasimojo, line 336:
```suggestion
		legacy := false
		for _, lc := range legacyConditions {
			if string(c.Type) == lc {
				legacy = true
				break
			}
		}
		if !legacy {
			newConditions = append(newConditions, c)
		} else {
			cdLog.Infof("removing legacy condition: %v", c)
		}
```
?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -317,6 +316,35 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
...
+	}
+	if len(newConditions) < len(cd.Status.Conditions) {
+		cd.Status.Conditions = newConditions
+		cdLog.Info("removed legacy conditions")
```
2uasimojo, line 340: Seems unnecessary with the log(s) on L334.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Whoops!

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Hmm perhaps if someone injected some code in the middle of these statements. I'll add a bool.

dgoodwin:
Updates pushed.

PR #1432, opened 2021-06-24
Include cloud errors in dnszone condition

suhanime:
Now that DNSZone controller is rate limited while querying for errors,
we can safely include dnszone related errors in DNSNotReady cluster
deployment condition.

To facilitate this, a generic error condition is introduced as a part of
DNSZone CR, which is used to pass-on the major errors to cluster
deployment.

This PR also includes the error-scrubing facility from [CCO](https://github.com/openshift/cloud-credential-operator/commit/d3ad0d4aafec8aaf3cc0d59eb074266e5421273a), that strips
the request ID from the error message.

xref: https://issues.redhat.com/browse/HIVE-1554

/assign @abutcher 
/cc @dgoodwin


pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -506,6 +506,88 @@ func TestReconcileDNSProviderForAzure(t *testing.T) {
...
+
+			setFakeDNSZoneInKube(mocks, tc.dnsZone)
+
+			if tc.setupAWSMock != nil {
+				tc.setupAWSMock(mocks.mockAWSClient.EXPECT())
+			}
+
+			// Act
+			_, err := r.reconcileDNSProvider(zr, tc.dnsZone)
+			zr.SetConditionsForError(err)
```
suhanime, line 569: I might have taken the easier route here - implementing entire Reconcille Test was too complicated

pkg/controller/utils/errorscrub.go
```diff
@@ -0,0 +1,18 @@
+package utils
+
+import (
+	"regexp"
+)
+
+var (
+	awsRequestIDRE = regexp.MustCompile(`(, )*(request id|Request ID): ([-0-9a-f]+)`)
```
suhanime, line 8: This is very AWS specific regex but I have still used it for GCP and Azure actuators as well just to be safe

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -152,6 +154,19 @@ func (r *ReconcileDNSZone) Reconcile(ctx context.Context, request reconcile.Requ
...
+			hivev1.GenericDNSErrorsCondition,
+			corev1.ConditionTrue,
+			"RelocationError",
+			controllerutils.ErrorScrub(err),
+			controllerutils.UpdateConditionIfReasonOrMessageChange)
+		if changed {
+			if err := r.Status().Update(context.Background(), desiredState); err != nil {
+				dnsLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update dnszone status")
+			}
+		}
```
dgoodwin, line 169: Would this get cleared if it was set, and then the problem resolved?

pkg/controller/dnszone/dnszone_controller.go
```diff
@@ -204,6 +219,19 @@ func (r *ReconcileDNSZone) Reconcile(ctx context.Context, request reconcile.Requ
...
+			hivev1.GenericDNSErrorsCondition,
+			corev1.ConditionTrue,
+			"ActuatorNotInitialized",
+			"error instantiating actuator: "+controllerutils.ErrorScrub(err),
+			controllerutils.UpdateConditionIfReasonOrMessageChange)
+		if changed {
+			if err := r.Status().Update(context.Background(), desiredState); err != nil {
+				dnsLog.WithError(err).Log(controllerutils.LogLevel(err), "failed to update dnszone status")
+			}
+		}
```
dgoodwin, line 234: Same here, we need to be sure to clear this when everything is ok. If i've missed it then my bad. :)

pkg/controller/dnszone/dnszone_controller_test.go
dgoodwin: I think it's completely fair to break out a separate test, mega suites that do all kinds of complicated things are not a pattern I like.

However I am confused by these two lines. zr is an actuator, it's passed into the reconciler r, which we call, why would we explicitly call SetConditionsForError on the actuator after running reconcile, which returned the error?

pkg/controller/dnszone/azureactuator.go
```diff
@@ -201,5 +202,30 @@ func (a *AzureActuator) UpdateMetadata() error {
...
+	var cloudErrorsConds []hivev1.DNSZoneCondition
+	var cloudErrorsCondsChanged bool
+	if err == nil {
+		cloudErrorsConds, cloudErrorsCondsChanged = controllerutils.SetDNSZoneConditionWithChangeCheck(
+			a.dnsZone.Status.Conditions,
+			hivev1.GenericDNSErrorsCondition,
+			corev1.ConditionFalse,
+			dnsNoCloudErrorReason,
+			"No cloud errors occurred",
+			controllerutils.UpdateConditionIfReasonOrMessageChange,
```
2uasimojo, line 215: I noticed this was `controllerutils.UpdateConditionNever` for awsactuator [above](https://github.com/openshift/hive/pull/1432/files#diff-2257da48fedf6f2c5593b99c2ea746f86c885d246da2758eb98bb4b2e0038983R617). I'm still trying to wrap my head around how that whole thing works, but IIUC it won't make a difference since
- a status change will always trigger an update; and
- if the status hasn't changed, neither has the reason/message, since it's hardcoded on success.

pkg/controller/dnszone/gcpactuator.go
```diff
@@ -236,7 +236,32 @@ func (a *GCPActuator) Refresh() error {
...
+	var cloudErrorsConds []hivev1.DNSZoneCondition
+	var cloudErrorsCondsChanged bool
+	if err == nil {
+		cloudErrorsConds, cloudErrorsCondsChanged = controllerutils.SetDNSZoneConditionWithChangeCheck(
+			a.dnsZone.Status.Conditions,
+			hivev1.GenericDNSErrorsCondition,
+			corev1.ConditionFalse,
+			dnsNoCloudErrorReason,
+			"No cloud errors occurred",
+			controllerutils.UpdateConditionIfReasonOrMessageChange,
```
2uasimojo, line 249: as above

pkg/controller/utils/errorscrub.go
2uasimojo: Could use case-insensitive here:
```suggestion
	awsRequestIDRE = regexp.MustCompile(`(, )*(?i:request id): ([-0-9a-f]+)`)
```
Also, I'm curious why you're matching & scrubbing the `, `s.

pkg/controller/utils/errorscrub.go
```diff
@@ -0,0 +1,18 @@
...
+	newlineTabRE   = regexp.MustCompile(`(\n\t)`)
+)
+
+// ErrorScrub scrubs cloud error messages destined for CRD status to remove things that
+// change every attempt, such as request IDs, which subsequently cause an infinite update/reconcile loop.
+func ErrorScrub(err error) string {
+	s := awsRequestIDRE.ReplaceAllString(err.Error(), "")
+	s = newlineTabRE.ReplaceAllString(s, ", ")
```
2uasimojo, line 16: Do you want to swap the order of these, so you've already translated `\n\t` to `, ` by the time you scrub the latter?

pkg/controller/utils/errorscrub_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestErrorScrubber(t *testing.T) {
+	cases := []struct {
+		name string
+
+		input    string
+		expected string
+	}{
+		{
+			name:     "aws request id",
+			input:    "failed to grant creds: error syncing creds in mint-mode: AWS Error: LimitExceeded - LimitExceeded: Cannot exceed quota for UsersPerAccount: 5000\n\tstatus code: 409, request id: 0604c1a4-0a68-4d1a-b8e6-cdcf68176d71",
```
2uasimojo, line 19: Is this input string realistic, or would it be
```suggestion
			input:    "failed to grant creds: error syncing creds in mint-mode: AWS Error: LimitExceeded - LimitExceeded: Cannot exceed quota for UsersPerAccount: 5000\n\tstatus code: 409\n\trequest id: 0604c1a4-0a68-4d1a-b8e6-cdcf68176d71",
```
? (see above comment)

pkg/controller/utils/errorscrub_test.go
```diff
@@ -0,0 +1,38 @@
...
+		{
+			name:     "request id mid",
+			input:    "AWS Error: LimitExceeded - LimitExceeded: Cannot exceed quota for UsersPerAccount: 5000, request id: 0604c1a4-0a68-4d1a-b8e6-cdcf68176d71, something else",
+			expected: "AWS Error: LimitExceeded - LimitExceeded: Cannot exceed quota for UsersPerAccount: 5000, something else",
+		},
+		{
+			name:     "request id start",
+			input:    "request id: 0604c1a4-0a68-4d1a-b8e6-cdcf68176d71, something else", // shouldn't really happen
+			expected: ", something else",                                                 // not pretty but what I want to verify happens
+		},
```
2uasimojo, line 31: Could include a test with `Request ID` (capitals) since you're accounting for that in the regex.

vendor/modules.txt
```diff
@@ -114,6 +114,7 @@ github.com/aws/aws-sdk-go/service/elbv2
 github.com/aws/aws-sdk-go/service/elbv2/elbv2iface
 github.com/aws/aws-sdk-go/service/iam
 github.com/aws/aws-sdk-go/service/iam/iamiface
+github.com/aws/aws-sdk-go/service/kinesis
```
2uasimojo, line 117: Expected to see a go.mod/go.sum update with this.

pkg/controller/utils/errorscrub.go
2uasimojo: I wonder if, rather than chopping this whole chunk out of the string, you might prefer to simply replace the ID itself with a static string.

```go
	myre := regexp.MustCompile("(?i)(request id: )(?:[-[:xdigit:]]+)")
	fmt.Println(myre.ReplaceAllString("foo, Request id: abc-123, bar", "${1}XXXX"))
```
Result: `foo, Request id: XXXX, bar`

pkg/controller/dnszone/azureactuator.go
suhanime: That's correct - it wouldn't matter in this instance whether it is `UpdateConditionIfReasonOrMessageChange` or `UpdateConditionNever`, since the next update would only be when the status is changed, but given that this is a generic error condition, I just wanted to keep the opportunity in the future for the condition status to still be True but for some other reason. Either way it wouldn't make a difference right now.

pkg/controller/dnszone/dnszone_controller.go
suhanime: Yes, it will be [here](https://github.com/openshift/hive/pull/1432/files#diff-3054a0cfe6b1c7c844ae261bfb40f5ad58f26b7c4615079cf3dbff0b034e6c4fL217). I made some changes and also ensured there's a [test-case](https://github.com/openshift/hive/pull/1432/files#diff-ae52ca54b986b416ec91f5fe092abedf45e6fd73b413fc92791620e459fd2a91R540-R564) for this scenario

pkg/controller/dnszone/dnszone_controller.go
suhanime: Same as above - so reconcileDNSProvider is called, and then setErrorConditions checks if there was an error, otherwise sets those error conditions to false. That code block will always execute for a successful reconcile, so I kept the code there in order to limit the update events

pkg/controller/dnszone/dnszone_controller_test.go
suhanime: So there's no test in dnsZone controller that tests the Reconcile itself. There are tests for reconcileDNSProvider and SetConditionsForError. For this particular situation, I needed to call both methods because actuators return an error(if any) via reconcileDNSProvider, whereas SetConditionsForError parses the error and decides how the conditions would be set. It mimics [this](https://github.com/openshift/hive/blob/master/pkg/controller/dnszone/dnszone_controller.go#L216-L217) code block of Reconcile.

pkg/controller/utils/errorscrub.go
suhanime: Just a message to add to the reason (or logs in the future?) so no need to display the request id IMO

pkg/controller/utils/errorscrub.go
suhanime: That's a good idea

pkg/controller/utils/errorscrub_test.go
suhanime: updated

pkg/controller/utils/errorscrub_test.go
suhanime: updated

vendor/modules.txt
suhanime: only another module of `github.com/aws/aws-sdk-go/service/` was included - others were already present, so no go.mod or go.sum changes needed (or were caught as a part of `make vendor`)

2uasimojo:
Changes from my comments look good, but leaving the final approval for Devan.

PR #1434, opened 2021-06-28
RFE: ClusterPool hot spares

2uasimojo:
Add enhancement doc proposing ways to keep Active clusters in a ClusterPool for faster claiming.

[HIVE-1576](https://issues.redhat.com/browse/HIVE-1576)

### DIFF
--- a/docs/enhancements/cluster_pool_hot_spares.md
+++ b/docs/enhancements/cluster_pool_hot_spares.md
@@ -1,0 +1,169 @@
+# Maintaining Running Clusters in a ClusterPool
+
+[HIVE-1576](https://issues.redhat.com/browse/HIVE-1576)
+
+- [Maintaining Running Clusters in a ClusterPool](#maintaining-running-clusters-in-a-clusterpool)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [User Story](#user-story)
+  - [Proposal](#proposal)
+    - [`spec.runningCount`](#specrunningcount)
+    - [Decouple PowerState on Creation](#decouple-powerstate-on-creation)
+    - [ClusterClaim Assignment by `creationTimestamp`](#clusterclaim-assignment-by-creationtimestamp)
+    - [Hibernation Controller Must Account For `hibernateAfter` in Pool Clusters](#hibernation-controller-must-account-for-hibernateafter-in-pool-clusters)
+    - [A Note About Tuning](#a-note-about-tuning)
+  - [Alternatives](#alternatives)
+    - [Don't Hibernate New ClusterDeployments](#dont-hibernate-new-clusterdeployments)
+    - [`runningPercent`](#runningpercent)
+
+## Summary
+
+Support configuring ClusterPools to maintain some number of clusters (`0 <= N <= spec.size`) in Running state.
+
+## Motivation
+
+Today ClusterPools immediately hibernate new clusters.
+This means a ClusterClaim is guaranteed to suffer at least Resuming time before the cluster is available for use.
+While this is still faster than having to create a cluster from scratch, it is a potentially-unnecessary delay.
+As an example, CI runs could be reduced by ~5m if a ClusterClaim were able to grab a Running cluster rather than having to Resume a Hibernating one.
+
+### Goals
+
+Ability to tune a ClusterPool such that a ClusterClaim gets a Running cluster immediately in the common case.
+
+(Stretch goal) Solving world peace.
+
+### User Story
+
+As a consumer of a ([properly tuned](#a-note-about-tuning)) ClusterPool,
+
+I want ClusterClaim to give me a Running cluster immediately in the common case,
+
+So that I can get my work done more quickly.
+
+(Ideally without sacrificing the cost savings of having some clusters in the pool Hibernating.)
+
+## Proposal
+
+### `spec.runningCount`
+
+Add a field to ClusterPool indicating a portion of the pool to maintain in Running state.
+`spec.runningCount` is an integer absolute number of Running clusters to strive for.
+It is constrained to `0 <= runningCount <= Size`.
+(We could enforce this by webhook; or we could update the spec accordingly (explicitly set `runningCount = min(runningCount, size)`); but instead we just follow the pattern of the other limits and use the min of `runningCount` and `size` where appropriate.)
+
+If unspecified, the default behavior is to hibernate all CDs in the pool, as today.
+This corresponds to `runningCount=0`, conveniently the golang nil value of an int.
+
+### Decouple PowerState on Creation
+
+Today we create all new ClusterDeployments with `powerState=Hibernating` and update that to `Running` at claim time.
+With this feature, we will create with `powerState` unset.
+Then in a separate code path:
+- Sort unassigned (installing + ready) CDs by `creationTimestamp`
+- Update the oldest `runningCount` CDs with `powerState=Running`
+- Update the remainder (the newest `size - runningCount`) with `powerState=Hibernating`
+
+To be safe, and to continue correctly supporting pools with no/zero `runningCount`, we will still always set `powerState=Running` when assigning a claim.
+
+(In both cases, we should only perform an actual `Update()` for a given CD if its state is changed.)
+
+(Note: This means it is possible for a CD's PowerState to be updated several times even while it's installing.
+That's fine -- though we should make sure the hibernation controller handles it sanely.)
+
+### ClusterClaim Assignment by `creationTimestamp`
+
+When determining which ClusterDeployments to assign to new claims, we can simply sort the list by `creationTimestamp`.
+Since we're [prioritizing PowerState the same way](#decouple-powerstate-on-creation), this should serve to satisfy both:
+- FIFO behavior. Oldest clusters will be claimed first.
+- Minimum response time.
+  We always activate the oldest cluster, and we always claim the oldest cluster, so the first claim will pick up the least-recently-resumed cluster, which is the most likely to be Running, or will be there the soonest.
+
+Caveats:
+- Due to variability in install or resume time, it is technically possible that there would be a "better" (faster) choice for a given assignment.
+  But it'll be real close; and we prefer to code more strictly to the FIFO behavior.
+- Manually resuming a pool CD will not cause it to be claimed earlier.
+  In fact, it is subject to being re-hibernated by the controller.
+  This is an acceptable design point: You can manage power states by configuring your ClusterPool; don't expect to be able to do so manually.
+
+### Hibernation Controller Must Account For `hibernateAfter` in Pool Clusters
+
+If the ClusterPool has `hibernateAfter` set, that clock starts on a given ClusterDeployment from the time its
+[Hibernating condition is Running](https://github.com/openshift/hive/blob/28f67bab3dac5fb9dec405e0536d64187faf964d/pkg/controller/hibernation/hibernation_controller.go#L240-L244);
+or, if the status of that condition is Unknown, from
+[when installation completed](https://github.com/openshift/hive/blob/28f67bab3dac5fb9dec405e0536d64187faf964d/pkg/controller/hibernation/hibernation_controller.go#L229-L239).
+If we don't account for this, Running unassigned CDs in the pool will hibernate, and then the [new logic](#decouple-powerstate-on-creation) will immediately wake them back up, resulting in unnecessary churn and potential claim delays.
+So we'll:
+- Add Status field to ClusterDeployment called `ClaimedTimestamp`, corollary to `InstalledTimestamp`.
+  We'll set this to `Now()` at the same time we set the `Spec.ClusterPoolRef.ClaimName`.
+  Like `InstalledTimestamp` (and `ClaimName`, for that matter), once set, it never changes.
+- Add logic to the hibernation controller as follows:
+  - If the CD is part of a ClusterPool
+    - If it is unclaimed, ignore `hibernateAfter`* -- never hibernate.
+    - If it is claimed, use as the base time the _later_ of
+      - the `ClaimedTimestamp`
+      - the existing algorithm's base time (this accounts for hibernating/resuming during the normal post-claim lifecycle of the CD).
+  - Else (not part of a ClusterPool) do what we do today.
+
+*This also means a manually-resumed unclaimed CD from a pool will never hibernate at the behest of `hibernateAfter`.
+As noted [above](#clusterclaim-assignment-by-creationtimestamp), you shouldn't be trying to manage the PowerState of pool-owned CDs manually.
+This goes double if you're using `hibernateAfter` with ClusterPools.
+
+### A Note About Tuning
+
+Proper tuning will vary wildly depending on the consumer, of course.
+We're giving them the knob to tune with.
+As a starting point, perhaps:
+```
+runningCount = round_up(claim_frequency * install_time)
+```
+For example, if your pool sees one claim every 15 minutes, and it takes 40 minutes to install a cluster:
+```
+runningCount = round_up(1/15 * 40) = 3
+```
+Further tuning would depend on e.g.
+- How "smoothly" claims come in. If your average of 1/15m results from 11 per hour from 9a-5p and 8 total overnight, maybe you want to use the 11/h (`runningCount = 8`) to ensure you can handle peak demand.
+- Whether your priority is cost savings or responsiveness. If the former, perhaps you want to lean towards the 0.5/h (`runningCount = 1`) -- or not use a `runningCount` at all.
+- How much it "costs" you to retune the pool frequently. E.g. if you have the means for automation and a predictable usage curve, perhaps you configure a job to set `runningCount = 8` at 8:50am M-F and `runningCount = 0` at 5p M-F.
+
+
+## Alternatives
+
+The following alternatives were considered:
+
+### Don't Hibernate New ClusterDeployments
+
+Today when ClusterDeployments are created to fill out the pool, they are created with `PowerState = HibernatingClusterPowerState`.
+Add an option to ClusterPool to not do that.
+
+**Pros:**
+- Dead simple
+
+**Cons:**
+- In periods of low activity, you're burning money by having Running vs. Hibernating clusters.
+  (I suppose you could manually hibernate clusters in the pool. Not the best UX.)
+
+### `runningPercent`
+
+Use a percentage instead of an absolute number to indicate how many clusters in the pool should remain Running.
+
+Quoting a conversation with [Gurney Buchanan](https://github.com/gurnben) (ACM):
+
+> Eric Fried:
+> The main decision being: by percentage or by count?
+>
+> Gurney Buchanan:
+> Hmm - we use small clusterpools so I would wager count (percentages would get messy) - plus with count, you can know (without maths):
+> - How many running instances you’ll have in the pool
+> - How much vCPU you need available (for future vsphere pools)
+> - Control cost
+> - **Critically** you can control cost/number of cold spares without resizing the pools.  If a prod deploy needs 3 clusters for example and you know you might need to roll one at any time, you can keep 3 hot spares and as long as another deploy doesn’t have to happen in the next 5 minutes you’ll have hot spares for the next,  and the user can tune the size of the pool to ensure that they always have enough clusters ready for deploys to cycle.
+>
+> Yepp I’m convinced it should be count
+>
+> Eric Fried:
+> I had been thinking percentages could be more convenient if you're frequently resizing the pool -- but compared to what you've said, the minor inconvenience of having to edit two fields instead of one when resizing is insignificant.
+>
+> Gurney Buchanan:
+> That’s what I realized as I was typing up the use-case @Eric Fried :laugh1:


docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,111 @@
...
+(Ideally without sacrificing the cost savings of having some clusters in the pool hibernating.)
+
+## Proposals
+
+Several implementation alternatives have already come to mind.
+These are not necessarily mutually exclusive.
+
+### ClusterClaim Prioritizes Active Clusters
+
+For most/all of the options below, it makes sense to add logic to ClusterClaim to preferentially grab an Active cluster first.
```
petr-muller, line 57: With sufficiently bad claim patterns this can lead to oldest hibernating clusters to not being claimed for prolonged time period. Not sure if that's a problem on its own, but we're seeing problems with clusters that spend too long in hibernation after installing (24h+), and claims not being FIFO could increase the odds of that.

2uasimojo: I'm curious what problems you've seen with old clusters getting stale.

There's a known bug (yet to be filed) where, if you change the ClusterPool's spec in ways that result in different ClusterDeployments, you're now lying about the CDs that were already in the pool. Is it that, or something else?

petr-muller: No, it was the clusters massively restarting (thought to be cert rotation) after hibernated for 24h+: https://coreos.slack.com/archives/CE3ETN3J8/p1624379289071000?thread_ts=1624376710.056300&cid=CE3ETN3J8

petr-muller: HIVE-1570

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,111 @@
...
+The simplest implementation is to sort the list of ClusterDeployments into three buckets: Active, Resuming, and Hiberating, claimed in that order.
+A smarter implementation could take into account when Resuming clusters were asked to resume, with the most recent at the end, on the theory that the least-recently-resumed will be ready first.
+
+### ClusterClaim Must Account For `hibernateAfter`
+
+This applies to most/all of the options below.
+
+If the ClusterPool has `hibernateAfter` set, that clock starts running as soon as the ClusterDeployment is ready.
+(TODO: Confirm this -- does the clock start at creation, or `installed == True`, or something else?)
```
2uasimojo, line 66: Today, the hibernation_controller adds `hibernateAfter` to a base time to decide whether to hibernate.
To determine the base time, it looks for the `Hibernating` Condition on the CD.
- If `Unknown`, uses the `status.installedTimestamp` (when installation completed).
- If `False`, uses the `lastTransitionTime` of the Condition (when it woke up from the most recent previous hibernation).

What we could do here is add another Condition, e.g. `Claimed`, indicating claim assignment status. (We already have the equivalent on the claim.) Then change the above to:
- If the CD is part of a ClusterPool
  - If it is unclaimed, ignore `hibernateAfter` [1] -- never hibernate.
  - If it is claimed, use as the base time the _later_ of
    - the `lastTransitionTime` of the new `Claimed` Condition
    - the above algorithm's base time (this accounts for hibernating/resuming during the normal post-claim lifecycle of the CD).
- Else the above algorithm as is

[1] This also means a manually-resumed unclaimed CD from a pool will never hibernate at the behest of `hibernateAfter`. IMO we can live with that. If you're using `hibernateAfter` with ClusterPools, you really oughtta let the code manage things.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,111 @@
...
+Add an option to ClusterPool to not do that.
+
+**Pros:**
+- Dead simple
+
+**Cons:**
+- In periods of low activity, you're burning money by having Active vs. Hibernating clusters.
+  (I suppose you could manually hibernate clusters in the pool. Not the best UX.)
+
+### `activeCount`/`activePercent`
```
2uasimojo, line 86: @gurnben convinced me in [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1627311417155200?thread_ts=1627057481.125300&cid=CE3ETN3J8) that it should be `activeCount`.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,111 @@
...
+Add knob(s) to ClusterPool indicating a portion of the pool to maintain in Active state.
+Either or both (if both, they should be mutually exclusive, or one should take precedence) of:
+
+- `activeCount`: Integer absolute number of Active clusters to strive for.
+  Constrained to `0 (default) <= activeCount <= (Max?)Size`.
+- `activePercent`: Float `0.0 (default) <= activePercent <= 100.0` percentage of `(Max?)Size` clusters to keep Active.
+  (How should we round this?)
+
+Either way, this translates to an integer number of clusters the ClusterPool controller will try to keep Active.
+- When creating a new cluster for the pool, set `PowerState = HibernatingClusterPowerState` iff we're at/above that threshold.
```
2uasimojo, line 97: We need to be careful about this algorithm so we don't end up LIFOing CDs. Viz:
- Pool size is 3, activeCount is 1. We set up 2 hibernating and 1 active
- Claim 1. You get the active one.
- We spin up a new CD to bring the count back to 3. We check against active count to figure out whether to hibernate the new one or not. We decide not.

So the newest one is the active one. Repeat. The two hibernating ones never get used.

I think to mitigate this, we need to fully decouple hibernation state from new CD creation. In the reconcile loop, *after* we've done assignments and kicked off any installations, we calculate how many clusters need to be hibernated or resumed. We decide which ones based on overall age (`creationTimestamp`), which means we'll prioritize hibernating Installing clusters and resuming the oldest Hibernating clusters.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: Latest rev of the doc enforces strict FIFO by claiming in `creationTimestamp` order.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,111 @@
...
+## Proposals
+
+Several implementation alternatives have already come to mind.
+These are not necessarily mutually exclusive.
+
+### ClusterClaim Prioritizes Active Clusters
+
+For most/all of the options below, it makes sense to add logic to ClusterClaim to preferentially grab an Active cluster first.
+The simplest implementation is to sort the list of ClusterDeployments into three buckets: Active, Resuming, and Hiberating, claimed in that order.
+A smarter implementation could take into account when Resuming clusters were asked to resume, with the most recent at the end, on the theory that the least-recently-resumed will be ready first.
```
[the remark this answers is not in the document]
2uasimojo: Made it simpler than that: strict FIFO by sorting by `creationTimestamp`.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: Added to doc.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: Updated doc accordingly.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: Updated doc accordingly.

gurnben:
Everything looks good to me @2uasimojo - you've done a particularly good job of enumerating the edge cases, reasoning, and alternatives!  We're approaching a point where we might even keep a hot spare or two to speed up CI!

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+
+   So that I can get my work done more quickly.
+
+   (Ideally without sacrificing the cost savings of having some clusters in the pool Hibernating.)
+
+2. As a user of a (hypothetical future) ClusterDeployment (or sibling CRD) for a cluster type that doesn't support hibernation,
+
+   I want to be able to configure a ClusterPool that never hibernates its ClusterDeployments
+
+   So that I can use ClusterPools for that kind of ClusterDeployment.
```
abhinavdahiya, line 54: hmm.. i don't think the clusterpool api mentions that the clusters it maintains have to be in hibernating state. So for a type of cd that doesn't support hibernation it can continue to maintain them as running.
so this user story imo is not necessary.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Proposal
+
+### `spec.runningCount`
+
+Add a field to ClusterPool indicating a portion of the pool to maintain in Running state.
+`spec.runningCount` is an integer absolute number of Running clusters to strive for.
+It is constrained to `0 <= runningCount <= Size`.
+We could enforce this by webhook; or we could just update the spec accordingly.
```
abhinavdahiya, line 63: > we could just update the spec accordingly.

can you clarify what you mean by this?

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+
+If unspecified, the default behavior is to hibernate all CDs in the pool, as today.
+This corresponds to `runningCount=0`, conveniently the golang nil value of an int.
+
+### Decouple PowerState on Creation
+
+Today we create all new ClusterDeployments with `powerState=Hibernating` and update that to `Running` at claim time.
+With this feature, we will create with `powerState` unset.
+Then in a separate code path:
+- Sort unassigned (installing + ready) CDs by `creationTimestamp`
```
abhinavdahiya, line 73: why do we include the installing cds in this? those can be acted upon when they are installed right?

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+- Update the oldest `runningCount` CDs with `powerState=Running`
+- Update the remainder (the newest `size - runningCount`) with `powerState=Hibernating`
```
abhinavdahiya, line 75: with this flow, some previously hibernated clusters will be made to transition to match the running count.
how do we handle case where the cluster fails to go from hibernated -> running ?

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+While this is still faster than having to create a cluster from scratch, it is a potentially-unnecessary delay.
+As an example, CI runs could be reduced by ~5m if a ClusterClaim were able to grab a Running cluster rather than having to Resume a Hibernating one.
+
+### Goals
+
+Ability to tune a ClusterPool such that a ClusterClaim gets a Running cluster immediately in the common case.
+
+### Non-Goals
+
+Solving world peace.
```
dgoodwin, line 37: Disagree. This is needlessly negative. If we solve world peace with hot spares then so be it. And we might.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+  We always activate the oldest cluster, and we always claim the oldest cluster, so the first claim will pick up the least-recently-resumed cluster, which is the most likely to be Running, or will be there the soonest.
+
+Caveats:
+- Due to variability in install or resume time, it is technically possible that there would be a "better" (faster) choice for a given assignment.
+  But it'll be real close; and we prefer to code more strictly to the FIFO behavior.
+- Manually resuming a pool CD will not cause it to be claimed earlier.
+  In fact, it is subject to being re-hibernated by the controller.
+  This is an acceptable design point: You can manage power states by configuring your ClusterPool; don't expect to be able to do so manually.
+
+### Hibernation Controller Must Account For `hibernateAfter` in Pool Clusters
```
abhinavdahiya, line 99: what if we look at a general example?

the clusterdeployment starts it's life without hibernateAfter but later on at some point the owner adds the hibernateAfter ..
should the cluster shutdown from the time the hibernateAfter was set or from the last running/installed?

if we think the first hibernation should happen after the duration when hibernateAfter was set, i think we should move towards that change and have the clusterpool set the hibernateAfter when the cluster was assigned to the claim.

docs/enhancements/cluster_pool_hot_spares.md
dgoodwin: Can we handle that generically in https://issues.redhat.com/browse/HIVE-1602

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+**Pros:**
+- Dead simple
+
+**Cons:**
+- In periods of low activity, you're burning money by having Running vs. Hibernating clusters.
+  (I suppose you could manually hibernate clusters in the pool. Not the best UX.)
+
+### `runningPercent`
+
+Use a percentage instead of an absolute number to indicate how many clusters in the pool should remain Running.
```
abhinavdahiya, line 142: the kubeapis have IntOrString type that allows users to decide whether they want to use exact numbers or percentages.
for example see https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/#DeploymentSpec `strategy.rollingUpdate.maxSurge`

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,162 @@
...
+Ability to tune a ClusterPool such that a ClusterClaim gets a Running cluster immediately in the common case.
+
+### Non-Goals
+
+Solving world peace.
+
+
+### User Stories
+
+1. As a consumer of a (properly tuned) ClusterPool,
```
abhinavdahiya, line 42: i think it makes sense for defining or prescribing what a properly tuned clusterpool will look like given the use case from user. esp in terms of having the runningCount set to X, what does it really mean when users are claiming from this clusterpool.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: That's neat. Though (Gurney and) I feel the reasons for not choosing to use a percentage stand on their own, beyond the logistics of having to deal with the data types.

docs/enhancements/cluster_pool_hot_spares.md
abhinavdahiya: > Punting that knowledge to the consumer via this knob would be a more flexible answer. E.g. hibernation for Zaphod clusters was added at version 3.0. Consumer configures their Zaphod 2.5 pool with runningCount==size but their 3.x pools with runningCount=0.

to me this having the user remember that size matches runningCount for cluster types that do not support hibernation is bad.
- because it allows the user to setup clusterpools that will cause errors in the hive controllers. invalid inputs should be rejected fast instead of waiting for errors. the error reporting from cluster fleet is already difficult because of noise.
- having the users know whether their configuration will support hibernation just so that the hive controller doesn't have to know that seems like bad user experience.

i personally feel that this is not a good user story. And it does not sell the runningCount feature anymore than the first use-case

docs/enhancements/cluster_pool_hot_spares.md
abhinavdahiya: > Editing the power state of an installing cluster is cheap

This is based on the assumption that the hibernation controller does not act on the powerState request unless the cluster is installed.
i think a) we need to solidify that this behavior that powerState change is allowed on installing clusters b) hibernation controller only looks at changing state of installed clusters else no-ops

without those two, it seems like usage of an undefined behavior like the `partway hibernating`

docs/enhancements/cluster_pool_hot_spares.md
abhinavdahiya: > I still hold that this entails a mostly-unrelated change to the design/contract of hibernateAfter and is out of scope for this design.

the current recommendation includes updating the hibernation controller to understand what clusters are claimed/or not. I don't think the hibernation controller will be well served to understand a concept that it technically has no reason knowing except that the clusterpool controller sets hibernateAfter but expects that the timer start when clusterpool controller decide.

my recommendation is that we solve this by updating the hibernation controller to support a behavior in general sense where the user wants the timer for hibernation to start when it is set vs always bounded on installation complete of the cluster.

and the clusterpool controller use that behavior to set the hibernateAfter when it is assigning that cluster to a claim.

docs/enhancements/cluster_pool_hot_spares.md
abhinavdahiya: also if this usecase is that important, i think we should make changes in the proposal that would help improve the UX of size == runningCount . for example supporting percentages in addition to count will make the pool set once and forget by setting runningCount to 100%

2uasimojo: I've removed the user story.

However:

> having the users know whether their configuration will support hibernation just so that the hive controller doesn't have to know that seems like bad user experience.

I don't disagree; but this assumes hive *can* know. If hive *can't* know (which I contend will be the case at least some of the time) then the user is helpless.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: > assumption that the hibernation controller does not act on the powerState request unless the cluster is installed

General principle of idempotence / eventual consistency. If it doesn't behave nicely in these circumstances, it's a bug.

However, I'm liking the idea of reorganizing when we do what. I was talking to @suhanime this morning about [fixing broken claims (HIVE-1602)](https://issues.redhat.com/browse/HIVE-1602) and we discussed something like:

**Don't consider a CD assignable until it is Running.**

which, in the absence of `runningCount`, entails:

**For each unassigned claim, strive for an equal number of Running CDs.**

So we need to subdivide Installed, non-Deleting CDs by Hibernation status:
- **Running**: These can be assigned to unassigned claims.
- **Resuming**: These count against the number of unassigned claims, but aren't yet ready for assignment.
- **Hibernating**: These are eligible to be kicked over to `PowerState: Running` to (eventually) satisfy claims. We will do so when `numUnassignedClaims > (numResumingCDs + numRunningCDs)`

(Note that with this logic we can still starve claims if CDs get stuck Resuming -- but that's what HIVE-1602 is going to address.)

Factoring in `runningCount` simply means we strive for `numUnassignedClaims + runningCount` active CDs instead of just `numUnassignedClaims`.

Thoughts?

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: I can get behind the idea of waiting until we assign the CD to set its `hibernateAfter`.

I'm still not seeing a (non-ugly) way of keeping track of when that field was set -- or, for that matter, edited. E.g. if I set the field to 4h, then an hour goes by, then I edit the field from 4h to 2h, do I want my cluster to hibernate in 1h (2h from when the field was initially set) or 2h (2h from when it was updated)? And again, any answer that isn't "2h from installation / the last time the cluster woke up" (the current behavior) represents an API change, and I don't know that we can get away with that.

Suggestions?

docs/enhancements/cluster_pool_hot_spares.md
abhinavdahiya: > General principle of idempotence / eventual consistency. If it doesn't behave nicely in these circumstances, it's a bug.

Well technically there is a case to be made that such an Update should be rejected, and not eventual consistency (you can't hibernate something that hasn't reached the installed state in the first place). So unless this behavior is documented in the API, i don't think it's safe to make this assumption.

docs/enhancements/cluster_pool_hot_spares.md
abhinavdahiya: For now, solving the issue with some tie between clusterpool and hibernation seems like acceptable.
We can look at hibernateAfter sometime later on, it seems like too much of behavior change for this enhancement.

abhinavdahiya:
LGTM

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,171 @@
...
+(Ideally without sacrificing the cost savings of having some clusters in the pool Hibernating.)
+
+## Proposal
+
+### `spec.runningCount`
+
+Add a field to ClusterPool indicating a portion of the pool to maintain in Running state.
+`spec.runningCount` is an integer absolute number of Running clusters to strive for.
+It is constrained to `0 <= runningCount <= Size`.
+We could enforce this by webhook; or we could just update the spec accordingly (explicitly set `runningCount = min(runningCount, size)`).
```
dgoodwin, line 54: Typically webhook in our code, the alternative could be a little confusing and unexpected for users.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,171 @@
...
+If unspecified, the default behavior is to hibernate all CDs in the pool, as today.
+This corresponds to `runningCount=0`, conveniently the golang nil value of an int.
+
+### Decouple PowerState on Creation
+
+Today we create all new ClusterDeployments with `powerState=Hibernating` and update that to `Running` at claim time.
+With this feature, we will create with `powerState` unset.
+Then in a separate code path:
+- Sort unassigned (installing + ready) CDs by `creationTimestamp`
+- Update the oldest `runningCount` CDs with `powerState=Running`
```
dgoodwin, line 65: Will this conflict with the stale cluster deletion, which is also based on oldest?

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,171 @@
...
+- Due to variability in install or resume time, it is technically possible that there would be a "better" (faster) choice for a given assignment.
+  But it'll be real close; and we prefer to code more strictly to the FIFO behavior.
+- Manually resuming a pool CD will not cause it to be claimed earlier.
+  In fact, it is subject to being re-hibernated by the controller.
+  This is an acceptable design point: You can manage power states by configuring your ClusterPool; don't expect to be able to do so manually.
+
+### Hibernation Controller Must Account For `hibernateAfter` in Pool Clusters
+
+If the ClusterPool has `hibernateAfter` set, that clock starts on a given ClusterDeployment from the time its
+[Hibernating condition is Running](https://github.com/openshift/hive/blob/28f67bab3dac5fb9dec405e0536d64187faf964d/pkg/controller/hibernation/hibernation_controller.go#L240-L244);
```
dgoodwin, line 93: Will soon be a Running condition = true for this.

docs/enhancements/cluster_pool_hot_spares.md
```diff
@@ -0,0 +1,171 @@
...
+runningCount = round_up(claim_frequency * install_time)
+```
+For example, if your pool sees one claim every 15 minutes, and it takes 40 minutes to install a cluster:
+```
+runningCount = round_up(1/15 * 40) = 3
+```
+Further tuning would depend on e.g.
+- How "smoothly" claims come in. If your average of 1/15m results from 11 per hour from 9a-5p and 8 total overnight, maybe you want to use the 11/h (`runningCount = 8`) to ensure you can handle peak demand.
+- Whether your priority is cost savings or responsiveness. If the former, perhaps you want to lean towards the 0.5/h (`runningCount = 1`) -- or not use a `runningCount` at all.
+- How much it "costs" you to retune the pool frequently. E.g. if you have the means for automation and a predictable usage curve, perhaps you configure a job to set `runningCount = 8` at 8:50am M-F and `runningCount = 0` at 5p M-F.
```
dgoodwin, line 130: A histogram metric tracking time from claim request to fulfillment would be great to see in heavily loaded envs, and would help inform if you've got enough of a running count or not.

dgoodwin (commented):
This looks pretty good to me

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: Yes (which wasn't a thing when I wrote this :) )

Today we don't take staleness into account when assigning: we'll assign the oldest installed cluster even if it's stale; but then that takes that cluster out of consideration for deletion due to staleness.

If we don't treat stale clusters specially:
- Kick the oldest cluster, which happens to be stale, to running
- Delete the oldest stale cluster, which is ^ that one
- Replace that cluster. While it's installing...
- Kick the oldest cluster, which happens to be stale, to running

We won't delete another stale cluster until the replacement is done installing. As long as a claim comes in before then, we'll assign the now-running stale cluster that was the second-oldest when we started. If not, we'll repeat the above sequence. It's not really thrashy since it's only happening at worst on a ~40m interval.

Is this something you feel we need to solve?

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: I like it.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: > "ClusterPool should handle (trash and replace) broken CDs" -- which can really happen for any number of reasons, not just hibernate/resume.

One such reason being ProvisionStopped, per [HIVE-1615](https://issues.redhat.com/browse/HIVE-1615)/#1524.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: This all felt wrong as I started prototyping. This section proposes a new Condition on ClusterDeployment just so we can tell the claim time.

As noted here, the times the controller looks at today are last resume and when installation completed.
Last resume is in a Condition.
Installation completion is directly in a status field.
Resuming is a thing that can happen over and over.
Installation completes once; that field should never get updated again.

Claiming, like install completion, is a thing that happens once. So it makes more sense for it to be a status field.

So I'd like to change this to propose `status.claimedTimestamp` instead, and use it if set.

(It also occurs to me that I should describe what happens on "upgrade": We'll never set status.claimedTimestamp for CDs that have flown the coop, but that's okay because they also wouldn't have been subject to `runningCount`, so would have been resumed from hibernating when claimed, so would be using the last resumed time as the baseline for `hibernateAfter` anyway.)

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: ...but it seems only partially related to this feature, so I'd like to keep it separate. I've opened HIVE-1643 to track it.

docs/enhancements/cluster_pool_hot_spares.md
2uasimojo: I think the important thing to note here is that we're preserving all of these rules separately (because trying to coordinate them would be insanely complicated):
- Claiming gets the oldest ready cluster -- FIFO
- runningCount hits the oldest ready clusters -- again for FIFO purposes so a given :arrow_up: claim is most likely to get a running cluster
- Delete the oldest stale cluster first -- again for FIFO purposes.

PR #1437, opened 2021-06-29
handle the possibility that an error isn't an awsError

joelddiaz:
Need to properly account for an error not being an awsError. Update the
code flow to handle these cases so that the right conditions are being
set/cleared.

Update test cases to exercise this non-AWS error codepath.


pkg/controller/dnszone/dnszone_controller_test.go
```diff
@@ -973,16 +1040,25 @@ func TestSetConditionsForErrorForAWS(t *testing.T) {
...
+					for _, cond := range zr.dnsZone.Status.Conditions {
+						if expectedCondition.Type == cond.Type {
+							foundCondition = true
+							assert.Equal(t, cond.Status, expectedCondition.Status)
+							assert.Equal(t, cond.Reason, expectedCondition.Reason)
+							assert.Equal(t, cond.Message, expectedCondition.Message)
+						}
+
+					}
+					assert.True(t, foundCondition, "Did not find expected condition of type %s", expectedCondition.Type)
```
suhanime, line 1055: Can you change it to [this](https://github.com/openshift/hive/blob/master/pkg/test/assert/assertions.go#L68-L77) pattern? We don't want to ever be reliant on the sequence of conditions when we check for them since conditions can be applied in any order

PR #1438, opened 2021-06-29
Fix panic in dnszone controller

suhanime:
While trying to get the actuator in dnszone controller, code was trying
to parse a re-assigned error variable, causing a panic

/assign @joelddiaz


joelddiaz (commented):
/lgtm

PR #1439, opened 2021-06-30
hibernation: include machines in undesired state when hibernating/resuming

abhinavdahiya:
Currently when hibernation is in progress, the controller checks if all
machines for the cluster are in desired state (stopped when hibernating
and running when running). If all machines are not in that state, the
controller returns and retries after some timeout.
This means that there is no information from the controller about the
progress made and machines in mis-matched state.

With this change, when machines are in mis-matched state, we update the
Hibernating condition to include a message listing those machines.

xref: https://issues.redhat.com/browse/HIVE-1564

/assign @joelddiaz 
/cc @2uasimojo

### DIFF
--- a/pkg/controller/hibernation/aws_actuator.go
+++ b/pkg/controller/hibernation/aws_actuator.go
@@ -95,35 +95,37 @@ func (a *awsActuator) StartMachines(cd *hivev1.ClusterDeployment, c client.Clien
 }
 
 // MachinesRunning will return true if the machines associated with the given
-// ClusterDeployment are in a running state.
-func (a *awsActuator) MachinesRunning(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) (bool, error) {
+// ClusterDeployment are in a running state. It also returns a list of machines that
+// are not running.
+func (a *awsActuator) MachinesRunning(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) (bool, []string, error) {
 	logger = logger.WithField("cloud", "aws")
 	logger.Infof("checking whether machines are running")
 	awsClient, err := a.awsClientFn(cd, c, logger)
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
 	instanceIDs, err := getClusterInstanceIDs(cd, awsClient, notRunningStates, logger)
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
-	return len(instanceIDs) == 0, nil
+	return len(instanceIDs) == 0, aws.StringValueSlice(instanceIDs), nil
 }
 
 // MachinesStopped will return true if the machines associated with the given
-// ClusterDeployment are in a stopped state.
-func (a *awsActuator) MachinesStopped(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) (bool, error) {
+// ClusterDeployment are in a stopped state. It also returns a list of machines
+// that have not stopped.
+func (a *awsActuator) MachinesStopped(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) (bool, []string, error) {
 	logger = logger.WithField("cloud", "aws")
 	logger.Infof("checking whether machines are stopped")
 	awsClient, err := a.awsClientFn(cd, c, logger)
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
 	instanceIDs, err := getClusterInstanceIDs(cd, awsClient, notStoppedStates, logger)
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
-	return len(instanceIDs) == 0, nil
+	return len(instanceIDs) == 0, aws.StringValueSlice(instanceIDs), nil
 }
 
 func getAWSClient(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) (awsclient.Client, error) {
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -3,6 +3,8 @@ package hibernation
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/blang/semver/v4"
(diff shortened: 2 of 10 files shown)


pkg/controller/hibernation/aws_actuator.go
```diff
@@ -96,34 +96,34 @@ func (a *awsActuator) StartMachines(cd *hivev1.ClusterDeployment, c client.Clien
...
-		return false, err
+		return false, nil, err
 	}
 	instanceIDs, err := getClusterInstanceIDs(cd, awsClient, notRunningStates, logger)
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
-	return len(instanceIDs) == 0, nil
+	return len(instanceIDs) == 0, aws.StringValueSlice(instanceIDs), nil
```
2uasimojo, line 110: Note: `StringValueSlice` will never return `nil`. If `instanceIDs` is `nil` or zero length, it will return `[]string{}`. Same applies to all impls for both funcs.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -172,7 +172,7 @@ func TestReconcile(t *testing.T) {
 			cd:   cdBuilder.Options(o.shouldHibernate, o.stopping).Build(),
 			cs:   csBuilder.Build(),
 			setupActuator: func(actuator *mock.MockHibernationActuator) {
-				actuator.EXPECT().MachinesStopped(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(true, nil)
+				actuator.EXPECT().MachinesStopped(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(true, nil, nil)
```
2uasimojo, line 175: nit: As noted above, `MachinesStopped` will never return `nil` for the second return value (except in error cases). It totally doesn't affect anything, but if you wanted to be pedantic:
```suggestion
				actuator.EXPECT().MachinesStopped(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(true, []string{}, nil)
```
Same applies for `MachinesRunning` below.

2uasimojo (commented):
IIUC, with this change, seeing this new condition message would be a normal part of the hibernation/resuming process, right? Because the initial "Stopping cluster machines" message is set the first time the controller reconciles after the PowerState is changed; but then anything that triggers another CD reconcile before the hibernation finishes will cause the message to change to "Some machines have not yet stopped: X,Y,Z".

I wonder if that's confusing UX. Might it be better for the message to be the same throughout the process? E.g.
- get rid of the `stopMachines` func entirely, and use `checkClusterStopped` (possibly renamed? `ensureClusterStopped`?) instead, since it's now effectively a superset anyway
- adjust the message to say something like, "Stopping cluster machines. The followng are not yet stopped: X,Y,Z"

Not blocking on this, just food for thought.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -348,7 +365,8 @@ func (r *hibernationReconciler) checkClusterResumed(cd *hivev1.ClusterDeployment
 		logger.Warning("No compatible actuator found to check machine status")
 		return reconcile.Result{}, nil
 	}
-	running, err := actuator.MachinesRunning(cd, r.Client, logger)
+
+	running, notRunning, err := actuator.MachinesRunning(cd, r.Client, logger)
```
joelddiaz, line 369: Same comment about the var names

pkg/controller/hibernation/hibernation_controller_test.go
abhinavdahiya: for consumers of slices, both should have the same behavior. So i would rather just keep it nil for clarity sake.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -330,14 +332,29 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 		logger.Warning("No compatible actuator found to check machine status")
 		return reconcile.Result{}, nil
 	}
-	stopped, err := actuator.MachinesStopped(cd, r.Client, logger)
+
+	stopped, notStopped, err := actuator.MachinesStopped(cd, r.Client, logger)
```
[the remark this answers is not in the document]
abhinavdahiya: switched to using remaining for the list. i usually prefer not duplicating type in the variable names and i thinking remaining makes the usage clear enough. wdyt?

abhinavdahiya:
> get rid of the stopMachines func entirely, and use checkClusterStopped (possibly renamed? ensureClusterStopped?) instead, since it's now effectively a superset anyway

Yes, that's true. but i didn't want to change the code flow for this PR that makes sure condition messages are clearer. Removing that code will affect the unit tests in a way i was not comfortable handling here. maybe in a separate pr.

> adjust the message to say something like, "Stopping cluster machines. The followng are not yet stopped: X,Y,Z"

the condition reason is already Resuming/Stopping and with that context, including "Stopping cluster machines" just seems more duplication imo. But i don't mind including that if that makes it easy for users.

2uasimojo:
Updates lgtm, thanks @abhinavdahiya.

> the condition reason is already Resuming/Stopping and with that context, including "Stopping cluster machines" just seems more duplication imo.

(Not asking for changes, just trying to explain my thinking. Feel free to ignore.)

It's in the [initial message](https://github.com/openshift/hive/blob/bf40780e4046c623096411568b9473ab117981b8/pkg/controller/hibernation/hibernation_controller.go#L326). I was thinking about the flow where, as a user, I'm watching the Condition: If I see it transition from

`Stopping / True / "Stopping cluster machines"`

to

`Stopping / True / "Some machines have not yet stopped: X,Y,Z"`

...I might think there's a Problem.

Put another way, if we're worried about redundancy, perhaps the initial message shouldn't say anything :)

But this is probably just me still being uncomfortable with how Conditions are structured.

pkg/controller/hibernation/hibernation_controller.go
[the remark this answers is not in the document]
2uasimojo: +1 from me. (FWIW I had the same thought as Joel when I read the PR originally.)

joelddiaz (commented):
:+1:

dgoodwin:
Great UX improvement, thanks!

2uasimojo:
/test e2e

infra timeout

PR #1444, opened 2021-07-06
Document cluster adoption

dgoodwin:
x-ref: https://issues.redhat.com/browse/HIVE-1566

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -10,7 +10,6 @@
     - [Cloning the repository](#cloning-the-repository)
   - [Deploying with Kubernetes In Docker (kind)](#deploying-with-kubernetes-in-docker-kind)
   - [Deploying with kind and Podman (EXPERIMENTAL)](#deploying-with-kind-and-podman-experimental)
-  - [Adopting ClusterDeployments](#adopting-clusterdeployments)
   - [Writing/Testing Code](#writingtesting-code)
     - [Run Hive Operator](#run-hive-operator)
       - [Directly from source](#directly-from-source)
@@ -261,24 +260,6 @@ NOTE: The cluster name and domain used to create the certificate must match the 
 
 Example:
 `hiveutil create-cluster mycluster --serving-cert=$HOME/mycluster.crt --serving-cert-key=$HOME/mycluster.key`
-
-## Adopting ClusterDeployments
-
-It is possible to adopt cluster deployments into Hive, potentially even fake or kind clusters. This can be useful for developers who would like to work on functionality separate from actual provisioning.
-
-To create a kind cluster and adopt:
-
-```bash
-./hack/create-kind-cluster.sh cluster1
-bin/hiveutil create-cluster --base-domain=new-installer.openshift.com kind-cluster1 --adopt --adopt-admin-kubeconfig=/path/to/cluster/admin/kubeconfig --adopt-infra-id=fakeinfra --adopt-cluster-id=fakeid
-```
-
-NOTE: when using a kind cluster not all controllers will be functioning properly as it is not an OpenShift cluster and thus lacks some of the CRDs our controllers use. (ClusterState, RemoteMachineSet, etc)
-
-Alternatively you can use any valid kubeconfig for live or since deleted clusters.
-
-Deprovision will run but find nothing to delete if no resources are tagged with your fake infrastructure ID.
-
 
 ## Code editors and multi-module repositories
 
--- a/pkg/remoteclient/remoteclient.go
+++ b/pkg/remoteclient/remoteclient.go
@@ -140,7 +140,7 @@ func InitialURL(c client.Client, cd *hivev1.ClusterDeployment) (string, error) {
 // the ClusterDeployment to determine if the remote cluster is reachable.
 func Unreachable(cd *hivev1.ClusterDeployment) (unreachable bool, lastCheck time.Time) {
 	cond := utils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.UnreachableCondition)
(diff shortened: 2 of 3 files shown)


pkg/remoteclient/remoteclient.go
```diff
@@ -140,7 +140,7 @@ func InitialURL(c client.Client, cd *hivev1.ClusterDeployment) (string, error) {
 // the ClusterDeployment to determine if the remote cluster is reachable.
 func Unreachable(cd *hivev1.ClusterDeployment) (unreachable bool, lastCheck time.Time) {
 	cond := utils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.UnreachableCondition)
-	if cond.Status == corev1.ConditionUnknown {
+	if cond == nil || cond.Status == corev1.ConditionUnknown {
```
2uasimojo, line 143: This seems unrelated, should be a separate PR.

docs/developing.md
```diff
@@ -262,24 +261,6 @@ NOTE: The cluster name and domain used to create the certificate must match the
 Example:
 `hiveutil create-cluster mycluster --serving-cert=$HOME/mycluster.crt --serving-cert-key=$HOME/mycluster.key`
```
2uasimojo, line 263: Should we keep a link to the new doc from here?

docs/cluster-adoption.md
```diff
@@ -0,0 +1,49 @@
...
+## Adopting ClusterDeployments
+
+It is possible to adopt cluster deployments into Hive. To do so you will need to create a ClusterDeployment with Spec.Installed set to True, no Spec.Provisioning section, and include the following:
+
+  * cluster INFRAID (obtained from `oc get infrastructure cluster -o json | jq .status.infrastructureName`)
+  * cluster ID (obtained from `oc get clusterversion version -o json | jq .spec.clusterID`)
+  * reference to a properly formatted admin kubeconfig Secret:  `oc create secret generic mycluster-admin-kubeconfig --from-file=kubeconfig=/tmp/admin.kubeconfig`
+  * Spec.Platform.YourCloudProvider for your cluster, most importantly region and a properly formatted credentials Secret
+
+Use Spec.PreserveOnDelete if you do not want Hive to deprovision resources when the ClusterDeployment is deleted.
```
2uasimojo, line 12:
```suggestion
Use `Spec.PreserveOnDelete = true` if you do not want Hive to deprovision resources when the ClusterDeployment is deleted.
```

docs/cluster-adoption.md
```diff
@@ -0,0 +1,49 @@
...
+      credentialsSecretRef:
+        name: my-gcp-creds
+      region: us-east1
+  pullSecretRef:
+    name: pull-secret
+```
+
+## Adopting with hiveutil
+
+hiveutil is a development focused CLI tool which can be build from the hive repo. To adopt a cluster specify the following flags:
```
2uasimojo, line 42: Link to the hiveutil doc?
```suggestion
[hiveutil](hiveutil.md) is a development focused CLI tool which can be build from the hive repo. To adopt a cluster specify the following flags:
```

docs/cluster-adoption.md
```diff
@@ -0,0 +1,49 @@
...
+
+```bash
+bin/hiveutil create-cluster --base-domain=example.com mycluster --adopt --adopt-admin-kubeconfig=/path/to/cluster/admin/kubeconfig --adopt-infra-id=[INFRAID] --adopt-cluster-id=[CLUSTERID]
```
2uasimojo, line 45: Is there an option to set `PreserveOnDelete` via hiveutil?

...looks like not :(  Should there be?

docs/cluster-adoption.md
```diff
@@ -0,0 +1,49 @@
+# Cluster Adoption
```
2uasimojo, line 1: We should link to this doc from... somewhere. Do we consider adoption a dev-only thing? If so, perhaps the suggested link from developing.md is sufficient. Otherwise maybe from using-hive.md?

2uasimojo: > Do we consider adoption a dev-only thing?

Looking at the [card](https://issues.redhat.com/browse/HIVE-1566) I can see the answer is "not anymore", so...

pkg/remoteclient/remoteclient.go
dgoodwin: We actually crash if you adopt without this, I will retitle the PR.

docs/cluster-adoption.md
dgoodwin: Probably not, we agreed at the outset it was never meant for 1-1 compat with the API, and you could use -o yaml to make edits you need. I still consider this pretty fringe and not really necessary for dev focused hiveutil.

docs/developing.md
dgoodwin: I'll add a link in the top level README

docs/cluster-adoption.md
dgoodwin: I'm thinking top level readme. Thought looking at what's there, maybe using hive is a better fit.

docs/developing.md
dgoodwin: Ended up moving the whole thing into using-hive.md given it's size.

dgoodwin:
Updated.

2uasimojo:
lgtm

Leaving for assignee to stamp.

Retitle/adjust description any time (can be done after merge, so no reason to hold this up). Is there a bug report we can link to?

dgoodwin:
No bug report no, this wasn't really a supported thing until this PR but it sounds like ACM has some use for it, thus why I started testing and found the bug.

PR #1445, opened 2021-07-06
Remove next provision details from ProvisionFailed condition.

dgoodwin:
x-ref: https://issues.redhat.com/browse/HIVE-1583

Per the Jira these details are not a good fit for exposing to users, and
can be removed.


2uasimojo:
Easy one. Does what it says. Passes tests, so we didn't have a test that vets the message (which is probably fine). lgtm, but leaving for assignee to stamp.

PR #1446, opened 2021-07-07
Safely delete ClusterDeployments

2uasimojo:
We're deleting ClusterDeployments from two different controllers: clusterpool and clusterdeployment. This means there's the possibility of timing problems whereby e.g.
- controller 1 loads a CD and decides it needs to be deleted
- controller 2 loads the same CD and updates it such that it should no longer be deleted
- controller 1, having already decided to delete, deletes

This commit adds a `SafeDelete` utility that asks the server to check the `ResourceVersion` of the object being deleted, and fail with a `Conflict` error if it's out of date. It then changes all deletions of ClusterDeployments to use that new method.


2uasimojo:
/hold

for testing

dgoodwin:
Looks good but could you elaborate on the details of the two controllers that might do this? Would would controller 1 (I'm assuming clusterdeployment_controller) delete a CD that controller 2 (presumably clusterclaim_controller) be grabbing?

2uasimojo:
> could you elaborate

Yeah, this is more or less hypothetical at this point. We don't need to merge it necessarily, but I wanted to get it working.

It's not even just about two controllers -- theoretically an outside agent (like ACM :) ) could race with one controller.

IMO Delete should check ResourceVersion by default like Update does; and the consumer should have to "force" explicitly in the cases where we're sure we don't care about the state of the object we're deleting. With SafeDelete we could implement this policy throughout hive if we wanted to.

2uasimojo:
Actually I think what led me to this was the race between { clusterclaim_controller [Updating](https://github.com/openshift/hive/blob/f00064e864a5d1f331cb80df36f467856704066a/pkg/controller/clusterclaim/clusterclaim_controller.go#L417) the cd to assign a claim } and { clusterpool_controller [Deleting](https://github.com/openshift/hive/blob/f00064e864a5d1f331cb80df36f467856704066a/pkg/controller/clusterpool/clusterpool_controller.go#L656) excess CDs }. I originally thought that was the cause of the behavior @jnpacker was seeing.

dgoodwin:
Ahh I see. Sounds good.

2uasimojo:
Rebased this to pick up the controller-runtime update, which allowed me to fix the unit test.

PR #1447, opened 2021-07-07
Claim: Add ClusterRunning=>ClusterDeleting condition

2uasimojo:
This is a UX improvement to reflect that a CD is deleting in the status of its assigned Claim: We set the ClusterRunning condition to False with Reason: ClusterDeleting.

Note that, although this triggers on both deletion timestamp and the `remove-claimed-cluster-from-pool` annotation, we are unlikely to notice in the latter case because the claim itself is deleted almost immediately in that code path. However, the code as written accounts for a future where that deletion may wait on some other finalizer.

### DIFF
--- a/pkg/controller/clusterclaim/clusterclaim_controller.go
+++ b/pkg/controller/clusterclaim/clusterclaim_controller.go
@@ -254,6 +254,8 @@ func (r *ReconcileClusterClaim) Reconcile(ctx context.Context, request reconcile
 	switch err := r.Get(context.Background(), client.ObjectKey{Namespace: clusterName, Name: clusterName}, cd); {
 	case apierrors.IsNotFound(err):
 		return r.reconcileForDeletedCluster(claim, logger)
+	case cd.DeletionTimestamp != nil || controllerutils.IsClaimedClusterMarkedForRemoval(cd):
+		return r.setDeletingStatus(claim, logger)
 	case err != nil:
 		logger.Log(controllerutils.LogLevel(err), "error getting ClusterDeployment")
 		return reconcile.Result{}, err
@@ -267,6 +269,23 @@ func (r *ReconcileClusterClaim) Reconcile(ctx context.Context, request reconcile
 	default:
 		return r.reconcileForAssignmentConflict(claim, logger)
 	}
+}
+
+func (r *ReconcileClusterClaim) setDeletingStatus(claim *hivev1.ClusterClaim, logger log.FieldLogger) (reconcile.Result, error) {
+	if conds, changed := controllerutils.SetClusterClaimConditionWithChangeCheck(
+		claim.Status.Conditions,
+		hivev1.ClusterRunningCondition,
+		corev1.ConditionFalse,
+		"ClusterDeleting",
+		"Assigned cluster has been marked for deletion",
+		controllerutils.UpdateConditionIfReasonOrMessageChange); changed {
+		claim.Status.Conditions = conds
+		if err := r.Status().Update(context.Background(), claim); err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "could not update status of ClusterClaim")
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
 }
 
 // getClaimLifetime returns the lifetime for a claim taking into account the lifetime set on the pool
--- a/pkg/controller/clusterclaim/clusterclaim_controller_test.go
+++ b/pkg/controller/clusterclaim/clusterclaim_controller_test.go
@@ -190,6 +190,20 @@ func TestReconcileClusterClaim(t *testing.T) {
 			}},
 		},
 		{
+			name:  "deleting cluster",
+			claim: initializedClaimBuilder.Build(testclaim.WithCluster(clusterName)),
+			cd: cdBuilder.GenericOptions(testgeneric.Deleted()).Build(
+				testcd.WithClusterPoolReference(claimNamespace, "test-pool", claimName),
+			),
+			expectCompletedClaim: true,
+			expectedConditions: []hivev1.ClusterClaimCondition{{
+				Type:    hivev1.ClusterRunningCondition,
+				Status:  corev1.ConditionFalse,
+				Reason:  "ClusterDeleting",
+				Message: "Assigned cluster has been marked for deletion",
+			}},
+		},
+		{
 			name:  "deleted cluster",
 			claim: initializedClaimBuilder.Build(testclaim.WithCluster(clusterName)),
 			expectedConditions: []hivev1.ClusterClaimCondition{{


2uasimojo:
/assign @suhanime 

/hold

for testing

pkg/controller/clusterclaim/clusterclaim_controller_test.go
```diff
@@ -208,6 +221,35 @@ func TestReconcileClusterClaim(t *testing.T) {
 			}},
 			expectAssignedClusterDeploymentDeleted: true,
 		},
+		{
+			name:  "deleted-pending cluster",
```
suhanime, line 225: `cluster marked for deletion` makes more sense

pkg/controller/clusterclaim/clusterclaim_controller_test.go
```diff
@@ -208,6 +221,35 @@ func TestReconcileClusterClaim(t *testing.T) {
...
+			expectedConditions: []hivev1.ClusterClaimCondition{{
+				Type:    hivev1.ClusterClaimClusterDeletedCondition,
+				Status:  corev1.ConditionTrue,
+				Reason:  "ClusterDeleted",
+				Message: "Assigned cluster has been deleted",
+			}},
+			expectCompletedClaim: true,
+		},
+		{
+			name:  "deleted-annotated cluster",
```
suhanime, line 239: cluster marked to be removed from pool?

pkg/controller/clusterclaim/clusterclaim_controller_test.go
```diff
@@ -536,6 +620,12 @@ func TestReconcileClusterClaim(t *testing.T) {
 					Reason:  "Resuming",
 					Message: "Waiting for cluster to be running",
 				},
+				{
```
suhanime, line 623: `expectedConditions: []hivev1.ClusterClaimCondition{
				{
					Type:    hivev1.ClusterClaimPendingCondition,
					Status:  corev1.ConditionFalse,
					Reason:  "ClusterClaimed",
					Message: "Cluster claimed",
				},
				{
					Type:    hivev1.ClusterRunningCondition,
					Status:  corev1.ConditionFalse,
					Reason:  "Resuming",
					Message: "Waiting for cluster to be running",
				},
				{
					Type:    hivev1.ClusterClaimClusterDeletedCondition,
					Status:  corev1.ConditionFalse,
					Reason:  "ClusterNotDeleted",
					Message: "Assigned cluster has not been deleted",
				},
			},`
This set of expected condition status are repeated many times, is it possible to break it out in a separate function for happy path?

2uasimojo: ✓

pkg/controller/clusterclaim/clusterclaim_controller_test.go
2uasimojo: ✓

pkg/controller/clusterclaim/clusterclaim_controller_test.go
2uasimojo: ✓

2uasimojo:
NB: Latest force-push includes the same go.sum updates as https://github.com/openshift/hive/pull/1450. These should wash out in a rebase and/or on merge.

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -253,11 +253,28 @@ func (r *ReconcileClusterClaim) Reconcile(ctx context.Context, request reconcile
 
 	cd := &hivev1.ClusterDeployment{}
 	switch err := r.Get(context.Background(), client.ObjectKey{Namespace: clusterName, Name: clusterName}, cd); {
-	case apierrors.IsNotFound(err):
-		return r.reconcileForDeletedCluster(claim, logger)
-	case err != nil:
+	case err != nil && !apierrors.IsNotFound(err):
 		logger.Log(controllerutils.LogLevel(err), "error getting ClusterDeployment")
 		return reconcile.Result{}, err
+	case apierrors.IsNotFound(err) || cd.DeletionTimestamp != nil || controllerutils.IsClaimedClusterMarkedForRemoval(cd):
```
abhinavdahiya, line 259: small nit:
this switch is no longer readable imo. 
just add the new conditions to separate if after the switch.

abhinavdahiya:
> Previously the ClusterClaim controller would only reconcile for a deleted ClusterDeployment if the CD was actually gone. Now it also notices if the CD has a deletionTimestamp or the remove-claimed-cluster-from-pool annotation.

what is the benefit of this change?

2uasimojo:
> what is the benefit of this change?

It *should* prevent you from claiming a CD that's already pending deletion.

abhinavdahiya:
> > what is the benefit of this change?
> 
> It _should_ prevent you from claiming a CD that's already pending deletion.

the clusterpool controller shouldn't attach a cluster to claim that is not ready already?
https://github.com/openshift/hive/blob/616301492b6f2ff2e9eeadc156d4b780ce0f04de/pkg/controller/clusterpool/clusterpool_controller.go#L329-L331

ready => https://github.com/openshift/hive/blob/616301492b6f2ff2e9eeadc156d4b780ce0f04de/pkg/controller/clusterpool/clusterpool_controller.go#L261-L285

pkg/controller/clusterclaim/clusterclaim_controller.go
2uasimojo: I tried refactoring this a couple of ways, ended up pulling the complicated bit into a helper func. Hope this suits.

My mental notes:
We want to blow up if:
- The error is not nil AND
- The error is something other than `IsNotFound`
We want to `reconcileForDeletedCluster` if:
- The error `IsNotFound`; OR
- There is no error AND (the CD has a deletion timestamp OR the removal annotation)
Continue otherwise, i.e. if:
- There is no error AND
- The CD has no deletion timestamp AND
- The CD has no removal annotation

2uasimojo:
> shouldn't attach a cluster to claim that is not ready

Correct. This handles timing issues where some other thread/agent deletes the CD after clusterpool_controller adds the claim info but before we get here.

2uasimojo:
You're right @abhinavdahiya, it should be impossible, in which case we're no worse off than we were before.

2uasimojo:
/test e2e

test env fixed

2uasimojo:
I'd like to land https://github.com/openshift/hive/pull/1460 and then revisit this.

2uasimojo:
#1460 removed the ClusterDeleted condition, so that part of this PR is obsolete.

The other part expands the invocation of `reconcileForDeletedCluster`, previously only triggered if `IsNotFound(cd)`, to include when:
- The CD's `remove-claimed-cluster-from-pool` annotation is set. This gets set to `true` by the clusterclaim_controller when the CD's claim is being deleted.
- The CD's `deletionTimestamp` is set. This is done by the clusterpool_controller upon noticing the above label, as long as doing so won't exceed the pool's `MaxConcurrent` operations.

Since #1460, `reconcileForDeletedCluster` sets the claim's `ClusterRunning` condition to
```
Type: ClusterRunning
Status: False
Reason: ClusterDeleted
Message: Assigned cluster has been deleted
```

Arguably, we could have an interim phase like
```
Type: ClusterRunning
Status: False
Reason: ClusterDeleting
Message: Assigned cluster has been marked for deletion
```
which we set at the same time as updating the CD with `remove-claimed-cluster-from-pool`. This Reason would appear in the table output like:
```
[efried@efried ~]$ oc get clusterclaim claim-gllgp
NAME          POOL                     PENDING          CLUSTERNAMESPACE               CLUSTERRUNNING   AGE
claim-gllgp   fake-ocp-4-7-amd64-aws   ClusterClaimed   fake-ocp-4-7-amd64-aws-m2kkp   ClusterDeleting  4m23s
```
...which IMO would be better UX than it staying `Running` during the deletion phase.

@suhanime @dgoodwin thoughts?

dgoodwin:
I like it.

2uasimojo:
Done, with a rebase, and stacked on #1472 to make sure we're running the whole test :)

2uasimojo:
Tested live, works. Watched a claim while issuing `oc delete cd`:

```
[efried@efried hive]$ oc get clusterclaim -n fake-cluster-pool   claim-z9lzf  -w
NAME          POOL                     PENDING          CLUSTERNAMESPACE               CLUSTERRUNNING   AGE
claim-z9lzf   fake-ocp-4-7-amd64-aws   ClusterClaimed   fake-ocp-4-7-amd64-aws-lh2z7   Resuming         6d21h
claim-z9lzf   fake-ocp-4-7-amd64-aws   ClusterClaimed   fake-ocp-4-7-amd64-aws-lh2z7   ClusterDeleting   6d21h
claim-z9lzf   fake-ocp-4-7-amd64-aws   ClusterClaimed   fake-ocp-4-7-amd64-aws-lh2z7   ClusterDeleted    6d21h
```

/hold cancel

2uasimojo:
And rebased again now that #1472 merged, so this shows up as one commit like it should.

/assign @suhanime 
...in Devan's absence.

PR #1449, opened 2021-07-08
Fix missed ArgoCD config deepcopy.

dgoodwin:
Not sure how this was missed, seems like something verify should have caught.

I've got this from running make update and make vendor, after removing _output dir to let it fetch latest controller gen.

/assign @abutcher 
/cc @2uasimojo


2uasimojo:
Okay, this looks sane and right -- but I can't reproduce, which means something is wrong with the tooling. We don't need to hold up this PR, but I want to get to the bottom of it.

2uasimojo:
FWIW, controller-gen isn't involved in this flow, so it's not that. And the script we're using is coming from vendor/ and installs deepcopy-gen etc. from *itself*, so there shouldn't be a possibility of a version mismatch there.

What's most worrying that CI isn't failing *before or after* your change.

2uasimojo:
@dgoodwin now that #1450 merged, you should be able to rebase this and it'll go back to just the deepcopy deltas.

dgoodwin:
Pushed.

2uasimojo:
I suspect (but don't feel like trying to prove) that this would have been caught via #1676.

PR #1451, opened 2021-07-09
Update controller-runtime to (unreleased) master

2uasimojo:
We were using a [controller-runtime fork](github.com/openshift-hive/controller-runtime) so we could [disable a metric](https://github.com/openshift-hive/controller-runtime/commit/e847e4b49034d7156179568639645920ec1df892) that was causing a cardinality explosion and memory problems.

Upstream controller-runtime has now [disabled that metric by default](https://github.com/kubernetes-sigs/controller-runtime/pull/1587); so this commit switches back to using upstream.

NOTE: We're using an unreleased upstream version.


dgoodwin:
Looks like some legit failures due to new code,  but this will be great once passing.

2uasimojo:
@dgoodwin you okay with using unreleased?

dgoodwin:
I can go either way, do we know how long until release expected?

2uasimojo:
According to Alvaro, it will ["take some time"](https://github.com/kubernetes-sigs/controller-runtime/pull/1587#issuecomment-877398019).

dgoodwin:
Probably best to keep what we have now and is stable until it's released, and then make the move to 0.9.x which we also have a card for.

2uasimojo:
Okay. (It'll be 0.10.x when it happens, since this is a breaking change.)

2uasimojo:
Well.
I just looked at what's changed since 0.9.2, other than my two commits that we were planning to cherry-pick anyway.
It's just https://github.com/kubernetes-sigs/controller-runtime/pull/1573. Can we agree the risk of pulling this in is negligible and move forward with this change?

2uasimojo:
I'm working on updating the tests. The major change in controller-runtime that our reconcile tests now have to accommodate is that the fake client now does Delete more closely to reality, in terms of handling finalizers properly. Many (*many*) tests were written against the old (incorrect) behavior: deleting the object regardless of finalizers.

dgoodwin:
> Well.
> I just looked at what's changed since 0.9.2, other than my two commits that we were planning to cherry-pick anyway.
> It's just [kubernetes-sigs/controller-runtime#1573](https://github.com/kubernetes-sigs/controller-runtime/pull/1573). Can we agree the risk of pulling this in is negligible and move forward with this change?

Yeah that's fine.

PR #1453, opened 2021-07-12
hibernation: support hibernating spot instance for AWS clusters

abhinavdahiya:
OpenShift supports running spot instances for compute. Therefore when
cluster is hibernated these instances must also be stopped and started
like the on-demand instances.

But due to the nature of the spot instances, these instances are not
guaranteed. So if these instances are started, these might not start if
the spot instance constraints are not met. Therefore aws supports
stopping and starting spot instances only when the spot instance request
are persistent (compared to one-time). see [1] for more info.

machine-api add support for spot instances in AWS detailed in [2].
Some of the things to note are,
- machine-api has decided that one spot instance in aws is tied to one
  Machine object in api.
- which means, the spot instances are created as "one-time" type. Once
  the backing instance is gone (either terminated by aws or user) the
  Machine object goes into Failed state and stays there as-is.
- For new Machine to take the place of the previous Machine, the
  machine-api deploys MachineHealthCheck object that deletes Machines
  that were "deleted by AWS" only. If user deletes/terminated the
  instance, the user must also remove the object from API.

To shut down a "one-time" spot instance, the only option is to terminate
the instance. So for running -> hibernation, aws actuator terminated
these instances instead of stopping them like it does for on-demand
instances.

Since the spot instances were terminated, for hibernation -> running
state change the aws actuator must somehow replace the previous
instances with new ones.
The only way is to delete the Machine objects for the previously
terminated spot instance, and letting the MachineSet replace those to
match the expected replica count.
The aws actuator now looks for these Machines that need to be replaced
by
- finding machines with "machine.openshift.io/interruptible-instance"
  label
- filtering out machines that are "not failed" given that their status
  was updated after hibernation was started
The actuator then deletes those machines by
- updating the machine objects to skip draining. since the instances are
  already known to be gone, skipping draining prevents machine-api
  getting stuck for pods to scheduled.
- deleting those machine objects.

[1]: https://aws.amazon.com/about-aws/whats-new/2020/01/amazon-ec2-spot-instances-stopped-started-similar-to-on-demand-instances/
[2]: https://github.com/openshift/enhancements/blob/master/enhancements/machine-api/spot-instances.md

xref: https://issues.redhat.com/browse/HIVE-1195

/cc @dgoodwin @2uasimojo 
/assign @joelddiaz

### DIFF
--- a/pkg/controller/hibernation/aws_actuator.go
+++ b/pkg/controller/hibernation/aws_actuator.go
@@ -1,19 +1,22 @@
 package hibernation
 
 import (
+	"context"
 	"fmt"
 	"os"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 
+	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
-
 	awsclient "github.com/openshift/hive/pkg/awsclient"
 	"github.com/openshift/hive/pkg/constants"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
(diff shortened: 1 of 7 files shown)


pkg/controller/hibernation/aws_actuator.go
```diff
@@ -149,7 +275,32 @@ func getAWSClient(cd *hivev1.ClusterDeployment, c client.Client, logger log.Fiel
...
+		n++
+	}
+	instances = instances[:n]
+	return instances, spotInstances
+}
+
+func instanceIDs(instances []*ec2.Instance) []string {
+	result := make([]string, len(instances))
+	for idx, i := range instances {
+		result[idx] = aws.StringValue(i.InstanceId)
```
2uasimojo, line 298: Why is `aws.StringValue()` needed here? Are we worried about the backing list changing under us? Keeping the IDs as pointers (making this func return `[]*string`) avoids the need to use e.g. `aws.StringSlice()` elsewhere. (And it saves a teeny bit of memory.)

[Later] Okay, I see that sometimes we're addressing and sometimes dereferencing. So meh, either way. But still a slight preference for keeping the pointers as much as possible.

pkg/controller/hibernation/aws_actuator.go
```diff
@@ -45,87 +48,210 @@ func (a *awsActuator) CanHandle(cd *hivev1.ClusterDeployment) bool {
...
-	return err
+	return nil
+}
+
+func (a *awsActuator) stopSpotInstances(awsClient awsclient.Client, instanceIDs []string, logger log.FieldLogger) error {
+	if len(instanceIDs) == 0 {
+		return nil
+	}
+	logger.WithField("instanceIDs", instanceIDs).Info("Terminating spot cluster instances")
+	_, err := awsClient.TerminateInstances(&ec2.TerminateInstancesInput{
```
2uasimojo, line 98: It would be nice to include a comment (here or at the top of the func) with some of the text from the commit message describing why this is `TerminateInstances` rather than `StopInstances`.

pkg/controller/hibernation/aws_actuator.go
```diff
@@ -45,87 +48,210 @@ func (a *awsActuator) CanHandle(cd *hivev1.ClusterDeployment) bool {
...
 	}
-	instanceIDs, err := getClusterInstanceIDs(cd, awsClient, stoppedOrStoppingStates, logger)
+
+	if err := a.startOnDemandInstances(cd, awsClient, logger); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (a *awsActuator) startOnDemandInstances(cd *hivev1.ClusterDeployment, awsClient awsclient.Client, logger log.FieldLogger) error {
```
2uasimojo, line 122: This doesn't look like it's actually doing anything differently than before. Should it be?
```suggestion
```

pkg/controller/hibernation/aws_actuator.go
```diff
@@ -45,87 +48,210 @@ func (a *awsActuator) CanHandle(cd *hivev1.ClusterDeployment) bool {
...
+
+		toBeReplaced = append(toBeReplaced, m)
+	}
+
+	logger.WithField("machines", machineNames(toBeReplaced)).Debug("Preemptible Machine objects will be replaced")
+	var replaced bool
+	var errs []error
+	for _, m := range toBeReplaced {
+		// We want the machine-api to skip the draining
+		// since we already know these nodes where terminated
```
2uasimojo, line 218:
```suggestion
		// since we already know these nodes were terminated
```

pkg/controller/hibernation/hibernation_actuator.go
```diff
@@ -28,3 +28,12 @@ type HibernationActuator interface {
 	// that have not stopped.
 	MachinesStopped(cd *hivev1.ClusterDeployment, hiveClient client.Client, logger log.FieldLogger) (bool, []string, error)
 }
+
+// HibernationActuator is the interface that the hibernation controller
+// for preemptible instances on cloud providers.
+type HibernationPreemptibleMachines interface {
+	// ReplaceMachines uses the remote client to replace the preemptible machines
+	// belonging to the given ClusterDeployment. Since it uses the remote client
+	// it should only be called when the API is reachable.
```
2uasimojo, line 37: What is the significance of the `replaced` return value? From reading the code, it returns `true` if *any* machine was replaced. Is that the intent (vs. if *all* relevant machines were replaced)?

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -413,6 +417,21 @@ func (r *hibernationReconciler) checkClusterResumed(cd *hivev1.ClusterDeployment
...
+		replaced, err := preemptibleActuator.ReplaceMachines(cd, remoteClient, logger)
+		if err != nil {
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "Failed to replace Preemptible machines")
+			return reconcile.Result{}, err
+		}
+		if replaced {
+			// when machines were replaced we must give some time before new nodes
+			// appear.
+			return reconcile.Result{RequeueAfter: nodeCheckWaitTime}, nil
+		}
```
2uasimojo, line 432: Making sure I'm understanding the logic here:

As written, the err will cause an immediate requeue, even if *some* machines were replaced. So the subsequent reconcile will find the as-yet-unreplaced machines and try to replace them. Eventually we'll get to a point where the last errant machine has been replaced and there is no error. So it's at that point that we'll wait `nodeCheckWaitTime` to requeue.

Or, shorter: we'll wait `nodeCheckWaitTime` to requeue when we've successfully replaced the *last* machine.

pkg/controller/hibernation/aws_actuator_test.go
```diff
@@ -118,6 +125,62 @@ func TestStopAndStartMachines(t *testing.T) {
...
+		spot:     2,
+		setupClient: func(t *testing.T, c *mockawsclient.MockClient) {
+			c.EXPECT().StopInstances(gomock.Any()).Do(
+				func(input *ec2.StopInstancesInput) {
+					input.InstanceIds = aws.StringSlice([]string{"onDemand-0"})
+				}).Return(nil, nil)
+			c.EXPECT().TerminateInstances(gomock.Any()).Do(
+				func(input *ec2.TerminateInstancesInput) {
+					input.InstanceIds = aws.StringSlice([]string{"spot-0", "spot-1"})
+				}).Return(nil, nil)
```
2uasimojo, line 166: At least for this test case, it would be nice to assert the number of calls to each method.
```suggestion
			c.EXPECT().StopInstances(gomock.Any()).Times(1).Do(
				func(input *ec2.StopInstancesInput) {
					input.InstanceIds = aws.StringSlice([]string{"onDemand-0"})
				}).Return(nil, nil)
			c.EXPECT().TerminateInstances(gomock.Any()).Times(2).Do(
				func(input *ec2.TerminateInstancesInput) {
					input.InstanceIds = aws.StringSlice([]string{"spot-0", "spot-1"})
				}).Return(nil, nil)
```

2uasimojo (commented):
Mostly code-ish (rather than functionality-ish) comments inline, nothing blocking. I'm having to take it on faith that this is The Right Thing™ for the AWS and machine APIs.

pkg/controller/hibernation/aws_actuator.go
```diff
@@ -149,7 +275,32 @@ func getAWSClient(cd *hivev1.ClusterDeployment, c client.Client, logger log.Fiel
...
+	for _, i := range instances {
+		if aws.StringValue(i.InstanceLifecycle) == "spot" {
+			spotInstances = append(spotInstances, i)
+			continue
+		}
+		instances[n] = i
+		n++
+	}
+	instances = instances[:n]
+	return instances, spotInstances
```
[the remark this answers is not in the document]
abhinavdahiya: this is a common technique of filtering a slice. https://github.com/golang/go/wiki/SliceTricks#filter-in-place
IMO, the name of the function makes it clear that it will remove certain objects from the slice.

pkg/controller/hibernation/aws_actuator.go
abhinavdahiya: the list of instance ID strings is useful for reporting/logging the work in action. pointers of string is not useful anywhere except when calling AWS API, and the SDK recognizes that by provding aws.StringSlice function.

my preference will be to keep a list of strings vs pointers to them..

pkg/controller/hibernation/aws_actuator_test.go
abhinavdahiya: there are only 2 calls at max , one call for all on-demand, and another for all spot instances.

pkg/controller/hibernation/hibernation_actuator.go
abhinavdahiya: at least one was replaced.

pkg/controller/hibernation/hibernation_controller.go
abhinavdahiya: > Or, shorter: we'll wait nodeCheckWaitTime to requeue when we've successfully replaced the last machine.

wait after replacing **any** machine. This is allow for new machine is boot in replacement and join as a node.

2uasimojo: As written, it won't wait if there's any error.

pkg/controller/hibernation/hibernation_actuator.go
2uasimojo: Would you mind adding that to the docstring?

pkg/controller/hibernation/aws_actuator_test.go
2uasimojo: orite. I wonder why the above change didn't fail for me?

pkg/controller/hibernation/hibernation_controller.go
abhinavdahiya: which is correct. failing on error moves the work to rate-limited queue. so transient errors are retired and given up (very large next interval) on continued failures.

pkg/controller/hibernation/aws_actuator.go
joelddiaz: or perhaps just call the function `terminateSpotInstnaces`

pkg/controller/hibernation/aws_actuator_test.go
joelddiaz: Maybe it didn't fail b/c there is no `defer ctrl.Finish()` after the `gomock.NewController(t)` line???

pkg/controller/hibernation/aws_actuator_test.go
```diff
@@ -118,6 +125,62 @@ func TestStopAndStartMachines(t *testing.T) {
...
+		},
+	}}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
```
joelddiaz, line 171: Are we missing a `defer ctrl.Finish()` here?

joelddiaz (commented):
I read through the changes to the aws actuator, and I think I'm left with one question that wasn't clear to me after reading through it.
If, prior to cluster hibernation, the cluster has some spot instances in a failed state (b/c someone was manually messing with the spot instances), does a running -> hibernating -> running set of state transitions leave us in a place where the previously failed spot instances are now noticed and restored by the aws actuator codepath?

pkg/controller/hibernation/aws_actuator_test.go
2uasimojo: Goooood call.

pkg/controller/hibernation/aws_actuator_test.go
abhinavdahiya: Ahh, fixed @joelddiaz

abhinavdahiya:
> If, prior to cluster hibernation, the cluster has some spot instances in a failed state (b/c someone was manually messing with the spot instances), does a running -> hibernating -> running set of state transitions leave us in a place where the previously failed spot instances are now noticed and restored by the aws actuator codepath?

@joelddiaz I think you are correct. the hive aws actuator will terminate the previously failed spot instances when we transition from running->hibernation and cluster aws actuator will try re creating them when we delete the old objects and they get replaced with new ones during the hibernating -> running transition.

did you have any specific concerns with this behavior in mind? it seems okay to me..

joelddiaz:
> @joelddiaz I think you are correct. the hive aws actuator will terminate the previously failed spot instances when we transition from running->hibernation and cluster aws actuator will try re creating them when we delete the old objects and they get replaced with new ones during the hibernating -> running transition.
> 
> did you have any specific concerns with this behavior in mind? it seems okay to me..

No concerns. Just making sure I'm reading understanding the behavior correctly. :+1:

abhinavdahiya:
I think i answered all the comments. 👍🏽

2uasimojo:
Good for me. Leaving to assignee to stamp.

PR #1454, opened 2021-07-12
Scrub frequent Azure cloud dns errors

suhanime:
When dns controller hits Invalid credentials on Azure cloud, Azure API
constantly tries to login and refresh the token. This results in
frequent update of dns condition for generic cloud errors.
While we do have rate limiter on cloud errors for dns, we can still
scrub the Azure error and bubble up only the error description. This
will make the error easier to interpret and restrict frequent status
updates.

xref: https://issues.redhat.com/browse/HIVE-1582

### DIFF
--- a/pkg/controller/utils/errorscrub.go
+++ b/pkg/controller/utils/errorscrub.go
@@ -5,8 +5,11 @@ import (
 )
 
 var (
+	newlineTabRE = regexp.MustCompile(`\n\t`)
+	// aws
 	awsRequestIDRE = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
-	newlineTabRE   = regexp.MustCompile(`(\n\t)`)
+	// azure
+	azureErrorDescriptionRE = regexp.MustCompile(`\"error_description\":\"(.*?)\\r\\n`)
 )
 
 // ErrorScrub scrubs cloud error messages destined for CRD status to remove things that
@@ -17,5 +20,10 @@ func ErrorScrub(err error) string {
 	}
 	s := newlineTabRE.ReplaceAllString(err.Error(), ", ")
 	s = awsRequestIDRE.ReplaceAllString(s, "")
+	// if Azure error, return just the error description
+	match := azureErrorDescriptionRE.FindStringSubmatch(s)
+	if len(match) > 0 {
+		return match[1]
+	}
 	return s
 }
--- a/pkg/controller/utils/errorscrub_test.go
+++ b/pkg/controller/utils/errorscrub_test.go
@@ -34,6 +34,11 @@ func TestErrorScrubber(t *testing.T) {
 			input:    nil,
 			expected: "",
 		},
+		{
+			name:     "azure error",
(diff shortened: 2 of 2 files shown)


pkg/controller/utils/errorscrub.go
```diff
@@ -2,11 +2,15 @@ package utils
 
 import (
 	"regexp"
+	"strings"
 )
 
 var (
+	newlineTabRE = regexp.MustCompile(`(\n\t)`)
```
2uasimojo, line 9: This isn't necessary. You're already chopping off at `\\r\\n` which will be earlier than `\n\t` (if that in fact exists in the string at all).

(Aside: the parens would be redundant anyway, since you're not using the capture group with `ReplaceAllString`.)

pkg/controller/utils/errorscrub.go
```diff
@@ -2,11 +2,15 @@ package utils
...
+	// aws
 	awsRequestIDRE = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
-	newlineTabRE   = regexp.MustCompile(`(\n\t)`)
+	// azure
+	azureErrorDescriptionRE = regexp.MustCompile(`error_description\":\"([^\r\n]+)`)
```
2uasimojo, line 13: The `[^\r\n]+` part of this isn't doing what you expect, because a) those sequences are still escaped in the string (`\\r\\n`), and b) the square brackets represent alternation ("any single character that is neither a line feed nor a newline"). So, mostly by luck, `match[1]` is grabbing everything to the end of the string, and it's the `Split` on L27 that's cutting down to the part you care about.

I would suggest rewriting this as follows:
```suggestion
	azureErrorDescriptionRE := regexp.MustCompile(`\"error_description\":\"(.*?)\\r\\n`)
```
- First, I would match the leading `\"` in `\"error_description\"`. That's a nit.
- I'm matching `\\r\\n` (which is not the same as `\r\n`!) outside of the capture group...
- ...but using `?` to make the `.*` in the capture group non-greedy, so it only goes up to the first `\\r\\n`. (Normally we would do this with lookarounds, but go doesn't support them.)

Then below you don't need to do the `Split` -- see suggestion inline.

pkg/controller/utils/errorscrub.go
```diff
@@ -17,5 +21,11 @@ func ErrorScrub(err error) string {
 	}
 	s := newlineTabRE.ReplaceAllString(err.Error(), ", ")
 	s = awsRequestIDRE.ReplaceAllString(s, "")
+	// if Azure error, return just the error description
+	match := azureErrorDescriptionRE.FindStringSubmatch(s)
+	if len(match) > 0 {
+		matchValue := strings.Split(match[1], "\\r\\n")
+		return matchValue[0]
```
2uasimojo, line 28:
```suggestion
		return match[1]
```

2uasimojo (changes requested):

2uasimojo:
btw, what you have will work; it's just got a bunch of redundant pieces. Regexes can be tough to read at best -- no need to make it harder :)

pkg/controller/utils/errorscrub.go
suhanime: This makes way more sense - updated. Thank you!!

pkg/controller/utils/errorscrub.go
suhanime: This one's for scrubbing of AWS Request ID I think, but didn't end up needing any change

2uasimojo: Oh, just realized this was latent; and also that it is indeed necessary for some of the other message types. In a fup we could still:
```suggestion
	newlineTabRE = regexp.MustCompile(`\n\t`)
```

2uasimojo (commented):
/lgtm

2uasimojo:
/lgtm

Thanks!

PR #1457, opened 2021-07-13
Enforce format of Duration fields

2uasimojo:
Add kubebuilder directives to validate Duration fields.

[HIVE-1587](https://issues.redhat.com/browse/HIVE-1587)


2uasimojo:
/assign @suhanime 
/cc @dgoodwin 

Note: this is currently on top of #1455, which will merge as soon as e2e decides to cooperate. Or we could abandon that one and squash these two commits. Whatever.

PR #1458, opened 2021-07-14
Make metadata admin password secret optional.

dgoodwin:
Hive doesn't need this value, and we should not require users adopting
clusters to provide it.

/assign @2uasimojo 
/cc @lwan-wanglin


2uasimojo (commented):
Sane.

Is the field required for non-adoption scenarios? If so, we should:
- document that in the description comment
- add a check for that

dgoodwin:
I don't think it should be required anywhere. In our code that sets it we'd like to make sure it's populated, which we do, but in the api itself I don't think we should try to make assumptions as it doesn't really affect anything.

2uasimojo:
Yeah, that makes sense.

/lgtm

PR #1461, opened 2021-07-16
Fix build-machinery-go dependencies

2uasimojo:
This commit fixes two things about our build-machinery-go includes:
- The `images.mk` include uses `ensure-imagebuilder`, which comes from `imagebuilder.mk`. We shouldn't have to include the latter explicitly. https://github.com/openshift/build-machinery-go/pull/50 is fixing that, but until that merges and we can vendor it in, include it explicitly.
- `ensure-kustomize` was working as far as it went, but it was installing kustomize into a subdirectory which is (usually) not in
`$PATH`, so things like `make deploy` were either breaking (bad) or picking up kustomize from somewhere else (potentially worse). Fix our usage of kustomize to ref the variable that `ensure-kustomize` sets up to point to the binary it installs. This is also a different version of kustomize (4.1.3) from what our e2e test was installing (2.0.0) and it behaves slightly differently when constructing the ServiceAccount for the deployment: the namespace hardcoded into the rolebinding manifest is not overridden as expected. Removing the namespace from that manifest fixes the problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,16 @@
 all: vendor update test build
 
 # Include the library makefile
+# TODO: Get rid of the imagebuilder.mk include once
+# https://github.com/openshift/build-machinery-go/pull/50 is merged and
+# vendored.
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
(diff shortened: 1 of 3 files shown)


2uasimojo (commented):
/test e2e

PR #1463, opened 2021-07-16
Don't allow excess claims to create extra CDs

2uasimojo:
Previously, due to a math error, if the number of pending claims exceeded the pool size, we would create the surplus number of CDs, assign them, and then delete them. This commit fixes the math error: any excess claim will remain Pending until capacity becomes available to create a CD for it.

**NOTE:**
The above also applies to the edge case where the pool size is zero. A claim against such a pool will remain Pending *until the pool is edited to have a nonzero size*.

This differs from other scenarios when we're at capacity:
- If we've reached `MaxSize`, capacity becomes available when a *CD* is deleted (which may be prompted by its *claim* being deleted).
- If we've reached `MaxConcurrent`, capacity becomes available when a *CD* finishes provisioning or deprovisioning.

Due to this discrepancy, the argument could be made to treat a zero-size pool specially and simply fail any claim against it. We can debate that separately; this commit doesn't address it.

[HIVE-1593](https://issues.redhat.com/browse/HIVE-1593)

### DIFF
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -325,6 +325,10 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 
 	// reserveSize is the number of clusters that the pool currently has in reserve
 	reserveSize := len(installingCDs) + len(readyCDs) - len(pendingClaims)
+	// Don't allow excess claims to add CDs
+	if reserveSize < 0 {
+		reserveSize = 0
+	}
 
 	readyCDs, err = r.assignClustersToClaims(pendingClaims, readyCDs, logger)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


abhinavdahiya:
> https://github.com/openshift/hive/blob/667abfc4ca5094c2efcf62d1cefc720544284a76/apis/hive/v1/clusterpool_types.go#L19-L22
> https://github.com/openshift/hive/blame/master/docs/clusterpools.md#L5-L8

those things make it sound like the pool is supposed to keep some warm clusters for immediate claim, but technically allow as many claims are requested.. (unless limited my other configuration)

so "Don't allow excess claims to create extra CDs" kind of seems to go against the purpose?

abhinavdahiya:
> we would create the surplus number of CDs, assign them, and then delete them

the delete part seem like the incorrect behavior based on the current expectations.. 🤔

2uasimojo:
To me, the documentation as written makes no indication one way or the other as to what the behavior of excess claims should be. One could make a case for either interpretation of "pool size". It's just an architectural decision we need to make. @dgoodwin care to make the call here?

2uasimojo:
We decided in scrum today (20210721) that we *do* want to allow claims to generate CDs in excess of the pool size. The capacity is controlled by MaxSize. Closing this in favor of whatever comes out of HIVE-1599.

PR #1464, opened 2021-07-19
Index ClusterDeployments by Pool in clusterpool_controller

2uasimojo:
The clusterpool_controller needs a list of all the ClusterDeployments in the pool being reconciled. Previously it was filtering these on the client side. This commit adds an index for the ClusterDeployment.Spec.ClusterPoolRef so we can filter on the server side.


pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -78,6 +79,16 @@ func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) *Re
 	}
 }
 
+func indexCDByClusterPool(o client.Object) []string {
```
abhinavdahiya, line 82: we don't seem to be testing/reusing this func.. so my personal preference would be to keep this inline.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -78,6 +79,16 @@ func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) *Re
...
+	cd := o.(*hivev1.ClusterDeployment)
+	if cd.Spec.ClusterPoolRef != nil {
+		return []string{
+			cd.Spec.ClusterPoolRef.Namespace + "/" + cd.Spec.ClusterPoolRef.PoolName,
```
abhinavdahiya, line 86: both the indexer and filter needs to perform this task. so
- we use something that is standard and know like types.NamespacedName{}.String()
- or have a function that creates these keys

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -78,6 +79,16 @@ func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) *Re
 	}
 }
 
+func indexCDByClusterPool(o client.Object) []string {
+	cd := o.(*hivev1.ClusterDeployment)
```
abhinavdahiya, line 83: in case there is some bug and we don't get clusterdeployment, this will panic and stop entire controller.. maybe handle and return empty?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -43,6 +43,7 @@ const (
 	clusterPoolAdminRoleName        = "hive-cluster-pool-admin"
 	clusterPoolAdminRoleBindingName = "hive-cluster-pool-admin-binding"
 	icSecretDependent               = "install config template secret"
+	cdClusterPoolIndex              = "spec.clusterpool.namespacedname"
```
abhinavdahiya, line 46: if possible we should try to mimic the actual path of the value or as close as we can be. so my personal preference would be `spec.clusterPoolRef`

2uasimojo: I've noticed other indices don't mirror actual paths exactly.

https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/controller/machinemanagement/machinemanagement_controller.go#L88
https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/controller/clusterdeployment/clusterinstalls.go#L223
https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/operator/hive/hive_controller.go#L160

I don't know if there's a technical reason -- like avoiding actual conflicts somewhere in c-r/apiserver -- but
- It actually helped me when figuring out how to do this, because it made it clear that we were *not* referencing/indexing exact fields; and
- In this case we're not indexing on the whole ClusterPoolRef. In particular, we're not using the ClaimName.

I don't mind changing the string if you have another suggestion, but I would like it *not* to be `spec.clusterPoolRef`.

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: Done.

Though the CD controller's [indexClusterInstall](https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/controller/clusterdeployment/clusterinstalls.go#L225) does the same.

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: IMO that's a bit too paranoid, considering the parameter two before this one in the `IndexField` func is the type in question. Also, none of the other indexer funcs are doing that type check.

https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/controller/machinemanagement/machinemanagement_controller.go#L90
https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/controller/clusterdeployment/clusterinstalls.go#L227
https://github.com/openshift/hive/blob/33be31582aa140ce772333d0437eb77cd723cee2/pkg/operator/hive/hive_controller.go#L162

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: The indexer is building the key from the CD's ClusterPoolRef, whereas the filter is building it from the pool's namespacedname, so the gain isn't huge, but done.

PR #1244, opened 2021-01-06
Add install log regex for GCPQuotaSSDTotalGBExceeded.

dgoodwin:
/assign @akhil-rane

PR #1466, opened 2021-07-20
add a match for GCP CPU quota

yithian:
i was investigating a [cluster provisioning delay](https://redhat.pagerduty.com/incidents/PP1XFXV) and this error was the cause

per [the SOP](https://github.com/openshift/hive-sops/blob/master/sop/ClusterProvisioningFailure.md#check-clusterdeployment-conditions), i opened this to add the error to the configmap

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -93,6 +93,11 @@ data:
       - "Quota \'SSD_TOTAL_GB\' exceeded"
       installFailingReason: GCPQuotaSSDTotalGBExceeded
       installFailingMessage: GCP quota SSD_TOTAL_GB exceeded
+    - name: GCPComputeQuota
+      searchRegexStrings:
+      - "compute\\.googleapis\\.com/cpus is not available in [a-z0-9-]* because the required number of resources \\([0-9]*\\) is more than remaining quota"
+      installFailingReason: GCPComputeQuotaExceeded
+      installFailingMessage: GCP CPUs quota exceeded
     # Bare Metal
     - name: LibvirtSSHKeyPermissionDenied
       searchRegexStrings:
--- a/pkg/controller/clusterprovision/installlogmonitor_test.go
+++ b/pkg/controller/clusterprovision/installlogmonitor_test.go
@@ -28,6 +28,7 @@ const (
 	pendingVerificationLog  = "blahblah\naws_instance.master.2: Error launching source instance: PendingVerification: Your request for accessing resources in this region is being validated, and you will not be able to launch additional resources in this region until the validation is complete. We will notify you by email once your request has been validated. While normally resolved within minutes, please allow up to 4 hours for this process to complete. If the issue still persists, please let us know by writing to awsa\n\nblahblah"
 	gcpInvalidProjectIDLog  = "blahblah\ntime=\"2020-11-13T16:05:07Z\" level=fatal msg=\"failed to fetch Master Machines: failed to load asset \"Install Config\": platform.gcp.project: Invalid value: \"o-6b20f250\": invalid project ID\nblahblah"
 	gcpSSDQUotaLog          = "blahblah\ntime=\"2021-01-06T03:35:44Z\" level=error msg=\"Error: Error waiting for instance to create: Quota 'SSD_TOTAL_GB' exceeded. Limit: 500.0 in region asia-northeast2.\nblahblah"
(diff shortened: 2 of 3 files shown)


2uasimojo:
`make update` ought to fix the verify failure.

Can you please add some context for this change in the commit/PR message?

yithian:
i was investigating a [cluster provisioning delay](https://redhat.pagerduty.com/incidents/PP1XFXV) and this error was the cause

per [the SOP](https://github.com/openshift/hive-sops/blob/master/sop/ClusterProvisioningFailure.md#check-clusterdeployment-conditions), i opened this to add the error to the configmap

2uasimojo:
I noticed some of the previous commits in this vein added unit tests (e.g. https://github.com/openshift/hive/pull/1244, https://github.com/openshift/hive/pull/1323). Would you mind doing that?

Also please squash when done. Thanks!

/cc @dgoodwin 

(Sorry for the multiple review iterations here. I haven't looked at one of these before.)

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -93,6 +93,11 @@ data:
       - "Quota \'SSD_TOTAL_GB\' exceeded"
       installFailingReason: GCPQuotaSSDTotalGBExceeded
       installFailingMessage: GCP quota SSD_TOTAL_GB exceeded
+    - name: GcpComputeQuota
+      searchRegexStrings:
+      - "compute.googleapis.com/cpus is not available in [A-Z_]* because the required number of resources ([0-9]*) is more than the remaining quota"
+      installFailingReason: GcpComputeQuotaExceeded
```
dgoodwin, line 99: Nit: please case "GCP" instead of "Gcp".

dgoodwin:
Unit test would be good if you're at all worried about your regex, which for me is 100% of my regexes.

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -93,6 +93,11 @@ data:
       - "Quota \'SSD_TOTAL_GB\' exceeded"
       installFailingReason: GCPQuotaSSDTotalGBExceeded
       installFailingMessage: GCP quota SSD_TOTAL_GB exceeded
+    - name: GCPComputeQuota
+      searchRegexStrings:
+      - "compute.googleapis.com/cpus is not available in [a-z0-9-]* because the required number of resources"
```
2uasimojo, line 98: I was just reviewing the previous iteration when you uploaded this. The notes still apply, especially if you want to keep the full regex you had before:

Interpreting a regex string through yaml is icky. This needs to be:
```suggestion
      - "compute\\.googleapis\\.com/cpus is not available in [a-z0-9-]* because the required number of resources \\([0-9]*\\) is more than remaining quota"
```
- Double backslash gets interpreted first by yaml into single backslash, which is what you want as an escape in the actual regex to remove special meaning from the subsequent character.
- Escape the periods in `compute.googleapis.com`; otherwise they'll match any single character. (Including period, so it'd still work, but your intent is to match exactly period.)
- Slash isn't special, so it doesn't need to be escaped. (You're probably thinking of needing to escape slash in a `sed` expression that's using slash as the regex delimiter. But e.g. `s/foo\/bar/foo\/baz/` is the same as `s,foo/bar,foo/baz,`. In this case you can think of double-quote as the delimiter (sort of).)
- Escaping to match literal parens is indeed necessary; but the backslashes need to be doubled as noted above.

pkg/controller/clusterprovision/installlogmonitor_test.go
```diff
@@ -262,6 +263,26 @@ func TestParseInstallLog(t *testing.T) {
...
+			name: "GCP compute quota",
+			log:  pointer.StringPtr(gcpCPUQuotaLog),
+			existing: []runtime.Object{&corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      regexConfigMapName,
+					Namespace: constants.DefaultHiveNamespace,
+				},
+				Data: map[string]string{
+					"regexes": `
+- name: GCPComputeQuota
```
2uasimojo, line 276: I can confirm that you need to copy the yaml stanza in here to make the test work... which sucks. Good future improvement to the test suite to load up the actual yaml file instead, so there's no possibility of drift.

config/configmaps/install-log-regexes-configmap.yaml
yithian: oh, dang, thanks! i was trying to figure out why this regex wasn't matching even after escaping the parens (once)

2uasimojo:
Phew, I think we've beaten this to death. Thanks for sticking with it @yithian.

/lgtm
/approve

yithian:
> Phew, I think we've beaten this to death. Thanks for sticking with it @yithian.

thanks for the assist!

PR #1467, opened 2021-07-21
Don't add apiextensions to scheme

2uasimojo:
apiextensions were not being used by the manager or operator.


2uasimojo:
Never mind, I see where we're using it. Follow #1468 for the real fix.

PR #1471, opened 2021-07-23
remotemachineset: aws: fallback to using kubernetes.io/role/elb tag for public subnets when route tables are not enough

abhinavdahiya:
Currently the discovery codepath only decides whether a subnet is public
by using the routetables.

There are many network deployments which do not directly attach an igw
to the public subnet, but usage of these subnets is clearly defined by
the user by tagging these subnets with PublicSubnetTagName. So it
would be benefitial if the discovery codepath respect the indication on
the public subnets set by the admins of the cluster.

Similarly the aws-load-balancer-controller [1] now solely depends on
PublicSubnetTagName to discover the public subnets for ELB.

In this change, we update the isSubnetPublic function such that when it
fails to decide that a subnet is public using route table rules, if
checks if the subnet has the PublicSubnetTagName, and if the tag exists
and equal to empty("") and "1" value [2] it returns true. So the main
check is
still route tables based, and the tag based becomes a fallback.

This is based on a change that is also being floated in upstream [3].

Since the upstream change hasn't been merged. This only helps creating
machines in these subnets. If a user was to create a Service type
Loadbalancer, it would still require that some subnets exists which are
identifable as public with just route tables.

[1]:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/networking/subnet_resolver.go#L114
[2]:
https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html
[3]: https://github.com/kubernetes/kubernetes/pull/103861

xref: https://issues.redhat.com/browse/HIVE-1568

/cc @suhanime 
/assign @dgoodwin

### DIFF
--- a/pkg/controller/remotemachineset/awsactuator.go
+++ b/pkg/controller/remotemachineset/awsactuator.go
@@ -366,7 +366,7 @@ func (a *AWSActuator) getPrivateSubnetsByAvailabilityZone(pool *hivev1.MachinePo
 
 	var privateSubnets, publicSubnets = map[string]ec2.Subnet{}, map[string]ec2.Subnet{}
 	for _, subnet := range results.Subnets {
-		isPublic, err := isSubnetPublic(routeTables.RouteTables, *subnet.SubnetId, a.logger)
+		isPublic, err := isSubnetPublic(routeTables.RouteTables, subnet, a.logger)
 		if err != nil {
 			return nil, errors.Wrap(err, "error describing route tables")
 		}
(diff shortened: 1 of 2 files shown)


dgoodwin:
looks good to me, cc @jharrington22

abhinavdahiya:
/hold

doing some testing on my side to verify some things. will lift as soon as that's done.

thanks for the lgtm @suhanime

PR #1475, opened 2021-07-26
fix app-sre build script current hash grab

gregsheremeta:
hack/app_sre_create_image_catalog.sh was grabbing the
hive-production v3 shard as the currently deployed hash,
and that's wrong because it's pinned to an old commit.
Use hivep01ue1's hash.

/assign @2uasimojo 
/cc @akhil-rane


2uasimojo (approved):
/lgtm

2uasimojo:
/test unit

We should really fix that one.

gregsheremeta:
fixed it up, please re-review @2uasimojo

2uasimojo (approved):
/lgtm

PR #1477, opened 2021-07-29
Hibernation bug with fake clusters

suhanime:
As a part of previous change to support hibernation for fake clusters,
it was found that
- some spec and/or status updates were missing
- code coverage for hibernation due to powerstate change (and not
  hibernateafter) was missing
- switching to running condition from initialized hibernation condition
  was not covered

This commit rearranges some code to resolve the above stated bugs

xref: https://issues.redhat.com/browse/HIVE-1585

/assign @2uasimojo 
/cc @abhinavdahiya

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -208,6 +208,12 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 				return r.setHibernatingCondition(cd, hivev1.SyncSetsNotAppliedReason, "Cluster SyncSets have not been applied", corev1.ConditionFalse, cdLog)
 			}
 		}
+
+		// Hibernate fake cluster if hibernate after is not set
+		if controllerutils.IsFakeCluster(cd) && cd.Spec.HibernateAfter == nil && hibernatingCondition.Status != corev1.ConditionTrue {
+			r.setHibernatingCondition(cd, hivev1.HibernatingHibernationReason, "Fake cluster is stopped",
+				corev1.ConditionTrue, cdLog)
+		}
 	}
 
 	// Clear any lingering unsupported hibernation condition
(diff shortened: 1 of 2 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -208,6 +208,12 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 				return r.setHibernatingCondition(cd, hivev1.SyncSetsNotAppliedReason, "Cluster SyncSets have not been applied", corev1.ConditionFalse, cdLog)
 			}
 		}
+
+		// Hibernate fake cluster if hibernate after is not set
+		if controllerutils.IsFakeCluster(cd) && cd.Spec.HibernateAfter == nil && hibernatingCondition.Status != corev1.ConditionTrue {
+			r.setHibernatingCondition(cd, hivev1.HibernatingHibernationReason, "Fake cluster is stopped",
```
2uasimojo, line 214: This chunk is [not being covered by UT](https://app.codecov.io/gh/openshift/hive/compare/1477/tree/pkg/controller/hibernation/hibernation_controller.go#D1L214) -- see below.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -392,10 +392,11 @@ func TestReconcile(t *testing.T) {
 		},
```
2uasimojo, line 392: The above test is hibernating based on `hibernateAfter`, so it's not hitting the new chunk of code referenced previously. We should add a new test with a CD with:
- *no* `hibernateAfter`
- `powerState=Hibernating`
- the hibernating condition `False` or `Unknown` (with whatever `Reason`)
and assert that the final `powerState` and condition represent hibernating.

2uasimojo (commented):
I'm confident the code change is good, but would like to see an additional test case, noted inline.

pkg/controller/hibernation/hibernation_controller_test.go
suhanime: Added - actually modified an existing one since there are other tests for hibernateafter. Also, when hibernateafter is not set, we change the powerstate, so an assert on that would be not necessary

2uasimojo: > when hibernateafter is not set, we change the powerstate, so an assert on that would be not necessary

I would think that would be a really good reason to include the assertion :)

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -378,7 +378,7 @@ func TestReconcile(t *testing.T) {
 		{
 			name: "hibernate fake cluster",
 			cd: cdBuilder.Build(
-				testcd.WithHibernateAfter(1*time.Hour),
+				o.shouldHibernate,
 				testcd.InstalledTimestamp(time.Now().Add(-1*time.Hour)),
```
2uasimojo, line 382: nit: This is now unnecessary (but harmless)

2uasimojo (approved):
Thanks @suhanime!

/lgtm

PR #1482, opened 2021-08-02
Introduce e2e-pool test

2uasimojo:
[HIVE-1605](https://issues.redhat.com/browse/HIVE-1605)

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -1,0 +1,182 @@
+max_tries=60
+sleep_between_tries=10
+# Set timeout for the cluster deployment to install
+# timeout = sleep_between_cluster_deployment_status_checks * max_cluster_deployment_status_checks
+max_cluster_deployment_status_checks=90
+sleep_between_cluster_deployment_status_checks="1m"
+
+export CLUSTER_NAMESPACE="${CLUSTER_NAMESPACE:-cluster-test}"
+
+# In CI, HIVE_IMAGE and RELEASE_IMAGE are set via the job's `dependencies`.
+if [[ -z "$HIVE_IMAGE" ]]; then
+    echo "The HIVE_IMAGE environment variable was not found." >&2
+    echo "It must be set to the fully-qualified pull spec of a hive container image." >&2
+    echo "E.g. quay.io/my-user/hive:latest" >&2
+    exit 1
+fi
+if [[ -z "$RELEASE_IMAGE" ]]; then
+    echo "The RELEASE_IMAGE environment variable was not found." >&2
+    echo "It must be set to the fully-qualified pull spec of an OCP release container image." >&2
+    echo "E.g. quay.io/openshift-release-dev/ocp-release:4.7.0-x86_64" >&2
+    exit 1
+fi
+
+echo "Running ${TEST_NAME} with HIVE_IMAGE ${HIVE_IMAGE}"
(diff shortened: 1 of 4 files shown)


2uasimojo:
/hold

Not quite done with this yet.

hack/e2e-common.sh
```diff
@@ -0,0 +1,181 @@
+max_tries=60
```
2uasimojo, line 1: Note to reviewers: There's quite a bit of cleanup/refactoring that could be done here, but I explicitly didn't do that so it'd be easier to review: use your favorite visual diff tool to compare this file to the original e2e-test.sh.

2uasimojo:
/hold cancel

Working. Ready for review.

hack/e2e-common.sh
abhinavdahiya: do you mind moving this move to separate commit. that would be a lot more read able.

hack/e2e-common.sh
```diff
@@ -0,0 +1,181 @@
...
+  echo -n "Creating namespace ${CLUSTER_NAMESPACE}. Try #${i}/${max_tries}... "
+  if oc create namespace "${CLUSTER_NAMESPACE}"; then
+    echo "Success"
+    break
+  else
+    echo -n "Failed, "
+  fi
+
+  i=$((i + 1))
+done
```
joelddiaz, line 44: should we not need to verify that the namespace creation was successfull when we exit this loop (so we can have a clearer error message in the even that this never succeeded)?

2uasimojo: Absolutely. There's a couple of things here, including the fact that the command to restore the original namespace doesn't work at all for me. I'd like to defer that to a separate PR if possible.

hack/e2e-common.sh
2uasimojo: I'm not sure how much more readable it would make anything. That commit would simply comprise e2e-common.sh and e2e-test.sh in the exact form they exist in this PR.

dgoodwin: Possible misunderstanding here, refactoring that *could* be done, but isn't in this PR.

hack/e2e-common.sh
```diff
@@ -0,0 +1,181 @@
...
+	CLUSTER_DOMAIN="${BASE_DOMAIN}"
+fi
+
+
+echo "Using cluster base domain: ${CLUSTER_DOMAIN}"
+
+function capture_manifests() {
+    oc get clusterdeployment -A -o yaml &> "${ARTIFACT_DIR}/hive_clusterdeployment.yaml" || true
+    oc get clusterimageset -o yaml &> "${ARTIFACT_DIR}/hive_clusterimagesets.yaml" || true
+    oc get clusterprovision -A -o yaml &> "${ARTIFACT_DIR}/hive_clusterprovision.yaml" || true
```
dgoodwin, line 149: could you add clusterstate -A -o yaml here as well, may give us some insight into operators in the cluster.

dgoodwin (commented):
One nit but this lgtm.

hack/e2e-common.sh
2uasimojo: Done.

hack/e2e-common.sh
2uasimojo: Correct: I kept the existing code from e2e-test.sh, warts and all, and just moved it to e2e-common.sh with as little change as possible, precisely to make this PR more scrutable. We can refactor/reorganize later.

PR #1483, opened 2021-08-03
Fix transfer of PreserveOnDelete to DNSZone for installed clusters.

dgoodwin:
Previous attempt only tested against installing clusters, which is
bypassed if the cluster is already installed.

Move the sync of PreserveOnDelete prior to both portions of the
reconcile loop.

x-ref: https://issues.redhat.com/browse/HIVE-1594

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -557,28 +557,26 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, nil
 	}
 
-	// Ensure that if DNSZone already exists, it's PreserveOnDelete setting matches the ClusterDeployment:
 	if cd.Spec.ManageDNS {
-		dnsZone := &hivev1.DNSZone{}
-		nsName := types.NamespacedName{Namespace: cd.Namespace, Name: controllerutils.DNSZoneName(cd.Name)}
-		logger := cdLog.WithField("zone", nsName.String())
-
-		switch err := r.Get(context.TODO(), nsName, dnsZone); {
-		case err != nil && !apierrors.IsNotFound(err):
-			logger.WithError(err).Error("failed to fetch DNS zone")
+		dnsZone, err := r.ensureManagedDNSZone(cd, cdLog)
+		if err != nil {
 			return reconcile.Result{}, err
-		case err == nil:
-			if dnsZone.Spec.PreserveOnDelete != cd.Spec.PreserveOnDelete {
-				logger.Infof("setting DNSZone PreserveOnDelete to match ClusterDeployment PreserveOnDelete: %v", cd.Spec.PreserveOnDelete)
-				dnsZone.Spec.PreserveOnDelete = cd.Spec.PreserveOnDelete
-				err := r.Update(context.TODO(), dnsZone)
-				if err != nil {
-					logger.WithError(err).Log(controllerutils.LogLevel(err), "error updating DNSZone")
-				}
-				return reconcile.Result{}, err
+		}
+		if dnsZone == nil {
+			// dnsNotReady condition was set.
+			if isSet, _ := isDNSNotReadyConditionSet(cd); isSet {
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -557,6 +557,30 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, nil
 	}
 
+	// Ensure that if DNSZone already exists, it's PreserveOnDelete setting matches the ClusterDeployment:
```
2uasimojo, line 560:
```suggestion
	// Ensure that if DNSZone already exists, its PreserveOnDelete setting matches the ClusterDeployment:
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -557,6 +557,30 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+	if cd.Spec.ManageDNS {
+		dnsZone := &hivev1.DNSZone{}
+		nsName := types.NamespacedName{Namespace: cd.Namespace, Name: controllerutils.DNSZoneName(cd.Name)}
+		logger := cdLog.WithField("zone", nsName.String())
+
+		switch err := r.Get(context.TODO(), nsName, dnsZone); {
+		case err != nil && !apierrors.IsNotFound(err):
```
2uasimojo, line 567: IIUC, the first pass through for a *fresh* cluster would now hit here with `IsNotFound` and fall out of this whole chunk.

Then further down we `ensureManagedDNSZone` which creates the DNSZone.

That triggers another Reconcile here, because we watch DNSZones and enqueue for the owning CD.

And it's on that _second_ reconcile we would hit this new code path to sync up the `PreserveOnDelete` settings.

2uasimojo (commented):
This seems right. It seems to take a long way around in exchange for avoiding calling the same code twice, but I reckon that's the k8s way :)

/lgtm

pkg/controller/clusterdeployment/clusterdeployment_controller.go
suhanime, line 560: Is it possible to use `ensureManagedDNSZone` here, with maybe a bool stating if managed dns zone not found then don't create it?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -557,6 +557,30 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+			logger.WithError(err).Error("failed to fetch DNS zone")
+			return reconcile.Result{}, err
+		case err == nil:
+			if dnsZone.Spec.PreserveOnDelete != cd.Spec.PreserveOnDelete {
```
suhanime, line 571: And this check can be useful in `ensureManagedDNSZone` too

suhanime:
/hold 
for review comments

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: I didn't think it would trigger, if it is an IsNotFound error, we don't have a case for that in the switch, which I think would be a no-op and continue with the code normally to create the DNSZone, which will get the flag on creation. Does that seem right?

2uasimojo: > which will get the flag on creation

Ahh, I see that now. Cool.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: I don't think it's necessary to run the same code twice, but I am wondering if we should just move ensureManagedDNSZone call up here, it would help repair a missing dnszone for an existing cluster, which wouldn't happen today. Checking if this is feasible.

dgoodwin:
How does this look with the full ensure loop moved up prior to the installed cluster block? It will now repair a deleted DNSZone immediately.

PR #1484, opened 2021-08-03
ClusterPool:ClusterDeployment version & condition

2uasimojo:
This commit adds a mechanism for calculating a hash representing the following fields in ClusterPool.Spec:
- Platform
- BaseDomain
- ImageSetRef
- InstallConfigSecretTemplateRef

When creating a new ClusterDeployment for the pool, this hash is stored in a new annotation on the ClusterDeployment.

The clusterpool controller checks all unassigned (installing or ready) ClusterDeployments and sets a new condition on the ClusterPool. This condition is:
- "True" when all CDs are up to date with the pool (including the edge case where there are no CDs);
- "False" when one or more CDs are stale;
- "Unknown" if the annotation is missing from one or more CDs.
- (Also "Unknown" if we haven't had a chance to figure it out yet.)

Of note:
- When "False" or "Unknown", the condition's Message contains a list of the offending CDs (in the spirit of ClusterSync's "Failed" condition).
- "False" takes precedence over "Unknown" (including: we don't try to combine the CD lists in the Message).

[HIVE-1058](https://issues.redhat.com/browse/HIVE-1058)

### DIFF
--- a/apis/hive/v1/clusterpool_types.go
+++ b/apis/hive/v1/clusterpool_types.go
@@ -138,6 +138,9 @@ const (
 	// ClusterPoolCapacityAvailableCondition is set to provide information on whether the cluster pool has capacity
 	// available to create more clusters for the pool.
 	ClusterPoolCapacityAvailableCondition ClusterPoolConditionType = "CapacityAvailable"
+	// ClusterPoolAllClustersCurrentCondition indicates whether all unassigned (installing or ready)
+	// ClusterDeployments in the pool match the current configuration of the ClusterPool.
+	ClusterPoolAllClustersCurrentCondition ClusterPoolConditionType = "AllClustersCurrent"
 )
 
 // +genclient
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/aws/aws-sdk-go v1.38.41
 	github.com/blang/semver/v4 v4.0.0
 	github.com/davecgh/go-spew v1.1.1
+	github.com/davegardnerisme/deephash v0.0.0-20210406090112-6d072427d830
 	github.com/evanphx/json-patch v4.11.0+incompatible
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -10,6 +10,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
+	"github.com/davegardnerisme/deephash"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -30,7 +31,6 @@ import (
 	"github.com/openshift/hive/pkg/clusterresource"
 	"github.com/openshift/hive/pkg/constants"
 	hivemetrics "github.com/openshift/hive/pkg/controller/metrics"
-	"github.com/openshift/hive/pkg/controller/utils"
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 )
 
@@ -55,6 +55,7 @@ var (
 	clusterPoolConditions = []hivev1.ClusterPoolConditionType{
 		hivev1.ClusterPoolMissingDependenciesCondition,
 		hivev1.ClusterPoolCapacityAvailableCondition,
+		hivev1.ClusterPoolAllClustersCurrentCondition,
 	}
 )
 
@@ -347,12 +348,14 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 	for _, cd := range toRemoveClaimedCDs[:toDel] {
 		cdLog := logger.WithField("cluster", cd.Name)
 		cdLog.Info("deleting cluster deployment for previous claim")
-		if err := utils.SafeDelete(r, context.Background(), cd); err != nil {
+		if err := cds.Delete(r.Client, cd.Name); err != nil {
 			cdLog.WithError(err).Error("error deleting cluster deployment")
 			return reconcile.Result{}, err
 		}
 	}
 	availableCurrent -= toDel
+
+	poolVersion := calculatePoolVersion(clp)
 
 	switch drift := reserveSize - int(clp.Spec.Size); {
 	// activity quota exceeded, so no action
@@ -364,7 +367,7 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 	// If too many, delete some.
 	case drift > 0:
 		toDel := minIntVarible(drift, availableCurrent)
-		if err := r.deleteExcessClusters(cds.Installing(), cds.Assignable(), toDel, logger); err != nil {
+		if err := r.deleteExcessClusters(cds, toDel, logger); err != nil {
 			return reconcile.Result{}, err
 		}
 	// If too few, create new InstallConfig and ClusterDeployment.
@@ -373,10 +376,21 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 			break
 		}
 		toAdd := minIntVarible(-drift, availableCapacity, availableCurrent)
-		if err := r.addClusters(clp, toAdd, logger); err != nil {
+		// TODO: Do this in cdLookup so we can add the new CDs to Installing(). For now it's okay
+		// because the new CDs are guaranteed to have a matching PoolVersion, and that's the only
+		// thing we need to keep consistent.
+		if err := r.addClusters(clp, poolVersion, toAdd, logger); err != nil {
 			log.WithError(err).Error("error adding clusters")
 			return reconcile.Result{}, err
 		}
+	}
+
+	// One more (possible) status update: wait until the end to detect whether all unassigned CDs
+	// are current with the pool config, since assigning or deleting CDs may eliminate the last
+	// one that isn't.
+	if err := setCDsCurrentCondition(r.Client, cds, clp, poolVersion); err != nil {
+		log.WithError(err).Error("error updating 'ClusterDeployments current' status")
+		return reconcile.Result{}, err
 	}
 
 	if err := r.reconcileRBAC(clp, logger); err != nil {
@@ -385,6 +399,23 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 	}
 
 	return reconcile.Result{}, nil
+}
+
+// calculatePoolVersion computes a hash of the important (to ClusterDeployments) fields of the
+// ClusterPool.Spec. This is annotated on ClusterDeployments when the pool creates them, which
+// subsequently allows us to tell whether all unclaimed CDs are up to date and set a condition
+// accordingly.
+// NOTE: If we change this algorithm, we're guaranteed to think that all CDs are stale at the
+// moment that code update rolls out. We may wish to consider a way to support detecting the old
+// value as "current" in that case.
+func calculatePoolVersion(clp *hivev1.ClusterPool) string {
+	ba := []byte{}
+	ba = append(ba, deephash.Hash(clp.Spec.Platform)...)
+	ba = append(ba, deephash.Hash(clp.Spec.BaseDomain)...)
+	ba = append(ba, deephash.Hash(clp.Spec.ImageSetRef)...)
+	ba = append(ba, deephash.Hash(clp.Spec.InstallConfigSecretTemplateRef)...)
+	// Hash of hashes to ensure fixed length
+	return fmt.Sprintf("%x", deephash.Hash(ba))
 }
 
 func minIntVarible(v1 int, vn ...int) (m int) {
@@ -497,6 +528,7 @@ func (r *ReconcileClusterPool) applyRoleBinding(desired, observed *rbacv1.RoleBi
 
 func (r *ReconcileClusterPool) addClusters(
 	clp *hivev1.ClusterPool,
+	poolVersion string,
 	newClusterCount int,
 	logger log.FieldLogger,
 ) error {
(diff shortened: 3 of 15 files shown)


2uasimojo:
This is currently stacked on #1474.

/hold

until that merges. Meanwhile, review only the second commit.

2uasimojo:
/cc @gurnben @stevekuznetsov @abutcher @abhinavdahiya 
per [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1627586516278200?thread_ts=1627578625.257700&cid=CE3ETN3J8)

2uasimojo:
These test failures are bogus*. No idea what's going on here. I even tried deleting the builds in the CI env. Anyway, I imagine we'll kick something over when this gets rebased after #1474 merges.

*except for e2e-pool, duh -- #1482 hasn't merged yet :roll_eyes:

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/aws/aws-sdk-go v1.38.41
 	github.com/blang/semver/v4 v4.0.0
 	github.com/davecgh/go-spew v1.1.1
+	github.com/davegardnerisme/deephash v0.0.0-20210406090112-6d072427d830
```
abhinavdahiya, line 13: maybe the use what we do elsewhere in the code, https://github.com/openshift/hive/blob/57147b24cb33b05de91c8070b459b66c24556e06/pkg/operator/hive/controllersconfig.go#L86-L91

apis/hive/v1/clusterpool_types.go
```diff
@@ -103,6 +103,14 @@ type ClusterPoolStatus struct {
 	// Ready is the number of unclaimed clusters that have been installed and are ready to be claimed.
 	Ready int32 `json:"ready"`
 
+	// PoolVersion is an opaque value representing the state of the important (to
+	// ClusterDeployments) fields of the ClusterPool. When creating a new ClusterDeployment,
+	// this value is copied into its Spec.ClusterPoolReference.PoolVersion. Comparing these values
+	// gives a quick indication as to whether the ClusterDeployment's configuration matches that of
+	// the pool.
+	// +optional
+	PoolVersion string `json:"poolVersion,omitempty"`
```
abhinavdahiya, line 112: Version to me has a lot of baggage in terms of the format, an opaque digest doesn't fit well. Can we use some other name here?
🤔

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -249,6 +249,12 @@ type ClusterPoolReference struct {
 	Namespace string `json:"namespace"`
 	// PoolName is the name of the ClusterPool for which the cluster was created.
 	PoolName string `json:"poolName"`
+	// PoolVersion is an opaque value representing the state of the important (to
+	// ClusterDeployments) fields of the ClusterPool at the time this CD was created. Comparing
+	// this value to the ClusterPool.Status.PoolVersion is a quick indication as to whether the
+	// ClusterDeployment's configuration matches that of the pool.
+	// +optional
+	PoolVersion string `json:"poolVersion,omitempty"`
```
abhinavdahiya, line 257: personally it feels odd to have an internal detail like that from the clusterpool to live here. i would have preferred an annotation instead of including that in object's spec api.

pkg/controller/clusterpool/collections.go
```diff
@@ -344,20 +344,71 @@ func (cds *cdCollection) SyncClaimAssignments(c client.Client, claims *claimColl
...
+		clp.Status.Conditions,
+		hivev1.ClusterPoolClusterDeploymentsCurrentCondition,
+		status, reason, message, controllerutils.UpdateConditionIfReasonOrMessageChange)
+	if changed {
+		clp.Status.Conditions = conds
+		if err := c.Status().Update(context.Background(), clp); err != nil {
+			return err
+		}
+	}
+	return nil
```
abhinavdahiya: this seems out of place from the cd collection here? can we just expect cds as input?

apis/hive/v1/clusterpool_types.go
2uasimojo: I was following the naming of `ResourceVersion`, which is a close corollary to this field: an opaque string that you're only supposed to compare for (in)equality to see if something has changed. Open to other suggestions.

apis/hive/v1/clusterdeployment_types.go
2uasimojo: I have a healthy dislike of annotations as a rule, but let me think about this some more.

go.mod
2uasimojo: I'll look into that. Is `%v` guaranteed to produce the same thing for (different instances of) identical objects? I know some languages [don't/didn't guarantee map ordering](https://docs.python.org/2.7/library/stdtypes.html#dict.items) for example.

pkg/controller/clusterpool/collections.go
2uasimojo: Yeah, that makes sense, will do.

go.mod
2uasimojo: So yeah, I can't prove that `%v` breaks with map key order, but I can prove it doesn't play nice with pointers, whereas `deephash` does The Right Thing™: https://play.golang.org/p/1SQUEeJfOND

apis/hive/v1/clusterdeployment_types.go
2uasimojo: Thought about it some more. Totally agree. Changed it out.

apis/hive/v1/clusterpool_types.go
2uasimojo: I named the annotation `cluster-pool-spec-hash`. There are still vestiges of `poolVersion` in comments, variable names, method names. Hope that's acceptable. (I really didn't want to be naming variables `clusterPoolSpecHash`...)

2uasimojo:
Updated.

This is still stacked on #1474, which is now stacked on #1489, so I'll keep the hold. Review just the last commit.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -386,6 +401,15 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 	return reconcile.Result{}, nil
 }
 
+func calculatePoolVersion(clp *hivev1.ClusterPool) string {
+	ba := []byte{}
```
abhinavdahiya, line 405: since we store the hashes in the annotation, what is the expected max length of the hash? the append here makes it seems like it grows with more fields we include?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -557,6 +583,13 @@ func (r *ReconcileClusterPool) createCluster(
 	}
 	logger.WithField("cluster", ns.Name).Info("Creating new cluster")
 
+	annotations := clp.Spec.Annotations
+	// Annotate the CD with the hash of the (important) pool spec fields so we can distinguish "stale" CDs
```
abhinavdahiya, line 587: i think we can abstract that to the pool version itself instead of re stating how we get the pool version..

pkg/controller/clusterpool/collections.go
```diff
@@ -371,20 +373,102 @@ func (cds *cdCollection) SyncClaimAssignments(c client.Client, claims *claimColl
...
+
+// setCDsCurrentCondition idempotently sets the ClusterDeploymentsCurrent condition on the
+// ClusterPool according to whether all unassigned CDs have the same PoolVersion as the pool.
+func setCDsCurrentCondition(c client.Client, cds *cdCollection, clp *hivev1.ClusterPool, poolVersion string) error {
+	// CDs with mismatched poolVersion
+	mismatchedCDs := make([]string, 0)
+	// CDs with no poolVersion
+	unknownCDs := make([]string, 0)
+
+	for _, cd := range append(cds.assignable, cds.installing...) {
```
abhinavdahiya, line 423: directly using private fields?

pkg/controller/clusterpool/collections.go
```diff
@@ -371,20 +373,102 @@ func (cds *cdCollection) SyncClaimAssignments(c client.Client, claims *claimColl
...
+		annotations := cd.Annotations
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
```
abhinavdahiya, line 427: you only need to initailize when you want to write to the map, the reads are fine with nil maps.

pkg/controller/clusterpool/collections.go
```diff
@@ -371,20 +373,102 @@ func (cds *cdCollection) SyncClaimAssignments(c client.Client, claims *claimColl
...
+		if cdPoolVersion, ok := annotations[constants.ClusterDeploymentPoolSpecHashAnnotation]; !ok {
+			// Annotation is either missing or empty. This could be due to upgrade (this CD was
```
abhinavdahiya, line 429: we are not checking for the value to be empty in the conditional? how can we be sure that annotation is empty?

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -46,6 +47,8 @@ func TestReconcileClusterPool(t *testing.T) {
 	corev1.AddToScheme(scheme)
 	rbacv1.AddToScheme(scheme)
 
+	initialPoolVersion := "21a89f16034caf029cebda7cced933a42d4d5b548134005dcbf29ce484222325"
```
abhinavdahiya, line 50: maybe we should include some comments on how to update it when we have to change the poolBuilder.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -162,7 +206,10 @@ func TestReconcileClusterPool(t *testing.T) {
 			existing: []runtime.Object{
 				initializedPoolBuilder.Build(testcp.WithSize(5), testcp.WithMaxSize(3)),
 				testclaim.FullBuilder(testNamespace, "test", scheme).Build(testclaim.WithPool(testLeasePoolName)),
-				cdBuilder("c1").Build(testcd.Installed(), testcd.WithClusterPoolReference(testNamespace, testLeasePoolName, "test")),
+				cdBuilder("c1").Build(
+					testcd.Installed(),
+					testcd.WithClusterPoolReference(testNamespace, testLeasePoolName, "test"),
+					testcd.WithPoolVersion(initialPoolVersion)),
```
abhinavdahiya, line 212: should we upate the cdbuilder to include the poolversion instead of setting the default everytime?

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -108,6 +119,39 @@ func TestReconcileClusterPool(t *testing.T) {
...
+				initializedPoolBuilder.Build(testcp.WithImageSet("abc123")),
+			},
+			expectPoolVersionChanged: true,
+		},
+		{
+			name: "poolVersion changes with InstallConfigSecretTemplate",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(testcp.WithInstallConfigSecretTemplateRef("abc123")),
+			},
+			expectPoolVersionChanged: true,
```
abhinavdahiya, line 154: this could be a smaller unit test just for the calculatePoolVersion function instead of putting it into this Reconcile test.

2uasimojo (commented):
Thanks for the review @abhinavdahiya!

2uasimojo:
> This is currently stacked on #1474.
> 
> /hold
> 
> until that merges.

#1474 is merged. Rebased.

/hold cancel

apis/hive/v1/clusterpool_types.go
```diff
@@ -138,6 +138,9 @@ const (
 	// ClusterPoolCapacityAvailableCondition is set to provide information on whether the cluster pool has capacity
 	// available to create more clusters for the pool.
 	ClusterPoolCapacityAvailableCondition ClusterPoolConditionType = "CapacityAvailable"
+	// ClusterPoolClusterDeploymentsCurrentCondition indicates whether all unassigned (installing or ready)
+	// ClusterDeployments in the pool match the current configuration of the ClusterPool.
+	ClusterPoolClusterDeploymentsCurrentCondition ClusterPoolConditionType = "ClusterDeploymentsCurrent"
```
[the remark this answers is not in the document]
abhinavdahiya: personally my preference is to not include ClusterDeployments in the condition name.. so the best i could come up was `AllClustersCurrent`

pkg/controller/clusterpool/clusterpool_controller.go
abhinavdahiya: > Fortunately for us, there's effectively no limit to the length of an annotation value :)

even though there is no limitation, i would expect that our hash is capped/constant at a length like most hashes.. our hashed value that keep increasing in length just looks odd and might create confusion in future imo.

pkg/controller/clusterpool/clusterpool_controller_test.go
abhinavdahiya: i think those are separate, the reconcile test makes sure that some hash is set. while the unit test for calculatePoolVersion tests that all the necessary fields are included in hash changes.

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: Changed to hash the hashes, which fixes the length at 16c.

(ed. Notwithstanding what I said above, anyone confused about the actual value of the string shouldn't be looking at the string. It's supposed to be an opaque value -- again like kube's `resourceVersion`.)

apis/hive/v1/clusterpool_types.go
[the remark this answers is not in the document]
2uasimojo: <IMO>Saying that the *clusters* are current implies a lot more than we're actually verifying. And the word "All" confuses the fact that we're only validating *unclaimed* CDs.</IMO>

But I've changed it per your suggestion.

pkg/controller/clusterpool/clusterpool_controller_test.go
2uasimojo: I just don't agree here. This controller does not have unit tests for individual helper functions. It does have cases that exercise them through the entire reconcile. The broader we can make a test for a given level of coverage, the better the test is. I'd like to leave this alone.

dgoodwin: Let's roll with it as is, I like separate unit tests for helpers of complexity but this the nature of this one doesn't scream separate to me and seems ok in a broader unit test.

2uasimojo:
Updated (forgot to `make vendor` :roll_eyes:).

I don't know wtf is wrong with e2e :angry:

PR #1486, opened 2021-08-04
Add templates, scripts, targets for OLM-less deploy

2uasimojo:
New `make` targets:
- `build-app-sre-template`: Makes sure hack/app-sre/saas-template.yaml is up to date with the current state of the repo. More below.
- `verify-app-sre-template`: Added to the `verify` chain. Makes sure the above has been run.

The `build-app-sre-template` target
- Syncs the resources in the new `kustomization.yaml` to make sure we're picking up all the CRDs under config/crds. The kustomization makes sure the template will contain variables, which will be substituted by app-sre via `oc process`, for the URI of the container image in the deployment.
- Uses that kustomization to generate/overwrite the new `saas-objects.yaml` to contain all the objects we wish to deploy via
app-sre.
- Runs the new `generate-saas-template.py` script, which melds the `saas-objects.yaml` with the new `saas-template-stub.yaml` to generate/overwrite the new `saas-template.yaml`.

This `saas-template.yaml` file is what app-sre uses for direct deploy.

The other new artifact is `hack/app-sre/build-push.sh`. This is the script that will eventually be run by app-sre's ci-int job for direct deploy. It is a subset of `hack/app_sre_build_deploy.sh`: it only builds the hive operator image and pushes it to quay. (Note that this script will not be used by appsre until all environments are migrated to direct deploy.)

[HIVE-1588](https://issues.redhat.com/browse/HIVE-1588)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,6 +203,26 @@ update-codegen:
 	hack/update-codegen.sh
 update: update-codegen
 
+# Build the template file used for direct (OLM-less) deploy by app-sre
+build-app-sre-template: ensure-kustomize
+	# Sync CRDs into kustomize resources
+	cd hack/app-sre && ../../$(KUSTOMIZE) edit add resource ../../config/crds/*.yaml
+	# Generate temporary saas object file
+	$(KUSTOMIZE) build --load-restrictor=LoadRestrictionsNone hack/app-sre --output hack/app-sre/saas-objects.yaml
+	# Generate saas template
+	./hack/app-sre/generate-saas-template.py hack/app-sre/saas-template-stub.yaml hack/app-sre/saas-objects.yaml hack/app-sre/saas-template.yaml
+	# Remove temporary saas object file
+	rm hack/app-sre/saas-objects.yaml
+
+
+# This needs to go after codegen so the CRDs are up to date
+verify-app-sre-template: build-app-sre-template
+	git diff --exit-code hack/app-sre/
+verify: verify-app-sre-template
+
+# This needs to go after codegen so the CRDs are up to date
+update: build-app-sre-template
+
 # Check import naming
 .PHONY: verify-imports
 verify-imports: build
--- a/hack/app-sre/kustomization.yaml
+++ b/hack/app-sre/kustomization.yaml
@@ -1,0 +1,39 @@
+# This kustomize configuration deploys what is required for the hive-operator
+# *without* OLM.
+# Except:
+# It does not contain a Namespace or a HiveConfig. These are created by app-sre.
+
+# The CRD resources are synced automatically by the `build-app-sre-template` make target.
+# The other resources should be curated manually.
+resources:
+- ../../config/operator/operator_role.yaml
+- ../../config/operator/operator_role_binding.yaml
+- ../../config/operator/operator_deployment.yaml
+- ../../config/crds/hiveinternal.openshift.io_clustersyncleases.yaml
+- ../../config/crds/hiveinternal.openshift.io_clustersyncs.yaml
+- ../../config/crds/hiveinternal.openshift.io_fakeclusterinstalls.yaml
+- ../../config/crds/hive.openshift.io_checkpoints.yaml
+- ../../config/crds/hive.openshift.io_clusterclaims.yaml
+- ../../config/crds/hive.openshift.io_clusterdeployments.yaml
+- ../../config/crds/hive.openshift.io_clusterdeprovisions.yaml
+- ../../config/crds/hive.openshift.io_clusterimagesets.yaml
+- ../../config/crds/hive.openshift.io_clusterpools.yaml
+- ../../config/crds/hive.openshift.io_clusterprovisions.yaml
+- ../../config/crds/hive.openshift.io_clusterrelocates.yaml
+- ../../config/crds/hive.openshift.io_clusterstates.yaml
+- ../../config/crds/hive.openshift.io_dnszones.yaml
+- ../../config/crds/hive.openshift.io_hiveconfigs.yaml
+- ../../config/crds/hive.openshift.io_machinepoolnameleases.yaml
+- ../../config/crds/hive.openshift.io_machinepools.yaml
+- ../../config/crds/hive.openshift.io_selectorsyncidentityproviders.yaml
+- ../../config/crds/hive.openshift.io_selectorsyncsets.yaml
+- ../../config/crds/hive.openshift.io_syncidentityproviders.yaml
+- ../../config/crds/hive.openshift.io_syncsets.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+# Use app-sre-supplied variables to pull the image for the current commit
+# TODO: Pull by digest instead. Don't forget to address Disappearing Digest Syndrome.
+images:
+- name: registry.ci.openshift.org/openshift/hive-v4.0:hive
+  newName: ${REGISTRY_IMG}
+  newTag: ${IMAGE_TAG}
--- a/hack/app-sre/saas-template.yaml
+++ b/hack/app-sre/saas-template.yaml
@@ -1,0 +1,6511 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: hive-saas-template
+objects:
+- apiVersion: apiextensions.k8s.io/v1
+  kind: CustomResourceDefinition
+  metadata:
+    annotations:
+      controller-gen.kubebuilder.io/version: v0.6.0
+    creationTimestamp: null
+    name: checkpoints.hive.openshift.io
+  spec:
+    group: hive.openshift.io
+    names:
+      kind: Checkpoint
+      listKind: CheckpointList
+      plural: checkpoints
+      singular: checkpoint
+    scope: Namespaced
+    versions:
+    - name: v1
+      schema:
+        openAPIV3Schema:
+          description: Checkpoint is the Schema for the backup of Hive objects.
+          properties:
+            apiVersion:
+              description: 'APIVersion defines the versioned schema of this representation
+                of an object. Servers should convert recognized schemas to the latest
+                internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+              type: string
+            kind:
+              description: 'Kind is a string value representing the REST resource
+                this object represents. Servers may infer this from the endpoint the
+                client submits requests to. Cannot be updated. In CamelCase. More
+                info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+              type: string
+            metadata:
+              type: object
+            spec:
+              description: CheckpointSpec defines the metadata around the Hive objects
+                state in the namespace at the time of the last backup.
+              properties:
+                lastBackupChecksum:
+                  description: LastBackupChecksum is the checksum of all Hive objects
+                    in the namespace at the time of the last backup.
+                  type: string
+                lastBackupRef:
+                  description: LastBackupRef is a reference to last backup object
+                    created
+                  properties:
+                    name:
+                      type: string
+                    namespace:
+                      type: string
+                  required:
+                  - name
+                  - namespace
+                  type: object
+                lastBackupTime:
+                  description: LastBackupTime is the last time we performed a backup
+                    of the namespace
+                  format: date-time
+                  type: string
+              required:
+              - lastBackupChecksum
+              - lastBackupRef
+              - lastBackupTime
+              type: object
+            status:
+              description: CheckpointStatus defines the observed state of Checkpoint
+              type: object
+          type: object
+      served: true
+      storage: true
+      subresources:
+        status: {}
+  status:
+    acceptedNames:
+      kind: ''
+      plural: ''
+    conditions: []
+    storedVersions: []
+- apiVersion: apiextensions.k8s.io/v1
+  kind: CustomResourceDefinition
+  metadata:
+    annotations:
+      controller-gen.kubebuilder.io/version: v0.6.0
+    creationTimestamp: null
+    name: clusterclaims.hive.openshift.io
+  spec:
+    group: hive.openshift.io
+    names:
+      kind: ClusterClaim
+      listKind: ClusterClaimList
+      plural: clusterclaims
+      singular: clusterclaim
+    scope: Namespaced
+    versions:
+    - additionalPrinterColumns:
+      - jsonPath: .spec.clusterPoolName
+        name: Pool
+        type: string
+      - jsonPath: .status.conditions[?(@.type=='Pending')].reason
+        name: Pending
+        type: string
+      - jsonPath: .spec.namespace
+        name: ClusterNamespace
+        type: string
+      - jsonPath: .status.conditions[?(@.type=='ClusterRunning')].reason
+        name: ClusterRunning
+        type: string
+      - jsonPath: .metadata.creationTimestamp
+        name: Age
+        type: date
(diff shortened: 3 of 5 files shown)


2uasimojo:
This goes first. Other than the new `verify` dep, it's a no-op until the app-sre side lands.

hack/app-sre/build-push.sh
```diff
@@ -0,0 +1,31 @@
...
+# app-interface saas file: data/services/hive/cicd/ci-int/saas-hive.yaml
+QUAY_IMAGE="quay.io/app-sre/${BASE_IMG}"
+IMG="${BASE_IMG}:latest"
+
+# This must correspond to the IMAGE_TAG generated by app-sre: As long as we
+# *don't* set `use_channel_in_image_tag: true` in the saas file, this is the
+# first $NUM digits of the current commit hash. $NUM is 7 by default (if we
+# don't override it by setting `hash_length` in the saas file).
+GIT_HASH=`git rev-parse --short=7 HEAD`
+
```
2uasimojo, line 20: If we're going to ref images by digest, we need to [short out if the image exists](https://github.com/openshift/boilerplate/blob/328a6a5330ca758439829122dbdfafca9dcc79cf/boilerplate/_lib/common.sh#L85), both here and in app_sre_build_deploy.sh.

hack/app-sre/kustomization.yaml
```diff
@@ -0,0 +1,46 @@
...
+- ../../config/crds/hive.openshift.io_machinepools.yaml
+- ../../config/crds/hive.openshift.io_selectorsyncidentityproviders.yaml
+- ../../config/crds/hive.openshift.io_selectorsyncsets.yaml
+- ../../config/crds/hive.openshift.io_syncidentityproviders.yaml
+- ../../config/crds/hive.openshift.io_syncsets.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+# Use app-sre-supplied variables to pull, by digest, the image for the current commit
+images:
+- digest: ${IMAGE_DIGEST}
```
2uasimojo, line 40: Alternative: use the tag and don't worry about Disappearing Digest Syndrome.

2uasimojo (commented):
/hold

Address DDS one way or another.

hack/app-sre/kustomization.yaml
```diff
@@ -0,0 +1,46 @@
...
+  name: registry.ci.openshift.org/openshift/hive-v4.0:hive
+  newName: ${REGISTRY_IMG}
+# Label all the deployed artifacts with the commit hash so we can tell everything is in sync
+commonLabels:
+  managed.openshift.io/gitHash: ${IMAGE_TAG}
```
maorfr, line 45: this will cause all resources to be applied on every deployment. i don't think this is desired. smells like fear of OLM ;)

hack/app-sre/kustomization.yaml
```diff
@@ -0,0 +1,46 @@
+# This kustomize configuration deploys what is required for the hive-operator
+# *without* OLM.
+# Except:
+# It does not contain a HiveConfig.
+
+# Adds namespace to all resources.
+namespace: hive
```
maorfr, line 7: you shouldn't add the namespace to the resources, this is determined by app-interface (think of a possible future hive deployment in multiple namespaces on the same cluster)

hack/app-sre/kustomization.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# The CRD resources are synced automatically by the `build-app-sre-template` make target.
+# The other resources should be curated manually.
+resources:
+- ../../config/namespace.yaml
```
maorfr, line 12: the namespace is created by app-interface and shouldn't be applied.

hack/app-sre/kustomization.yaml
maorfr: i think we should go as conventional as possible and deploy this with IMAGE_TAG?

2uasimojo: Yeah, probably. And cut over to by-digest separately. Principle of least change.

hack/app-sre/kustomization.yaml
2uasimojo: > this will cause all resources to be applied on every deployment.

Because appsre is comparing old/new and skipping resources if they match, or because `oc apply` is doing that?

> i don't think this is desired.

Why not? IIUC `oc apply` does it with an update, not a delete/create, so it should be non-invasive.

> smells like fear of OLM ;)

I mean, yes, I'm doing it deliberately as an answer to the OLM feature where, if the CSV is applied successfully, you can be sure* that all the resources are at the right version.

We're going to need an SOP for debugging problems in this new world, and "make sure all the resources are at the right version" needs to be a doable step. Do you have suggestions for other ways we could do that?

*for most values of "sure"

hack/app-sre/kustomization.yaml
2uasimojo: Ack, removed.

hack/app-sre/kustomization.yaml
maorfr: app-interface takes care of this for you. if a resource can not be applied, the deployment will fail. we don't want all resources applied every time. there is not another service we run that does this.

hack/app-sre/kustomization.yaml
2uasimojo: Done.

hack/app-sre/build-push.sh
2uasimojo: n/a, but left TODO

2uasimojo:
Updated. I'll keep the /hold on this until we've resolved the [discussion](https://github.com/openshift/hive/pull/1486#discussion_r683157900) with @maorfr about labeling objects with the git hash.

hack/app-sre/kustomization.yaml
gregsheremeta: > we don't want all resources applied every time

Can you explain why? I most certainly thought we did want that. What is the downside or risk of just always applying everything? Why wouldn't we just let `kubectl apply -f` figure it out?

dgoodwin: I would assume that is what app-interface is doing, kubectl apply everything, if it fails fail the deployment. Is this correct @maorfr ?

dgoodwin: I think I'd agree with @maorfr, we shouldn't need to verify the deployed version, our gitops tooling should do that for us and there shouldn't be any need to update resources that haven't changed.

hack/app-sre/build-push.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
abhinavdahiya, line 1: i don't see this script in use in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/22093/diffs ??

hack/app-sre/build-push.sh
```diff
@@ -0,0 +1,31 @@
...
+IMG="${BASE_IMG}:latest"
+
+# This must correspond to the IMAGE_TAG generated by app-sre: As long as we
+# *don't* set `use_channel_in_image_tag: true` in the saas file, this is the
+# first $NUM digits of the current commit hash. $NUM is 7 by default (if we
+# don't override it by setting `hash_length` in the saas file).
+GIT_HASH=`git rev-parse --short=7 HEAD`
+
+# build the image
+BUILD_CMD="docker build" IMG="$IMG" make GO_REQUIRED_MIN_VERSION:= docker-build
```
abhinavdahiya, line 22: what was previous method of building hive and pushing the container image? just to compare that with this new flow??

hack/app-sre/kustomization.yaml
```diff
@@ -0,0 +1,43 @@
+# This kustomize configuration deploys what is required for the hive-operator
```
abhinavdahiya, line 1: https://github.com/openshift/hive/blob/master/overlays/template/kustomization.yaml

why do we need to make a new one?

abhinavdahiya: there's also https://github.com/openshift/hive/blob/master/config/kustomization.yaml

abhinavdahiya: we already do make deploy to deploy using kustomize to our dev enviornment. how different is the app-sre that demands a completely separate step.

hack/app-sre/saas-objects.yaml
```diff
@@ -0,0 +1,6504 @@
+apiVersion: apiextensions.k8s.io/v1
```
abhinavdahiya, line 1: is it necessary to to store this duplicated data? also do we update this everytime we update the CRD resources?

hack/app-sre/saas-template.yaml
```diff
@@ -0,0 +1,6584 @@
+apiVersion: v1
```
abhinavdahiya, line 1: this again looks like duplicated data?

abhinavdahiya (commented):
i'm not sure why a completely new workflow outside of make deploy is required? maybe that's just me as i don't know how app-sre expects the input to look like?

hack/app-sre/kustomization.yaml
2uasimojo: Okay, I've taken it out. Easy enough to put back later if we find we need it.

hack/app-sre/kustomization.yaml
2uasimojo: Different enough :) Or I would have done it with the existing kustomizations.

The whole point of kustomize is to make it easy to do this kind of thing without needing multiple copies of the manifests.

hack/app-sre/saas-objects.yaml
2uasimojo: The duplication is no different in spirit to stuffing all the CRDs etc into the OLM bundle. (Or vendoring deps for that matter.)

> also do we update this everytime we update the CRD resources?

Yes, this file is generated and put in the `make verify` path for exactly that reason: so that the various manifests can live in one place, and we're forced to update automatically any time they change.

hack/app-sre/saas-template.yaml
2uasimojo: As above.

hack/app-sre/build-push.sh
2uasimojo: See [hack/app_sre_build_deploy.sh](https://github.com/openshift/hive/blob/0544ad7891b18e6ed34cd5ce39b98fb604e7ce8a/hack/app_sre_build_deploy.sh). This is a subset: the first part (build/push the operator image) but not the second part (build/push the OLM bundles).

maorfr: i propose to keep using the original script. when the time comes and we are off OLM - you can just remove the redundant section from the script.

hack/app-sre/kustomization.yaml
```diff
@@ -0,0 +1,39 @@
...
+# Except:
+# It does not contain a Namespace or a HiveConfig. These are created by app-sre.
+
+# The CRD resources are synced automatically by the `build-app-sre-template` make target.
+# The other resources should be curated manually.
+resources:
+- ../../config/operator/operator_role.yaml
+- ../../config/operator/operator_role_binding.yaml
+- ../../config/operator/operator_deployment.yaml
+- ../../config/crds/hiveinternal.openshift.io_clustersyncleases.yaml
```
abhinavdahiya, line 12: question: can we separate out the resources into sections that clearly show these are generated (do not touch) and these aren't

```
resources:
- curated-manually1
- curated-manually-2

## GENERATED: DO NOT TOUCH
resources:
- generated1
- generated2
```

or something of that sort?

hack/app-sre/saas-objects.yaml
abhinavdahiya: would it better to keep the scripts in hack, but actual objects in a directory that doesn't say hack ?
when making a PR, this will create a diff in hack directory which can look odd.

also is it possible to atleast remove this objects list file? and just maybe create this as temp file and create the saas bundle from it, and then delete this?
one less diff when making changes to CRDs

maorfr:
it will probably also be good to add a check to the pr-check that the templates were generated (generate and check that git is not in a dirty state?)

Makefile
```diff
@@ -195,6 +195,19 @@ deploy: ensure-kustomize install
 	# Create a default basic HiveConfig so the operator will deploy Hive
 	oc process --local=true -p HIVE_NS=${HIVE_NS} -p LOG_LEVEL=${LOG_LEVEL} -f config/templates/hiveconfig.yaml | oc apply -f -
 
+# Build the template file used for direct (OLM-less) deploy by app-sre
+build-app-sre-template: ensure-kustomize
```
2uasimojo, line 199: Need to put this in the way of `update` as well.

hack/app-sre/build-push.sh
2uasimojo: Right, I've updated the commit message accordingly:

> The other new artifact is hack/app-sre/build-push.sh. This is the script that will eventually be run by app-sre's ci-int job for direct deploy. It is a subset of hack/app_sre_build_deploy.sh: it only builds the hive operator image and pushes it to quay. **(Note that this script will not be used by appsre until all environments are migrated to direct deploy.)**

hack/app-sre/build-push.sh
2uasimojo: Right, we'll keep using the original for now (as updated in [!22093](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/22093)). Eventually I'd like to move (so everything app-sre is under hack/app-sre) and rename (to reflect what it's actually doing). If someone feels strongly about it, I can get rid of this script from this PR.

hack/app-sre/saas-objects.yaml
2uasimojo: > would it better to keep the scripts in hack, but actual objects in a directory that doesn't say hack ?

FWIW, the [OLM template](https://github.com/openshift/hive/blob/master/hack/olm-registry/olm-artifacts-template.yaml) is under hack. Also, I really want to group everything app-sre into the same directory. (TBH, not sure why the initial dir needs to be named "hack" in the first place.)

> also is it possible to atleast remove this objects list file?

Yeah, I can do that.

Makefile
2uasimojo: Done.

2uasimojo:
> it will probably also be good to add a check to the pr-check that the templates were generated (generate and check that git is not in a dirty state?)

Yes, [this target](https://github.com/openshift/hive/blob/93e02ef3c0ba91ba7a5db43dc6f33ac6a7f6b777/Makefile#L219-L220) does the generate-and-dirty-check, and [this dep](https://github.com/openshift/hive/blob/93e02ef3c0ba91ba7a5db43dc6f33ac6a7f6b777/Makefile#L221) adds it to the `verify` target which is the subject of an existing [CI job](https://github.com/openshift/release/blob/ef6e4745d0b734207082dfc382c59a597d8f2415/ci-operator/config/openshift/hive/openshift-hive-master.yaml#L42-L45) of the same name.

hack/app-sre/kustomization.yaml
2uasimojo: Sorry, no, I tried it and kustomize doesn't like it. We could probably do something similar by using separate kustomization files for those resource groups, but I _really_ dislike that kind of multi-file spread. (Bad enough the number of different files we're having to maintain as it is.)

hack/app-sre/build-push.sh
2uasimojo: I removed it.

2uasimojo:
I think I've addressed all the concerns now. Ready for hopefully-final review.

/hold cancel

PR #1487, opened 2021-08-05
Hide internal crs

suhanime:
This commit adds the annotation to the CSV generated for OperatorHub,
that would hide the following Hive CRs from the UI
- ClusterSync
- ClusterSyncLease
- FakeClusterInstall
- Checkpoints
- ClusterDeprovision
- ClusterProvision
- ClusterState
- MachinePoolLease

The annotation added is
`operators.operatorframework.io/internal-objects` as per [OpenShift Docs](https://docs.openshift.com/container-platform/4.7/operators/operator_sdk/osdk-generating-csvs.html#osdk-hiding-internal-objects_osdk-generating-csvs)

xref: https://issues.redhat.com/browse/HIVE-1575

/assign @dgoodwin 
/cc @2uasimojo

### DIFF
--- a/config/templates/hive-csv-template.yaml
+++ b/config/templates/hive-csv-template.yaml
@@ -14,6 +14,7 @@ metadata:
     support: Hive Team
     alm-examples: |-
       [{"apiVersion":"hive.openshift.io/v1","kind":"HiveConfig","metadata":{"name":"hive"},"spec":{"managedDomains":[{"aws":{"credentialsSecretRef":{"name":"my-route53-creds"}},"domains":["my-base-domain.example.com"]}]}}]
+    operators.operatorframework.io/internal-objects: '["checkpoints.hive.openshift.io","clusterdeprovisions.hive.openshift.io","clusterprovisions.hive.openshift.io","clusterstates.hive.openshift.io","machinepoolnameleases.hive.openshift.io","clustersyncleases.hiveinternal.openshift.io","clustersyncs.hiveinternal.openshift.io","fakeclusterinstalls.hiveinternal.openshift.io"]'
 spec:
   displayName: Hive for Red Hat OpenShift
   icon:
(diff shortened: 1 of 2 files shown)


config/templates/hive-csv-template.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
     support: Hive Team
     alm-examples: |-
       [{"apiVersion":"hive.openshift.io/v1","kind":"HiveConfig","metadata":{"name":"hive"},"spec":{"managedDomains":[{"aws":{"credentialsSecretRef":{"name":"my-route53-creds"}},"domains":["my-base-domain.example.com"]}]}}]
+    operators.operatorframework.io/internal-objects: '["checkpoints.hive.openshift.io","clusterdeprovisions.hive.openshift.io","clusterprovisions.hive.openshift.io","clusterstates.hive.openshift.io","machinepoolnameleases.hive.openshift.io","clustersyncleases.hiveinternal.openshift.io","clustersyncs.hiveinternal.openshift.io","fakeclusterinstalls.hiveinternal.openshift.io"]'
```
2uasimojo, line 17: If `"checkpoints.hive.openshift.io","clusterdeprovisions.hive.openshift.io","clusterprovisions.hive.openshift.io","clusterstates.hive.openshift.io","machinepoolnameleases.hive.openshift.io"` are to be considered internal, is there a reason we don't move them to the `hiveinternal` group?

2uasimojo (commented):
This seems sane and pretty harmless.

config/templates/hive-csv-template.yaml
dgoodwin: Some of them I'd definitely like to be move to internal but the process to do so is astonishingly hard and just not worth it. Others I would consider fine to programatically use like deprovision/provision, clusterstate. leases for sure i'd like in internal and checkpoints maybe.

PR #1490, opened 2021-08-06
Scrub errors for clusters unreachable due to cert dates.

dgoodwin:
Still lightly hotlooping because of the current time in this error message. 

We need to look into Abhinav's don't update if the reason hasn't changed for more controllers, but this should be a quick fix today for Monday's push.

### DIFF
--- a/pkg/controller/utils/errorscrub.go
+++ b/pkg/controller/utils/errorscrub.go
@@ -5,7 +5,8 @@ import (
 )
 
 var (
-	newlineTabRE = regexp.MustCompile(`\n\t`)
+	newlineTabRE           = regexp.MustCompile(`\n\t`)
+	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
 	// aws
 	awsRequestIDRE = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
 	// azure
@@ -20,6 +21,7 @@ func ErrorScrub(err error) string {
 	}
 	s := newlineTabRE.ReplaceAllString(err.Error(), ", ")
 	s = awsRequestIDRE.ReplaceAllString(s, "")
+	s = certificateTimeErrorRE.ReplaceAllString(s, "")
 	// if Azure error, return just the error description
 	match := azureErrorDescriptionRE.FindStringSubmatch(s)
 	if len(match) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/controller/utils/errorscrub.go
```diff
@@ -5,7 +5,8 @@ import (
 )
 
 var (
-	newlineTabRE = regexp.MustCompile(`\n\t`)
+	newlineTabRE           = regexp.MustCompile(`\n\t`)
+	certificateTimeErrorRE = regexp.MustCompile(`(: current time .* is after .*)`)
```
2uasimojo, line 9: nit: The parens should be unnecessary (you're not using the capture group).

I'm also a little concerned about the broadness of the `.*`s -- I've seen regexes like this unintentionally match huge chunks of a string because something _way later_ happens to also say "is after". Perhaps at least match non-whitespace for the timestamps?
```suggestion
	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
```

pkg/controller/utils/errorscrub_test.go
```diff
@@ -39,6 +39,11 @@ func TestErrorScrubber(t *testing.T) {
...
 			expected: "AADSTS7000215: Invalid client secret is provided.",
 		},
+		{
+			name:     "invalid certificate date",
+			input:    errors.New("[an error on the server (\"\") has prevented the request from succeeding, Get \"https://api.foobar.openshiftapps.com:6443/api?timeout=32s\": x509: certificate has expired or is not yet valid: current time 2021-08-06T11:54:02Z is after 2021-07-26T09:26:41Z]"),
+			expected: "[an error on the server (\"\") has prevented the request from succeeding, Get \"https://api.foobar.openshiftapps.com:6443/api?timeout=32s\": x509: certificate has expired or is not yet valid",
```
2uasimojo, line 45: FWIW, backtick strings are allowed anywhere (not just regexes and kube types) to avoid all the backslashes:
```suggestion
			input:    errors.New(`[an error on the server ("") has prevented the request from succeeding, Get "https://api.foobar.openshiftapps.com:6443/api?timeout=32s": x509: certificate has expired or is not yet valid: current time 2021-08-06T11:54:02Z is after 2021-07-26T09:26:41Z]`),
			expected: `[an error on the server ("") has prevented the request from succeeding, Get "https://api.foobar.openshiftapps.com:6443/api?timeout=32s": x509: certificate has expired or is not yet valid`,
```

2uasimojo (commented):
Works.

One probably-excessively-paranoid concern and a couple of niggles noted inline. Fix if you wish, or unhold and merge :)

/lgtm
/hold

dgoodwin:
Thanks, I have pushed an update with those changes.

@abutcher would love to get this out today to calm down the logs.

PR #1492, opened 2021-08-08
change admissionregistration.k8s.io api from v1beta1 to v1

osherdp:
/cc @osherdp 
/hold
(just wanted to see how the builds are with this change)


osherdp:
went into dependency hell
/close

PR #1493, opened 2021-08-09
Fix transfer of PreserveOnDelete to DNSZone when deleted.

dgoodwin:
Second attempt was still bad if the objects were already deleted, the code path was skipped. :(

/assign @2uasimojo 
/cc @suhanime

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -494,6 +494,16 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, err
 	}
 
+	if cd.Spec.ManageDNS {
+		changed, err := r.ensureDNSZonePreserveOnDelete(cd, cdLog)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if changed {
+			return reconcile.Result{}, nil
+		}
+	}
+
 	if cd.DeletionTimestamp != nil {
 		if !controllerutils.HasFinalizer(cd, hivev1.FinalizerDeprovision) {
 			// Make sure we have no deprovision underway metric even though this was probably cleared when we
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1426,6 +1436,35 @@ func (r *ReconcileClusterDeployment) setDNSDelayMetric(cd *hivev1.ClusterDeploym
 	return modified, nil
 }
 
+// ensureDNSZonePreserveOnDelete makes sure the DNSZone, if one exists, has a
+// matching PreserveOnDelete setting with it's ClusterDeployment. Returns true
```
2uasimojo, line 1440:
```suggestion
// matching PreserveOnDelete setting with its ClusterDeployment. Returns true
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1426,6 +1436,35 @@ func (r *ReconcileClusterDeployment) setDNSDelayMetric(cd *hivev1.ClusterDeploym
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		logger.WithError(err).Error("failed to fetch DNS zone")
+		return false, err
+	} else if err != nil {
+		// DNSZone doesn't exist yet
+		return false, nil
+	}
+
+	if dnsZone.Spec.PreserveOnDelete != cd.Spec.PreserveOnDelete {
+		logger.Infof("setting DNSZone PreserveOnDelete to match ClusterDeployment PreserveOnDelete: %v", cd.Spec.PreserveOnDelete)
```
2uasimojo, line 1456: Structured logs?

Also could use `%t` for the bool value.

2uasimojo (commented):
So with *this* PR we may need multiple reconciles on a new CD before this code path gets hit.

The first time through we haven't done `ensureManagedDNSZone` yet, so we'll bail out of `ensureDNSZonePreserveOnDelete early. That reconcile will [createManagedDNSZone](https://github.com/openshift/hive/blob/32a0a5cce601f4d35aae01ca9ddcfe347a0c574c/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1489) and then (afaict) [return without requeueing](https://github.com/openshift/hive/blob/32a0a5cce601f4d35aae01ca9ddcfe347a0c574c/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L583). Since this controller is [watching DNSZones](https://github.com/openshift/hive/blob/32a0a5cce601f4d35aae01ca9ddcfe347a0c574c/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L221), the creation of the DNSZone will trigger the subsequent reconcile that will get us into your new update code.

/lgtm

/hold in case you want to fix my nits, or in case I've gotten my analysis wrong above.

dgoodwin:
> So with _this_ PR we may need multiple reconciles on a new CD before this code path gets hit.
> 
> The first time through we haven't done `ensureManagedDNSZone` yet, so we'll bail out of `ensureDNSZonePreserveOnDelete early. That reconcile will [createManagedDNSZone](https://github.com/openshift/hive/blob/32a0a5cce601f4d35aae01ca9ddcfe347a0c574c/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1489) and then (afaict) [return without requeueing](https://github.com/openshift/hive/blob/32a0a5cce601f4d35aae01ca9ddcfe347a0c574c/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L583). Since this controller is [watching DNSZones](https://github.com/openshift/hive/blob/32a0a5cce601f4d35aae01ca9ddcfe347a0c574c/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L221), the creation of the DNSZone will trigger the subsequent reconcile that will get us into your new update code.

The creation of the DNSZone takes PreserveOnDelete into account, so this new code won't trigger if PreserveOnDelete was already set at DNSZone creation time. This would only trigger when adding or removing it later.

dgoodwin:
Nits addressed! Thanks.

PR #1494, opened 2021-08-09
More updates for k8s 1.22

2uasimojo:
OCP 4.9 uses k8s 1.22, where several deprecated objects have been removed [1]. We got partway via previous commits [2][3]. This hopefully finishes the job, addressing:
- ValidatingWebhookConfiguration
- CertificateSigningRequest
- some remaining CustomResourceDefinition references in test code

[1] https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22
[2] openshift#1395
[3] openshift#1468

[HIVE-1626](https://issues.redhat.com/browse/HIVE-1626)
[HIVE-1627](https://issues.redhat.com/browse/HIVE-1627)

### DIFF
--- a/config/hiveadmission/clusterdeployment-webhook.yaml
+++ b/config/hiveadmission/clusterdeployment-webhook.yaml
@@ -1,10 +1,12 @@
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: clusterdeploymentvalidators.admission.hive.openshift.io
 webhooks:
 - name: clusterdeploymentvalidators.admission.hive.openshift.io
+  admissionReviewVersions:
+  - v1beta1
   clientConfig:
     service:
       # reach the webhook via the registered aggregated API
--- a/pkg/operator/util/admissionregistration.go
+++ b/pkg/operator/util/admissionregistration.go
@@ -1,14 +1,14 @@
 package util
 
 import (
-	admregv1 "k8s.io/api/admissionregistration/v1beta1"
+	admregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 )
 
-// ReadValidatingWebhookConfigurationV1Beta1OrDie reads a ValidatingWebhookConfiguration,
+// ReadValidatingWebhookConfigurationV1OrDie reads a ValidatingWebhookConfiguration,
 // as this is not yet added to library-go.
-func ReadValidatingWebhookConfigurationV1Beta1OrDie(objBytes []byte, scheme *runtime.Scheme) *admregv1.ValidatingWebhookConfiguration {
+func ReadValidatingWebhookConfigurationV1OrDie(objBytes []byte, scheme *runtime.Scheme) *admregv1.ValidatingWebhookConfiguration {
 	apiExtensionsCodecs := serializer.NewCodecFactory(scheme)
 
 	requiredObj, err := runtime.Decode(apiExtensionsCodecs.UniversalDecoder(admregv1.SchemeGroupVersion), objBytes)
--- a/test/e2e/common/client.go
+++ b/test/e2e/common/client.go
@@ -7,7 +7,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
-	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/client-go/dynamic"
 	kclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -21,7 +21,7 @@ import (
 )
 
 func init() {
-	apiextv1beta1.AddToScheme(scheme.Scheme)
+	apiextv1.AddToScheme(scheme.Scheme)
 	hiveapis.AddToScheme(scheme.Scheme)
 	admissionv1beta1.AddToScheme(scheme.Scheme)
 	autoscalingv1.SchemeBuilder.AddToScheme(scheme.Scheme)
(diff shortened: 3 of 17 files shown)


dgoodwin:
config/hiveadmission/ still referencing v1beta1, should this be changed?

What happens on rollout to an existing hive with v1beta1 hooks?

joelddiaz:
> What happens on rollout to an existing hive with v1beta1 hooks?
IIRC based on what happened with CCO, it "just worked". You can in-place update your CRD to v1.

2uasimojo:
> config/hiveadmission/ still referencing v1beta1, should this be changed?

Most likely, yes. I should have marked this WIP (though I was fairly confident e2e would fail anyway).

/hold

> > What happens on rollout to an existing hive with v1beta1 hooks?

> IIRC based on what happened with CCO, it "just worked". You can in-place update your CRD to v1.

Here's hoping :)

I'm working on setting up a 4.9-hub version of e2e, which should help us verify at least some parts of this. Testing the upgrade... might have to do that manually.

osherdp:
/lgtm
(at least to my eyes)

config/hiveadmission/clusterdeployment-webhook.yaml
```diff
@@ -1,10 +1,12 @@
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: clusterdeploymentvalidators.admission.hive.openshift.io
 webhooks:
 - name: clusterdeploymentvalidators.admission.hive.openshift.io
+  admissionReviewVersions:
+  - v1
```
2uasimojo, line 9: [Now a required field](https://docs.openshift.com/container-platform/4.4/rest_api/extension_apis/validatingwebhookconfiguration-admissionregistration-k8s-io-v1.html).

2uasimojo:
/hold cancel

This should be good to go now.

(Note: I would have liked to land https://github.com/openshift/release/pull/21007 (run e2e on 4.9) first, but it's blocked on https://issues.redhat.com/browse/DPTP-2448 and we don't know when that's going to be fixed.)

/assign @joelddiaz 

Thanks for the help Joel!

joelddiaz (commented):
/lgtm

joelddiaz:
i was trying to reproduce these CI failures, and it looks like some of the utility functions need updating to return v1 objects (called from the hive operator)
https://github.com/openshift/hive/blob/master/pkg/operator/util/admissionregistration.go#L11
https://github.com/openshift/hive/blob/master/pkg/operator/util/admissionregistration.go#L23

2uasimojo:
> i was trying to reproduce these CI failures, and it looks like some of the utility functions need updating to return v1 objects (called from the hive operator)
> https://github.com/openshift/hive/blob/master/pkg/operator/util/admissionregistration.go#L11
> https://github.com/openshift/hive/blob/master/pkg/operator/util/admissionregistration.go#L23

And https://github.com/openshift/hive/blob/master/pkg/operator/hive/hiveadmission.go#L28?

pkg/operator/util/admissionregistration.go
```diff
@@ -1,14 +1,14 @@
...
+	admregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 )
 
-// ReadValidatingWebhookConfigurationV1Beta1OrDie reads a ValidatingWebhookConfiguration,
+// ReadValidatingWebhookConfigurationV1OrDie reads a ValidatingWebhookConfiguration,
 // as this is not yet added to library-go.
-func ReadValidatingWebhookConfigurationV1Beta1OrDie(objBytes []byte, scheme *runtime.Scheme) *admregv1.ValidatingWebhookConfiguration {
+func ReadValidatingWebhookConfigurationV1OrDie(objBytes []byte, scheme *runtime.Scheme) *admregv1.ValidatingWebhookConfiguration {
```
joelddiaz, line 11: Are there no users of the ReadMutatingWebhookConfigurationV1Beta1OrDie()? Can we just remove it?

test/e2e/common/client.go
```diff
@@ -6,8 +6,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
+	admissionv1 "k8s.io/api/admission/v1"
```
joelddiaz, line 9: Do we really need to keep admissionv1beta1 in the line below?

config/hiveadmission/clusterdeployment-webhook.yaml
```diff
@@ -1,10 +1,13 @@
...
+  - v1beta1
```
joelddiaz, line 10: How does registering both v1beta1 and v1 works with our present webhook code?

pkg/operator/util/admissionregistration.go
2uasimojo: I renamed the func and here and its (one) [usage](https://github.com/openshift/hive/pull/1494/files#diff-b90095902244c03464a07cf8fd6ab580c5d445534cc97b1c750b2382792decebR143).

test/e2e/common/client.go
2uasimojo: It seems to be the other way around. I can't figure out where the v1beta1 scheme is getting registered in the operator, but apparently it is, because the error if I keep both v1 and v1beta1 in `admissionReviewVersions` in the webhook configs indicates that the v1 scheme isn't registered; but if I take it out I don't get that complaint. My theory was that I need to register both in order to list both in the configs.

Anyway, I'm going to try to remove v1 here and in the configs and see if that works.

config/hiveadmission/clusterdeployment-webhook.yaml
2uasimojo: I had thought this was a list of versions to *try*, in order, and it would accept the first one that worked. But it's apparently not that simple. I'm removing v1 from these.

2uasimojo:
/test e2e

This looks like it was a “normal flake” 🤞

pkg/operator/util/admissionregistration.go
joelddiaz: I was asking about ReadMUTATINGWebhookConfigurationV1OrDie() (not Validating). You didn't touch it in the PR, but it the only other function in this file.

2uasimojo: Oh, my bad.

How about I remove it in a [separate PR](https://github.com/openshift/hive/pull/1508)?

joelddiaz (commented):
/lgtm

PR #1495, opened 2021-08-10
add NoWorkerNodes to install-log-regexes-configmap

gregsheremeta:
It is a common problem that STS-mode clusters have something
wrong with the operator roles -- typically they're bound to
the wrong (old) OIDC provider. The first place an install notices
that is when machine-api tries to create the workers.

/assign @abutcher


gregsheremeta:
/retest

/cc @dgoodwin 
for ack

PR #1497, opened 2021-08-11
Add requirements met condition to provision delay metric labels.

dgoodwin:
Encountered a cluster today with a missing image set, requirements met false, but the metric did not inform us.

### DIFF
--- a/pkg/controller/metrics/provision_underway_collector.go
+++ b/pkg/controller/metrics/provision_underway_collector.go
@@ -18,6 +18,7 @@ var (
 	// ClusterDeployment conditions that could indicate provisioning problems with a cluster
 	// First condition appearing True will be used in the metric labels.
 	provisioningDelayCondition = [...]hivev1.ClusterDeploymentConditionType{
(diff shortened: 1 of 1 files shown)


pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -18,6 +18,7 @@ var (
 	// ClusterDeployment conditions that could indicate provisioning problems with a cluster
 	// First condition appearing True will be used in the metric labels.
 	provisioningDelayCondition = [...]hivev1.ClusterDeploymentConditionType{
+		hivev1.RequirementsMet,
```
suhanime, line 21: The reason I didn't suggest it before is because I didn't want alerts for every tried - but not updated provision, but I can see a value in it from OSD perspective

suhanime (commented):
@dgoodwin do you need to update the api?

>  vet: pkg/controller/metrics/provision_underway_collector.go:21:10: RequirementsMet not declared by package v1
make: *** [verify-govet] Error 2

PR #1500, opened 2021-08-12
MachinePool documentation updates

showeimer:
Signed-off-by: Sho Weimer <sweimer@redhat.com>

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -486,6 +486,8 @@ openstack:
 
 ### Machine Pools
 
+`MachinePool` is a YAML configuration by which you can create and scale work nodes on a deployed cluster. A `MachinePool` will create `MachineSet` resources on the deployed cluster equivalent to the number of configured Availability Zones (AZ).
+
 To manage `MachinePools` Day 2, you need to define these as well. The definition of the worker pool should mostly match what was specified in `InstallConfig` to prevent replacement of all worker nodes.
 
 `InstallConfig` is limited to the one worker pool, but Hive can sync additional machine pools Day 2.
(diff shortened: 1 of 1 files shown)


docs/using-hive.md
```diff
@@ -558,6 +560,73 @@ openstack:
...
+  autoscaling:
+    minReplicas: 3
+    maxReplicas: 6
+```
+
+The number of minimum replicas must be equivalent to the number of configured Availability Zones.
+
+The `spec.replicas` and `spec.autoscaling` configurations cannot be configured simultaneously.
+
+The `spec.autoscaling.maxReplicas` is an optional field, if it is not configured, then nodes will be auto-scaled without restriction based on resource utilization needs.
```
2uasimojo, line 622:
```suggestion
The `spec.autoscaling.maxReplicas` is an optional field. If it is not configured, then nodes will be auto-scaled without restriction based on resource utilization needs.
```

docs/using-hive.md
```diff
@@ -486,6 +486,8 @@ openstack:
 
 ### Machine Pools
 
+`MachinePool` is a YAML configuration by which you can create and scale work nodes on a deployed cluster. A `MachinePool` will create `MachineSet` resources on the deployed cluster equivalent to the number of configured Availability Zones (AZ).
```
2uasimojo, line 489:
```suggestion
`MachinePool` is a YAML configuration by which you can create and scale worker nodes on a deployed cluster. A `MachinePool` will create `MachineSet` resources on the deployed cluster equivalent to the number of configured Availability Zones (AZ).
```

docs/using-hive.md
```diff
@@ -558,6 +560,73 @@ openstack:
...
+
+The number of minimum replicas must be equivalent to the number of configured Availability Zones.
+
+The `spec.replicas` and `spec.autoscaling` configurations cannot be configured simultaneously.
+
+The `spec.autoscaling.maxReplicas` is an optional field, if it is not configured, then nodes will be auto-scaled without restriction based on resource utilization needs.
+
+##### Integration with Horiztonal Pod Autoscalers
+
+A `MachinePool` configured to auto-scaling mode creates a `ClusterAutoscaler` on the deployed cluster. `ClusterAutoscalers` can co-exist and work with Horiztonal Pod Autoscalers to ensure that there are enough available nodes to meet the auto-scaled pod replica count requirements. See excerpt from OpenShift documentation:
```
2uasimojo, line 626: Can this be linked to said documentation?

2uasimojo (commented):
Thanks for adding this!

I'm only useful for minor typographical comments -- we'll want someone more familiar with the actual content to do a real review.

docs/using-hive.md
showeimer: Hey Eric, thanks for the quick reviews.  I can change the copy/paste to a link, but I wasn't sure if there would be some concern about the link getting stale since it would be pinned to a specific OCP version doc.

2uasimojo: Nono, keep the quote -- just add the link as well.

(Agree about stuff getting stale, it's always a problem.)

docs/using-hive.md
```diff
@@ -486,6 +486,8 @@ openstack:
 
 ### Machine Pools
 
+`MachinePool` is a YAML configuration by which you can create and scale worker nodes on a deployed cluster. A `MachinePool` will create `MachineSet` resources on the deployed cluster equivalent to the number of configured Availability Zones (AZ).
```
dgoodwin, line 489:
```suggestion
`MachinePool` is a YAML configuration by which you can create and scale worker nodes on a deployed cluster. A `MachinePool` will create `MachineSet` resources on the deployed cluster. If supported on your cloud, those MachineSets will automatically span all AZs, or you can specify an explicit list.
```

docs/using-hive.md
```diff
@@ -558,6 +560,73 @@ openstack:
...
+        size: 22
+        type: gp2
+      type: m4.xlarge
+      zones:
+        - us-east-1a
+        - us-east-1b
+  replicas: 3
+```
+
+If the Availability Zones are not configured in the `MachinePool`, then all of the AZs in the region will be used and a `MachineSet` resource will be created for each AZ.
```
dgoodwin, line 590: Likely only true on AWS, GCP, and Azure.

docs/using-hive.md
```diff
@@ -558,6 +560,73 @@ openstack:
...
+
+#### Auto-scaling
+
+`MachinePools` can be configured to auto-scale to scale up or down worker nodes as needed based on resource utilization of the deployed cluster by creating a `ClusterAutoscaler` on the deployed cluster.
```
dgoodwin, line 594:
```suggestion
`MachinePools` can be configured to auto-scale the number of worker nodes as needed based on resource utilization of the deployed cluster. (this feature creates a `ClusterAutoscaler` resource in the deployed cluster)
```

docs/using-hive.md
```diff
@@ -558,6 +560,73 @@ openstack:
...
+    maxReplicas: 6
+```
+
+The number of minimum replicas must be equivalent to the number of configured Availability Zones.
+
+The `spec.replicas` and `spec.autoscaling` configurations cannot be configured simultaneously.
+
+The `spec.autoscaling.maxReplicas` is an optional field. If it is not configured, then nodes will be auto-scaled without restriction based on resource utilization needs.
+
+##### Integration with Horiztonal Pod Autoscalers
```
dgoodwin, line 624:
```suggestion
##### Integration with Horizontal Pod Autoscalers
```

dgoodwin (commented):
Thanks so much @showeimer

docs/using-hive.md
```diff
@@ -558,6 +560,73 @@ openstack:
   flavor: m1.large
 ```
 
+#### Configuring Availability Zones
+
+The desired Availability Zones (AZ) to create new worker nodes in can be specified in the `MachinePool` YAML (`spec.platform.<provider>.zones`), for example:
```
abhinavdahiya, line 565: > specified in the `MachinePool` YAML (`spec.platform.<provider>.zones`),

might want to reword it on the lines of ``specified in the `MachinePool` YAML usually `spec.platform.<provider>.zones`,`` to make it clear that different platforms can have different paths for this value.

PR #1502, opened 2021-08-17
.gitignore: exclude *.out only in repo root

2uasimojo:
The .gitignore rule `*.out` was sweeping up `vendor/github.com/go-errors/errors/cover.out`, which made vendoring inconsistent, which was caught by downstream builds (see linked issue).

We should be able to live with only excluding `*.out` (specifically `coverage.out`, from `make coverage`) from the repo root, so do that.

fixes https://github.com/open-cluster-management/backlog/issues/15331


arewm (approved):
Thanks!

PR #1504, opened 2021-08-17
Unhack: Use a patch to set CRB subject ns for direct deploy

2uasimojo:
This reverts f5eca7d71 / #1498, instead using a patch in the kustomization file to yield the same result.

[HIVE-1588](https://issues.redhat.com/browse/HIVE-1588)


abhinavdahiya:
I like the patch approach. 👍🏽 

/lgtm

PR #1506, opened 2021-08-18
RFE: Detect and replace broken clusters

suhanime:
Add enhancement doc to propose separate bucket for stand by clusters and
validate a hibernating cluster can be successfully transitioned to
running state before assigning it to a claim

xref: https://issues.redhat.com/browse/HIVE-1602

/assign @2uasimojo 
/cc @abhinavdahiya @dgoodwin

### DIFF
--- a/docs/enhancements/replace_broken_clusters.md
+++ b/docs/enhancements/replace_broken_clusters.md
@@ -1,0 +1,76 @@
+# Automatically detect and replace broken Clusters
+
+## Summary
+
+Hibernating clusters often fail to transition to running state, but are assigned to ClusterClaims today. We would want to 
+validate a cluster can be successfully resumed and is running as expected before assigning it to a claim.
+
+## Motivation
+
+As we move towards being more proactive with having clusters ready as [hot spares](https://github.com/openshift/hive/pull/1434),
+we could come across instances where stale hibernated clusters could fail to resume and still be assigned to ClusterClaims.
+There's a need to detect broken clusters and replace them automatically.
+
+### Goals
+
+Always strive to assign a running cluster to a claim
+
+### Non Goals
+
+Fixing clusters (or attached claims) once a running cluster has already been assigned.
+
+## Proposal
+
+### Stand-by clusters
+
+Currently, we consider all installed clusters which are not set to be deleted as Assignable clusters. 
+Introduce another bucket of stand by clusters, which are installed and are not set for deletion, but are hibernating.
+A cluster is only assignable if it is running. A hibernating cluster has to be resumed successfully to be added to the 
+assignable clusters bucket.
+Some factors that might inhibit successful resume of a cluster that has been hibernating for a while include unapproved 
+certificates, expired credentials and operator bugs. Introduction of a separate bucket for hibernating clusters allows 
+for a validation before it is assigned to a claim.
+
+### Transition to Assignable clusters
+
+If a claim is created and there are no assignable clusters, we would resume a hibernating cluster and check if the resume
+was successful. The validating condition could be the expected powerstate matches the actual powerstate. If we encounter
+an error, we can mark that cluster to be deleted and try to transition the next available stand-by cluster.
+Figuring out how long we wait for the resume process to complete can be tricky - as there are instances when we can wait
+indefinitely. So we can consider a timeout for that wait. Fine-tuning the said timeout would be dependent on the 
+platform and the infrastructure, hence left for the admins to decide. Default is no time-out - so waiting indefinitely 
+unless an error is encountered.
+For a cluster that is determined broken and hence marked to be deleted, the deletion would be immediate if within the
+maxConcurrent budget. As per the current ClusterPool logic, that cluster will be deleted eventually and a new cluster
+would be brought up in its place.
+
+### Timeouts in Hibernation controller
+
+Introduce the following optional timeouts in cluster deployments and corresponding fields in cluster pool 
+- spec.powerstateTransitionToRunningTimeout
+  - maximum wait time for a cluster to resume once the hibernating condition reflects cluster resuming reason
+- spec.powerstateTransitionToHibernatingTimeout
+  - maximum wait time for a cluster to hibernate once the hibernating condition reflects cluster stopping reason
+
+Once the powerstate is changed and hibernating condition reflects Resuming or Stopping reason, controller should allow 
+for a maximum wait time counting from the last transition time of the condition equal to the relevant timeout before 
+determining that the cluster is broken.
+If a cluster is broken, the hibernating condition can be updated with FailedToStart or FailedToStop reason and with a 
+message indicating it timed out waiting for the cluster to stop or start, and the cluster can be marked for deletion.
+
+These timeouts are only applicable to non-fake clusters as they begin counting from the interim state - whereas the 
+powerstate transition of fake clusters is immediate.
+
+## Risks and Mitigations
+
+### Timeout too little
+
+Cluster admins will be in charge of ensuring the mentioned timeouts are not too little, to give the cluster a fighting 
+chance to be successfully transitioned.
+
+### Cluster is running but needs additional validation
+
+If a cluster has never been hibernated or if has been running for a while, it might need some additional validation to 
+ensure it is ripe for assignment. Some known reasons for these are expired credentials, unapproved certificates and 
+cluster operators in degraded state. Validations for the same are worthwhile but out of the scope for this enhancement
+and are being pursued separately.


docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+was successful. The validating condition could be the expected powerstate matches the actual powerstate. If we encounter
+an error, we can mark that cluster to be deleted and try to transition the next available stand-by cluster.
+Figuring out how long we wait for the resume process to complete can be tricky - as there are instances when we can wait
+indefinitely. So we can consider a timeout for that wait. Fine-tuning the said timeout would be dependent on the 
+platform and the infrastructure, hence left for the admins to decide. Default is no time-out - so waiting indefinitely 
+unless an error is encountered.
+
+#### Timeout in Hibernation controller
+
+Introduce an optional spec.powerstateTransitionTimeout in cluster deployments. Once the powerstate is changed and
```
2uasimojo, line 46: I wonder if it makes sense to have separate timeouts for stopping vs resuming. Do/can those take significantly different amounts of time?

Also:
```suggestion
Introduce an optional spec.powerstateTransitionTimeout in cluster deployments, and a corresponding field in ClusterPool. Once the powerstate is changed and 
```

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+indefinitely. So we can consider a timeout for that wait. Fine-tuning the said timeout would be dependent on the 
+platform and the infrastructure, hence left for the admins to decide. Default is no time-out - so waiting indefinitely 
+unless an error is encountered.
+
+#### Timeout in Hibernation controller
+
+Introduce an optional spec.powerstateTransitionTimeout in cluster deployments. Once the powerstate is changed and 
+hibernating condition reflects Resuming or Stopping reason, controller should allow for a maximum wait time counting 
+from the last transition time of the condition equal to this timeout before marking the cluster as broken. This can be 
+done via updating the hibernation condition to reflect the reason as TimedOutWaitingToResume or TimedOutWaitingToStop.
```
2uasimojo, line 49: Alternative: Use the existing `FailedToStop`/`FailedToStart` Reasons, with the Message indicating, "Timed out waiting for the cluster to {stop|start}." or similar.

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+Once ClusterPool controller comes across a broken cluster, it will mark it for deletion and create another cluster.
+
+## Risks and Mitigations
+
+### Timeout too little
+
+For non-fake clusters, it takes about 5 minutes for a cluster to be successfully resumed. We can include a validation
```
2uasimojo, line 56: Nah, IMO that's going too far. Let the customer shoot themselves in the foot if they want.

Not to mention that we really couldn't say what a reasonable governor value should be. And imposing one could make this pretty hard to test :)

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+### Timeout too little
+
+For non-fake clusters, it takes about 5 minutes for a cluster to be successfully resumed. We can include a validation 
+that timeout cannot be set for less than 10 minutes to ensure a cluster is given a fighting chance to be resumed before
+assuming it is broken
+
+### Cluster is running but needs additional validation
+
+If a cluster has never been hibernated or if has been running for a while, it might need some additional validation to 
+ensure it is ripe for assignment. Possible scenario - expired credentials.
```
2uasimojo, line 63: Right; I feel this is the subject of https://issues.redhat.com/browse/HIVE-1601, out of scope for this enhancement.

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Alternatives
+
+Consider a timeout if a cluster has been assigned to a claim but the ClusterClaim's ClusterRunning condition has not
```
2uasimojo, line 67: Totally happy keeping this as an "alternative" :)

IMHO we should absolutely not try to muck with claims/CDs after they've "flown the coop" (except for `Lifetime`).

2uasimojo (commented):
Thanks @suhanime, this is concise and well stated. A few discussion points inline. Also want to keep this open long enough for others to get a crack at it.

docs/enhancements/replace_broken_clusters.md
suhanime: I think that's a good idea. I was testing for https://issues.redhat.com/browse/HIVE-1630 and observed that resume time is significantly greater than stop time.
image
My cluster was fresh, I am assuming a stale cluster (or a cluster that has been hibernating for more than 24 hours) would take longer to resume

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+There's a need to detect broken clusters and replace them automatically.
+
+### Goals
+
+Always strive to assign a running cluster to a claim
+
+### Non Goals
+
+Fixing clusters (or attached claims) once a running cluster has already been assigned.
+
```
abhinavdahiya, line 21: I was hoping to understand how this change will affect the users.
1. Today we assign a cluster to claim and then it changes power state. the user watch the cluster running condition to see when the cluster is ready. How will this change this behavior?
2. How do we provide information to the claim owner about the progress being made.

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+assignable clusters bucket.
+Some of the factors that might inhibit successful resume of a cluster that has been hibernating for a while include  
+unapproved certificates and expired credentials. Introduction of a separate bucket for hibernating clusters 
+allows for a validation before it is assigned to a claim.
+
+### Transition to Assignable clusters
+
+If a claim is created and there are no assignable clusters, we would resume a hibernating cluster and check if the resume
+was successful. The validating condition could be the expected powerstate matches the actual powerstate. If we encounter
+an error, we can mark that cluster to be deleted and try to transition the next available stand-by cluster.
```
abhinavdahiya, line 38: Can we expand on the workflow for how to decide when resume failed and how to handle that by deleting the cluster? deleting the cluster needs to be done in maxConcurrent budget etc.

docs/enhancements/replace_broken_clusters.md
```diff
@@ -0,0 +1,69 @@
...
+## Proposal
+
+### Stand-by clusters
+
+Currently, we consider all installed clusters which are not set to be deleted as Assignable clusters. 
+Introduce another bucket of stand by clusters, which are installed and are not set for deletion, but are hibernating.
+A cluster is only assignable if it is running. A hibernating cluster has to be resumed successfully to be added to the 
+assignable clusters bucket.
+Some of the factors that might inhibit successful resume of a cluster that has been hibernating for a while include  
+unapproved certificates and expired credentials. Introduction of a separate bucket for hibernating clusters
```
dgoodwin, line 31: Also operator bugs. (soon we will watch for operators not healthy)

docs/enhancements/replace_broken_clusters.md
dgoodwin: Do we need it on ClusterDeployment at all? It seems like it could just live on Pool and handle timeouts there. Is there any value exposing it on standalone cluster deployments?

docs/enhancements/replace_broken_clusters.md
dgoodwin: :+1:

docs/enhancements/replace_broken_clusters.md
dgoodwin: Yeah I like the main proposal better than this alternative, which would imply scripters need to be aware that you might get assigned a cluster, and then it changes on you.

docs/enhancements/replace_broken_clusters.md
2uasimojo: Good points. My take:
1. The cluster is ready as soon as the claim is assigned. (That's arguably better UX anyway.)
2. IMO nothing changes here: the `ClusterClaimPendingCondition` remains `True` with reason `NoClusters` ("No clusters in pool are ready to be claimed"). It's just that we've now got an additional stage during which this is the case. (I don't think it's a good idea to try to add another Reason here, like `ResumingCluster`, because we haven't done (and shouldn't do) anything to associate the claim with the cluster we're resuming.)

Also note that once `runningCount` is a thing and properly tuned, most/many claims will be assigned immediately anyway.

docs/enhancements/replace_broken_clusters.md
2uasimojo: > value exposing it on standalone cluster deployments?

The thing we're timing out is the Hibernating condition of the CD. It makes sense to do that via the hibernation controller, which doesn't have any pool logic at the moment.

But aside from code organization reasons, doesn't this seem like a generally useful knob to have on CDs? Today a consumer resuming a CD (whether or not it came from a pool) need some higher-level logic to keep time if they want to flag "this is taking too long". With this change, we would be giving them a simple condition reason to watch for instead.

dgoodwin: Ah I thought we already did a timeout condition reason for this, that may not be the case. Making it configurable amount of time, maybe worthwhile. If so then lets put a sub-struct for power state on the list for v2 API changes. :)

2uasimojo: > I thought we already did a timeout condition reason for this

We'll wait to kick off hibernation until syncsets have been applied; but [only for 10m, after which we'll hibernate anyway](https://github.com/openshift/hive/blob/6f060f04b8ab4fb505b20486b09bba54c631974e/pkg/controller/hibernation/hibernation_controller.go#L203-L210). Other than that and `hibernateAfter` logic, I don't see anything around timeouts. (At least not in the hibernation controller. No idea what's happening on the cloud -- except that we've observed "stuck resuming" lasting indefinitely, so...)

> sub-struct for power state on the list for v2 API changes.

Something like this:

```
spec:
  ...
  powerState:
    desiredPowerState: {Running|Hibernating}
    runningTimeout: [Duration]
    hibernatingTimeout: [Duration]
  ...
```
?

suhanime:
@2uasimojo Is this okay to merge?

PR #1507, opened 2021-08-18
Scrub AWS not authorized message which includes AWS resource ARNs

abutcher:
I started out trying to hit the AWS resource ARNs with the regex but I think it will be simpler to just target this message. There is a request id and the ARN being updated in the message.

### DIFF
--- a/pkg/controller/utils/errorscrub.go
+++ b/pkg/controller/utils/errorscrub.go
@@ -22,7 +22,7 @@ func ErrorScrub(err error) string {
 	}
 	s := newlineTabRE.ReplaceAllString(err.Error(), ", ")
 	s = awsRequestIDRE.ReplaceAllString(s, "")
-	s = awsNotAuthorized.ReplaceAllString(s, `$1 XXX $2 XXX $3 XXX $4`)
+	s = awsNotAuthorized.ReplaceAllString(s, `${1}XXX${2}XXX${3}XXX${4}`)
 	s = certificateTimeErrorRE.ReplaceAllString(s, "")
 	// if Azure error, return just the error description
 	match := azureErrorDescriptionRE.FindStringSubmatch(s)
--- a/pkg/controller/utils/errorscrub.go
+++ b/pkg/controller/utils/errorscrub.go
@@ -8,7 +8,8 @@ var (
 	newlineTabRE           = regexp.MustCompile(`\n\t`)
 	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
 	// aws
-	awsRequestIDRE = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
+	awsRequestIDRE   = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
+	awsNotAuthorized = regexp.MustCompile(`User\: \S+ is not authorized to perform\: \S+ on resource\: \S+`)
 	// azure
 	azureErrorDescriptionRE = regexp.MustCompile(`\"error_description\":\"(.*?)\\r\\n`)
 )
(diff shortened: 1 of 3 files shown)


2uasimojo (commented):
Let's

/hold

onto this for a bit. A couple of issues need to be addressed:
1. As written, this won't address the hot loop (it's not affecting requeuing, which is still based on having received the API error in the first place) or the alert (which is based on the reconcile erroring frequently, which it still does with this PR).
2. The information in the error message is (IMO) too useful to simply scrub out. We need to look at other ways of avoiding thrashing the message. (We've been discussing a new `UpdateConditionIfReasonChanged` predicate. That may or may not end up being the answer here.)

2uasimojo:
> 1. As written, this won't address the hot loop (it's not affecting requeuing, which is still based on having received the API error in the first place) or the alert (which is based on the reconcile erroring frequently, which it still does with this PR).

As @abutcher [pointed out](https://coreos.slack.com/archives/CE3ETN3J8/p1629385174043900?thread_ts=1629379363.034100&cid=CE3ETN3J8), I may be wrong about this. Returning an error from reconcile will requeue the request, but that's subject to backoff; whereas doing an actual update of the object will trigger the watchers and queue immediately without backoff.

pkg/controller/utils/errorscrub.go
```diff
@@ -8,7 +8,8 @@ var (
 	newlineTabRE           = regexp.MustCompile(`\n\t`)
 	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
 	// aws
-	awsRequestIDRE = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
+	awsRequestIDRE   = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
+	awsNotAuthorized = regexp.MustCompile(`User: \S+ is not authorized to perform: \S+ on resource: \S+`)
```
2uasimojo, line 12: If we wanted to preserve a bit more information here, we could do something like
```suggestion
	awsNotAuthorized = regexp.MustCompile(`(User:) \S+ (is not authorized to perform: \S+ on resource:) \S+`)
```
...and then on L25:

```go
	s = awsNotAuthorized.ReplaceAllString(s, "$1 XXX $2 XXX;")
```

Resulting in: `"failed to describe load balancer for the cluster: AccessDenied: User: XXX is not authorized to perform: sts:AssumeRole on resource: XXX; status code: 403"`

Or if we're confident in the format of the user/resource, even:
```suggestion
	awsNotAuthorized = regexp.MustCompile(`(User: arn:aws:sts::)[^:]+(:assumed-role/[^/]+/)\S+( is not authorized to perform: \S+ on resource: arn:aws:iam::)[^:]+(:\S+)`)
```
and
```go
	s = awsNotAuthorized.ReplaceAllString(s, "${1}XXX${2}XXX${3}XXX${4}")
```
resulting in `"failed to describe load balancer for the cluster: AccessDenied: User: arn:aws:sts::XXX:assumed-role/RH-Managed-OpenShift-Installer/XXX is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::XXX:role/ManagedOpenShift-Installer-Role, status code: 403"`

2uasimojo (commented):
I'm ready to land this, but wanted to get your thoughts on the suggestions inline first.

pkg/controller/utils/errorscrub.go
abutcher: I like the second approach where we replace the identifiers. I've committed that with a few changes. I'm not sure how to remove the extra spaces that get inserted in the replace though.

pkg/controller/utils/errorscrub.go
```diff
@@ -21,6 +22,7 @@ func ErrorScrub(err error) string {
 	}
 	s := newlineTabRE.ReplaceAllString(err.Error(), ", ")
 	s = awsRequestIDRE.ReplaceAllString(s, "")
+	s = awsNotAuthorized.ReplaceAllString(s, `$1 XXX $2 XXX $3 XXX $4`)
```
2uasimojo, line 25: You can get rid of the spaces by using curly braces:
```suggestion
	s = awsNotAuthorized.ReplaceAllString(s, `${1}XXX${2}XXX${3}XXX${4}`)
```

pkg/controller/utils/errorscrub.go
2uasimojo: See below

pkg/controller/utils/errorscrub.go
abutcher: Fantastic

2uasimojo (commented):
Beauty

/lgtm

PR #1510, opened 2021-08-23
Add PowerState Transition time metric

suhanime:
Once spec.PowerState is changed, clusters can take some time to
transition to the desired state, and sometimes they fail to transition
as well. This commit adds a metric for that transition time.

/assign @2uasimojo 
/cc @abutcher

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -87,6 +87,15 @@ var (
 		},
 		[]string{"cluster_deployment", "namespace", "cluster_type"},
 	)
+	// metricClusterDeploymentPowerStateTransitionSeconds is a prometheus metric for the number of seconds it takes
+	// for a cluster to transition to the desired power state. It is cleared once the desired state is reached
+	metricClusterDeploymentPowerStateTransitionSeconds = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "hive_cluster_deployment_powerstate_transition_seconds",
+			Help: "Length of time for a cluster to transition to the desired power state",
+		},
+		[]string{"cluster_deployment", "desired_powerstate"},
+	)
 	// metricControllerReconcileTime tracks the length of time our reconcile loops take. controller-runtime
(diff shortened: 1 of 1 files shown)


2uasimojo (commented):
This looks straightforward enough.

Can we add UT for it?

pkg/controller/metrics/metrics.go
```diff
@@ -225,6 +235,27 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+					} else {
+						cleared := metricClusterDeploymentPowerStateTransitionSeconds.Delete(map[string]string{
+							"cluster_deployment": cd.Name,
+							"desired_powerstate": string(cd.Spec.PowerState),
+						})
+						if cleared {
+							mcLog.Infof("cleared metric: %v", metricClusterDeploymentPowerStateTransitionSeconds)
+						}
+					}
+				}
```
dgoodwin, line 258: I don't think a per cluster labelled metric is a good idea here, that should be a last resort as we've learned with the provisioning delay issues and the cardinality we hit. 

There are a few other ways that would be less intensive on prometheus.

(1) Use the pre-existing cluster condition counts with and an alert to watch for a timeout reason, and separate tooling to tell you which clusters are in that state for SRE.
(2) A new gauge metric for stuck resuming/hibernating clusters beyond a threshold, separate tooling to find you which clusters those are.
(2) A whole system histogram for reporting how long each successful transition took, allowing us to see the average across all operations. This of course would not have entries for stuck transitioning items that never make it, which is a big flaw.

I think the per cluster thing is an anti-pattern we should avoid if possible, if grafana keeps at least 2 weeks of results, this can be a whole lot of time series in a 1k cluster with a lot of hibernating resuming going on.

2uasimojo: > per cluster labelled

D'oh! Sorry, I totally missed that :(

suhanime:
Closing because we would like to move away from per cluster metrics and alerts.
Will have to wait for [1527](https://github.com/openshift/hive/pull/1527) to go in first.

PR #1511, opened 2021-08-23
Clean up App-SRE OLMisms

2uasimojo:
We've now moved all App-SRE environments to direct deploy (see HIVE-1590) so we can get rid of the scripts and other artifacts that were used to produce those bundles etc.

[HIVE-1589](https://issues.redhat.com/browse/HIVE-1589)


joelddiaz (commented):
:+1: i guess we wait for a second lgtm on CC

2uasimojo:
>  i guess we wait for a second lgtm on CC

wdym?

joelddiaz:
I thought for any non-trivial change, we were hoping to get a :+1: from the each of the assigned and the CC individuals. But if this is considered trivial enough, I'm fine giving it the lgtm. This whole area of publishing Hive is way out of my day-to-day knowledge of this repo :). I don't know if those scripts have other uses outside of this one usecase ¯\_ (ツ)_/¯

2uasimojo:
Oh, if there were such nuances to /assign and /cc, I wasn't aware of them, sorry. I assigned people I thought would be qualified to review, and cc'd folks I thought would be interested. (I would have just assigned Devan, but he's out, and I sorta wanted to have this whole thing wrapped up in a bow by the time he got back.)

Anyway, if you don't feel good about reviewing this, don't worry about it.

dgoodwin:
It's like Christmas around here.

/lgtm

Thank you @2uasimojo !

PR #1512, opened 2021-08-24
Slowly delete stale pool clusters

2uasimojo:
3dc3c3c7c / #1484 added setting and discovery of a version marker for ClusterPool and ClusterDeployment so we could tell whether they match. This commit goes a step further and uses that information to replace stale CDs such that the pool will eventually be consistent.

The algorithm we use is as follows:
If the pool is in steady state -- i.e. we're at capacity and all unclaimed CDs are finished installing -- delete *one* stale CD. This triggers another reconcile wherein the deleted CD is replaced. We're no longer in steady state until (at least) that CD finishes installing, whereupon we're eligible to repeat.

Reasoning:
- We don't want to immediately trash all stale CDs, as this could result in significant downtime following a pool edit.
- We don't want to exceed capacity (size or maxSize) to do this rotation, so we go "down" instead of "up".
- It would have been nice to just wait for the *replacement* CDs to finish installing; but that would have entailed tracking those somehow as being distinct from CDs added through other code paths. That could get tricky, especially across multiple pool edits.

TODO:
- As written, we'll even wait for *stale* CDs to finish installing before we start deleting. We should figure out a way to discount those when checking whether we're eligible for a deletion.
- Consider exposing a knob allowing the consumer to tune how aggressively stale CDs are replaced.

[HIVE-1058](https://issues.redhat.com/browse/HIVE-1058)

### DIFF
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -392,7 +392,7 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 			logger.WithError(err).Error("error deleting cluster deployment")
 			return reconcile.Result{}, err
 		}
-
+		metricStaleClusterDeploymentsDeleted.WithLabelValues(clp.Namespace, clp.Name).Inc()
 	}
 
 	// One more (possible) status update: wait until the end to detect whether all unassigned CDs
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -285,7 +285,9 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 		return reconcile.Result{}, nil
 	}
 
-	cds, err := getAllClusterDeploymentsForPool(r.Client, clp, logger)
+	poolVersion := calculatePoolVersion(clp)
+
+	cds, err := getAllClusterDeploymentsForPool(r.Client, clp, poolVersion, logger)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -355,8 +357,6 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 	}
 	availableCurrent -= toDel
 
-	poolVersion := calculatePoolVersion(clp)
-
 	switch drift := reserveSize - int(clp.Spec.Size); {
 	// activity quota exceeded, so no action
 	case availableCurrent <= 0:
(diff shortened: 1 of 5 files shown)


2uasimojo:
/hold

Needs UT & live test

/cc @stevekuznetsov @gurnben

dgoodwin:
LGTM so far.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -383,6 +383,16 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 			log.WithError(err).Error("error adding clusters")
 			return reconcile.Result{}, err
 		}
+	// Special case for stale CDs: allow deleting one if all CDs are installed.
+	case drift == 0 && len(cds.Installing()) == 0 && len(cds.Stale()) > 0:
```
dgoodwin, line 387: Actually does this need a check to make sure none are deleting? Otherwise it kinda looks like it might immediately wipe them all out.

2uasimojo: It shouldn't need that:
- On this reconcile we `Delete` one.
- The next reconcile will miss this `case` because we'll be one short of `pool.size` so `drift` will be `< 0`. It'll add a CD  via L382 which will go to `Installing`.
- The next reconcile (and subsequent ones for the next 40m) will see nonzero `Installing` clusters, so will skip this `case`.

Did I logic that right?

dgoodwin: You are probably right, there are cases where the cache isn't updated for new objects and may not see the new Installing cluster, which is where we have to use expectations. If I'm correct that *might* happen here, it's possible it could wipe out a couple at once before the installing start showing up. I'm not super confident in any of this so we can see how it goes.

dgoodwin: Could we do a counter metric here so we can see the rate at which clusters are getting replaced due to pool changes.

dgoodwin: Counters are since the process started, prometheus handles the past data and adds the two together. In this case it's not so much for the value of the counter, more so for the rate at which it's occurring.

A gauge does sound useful for the total number of stale at any given time, rather than the rate at which we're doing replacements.

pkg/controller/metrics/metrics.go
```diff
@@ -109,6 +109,14 @@ var (
 		},
 		[]string{"cluster_deployment", "namespace", "cluster_type"},
 	)
+
+	// MetricClusterPoolStaleClusterDeploymentsDeleted tracks the total number of CDs we delete
+	// because they became "stale". That is, the ClusterPool was modified in a substantive way such
+	// that these CDs no longer match its spec.
```
dgoodwin, line 115:
```suggestion
	// that these CDs no longer match its spec. Note that there may be additional stale CDs that will only show up in this counter when they can be safely replaced.
```

pkg/controller/metrics/metrics.go
```diff
@@ -109,6 +109,14 @@ var (
...
+	MetricClusterPoolStaleClusterDeploymentsDeleted = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "hive_clusterpool_stale_clusterdeployments_deleted",
+		Help: "The number of ClusterDeployments deleted because they no longer match the spec of their ClusterPool.",
+	}, []string{"clusterpool_namespace", "clusterpool_name"})
```
dgoodwin, line 119: As far as I know, it's ok and probably best to define and register these where they're used. This routine handles a lot of the gauges for which we can't assume a controller is reconciling reliably, so we run this block every 2m, but other controllers have metrics and they're defined in that package. 

If you're going to do a gauge, I think you might want to do it in this module though, as your reconcile loop is only for a specific pool, whereas we need to see the number across all pools.

2uasimojo: Got it.

I'll leave the gauge out for now. YAGNI. We can add it later if there's a call for it. (I'm kind of guessing this counter won't be used much either.)

2uasimojo:
UT is in. Still need to test live.

2uasimojo:
Tested live, works.

/hold cancel
/assign @dgoodwin

PR #1513, opened 2021-08-25
add install log regexp for invalid AWS tags

gregsheremeta:
add install log regexp for invalid AWS tags

/assign @dgoodwin 
/cc @abutcher

/cc @vkareh - might be good to precheck this in rosa cli too


2uasimojo:
These are being moved to app-interface I thought.

dgoodwin:
> These are being moved to app-interface I thought.

It's a bit complicated, app-interface was meant to be a temporary way for us to roll these out for urgent issues, without waiting for a full hive release, typically weekly. We still want to ship them in hive for any users to have access too. 

The app-interface path was originally asked to be a "let us modify a configmap live in the clusters", but became app-interface. I don't think it's ever been used but it may still prove helpful someday. It is however meant to be a short term thing, anything there should make it's way to hive's codebase.

2uasimojo:
Okay... @abutcher does this gel with your understanding?

PR #1514, opened 2021-08-25
fix GCP CPU quota install log regexp

gregsheremeta:
fix GCP CPU quota install log regexp. GCP CPU quota install log regexp
was too specific, and the tail end of the message in GCP changed.

/assign @dgoodwin 
/cc @abutcher


dgoodwin:
Just needs a make update.

PR #1515, opened 2021-08-25
Remove encoded error message in VPCEndpointServiceReconcileFailed condition.

abutcher:


pkg/controller/utils/errorscrub.go
```diff
@@ -8,8 +8,9 @@ var (
 	newlineTabRE           = regexp.MustCompile(`\n\t`)
 	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
 	// aws
-	awsRequestIDRE   = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
-	awsNotAuthorized = regexp.MustCompile(`(User: arn:aws:sts::)\S+(:assumed-role/[^/]+/)\S+( is not authorized to perform: \S+ on resource: arn:aws:iam::)[^:]+(:\S+)`)
+	awsRequestIDRE    = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
+	awsNotAuthorized  = regexp.MustCompile(`(User: arn:aws:sts::)\S+(:assumed-role/[^/]+/)\S+( is not authorized to perform: \S+ on resource: arn:aws:iam::)[^:]+(:\S+)`)
+	awsEncodedMessage = regexp.MustCompile(`(Encoded authorization failure message: )\S+(,)`)
```
2uasimojo, line 13: Is `\S+` really the right thing? Anybody know what kind of encoding this is? Perhaps if we know it doesn't have commas, we can just use `[^,]+`.

Also, parens around the comma are unnecessary, since we're not using the capture group.

PR #1520, opened 2021-08-26
e2e: Add retry for managed DNS

2uasimojo:
We've been intermittently seeing this flake in e2e:

`time="2021-08-26T18:12:59Z" level=fatal msg="error updating HiveConfig" error="Operation cannot be fulfilled on hiveconfigs.hive.openshift.io \"hive\": the object has been modified; please apply your changes to the latest version and try again"`

Add a dumb retry around enabling managed DNS.

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -133,8 +133,14 @@ if $USE_MANAGED_DNS; then
 	# This is to prevent name conflicts across customer clusters.
 	CLUSTER_SHARD=$(cat /dev/urandom | tr -dc 'a-z' | fold -w 8 | head -n 1)
 	CLUSTER_DOMAIN="${CLUSTER_SHARD}.${BASE_DOMAIN}"
-	go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" adm manage-dns enable ${BASE_DOMAIN} \
-		--creds-file="${CREDS_FILE}" --cloud="${CLOUD}"
+	# We've seen this failing 409 (concurrent updates) recently. So try a couple times.
+	# TODO: Check output for "the object has been modified" and only retry if it matches.
+	echo "enabling managed DNS"
+	for i in 1 2 3; do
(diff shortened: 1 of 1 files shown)


hack/e2e-common.sh
```diff
@@ -133,8 +133,14 @@ if $USE_MANAGED_DNS; then
...
 	CLUSTER_DOMAIN="${CLUSTER_SHARD}.${BASE_DOMAIN}"
-	go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" adm manage-dns enable ${BASE_DOMAIN} \
-		--creds-file="${CREDS_FILE}" --cloud="${CLOUD}"
+	# We've seen this failing 409 (concurrent updates) recently. So try a couple times.
+	# TODO: Check output for "the object has been modified" and only retry if it matches.
+	echo "enabling managed DNS"
+	for i in 1 2 3; do
+	  go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" adm manage-dns enable ${BASE_DOMAIN} \
+	    --creds-file="${CREDS_FILE}" --cloud="${CLOUD}" && break
+	  echo "Retrying..."
```
joelddiaz, line 142: should we put a small sleep here?

2uasimojo: Shouldn't be necessary. The reason the 409 happens is because some update comes in between the GET and the POST within the execution of this single command.

I guess a sleep would help if we think _multiple_ updates are happening and need a few seconds to settle. WDYT?

joelddiaz: It's probably safe to just do the 3 retries. If we're competing with anything it would be an in-cluster controller, and hopefully we don't have controllers that are doing multiple writes during a single reconciliation. I'm fine either way. Let me know and i'll lgtm it

2uasimojo: Yeah, let's just go with it. I saw this error a couple of times on a couple of PRs over a couple of days, but haven't seen it since, so it's clearly not a high frequency thing in any case.

PR #1522, opened 2021-08-31
Add polarity check for Provision Underway metric

suhanime:
This commit adds a polarity check of the given cluster deployment
condition before deciding if it is responsible for provision failure or
delay. This fixes the bug where the collector incorrectly interprets the
error status of RequirementsMet condition

/assign @dgoodwin


2uasimojo:
Legit.

/lgtm

PR #1525, opened 2021-09-02
hive-operator: support exporting metrics to cluster

abhinavdahiya:
Adds a configuration to HiveConfig so that the operator can export
metrics for the controllers to prometheus instance running on the
cluster.

The operator deploys,
- the service monitors in target namespace for controllers,
- rbac resources so the in-cluster prometheus can access the monitoring
  resources and endpoints/services in the target namespace

An alternative was to use OLM supported method [1] to expose the
metrics. Hive uses CSV deployment method and the prometheus manifests
are only supported using operator bundle method. So using the OLM method
is not usefule.

[1]: https://github.com/openshift/enhancements/blob/master/enhancements/olm/olm-managed-operator-metrics.md

xref: https://issues.redhat.com/browse/HIVE-1581

/assign @joelddiaz 
/cc @stevekuznetsov


docs/monitoring.md
```diff
@@ -0,0 +1,25 @@
+# Use OpenShift Cluster Monitoring to expose Hive metrics
+
+OpenShift ships with monitoring stack that can be used to export various
+Hive metrics.
+
+## Enabling monitoring for user-defined projects
+
+Follow the guidelines as defined in [OpenShift documentation][user-projects-monitoring]
```
abhinavdahiya, line 8: By enabling monitoring for user-defined projects in OpenShift, we can export metrics even for `hive` namespace. So this makes the previously discussed requirement of `openshift-` prefix no longer required.

config/monitoring/role_binding.yaml
```diff
@@ -0,0 +1,15 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
+- kind: ServiceAccount
+  name: prometheus-user-workload
+  namespace: openshift-user-workload-monitoring
```
joelddiaz, line 15: What is the implication of always setting up this RoleBinding to **both** the usual `openshift-monitoring` ServiceAccount and the user-workload ServiceAccount?

config/monitoring/hive_clustersync_servicemonitor.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: hive-clustersync
+spec:
+  endpoints:
+  - interval: 30s
+    path: /metrics
+    port: metrics
+    scheme: http
```
joelddiaz, line 10: I've received some grief in the past for exporting metrics over http on CCO. If we're sure it is acceptable to export metrics over http then disregard this.

apis/hive/v1/hiveconfig_types.go
```diff
@@ -141,6 +141,13 @@ type HiveConfigSpec struct {
 	ArgoCD ArgoCDConfig `json:"argoCDConfig,omitempty"`
 
 	FeatureGates *FeatureGateSelection `json:"featureGates,omitempty"`
+
+	// ExtractMetrics specifies whether the operator should enable metrics for hive controllers
+	// to be extracted for prometheus.
+	// When set to true, the operator deploys ServiceMonitors so that the prometheus instances that
+	// extract metrics. The operator also sets up RBAC in the TargetNamespace so that openshift
+	// prometheus in the cluster can list/access objects required to pull metrics.
+	ExtractMetrics bool `json:"extractMetrics,omitempty"`
```
joelddiaz, line 150: I see a disconnect between the name `ExtractMetrics` here and the `exportMetrics` in the docs/monitoring.md file. FWIW, my vote is for `ExportMetrics`.

pkg/operator/hive/hive.go
```diff
@@ -513,3 +514,47 @@ func (r *ReconcileHiveConfig) deleteAllSyncSetInstances(hLog log.FieldLogger) (n
...
+		// add the label
+		if ns.Labels == nil {
+			ns.Labels = map[string]string{}
+		}
+		ns.Labels["openshift.io/cluster-monitoring"] = "true"
+		if err := r.Client.Update(context.TODO(), ns); err != nil {
+			hLog.WithError(err).Error("error updating hive namespace label to enable monitoring")
+			return err
+		}
+	}
```
joelddiaz, line 542: Is this strictly necessary when running with monitoring for a namespace not prefixed with `openshift-`? Or is this just harmless in that case?

config/monitoring/role_binding.yaml
abhinavdahiya: Always adding the role binding would mean that if the user created a service monitor, like SD does on it's own, the openshift in built monitoring will start capturing metrics which the user might not really want.
so making sure we add the service monitors and rbac when exportMetrics is true keeps things more predictable.

config/monitoring/hive_clustersync_servicemonitor.yaml
abhinavdahiya: This setup is as-is by service delivery currently, so it seems like it has been acceptable.

running metrics on https would need more work in the codebase itself, and we can add that if a requirement comes in.

pkg/operator/hive/hive.go
abhinavdahiya: definitely required for `openshift-`, will have to test if user-workload monitoring requires it. seems harmless to cover both cases by keeping it.

joelddiaz:
this looks good to me once the conflicts are resolved and this is mergeable.
@stevekuznetsov you want to give this a once-over?

abhinavdahiya:
Rebased on master for fixing merge conflicts. cc @joelddiaz

joelddiaz (commented):
/lgtm

pkg/operator/hive/hive.go
```diff
@@ -517,3 +518,47 @@ func (r *ReconcileHiveConfig) deleteAllSyncSetInstances(hLog log.FieldLogger) (n
...
 	return
 }
+
+func (r *ReconcileHiveConfig) deployMonitoring(hLog log.FieldLogger, h resource.Helper, instance *hivev1.HiveConfig) error {
+	hiveNSName := getHiveNamespace(instance)
+
+	if !instance.Spec.ExportMetrics {
+		hLog.Info("metrics were not requested to be extracted, so skipping")
+		return nil
+	}
```
2uasimojo, line 528: Here's where we need to _undo_ these changes if previously applied, per HIVE-1655.

PR #1526, opened 2021-09-03
HIVE-1637: Update installer vendor

abutcher:
/cc @abhinavdahiya


pkg/controller/remotemachineset/ovirt_test.go
```diff
@@ -39,7 +39,7 @@ func TestOvirtActuator(t *testing.T) {
 			clusterDeployment: testOvirtClusterDeployment(),
 			pool:              testOvirtPool(),
 			expectedMachineSetReplicas: map[string]int64{
-				fmt.Sprintf("%s-worker-0", testInfraID): 3,
+				fmt.Sprintf("%s-worker", testInfraID): 3,
```
abutcher, line 42: This is worrisome.

abhinavdahiya: just so i understand , previously they had -0 suffix even when there was only on machineset for a machinepoolm, but they have replaced the name to no suffix. and our worry is that it will affect our existing clusters as they will now have a new machineset and abandon the last one?

abutcher: Correct. My worry is that if the naming changes we will replace machinesets for existing ovirt clusters.

2uasimojo:
/retest

Most recent failures look like busted resume of CI pool clusters.

abutcher:
Trying out a rebase

pkg/controller/remotemachineset/ovirt_test.go
abutcher: Added a function to preserve the suffix.

2uasimojo:
(https://github.com/openshift/hive/pull/1534 should address the most recent flake)

PR #1528, opened 2021-09-03
ClusterPool RunningCount

2uasimojo:
Adds ClusterPool.Spec.RunningCount, the number of pool clusters to try to keep active at any time.

Note that when ClusterPool.Spec.HibernateAfter is set, we count from when it was claimed rather than when it was started so it doesn't hibernate early.

[HIVE-1576](https://issues.redhat.com/browse/HIVE-1576)

### DIFF
--- a/apis/hive/v1/clusterpool_types.go
+++ b/apis/hive/v1/clusterpool_types.go
@@ -21,6 +21,12 @@ type ClusterPoolSpec struct {
 	// +required
 	Size int32 `json:"size"`
 
+	// RunningCount is the number of clusters we should keep running. The remainder will be kept hibernated until claimed.
+	// By default no clusters will be kept running (all will be hibernated).
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	RunningCount int32 `json:"runningCount,omitempty"`
+
 	// MaxSize is the maximum number of clusters that will be provisioned including clusters that have been claimed
 	// and ones waiting to be used.
 	// By default there is no limit.
--- a/pkg/controller/clusterpool/clusterpool_controller_test.go
+++ b/pkg/controller/clusterpool/clusterpool_controller_test.go
@@ -1251,6 +1251,127 @@ func TestReconcileClusterPool(t *testing.T) {
 			expectedObservedSize:   2,
 			expectedTotalClusters:  3,
 		},
+		{
+			name: "runningCount < size",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(
+					testcp.WithSize(4),
+					testcp.WithRunningCount(2),
+				),
+			},
+			expectedTotalClusters: 4,
+			expectedRunning:       2,
+		},
+		{
+			name: "runningCount == size",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(
+					testcp.WithSize(4),
+					testcp.WithRunningCount(4),
+				),
+			},
+			expectedTotalClusters: 4,
+			expectedRunning:       4,
+		},
+		{
+			name: "runningCount > size",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(
+					testcp.WithSize(4),
+					testcp.WithRunningCount(400),
+				),
+			},
+			expectedTotalClusters: 4,
+			expectedRunning:       4,
+		},
+		{
+			name: "runningCount restored after deletion",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(
+					testcp.WithSize(4),
+					testcp.WithRunningCount(2),
+				),
+				unclaimedCDBuilder("c1").Build(testcd.Generic(testgeneric.Deleted()), testcd.WithPowerState(hivev1.RunningClusterPowerState)),
+				unclaimedCDBuilder("c2").Build(testcd.WithPowerState(hivev1.RunningClusterPowerState)),
+				unclaimedCDBuilder("c3").Build(),
+				unclaimedCDBuilder("c4").Build(),
+			},
+			expectedObservedSize:  3,
+			expectedTotalClusters: 5,
+			expectedRunning:       3,
+		},
+		{
+			name: "runningCount restored after claim",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(
+					testcp.WithSize(4),
+					testcp.WithRunningCount(2),
+				),
+				unclaimedCDBuilder("c1").Build(testcd.WithPowerState(hivev1.RunningClusterPowerState), testcd.Installed()),
+				unclaimedCDBuilder("c2").Build(testcd.WithPowerState(hivev1.RunningClusterPowerState), testcd.Installed()),
+				unclaimedCDBuilder("c3").Build(testcd.Installed()),
+				unclaimedCDBuilder("c4").Build(testcd.Installed()),
+				testclaim.FullBuilder(testNamespace, "test-claim", scheme).Build(testclaim.WithPool(testLeasePoolName)),
+			},
+			expectedObservedSize:   4,
+			expectedObservedReady:  4,
+			expectedTotalClusters:  5,
+			expectedRunning:        3,
+			expectedAssignedCDs:    1,
+			expectedAssignedClaims: 1,
+		},
+		{
+			name: "pool size > number of claims > runningCount",
+			existing: []runtime.Object{
+				initializedPoolBuilder.Build(
+					testcp.WithSize(4),
+					testcp.WithRunningCount(2),
+				),
+				unclaimedCDBuilder("c1").Build(testcd.WithPowerState(hivev1.RunningClusterPowerState), testcd.Installed()),
+				unclaimedCDBuilder("c2").Build(testcd.WithPowerState(hivev1.RunningClusterPowerState), testcd.Installed()),
+				unclaimedCDBuilder("c3").Build(testcd.Installed()),
+				unclaimedCDBuilder("c4").Build(testcd.Installed()),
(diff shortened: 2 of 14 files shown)


2uasimojo:
WIP: Test

2uasimojo:
Rebase picks up #1534 which should reduce e2e-pool flakes.
Still likely to run into #1535 though.

pkg/test/clusterdeployment/clusterdeployment.go
```diff
@@ -136,6 +136,8 @@ func WithClusterPoolReference(namespace, poolName, claimName string) Option {
 			PoolName:  poolName,
 			ClaimName: claimName,
 		}
+		now := metav1.Now()
+		clusterDeployment.Status.ClaimedTimestamp = &now
```
2uasimojo, line 140: NTS: This was the fun test bug where the extra (Status) Update() to [sync this](https://github.com/openshift/hive/pull/1528/files#diff-bbf4d3cb24b8cac87951b84ca973747902c00d4a9d662648c8e760a37569581dR625) would cause the fake client to actually delete a CD with its DeletionTimestamp set.

Having it set on a claimed CD is The Right Thing™ in any case -- we just happened to have a test case that made sure we did it, in a roundabout way.

2uasimojo:
Tested live. Works gooood. Still pending UT.

2uasimojo:
UT done

/hold cancel

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -224,29 +224,58 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
...
+			hibLog = hibLog.WithField("claimedSince", claimedSince)
+		} else {
+			// This means it's not a pool cluster (!isUnclaimedPoolCluster && ClaimedTimestamp == nil)
+			hibLog.Debug("cluster does not belong to a clusterpool")
+		}
+
 		cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
 		if cond.Status == corev1.ConditionUnknown {
-			hibLog.Debug("cluster has never been hibernated, using installed time")
+			hibLog.Debug("cluster has never been hibernated")
```
joelddiaz, line 260: Two lines above, it seems like `cond` can potentially be nil. Although I understand that we set conditions to default values, can we defensively code this to avoid inadvertent panics? Or is there some mechanism that is guaranteeing that this simply cannot happen?

apis/hive/v1/clusterpool_types.go
```diff
@@ -21,6 +21,12 @@ type ClusterPoolSpec struct {
 	// +required
 	Size int32 `json:"size"`
 
+	// RunningCount is the number of clusters we should keep running. The remainder will be kept hibernated until claimed.
+	// By default no clusters will be kept running (all will be hibernated).
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	RunningCount int32 `json:"runningCount"`
```
abhinavdahiya, line 28: can we have the json tag to include omitempty since this is optional field.

apis/hive/v1/clusterpool_types.go
```diff
@@ -21,6 +21,12 @@ type ClusterPoolSpec struct {
 	// +required
 	Size int32 `json:"size"`
 
+	// RunningCount is the number of clusters we should keep running. The remainder will be kept hibernated until claimed.
+	// By default no clusters will be kept running (all will be hibernated).
```
abhinavdahiya, line 25: nit: can we include what the default value is in the api doc, or do you think most users will understand that 0 is the default value for interger

pkg/controller/clusterpool/collections.go
```diff
@@ -395,6 +417,13 @@ func (cds *cdCollection) Stale() []*hivev1.ClusterDeployment {
 	return append(cds.unknownPoolVersion, cds.mismatchedPoolVersion...)
 }
 
+// RegisterNewCluster adds a freshly-created cluster to the cdCollection, assuming it is installing.
+func (cds *cdCollection) RegisterNewCluster(cd *hivev1.ClusterDeployment) {
+	cds.byCDName[cd.Name] = cd
+	cds.installing = append(cds.installing, cd)
+	cds.sortInstalling()
```
abhinavdahiya, line 424: is it better to just insert the cd to the right location instead of appending and then sorting. i'm assuming he installing was already sorted right?

joelddiaz:
One overarching thought I've had around this PR is adding the new field to the ClusterDeployment for tracking the claimed timestamp. I understand its purpose, but it does feel a bit like leaking details about ClusterPools down into the ClusterDeployment (which we've already done as I look through other part of the ClusterDeployment.Spec).

I suppose my question is: does everyone feel okay with this (I don't know what to call it: separation of concerns or leaky abstractions or something else)?

apis/hive/v1/clusterpool_types.go
2uasimojo: Between the statement on L25 and the intuitiveness of "runningCount==0 means no running clusters" I think it's sufficient. MaxSize and MaxConcurrent also mention the default behavior, not the default value (though admittedly in those cases the default value is less intuitive).

apis/hive/v1/clusterpool_types.go
2uasimojo: Good catch, done.

(Though since this is a primitive type with a nil value, I suspect `omitempty` is redundant.)

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -1251,6 +1251,127 @@ func TestReconcileClusterPool(t *testing.T) {
...
+				testclaim.FullBuilder(testNamespace, "test-claim5", scheme).Build(testclaim.WithPool(testLeasePoolName)),
+			},
+			expectedObservedSize:  4,
+			expectedObservedReady: 4,
+			expectedTotalClusters: 9,
+			// The four original pool CDs got claimed. Two were already running; we started the other two.
+			// We create five new CDs to satisfy the pool size plus the additional claim.
+			// Of those, we start two for runningCount, and hibernate the rest.
+			// (Intuitively, perhaps the one for the excess claim should start off in running state.
+			// But that's not how the logic works today. Should it?)
```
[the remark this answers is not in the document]
2uasimojo: Yeah.

We had to do some [nontrivial work](https://github.com/openshift/hive/pull/1474) for these kinds of scenarios (claims in excess of pool size) to prevent the "overdrawn" clusters from getting deleted to get us back to capacity.

This would behave the same way: those clusters would get `PowerState=Hibernating` until they finished installing. Then -- and this is important -- _the oldest clusters_ (which are not necessarily the ones just installed) would get assigned and set `PowerState=Running`. In fact, those aren't necessarily synchronous steps: if one or more older clusters were already installed/installing, they would be activated and assigned before the surplus ones finished.

It would be nice if, in all cases (regardless of `runningCount`) we could set an extra {number of surplus claims} clusters to running. I'm just worried about the complexity and getting the math right.

So I think I'd like to look into this in a followon. If that's okay with you, I'll open a card.

pkg/controller/clusterpool/collections.go
2uasimojo: It was, yes.

The right place would be the front of the list, since we're sorting these by reverse age.

However, I had to figure that out by looking at the `sortInstalling` function. I'm preferring to sacrifice a little bit of performance for the benefit of keeping the ordering logic in a single spot. Like, if we decided to change the ordering later, we could easily change `sortInstalling` and forget to change this function, leading to some tough-to-find bugs.

(I suppose I could get rid of `sortInstalling` and instead make an `addInstalling` that's used from both here and [L299](https://github.com/openshift/hive/pull/1528/files#diff-bbf4d3cb24b8cac87951b84ca973747902c00d4a9d662648c8e760a37569581dR299). But then I'm writing my own insertion sort, which is likely to be way less efficient than the builtin.)

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: Legit. Fixed.

2uasimojo:
> One overarching thought I've had around this PR is adding the new field to the ClusterDeployment for tracking the claimed timestamp. I understand its purpose, but it does feel a bit like leaking details about ClusterPools down into the ClusterDeployment (which we've already done as I look through other part of the ClusterDeployment.Spec).

Would it be more comfortable if it was `ClusterDeployment.Spec.ClusterPoolRef.ClaimedTimestamp` instead? There are tradeoffs either way.
- The timestamp seems like more of a status-y thing than a spec-y thing.
- It feels a lot like `InstalledTimestamp`.
- It's used by the hibernation controller right alongside :arrow_up: 

...but it's set a the same time as `ClusterPoolRef.ClaimName`; and putting it in there would save us having to do a separate status update.

I'm happy either way, just lmk.

2uasimojo:
/hold

Team met today and we decided to do this:

> Would it be more comfortable if it was ClusterDeployment.Spec.ClusterPoolRef.ClaimedTimestamp instead?

2uasimojo:
4acce17 rebase only. ClaimedTimestamp changes still forthcoming.

2uasimojo:
> Would it be more comfortable if it was ClusterDeployment.Spec.ClusterPoolRef.ClaimedTimestamp instead?

This is done.

/hold cancel

pkg/controller/clusterpool/clusterpool_controller_test.go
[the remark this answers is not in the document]
2uasimojo: https://issues.redhat.com/browse/HIVE-1651

PR #1530, opened 2021-09-07
hivev2: ClusterTemplate

2uasimojo:
Add a section proposing a ClusterTemplate CR that would contain the meat of ClusterDeployment configuration and be used by both ClusterDeployment and ClusterPool.

### DIFF
--- a/docs/enhancements/hivev2.md
+++ b/docs/enhancements/hivev2.md
@@ -15,6 +15,10 @@
     - [ClusterProvision](#clusterprovision)
     - [ClusterRelocate](#clusterrelocate)
     - [ClusterState](#clusterstate)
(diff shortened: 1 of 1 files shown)


gregsheremeta:
+1 to the idea -- I love it.

It reminds me of VM templates in RHV. RHV has both flavours and templates, but in practice everyone used only the templates. https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/virtual_machine_management_guide/creating_a_template_from_an_existing_virtual_machine

abhinavdahiya:
one thing I would like to consider is if the template function needs to be provided by the hive apis. 
there are tools like kustomize that can better solve the needs for users by creating a better experience of DRY and composition that i don't think is very much possible with a live k8s object.

PR #1535, opened 2021-09-10
e2e: Force DNSZone resync on ActuatorNotInitialized flake

2uasimojo:
We've seen frequent e2e flakes where the ClusterDeployment times out waiting to install, and it contains a status condition like:

```
   - lastProbeTime: "2021-09-10T00:24:40Z"
      lastTransitionTime: "2021-09-10T00:24:40Z"
      message: 'error instantiating actuator: Secret "hiveci-613aa513-5fd-aws-creds" not found'
      reason: ActuatorNotInitialized
      status: "True"
      type: DNSNotReady
```

This is a timing problem whereby hiveutil create-cluster is creating the secret and the ClusterDeployment in quick succession, and the dnszone controller kicks in before the secret has come to life (or appeared in the cache?). If the dnszone controller doesn't find the secret, it goes to sleep for 2h, during which time the CD doesn't progress; and we time out of the test after ~90m.

Since this is not a situation that could reasonably happen in real life, we can't justify changing the logic of the clusterpool or dnszone controllers. So with this commit we hack the test case: While waiting for the ClusterDeployment to install, if we notice this condition, we reset the `lastSyncGeneration` of the DNSZone object, which forces the dnszone controller to resync it. Hopefully by this time the secret exists and things will proceed as normal.

While we're in here, we do a couple of other small things:
- Send failure and timeout messages to stderr, which (I think?) should make them show up in the CI output in eye-catching red, and not be collapsed.
- On timeout, print the ClusterDeployment's status conditions so we don't have to go hunting through the artifacts directory for them.
- Save dnszones, machinepools, and (a summary of) secrets into the artifacts dir to aid debugging.

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -156,6 +156,10 @@ function capture_manifests() {
     oc get clusterimageset -o yaml &> "${ARTIFACT_DIR}/hive_clusterimagesets.yaml" || true
     oc get clusterprovision -A -o yaml &> "${ARTIFACT_DIR}/hive_clusterprovision.yaml" || true
     oc get clusterstate -A -o yaml &> "${ARTIFACT_DIR}/hive_clusterstate.yaml" || true
+    oc get dnszone -A -o yaml &> "${ARTIFACT_DIR}/hive_dnszones.yaml" || true
+    oc get machinepool -A -o yaml &> "${ARTIFACT_DIR}/hive_machinepools.yaml" || true
+    # Don't get the contents of the secrets, since they're sensitive; hopefully just listing them will be helpful.
+    oc get secrets -A &> "${ARTIFACT_DIR}/secret_list.txt" || true
 }
 
 function capture_cluster_logs() {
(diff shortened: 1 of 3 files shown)


2uasimojo:
/hold

I want to repeat e2e on this at least until we hit the condition in question and prove the hack is working.

/assign @abhinavdahiya

2uasimojo:
Well, at least we hit the code path :P

2uasimojo:
/test e2e

Of course, the one time we _want_ to hit the flake.

2uasimojo:
/test e2e

wrong flake this time

2uasimojo:
/test e2e

wonder if the revendor tweaked the timing so we’re not hitting this window anymore.

2uasimojo:
```
Waiting for the ClusterDeployment hiveci-613cf76c-5ff to install
Found DNSNotReady=>ActuatorNotInitialized condition. Forcing DNSZone to resync...
Retrieving object...
Noninteractive mode. Applying jq...
Starting to serve on /tmp/statuspatch.19Wk.sock
Success
Cleaning up /tmp/statuspatch.19Wk
Still waiting for the ClusterDeployment hiveci-613cf76c-5ff to install. Status check #1/90... 
```

this is working

/hold cancel

2uasimojo:
/test e2e

terraform

PR #1536, opened 2021-09-10
Better namespace in clusterpool doc examples

2uasimojo:
The examples in clusterpools.md created the pool and other artifacts in the `hive` namespace, which wouldn't usually be a good practice. Change it to something generic and obviously example-ish.

[HIVE-1646](https://issues.redhat.com/browse/HIVE-1646)


joelddiaz (commented):
:+1:

PR #1538, opened 2021-09-14
e2e-pool: Move AllClustersCurrent check into wait_for_pool_to_be_ready

2uasimojo:
We've started seeing flakes where the pool's ready/size statuses are copacetic before the conditions settle. This commit moves the check for AllClustersCurrent into the same loop that checks the ready/size statuses so we get a chance to retry in this case.

### DIFF
--- a/hack/e2e-pool-test.sh
+++ b/hack/e2e-pool-test.sh
@@ -67,7 +67,7 @@ function wait_for_pool_to_be_ready() {
   local poolname=$1
   local i=0
   # NOTE: This will need to change if we add a test with a zero-size pool.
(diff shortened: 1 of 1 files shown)


2uasimojo:
/hold

This is clearly not working, though I'll be darned if I can figure out why.

2uasimojo:
> /test unit

This flake addressed via #1541

2uasimojo:
@suhanime silly logic error, fixed in latest update. Please re-review. Thanks!

2uasimojo:
I thought I already

/hold cancel

ed this.

PR #1540, opened 2021-09-20
controller/dnszone: add condition to track dns api opt-in required failures

abhinavdahiya:
Adds a special purpose condition to DNSZone like the
AuthenticationFailure condition to provide details when the controller
requires DNS apis to be enabled for success.

Currently, only the AWS actuator sets this condition based on request
from HIVE-1469.

The clusterdeployment controller tracks the conditions from DNSZone, and
it now also tracks the new condition to update the DNSReady condition on
the clusterdeployment.

/assign @joelddiaz 
/cc @suhanime

### DIFF
--- a/pkg/controller/dnszone/awsactuator.go
+++ b/pkg/controller/dnszone/awsactuator.go
@@ -572,6 +572,39 @@ func (a *AWSActuator) setInsufficientCredentialsConditionToTrue(message string) 
 	return accessDeniedCondsChanged
 }
 
+func (a *AWSActuator) setAPIOptInRequiredConditionToFalse() bool {
+	apiOptInConds, apiOptInCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.APIOptInRequiredCondition,
+		corev1.ConditionFalse,
+		apiOptInNotRequiredReason,
+		"route53 apis are enabled",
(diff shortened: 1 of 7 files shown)


2uasimojo:
/retest

The e2e-pool flake is addressed by #1538. Feel free to merge that.

pkg/controller/dnszone/awsactuator.go
```diff
@@ -572,6 +572,39 @@ func (a *AWSActuator) setInsufficientCredentialsConditionToTrue(message string)
...
+	return apiOptInCondsChanged
+}
+
+func (a *AWSActuator) setAPIOptInRequiredConditionToTrue(message string) bool {
+	apiOptInConds, apiOptInCondsChanged := controllerutils.SetDNSZoneConditionWithChangeCheck(
+		a.dnsZone.Status.Conditions,
+		hivev1.APIOptInRequiredCondition,
+		corev1.ConditionTrue,
+		apiOptInRequiredReason,
+		message,
```
joelddiaz, line 597: Do we not need to worry about sanitizing the message to remove the UUID-type pieces from the string?

joelddiaz: I was more concerned about the hot-loops, but as you pointed out, this controller has been converted to rate limit on updates.

joelddiaz (commented):
/lgtm

PR #1541, opened 2021-09-20
Work around clustersync float compare test flake

2uasimojo:
We frequently see flakes like:

```
--- FAIL: TestReconcileClusterSync_NewSyncSetApplied (0.01s)
    assertion_compare.go:323:
        	Error Trace:	clustersync_controller_test.go:161
        	            				clustersync_controller_test.go:840
        	Error:      	"3599.994122811" is not greater than or equal to "3599.998926313"
        	Test:       	TestReconcileClusterSync_NewSyncSetApplied
        	Messages:   	[requeue after too small]
```

This isn't because the controller is doing something wrong; it's because
adding and substracting floating point numbers is subject to small
errors due to internal representation.

We don't really care about these requeue intervals beyond about 0.1s
granularity; so this commit truncates them to 2 decimal places to avert
this flake.

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller_test.go
+++ b/pkg/controller/clustersync/clustersync_controller_test.go
@@ -3,6 +3,7 @@ package clustersync
 import (
 	"context"
 	"fmt"
+	"math"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


joelddiaz (commented):
/lgtm

PR #1544, opened 2021-09-21
validate username/password set for vsphere installconfig

abhinavdahiya:
since username and password must be set in install-config.yaml for vSphere platform in addition
to the credentialsSecretRef in clusterdeployment. We add a quick check to make sure some value
is set to give users pre-flight style check before running.
similar to https://github.com/openshift/hive/pull/1403

xref: https://issues.redhat.com/browse/OCPBUGSM-33627

/assign @abutcher 
/cc @suhanime

### DIFF
--- a/pkg/controller/clusterdeployment/installconfigvalidation.go
+++ b/pkg/controller/clusterdeployment/installconfigvalidation.go
@@ -10,10 +10,13 @@ import (
 )
 
 const (
-	noAWSPlatformErr   = "install config did not contain an AWS platform"
-	noGCPPlatformErr   = "install config did not contain a GCP platform"
-	noAzurePlatformErr = "install config did not contain an Azure platform"
-	regionMismatchErr  = "install config region does not match cluster deployment region"
+	noAWSPlatformErr     = "install config did not contain an AWS platform"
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/installconfigvalidation.go
```diff
@@ -45,6 +48,13 @@ func ValidateInstallConfig(cd *hivev1.ClusterDeployment, installConfig []byte) e
 		if ic.Platform.Azure.Region != cd.Spec.Platform.Azure.Region {
 			return errors.New(regionMismatchErr)
 		}
+	case platform.VSphere != nil:
+		if ic.Platform.VSphere == nil {
+			return errors.New(novSpherePlatformErr)
+		}
+		if ic.Platform.VSphere.Username == "" || ic.Platform.VSphere.Password == "" {
```
2uasimojo, line 55: Are we sure an empty password is never legitimate? (Not IRL of course; I'm thinking for test environments.)

abhinavdahiya: openshift-installer will reject an installconfig that has empty credentials. so this check should help improve the problem highlighted in OCPBUGSM-33627

PR #1545, opened 2021-09-22
detect install failures when the installer can't delete an IAM role

yithian:


config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -82,6 +82,11 @@ data:
       - "platform\\.aws\\.userTags.*: Invalid value:.*value contains invalid characters"
       installFailingReason: InvalidAWSTags
       installFailingMessage: You have specified an invalid AWS tag value. Verify that your tags meet AWS requirements and try again.
+    - name: ErrorDeletingIAMRole
+      searchRegexStrings:
+        - "Error deleting IAM Role \\([a-zA-Z0-9-]*\\): DeleteConflict: Cannot delete entity, must detach all policies first."
```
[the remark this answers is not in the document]
yithian: yeah it's the role id

now that you mention it, i just assumed that ids would be alphanumericand hyphen, i should see if aws has any specs on that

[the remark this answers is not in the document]
2uasimojo: Per https://docs.aws.amazon.com/IAM/latest/APIReference/API_Role.html:
If it's the role *name*, then what you've changed it to is good.
If it's the role *ID*, `\w+` is good.
As noted on the last PR, if we don't want to try to figure out (and keep up with if it changes!) the actual format, we can just use something more like `\\([^)]+\\)` ("open paren, one or more not-a-close-paren, close paren").

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -82,6 +82,11 @@ data:
       - "platform\\.aws\\.userTags.*: Invalid value:.*value contains invalid characters"
       installFailingReason: InvalidAWSTags
       installFailingMessage: You have specified an invalid AWS tag value. Verify that your tags meet AWS requirements and try again.
+    - name: ErrorDeletingIAMRole
+      searchRegexStrings:
+        - "Error deleting IAM Role \\([a-zA-Z0-9+=,.@-]+\\): DeleteConflict: Cannot delete entity, must detach all policies first."
```
gregsheremeta, line 87:
```suggestion
        - "Error deleting IAM Role .* DeleteConflict: Cannot delete entity, must detach all policies first."
```

gregsheremeta: when we make these too expressive, they become brittle / prone to breaking in subtle ways. Don't be afraid of `.*`

pkg/controller/clusterprovision/installlogmonitor_test.go
```diff
@@ -304,6 +305,26 @@ func TestParseInstallLog(t *testing.T) {
...
+			existing: []runtime.Object{&corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      regexConfigMapName,
+					Namespace: constants.DefaultHiveNamespace,
+				},
+				Data: map[string]string{
+					"regexes": `
+    - name: ErrorDeletingIAMRole
+      searchRegexStrings:
+        - "Error deleting IAM Role \\([a-zA-Z0-9+=,.@-]+\\): DeleteConflict: Cannot delete entity, must detach all policies first."
```
gregsheremeta, line 320:
```suggestion
        - "Error deleting IAM Role .* DeleteConflict: Cannot delete entity, must detach all policies first."
```

pkg/operator/assets/bindata.go
```diff
@@ -1640,6 +1640,11 @@ data:
       - "platform\\.aws\\.userTags.*: Invalid value:.*value contains invalid characters"
       installFailingReason: InvalidAWSTags
       installFailingMessage: You have specified an invalid AWS tag value. Verify that your tags meet AWS requirements and try again.
+    - name: ErrorDeletingIAMRole
+      searchRegexStrings:
+        - "Error deleting IAM Role \\([a-zA-Z0-9+=,.@-]+\\): DeleteConflict: Cannot delete entity, must detach all policies first."
```
gregsheremeta, line 1645:
```suggestion
        - "Error deleting IAM Role .* DeleteConflict: Cannot delete entity, must detach all policies first."
```

PR #1539, opened 2021-09-16
add install log failure regexes

yithian:
for when:
* a subnet that doesn't exist was specified
* aws credentials with insufficient permissions were provided

PR #1548, opened 2021-09-23
Read default regexes from configmap for tests.

abutcher:
@yithian @gregsheremeta 
/hold

### DIFF
--- a/pkg/controller/clusterprovision/installlogmonitor_test.go
+++ b/pkg/controller/clusterprovision/installlogmonitor_test.go
@@ -1,9 +1,11 @@
 package clusterprovision
 
 import (
+	"io/ioutil"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterprovision/installlogmonitor_test.go
```diff
@@ -324,62 +298,16 @@ func TestParseInstallLog(t *testing.T) {
...
-  searchRegexStrings:
-  - "Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane"
-  installFailingReason: KubeAPIWaitFailed
-  installFailingMessage: Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane
-`,
-		},
+// buildRegexConfigMap reads the install log regexes configmap from within config/configmaps/install-log-regexes-configmap.yaml
+func buildRegexConfigMap() runtime.Object {
+	decode := serializer.NewCodecFactory(scheme.Scheme).UniversalDeserializer().Decode
+	stream, err := ioutil.ReadFile("../../../config/configmaps/install-log-regexes-configmap.yaml")
```
2uasimojo, line 304: Let's bindata this puppy.

2uasimojo: Oh, look, it [already is](https://github.com/openshift/hive/blob/c0e9a5bfbb68a790101c2f56e1fc40372f4bf102/pkg/operator/assets/bindata.go#L1559).

2uasimojo: To be clear, I'm asking you to use the bindata asset instead of loading up the file "by hand" here.

abhinavdahiya: bindata means we need to make update every time and it's not always clear why my changes come up. since go test runs from the source code directly, it's imo better to use the file directly.

2uasimojo: Our `verify` CI test already enforces that bindata is in sync.

This is what bindata is for.

abutcher: I could see the workflow of adding a new regex/test being a little confusing if you have to make bindata before running this test specifically.

2uasimojo: Fair enough.

pkg/controller/clusterprovision/installlogmonitor_test.go
```diff
@@ -265,43 +267,15 @@ func TestParseInstallLog(t *testing.T) {
...
-  searchRegexStrings:
-  - "compute\\.googleapis\\.com/cpus is not available in [a-z0-9-]* because the required number of resources \\([0-9]*\\) is more than remaining quota"
-  installFailingReason: GCPComputeQuotaExceeded
-  installFailingMessage: GCP CPUs quota exceeded
-`,
-				},
-			}},
+			name:           "GCP compute quota",
+			log:            pointer.StringPtr(gcpCPUQuotaLog),
+			existing:       []runtime.Object{buildRegexConfigMap()},
```
2uasimojo, line 272: This is latent, so could be handled in a separate PR, but is there any reason to build this on every test case? It's read-only, right?

2uasimojo (commented):
There are several other test cases with inlined configmaps. If those regexes aren't part of the official config, do we still need to test them?

pkg/controller/clusterprovision/installlogmonitor_test.go
abutcher: There are some test cases that want a missing config or a malformed regex so that's why I have it per test case atm.

abutcher: Could split them in a separate PR

abutcher:
If this goes before https://github.com/openshift/hive/pull/1539 @yithian will have to rebase again so
/hold

pkg/controller/clusterprovision/installlogmonitor_test.go
2uasimojo: > some test cases that want a missing config or a malformed regex

So there are two different issues here.
1. This comment was about all test cases where `existing` is simply using `buildRegexConfigMap()`. All of those test cases should be able to reuse a common variable built once at the top of the function. That's the bit I'm saying was latent and could be handled in a separate PR.
2. Some of the test cases don't use `buildRegexConfigMap()` but instead use inlined configmaps. Are all of those specifically testing edge cases with missing/bad configmaps?

2uasimojo (commented):
In any case, this is a good start. We can address further cleanup in subsequent PRs.

/lgtm

PR #1549, opened 2021-09-24
RFE: ClusterPool Inventory (for VSphere DNS)

2uasimojo:
Add an enhancement doc describing a new `ClusterPool.Spec.Inventory`
field. The immediate goal is adding VSphere support to ClusterPools by
solving the DNS problem (see referenced jira); but we try to leave the
door open to future expansion for other cloud providers such as
baremetal.

[HIVE-1367](https://issues.redhat.com/browse/HIVE-1367)

### DIFF
--- a/docs/enhancements/cluster-pool-inventory.md
+++ b/docs/enhancements/cluster-pool-inventory.md
@@ -1,0 +1,173 @@
+# ClusterPool Inventory
+
+- [ClusterPool Inventory](#clusterpool-inventory)
+  - [User Stories](#user-stories)
+  - [Problem Statement](#problem-statement)
+  - [Scope](#scope)
+  - [Proposal](#proposal)
+    - [Summary](#summary)
+    - [`ClusterPool.Spec.Inventory`](#clusterpoolspecinventory)
+    - [How To Use](#how-to-use)
+    - [Validation](#validation)
+    - [`Size` and `MaxSize`](#size-and-maxsize)
+    - [Pool Version](#pool-version)
+    - [Handling Inventory Updates](#handling-inventory-updates)
+      - [Adding An Inventory](#adding-an-inventory)
+      - [Adding An Entry](#adding-an-entry)
+      - [Removing An Entry](#removing-an-entry)
+      - [Deleting The Inventory](#deleting-the-inventory)
+    - [Racing For Names](#racing-for-names)
+    - [Fairness](#fairness)
+
+## User Stories
+As a cluster administrator, I want to be able to use ClusterPools to provision VSphere clusters.
+
+(Future) As a cluster administrator, I want to be able to use ClusterPools to provision bare metal clusters.
+
+## Problem Statement
+Provisioning a cluster requires configuration unique to that cluster.
+For example, the cluster's name is used to build the hostnames for the API and console.
+_Someone_ has to handle the DNS that resolves those hostnames to the IPs of the services created by the installer.
+One option is to create your ClusterDeployment with `ManageDNS=True` and we'll create a DNSZone on the fly.
+
+But that doesn't work for some cloud providers, such as VSphere, the focus of this document.
+
+In the case of VSphere, one solution is to configure DNS manually and then create your ClusterDeployment with the right `Name` and `BaseDomain` so that the assembled hostnames match those entries. This is fine if you're creating ClusterDeployments by hand, but breaks down if ClusterDeployment names are being generated with random slugs, as is the case with ClusterPools.
+
+## Scope
+This feature is limited to solving DNS for VSphere.
+Bare metal will come later, _but_ we will try to be forward-thinking in this design to allow for future expansion to cover that user story.
+
+## Proposal
+### Summary
+Allow ClusterPool to accept an inventory of per-cluster details to be used when generating ClusterDeployments for the pool.
+For the current design, only the cluster (short) name is supported.
+
+### `ClusterPool.Spec.Inventory`
+Add a field to `ClusterDeployment.Spec` called `Inventory`.
+It is a dict.
+For now, the dict can have one member, `ClusterDeployments`, which is a list of dicts.
+For now, each dict in the list can have one member, `Name`, which is a string.
+Example:
+
+```yaml
+spec:
+  inventory:
+    clusterDeployments:
+    - name: foo
+    - name: bar
+    - name: baz
+```
+
+(This may seem overengineered, but the idea is to leave room for future expansion.
+For example, perhaps we'll need to specify IP addresses that are tied to each name.
+If we designed the inventory as simply a list of name strings, there would be nowhere to put those IPs.)
+
+When adding a ClusterDeployment, if such an `Inventory` is present, ClusterPool will:
+- Load up the inventory list.
+- Load up the existing pool CDs (it already does this).
+- Subtract the set of names of existing CDs from those in the inventory list.
+- Pick a name from the set that remains.
+- Use the name for the new CD:
+  - As the `.spec.clusterName`.
+  - As the seed for the `.metadata.namespace` and `.metadata.name`, to which a random slug will be appended as usual.
+
+Absent an `Inventory`, ClusterPool will continue to use the generated namespace/name (`{pool name}-{random slug}`) as the `.spec.clusterName` as it does today.
+
+(**Note:** We will not restrict the use of `Inventory` to VSphere.
+I don't know why you would want to use it for other cloud providers, but you can if you like.
+In fact, that's 100% how I plan to test this.)
+
+### How To Use
+For the VSphere case, this allows the administrator to:
+- Preconfigure DNS.
+- Provide the list of (short) hostnames to `ClusterPool.Spec.Inventory`.
+
+For example, if DNS is configured with:
+```
+10.0.0.10  api.foo.example.com
+10.0.0.11  apps.foo.example.com
+10.0.0.12  console.foo.example.com
+
+10.0.1.10  api.bar.example.com
+10.0.1.11  apps.bar.example.com
+10.0.1.12  console.bar.example.com
+```
+the ClusterPool could be configured with:
+```yaml
+...
+spec:
+  baseDomain: example.com
+  inventory:
+    clusterDeployments:
+    - name: foo
+    - name: bar
+  ...
+```
+
+### Validation
+Webhook validation will ensure that, if `Inventory.ClusterDeployments` is specified
+- Each entry contains the `Name` key.
+- The list of `Name` values contains no empties or duplicates.
+
+This validation can be loosened in the future if the need arises without violating the API contract.
+Like maybe for baremetal, we need to supply a preconfigured hardware inventory for each cluster, but it's okay to generate the names.
+
+### `Size` and `MaxSize`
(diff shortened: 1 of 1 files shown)


docs/enhancements/cluster-pool-inventory.md
```diff
@@ -0,0 +1,158 @@
...
+
+(This may seem overengineered, but the idea is to leave room for future expansion.
+For example, perhaps we'll need to specify IP addresses that are tied to each name.
+If we designed the inventory as simply a list of name strings, there would be nowhere to put those IPs.)
+
+When adding a ClusterDeployment, if such an `Inventory` is present, ClusterPool will:
+- Load up the inventory list.
+- Load up the existing pool CDs (it already does this).
+- Subtract the set of names of existing CDs from those in the inventory list.
+- Pick a name from the set that remains and use it for the new CD.
```
dgoodwin, line 68: This code will be in a controller that watches ClusterPool, giving you some guarantee that we're not concurrently doing this calculation for more than one request and might select the same name twice?

Probably will want to watch in the calculation for two hits for the same name and do... something... I don't know what. :)

docs/enhancements/cluster-pool-inventory.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Deleting The Inventory
+This will change the [pool version](#pool-version), rendering existing unclaimed clusters stale and causing the controller to replace them gradually.
+Absent the inventory, the replacements will have generated names as "normal" (before this feature).
+The administrator may wish to speed up this process by manually deleting CDs, or scaling the pool size to zero and back.
+
+### Challenge: Namespace Collision
+Today, ClusterPool uses the same Name and Namespace for a given ClusterDeployment.
+There is code that relies on these being the same.
+If we continue this practice, a user-specified inventory has the potential to contain a name that would conflict with an existing namespace.
```
dgoodwin, line 146: I think you have some flexibility in terms of what the clusterdeployment is named, and the actual cluster name fields in the clusterdeployment. (spec.clusterName, spec.baseName) Could you work out a way to take the given name and append something random when naming your cd and namespace, even though the cluster name will be consistent and what the user requests in their inventory?

dgoodwin (commented):
This is pretty cool, tucking inventory in there is consistent with the concept of a ClusterPool.

docs/enhancements/cluster-pool-inventory.md
joelddiaz: This kind of implementation kind of guarantees that we can only run one replica of hive-controllers, right?

dgoodwin: Leader election ensures that only one of those pods is actively processing at a time even if you're using multiple replicas. They're more like hot backups which is sort of why we weren't super worried about the multiple replicas.

That said Thomas' clustersync thing we did for multiple pods each acting on only a subset if the hash matches is another option we might use in future, but even there it should be close to impossible for two pods to reconcile one object in the same controller. We don't have full guarantees here as we're not a transactional db, but it should be pretty reliable day to day, you just might want to keep an eye out for it going wrong and react or warn somehow.

joelddiaz (commented):
IIRC the vSphere provisioning process, the generated install-config needs to contain the platform.vsphere values of apiVIP and ingressVIP. How is this information being populated? I see that the installer has this as optional fields in https://github.com/openshift/installer/blob/master/pkg/types/vsphere/platform.go , but our docs don't seem to cover them as optional. Whether that requirement is specific to the environment DPTP has set up for us or not, it doesn't seem unreasonable that someone else's vSphere config might also require these per-cluster static pieces of information.

More generally, I have concerns that the inventory structure will by necessity be platform-specific across vsphere, openstack, and metal (and really all non-auto-dns-platforms). You can see that the field names (and the quantity of fields) can differ https://github.com/openshift/installer/blob/master/pkg/types/openstack/platform.go

What I think we need is the opportunity for some more knowledgeable piece of code (knowledgeable about the specifics of the underlying infrastructure) to make changes to an install-config before the clusterdeployment controller tries to launch a provision job. I have some not-fully-thought-through ideas, but I'd be curious what other people dream up.

Or have I missed some bigger picture here, and these things are simply not to be worried about or solved in some other way?

abhinavdahiya:
> IIRC the vSphere provisioning process, the generated install-config needs to contain the platform.vsphere values of apiVIP and ingressVIP. How is this information being populated? I see that the installer has this as optional fields in https://github.com/openshift/installer/blob/master/pkg/types/vsphere/platform.go , but our docs don't seem to cover them as optional. Whether that requirement is specific to the environment DPTP has set up for us or not, it doesn't seem unreasonable that someone else's vSphere config might also require these per-cluster static pieces of information.
> 
> More generally, I have concerns that the inventory structure will by necessity be platform-specific across vsphere, openstack, and metal (and really all non-auto-dns-platforms). You can see that the field names (and the quantity of fields) can differ https://github.com/openshift/installer/blob/master/pkg/types/openstack/platform.go
> 
> What I think we need is the opportunity for some more knowledgeable piece of code (knowledgeable about the specifics of the underlying infrastructure) to make changes to an install-config before the clusterdeployment controller tries to launch a provision job. I have some not-fully-thought-through ideas, but I'd be curious what other people dream up.
> 
> Or have I missed some bigger picture here, and these things are simply not to be worried about or solved in some other way?

One idea i had was to use partially filled install-config.yaml as definition of one item in inventory. and the inventory is a collection of install-config yamls. the controller can read the install-config, update the fields it controls and use that create the cluster...

docs/enhancements/cluster-pool-inventory.md
2uasimojo: Worst case we could explicitly keep track -- e.g. ClusterPool.Status.ClusterNamesInUse, and bouncing the status update with a 409 is our concurrency protection -- but I'd like to avoid that if possible. If we can solve the namespace collision thing in a way that lets us keep the names deterministic, that would also work.

Still thinking this through...

docs/enhancements/cluster-pool-inventory.md
2uasimojo: Yes!

The simplest thing, without mucking with [Builder.Name](https://github.com/openshift/hive/blob/8671ed2067e47e5ed0b6e718e7163a6741ed7b39/pkg/clusterresource/builder.go#L27-L30), would be to keep [using Builder as is](https://github.com/openshift/hive/blob/8671ed2067e47e5ed0b6e718e7163a6741ed7b39/pkg/controller/clusterpool/clusterpool_controller.go#L665-L679) and then change the cd.spec.clusterName at the same time we [set the pool reference](https://github.com/openshift/hive/blob/8671ed2067e47e5ed0b6e718e7163a6741ed7b39/pkg/controller/clusterpool/clusterpool_controller.go#L700-L701).

This solves the namespace collision problem nicely... but doesn't help with the cluster name concurrency problem discussed [above](https://github.com/openshift/hive/pull/1549#discussion_r716593307).

docs/enhancements/cluster-pool-inventory.md
2uasimojo: After a week of noodling, I think I'd like to address this with an annotation containing a JSON dict, keyed by the cluster name from the inventory list, of the namespace of the CD consuming it.

- When the value is empty/null/nil, that name hasn't been claimed yet.
- When we need to create a new cluster, we
  - Scan the list for an entry with a nil value
  - Generate a unique namespace name using that name as a stub
  - Post an update to the CD with the annotation updated with that namespace as the value for that entry. Here's where we bounce if two controllers try to grab the same name at the same time: the loser will bounce the update with a 409; we'll requeue and start over.
  - Proceed with existing algorithm to create new cluster (except setting cluster name explicitly as noted [below](https://github.com/openshift/hive/pull/1549#discussion_r719721440)).
- Add a top-level routine to ensure all replete entries have CDs associated with them. If not, attempt to create them through the same flow. This is to guard against leaking entries if a controller crashes between updating the annotation and creating the CD.

**Question:** Sorting. Is it important that we rotate through the list of supplied cluster names in any particular order, or with any nod to "fairness"? For example: If there are five names and the usage pattern happens to only use two at a time, is it okay to reuse the same two over and over, or must we strive to consume all of them evenly?

docs/enhancements/cluster-pool-inventory.md
```diff
@@ -0,0 +1,158 @@
...
+Like maybe for baremetal, we need to supply a preconfigured hardware inventory for each cluster, but it's okay to generate the names.
+
+### `Size` and `MaxSize`
+If `Inventory` is used, `ClusterPool.Spec.Size` and `.MaxSize` are implicitly constrained to the length of the list of nonempty `ClusterDeployments[].Names`.
+Setting either/both to a smaller number still works as you would expect.
+
+### Pool Version
+To make [adding](#adding-an-inventory) and [deleting](#deleting-the-inventory) `Inventory` work sanely, we will adjust the computation of the pool version used for [stale CD detection/replacement](https://issues.redhat.com/browse/HIVE-1058) as follows:
+- When `Inventory` is present, append an arbitrary fixed string before the [final hash operation](https://github.com/openshift/hive/blob/0b9229b91e6f8a3c2bf095efbdf017226e69d026/pkg/controller/clusterpool/clusterpool_controller.go#L479).
+  (We don't want to recalculate the hash any time the inventory changes, as this would treat *all* existing CDs as stale and replace them.)
```
akhil-rane, line 118: here you mean all existing `unclaimed` CDs right? if yes, what is the disadvantage of marking them stale and creating new CDs?

docs/enhancements/cluster-pool-inventory.md
```diff
@@ -0,0 +1,158 @@
...
+
+Options:
+- Validate each namespace.
+  We can't do this at ClusterPool config time (like with a webhook) because that doesn't prevent conflicting namespaces being created after the fact.
+  We would have to do it on the fly when we peeled off a name for a ClusterDeployment.
+  But then what would we do if we found a conflict?
+  The best we could do is log the failure and move on to the next name in the list.
+  That's not great UX.
+  (Note that this uniqueness check is still done with generated namespaces -- but in that case we'll keep trying.
+  With a predefined inventory, such a conflict would render that namespace unusable, potentially reducing the capacity of the cluster.)
```
akhil-rane, line 156: If I understand correctly

```suggestion
  With a predefined inventory, such a conflict would render that namespace unusable, potentially reducing the capacity of the clusterpool.)
```

2uasimojo: Good catch, thanks.

docs/enhancements/cluster-pool-inventory.md
2uasimojo: Per [slack discussion](https://coreos.slack.com/archives/CE3ETN3J8/p1634157465296300), moved to a status field.

2uasimojo:
@joelddiaz and Abhinav: regarding using install-config.yaml as a template, or allowing a hook to modify it per-CD: I like this idea, but it may be too heavy to be practical for consumers. I'd say we should avoid doing that until we have no alternative -- and in this case we have a simpler alternative.

2uasimojo:
Closing in favor of #1607.

PR #1550, opened 2021-09-26
HIVE-1604: Amalgamate build and operatorhub push scripts.

abutcher:
```
➜  hive (HIVE-1604) ✔ ./hack/bundle-gen.py --help                    
usage: bundle-gen.py [-h] [--verbose] [--dry-run] [--branch BRANCH] [--registry-auth-file REGISTRY_AUTH_FILE] [--github-user GITHUB_USER] [--hold]

Hive Bundle Generator and Publishing Script

optional arguments:
  -h, --help            show this help message and exit
  --verbose             Show more details while running
  --dry-run             Test run that skips pushing branches and submitting PRs
  --branch BRANCH       The branch (or commit-ish) from which to build and push the image. If unspecified, we assume `master`. If we're using `master`, we generate the bundle version
                        number based on ${HIVE_VERSION_PREFIX} and update the `alpha` channel. If BRANCH *also* corresponds to an `ocm-X.Y`, we'll also update that channel with the same
                        bundle.If BRANCH corresponds to an `ocm-X.Y` but *not* `master`, we'll generate the bundle version number based on `X.Y` and update only that channel.
  --registry-auth-file REGISTRY_AUTH_FILE
                        Path to registry auth file
  --github-user GITHUB_USER
                        User's github username, if different than $USER
  --hold                Adds a /hold comment in commit body to prevent the PR from merging (use "/hold cancel" to remove)
```

Default, no arg case updates alpha channel using the master branch only.
```
./hack/bundle-gen.py --dry-run
```

### DIFF
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -3,7 +3,7 @@
 import argparse
 import datetime
 import git
-import importlib
+import github as gh
 import json
 import os
 import shutil
@@ -90,16 +90,22 @@ def build_and_push_image(repo_dir, registry_auth_file, image_tag, branch, dry_ru
         print('Failed to checkout branch {}'.format(branch))
         raise
 
-    # build/push the thing
     container_name = '{}:{}'.format(OPERATORHUB_HIVE_IMAGE_DEFAULT, image_tag)
-    print('Building container {}'.format(container_name))
 
     if dry_run:
-        print('Skipping build due to dry-run')
+        print('Skipping build of container {} due to dry-run'.format(container_name))
         return
 
-    cmd = 'buildah bud --tag {} -f ./Dockerfile'.format(container_name).split()
-    subprocess.run(cmd)
+    # Did we already build it locally?
+    cp = subprocess.run('buildah images -nq {}'.format(container_name).split(), stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    if cp.returncode == 0:
+        print('Container {} already exists locally; not rebuilding.'.format(container_name))
+    else:
+        # build/push the thing
+        print('Building container {}'.format(container_name))
+
+        cmd = 'buildah bud --tag {} -f ./Dockerfile'.format(container_name).split()
+        subprocess.run(cmd)
 
     print("Pushing container")
     cmd = 'buildah push '
@@ -334,9 +340,9 @@ def open_pr(work_dir, fork_repo, upstream_repo, gh_username, bundle_source_dir, 
     print("Adding file")
     repo.git.add(HIVE_SUB_DIR)
 
-    print("Commiting {}".format(pr_title))
+    print("Committing {}".format(pr_title))
     try:
-        repo.git.commit('--signoff', '--message="{}"'.format(pr_title))
+        repo.git.commit('--signoff', '--message={}'.format(pr_title))
     except:
         print("Failed to commit")
         raise
@@ -344,15 +350,14 @@ def open_pr(work_dir, fork_repo, upstream_repo, gh_username, bundle_source_dir, 
 
     if not dry_run:
         print("Pushing branch {}".format(branch_name))
-        origin = repo.git.origin
+        origin = repo.remotes.origin
         try:
-            origin.push('--force', branch_name)
+            origin.push(branch_name, None, force=True)
         except:
             print("failed to push branch to origin")
             raise
 
         # open PR
-        gh = importlib.import_module('github')
         client = gh.GitHubClient(dest_github_org, dest_github_reponame, "")
 
         from_branch = "{}:{}".format(gh_username, branch_name)
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -1,0 +1,447 @@
+#!/usr/bin/env python
+
+import argparse
+import datetime
+import git
+import importlib
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import urllib3
+import yaml
+
+# HIVE_VERSION_PREFIX is the prefix of the Hive version that will be constructed by this script
+# the version used for images and bundles will be "{prefix}.{number of commits}-{git hash}" eg. "v1.2.3187-18827f6"
+HIVE_VERSION_PREFIX = "v1.2"
+
+HIVE_REPO = "git@github.com:openshift/hive.git"
+
+# Hive dir within both:
+# https://github.com/redhat-openshift-ecosystem/community-operators-prod
+# https://github.com/k8s-operatorhub/community-operators
+HIVE_SUB_DIR = 'operators/hive-operator'
+
+OPERATORHUB_HIVE_IMAGE_DEFAULT = 'quay.io/openshift-hive/hive'
+
+REGISTRY_AUTH_FILE_DEFAULT = '{}/.docker/config.json'.format(os.environ['HOME'])
+
+COMMUNITY_OPERATORS_HIVE_PKG_URL = 'https://raw.githubusercontent.com/redhat-openshift-ecosystem/community-operators-prod/main/operators/hive-operator/hive.package.yaml'
+
+CHANNEL_DEFAULT = 'alpha'
+BRANCH_DEFAULT = 'master'
+
+SUBPROCESS_REDIRECT = subprocess.DEVNULL
+
+
+def get_params():
+    parser = argparse.ArgumentParser(description='Hive Bundle Generator and Publishing Script')
+    parser.add_argument('--verbose',
+                        default=False,
+                        help='Show more details while running',
+                        action='store_true')
+    parser.add_argument('--dry-run',
+                        default=False,
+                        help='Test run that skips pushing branches and submitting PRs',
+                        action='store_true')
+    parser.add_argument('--branch',
+                        default=False,
+                        help='''The branch (or commit-ish) from which to build and push the image.
+                                If unspecified, we assume `master`. If we're using `master`, we
+                                generate the bundle version number based on the hive version prefix `{}` and
+                                update the `alpha` channel. If BRANCH *also* corresponds to an `ocm-X.Y`,
+                                we'll also update that channel with the same bundle.If BRANCH corresponds
+                                to an `ocm-X.Y` but *not* `master`, we'll generate the bundle version
+                                number based on `X.Y` and update only that channel.'''.format(HIVE_VERSION_PREFIX))
+    parser.add_argument('--registry-auth-file',
+                        default=REGISTRY_AUTH_FILE_DEFAULT,
+                        help='Path to registry auth file')
+    parser.add_argument('--github-user',
+                        default=os.environ['USER'],
+                        help="User's github username, if different than $USER")
+    parser.add_argument('--hold',
+                        default=False,
+                        help='Adds a /hold comment in commit body to prevent the PR from merging (use "/hold cancel" to remove)',
+                        action='store_true')
+    args = parser.parse_args()
+
+    if args.branch and args.branch == BRANCH_DEFAULT:
+        parser.error('--branch=master is assumed by this script and corresponds to the alpha channel')
+
+    if not os.path.isfile(args.registry_auth_file):
+        parser.error('--registry-auth-file ({}) does not exist, provide --registry-auth-file'.format(args.registry_auth_file))
+
+    if args.verbose:
+        global SUBPROCESS_REDIRECT
+        SUBPROCESS_REDIRECT = None
+
+    return args
+
+# build_and_push_image uses buildah to build the HIVE image (tagged with image_tag) and then pushes the image to quay
+def build_and_push_image(repo_dir, registry_auth_file, image_tag, branch, dry_run):
+    repo = git.Repo(repo_dir)
+
+    print('Checkout out {} branch'.format(branch))
+    try:
+        repo.git.checkout(branch)
+    except:
+        print('Failed to checkout branch {}'.format(branch))
+        raise
+
+    # build/push the thing
+    container_name = '{}:{}'.format(OPERATORHUB_HIVE_IMAGE_DEFAULT, image_tag)
+    print('Building container {}'.format(container_name))
+
+    if dry_run:
+        print('Skipping build due to dry-run')
+        return
+
+    cmd = 'buildah bud --tag {} -f ./Dockerfile'.format(container_name).split()
+    subprocess.run(cmd)
+
+    print("Pushing container")
+    cmd = 'buildah push '
+    if registry_auth_file != None:
+        cmd = cmd + ' --authfile={} '.format(registry_auth_file)
+    cmd = cmd + ' {}'.format(container_name)
+    subprocess.run(cmd.split())
+
+# gen_hive_version generates and returns the hive version eg. "v1.2.3187-18827f6"
+def gen_hive_version(repo_dir, branch, commit_hash, prefix):
+    repo = git.Repo(repo_dir)
+
+    print('Checkout out {} branch'.format(branch))
+    try:
+        repo.git.checkout(branch)
+    except:
+        print('Failed to checkout branch {}'.format(branch))
+        raise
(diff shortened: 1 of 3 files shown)


hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+    parser.add_argument('--hive-repo-dir',
+                        default=HIVE_REPO_DIR_DEFAULT,
+                        help='Path to root of repo to checkout/build images from')
+    parser.add_argument('--registry-auth-file',
+                        default=REGISTRY_AUTH_FILE_DEFAULT,
+                        help='Path to registry auth file')
+    parser.add_argument('--channel',
+                        default=[CHANNEL_DEFAULT],
+                        action='append',
+                        help='Update channel in OLM package to new version')
```
2uasimojo, line 56: I think we shouldn't need this arg. If we can't detect what channel(s) to update, something is wrong. (Unless this script is also going to support fast-forwarding OCM branches, which I think is probably too much.)

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+# build_and_push_image uses buildah to build the HIVE image (tagged with image_tag) and then pushes the image to quay
+def build_and_push_image(repo_dir, registry_auth_file, image_tag, commit_hash):
+    prev_dir = os.getcwd()
+    os.chdir(repo_dir)
+
+    try:
+        # sync up git repo with remote
+        cmd = "git fetch upstream".split()
+        subprocess.run(cmd)
+        cmd = "git fetch upstream --tags".split()
```
2uasimojo, line 97: I think this is a superset of the above; you don't need both.

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+        subprocess.run(cmd)
+
+        # checkout tag/commit
+        cmd = 'git checkout {}'.format(commit_hash).split()
+        subprocess.run(cmd)
```
2uasimojo, line 102: Don't we need to be error checking these commands?

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+        container_name = 'quay.io/openshift-hive/hive:{}'.format(image_tag)
+        cmd = 'buildah bud --tag {} -f ./Dockerfile'.format(container_name).split()
+        subprocess.run(cmd)
+
+        print("Pushing container")
+        cmd = 'buildah push '
+        if registry_auth_file != None:
+            cmd = cmd + ' --authfile={} '.format(registry_auth_file)
+        cmd = cmd + ' {}'.format(container_name)
+        subprocess.run(cmd.split())
```
2uasimojo, line 115: Subject this to `--dry-run`.

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+    # sha is the first 7 characters of commit_hash
+    sha = str(repo.rev_parse(commit_hash))[0:7]
+
+    # num commits is the number of git commits counted from the first commit to the provided commit_hash
+    # this is the equivalent of running:
+    # git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-9]{40}$"`..HEAD --count
+    parent = repo.git.rev_list('--max-parents=0', commit_hash)
+    num_commits = repo.git.rev_list('--count', '{parent}..{commit_hash}'.format(parent=parent, commit_hash=commit_hash))
+
+    return '{prefix}.{commits}-{sha}'.format(prefix=HIVE_VERSION_PREFIX, commits=num_commits, sha=sha)
```
2uasimojo, line 132: For older branches, it won't be`HIVE_VERSION_PREFIX`. That's where we'll want to do something like `branch_name.split('-', 1)[0]`

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+def get_previous_version(channel_name):
+    http = urllib3.PoolManager()
+    r = http.request('GET', COMMUNITY_OPERATORS_HIVE_PKG_URL)
+    pkg_yaml = yaml.load(r.data.decode('utf-8'), Loader=yaml.FullLoader)
+    try:
+        for channel in pkg_yaml['channels']:
+            if channel['name'] == channel_name:
+                return channel['currentCSV'].replace('hive-operator.', '')
+    except:
+        print('Unable to determine previous hive version from {}', COMMUNITY_OPERATORS_HIVE_PKG_URL)
```
2uasimojo, line 145: It would be worth printing the exception here, since a number of things could go wrong above.

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+if __name__ == "__main__":
+    args = get_params()
+    validate_params(args)
+
+    if not buildah_installed():
+        print('buildah not found, please install buildah')
+        sys.exit(1)
+
+    hive_version = gen_hive_version(args.hive_repo_dir, args.commit_hash)
+    previous_hive_version = get_previous_version(CHANNEL_DEFAULT)
```
2uasimojo, line 392: And here we'll want to be smart about which channel to use.

2uasimojo (commented):
Much of this might be obsolete given recent discussions, but if nothing else, notes to self...

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+# https://github.com/k8s-operatorhub/community-operators
+HIVE_SUB_DIR = 'operators/hive-operator'
+
+REGISTRY_AUTH_FILE_DEFAULT = '/home/{}/.docker/config.json'.format(os.environ['USER'])
+
+COMMUNITY_OPERATORS_HIVE_PKG_URL = 'https://raw.githubusercontent.com/redhat-openshift-ecosystem/community-operators-prod/main/operators/hive-operator/hive.package.yaml'
+
+CHANNEL_DEFAULT = 'alpha'
+
+CSV_DEFAULT = 'operatorhub'
```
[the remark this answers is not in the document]
abutcher: The original bundle generation script had a distinction between `operatorhub` and `osd` and this is a remnant of that. Removing.

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+
+# HIVE_VERSION_PREFIX is the prefix of the Hive version that will be constructed by this script
+# the version used for images and bundles will be "{prefix}.{number of commits}-{git hash}" eg. "v1.2.3187-18827f6"
+HIVE_VERSION_PREFIX = "v1.2"
+
+# HIVE_COMMIT_HASH_DEFAULT is the default hive commit hash referenced by this script for building
+# the hive image, generating the version, etc.
+HIVE_COMMIT_HASH_DEFAULT = 'HEAD'
+
+HIVE_REPO_DIR_DEFAULT = '/home/{}/go/src/github.com/openshift/hive'.format(os.environ['USER'])
```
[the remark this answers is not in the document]
abutcher: Thoughts on just using `os.getcwd()`? Current commit is using `os.getcwd()`.

hack/bundle-gen.py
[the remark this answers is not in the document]
abutcher: The versioning was slightly different for osd https://github.com/openshift/hive/blob/master/hack/generate-operator-bundle-operatorhub.py#L44

hack/bundle-gen.py
```diff
@@ -0,0 +1,410 @@
...
+import yaml
+
+
+# HIVE_VERSION_PREFIX is the prefix of the Hive version that will be constructed by this script
+# the version used for images and bundles will be "{prefix}.{number of commits}-{git hash}" eg. "v1.2.3187-18827f6"
+HIVE_VERSION_PREFIX = "v1.2"
+
+# HIVE_COMMIT_HASH_DEFAULT is the default hive commit hash referenced by this script for building
+# the hive image, generating the version, etc.
+HIVE_COMMIT_HASH_DEFAULT = 'HEAD'
```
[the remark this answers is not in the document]
abutcher: Everything should be branch based. I've left a few references to commit that may be useful for a future `--commit-hash` sort of arg when you don't want to build the `HEAD` of `--branch`.

hack/bundle-gen.py
```diff
@@ -0,0 +1,442 @@
...
+    # this is the equivalent of running:
+    # git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-9]{40}$"`..HEAD --count
+    parent = repo.git.rev_list('--max-parents=0', commit_hash)
+    num_commits = repo.git.rev_list('--count', '{parent}..{commit_hash}'.format(parent=parent, commit_hash=commit_hash))
+
+    return '{prefix}.{commits}-{sha}'.format(prefix=HIVE_VERSION_PREFIX, commits=num_commits, sha=sha)
+
+# bump_version_zstream bumps the zstream on provided prev_version
+# eg. bump_version_zstream('hive-operator.v1.0.11') = 'hive-operator.v1.0.12'
+def bump_version_zstream(prev_version):
```
abutcher, line 135: This is used to generate the version for a `--branch` whose commit doesn't match master. This covers branches <= ocm-2.3. When ocm-2.4 no longer tracks master for an update we would want to use `gen_hive_version` to bump the commit count.

abutcher: Need to special case that somehow where this is called.

abutcher: Committed the prefix swap separately.

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+                        default=os.environ['USER'],
+                        help="User's github username, if different than $USER")
+    parser.add_argument('--hold',
+                        default=False,
+                        help='Adds a /hold comment in commit body to prevent the PR from merging (use "/hold cancel" to remove)',
+                        action='store_true')
+    args = parser.parse_args()
+
+    if args.branch and args.branch == BRANCH_DEFAULT:
+        parser.error('--branch=master is assumed by this script and corresponds to the alpha channel')
```
2uasimojo, line 68: I don't know that we want to error if they specify the default explicitly. But if we're going to do this check, we need to be cleverer about detecting whether whatever commit-ish they specify corresponds to master.

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+    if not buildah_installed():
+        print('buildah not found, please install buildah')
+        sys.exit(1)
+
+    update_channels = [CHANNEL_DEFAULT]
+    if args.branch:
+        if branch_is_on_master_commit(args.branch, os.getcwd()):
+            update_channels.append(args.branch)
+        else:
+            update_channels = [args.branch]
```
2uasimojo, line 395: This is nice. But it won't work if I specify a commit hash. So we should probably check for that. (Not sure how.)

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+        hive_version = gen_hive_version(os.getcwd(), BRANCH_DEFAULT, 'HEAD', HIVE_VERSION_PREFIX)
+        build_and_push_image(os.getcwd(), args.registry_auth_file, hive_version, BRANCH_DEFAULT, args.dry_run)
+        previous_hive_version = get_previous_version(CHANNEL_DEFAULT)
+        generate_csv_base(bundle_dir.name, hive_version, previous_hive_version, OPERATORHUB_HIVE_IMAGE_DEFAULT, CHANNEL_DEFAULT)
+    else:
+        previous_hive_version = get_previous_version(args.branch)
+        # the version prefix is determined by the previous version when the
+        # provided branch does not correspond with the master branch
+        # eg. previous_hive_version = "hive-operator.v1.1.9-1.1.8.2" -> hive_version_prefix = "v1.1"
+        hive_version_prefix = previous_hive_version.replace('hive-operator.', '')[0:4]
```
2uasimojo, line 411: This would break down when we hit e.g. `v1.10`. Perhaps
```suggestion
        hive_version_prefix = previous_hive_version.replace('hive-operator.', '').split('.', 2)[0..1].join('.')
```
ish?

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+            update_channels.append(args.branch)
+        else:
+            update_channels = [args.branch]
+
+    hive_version = ""
+    bundle_dir = tempfile.TemporaryDirectory(prefix='hive-operator-bundle-')
+    work_dir = tempfile.TemporaryDirectory(prefix='operatorhub-push-')
+
+    if CHANNEL_DEFAULT in update_channels:
+        hive_version = gen_hive_version(os.getcwd(), BRANCH_DEFAULT, 'HEAD', HIVE_VERSION_PREFIX)
```
2uasimojo, line 402: The first and third arguments are the same for both calls to this; so can we just hardcode them in the function?

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+        build_and_push_image(os.getcwd(), args.registry_auth_file, hive_version, BRANCH_DEFAULT, args.dry_run)
+        previous_hive_version = get_previous_version(CHANNEL_DEFAULT)
+        generate_csv_base(bundle_dir.name, hive_version, previous_hive_version, OPERATORHUB_HIVE_IMAGE_DEFAULT, CHANNEL_DEFAULT)
+    else:
+        previous_hive_version = get_previous_version(args.branch)
+        # the version prefix is determined by the previous version when the
+        # provided branch does not correspond with the master branch
+        # eg. previous_hive_version = "hive-operator.v1.1.9-1.1.8.2" -> hive_version_prefix = "v1.1"
+        hive_version_prefix = previous_hive_version.replace('hive-operator.', '')[0:4]
+        hive_version = gen_hive_version(os.getcwd(), BRANCH_DEFAULT, 'HEAD', hive_version_prefix)
```
2uasimojo, line 412: Should this be
```suggestion
        hive_version = gen_hive_version(os.getcwd(), args.branch, 'HEAD', hive_version_prefix)
```
?

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+        else:
+            update_channels = [args.branch]
+
+    hive_version = ""
+    bundle_dir = tempfile.TemporaryDirectory(prefix='hive-operator-bundle-')
+    work_dir = tempfile.TemporaryDirectory(prefix='operatorhub-push-')
+
+    if CHANNEL_DEFAULT in update_channels:
+        hive_version = gen_hive_version(os.getcwd(), BRANCH_DEFAULT, 'HEAD', HIVE_VERSION_PREFIX)
+        build_and_push_image(os.getcwd(), args.registry_auth_file, hive_version, BRANCH_DEFAULT, args.dry_run)
```
2uasimojo, line 403: similarly here, can we factor out `os.getcwd()`?

hack/bundle-gen.py
```diff
@@ -0,0 +1,429 @@
...
+        previous_hive_version = get_previous_version(CHANNEL_DEFAULT)
+        generate_csv_base(bundle_dir.name, hive_version, previous_hive_version, OPERATORHUB_HIVE_IMAGE_DEFAULT, CHANNEL_DEFAULT)
```
2uasimojo, line 405: Likewise don't need `OPERATORHUB_HIVE_IMAGE_DEFAULT` here.

hack/bundle-gen.py
2uasimojo: Oh, wait. I think we want the prefix to come from the branch/channel name, not the version tag.

hack/bundle-gen.py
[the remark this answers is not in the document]
abutcher: Now cloning hive to a tmpdir from git URL.

hack/bundle-gen.py
abutcher: Everything but the `commit_hash` arg could change for now in the two calls to this.

2uasimojo:
Added a commit that gets this working (produced https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/276 and https://github.com/k8s-operatorhub/community-operators/pull/234 with it). Going to merge this as is and add moar stuffs later.

/lgtm

PR #1553, opened 2021-09-28
bundle-gen: smart branch/channel detection

2uasimojo:
Add some logic to map whatever branch you specify to the right
- commit to build from
- semver prefix
- channel(s) to update

Plus a bit of general cleanup/refactoring.

[HIVE-1604](https://issues.redhat.com/browse/HIVE-1604)


abutcher (commented):
Looks great. Tested with no arguments and with `--branch=ocm-2.3`.
/lgtm

PR #1560, opened 2021-10-01
OSD: Pull operator/controller images by digest

2uasimojo:
To accommodate ICSP, guarding against a quay outage, pull the operator images by digest for OSD, causing the operator to pull the (same) image for the controllers by digest as well.

OperatorHub is not affected.

[HIVE-1504](https://issues.redhat.com/browse/HIVE-1504)


joelddiaz:
:+1: from me. Just want to make sure this works for @jharrington22 and the rest of SRE

PR #1561, opened 2021-10-04
bundle-gen: Ensure working in hive clone when generating bundle.

abutcher:
/cc @2uasimojo

### DIFF
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -113,8 +113,8 @@ def get_params():
     return args
 
 
-# build_and_push_image uses buildah to build the HIVE image (tagged with "v{hive_version}"
-# eg. "v1.2.3187-18827f6") and then pushes the image to quay
+# build_and_push_image uses buildah to build the HIVE image from the current working directory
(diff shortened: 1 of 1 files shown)


hack/bundle-gen.py
```diff
@@ -186,7 +186,10 @@ def get_previous_version(channel_name):
         raise
 
 
-def generate_csv_base(bundle_dir, version, prev_version, channel):
+def generate_csv_base(hive_repo_dir, bundle_dir, version, prev_version, channel):
+    print("Working in {}".format(hive_repo_dir))
+    os.chdir(hive_repo_dir)
```
2uasimojo, line 191: I think this actually needs to go above the checkout on L552. (Otherwise we're landing in the default branch in the temp clone.)

abutcher: Moved!

abutcher: I squashed again to add comments about CWD.

2uasimojo (commented):
/lgtm

Thanks!

PR #1565, opened 2021-10-04
Fix Provisioned condition for adopted clusters

2uasimojo:
...or clusters that were already provisioned prior to being upgraded
to/past where the Provisioned condition was introduced.

Also address a latent
[buglet](https://github.com/openshift/hive/pull/1546#discussion_r721401994)
that wasn't actually affecting anything (it would have washed out on the
subsequent reconcile).

HIVE-1596

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -599,6 +599,18 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 				return reconcile.Result{Requeue: true}, nil
 			}
 		}
+		// Set the Provisioned status condition for adopted clusters (and in case we upgraded to/past where that condition was introduced)
+		if err := r.updateCondition(
+			cd,
+			hivev1.ProvisionedCondition,
+			corev1.ConditionTrue,
+			hivev1.ProvisionedProvisionedReason,
+			"Cluster is provisioned",
+			cdLog,
+		); err != nil {
+			cdLog.WithError(err).Error("Error updating Provisioned status condition")
+			return reconcile.Result{}, err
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -599,6 +599,18 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 				return reconcile.Result{Requeue: true}, nil
 			}
 		}
+		// Set the Provisioned status condition for adopted clusters (and in case we upgraded to/past where that condition was introduced)
```
suhanime, line 602: Shouldn't this be in the above `if` block where we are setting the installedtimestamp to ensure this only happens for already installed (and/or adopted) clusters?

2uasimojo: No; installed clusters will have their installedTimestamp set already.

suhanime: I meant setting the provisioned condition to true only if cd is already installed

2uasimojo: I'm missing something.

The new code block is inside the `if cd.Spec.Installed {` block started on L593. So it will only be triggered when `cd.Spec.Installed` is True.

It is _outside_ of the block where we check whether `InstalledTimstamp` is set because we want to trigger both setting that timestamp and setting the Provisioned condition based on `cd.Spec.Installed` being True.

suhanime: D'oh - github formatting mistake - sorry!

suhanime (commented):
/lgtm

PR #1566, opened 2021-10-05
Clear SyncSetsNotApplied Hibernating Reason

2uasimojo:
For a newly-created ClusterDeployment, the hibernation controller can
hit the window where the ClusterSync has been created by the clustersync
controller, but it has not yet succeeded. In this case the hibernation
controller sets the Hibernating condition's Reason to
SyncSetsNotApplied. However, when the ClusterSync does go green, we
weren't resetting that condition, so it stayed that way indefinitely
until the next time the cluster hibernated.

With this commit, we check for that and set the condition to a new
SyncSetsApplied state. The subsequent reconcile does the checks for
Running and sets that state if appropriate.

We also close a latent gap whereby we would silently ignore the case
where the ClusterSync hadn't yet been created. Now we requeue in that
case so we can wait for it to exist.

[HIVE-1657](https://issues.redhat.com/browse/HIVE-1657)

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -192,17 +192,18 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	shouldHibernate := cd.Spec.PowerState == hivev1.HibernatingClusterPowerState
 	hibernatingCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
 
+	clusterSync := &hiveintv1alpha1.ClusterSync{}
+	if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync); err != nil {
+		// This may be NotFound, which means the clustersync controller hasn't created the ClusterSync yet.
+		// Fail and requeue to wait for it to exist.
+		return reconcile.Result{}, fmt.Errorf("could not get ClusterSync: %v", err)
+	}
+
 	// Signal a problem if we should be hibernating or have requested hibernate after and the
 	// SyncSets have not yet been applied.
 	if shouldHibernate || cd.Spec.HibernateAfter != nil {
 		// Determine if SyncSets have been applied
-		clusterSync := &hiveintv1alpha1.ClusterSync{}
-		err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync)
-		if err != nil && !apierrors.IsNotFound(err) {
-			return reconcile.Result{}, fmt.Errorf("could not get ClusterSync: %v", err)
-		}
-		// SyncSets are present and have not been applied
-		if err == nil && clusterSync.Status.FirstSuccessTime == nil {
+		if clusterSync.Status.FirstSuccessTime == nil {
 			// Allow hibernation (do not set condition) if hibernateAfterSyncSetsNotApplied duration has passed since cluster
 			// installed and syncsets still not applied
 			if cd.Status.InstalledTimestamp != nil && time.Now().Sub(cd.Status.InstalledTimestamp.Time) < hibernateAfterSyncSetsNotApplied {
@@ -222,6 +223,12 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		if supported, msg := r.hibernationSupported(cd); supported {
 			return r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, msg, corev1.ConditionFalse, cdLog)
 		}
+	}
+
+	// Check on the SyncSetsNotApplied condition. Usually this is happening on a freshly installed cluster that's
+	// running; checkClusterResumed will discover that state and flip the condition appropriately.
+	if hibernatingCondition.Reason == hivev1.SyncSetsNotAppliedReason && clusterSync.Status.FirstSuccessTime != nil {
+		return r.setHibernatingCondition(cd, hivev1.SyncSetsAppliedReason, "SyncSets have been applied", corev1.ConditionFalse, cdLog)
 	}
 
 	// Check if HibernateAfter is set, and put it to sleep if necessary.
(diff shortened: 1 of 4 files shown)


2uasimojo:
This is working nicely now.

```
hive1657-3   hive1657-3-fl84l   aws        us-east-1                           Provisioned       Initialized   36m
hive1657-3   hive1657-3-fl84l   aws        us-east-1   4.7.0                   Provisioned       Unsupported   36m
hive1657-3   hive1657-3-fl84l   aws        us-east-1   4.7.0                   Provisioned       SyncSetsNotApplied   36m
hive1657-3   hive1657-3-fl84l   aws        us-east-1   4.7.0                   Provisioned       Initialized          37m
hive1657-3   hive1657-3-fl84l   aws        us-east-1   4.7.0                   Provisioned       Running              37m
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -224,6 +225,18 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// Clear any lingering SyncSetsNotApplied condition. Usually this is happening on a freshly installed cluster that's
+	// running; checkClusterResumed will discover that state and flip the condition appropriately. Otherwise we would end
+	// up going through Resuming, artificially subject to the nodeCheckWaitTime.
+	// TODO: Refactor this reconcile to be more idempotent, with "detect state, set conditions" in a single self-contained
+	// code block.
+	if hibernatingCondition.Reason == hivev1.SyncSetsNotAppliedReason && clusterSync.Status.FirstSuccessTime != nil {
+		if result, err := r.setHibernatingCondition(cd, hivev1.InitializedConditionReason, "Condition Initialized", corev1.ConditionUnknown, cdLog); err != nil {
```
suhanime, line 234: Please set this to syncsets applied reason instead of initialized

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -534,7 +547,8 @@ func (r *hibernationReconciler) nodesReady(cd *hivev1.ClusterDeployment, remoteC
 	if hibernatingCondition == nil {
 		return false, errors.New("cannot find hibernating condition")
 	}
-	if time.Since(hibernatingCondition.LastProbeTime.Time) < nodeCheckWaitTime {
+	// Don't delay nodeCheckWaitTime the first time through.
+	if hibernatingCondition.Reason != hivev1.InitializedConditionReason && time.Since(hibernatingCondition.LastProbeTime.Time) < nodeCheckWaitTime {
```
suhanime, line 551: check for status unknown instead?

2uasimojo: Done.

2uasimojo: This is now checking for the new SyncSetsApplied reason instead.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -138,6 +145,31 @@ func TestReconcile(t *testing.T) {
...
+				actuator.EXPECT().MachinesRunning(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(true, nil, nil)
+			},
+			setupRemote: func(builder *remoteclientmock.MockBuilder) {
+				fakeClient := fake.NewFakeClientWithScheme(scheme, readyNodes()...)
+				builder.EXPECT().Build().Times(1).Return(fakeClient, nil)
+			},
+			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
+				cond := getHibernatingCondition(cd)
+				require.NotNil(t, cond)
+				assert.NotEqual(t, hivev1.SyncSetsNotAppliedReason, cond.Reason)
```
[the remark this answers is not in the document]
2uasimojo: This is now explicitly checking for the new SyncSetsApplied/False condition.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -323,7 +329,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 				return r.startMachines(cd, cdLog)
 			}
 			return reconcile.Result{}, nil
-		case hivev1.ResumingHibernationReason:
+		case hivev1.ResumingHibernationReason, hivev1.SyncSetsAppliedReason:
```
suhanime, line 332: I know why you had to add it here, but for a human reading english, it makes more sense to say if we should be running (aka !shouldhibernate) and hibernating condition = false but status != running, check cluster resumed... You don't have to fix it, I would take care of it in the refactor

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -534,7 +540,8 @@ func (r *hibernationReconciler) nodesReady(cd *hivev1.ClusterDeployment, remoteC
 	if hibernatingCondition == nil {
 		return false, errors.New("cannot find hibernating condition")
 	}
-	if time.Since(hibernatingCondition.LastProbeTime.Time) < nodeCheckWaitTime {
+	// Don't delay nodeCheckWaitTime if we just discovered SyncSets have been applied
+	if hibernatingCondition.Reason != hivev1.SyncSetsAppliedReason && time.Since(hibernatingCondition.LastProbeTime.Time) < nodeCheckWaitTime {
```
suhanime, line 544: another thing that needs to make more sense when the refactor happens - not waiting because syncsets have applied makes no sense :weary:

suhanime (commented):
/lgtm

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: Yeah, I think the refactor should rename some things. Like `checkClusterResumed` should be `checkClusterRunning` -- in this case, we never hibernated, so we never resumed; and the function's job has nothing to do with resuming.

Et cetera.

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: Right... except with the zoomed-out view that SyncSetsNotApplied => SyncSetsApplied is part of the install (as opposed to resume) sequence, where it doesn't make sense to impose the delay designed for resuming.

PR #1568, opened 2021-10-05
ProvisionDelay metric shouldn't log unknown conds

suhanime:
Provision underway metric logged conditions even if they were of unknown status. Since we are initializing all the conditions now, it doesn't seem appropriate to do so anymore.

This commit ensures the provision underway metric is logged if and only if there are conditions present on the cluster deployment in undesired state. For conditions that are not part of the predetermined [list](https://github.com/openshift/hive/blob/master/pkg/controller/metrics/provision_underway_collector.go#L19-L26) of provision delay conditions, if there are conditions in error/undesired state, we would log the metric but condition and reason would be unknown. This should help with better alerts.

[HIVE-1660](https://issues.redhat.com/browse/HIVE-1660)

/assign @2uasimojo

### DIFF
--- a/pkg/controller/metrics/provision_underway_collector.go
+++ b/pkg/controller/metrics/provision_underway_collector.go
@@ -7,6 +7,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
+	corev1 "k8s.io/api/core/v1"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
@@ -60,9 +62,6 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			continue
 		}
 
-		// Add install failure details for stuck provision
-		condition, reason := getKnownConditions(cd.Status.Conditions)
-
 		platform := cd.Labels[hivev1.HiveClusterPlatformLabel]
 		imageSet := "none"
 		if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.ImageSetRef != nil {
@@ -74,6 +73,11 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			continue // skip reporting the metric for clusterdeployment until the elapsed time is at least minDuration
 		}
 
+		// Add install failure details for stuck provision
+		condition, reason, skip := getConditionAndReason(cd.Status.Conditions)
+		if skip {
+			continue
+		}
 		// For installing clusters we report the seconds since the cluster was created.
 		ch <- prometheus.MustNewConstMetric(
 			cc.metricClusterDeploymentProvisionUnderwaySeconds,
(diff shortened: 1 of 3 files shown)


pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -164,6 +175,15 @@ func (cc provisioningUnderwayInstallRestartsCollector) Collect(ch chan<- prometh
 			continue // skip reporting the metric for clusterdeployment until the InstallRestarts is at least minRestarts
 		}
 
+		// Skip logging the metric if conditions in desired state
```
2uasimojo, line 178: I would rather see all the logic embedded in `getKnownConditions`, so you can immediately (after L162) do something like:
```go
if condition == something {
	continue
}
```

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -74,6 +76,15 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			continue // skip reporting the metric for clusterdeployment until the elapsed time is at least minDuration
 		}
 
+		// Skip logging the metric if conditions in desired state
+		if condition == "Unknown" && len(cd.Status.Conditions) > 0 {
```
2uasimojo, line 80: I would rather see all the logic embedded in `getKnownConditions`, so you can immediately (after L66) do something like:
```go
if condition == something {
	continue
}
```

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -217,5 +228,14 @@ func getKnownConditions(conditions []hivev1.ClusterDeploymentCondition) (conditi
...
 	}
-	return condition, reason
+	// if known condition not found, but some condition is in undesired state, report it
+	if condition == "Unknown" && len(conditions) > 0 {
+		// conditions would be sorted to have error conditions on the top, so if the first condition is in desired state,
+		// we can skip reporting it
+		if conditions[0].Status == corev1.ConditionUnknown ||
+			controllerutils.IsConditionInDesiredState(conditions[0]) {
+			skip = true
+		}
```
2uasimojo, line 238: - It's still ooking me out a bit how tightly bound this is to the sorting algorithm housed entirely elsewhere.
- If the top condition is not in the desired state, why don't we report *that*, rather than "Unknown"?

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -74,6 +73,11 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 			continue // skip reporting the metric for clusterdeployment until the elapsed time is at least minDuration
 		}
 
+		// Add install failure details for stuck provision
+		condition, reason, skip := getConditionAndReason(cd.Status.Conditions)
+		if skip {
+			continue
+		}
```
2uasimojo, line 80: Yah, I like this a lot better.

pkg/controller/metrics/provision_underway_collector.go
```diff
@@ -217,5 +228,11 @@ func getKnownConditions(conditions []hivev1.ClusterDeploymentCondition) (conditi
 			}
 		}
 	}
-	return condition, reason
+	// skip reporting if all conditions are in their desired state
+	if condition == "Unknown" && len(conditions) > 0 {
+		if controllerutils.AreAllConditionsInDesiredState(conditions) {
+			skip = true
+		}
+	}
```
2uasimojo, line 236: nit:
```suggestion
	if condition == "Unknown" && len(conditions) > 0 && controllerutils.AreAllConditionsInDesiredState(conditions) {
		skip = true
	}
```

PR #1572, opened 2021-10-11
OSD-8194 : modifying order of operation for syncset update to avoid deleting new resource version

bdematte:
In case the syncset managing a resource changes, the resulting state isn't the expected one. 
Indeed, we first apply the new SyncSet, and then delete any resources attached to syncset which aren't existing anymore. In case of the same resource being now owned by a different syncset (like for user-worklad-monitoring which triggers the deployment of the same configmap but with a different content), this is leading to a first application of the syncset which is not valid (new version being eventually deleted at the end of the initial reconcile loop).
Goal of this MR is to enforce the deletion of any resource managed by a former syncset before applying changes, to ensure we don't delete new version of a resource.

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -544,16 +544,44 @@ func (r *ReconcileClusterSync) applySyncSets(
 		return syncSets[i].AsMetaObject().GetName() < syncSets[j].AsMetaObject().GetName()
 	})
 
+	deletionList := syncStatuses
+
 	for _, syncSet := range syncSets {
-		logger := logger.WithField(syncSetType, syncSet.AsMetaObject().GetName())
-		oldSyncStatus, indexOfOldStatus := getOldSyncStatus(syncSet, syncStatuses)
+		_, indexOfOldStatus := getOldSyncStatus(syncSet, deletionList)
 		// Remove the matching old sync status from the slice of sync statuses so that the slice only contains sync
 		// statuses that have not been matched to a syncset.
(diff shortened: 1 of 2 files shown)


dofinn:
@2uasimojo @abutcher anything oustanding on this PR ?

2uasimojo:
> @2uasimojo @abutcher anything oustanding on this PR ?

We were waiting for @bdematte to respond. We're on board with the concept.

abutcher:
We should include the tests from https://github.com/abutcher/hive/commit/ddcd9d0d38dc3fab510740240037713e08e5f57b and then this should be ready to go.

2uasimojo:
I would also like someone (anyone) to ack that they've tested this with a live cluster.

bdematte:
I updated the MR locally to add the test, and this is ready for MR update. However, I spent some time to try to run it on a live cluster and couldn't succeed as some needed packages aren't Mac-friendly (couldn't find workaround).
If someone can help with the validation on a live cluster, I can push the updated version with added test.

abutcher:
@bdematte Sounds good, I can help with validating on a cluster.

abutcher:
Tested by creating a syncset managing a configmap and then moving that configmap resource to a new syncset https://asciinema.org/a/uswmz6HMCzw6U71raSbHtR1fP

abutcher:
/lgtm
@2uasimojo Leaving hold, PTAL and remove if you're cool with the validation/test

2uasimojo:
Validated.

/hold cancel

Let's keep a pretty close eye on OSD as this gets merged and promoted.

PR #1573, opened 2021-10-12
Expand RBAC for ClusterPool CD namespaces

2uasimojo:
Subjects with a RoleBinding to the hive-cluster-pool-admin ClusterRole
have that RoleBinding copied into the namespaces of ClusterDeployments
created under a ClusterPool. Previously that ClusterRole only included
read (`get`/`list`/`watch`) access to ClusterDeployments and
ClusterProvisions. This change adds write (`update` and `patch`, but
notably not `create`) access, and extends the role to include
ClusterDeprovisions as well.

[HIVE-1664](https://issues.redhat.com/browse/HIVE-1664)

### DIFF
--- a/config/rbac/hive_clusterpool_admin.yaml
+++ b/config/rbac/hive_clusterpool_admin.yaml
@@ -29,10 +29,14 @@ rules:
   resources:
   - clusterdeployments
   - clusterprovisions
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
/lgtm

PR #1577, opened 2021-10-13
Unexport metrics

2uasimojo:
138c8e0 / #1525 / [HIVE-1581](https://issues.redhat.com/browse/HIVE-1581) added `HiveConfig.spec.exportMetrics` and
logic to the operator to enable exporting metrics when this setting is
`true`. This commit adds logic to undo same when it is changed from
`true` to `false` (or removed).

[HIVE-1655](https://issues.redhat.com/browse/HIVE-1655)

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -47,10 +47,21 @@ const (
 	hiveConfigHashAnnotation = "hive.openshift.io/hiveconfig-hash"
 
 	hiveClusterSyncStatefulSetSpecHashAnnotation = "hive.openshift.io/clustersync-statefulset-spec-hash"
+
+	// clusterMonitoringLabel on a namespace tells prometheus to monitor that namespace
+	clusterMonitoringLabel = "openshift.io/cluster-monitoring"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/operator/hive/hive.go
```diff
@@ -540,46 +540,46 @@ func monitoringLabelPatch(enable bool) string {
...
-	}
-
-	// Load namespaced assets, decode them, set to our target namespace, and apply:
-	for _, assetPath := range monitoringNamespacedAssets {
-		if err := util.ApplyAssetWithNSOverrideAndGC(h, assetPath, hiveNSName, instance); err != nil {
-			hLog.WithError(err).Error("error applying object with namespace override")
-			return err
-		}
-		hLog.WithField("asset", assetPath).Info("applied asset with namespace override")
+	var labelVal, verb1, verb2, verb3 string
```
joelddiaz, line 561: can we call those verbs something like 'desiredStateVerb' 'progressingVerb' 'completedVerb' , or just anything to make their purpose a bit more clear?

2uasimojo:
This is ready to go @joelddiaz 

Thanks!

PR #1578, opened 2021-10-14
provide a UBI-based Dockerfile

joelddiaz:
Introduce a new Dockerfile.ubi which will use the RHEL UBI container as
the base/runtime container image.

Add new CONTAINER_BUILD_FLAGS into the Makefile and set it to use the
original 'Dockefile'.

Update hack/app_sre_build_deploy.sh to overide this
CONTAINER_BUILD_FLAGS to use 'Dockerfile.ubi' as the Dockerfile to use.

Remove the unused BUILD_CMD flag from the hack/app_sre_build_deploy.sh
script as it is a no-op in our Makefile.

xref: https://issues.redhat.com/browse/HIVE-1668

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -1,0 +1,46 @@
+FROM registry.ci.openshift.org/openshift/release:golang-1.16 as builder
+RUN mkdir -p /go/src/github.com/openshift/hive
+WORKDIR /go/src/github.com/openshift/hive
+COPY . .
+RUN make build
+
+FROM registry.access.redhat.com/ubi8/ubi:latest
+
+# CentOS images do not get updates as they are meant to mirror ISO content, and thus this update
+# is strongly recommended for security updates.
+RUN dnf -y update && dnf clean all
+
+# ssh-agent required for gathering logs in some situations:
+RUN if ! rpm -q openssh-clients; then dnf install -y openssh-clients && dnf clean all && rm -rf /var/cache/dnf/*; fi
+
+# libvirt libraries required for running bare metal installer.
+RUN if ! rpm -q libvirt-devel; then dnf install -y libvirt-devel && dnf clean all && rm -rf /var/cache/dnf/*; fi
+
+COPY --from=builder /go/src/github.com/openshift/hive/bin/manager /opt/services/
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -269,7 +270,7 @@ $(addprefix generate-submodules-,$(GO_SUB_MODULES)):
 .PHONY: docker-build
 docker-build:
 	@echo "*** DEPRECATED: Use the image-hive target instead ***"
-	$(DOCKER_CMD) build -t ${IMG} .
+	$(DOCKER_CMD) build $(DOCKER_BUILD_FLAGS) -t ${IMG} .
```
2uasimojo, line 273:
```suggestion
	$(DOCKER_CMD) build $(CONTAINER_BUILD_FLAGS) -t ${IMG} .
```

2uasimojo (changes requested):
I think you forgot to `git add Dockerfile.ubi`.

Dockerfile.ubi
```diff
@@ -0,0 +1,46 @@
+FROM registry.ci.openshift.org/openshift/release:golang-1.16 as builder
```
2uasimojo, line 1: This file differs from `Dockerfile` only in this line, which seems like The Right Thing™

2uasimojo (commented):
This lgtm. However, I wonder if we want to merge it into a temporary topic branch first so we can control which environment(s) we deploy it to initially for validation (since there's really no way to test it without shoving it through the appsre pipeline).

2uasimojo (approved):

Dockerfile.ubi
2uasimojo: Sorry, I didn't mean this line...

Dockerfile.ubi
```diff
@@ -0,0 +1,46 @@
...
+RUN mkdir -p /go/src/github.com/openshift/hive
+WORKDIR /go/src/github.com/openshift/hive
+COPY . .
+RUN make build
+
+FROM registry.access.redhat.com/ubi8/ubi:latest
```
2uasimojo, line 7: ...I meant this line.

PR #1580, opened 2021-10-14
Refactor hibernation

suhanime:
Simplify the logic of hibernation controller in order to
better support the upcoming changes. Some noteworthy changes are
- streamline hibernateAfter and shouldHibernate logic to have no
  repeating code
- separate out and untangle decisions made on conditions and their
  reasons in order to eliminate haphazard patches and to support adding
  new condition down the line
- eliminate dead code paths
- minor bug fixes and renames

xref: https://issues.redhat.com/browse/HIVE-1672

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -194,11 +194,21 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		return r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, msg, corev1.ConditionFalse, cdLog)
 	}
 
+	isFakeCluster := controllerutils.IsFakeCluster(cd)
+	var clusterSyncExists = true
+
 	clusterSync := &hiveintv1alpha1.ClusterSync{}
 	if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync); err != nil {
-		// This may be NotFound, which means the clustersync controller hasn't created the ClusterSync yet.
-		// Fail and requeue to wait for it to exist.
-		return reconcile.Result{}, fmt.Errorf("could not get ClusterSync: %v", err)
+		if isFakeCluster && apierrors.IsNotFound(err) {
+			// fake clusters don't create clustersync object by default
+			// This check is used further in the code when we check for SyncSetsNotApplied
+			// TODO: fix this when fake clusters create clusterSync by default like real clusters
+			clusterSyncExists = false
+		} else {
+			// This may be NotFound, which means the clustersync controller hasn't created the ClusterSync yet.
+			// Fail and requeue to wait for it to exist.
+			return reconcile.Result{}, fmt.Errorf("could not get ClusterSync: %v", err)
+		}
 	}
 
 	// Check on the SyncSetsNotApplied condition. Usually this is happening on a freshly installed cluster that's
@@ -207,7 +217,6 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		return r.setHibernatingCondition(cd, hivev1.SyncSetsAppliedReason, "SyncSets have been applied", corev1.ConditionFalse, cdLog)
 	}
 
-	isFakeCluster := controllerutils.IsFakeCluster(cd)
 	shouldHibernate := cd.Spec.PowerState == hivev1.HibernatingClusterPowerState
 	// set readyToHibernate if hibernate after is ready to kick in hibernation
 	var readyToHibernate bool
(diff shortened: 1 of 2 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -541,7 +532,8 @@ func (r *hibernationReconciler) nodesReady(cd *hivev1.ClusterDeployment, remoteC
 		return false, errors.New("cannot find hibernating condition")
 	}
 	// Don't delay nodeCheckWaitTime if we just discovered SyncSets have been applied
-	if hibernatingCondition.Reason != hivev1.SyncSetsAppliedReason && time.Since(hibernatingCondition.LastProbeTime.Time) < nodeCheckWaitTime {
+	if hibernatingCondition.Reason == hivev1.SyncSetsAppliedReason &&
```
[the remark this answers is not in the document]
2uasimojo: It was correct before:
- If the hibernating condition is SyncSetsApplied, *don't* return; carry on reconciling
- Otherwise, decide whether we've delayed long enough. If so, carry on reconciling; otherwise return.

suhanime:
/test verify

Rerunning because succeeded locally

suhanime:
> I'm going to go through this now. One question though:
> 
> You said this fixes a regression for fake clusters. Having UT for that scenario would have prevented the regression. Can we add it here? Or... did we talk about adding that case to e2e-pool?

There are UTs present already - see [1](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller_test.go#L404-L418), [2](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller_test.go#L419-L432), [3](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller_test.go#L639-L652), [4](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller_test.go#L653-L664) and [5](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller_test.go#L665-L673). I do not know why they differ from live test environment though.
I have decided to add the test in e2e-pool and currently working on that - depending on when a round of review happens, it can go either with this PR or later

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -184,165 +184,156 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
...
+			return r.checkClusterStopped(cd, false, cdLog)
+		}
+	} else {
+		if shouldStartMachines(cd) {
 			if controllerutils.IsFakeCluster(cd) {
 				r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, "Fake cluster is running",
 					corev1.ConditionFalse, cdLog)
 			} else {
 				return r.startMachines(cd, cdLog)
 			}
```
2uasimojo, line 331:
```suggestion
				return r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, "Fake cluster is running",
					corev1.ConditionFalse, cdLog)
			}
			return r.startMachines(cd, cdLog)
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -184,165 +184,156 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
...
-			return reconcile.Result{}, nil
-		case hivev1.ResumingHibernationReason, hivev1.SyncSetsAppliedReason:
-			return r.checkClusterResumed(cd, cdLog)
+		} else if hibernatingCondition.Reason == hivev1.ResumingHibernationReason ||
```
2uasimojo, line 332: ...and then this `else` can go away.

2uasimojo (commented):
Some suggestions inline, but nothing that needs to be done right away. It's up to you.
I'm just running a quick sanity check on a couple of code paths. Will update first thing tomorrow.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -184,165 +184,160 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
...
+			return r.checkClusterStopped(cd, false, cdLog)
+		}
+	} else {
+		if shouldStartMachines(cd, hibernatingCondition) {
+			if isFakeCluster {
 				r.setHibernatingCondition(cd, hivev1.RunningHibernationReason, "Fake cluster is running",
 					corev1.ConditionFalse, cdLog)
-			} else {
-				return r.startMachines(cd, cdLog)
+				return reconcile.Result{}, nil
```
2uasimojo, line 332: This should return the error if setting the status fails. Just return the result of `setHibernatingCondition`.

PR #1583, opened 2021-10-15
docker/podman login for appsre

2uasimojo:


2uasimojo:
> /lgtm so much for alias docker='podman' ...

IKR?? Drop in replacement, they said. You won't even notice, they said. That has *never* been the case for me.

PR #1586, opened 2021-10-18
Add useful metrics for ClusterPools

2uasimojo:
[HIVE-1677](https://issues.redhat.com/browse/HIVE-1677)

### DIFF
--- a/pkg/controller/clusterpool/metrics.go
+++ b/pkg/controller/clusterpool/metrics.go
@@ -7,6 +7,38 @@ import (
 )
 
 var (
+	metricClusterDeploymentsAssignable = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_clusterpool_clusterdeployments_assignable",
+		Help: "The number of ClusterDeployments ready to be claimed. Contributes to Size and MaxSize.",
+	}, []string{"clusterpool_namespace", "clusterpool_name"})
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterpool/metrics.go
```diff
@@ -7,6 +7,40 @@ import (
...
+	metricClusterDeploymentsDeleting = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_clusterpool_clusterdeployments_deleting",
+		Help: "The number of ClusterDeployments marked for or actively deprovisioning and deleting. Contributes to MaxConcurrent.",
+	}, []string{"clusterpool_namespace", "clusterpool_name"})
+	metricClusterDeploymentsInstalling = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "hive_clusterpool_clusterdeployments_installing",
+		Help: "The number of ClusterDeployments being added to the pool, in the process of installing. Contributes to Size, MaxSize, and MaxConcurrent.",
+	}, []string{"clusterpool_namespace", "clusterpool_name"})
+	// This is always assignable+installing. But it's the number we strive to keep equal to the pool size.
+	// Is it useful to include?
```
suhanime, line 27: I think that's a simple prom query so I vote for scraping it

2uasimojo: But OTOH, the consumer of these metrics would have to understand that internal business logic. Including knowing that it should include "standby" once that's implemented. I'm leaning towards keeping it :)

PR #1588, opened 2021-10-20
HIVE-1669: Create Azure Client for Azure CloudName.

abutcher:
/assign @joelddiaz

### DIFF
--- a/pkg/azureclient/client.go
+++ b/pkg/azureclient/client.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2018-05-01/dns"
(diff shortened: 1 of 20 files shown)


pkg/azureclient/client.go
```diff
@@ -155,23 +156,26 @@ func newClient(authJSONSource func() ([]byte, error)) (*azureClient, error) {
 		return nil, errors.New("missing subscriptionId in auth")
 	}
 
-	config := auth.NewClientCredentialsConfig(clientID, clientSecret, tenantID)
+	env, err := azure.EnvironmentFromName(string(environmentName))
```
joelddiaz, line 159: can we do the common `if environmentName == "" { environmentName = azure.PublicCloud.Name }` dance here instead of sprinkled all around the various files?

ldpliu:
@abutcher this pr should be cherry-pick to ocm-2.4 branch. Thank you

PR #1591, opened 2021-10-21
Create clustersyncs for fake clusters.

abutcher:


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -162,20 +161,34 @@ func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) (*R
...
+	logger log.FieldLogger,
+) (
+	resource.Helper,
+	error,
+) {
+	if controllerutils.IsFakeCluster(cd) {
 		return resource.NewFakeHelper(logger), nil
 	}
+
+	restConfig, err := remoteClusterAPIClientBuilderFunc(cd).RESTConfig()
```
abutcher, line 186: Rest config moved to here within `resourceHelperBuilderFunc`. Tests no longer need to mock a call to `.RESTConfig()` since the test helper returns a fake resource helper.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -162,20 +161,34 @@ func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) (*R
...
 		},
+		resourceHelperBuilder: resourceHelperBuilderFunc,
 	}, nil
 }
 
-func resourceHelperBuilderFunc(restConfig *rest.Config, fakeCluster bool, logger log.FieldLogger) (resource.Helper, error) {
-	if fakeCluster {
+func resourceHelperBuilderFunc(
+	cd *hivev1.ClusterDeployment,
+	remoteClusterAPIClientBuilderFunc func(cd *hivev1.ClusterDeployment) remoteclient.Builder,
```
2uasimojo, line 176: It seems odd that we can't just access this via `r` from within `r.resourceHelperBuilder`... but I get it :(

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -162,20 +161,34 @@ func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) (*R
...
-		logger:                logger,
-		reapplyInterval:       reapplyInterval,
-		resourceHelperBuilder: resourceHelperBuilderFunc,
+		Client:          c,
+		logger:          logger,
+		reapplyInterval: reapplyInterval,
 		remoteClusterAPIClientBuilder: func(cd *hivev1.ClusterDeployment) remoteclient.Builder {
 			return remoteclient.NewBuilder(c, cd, ControllerName)
 		},
+		resourceHelperBuilder: resourceHelperBuilderFunc,
```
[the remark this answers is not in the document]
2uasimojo: But only because you moved `resourceHelperBulider` down. What was the need to do that?

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -199,16 +199,9 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 
 	clusterSync := &hiveintv1alpha1.ClusterSync{}
 	if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync); err != nil {
-		if isFakeCluster && apierrors.IsNotFound(err) {
-			// fake clusters don't create clustersync object by default
-			// This check is used further in the code when we check for SyncSetsNotApplied
-			// TODO: fix this when fake clusters create clusterSync by default like real clusters
-			clusterSyncExists = false
```
suhanime: This variable is no longer needed so should be dropped

PR #1607, opened 2021-11-08
Enhancement for clusterpool to work on on-prem clouds

akhil-rane:
Currently, clusterpool is only available for public clouds like
aws, gcp and azure. Installing cluster on on-prem clouds require
additional pre-configurations that cannot be done on a fly as
currently implemented in clusterpool. Through this enhancement
we propose how these pre-configurations can be fed into the
clusterpool through a new `inventory` field.

x-ref: https://issues.redhat.com/browse/HIVE-1367


akhil-rane:
@2uasimojo this is mostly based on your original proposal, but I have included the changes that we discussed. PTAL

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+      Here's where we bounce if two controllers try to grab the same InstallConfigPatch: the loser will bounce the update with a 409; we'll requeue and start over.
+    - Proceed with existing algorithm to create new cluster.
+- Add a top-level routine to ensure all unavailable InstallConfigPatches have valid ClusterDeployments associated with them.
+  If not, attempt to create them through the same flow.
+  This is to guard against leaking entries if a controller crashes between updating the lease on InstallConfigPatch and creating the CD.
+
+### Fairness
+We will assume it is **not** important that we rotate through the list of supplied install config patches in any particular order, or with any nod to "fairness".
+For example: If there are five patches and the usage pattern happens to only use two at a time, there is no guarantee that we will round-robin through all five.
+We may reuse the same two over and over, or pick at random, or something else -- any of which has the potential to "starve" some names.
```
2uasimojo, line 175:
```suggestion
We may reuse the same two over and over, or pick at random, or something else -- any of which has the potential to "starve" some patches.
```

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+        installConfigPatches:
+          - name: foo-install-config-patch
+    ```
+  
+### Validation
+Webhook validation will ensure that for a clusterpool
+- if `spec.inventory.installConfigPatchs` is specified, it is not an empty list.
+- if `spec.inventory.installConfigPatchs` is specified, `spec.maxSize` is equal to or less than length of list `spec.inventory.installConfigPatchs`.
+- if `spec.inventory.installConfigPatchs` is specified, `spec.size` is equal to or less than length of list `spec.inventory.installConfigPatchs`.
+
```
2uasimojo, line 123: Other validation scenarios to consider:
- At what point do we validate that an inventory entry has an InstallConfigPatch associated with it? What do we do if it doesn't? (I think we do this in the flow where we're trying to create a new CD; and I think we log an error and move on to the next one.)
- What do we do if a patch fails, e.g. because it's malformed? (I think we log an error, add a status condition to the InstallConfigPatch, and move on to the next one. But the status condition should be write-only -- e.g. we shouldn't subsequently ignore the patch, because we can't know whether, when, or how the user has "fixed" it.)

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+- When we need to build a new ClusterDeployment, we
+    - Load up the InstallConfigPatch list from the inventory.
+    - Do a GET on each entry in the list and check if the `PatchAvailable` condition is true
+    - For the first InstallConfigPatch that is available, we will use the json patch in the `spec.patch` field to update the default install config generated by the clusterpool.
+    - Update the `spec.clusterPoolReference.installConfigPatchRef`field in the ClusterDeployment with a reference to InstallConfigPatch
+    - Post an update to the InstallConfigPatch with the reference to the ClusterDeployment. Also update the `PatchAvailable` condition to false.
+      Here's where we bounce if two controllers try to grab the same InstallConfigPatch: the loser will bounce the update with a 409; we'll requeue and start over.
+    - Proceed with existing algorithm to create new cluster.
+- Add a top-level routine to ensure all unavailable InstallConfigPatches have valid ClusterDeployments associated with them.
+  If not, attempt to create them through the same flow.
```
2uasimojo, line 169: ...within the bounds of the various pool capacities.

(This should be interesting to figure out. I think we would end up slotting this into [this switch statement](https://github.com/openshift/hive/blob/d66896015d990af5d33503f6f7539bc28a7a2c17/pkg/controller/clusterpool/clusterpool_controller.go#L375), possibly embedded in [this case](https://github.com/openshift/hive/blob/d66896015d990af5d33503f6f7539bc28a7a2c17/pkg/controller/clusterpool/clusterpool_controller.go#L383-L388).)

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+        - [Validation](#validation)
+        - [`Size` and `MaxSize`](#size-and-maxsize)
+        - [Pool Version](#pool-version)
+        - [Handling Inventory Updates](#handling-inventory-updates)
+            - [Adding An Inventory](#adding-an-inventory)
+            - [Adding An Entry](#adding-an-entry)
+            - [Removing An Entry](#removing-an-entry)
+            - [Deleting The Inventory](#deleting-the-inventory)
+        - [Racing For Names](#racing-for-names)
+        - [Fairness](#fairness)
```
2uasimojo, line 21: TOC needs update. FWIW, I use the following VSCode extension, which keeps the TOC up to date automatically:

Name: Markdown All in One
Id: yzhang.markdown-all-in-one
Description: All you need to write Markdown (keyboard shortcuts, table of contents, auto preview and more)
Version: 3.4.0
Publisher: Yu Zhang
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one

2uasimojo (commented):
Very nice @akhil-rane! LMK if you want to talk through any of this.

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+When adding a ClusterDeployment, if such an `Inventory` is present, ClusterPool controller will:
+- Load up the inventory list.
+- For each reference it will do a GET to fetch InstallConfigPatch and check the if the PatchAvailable condition is true.
+- For the first InstallConfigPatch that is available, it will use the json patch in the `spec.patch` field to update the default install config generated by clusterpool. It will also update the status with reference to the cluster deployment that is using it and update the `PatchAvailable` condition to false.
+- Set the `spec.clusterPoolReference.installConfigPatchRef`field in the ClusterDeployment with a reference to InstallConfigPatch CR.
+
+Absent an `Inventory`, ClusterPool will continue to use the generated default install config as it does today.
+
+When ClusterDeployment is deleted, ClusterPool controller will: 
+- Find out if ClusterDeployment used an InstallConfigPatch from `spec.ClusterPoolReference.InstallConfigPatchRef`field.
```
joelddiaz, line 93: I assume the clusterpool controller will set a finalizer or use an existing finalizer to keep ClusterDeployments from disappearing before proper accounting can be taken?

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+- If yes, we simply update InstallConfigPatch, setting the `spec.clusterDeploymentRef` to nil and `PatchAvailable` to true.  
+
+### How To Use
+For the VSphere case, this allows the administrator to:
+- Preconfigure DNS with following entries (assuming cluster name is `foo`)
+    ```
+    10.0.0.10  api.foo.example.com
+    10.0.0.11  apps.foo.example.com
+    10.0.0.12  console.foo.example.com
```
joelddiaz, line 102: This is fine for the sake of an example, but I was under the impression that a cluster only needed 2 IPs. One for api.CLUSTER.domain and one for *.apps.CLUSTER.domain (where console-openshift-console.apps.CLUSTER.domain will just piggy back off of the *.apps.CLUSTER.domain entry).

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+- Add the name of InstallConfigPatch CR to`clusterPool.spec.inventory.installConfigPatches` list. For InstallConfigPatch with a name `foo-install-config-patch` the clusterpool should be configured as follows
+    ```yaml
+    spec:
+      inventory:
+        installConfigPatches:
+          - name: foo-install-config-patch
+    ```
+  
+### Validation
+Webhook validation will ensure that for a clusterpool
```
joelddiaz, line 119: It feels like it should just be a runtime condition if a user sets maxSize and size to nonsensical values. If you put size 5, and inventory contains 2 items, then you get 2 with a condition saying "cant get to size 5".

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Adding An Entry to the Inventory
+If `MaxSize` is unset or less than/equal to the length of the inventory, and `Size` allows, adding a new entry will cause a new ClusterDeployment to be added to the pool.
+If the pool is already at `[Max]Size` there is no immediate effect.
+
+#### Removing An Entry to the Inventory
+- If the entry is unused (no pool CD with that name exists), this is a no-op.
+- If an _unclaimed_ CD exists with that name, we delete it.
+  The controller will replace it, assuming an unused entry is available.
+- If a _claimed_ CD exists with that name, no-op.
```
joelddiaz, line 147: I feel like someone is inevitably going to ask "i deleted from the inventory. how do i know if it is presently in use?"

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+In such cases, one solution is to configure DNS manually and then create your ClusterDeployment with the right `Name` and `BaseDomain` so that the assembled hostnames match those entries. This is fine if you're creating ClusterDeployments by hand, but breaks down if ClusterDeployment names are being generated with random slugs, as is the case with ClusterPools.
+
+## Proposal
+### Summary
+Allow ClusterPool to accept an inventory of install config json patches to be used when generating ClusterDeployments for the pool.
+
+### `ClusterPool.Spec.Inventory`
+Add a field to `ClusterDeployment.Spec` called `Inventory`.
+
+It has an array of object references to InstallConfigPatch custom resource. This CR will have a base64 encoded json patch for default install config generated by clusterpool controller. The json patch will have cluster specific changes, for vSphere this corresponds to configured DNS hostnames
```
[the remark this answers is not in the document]
akhil-rane: As you mentioned, yes I was worried about escape quotes and that kind of stuff in JSON. I am cool with the inclined yaml option. Looks neat! Not entirely sure how to implement it though. Do we envision a separate struct?
```
struct PatchEntity {
   op string
   path string
   value string
}
```

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+
+```yaml
+spec:
+  inventory:
+    installConfigPatches:
+      - name: foo-install-config-patch
```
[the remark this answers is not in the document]
akhil-rane: Mentioned in line 44. Let me know if the wording needs to be changed

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+
+```json
+[
+  { "op": "replace", "path": "metadata/name", "value": "foo" }
+]
+```
+
+When adding a ClusterDeployment, if such an `Inventory` is present, ClusterPool controller will:
+- Load up the inventory list.
+- For each reference it will do a GET to fetch InstallConfigPatch and check the if the PatchAvailable condition is true.
```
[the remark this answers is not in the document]
akhil-rane: Any particular reason why? Is this a bad practice? Historically hive has chosen to use conditions for a status checks. We can go ahead and check for both `PatchAvailable` condition is true and  `.status.clusterDeploymentRef` == nil.

docs/enhancements/clusterpool-inventory.md
akhil-rane: Good points! Both your suggestions are fair enough, we would also like to add degraded condition on Clusterpool to notify problems with InstallConfigPatches

docs/enhancements/clusterpool-inventory.md
akhil-rane: should there be a finalizer on ClusterDeployment that prevents it from getting deleted before releasing the lease on InstallConfigPatch?

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+      status: "false"
+      type: PatchAvailable
+```
+
+if DNS is configured with the name `foo`, The base 64 encoded json content to patch vSphere install config will be as follows
+
+```json
+[
+  { "op": "replace", "path": "metadata/name", "value": "foo" }
+]
```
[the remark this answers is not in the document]
akhil-rane: I am good with RFC 6902, simple and easy to use

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+      inventory:
+        installConfigPatches:
+          - name: foo-install-config-patch
+    ```
+  
+### Validation
+Webhook validation will ensure that for a clusterpool
+- if `spec.inventory.installConfigPatchs` is specified, it is not an empty list.
+- if `spec.inventory.installConfigPatchs` is specified, `spec.maxSize` is equal to or less than length of list `spec.inventory.installConfigPatchs`.
+- if `spec.inventory.installConfigPatchs` is specified, `spec.size` is equal to or less than length of list `spec.inventory.installConfigPatchs`.
```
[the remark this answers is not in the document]
akhil-rane: agreed, removing this

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+
+### `ClusterPool.Spec.Inventory`
+Add a field to `ClusterDeployment.Spec` called `Inventory`.
+
+It has an array of object references to InstallConfigPatch custom resource. This CR will have a base64 encoded json patch for default install config generated by clusterpool controller. The json patch will have cluster specific changes, for vSphere this corresponds to configured DNS hostnames
+
+```yaml
+spec:
+  inventory:
+    installConfigPatches:
```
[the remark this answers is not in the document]
akhil-rane: Whatever we have in `clusterpool.spec.inventory` affects ClusterDeployments so `clusterDeploymentCustomization` may be? I am not good at naming @2uasimojo any suggestions?

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+
+and InstallConfigPatch CR will look like
+```yaml
+apiVersion: v1
+kind: InstallConfigPatch
+metadata:
+  name: foo-install-config-patch
+  namespace: my-project
+spec:
+  patch: <base64 encoded json patch>
```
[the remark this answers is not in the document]
akhil-rane: What do you think about the inclined yaml suggested by Eric [here](https://github.com/openshift/hive/pull/1607#discussion_r745983236)? That way we can have a list of patches

docs/enhancements/clusterpool-inventory.md
akhil-rane: yes, I will add that information to the doc

docs/enhancements/clusterpool-inventory.md
akhil-rane: you are right! only 2 IPs are enough.

docs/enhancements/clusterpool-inventory.md
akhil-rane: agreed, I will make that change. Also, maxSize (if set) should be equal to inventory size.

docs/enhancements/clusterpool-inventory.md
akhil-rane: We can just go ahead and check the status on the CRD

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,175 @@
...
+
+if DNS is configured with the name `foo`, The base 64 encoded json content to patch vSphere install config will be as follows
+
+```json
+[
+  { "op": "replace", "path": "metadata/name", "value": "foo" }
+]
+```
+
+When adding a ClusterDeployment, if such an `Inventory` is present, ClusterPool controller will:
```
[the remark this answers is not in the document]
akhil-rane: Assuming we have a controller that watches and ensures InstallConfigPatch is mapped to ClusterDeployment. The right order should be as follows
- clusterpool controller fetches the first available InstallConfigPatch
- Update finalizer for InstallConfigPatch, if it fails we just reconcile again
- Update status.ClusterDeploymentRef for InstallConfigPatch, if it fails we just reconcile again. The Finalizer of InstallConfigPatch will be cleared by the controller
- Add spec.ClusterPoolRef.InstallConfigPatchRef to ClusterDeployment and create cluster, if it fails we just reconcile again. The Finalizer and status of InstallConfigPatch will be cleared by the controller

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
joelddiaz: yes, the yaml patch list looks like it would work just fine

joelddiaz (commented):
One overarching question I have while reviewing this is related to having the ability to do more than just patch the install-config.
Can this mechanism/CRD be extended to make use of things like custom manifests ( https://github.com/openshift/hive/blob/master/apis/hive/v1/clusterdeployment_types.go#L191 )?

akhil-rane:
> One overarching question I have while reviewing this is related to having the ability to do more than just patch the install-config. Can this mechanism/CRD be extended to make use of things like custom manifests ( https://github.com/openshift/hive/blob/master/apis/hive/v1/clusterdeployment_types.go#L191 )?

Yes I believe, add one more field to spec and add implementation in the controller. We need to make CRD as generic as possible.

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
2uasimojo: If the data type is going to be a list of RFC6092 patches, I'm happy with `{prefix}Patches`, where `{prefix}` needs to refer to the object relative to which the `path` is computed.

That said, I don't think we can patch an installconfig using a json path relative to the CD spec.

So maybe we need
```yaml
clusterDeploymentCustomization:
  clusterDeploymentPatches:
    ...
  installConfigPatches:
    ...
```
or something.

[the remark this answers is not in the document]
akhil-rane: > 

This works for me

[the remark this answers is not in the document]
akhil-rane: this works for me.

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
2uasimojo: Nope, all good.

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,180 @@
...
+In such cases, one solution is to configure DNS manually and then create your ClusterDeployment with the right `Name` and `BaseDomain` so that the assembled hostnames match those entries. This is fine if you're creating ClusterDeployments by hand, but breaks down if ClusterDeployment names are being generated with random slugs, as is the case with ClusterPools.
+
+## Proposal
+### Summary
+Allow ClusterPool to accept an inventory of install config json patches to be used when generating ClusterDeployments for the pool.
+
+### `ClusterPool.Spec.Inventory`
+Add a field to `ClusterPool.Spec` called `Inventory`.
+
+It has an array of object references to InstallConfigPatch custom resource. This CR will have a base64 encoded json patch (RFC 6902) for default install config generated by clusterpool controller. The json patch will have cluster specific changes, for vSphere this corresponds to configured DNS hostnames
```
2uasimojo, line 44: Let's state that the patches will be applied in the order listed in the inventory.

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
2uasimojo: Okay, so
- Can we get rid of references to base64 encoding and fix the examples accordingly throughout?
- Let's rename this object to ClusterDeploymentCustomization to leave room for expansion.
- The .spec should for now just contain `installConfigPatches`, which would be `[]PatchEntity` as you [suggested](https://github.com/openshift/hive/pull/1607#discussion_r750639190). (Later we can add `clusterDeploymentPatches` which would also be `[]PatchEntity`.)

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
2uasimojo: Okay, so ClusterPool.Spec.Inventory will be []LocalObjectReference pointing to ClusterDeploymentCustomization (the artist formerly known as InstallConfigPatch -- see below), right?

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
2uasimojo: Cool. We can shake this out further in the implementation.

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
akhil-rane: yep

docs/enhancements/clusterpool-inventory.md
akhil-rane: Added on line 92

docs/enhancements/clusterpool-inventory.md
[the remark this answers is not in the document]
akhil-rane: Addressed

akhil-rane:
@abraverm you might want to review before we merge this enhancement

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,186 @@
...
+status:
+  clusterDeploymentRef: 
+    name: foo
+    namespace: foo-namespace
+  conditions:
+    - lastProbeTime: "2020-11-05T14:49:26Z"
+      lastTransitionTime: "2020-11-05T14:49:26Z"
+      message: Currently in use by cluster deployment foo of clusterpool foo-pool
+      reason: ClusterDeploymentCustomizationInUse
+      status: "false"
```
2uasimojo, line 76: We spell this `False` elsewhere. Let's keep it consistent.

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,186 @@
...
+  installConfigPatches:
+    - op: replace
+      path: metadata/name
+      value: foo
+```
+
+When adding a ClusterDeployment, if such an `Inventory` is present, ClusterPool controller will:
+- Load up the inventory list.
+- For each reference it will do a GET to fetch ClusterDeploymentCustomization and check if the `Available` condition is true and `.status.clusterDeploymentRef` is nil.
+- For the first ClusterDeploymentCustomization that is available, it will use the patches in the `spec.installConfigPatchs` field to update the default install config generated by clusterpool. The patches will be applied in the order listed in the inventory. It will also update the status with reference to the cluster deployment that is using it and update the `Available` condition to false.
```
2uasimojo, line 92:
```suggestion
- For the first ClusterDeploymentCustomization that is available, it will use the patches in the `spec.installConfigPatches` field to update the default install config generated by clusterpool. The patches will be applied in the order listed in the inventory. It will also update the status with reference to the cluster deployment that is using it and update the `Available` condition to false.
```

2uasimojo (commented):
I'm happy with this. A couple of nits inline that could be fixed now or in a followup.

(Now or) in a followup, I would like to add a couple of sections:

### Future
Talk about how we've left this open to extending `ClusterDeploymentCustomization` with `clusterDeploymentSpecPatches` etc. as necessary.

### Alternatives
Include some variant of the content of [this comment](https://issues.redhat.com/browse/HIVE-1367?focusedCommentId=19220630&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19220630) to explain why/how we landed on this approach.

/lgtm
/hold for Alex's review, or in case you want to address any of the above in this PR.

docs/enhancements/clusterpool-inventory.md
96cub0fs, line 76: I don't know how it was addressed in other cases, but assuming that a ClusterDeploymentCustomization is an object that will exist for a long time, might be used by number of inventories and will be checked regularly for availability, wouldn't the number of such log entries grow repidily and might become unusable?
Maybe keep such log entries in other resources such as CloudDeployment instance for availability check,  and validation in ClusterPool?

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,186 @@
...
+  installConfigPatches:
+    - op: replace
+      path: metadata/name
+      value: foo
+status:
+  clusterDeploymentRef: 
+    name: foo
+    namespace: foo-namespace
+  conditions:
+    - lastProbeTime: "2020-11-05T14:49:26Z"
```
abraverm, line 72: I don't know how it was addressed in other cases, but assuming that a ClusterDeploymentCustomization is an object that will exist for a long time, might be used by number of inventories and will be checked regularly for availability, wouldn't the number of such log entries grow repidily and might become unusable?
Maybe keep such log entries in other resources such as CloudDeployment instance for availability check, and validation in ClusterPool?

docs/enhancements/clusterpool-inventory.md
```diff
@@ -0,0 +1,186 @@
...
+  ```
+- Add the name of ClusterDeploymentCustomization CR to`clusterPool.spec.inventory.ClusterDeploymentCustomizations` list. For ClusterDeploymentCustomization with a name `foo-cluster-deployment-customization` the clusterpool should be configured as follows
+    ```yaml
+    spec:
+      inventory:
+        ClusterDeploymentCustomizations:
+          - name: foo-cluster-deployment-customization
+    ```
+  
+### Validation
```
abraverm, line 125: What happens in case that none of the entries are valid?
If one valid entry exists then it should update Clusterpool somehow, which field?

docs/enhancements/clusterpool-inventory.md
2uasimojo: This isn't a log -- each condition on a resource is identified by its `type` such that there's only ever one condition with that `type`. So this just represents the _current_ availability status of this ClusterDeploymentCustomization.

docs/enhancements/clusterpool-inventory.md
2uasimojo: L131-2 mention using a status condition on the ClusterPool to indicate when any of its ClusterDeploymentCustomizations are missing or invalid. To be more specific (and maybe we should go ahead and add this to the design @akhil-rane) perhaps that condition looks like:

```yaml
    - lastProbeTime: "2020-11-05T14:49:26Z"
      lastTransitionTime: "2020-11-05T14:49:26Z"
      message: "4 out of 7 ClusterDeploymentCustomizations are valid. Missing: foo, bar. Invalid: baz."
      reason: SomeMissingSomeInvalid
      status: "False"
      type: InventoryValid
```

If all entries are missing or invalid, that's just an edge case of "a subset are missing/invalid". Generally:
- The total number of clusters (claimed & unclaimed) will be min(MaxSize, NumValidInventoryEntries)
- The number of unclaimed clusters will be min(Size, MaxSize, ClaimedClusters - NumValidInventoryEntries)
So for the edge case where NumValidInventoryEntries is 0, the pool will simply not be able to produce any clusters.

akhil-rane: Added the status condition change on L92

2uasimojo:
> (Now or) in a followup, I would like to add a couple of sections:

https://github.com/openshift/hive/pull/1636

PR #1609, opened 2021-11-11
HiveConfig.Spec.FailedProvisionConfig.RetryReasons

2uasimojo:
Add support for a list of strings in HiveConfig which correspond to
`installFailingReason`s from [additional-]install-log-regexes.

- If the list is present and the most recent ClusterProvision has failed
with a matching reason, we will retry (assuming not otherwise blocked,
e.g. by `InstallAttemptsLimit`).
- If the list is present and the reason is _not_ in the list, we will
fail immediately (`ProvisionStopped`) even if `InstallAttemptsLimit` is
not reached.
- If the list is absent, the previous behavior is observed: we will
ignore the failure reason and proceed or stop based solely on factors
such as `InstallAttemptsLimit`.

[HIVE-1662](https://issues.redhat.com/browse/HIVE-1662)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterprovisions.go
+++ b/pkg/controller/clusterdeployment/clusterprovisions.go
@@ -44,16 +44,22 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		return reconcile.Result{}, err
 	}
 
-	for _, provision := range existingProvisions {
-		if provision.Spec.Stage != hivev1.ClusterProvisionStageFailed {
-			return reconcile.Result{}, r.adoptProvision(cd, provision, logger)
+	var lastFailedProvision *hivev1.ClusterProvision
+	for _, lastFailedProvision = range existingProvisions {
+		if lastFailedProvision.Spec.Stage != hivev1.ClusterProvisionStageFailed {
+			return reconcile.Result{}, r.adoptProvision(cd, lastFailedProvision, logger)
 		}
 	}
+	// If we get here, `lastFailedProvision` is either nil (no provisions exist yet) or points to
+	// the most recent (because `existingProvisions` is sorted by age) failed (because otherwise we
+	// went to adoptProvision) ClusterProvision.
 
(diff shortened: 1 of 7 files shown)


abutcher:
Looks good to me. UT covers everything I can think of. :surfer: 
/lgtm

pkg/controller/clusterdeployment/clusterprovisions.go
```diff
@@ -220,6 +234,39 @@ func (r *ReconcileClusterDeployment) startNewProvision(
...
+func (r *ReconcileClusterDeployment) shouldRetryBasedOnFailureReason(prov *hivev1.ClusterProvision, logger log.FieldLogger) (bool, error) {
+	// prov will be nil if there have been no failed provisions yet
+	if prov == nil {
+		logger.Debug("no failed provisions yet -- allowing retry")
+		return true, nil
+	}
+	// Load up HiveConfig
+	hc := &hivev1.HiveConfig{}
+	if err := r.Get(context.TODO(), types.NamespacedName{Name: constants.HiveConfigName}, hc); err != nil {
+		return false, err // caller logs
```
joelddiaz, line 246: Is this a common pattern in this repo? I was under the impression that we let the hive-operator read HiveConfig and pass any desired configuration into the hive-controllers Pod, but that hive-controllers wouldn't be reading the HiveConfig directly.

joelddiaz: depending on the info being passed, it can be as simple as env vars, or a dedicated cli parameter for the hive-controller binary, or as complex as passing in a configmap. for example the managedns setup uses a ConfigMap https://github.com/openshift/hive/blob/master/pkg/operator/hive/manageddns.go

2uasimojo: I guess in this case a configmap is appropriate since it's complex data (a list of strings).

PR #1611, opened 2021-11-12
move the install log regex for loadbalancer creation failure to its own condition

gregsheremeta:
move the install log regex for loadbalancer creation failure to its own condition

When this happens, the problem is slightly different than just a plain ol'
permission denied, and we need to call it out as such.

/assign @mrbarge @2uasimojo 
/cc @abutcher


2uasimojo (commented):
/lgtm

PR #1612, opened 2021-11-16
clarify when the ProvisionStopped condition would be set

joelddiaz:
It is meant to indicate that at least one cluster provision was
attempted.

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -369,7 +369,9 @@ const (
 	// DeprovisionLaunchErrorCondition is set when a cluster deprovision fails to launch.
 	DeprovisionLaunchErrorCondition ClusterDeploymentConditionType = "DeprovisionLaunchError"
 
-	// ProvisionStoppedCondition is set when cluster provisioning is stopped
+	// ProvisionStoppedCondition is set when cluster provisioning is stopped.
+	// This indicates that at least one provision attempt was made, but there will be no further
+	// retries (without InstallAttemptsLimit changes or other hive configuration stopping further retries).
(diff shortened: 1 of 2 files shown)


2uasimojo:
Pretty sure this will need a `make update` to regenerate the CRDs etc.

Should we also say that ProvisionStopped means we've run out of retries?

joelddiaz:
While the Conditions are part of the API in a roundabout way, they aren't part of the CRD. At least that's what I'm telling myself why there were no other changes after running 'make update'. I've added additional text that hopefully makes things more clear.

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -369,7 +369,9 @@ const (
 	// DeprovisionLaunchErrorCondition is set when a cluster deprovision fails to launch.
 	DeprovisionLaunchErrorCondition ClusterDeploymentConditionType = "DeprovisionLaunchError"
 
-	// ProvisionStoppedCondition is set when cluster provisioning is stopped
+	// ProvisionStoppedCondition is set when cluster provisioning is stopped.
+	// This indicates that at least one provision attempt was made, but there will be no further
+	// retries (without maxRetries changes or other hive configuration stopping futher retries).
```
2uasimojo, line 374: There's no such thing as `maxRetries` afaict. Should this refer to `InstallAttemptsLimit`?

Also typo "futher".
```suggestion
	// retries (without changing InstallAttemptsLimit or other hive configuration preventing further retries).
```

2uasimojo:
> While the Conditions are part of the API in a roundabout way, they aren't part of the CRD

Oh, duh, right.

PR #1613, opened 2021-11-17
Static name for managed DNS ConfigMap

2uasimojo:
When managed DNS is enabled, we copy the relevant data from HiveConfig
to a ConfigMap in the target namespace. Previously this ConfigMap was
dynamically named `managed-domains-XXXXX` and we would find it via a
label. For consistency with other similar ConfigMaps used by Hive, this
commit switches it to use a static name, `hive-managed-domains`,
instead. For upgrade purposes, we have to keep the old lookup mechanism
in place until we can be sure we've flushed out all instances of the old
naming scheme.

### DIFF
--- a/pkg/operator/hive/manageddns.go
+++ b/pkg/operator/hive/manageddns.go
@@ -18,15 +18,16 @@ import (
 )
 
 const (
-	managedDomainsConfigMapNamePrefix = "managed-domains-"
-	managedDomainsConfigMapKey        = "managed-domains"
-	configMapLabel                    = "managed-domains"
-	configMapMountPath                = "/data/config"
+	managedDomainsConfigMapName = "hive-managed-domains"
+	managedDomainsConfigMapKey  = "managed-domains"
+	configMapLabel              = "managed-domains"
+	configMapMountPath          = "/data/config"
 )
 
 // configureManagedDomains will create a new configmap holding the managed domains settings (if necessary), or simply
-// return the current configmap of the current deployment if the settings it contains match the desired settings.
-func (r *ReconcileHiveConfig) configureManagedDomains(logger log.FieldLogger, h resource.Helper, instance *hivev1.HiveConfig, namespacesToClean []string) (*corev1.ConfigMap, error) {
+// use the current configmap of the current deployment if the settings it contains match the desired settings. The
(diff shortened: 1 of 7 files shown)


2uasimojo:
/hold

Oh. I just figured out a problem dynamic naming was solving. The configmap is mounted into the controller pod. If the pod config doesn't change, we don't redeploy. By changing the name of the configmap when we change the config, we ensure the pod config changes, thereby redeploying it and picking up the changes.

There are other ways to solve this. But solve it I must.

2uasimojo:
/hold cancel

Now with a precursor refactor commit cause I just couldn't stand the duplication.

pkg/operator/hive/manageddns.go
```diff
@@ -18,36 +18,40 @@ import (
...
 
 	domains, err := json.Marshal(instance.Spec.ManagedDomains)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to marshal managed domains list into the configmap")
+		return "", errors.Wrap(err, "failed to marshal managed domains list into the configmap")
 	}
 
 	newConfigMapData := map[string]string{managedDomainsConfigMapKey: string(domains)}
 
+	// TODO: The algorithm from here down can be simplified once we're sure we've flushed out any
```
joelddiaz, line 49: Should we make a Jira to follow up on this in a month or two?

joelddiaz (commented):
/lgtm

pkg/operator/hive/manageddns.go
2uasimojo: I was thinking about that.

I don't think we ever really _know_ we've flushed out the old configmaps, so arguably it's _never_ safe to do this.

However, if we refactor this code accordingly and it gets deployed in such an env (e.g. upgrade from ACM 2.4 to 2.5?) the worst thing that happens is we leave a stale, unused configmap lying around. It doesn't have any sensitive data, so meh.

Thoughts?

joelddiaz: I'm fine doing a best-effort migration. I forgot about ACM, so perhaps we should put the Jira card out closer to 6 months into the future to make sure ACM picks up a release with this cleanup code.

2uasimojo: https://issues.redhat.com/browse/HIVE-1700

PR #1619, opened 2021-12-02
CD.Spec.Provisioning.InstallerImageOverride

2uasimojo:
Add a field to ClusterDeployment.Spec.Provisioning allowing
specification of an image URI to use in place of the installer image
gleaned from the ReleaseImage metadata.

[HIVE-1698](https://issues.redhat.com/browse/HIVE-1698)

### DIFF
--- a/pkg/imageset/updateinstaller.go
+++ b/pkg/imageset/updateinstaller.go
@@ -168,18 +168,26 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
 		return errors.Wrap(err, "unrecognized image-references in release payload")
 	}
 
-	installerTagName := "installer"
-	// If this is a bare metal install, we need to get the openshift-install binary from a different image with
-	// bare metal functionality compiled in. The binary is named the same and in the same location, so after swapping
-	// out what image to get it from, we can proceed with the code as we normally would.
-	if cd.Spec.Platform.BareMetal != nil {
-		installerTagName = "baremetal-installer"
+	var installerImage string
+	// There should be no way we get here with Provisioning == nil, but better safe.
+	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallerImageOverride != "" {
+		installerImage = cd.Spec.Provisioning.InstallerImageOverride
+		o.log.WithField("installerImage", installerImage).Info("installer image overridden")
+	} else {
+		// Glean the installer image from the release metadata
+		installerTagName := "installer"
+		// If this is a bare metal install, we need to get the openshift-install binary from a different image with
+		// bare metal functionality compiled in. The binary is named the same and in the same location, so after swapping
+		// out what image to get it from, we can proceed with the code as we normally would.
+		if cd.Spec.Platform.BareMetal != nil {
+			installerTagName = "baremetal-installer"
+		}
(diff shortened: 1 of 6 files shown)


pkg/imageset/updateinstaller.go
```diff
@@ -168,18 +168,26 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
...
+	} else {
+		// Glean the installer image from the release metadata
+		installerTagName := "installer"
+		// If this is a bare metal install, we need to get the openshift-install binary from a different image with
+		// bare metal functionality compiled in. The binary is named the same and in the same location, so after swapping
+		// out what image to get it from, we can proceed with the code as we normally would.
+		if cd.Spec.Platform.BareMetal != nil {
+			installerTagName = "baremetal-installer"
+		}
+		installerImage, err = findImageSpec(is, installerTagName)
```
2uasimojo, line 185: Note to reviewers: This `else` branch comprises the previous code (except for a colon on this line -- `installerImage` is now declared on L171).

pkg/imageset/updateinstaller_test.go
```diff
@@ -148,14 +158,25 @@ func testClusterDeploymentWithErrorCondition() *hivev1.ClusterDeployment {
...
+	return cd
+}
+
+func validateSuccessfulExecution(expectedInstallerImage string) func(*testing.T, *hivev1.ClusterDeployment) {
+	return func(t *testing.T, clusterDeployment *hivev1.ClusterDeployment) {
+		if clusterDeployment.Status.InstallerImage == nil {
+			t.Error("did not get an installer image in status")
+		} else if actual := *clusterDeployment.Status.InstallerImage; actual != expectedInstallerImage {
+			t.Errorf("did not get expected installer image in status: got %s, expected %s", actual, expectedInstallerImage)
+		}
```
joelddiaz, line 175: Can we just do some kind of require/assert.NotNil() for the clusterDeployment.Status.InstallerImage == nil check, and and equivalent assert/require for the clusteDeployment.Status.InstallerImage check?

pkg/imageset/updateinstaller_test.go
```diff
@@ -148,14 +158,25 @@ func testClusterDeploymentWithErrorCondition() *hivev1.ClusterDeployment {
...
+		condition := controllerutils.FindClusterDeploymentCondition(clusterDeployment.Status.Conditions, hivev1.InstallerImageResolutionFailedCondition)
+		if condition != nil && condition.Status != corev1.ConditionFalse {
```
joelddiaz, line 177: Similar comment about using assert/require for the checks.

pkg/imageset/updateinstaller_test.go
2uasimojo: That would be my preference, but I was conforming to how the existing code in this file was doing things. I can make a followon PR to cut the whole test suite over to use testify if ya like.

joelddiaz: works for me

2uasimojo: https://github.com/openshift/hive/pull/1622/commits/83d20afc84764395899904583a81baea761c97a9

PR #1621, opened 2021-12-02
e2e: Increase node wait timeout

2uasimojo:
Been seeing frequent flakes on this. Increasing the timeout,
understanding that it may be a deeper flake and we're just wasting more
time.


joelddiaz (commented):
/lgtm

PR #1627, opened 2021-12-06
add install error

katherinelc321:
for when there is not enough AWS quota


2uasimojo:
Duplicate of #1521. For the latest, see [this comment](https://github.com/openshift/hive/pull/1521#issuecomment-975697317).

Jason said he wasn't going to be able to look at it for a while, though, so perhaps you want to either take over that PR or make this one look like it. Clearly this is something OSD wants :)

PR #1628, opened 2021-12-07
OSD-9064 add install log handlers for proxy install failures

mrbarge:
This PR adds two new installation log handlers, `ProxyTimeout`, and  `ProxyInvalidCABundle`, to cover cases where a cluster installation using a cluster-wide proxy is impacted due to the proxy either being unavailable or serving untrusted certificates.

Refs [OSD-9064](https://issues.redhat.com/browse/OSD-9064) / [SDE-1471](https://issues.redhat.com/browse/SDE-1471)

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -131,6 +131,7 @@ data:
       installFailingReason: UserInitiatedShutdown
       installFailingMessage: User initiated shutdown of instances as the install was running
 
+
     # GCP Specific
     - name: GCPInvalidProjectID
       searchRegexStrings:
@@ -163,6 +164,7 @@ data:
       installFailingReason: GCPServiceAccountQuotaExceeded
       installFailingMessage: GCP Service Account quota exceeded
 
+    
     # Bare Metal
     - name: LibvirtSSHKeyPermissionDenied
       searchRegexStrings:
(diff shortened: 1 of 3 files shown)


gregsheremeta:
seems good to me

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -175,6 +177,20 @@ data:
       installFailingReason: LibvirtConnectionFailed
       installFailingMessage: "Could not connect to libvirt host"
 
+
+    # Proxy-enabled clusters
+    - name: ProxyTimeout
```
2uasimojo, line 182: Is this one [retryable](https://gitlab.cee.redhat.com/service/app-interface/-/blob/7b33783eb34219c03b28c2cbccbb1b29d5df2a87/resources/services/hive/integration/hive.hiveconfig.yaml#L27-30)?

2uasimojo (commented):
/lgtm

config/configmaps/install-log-regexes-configmap.yaml
gregsheremeta: I'll let @mrbarge answer, but both of those sound like things that need user intervention (so, no, not retryable)

mrbarge: For now I'd agree, if it's timing out, it's likely either a VPC configuration is wrong, or the wrong values have been supplied at install time, so likely not retryable. We can let our future experience be the judge on if that's correct or not. :) 

Thanks for the review folks!

PR #1629, opened 2021-12-09
feat: clusterprovision: exclude aws throttling

boranx:


joelddiaz (commented):
Is this really a "throttling" event? I'm used to seeing error messages with "Rate exceeded" (or something similar. This error sounds like a generic "route53 DNS resolution is taking a while" type of error.

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -120,6 +120,11 @@ data:
       - "current credentials insufficient for performing cluster installation"
       installFailingReason: AWSInsufficientPermissions
       installFailingMessage: AWS credentials are insufficient for performing cluster installation
+    - name: AWSThrottlingIssues
+      searchRegexStrings:
+      - "error waiting for Route53 Hosted Zone"
+      installFailingReason: AWSThrottlingIssues
+      installFailingMessage: AWS is having throttling issues, please retry your installation
```
gregsheremeta, line 127: I'd move this to line 86 (after Generic Throttling) and reuse the same Reason

```suggestion
    - name: AWSRoute53Throttling
      searchRegexStrings:
      - "error waiting for Route53 Hosted Zone .* creation: timeout while waiting for state to become 'INSYNC'"
      installFailingReason: AWSAPIRateLimitExceeded
      installFailingMessage: AWS API rate limit exceeded waiting for Route53
```

gregsheremeta: i made the regexp a little more aggressive too

joelddiaz:
I'm still confused about calling a log message that matches this regex a "throttling" event. This looks like DNS is just taking a while to settle down, not that AWS is throttling API calls. Can someone please explain how "error waiting for Route53 Hosted Zone .* creation: timeout while waiting for state to become 'INSYNC'" is a throttling event?

gregsheremeta:
according to @staebler (June 17) known issue that aws throttling causes this message

joelddiaz:
Looking at the terraform code that appeart to me to be "timing out" https://github.com/openshift/installer/blob/master/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/resource/state.go#L37-L51 , I wouldn't describe this as a "throttling" event. To me, this looks more like: "Route53 is taking longer than we're (terraform is) willing to wait".

Based on what I see in the code, I wouldn't describe this as "AWS is having throttling issues" because we are not making an excessive number of AWS API calls and getting slowed down (by AWS) as a result of too many requests.

If someone has some information that explains why this timeout is due to an underlying throttling event, then I suppose we can go ahead and bubble this up as a throttling event, but please put some comments next to the regex about why a "error waiting for ..." error is being interpreted as a throttling error.

staebler:
> Looking at the terraform code that appeart to me to be "timing out" https://github.com/openshift/installer/blob/master/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/resource/state.go#L37-L51 , I wouldn't describe this as a "throttling" event. To me, this looks more like: "Route53 is taking longer than we're (terraform is) willing to wait".
> 
> Based on what I see in the code, I wouldn't describe this as "AWS is having throttling issues" because we are not making an excessive number of AWS API calls and getting slowed down (by AWS) as a result of too many requests.
> 
> If someone has some information that explains why this timeout is due to an underlying throttling event, then I suppose we can go ahead and bubble this up as a throttling event, but please put some comments next to the regex about why a "error waiting for ..." error is being interpreted as a throttling error.

This issue is caused by AWS throttling the CreateHostedZone request. The terraform provider is not properly handling the throttling response and gets stuck in a state where it does not retry the request. Eventually, the terraform provider times out claiming that it is waiting for the hosted zone to be INSYNC.

joelddiaz:
> This issue is caused by AWS throttling the CreateHostedZone request. The terraform provider is not properly handling the throttling response and gets stuck in a state where it does not retry the request. Eventually, the terraform provider times out claiming that it is waiting for the hosted zone to be INSYNC.
Perfect. Thanks!  ^^^ is the kind of comment we should put next to that regex explaining that terraform is hiding the underlying throttling.

joelddiaz:
based on a chat with @staebler where he said "... I would expect the overwhelming majority of INSYNC timeouts to be throttling, however."

Can we tighten up the regex to watch for the INSYNC part of the error message so that we can more clearly differentiate the throttling issue vs a "Route53 is being slow" issue?

gregsheremeta:
closing in favor of #1638

PR #1630, opened 2021-12-09
HIVE-1696: machinepool support for alternative Azure image plans

abutcher:
Adds `pool.Spec.Platform.Azure.Image` to `hivev1.MachinePool` which is set for generated `MachineSets` within `ms.Spec.Template.Spec.ProviderSpec` before `MachineSets` are sent to the managed cluster.


pkg/controller/machinepool/azureactuator.go
```diff
@@ -95,6 +98,24 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
...
+	// Use image set for machinepool azure platform instead of the default
+	if image := pool.Spec.Platform.Azure.Image; (machineapi.Image{}) != image {
+		for _, ms := range installerMachineSets {
+			provider := ms.Spec.Template.Spec.ProviderSpec.Value.Object.(*azureprovider.AzureMachineProviderSpec)
+			provider.Image = azureprovider.Image{
+				Publisher:  image.Publisher,
+				Offer:      image.Offer,
+				SKU:        image.SKU,
+				Version:    image.Version,
+				ResourceID: image.ResourceID,
```
abutcher, line 111: Noting that (OpenShift Machine API) `Image.Type` is not passed on to the Azure Provider Image. These are the only fields supported by the Azure Provider.

config/crds/hive.openshift.io_selectorsyncidentityproviders.yaml
```diff
@@ -495,13 +495,28 @@ spec:
...
+                                provider to OpenShift for the user. If multiple claims
+                                are specified, the first one with a non-empty value
+                                is used.
+                              items:
+                                description: OpenIDClaim represents a claim retrieved
+                                  from an OpenID provider's tokens or userInfo responses
+                                minLength: 1
+                                type: string
+                              type: array
+                              x-kubernetes-list-type: atomic
```
joelddiaz, line 511: What are these selectorsyncset changes all about?

pkg/controller/machinepool/azureactuator.go
```diff
@@ -95,6 +98,24 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
 		workerRole,
 		workerUserDataName,
 	)
+
+	// Use image set for machinepool azure platform instead of the default
+	if image := pool.Spec.Platform.Azure.Image; (machineapi.Image{}) != image {
+		for _, ms := range installerMachineSets {
+			provider := ms.Spec.Template.Spec.ProviderSpec.Value.Object.(*azureprovider.AzureMachineProviderSpec)
+			provider.Image = azureprovider.Image{
```
joelddiaz, line 106: Do we need to worry about which version of OpenShift we are generating the MachineSets for (ie. if we set this for an old (but still supported) version of OpenShift, what happens?)?

config/crds/hive.openshift.io_selectorsyncidentityproviders.yaml
abutcher: These crd changes were generated after bumping mods in the api submodule.

pkg/controller/machinepool/azureactuator.go
abutcher: That is a good question. We probably do.

pkg/controller/machinepool/azureactuator.go
abutcher: These types are moving into the api repo. I'm going to revendor 4.11 to pick up those changes which include `Image.Type`.

pkg/controller/machinepool/azureactuator.go
abutcher: Discussed that extra fields would not produce an error. The managed cluster would ultimately not use the overridden image once machinesets were sent over and possibly error at that point.

abutcher:
I think this is good to go but may as well wait until we're back.

pkg/controller/machinepool/azureactuator.go
```diff
@@ -100,6 +100,7 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
 	)
 
 	// Use image set for machinepool azure platform instead of the default
+	// TODO: Pass the image to installazure.MachineSets when image able to be provided
```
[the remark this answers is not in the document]
abutcher: Depends on the timeline for the installer work vs aro I guess.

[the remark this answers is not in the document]
abutcher: Here's the installer card for the installer support https://issues.redhat.com/browse/CORS-1824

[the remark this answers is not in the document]
joelddiaz: Here's the current plan until I hear otherwise...I've put CORS-1824 as blocking HIVE-1696 for now, and we can wait for the installer support and update this PR when that lands.

abutcher:
Superseded by #1764

pkg/controller/machinepool/azureactuator.go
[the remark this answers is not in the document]
abutcher: #1764 vendors the changes from from the installer which add `OSImage` to azure platform.

PR #1631, opened 2021-12-10
DNM: Demonstrate multiline install-log-regex

2uasimojo:
If you need to match chunks on separate lines, add `(?s)`, which causes
`.` to match `\n`, which it normally doesn't.


2uasimojo:
Purpose served.

PR #1625, opened 2021-12-04
Document failed provision log upload

2uasimojo:
Document how to configure log upload to an AWS S3 bucket for failed
provisions in two places:
- The docstring for `hiveutil install-manager`
- The `using-hive` markdown document.

[HIVE-1697](https://issues.redhat.com/browse/HIVE-1697)

PR #1635, opened 2021-12-13
Consolidate S3 upload docs

2uasimojo:
When I wrote #1625 / 1566d49 I didn't realize we already had some
documentation in place for provision failure log upload. Remove the
duplicate bits; leave the appropriate bits in the appropriate docs.

[HIVE-1697](https://issues.redhat.com/browse/HIVE-1697)

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,26 +1,21 @@
 # Troubleshooting
 
+- [Troubleshooting](#troubleshooting)
+  - [Cluster Install Failure Logs](#cluster-install-failure-logs)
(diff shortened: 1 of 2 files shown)


2uasimojo:
/assign @abutcher 

Note: I could have consolidated this information to a single doc. I decided against it because I figured you would rather be aware of the feature while setting up ("using hive") than find out about it after something fails (while "troubleshooting") when it's too late. So I left the setup bits in "using hive" and the bits about how to consume the logs in "troubleshooting". Reasonable?

PR #1637, opened 2021-12-13
Add install error regex for AWSEC2QuotaExceeded

csheremeta:
see https://coreos.slack.com/archives/CCX9DB894/p1639325596354000

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -7,6 +7,11 @@ data:
   regexes: |
 
     # AWS Specific
+    - name: AWSEC2QuotaExceeded
+      searchRegexStrings:
+      - "failed to generate asset.*Platform Quota Check.*MissingQuota.*ec2"
+      installFailingReason: AWSEC2QuotaExceeded
+      installFailingMessage: AWS EC2 Quota Exceeded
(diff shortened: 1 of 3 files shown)


csheremeta:
related: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3548

2uasimojo:
- This is a dup of #1521.
- Please add a simple test case (see :arrow_up:)
- Also may want to consider the regex variations suggested there :arrow_up: 
- This will need a `make update` when you're done
- I'd like @gregsheremeta to look before we merge.

2uasimojo:
> * This is a dup of #1521.

...and this is the second dup I've seen (#1627). I'm starting to think y'all want this in. For the love of all that is holy, let's get *one* of these landed.

gregsheremeta:
> this is the second dup I've seen

I like this one the best. The other two are now closed.

gregsheremeta:
looks good to me after a test is added. I like this regexp the most -- it will catch specifically ec2, and it's not more precise than it needs to be

gregsheremeta:
I'm sorry, there is a test. Looks good to me (leaving to @2uasimojo to tag it)

2uasimojo (commented):
/lgtm

thanks all

PR #1639, opened 2021-12-14
add install log regexp for bootstrap failed

gregsheremeta:
/assign @joelddiaz 
/assign @2uasimojo

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -204,6 +204,12 @@ data:
 
 
     # Generic OpenShift Install
+    
+    - name: BootstrapFailed
+      searchRegexStrings:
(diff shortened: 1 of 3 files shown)


joelddiaz (commented):
IIUC, the regex matching happens top-to-bottom. This seems like a very generic error as bootstrap can fail for any number of reasons. Should this be near the bottom of the regex list, or am I missing the bigger picture on how the regex handling works?

gregsheremeta:
Where it's placed now should be good. I wanted it to come after all the cloud-specific things, but before all the other general and fallback problems. All of the other items in the `Generic OpenShift Install` section can only happen post bootstrap succeeding.

2uasimojo (commented):
/lgtm

2uasimojo (commented):
/lgtm

PR #1640, opened 2021-12-14
add install log regexps for terraform inconsistent apply (a known intstaller bug)

gregsheremeta:
/assign @joelddiaz 
/assign @2uasimojo


pkg/controller/clusterprovision/installlogmonitor_test.go
```diff
@@ -47,6 +47,7 @@ const (
...
 	proxyInvalidCABundleLog   = "time=\"2021-08-27T05:56:50Z\" level=info msg=\"[pull.q1w2.quay.rhcloud.com/openshift-release-dev/ocp-release@sha256:7047acb946649cc1f54d98a1c28dd7b487fe91479aa52c13c971ea014a66c8a8: error pinging docker registry pull.q1w2.quay.rhcloud.com: Get \\\"https://pull.q1w2.quay.rhcloud.com/v2/\\\": proxyconnect tcp: x509: certificate signed by unknown authority]): quay.io/openshift-release-dev/ocp-release@sha256:7047acb946649cc1f54d98a1c28dd7b487fe91479aa52c13c971ea014a66c8a8: error pinging docker registry quay.io: Get \\\"https://quay.io/v2/\\\": proxyconnect tcp: x509: certificate signed by unknown authority"
+	inconsistentResult        = "time=\"2021-12-01T16:08:46Z\" level=error msg=\"Error: Provider produced inconsistent result after apply\""
```
2uasimojo, line 50: Can you please add a test case that uses this?

gregsheremeta: oops! done

2uasimojo:
Since this needs a rebase anyway...

> (a known installer bug)

Should we include a comment pointing to the bug? Presumably once the bug gets resolved (and vendored in?) OSD would no longer want to suppress this error if it pops up?

gregsheremeta:
> OSD would no longer want to suppress this error if it pops up?
My plan is to make this one of the retryable conditions. If installer fixes the bug, I still think I'd want to retry if we ever saw it again. (I'd also want to alert on it somehow, but that's not Hive's concern.)

> Should we include a comment pointing to the bug?
Sure. I'll have to track it down though.

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -199,6 +199,11 @@ data:
 
 
     # Generic OpenShift Install
+    - name: InconsistentResult
+      searchRegexStrings:
+      - "Error: Provider produced inconsistent result after apply"
+      installFailingReason: InconsistentResult
```
joelddiaz, line 205: InconsistentResult sounds a little generic. Can we call it InconsistentTerraformResult?

gregsheremeta: done

gregsheremeta:
> I'll add it when I do the rebase

:heavy_check_mark:

PR #1642, opened 2021-12-16
Fix some bugs identified by code inspection

2uasimojo:
- A missed error return in the Reconcile loop if Delete() fails.
- Incomplete error path coverage in the unit test driver.

Adding unit test for the former entailed beefing up the fake controller-runtime client to allow overriding behavior for specific calls without mocking the whole thing. For this I stole/adapted a [solution](https://github.com/openshift/aws-efs-operator/blob/043676e0d01fc41279715ad740145c7ab22a2379/pkg/test/client.go) from 2020 Eric.

### DIFF
--- a/pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
+++ b/pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
@@ -44,10 +44,15 @@ func TestClusterDeprovisionReconcile(t *testing.T) {
 	apis.AddToScheme(scheme.Scheme)
 
 	tests := []struct {
-		name                           string
-		deployment                     *hivev1.ClusterDeployment
-		deprovision                    *hivev1.ClusterDeprovision
-		mockGetCallerIdentity          bool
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterdeprovision/clusterdeprovision_controller_test.go
```diff
@@ -340,6 +317,9 @@ func TestClusterDeprovisionReconcile(t *testing.T) {
 			if !test.expectErr && err != nil {
 				t.Errorf("Unexpected error: %v", err)
 			}
+			if test.expectErr && err == nil {
+				t.Error("Expected error but didn't get one")
+			}
```
akhil-rane, line 322: nit
```suggestion
			if test.expectErr {
				assert.NotNil(t, err, "Expected error but didn't get one")
			} else {
				assert.Nil(t, err, "Unexpected error: %v", err)
			}
			
```

akhil-rane (commented):
/lgtm

PR #1643, opened 2021-12-16
Stop assuming containers[0]

2uasimojo:
As hive-operator loads and modifies the manifests for the other
controllers, it used to assume that they only had one container. This is
not future-proof: if we added another container, it would be easy to
miss that we're now modifying the wrong one. Guard against this by
instead looking up the container by name.

There should be no functional change.


2uasimojo:
/hold

This is currently stacked on #1616. Rebase once that merges. Meanwhile, review only the second commit.

2uasimojo:
Live test ✓s out

joelddiaz (commented):
:+1: 
i think this just needs a rebase since it appears #1616 has merged

2uasimojo:
> i think this just needs a rebase since it appears #1616 has merged

Done.

/hold cancel

joelddiaz (commented):
/lgtm

PR #1644, opened 2021-12-23
hiveconfig additionalCertificateAuthoritiesSecretRef doc

joelddiaz:
mention that the Secret(s) referenced in
hiveConfig.spec.additionalCertificateAuthoritiesSecretRef[] must hold
the cert data in the key named 'ca.crt'.


joelddiaz:
/assign @akhil-rane 
From the BZ discussion in https://bugzilla.redhat.com/show_bug.cgi?id=2005759, I had to go digging through the code to figure out what exactly the Secret format should be for the additionalCertificateAuthoritiesSecretRef Secrets https://github.com/openshift/hive/blob/master/pkg/operator/hive/hive.go#L363 . Help out future versions of me by clarifying that detail in the API comments.

akhil-rane (commented):
/lgtm

PR #1645, opened 2022-01-04
HIVE-1465: VMware x509 error is not reflected in ClusterDeployment

abutcher:
Our previous test returned a nil value for the error received from validating credentials which made it seem like we would always set an auth failure condition in the code we were testing. We actually just logged the error and returned. The auth condition will now be set and a reconcile error will be returned after (and when) the condition has been set.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -665,14 +665,9 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return *result, err
 	}
 
-	// Sanity check the platform/cloud credentials.
-	validCreds, err := r.validatePlatformCreds(cd, cdLog)
-	if err != nil {
-		cdLog.WithError(err).Error("unable to validate platform credentials")
-		return reconcile.Result{}, err
-	}
-	// Make sure the condition is set properly.
-	_, err = r.setAuthenticationFailure(cd, validCreds, cdLog)
+	// Sanity check the platform/cloud credentials and set hivev1.AuthenticationFailureClusterDeploymentCondition
+	validCreds, authError := r.validatePlatformCreds(cd, cdLog)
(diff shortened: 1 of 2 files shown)


abutcher:
WIP
/hold

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -665,14 +665,9 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
-	if err != nil {
-		cdLog.WithError(err).Error("unable to validate platform credentials")
-		return reconcile.Result{}, err
-	}
-	// Make sure the condition is set properly.
-	_, err = r.setAuthenticationFailure(cd, validCreds, cdLog)
+	// Sanity check the platform/cloud credentials and set hivev1.AuthenticationFailureClusterDeploymentCondition
+	validCreds, authError := r.validatePlatformCreds(cd, cdLog)
+	_, err := r.setAuthenticationFailure(cd, validCreds, authError, cdLog)
 	if err != nil {
```
2uasimojo, line 671: If the status update succeeds, but `authError` is not `nil`, I think you want to `return reconcile.Result{}, authError`.

abutcher: That happens right below this block of code. We were always skipping it before.

abutcher: It is based on the condition having been set but could check `authError` instead.

2uasimojo: Ah, I see it now. Yeah, that reads kind of weird, but :man_shrugging:

ldpliu:
@abutcher Could you help to cherry-pick this pr to ocm-2.4 branch ? 
As this pr will fix an ACM 2.4 issue: https://github.com/stolostron/backlog/issues/10327

abutcher:
@ldpliu Sure!

PR #1648, opened 2022-01-10
Add metrics for transitioned clusters

suhanime:
Add cumulative metrics for transitioned clusters.
To keep cardinality in check, labels are cloud platform, cluster version
(down to patch) and clusterpool name if it belongs to a clusterpool

/assign @2uasimojo

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -351,7 +351,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		if shouldStopMachines(cd, hibernatingCondition) {
 			return r.stopMachines(cd, cdLog)
 		}
-		return r.checkClusterStopped(cd, false, cdLog)
+		return r.checkClusterStopped(cd, readyCondition, false, cdLog)
 	}
 	// If we get here, we're not supposed to be hibernating
 	if isFakeCluster {
@@ -371,7 +371,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	if shouldStartMachines(cd, hibernatingCondition, readyCondition) {
 		return r.startMachines(cd, cdLog)
 	}
-	return r.checkClusterRunning(cd, cdLog, readyCondition)
+	return r.checkClusterRunning(cd, cdLog, readyCondition, hibernatingCondition)
 }
 
 func (r *hibernationReconciler) startMachines(cd *hivev1.ClusterDeployment, logger log.FieldLogger) (reconcile.Result, error) {
@@ -434,7 +434,8 @@ func (r *hibernationReconciler) stopMachines(cd *hivev1.ClusterDeployment, logge
 	return reconcile.Result{}, err
 }
 
-func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment, expectRunning bool, logger log.FieldLogger) (reconcile.Result, error) {
+func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment, readyCondition *hivev1.ClusterDeploymentCondition,
+	expectRunning bool, logger log.FieldLogger) (reconcile.Result, error) {
 	actuator := r.getActuator(cd)
 	if actuator == nil {
 		logger.Warning("No compatible actuator found to check machine status")
@@ -474,12 +475,20 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 		if err := r.updateClusterDeploymentStatus(cd, logger); err != nil {
 			return reconcile.Result{}, err
 		}
+		poolNS, poolName := "", ""
+		if cd.Spec.ClusterPoolRef != nil {
+			poolNS = cd.Spec.ClusterPoolRef.Namespace
+			poolName = cd.Spec.ClusterPoolRef.PoolName
+		}
+		// logging with time since ready condition was set to StoppingOrHibernating state
+		hivemetrics.MetricClusterHibernationTransitionSeconds.WithLabelValues(cd.Labels[constants.VersionMajorMinorPatchLabel],
+			cd.Labels[hivev1.HiveClusterPlatformLabel], poolNS, poolName).Observe(time.Since(readyCondition.LastTransitionTime.Time).Seconds())
 	}
 	return reconcile.Result{}, nil
 }
 
 func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment, logger log.FieldLogger,
-	readyCondition *hivev1.ClusterDeploymentCondition) (reconcile.Result, error) {
+	readyCondition *hivev1.ClusterDeploymentCondition, hibernatingCondition *hivev1.ClusterDeploymentCondition) (reconcile.Result, error) {
 	actuator := r.getActuator(cd)
 	if actuator == nil {
 		logger.Warning("No compatible actuator found to check machine status")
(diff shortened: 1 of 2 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -225,6 +235,25 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+				readyCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterReadyCondition)
+				poolRef := ""
+				if cd.Spec.ClusterPoolRef != nil {
+					poolRef = cd.Spec.ClusterPoolRef.PoolName
+				}
+				if hibernatingCond.Reason == hivev1.ResumingOrRunningHibernationReason &&
+					readyCond.Reason != hivev1.RunningReadyReason {
+					metricClusterPowerStateTransitionSeconds.WithLabelValues(cd.Labels[constants.VersionMajorMinorPatchLabel],
+						getCloudPlatform(cd.Spec.Platform),
+						poolRef).Observe(time.Since(hibernatingCond.LastTransitionTime.Time).Seconds())
```
suhanime, line 250: I have worries regarding the clearing of metric - while it is possible to find and clear it for corresponding clusterpool, what about the other cds?

pkg/controller/metrics/metrics.go
```diff
@@ -540,6 +569,34 @@ func (ca *clusterAccumulator) processCluster(cd *hivev1.ClusterDeployment) {
 	}
 }
 
+func getCloudPlatform(platform hivev1.Platform) string {
```
2uasimojo, line 572: Is it not safe to consult the label `hive.openshift.io/cluster-platform` (`HiveClusterPlatformLabel`) instead?

But if it's not, how about we just move [getClusterPlatform](https://github.com/openshift/hive/blob/374913322b5d94efa5ff0fdcd1a6055df351476f/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L2092-L2111) to a utils package and make it public?

pkg/controller/metrics/metrics.go
```diff
@@ -225,6 +235,25 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
 				}
+
+				// Check if cluster is transitioning
+				hibernatingCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
+				readyCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterReadyCondition)
+				poolRef := ""
+				if cd.Spec.ClusterPoolRef != nil {
+					poolRef = cd.Spec.ClusterPoolRef.PoolName
+				}
+				if hibernatingCond.Reason == hivev1.ResumingOrRunningHibernationReason &&
```
2uasimojo, line 246: This seems an odd way/place to be observing/calculating this metric. If I understand the logic correctly, we're only reporting *during* the power state transition; and we're reporting the time taken *so far*; and we're doing that *every time this loop runs*. So for example if it takes 8m to resume and this loop runs every 2m and the first time it kicks in is 1m into my resume, we would end up reporting *four* intervals (1m, 3m, 5m, 7m) when what we actually wanted was only one (8m).

It would seem to make more sense to report this metric in the hibernation controller, at the exact point where we change the hibernating/ready reason to Hibernating/Running, respectively. (The `time.Since` should be based on the *other* condition's `LastTransitionTime`, as you've done here, but being careful to calculate that *before* the status update.) That way we only report one metric per transition, and it represents the entire time frame.

pkg/controller/metrics/metrics.go
```diff
@@ -225,6 +235,25 @@ func (mc *Calculator) Start(ctx context.Context) error {
 						mcLog.Infof("cleared metric: %v", metricClusterDeploymentSyncsetPaused)
 					}
 				}
+
+				// Check if cluster is transitioning
+				hibernatingCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
+				readyCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterReadyCondition)
+				poolRef := ""
+				if cd.Spec.ClusterPoolRef != nil {
+					poolRef = cd.Spec.ClusterPoolRef.PoolName
```
2uasimojo, line 244: We need to account for the pool's namespace as well as its name. Nothing stopping you from having pools of the same name in different namespaces. For other pool-ish metrics ([example](https://github.com/openshift/hive/blob/374913322b5d94efa5ff0fdcd1a6055df351476f/pkg/controller/clusterpool/metrics.go#L13)) I've used separate labels for the namespace and name -- not sure if that's the best thing, or if we should combine them à la NamespacedName.

pkg/controller/metrics/metrics.go
2uasimojo: We're still [discussing](https://coreos.slack.com/archives/CE3ETN3J8/p1641851588051600) this.

My current thinking is: don't clear anything, because it would be really hard to figure out the set of labels you would have to specify.

This would mean that you could pollute new pool "foo" with metrics from old, deleted pool "foo". IMO that's the lesser evil. Document it in the help text and call it good.

2uasimojo (commented):
Reminder from scrum: we also talked about making the reporting and/or labeling configurable, and "off" by default. (Perhaps "off" means we still gather, but don't use any labels.)

pkg/controller/metrics/metrics.go
suhanime: The intent of this change was for admins to be aware of clusters that take more time to transition than trends suggest (so for situations where clusters are stuck hibernating or resuming) - so I can't visualize how it would work in hibernation controller. Therefore I kept the metric in metrics controller - where we loop over all cds and log similar metrics, but I am open to suggestions.

2uasimojo: Discussed in scrum today. Summary:
- Keep the histogram(s -- one for hibernating, one for resuming) but move their computation to the hibernation controller as suggested, only tracking *completed* state transitions.
- Add two new pairs of metrics -- one for hibernating, one for resuming:
  - A gauge that counts the number of CDs *currently* transitioning
  - A gauge that sums the time spent so far in that transition for all such CDs

  This one needs to be computed here in the async loop. The use case is to divide these two numbers (yielding the "average time in active power state transitions") and get worried if that average starts to increase, indicating that one or more transitions are taking longer than they should.

suhanime:
@joelddiaz @2uasimojo - Added the relevant metrics, however metrics controller accumulates the metrics constantly - so either we need an effective strategy to remove the entries from the relevant maps once those clusters have transitioned, or drop/rethink the metric

suhanime:
/unhold
Dropped custom collector changes to go in as a separate PR

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -474,12 +475,20 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 		if err := r.updateClusterDeploymentStatus(cd, logger); err != nil {
 			return reconcile.Result{}, err
 		}
+		poolNS, poolName := "", ""
+		if cd.Spec.ClusterPoolRef != nil {
+			poolNS = cd.Spec.ClusterPoolRef.Namespace
+			poolName = cd.Spec.ClusterPoolRef.PoolName
+		}
+		// logging with time since ready condition was set to StoppingOrHibernating state
```
2uasimojo, line 483: nts: We don't record this metric for fake clusters, which is fine.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -600,9 +609,19 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+		if err := r.updateClusterDeploymentStatus(cd, logger); err != nil {
+			return reconcile.Result{}, err
+		}
+		poolNS, poolName := "", ""
+		if cd.Spec.ClusterPoolRef != nil {
+			poolNS = cd.Spec.ClusterPoolRef.Namespace
+			poolName = cd.Spec.ClusterPoolRef.PoolName
+		}
+		// logging with time since hibernating condition was set to ResumingOrRunning state
+		hivemetrics.MetricClusterReadyTransitionSeconds.WithLabelValues(cd.Labels[constants.VersionMajorMinorPatchLabel],
```
2uasimojo, line 621: likewise

2uasimojo (commented):
Straightforward and effective. Ramblings below, feel free to ignore.

/lgtm

My one worry is maintainability: if we end up with some other code path that sets one of these states, will we miss updating the metric in that code path? Example, right now each of those state transitions also exists for fake clusters. We're not observing the metric in those cases -- which is appropriate -- but maybe some future thing?

The only way I can think to be defensive about it would be to roll up each of those state transitions (including setting the status.PowerState and posting the metric) into its own function. That would at least increase the likelihood that future-us would notice and consider the metric.

suhanime:
> My one worry is maintainability: if we end up with some other code path that sets one of these states, will we miss updating the metric in that code path? Example, right now each of those state transitions also exists for fake clusters. We're not observing the metric in those cases -- which is appropriate -- but maybe some future thing?

One of the main principles of fake clusters with regards to hibernation is that powerstate change should be immediate - that is also one of the testing criteria for QE. So I think it is safe to assume we won't be recording this metric for fake clusters
And about some other path setting the powerstate - that is exactly what we don't ever want to do - that was why we refactored this controller. We go through too many states, and we need to ensure that powerstates are set (especially the final ones) at one place only
 
> The only way I can think to be defensive about it would be to roll up each of those state transitions (including setting the status.PowerState and posting the metric) into its own function. That would at least increase the likelihood that future-us would notice and consider the metric.

Yes, that would be a good idea, but I'm okay with not having it right now.

2uasimojo:
@gurnben @KevinFCormier This is (the easy) part of the metrics picture we discussed earlier today. It reports transition times for _completed_ transitions to hibernating or running. Which is cool, but it doesn't help you identify that one cluster that's _currently_ stuck resuming.

The other piece (#1712) will allow you to report on durations of in-flight state transitions, notably allowing you to alert on that specific case. (If we can ever get the logic figured out.)

PR #1649, opened 2022-01-12
Update bundle-gen.py to support docker

2uasimojo:
...as well as buildah.


2uasimojo:
/assign @abutcher 

I was not able to test this all the way through, as the docker build is still failing with centos:stream, but it worked up to that point.

Please regression test in your env (with buildah) before merging.

Thanks!

PR #1604, opened 2021-11-01
Check clusteroperators

suhanime:
Add functionality to check for clusteroperators to settle as we resume a hibernating cluster

To accomplish this, following things were added
- cd.Status.PowerState
- ClusterReady condition for cluster deployment, that expands on transitioning status when a cluster is resuming
- Intermediate reasons to reflect starting (and stopping) if machines, distinct from waiting for the machines to start (or stop)

xref: https://issues.redhat.com/browse/HIVE-1601

/cc @2uasimojo

PR #1651, opened 2022-01-14
Use cd.Status.PowerState for ClusterClaim Running condition

2uasimojo:
The ClusterDeployment Hibernating condition used to be the one-stop shop
for determining the target cluster's power state. In #1604 we changed
the semantics such that the Hibernating condition only reports things
about hibernation; a new Ready condition reports things about being
ready; and a new Status.PowerState field reports the cluster's power
state definitively.

The ClusterClaim controller uses the ClusterDeployment's power state to
set the ClusterClaim's ClusterRunning condition. It was previously using
the state of the ClusterDeployment's Hibernating condition to do this.
We missed changing it in #1604 to look at the Ready condition or
Status.PowerState instead. This commit does the latter.

[HIVE-1734](https://issues.redhat.com/browse/HIVE-1734)


joelddiaz (commented):
/lgtm

PR #1652, opened 2022-01-14
ClusterPool: Claim running clusters

2uasimojo:
With this PR, the cdCollection bucket called `assignable` now only contains installed, non-deleted/deleting, *running* clusters. Since we fulfill claims out of that bucket, the result is that, when your claim is assigned a cluster, that cluster is guaranteed to be running. Before this change, it might have been hibernating and we would initiate resume when assigning; so you would need to wait before using it.

To make this work, we also add a new bucket to cdCollection called `standby`. This contains the installed, non-deleted/deleting clusters that are *not* running.

The related metrics are changed accordingly:
- `hive_clusterpool_clusterdeployments_assignable` now only counts running clusters.
- `hive_clusterpool_clusterdeployments_running` is removed, as it would have been redundant with the above.
- `hive_clusterpool_clusterdeployments_standby` is added.

[HIVE-1679](https://issues.redhat.com/browse/HIVE-1679)

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -53,17 +53,50 @@ const (
 
 // ClusterPowerState is used to indicate whether a cluster is running or in a
 // hibernating state.
-// +kubebuilder:validation:Enum="";Running;Hibernating
 type ClusterPowerState string
 
 const (
-	// RunningClusterPowerState is the default state of a cluster after it has
+	// ClusterPowerStateRunning is the default state of a cluster after it has
 	// been installed. All of its machines should be running.
-	RunningClusterPowerState ClusterPowerState = "Running"
+	ClusterPowerStateRunning ClusterPowerState = "Running"
 
-	// HibernatingClusterPowerState is used to stop the machines belonging to a cluster
-	// and move it to a hibernating state.
-	HibernatingClusterPowerState ClusterPowerState = "Hibernating"
+	// ClusterPowerStateHibernating indicates the machines belonging to a cluster
+	// are stopped.
+	ClusterPowerStateHibernating ClusterPowerState = "Hibernating"
+
+	// ClusterPowerStateSyncSetsNotApplied indicates SyncSets have not yet been applied
+	// for the cluster based on ClusterSync.Status.FirstSucessTime
+	ClusterPowerStateSyncSetsNotApplied ClusterPowerState = "SyncSetsNotApplied"
+
+	// ClusterPowerStateStartingMachines is used to reflect attempt to list and start cloud VMs
+	ClusterPowerStateStartingMachines ClusterPowerState = "StartingMachines"
+
+	// ClusterPowerStateFailedToStartMachines
+	ClusterPowerStateFailedToStartMachines ClusterPowerState = "FailedToStartMachines"
+
+	// ClusterPowerStateStopping indicates the cluster is transitioning
+	// from a Running state to a Hibernating state.
+	ClusterPowerStateStopping ClusterPowerState = "Stopping"
+
+	// ClusterPowerStateFailedToStop is used when there was an error stopping machines
+	// to enter hibernation
+	ClusterPowerStateFailedToStop ClusterPowerState = "FailedToStop"
+
+	// ClusterPowerStateWaitingForMachinesToStop is used when waiting for cloud VMs to stop
+	ClusterPowerStateWaitingForMachinesToStop ClusterPowerState = "WaitingForMachinesToStop"
+
+	// ClusterPowerStateWaitingForMachines is used when waiting for cloud VMs to start.
+	ClusterPowerStateWaitingForMachines ClusterPowerState = "WaitingForMachines"
+
+	// ClusterPowerStateWaitingForNodes is used when waiting for nodes to become Ready.
+	ClusterPowerStateWaitingForNodes ClusterPowerState = "WaitingForNodes"
+
+	// ClusterPowerStatePausingForClusterOperatorsToSettle is used when pausing to let ClusterOperators start and post new status before we check it.
+	ClusterPowerStatePausingForClusterOperatorsToSettle ClusterPowerState = "PausingForClusterOperatorsToSettle"
+
+	// ClusterPowerStateWaitingForClusterOperators is used when waiting for ClusterOperators to
+	// get to a good state. (Available=True, Processing=False, Degraded=False)
+	ClusterPowerStateWaitingForClusterOperators ClusterPowerState = "WaitingForClusterOperators"
 )
 
 // ClusterDeploymentSpec defines the desired state of ClusterDeployment
@@ -134,6 +167,7 @@ type ClusterDeploymentSpec struct {
 
 	// PowerState indicates whether a cluster should be running or hibernating. When omitted,
 	// PowerState defaults to the Running state.
+	// +kubebuilder:validation:Enum="";Running;Hibernating
 	// +optional
 	PowerState ClusterPowerState `json:"powerState,omitempty"`
 
@@ -295,7 +329,7 @@ type ClusterDeploymentStatus struct {
 
 	// PowerState indicates the powerstate of cluster
 	// +optional
-	PowerState string `json:"powerState,omitempty"`
+	PowerState ClusterPowerState `json:"powerState,omitempty"`
 
 	// ProvisionRef is a reference to the last ClusterProvision created for the deployment
 	// +optional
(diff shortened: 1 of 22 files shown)


2uasimojo:
/assign @suhanime 

Recommend reviewing commits individually.

/hold for live test

2uasimojo:
Dear Monday Eric-

The computation of `excessSize` needs work. Right now your `runningCount=0` pool isn't satisfying any claims because it's not kicking any CDs to Running because `excessSize` is ending up at zero. I think it's because right now you're calculating `excessSize` based on all the unassigned clusters; whereas it should probably be calculated based only on the running ones. (Consider renaming `excessSize` to `extraRunningNeeded` or similar, since it's only used to affect how many we start.)

Thanks,
Friday Eric
.

2uasimojo:
Fixed.

Live test checks out.

/hold cancel

@suhanime ready for review. Again, recommend reviewing one commit at a time :)

2uasimojo:
> @suhanime ready for review. Again, recommend reviewing one commit at a time :)

Gah, forgot that some of my UT framework changes would affect other tests.

/hold

2uasimojo:
Tabular pool output's ready count isn't right yet.

2uasimojo:
Okay, should be good to go now.

/hold cancel

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -429,19 +415,22 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
...
-	excessCount int,
+	extraRunning int,
 	logger log.FieldLogger,
 ) error {
 	// If we're creating excess clusters to satisfy unassigned claims, add that many
 	// to the runningCount. They'll get snatched up immediately, bringing the number
 	// of running clusters back down to runningCount once the pool reaches steady state.
-	runningCount := int(clp.Spec.RunningCount) + excessCount
-	cdList := append(cds.Assignable(), cds.Installing()...)
+	runningCount := int(clp.Spec.RunningCount) + extraRunning
```
suhanime, line 431: we are changing runningCount = spec.RunningCount + claims.Unassigned - cds.Installing - cds.Assignable - cds.Broken to runningCount = spec.RunningCount + claims.Unassigned
Questions still present about original implementation because number of excessCount was set before broken clusters were deleted but I digress
Consider spec.runningCount=0. Now if claims.Unassigned < cds.Unassigned(false), then we are good - but what happens when claims.Unassigned > cds.Unassigned(false)?
Wait - that wouldn't happen because we have already added those clusters at this point so they would be installing - right?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -429,19 +415,22 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
...
-// set to running, and the remainder are set to hibernating.
+// reconcileRunningClusters ensures the oldest unassigned clusters are set to running, and the
+// remainder are set to hibernating. The number of clusters we set to running is determined by
+// adding the cluster's configured runningCount to the number of unsatisfied claims for which we're
+// spinning up new clusters.
 func (r *ReconcileClusterPool) reconcileRunningClusters(
 	clp *hivev1.ClusterPool,
 	cds *cdCollection,
-	excessCount int,
+	extraRunning int,
```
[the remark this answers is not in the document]
suhanime: And if your math is correct - can we rename it to just call it what it is - unassignedClaims?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -369,7 +352,10 @@ func (r *ReconcileClusterPool) Reconcile(ctx context.Context, request reconcile.
 	}
 	availableCurrent -= toDel
 
-	switch drift := reserveSize - int(clp.Spec.Size); {
+	// drift will indicate how many clusters we need to add or delete to get back to steady state
+	// of the pool's Size. This needs to take into account the clusters we're creating to satisfy
+	// the immediate demand of pending claims.
+	switch drift := len(cds.Unassigned(true)) - int(clp.Spec.Size) - len(claims.Unassigned()); {
```
suhanime, line 358: Question about original implementation - drift is calculated considering broken clusters, but we delete them before entering case of drift==0 - shouldn't drift be recalculated then? Maybe we need to add some clusters?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -741,31 +730,46 @@ func (r *ReconcileClusterPool) createRandomNamespace(clp *hivev1.ClusterPool) (*
 	return ns, err
 }
 
+func getClustersToDelete(cds *cdCollection, deletionsNeeded int, logger log.FieldLogger) []*hivev1.ClusterDeployment {
```
suhanime, line 733: A comment about the logic would be helpful here

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -741,31 +730,46 @@ func (r *ReconcileClusterPool) createRandomNamespace(clp *hivev1.ClusterPool) (*
...
+		return append(clustersToDelete, readyClusters[:deletionsNeeded]...)
+	}
+
+	logger.WithField("deletionsNeeded", origDeletionsNeeded).
+		WithField("installingClusters", len(installingClusters)).
+		WithField("standbyClusters", len(standbyClusters)).
+		WithField("readyClusters", len(readyClusters)).
+		Error("trying to delete more clusters than there are available")
+
+	return append(clustersToDelete, readyClusters...)
```
suhanime, line 762: This would result in error eventually when we try to delete the cd according to the logs above - right? Shouldn't we fail early?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -802,7 +806,7 @@ func (r *ReconcileClusterPool) reconcileDeletedPool(pool *hivev1.ClusterPool, lo
 	if err != nil {
 		return err
 	}
-	for _, cd := range append(cds.Assignable(), cds.Installing()...) {
+	for _, cd := range cds.Unassigned(true) {
```
suhanime, line 809: Why weren't considering broken cds before that we are now?

suhanime (commented):
A few questions to annoy you :/

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: Yeah, as noted above, the previous implementation was a bit sloppy, saved by the fact that we were ensuring the PowerState was set when we did the assignment.

Yes, by this point we have
- Deleted broken clusters. Though by definition none of those would have been running.
- Kicked off creation of new clusters, within `maxConcurrent`, to satisfy capacity and unclaimed clusters. Those also wouldn't be running yet.
- Assigned whatever claims we could. With this PR, those assignments consumed running clusters. Depending how many loops we've needed in non-steady state, that may or may not have reduced the number of running clusters below `runningCount`.

So at this point, we know we want the number of running clusters to be `runningCount` plus enough to accommodate whatever unassigned claims _remain_.

> Consider spec.runningCount=0. Now if claims.Unassigned < cds.Unassigned(false), then we are good - but what happens when claims.Unassigned > cds.Unassigned(false)?

I'm not sure if I've managed to answer this question with my ramblings above. Perhaps you can formulate it as a unit test case? Then we can either see if we already have that case covered, or create it if not.

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: Totally. Will do.

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: No, it shouldn't result in an error. If we get here it just means that the length of the list we return will be _less_ than the number of deletions that were originally requested. The list will simply contain all the unassigned clusters (except broken ones -- they got deleted in a previous reconcile).

pkg/controller/clusterpool/clusterpool_controller.go
2uasimojo: I believe that was likely a latent bug -- this code path was missed when we introduced the concept of broken clusters.

We invoke `reconcileDeletedPool` early in the reconcile, before we get a whack at cleaning up the broken clusters, and the code path that does that never gets any further if the pool is marked for deletion. So I think if you timed the deletion just right, you could have left zombie broken clusters around "forever".

pkg/controller/clusterpool/collections.go
```diff
@@ -190,8 +190,10 @@ func (c *claimCollection) Untrack(claimNames ...string) {
 }
 
 type cdCollection struct {
-	// Unclaimed installed clusters which belong to this pool and are not (marked for) deleting
+	// Unclaimed, installed, running clusters which belong to this pool and are not (marked for) deleting
 	assignable []*hivev1.ClusterDeployment
+	// Unclaimed, installed, non-deleted clusters which are not running and therefore not (yet) assignable
```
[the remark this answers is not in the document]
2uasimojo: ack

2uasimojo (commented):
Thanks for the review @suhanime!

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -429,32 +429,32 @@ var PositivePolarityClusterDeploymentConditions = []ClusterDeploymentConditionTy
...
-	// StoppingHibernationReason is used as the reason when the cluster is transitioning
+	ResumingOrRunningHibernatingReason = "ResumingOrRunning"
+	// StoppingHibernatingReason is used as the reason when the cluster is transitioning
 	// from a Running state to a Hibernating state.
-	StoppingHibernationReason = "Stopping"
+	StoppingHibernatingReason = "Stopping"
 	// WaitingForMachinesToStopHibernatingReason is used on the Hibernating condition when waiting for cloud VMs to stop
 	WaitingForMachinesToStopHibernatingReason = "WaitingForMachinesToStop"
-	// HibernatingHibernationReason is used as the reason when the cluster is in a
+	// HibernatingHibernatingReason is used as the reason when the cluster is in a
```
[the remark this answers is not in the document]
2uasimojo: Discussed in [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1642621718221000), decided to go with {condition name}Reason{value} throughout. Will update.

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -429,32 +429,32 @@ var PositivePolarityClusterDeploymentConditions = []ClusterDeploymentConditionTy
...
 	// Hibernating state.
-	HibernatingHibernationReason = "Hibernating"
-	// UnsupportedHibernationReason is used as the reason when the cluster spec
+	HibernatingHibernatingReason = string(HibernatingClusterPowerState)
```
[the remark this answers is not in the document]
2uasimojo: Discussed in [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1642621224215800) and decided it's okay to make changes to golang data types as long as the underlying json/k8s type doesn't change. Will reevaluate in this light.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -1422,7 +1421,7 @@ func TestReconcileClusterPool(t *testing.T) {
 						Build(testsecret.WithDataKeyValue("dummykey", []byte("dummyval"))),
 				)
 			}
-			fakeClient := fake.NewFakeClientWithScheme(scheme, test.existing...)
+			fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithRuntimeObjects(test.existing...).Build()
```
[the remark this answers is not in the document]
2uasimojo: Yah; I'm kinda hitting it as I touch affected files.

apis/hive/v1/clusterdeployment_types.go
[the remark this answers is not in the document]
2uasimojo: Okay, I redid the first commit, PTAL.

2uasimojo:
/test e2e

node timeout

2uasimojo:
/test e2e

c'mon now

PR #1653, opened 2022-01-14
Add error message if install pods get killed or evicted

akhil-rane:
This PR adds error message in the install log if we detect a
infraID set on cluster provision by previous install pod which
got killed or evicted midway through the installation process.
We also fail the cluster provision in this case. Earlier we use
to start a new install attempt which failed to update cluster
provision infraID - an immutable field giving a not so user
friendly admission webhook failure message.

    
This PR also fixes the bug where secrets associated with the
failed cluster provisions were getting leaked. We now store
the name of the previous provision and use it to delete
leaked secrets in the next provision attempt.

x-ref: https://issues.redhat.com/browse/HIVE-1725

### DIFF
--- a/apis/hive/v1/clusterprovision_types.go
+++ b/apis/hive/v1/clusterprovision_types.go
@@ -44,6 +44,9 @@ type ClusterProvisionSpec struct {
 
 	// PrevInfraID is the infra ID of the previous failed provision attempt.
 	PrevInfraID *string `json:"prevInfraID,omitempty"`
+
+	// PrevProvisionName is the name of the previous failed provision attempt.
+	PrevProvisionName *string `json:"prevProvisionName,omitempty"`
 }
 
 // ClusterProvisionStatus defines the observed state of ClusterProvision.
--- a/pkg/controller/clusterdeployment/clusterprovisions.go
+++ b/pkg/controller/clusterdeployment/clusterprovisions.go
@@ -175,10 +175,11 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		},
 	}
 
-	// Copy over the cluster ID and infra ID from previous provision so that a failed install can be removed.
-	if cd.Spec.ClusterMetadata != nil {
-		provision.Spec.PrevClusterID = &cd.Spec.ClusterMetadata.ClusterID
-		provision.Spec.PrevInfraID = &cd.Spec.ClusterMetadata.InfraID
+	// Copy over the name, cluster ID and infra ID from previous provision so that a failed install can be removed.
+	if lastFailedProvision != nil {
+		provision.Spec.PrevProvisionName = &lastFailedProvision.Name
+		provision.Spec.PrevClusterID = lastFailedProvision.Spec.ClusterID
+		provision.Spec.PrevInfraID = lastFailedProvision.Spec.InfraID
 	}
 
 	logger.WithField("derivedObject", provision.Name).Debug("Setting label on derived object")
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -344,12 +344,32 @@ func (m *InstallManager) Run() error {
 		}
 	}
 
-	// If the cluster provision has an infraID set, this implies we failed an install
-	// and are re-trying. Cleanup any resources that may have been provisioned.
-	m.log.Info("cleaning up from past install attempts")
-	if err := m.cleanupFailedInstall(cd, provision); err != nil {
-		m.log.WithError(err).Error("error while trying to preemptively clean up")
-		return err
+	// If the cluster provision has a prevInfraID set, this implies we failed a previous
+	// cluster provision attempt. Cleanup any resources that may have been provisioned.
+	if provision.Spec.PrevInfraID != nil {
+		m.log.Info("cleaning up resources from previous provision attempt")
+		if err := m.cleanupFailedInstall(cd, *provision.Spec.PrevInfraID, *provision.Spec.PrevProvisionName, m.Namespace); err != nil {
+			m.log.WithError(err).Error("error while trying to preemptively clean up")
+			return err
+		}
+	}
+
+	// If the cluster provision has an infraID set, this implies that previous install
+	// pod for the current cluster provision got unexpectedly killed or evicted. If we
+	// proceed, the current install will error out updating the infraID on the cluster
+	// provision. The infraID is an immutable write-once field, as we need it to clean
+	// up provisioned resources from a failed provision. So if infraID is already set,
+	// we do not proceed further and fail the provision.
+	if provision.Spec.InfraID != nil {
+		// This error won't be displayed as condition message on the cluster provision
+		// because install log is not generated until we run installer binary. We do
+		// have the option of faking an install log that can then be regexed.
+		m.log.Error("infraID is already set on the ClusterProvision. Unexpected install pod restart detected. Cleaning up resources from previous install attempt")
+		if err := m.cleanupFailedInstall(cd, *provision.Spec.InfraID, m.ClusterProvisionName, m.Namespace); err != nil {
+			m.log.WithError(err).Error("error while trying to preemptively clean up")
+			return err
+		}
+		return fmt.Errorf("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")
 	}
 
 	// Generate installer assets we need to modify or upload.
@@ -543,25 +563,16 @@ func (m *InstallManager) copyFile(src, dst string) error {
 }
 
 // cleanupFailedInstall allows recovering from an installation error and allows retries
-func (m *InstallManager) cleanupFailedInstall(cd *hivev1.ClusterDeployment, provision *hivev1.ClusterProvision) error {
-	infraID := provision.Spec.InfraID
-	if infraID == nil {
-		infraID = provision.Spec.PrevInfraID
-	}
-	if infraID != nil {
-		m.log.Info("InfraID set from failed install, running deprovison")
-		if err := m.cleanupFailedProvision(m.DynamicClient, cd, *infraID, m.log); err != nil {
-			return err
-		}
-	} else {
-		m.log.Warn("skipping cleanup as no infra ID set")
-	}
-
-	if err := m.cleanupAdminKubeconfigSecret(); err != nil {
+func (m *InstallManager) cleanupFailedInstall(cd *hivev1.ClusterDeployment, infraID, provisionName, provisionNamespace string) error {
+	if err := m.cleanupFailedProvision(m.DynamicClient, cd, infraID, m.log); err != nil {
 		return err
 	}
 
-	if err := m.cleanupAdminPasswordSecret(); err != nil {
+	if err := m.cleanupAdminKubeconfigSecret(provisionName, provisionNamespace); err != nil {
+		return err
+	}
+
+	if err := m.cleanupAdminPasswordSecret(provisionName, provisionNamespace); err != nil {
 		return err
 	}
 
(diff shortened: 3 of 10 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -351,6 +351,12 @@ func (m *InstallManager) Run() error {
 		m.log.WithError(err).Error("error while trying to preemptively clean up")
 		return err
 	}
+	// Fail the cluster provision if infraID is set. This can occur if previous install
+	// pod for current cluster provision is unexpectedly killed or evicted.
+	if provision.Spec.InfraID != nil {
+		m.log.WithError(err).Error("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")
+		return fmt.Errorf("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")
+	}
```
akhil-rane, line 359: This code will not update `ClusterProvisionFailed` condition with any relevant message about the failure. We have to check install pod logs to know the actual cause. The `ClusterProvisionFailed` gets updated by clusterprovision_controller with reason and message parsed from the install log. The install log is not stored until we run openshift-install binary. One option is we directly update ClusterProvisionFailed condition here. Any thoughts?

ClusterProvisionFailed condition with current code
```yaml
  - lastProbeTime: "2022-01-14T16:38:21Z"
    lastTransitionTime: "2022-01-14T16:38:21Z"
    message: Cluster install failed but installer log was not captured
    reason: UnknownError
    status: "True"
    type: ClusterProvisionFailed
```

pkg/installmanager/installmanager.go
```diff
@@ -351,6 +351,12 @@ func (m *InstallManager) Run() error {
 		m.log.WithError(err).Error("error while trying to preemptively clean up")
 		return err
 	}
+	// Fail the cluster provision if infraID is set. This can occur if previous install
+	// pod for current cluster provision is unexpectedly killed or evicted.
+	if provision.Spec.InfraID != nil {
+		m.log.WithError(err).Error("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")
```
2uasimojo, line 357: I can't tell where this `err` came from. I suspect you just meant to log your message. But I'm not sure you need to do that, as the controller-runtime will log the error (below) anyway. TL;DR: you can probably just remove this line.

pkg/installmanager/installmanager.go
```diff
@@ -351,6 +351,12 @@ func (m *InstallManager) Run() error {
 		m.log.WithError(err).Error("error while trying to preemptively clean up")
 		return err
 	}
+	// Fail the cluster provision if infraID is set. This can occur if previous install
+	// pod for current cluster provision is unexpectedly killed or evicted.
+	if provision.Spec.InfraID != nil {
```
2uasimojo, line 356: `cleanupFailedInstall` (invoked on L350) already has a code path that catches `provision.Spec.InfraID != nil` and cleans up the provision with that InfraID. So this code change says, "and also fail," right?

I don't have a better suggestion, but that was pretty confusing to figure out. I'll take another look next week.

pkg/installmanager/installmanager.go
2uasimojo: What would we update it to? If we don't have the install log, we don't have the install log.

2uasimojo (commented):
I need to zoom out and grok the bigger picture here, but some initial thoughts inline.

pkg/installmanager/installmanager.go
joelddiaz: I had to stare at this for a moment as well. If I read things correctly, prevInfraID is set so that a subsequent provision attempt cleans up a provision that "cleanly" failed (ie provision Pod that didn't suddenly stop/get deleted/was rescheduled to a new node). This is a *good thing*. The previous code also catches the case where an infraID was set, and cleans things up since infraID != nil  is an unexpected situation at this point in the codeflow.

Then this new code is handling the "why on earth is infraID set at this point". The previous code cleaned things up, and this new if clause is making sure we don't bother going forward with a provision, and marks this provision attempt as a "we've recovered from a strange state, and will mark this attempt as an error so that on next attempt things will work out as expected".

With this if clause, we don't unexpectedly error later when we try to update the generated infraID. We mark it as a clean "this install failed" attempt, and a subsequent provision attempt will the the prevInfraID fields, the code will clean up (a no-op since we've just cleaned up in this current provision attempt).

Assuming I have explained this accurately, can we spread some comments around to explain why infraID != nil is unexpected at this point, and how we handle it so that resources don't get leaked, and future provision attempts have a chance at success?

I can't see how infraID != nil AND prevInfraID != nil at the same time, but perhaps a sanity check and a loud error message wouldn't hurt since this code is relying on some subtle interactions and sequences of events.

pkg/installmanager/installmanager.go
akhil-rane: > What would we update it to? If we don't have the install log, we don't have the install log.

I was thinking of calling `setClusterProvisionCondtion` directly from here and set a `ClusterProvisionFailed` conditon along with an appropriate reason/message. Currently `clusterprovision` controller parses the install log to set this condition.

> If we did a `fmt.Println("HEY I FAILED BC INFRAID WAS NOT NIL")` would that be considered the "install log" for this provision attempt?
For this, we need to write stdout content to installer console log and then call a function to read the log and transfer it to clusterprovision. Currently, we start this process only after running the installer binary.

joelddiaz: I now see that the fmt.Println() calls are just so that this Pod's logs show the install output, but that isn't really considered the "install output".
Since we already are writing out to the ClusterProvision, I don't see why we couldn't also set this condition right here where we know a condition occurred. Just need to make sure the clusterprovision controller can gracefully handle the job itself updating those fields.

pkg/installmanager/installmanager.go
akhil-rane: nice catch! I will just remove this line.

pkg/installmanager/installmanager.go
akhil-rane: @joelddiaz Everything you have said here is correct. I will go ahead and add some more details in the comments.

pkg/installmanager/installmanager.go
```diff
@@ -351,6 +351,12 @@ func (m *InstallManager) Run() error {
...
+		m.log.WithError(err).Error("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")
+		return fmt.Errorf("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")
```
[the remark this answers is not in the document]
akhil-rane: [line 357 is going away](https://github.com/openshift/hive/pull/1653#discussion_r785204413) so I will just keep `return fmt.Errorf("infraID is already set on the ClusterProvision. Unexpected install pod restart detected")`

pkg/installmanager/installmanager.go
akhil-rane: > I can't see how infraID != nil AND prevInfraID != nil at the same time

Consider the following case
- cluster provision 1 fails due to a valid error. May be a throttling issue
- controller creates new provision 2 and transfers infraID from previous attempt to field prevInfraID.
- provision 2 install pod 1 gets killed/evicted after updating infraID
- provision 2 install pod 2 will see both infraID != nil and prevInfraID != nil

possible?

akhil-rane: We might have to run 2 separate cleanup runs here

pkg/installmanager/installmanager.go
akhil-rane:
```
time="2022-01-17T19:36:01Z" level=info msg="deleting existing object" installID=xjlzmk2s object=default/arane-05-0-4v524-admin-kubeconfig
time="2022-01-17T19:36:01Z" level=info msg="deleting existing object" installID=xjlzmk2s object=default/arane-05-0-4v524-admin-password
time="2022-01-17T19:36:01Z" level=error msg="infraID is already set on the ClusterProvision. Unexpected install pod restart detected" error="<nil>" installID=xjlzmk2s
time="2022-01-17T19:36:01Z" level=fatal msg="runtime error" error="infraID is already set on the ClusterProvision. Unexpected install pod restart detected"
```

Not controller runtime but returned error will be logged by [this](https://github.com/openshift/hive/blob/588f139c6122c62fd5232e7b47125bad103680fe/pkg/installmanager/installmanager.go#L171-L173) line of code

pkg/installmanager/installmanager.go
akhil-rane: No luck updating the cluster provision condition. The code [here](https://github.com/openshift/hive/blob/588f139c6122c62fd5232e7b47125bad103680fe/pkg/controller/clusterprovision/clusterprovision_controller.go#L383-L396) stomps over whatever condition I set. Since this is a rarely occurring situation we can get away with only an explicit message in the log I feel. Unless someone complains.

pkg/controller/clusterdeployment/clusterprovisions.go
```diff
@@ -175,10 +175,11 @@ func (r *ReconcileClusterDeployment) startNewProvision(
...
-	// Copy over the cluster ID and infra ID from previous provision so that a failed install can be removed.
-	if cd.Spec.ClusterMetadata != nil {
-		provision.Spec.PrevClusterID = &cd.Spec.ClusterMetadata.ClusterID
-		provision.Spec.PrevInfraID = &cd.Spec.ClusterMetadata.InfraID
+	// Copy over the name, cluster ID and infra ID from previous provision so that a failed install can be removed.
+	if lastFailedProvision != nil {
+		provision.Spec.PrevProvisionName = &lastFailedProvision.Name
+		provision.Spec.PrevClusterID = lastFailedProvision.Spec.ClusterID
+		provision.Spec.PrevInfraID = lastFailedProvision.Spec.InfraID
 	}
```
akhil-rane, line 183: I have no clue why we were fetching these details from cluster deployment before. Nothing went wrong during my testing after this change.

apis/hive/v1/clusterprovision_types.go
```diff
@@ -44,6 +44,9 @@ type ClusterProvisionSpec struct {
 
 	// PrevInfraID is the infra ID of the previous failed provision attempt.
 	PrevInfraID *string `json:"prevInfraID,omitempty"`
+
+	// PrevProvisionName is the name of the previous failed provision attempt.
+	PrevProvisionName *string `json:"prevProvisionName,omitempty"`
```
akhil-rane, line 49: I had to add this field to clear secrets from previously failed provisions. I tried avoiding it as part of this PR but that meant writing some weird-looking code that didn't make any sense. Something like if infraID != nil -> remove secrets but if prevInfraID != nil -> do not remove secrets.

pkg/controller/clusterdeployment/clusterprovisions.go
```diff
@@ -175,10 +175,11 @@ func (r *ReconcileClusterDeployment) startNewProvision(
...
 
-	// Copy over the cluster ID and infra ID from previous provision so that a failed install can be removed.
-	if cd.Spec.ClusterMetadata != nil {
-		provision.Spec.PrevClusterID = &cd.Spec.ClusterMetadata.ClusterID
-		provision.Spec.PrevInfraID = &cd.Spec.ClusterMetadata.InfraID
+	// Copy over the name, cluster ID and infra ID from previous provision so that a failed install can be removed.
+	if lastFailedProvision != nil {
+		provision.Spec.PrevProvisionName = &lastFailedProvision.Name
+		provision.Spec.PrevClusterID = lastFailedProvision.Spec.ClusterID
+		provision.Spec.PrevInfraID = lastFailedProvision.Spec.InfraID
```
akhil-rane, line 182: I have no clue why we were fetching these details from cluster deployment before. Nothing went wrong during my testing after this change.

joelddiaz (commented):
👍 this looks great. not sure if you want another set of eyes, but i'm okay with this.

pkg/controller/clusterdeployment/clusterprovisions.go
2uasimojo: The previous code implies that we've been copying the infra ID to the ClusterDeployment for each provision. This will affect the followon work we've been discussing to add webhooks to make that a write-once field. (No action, just an observation.)

pkg/installmanager/installmanager_test.go
```diff
@@ -145,6 +145,11 @@ func TestInstallManager(t *testing.T) {
 			expectPasswordSecret:          true,
 			expectProvisionMetadataUpdate: true,
 		},
+		{
+			name:        "infraID already set on cluster provision", // fatal error
+			existing:    []runtime.Object{testClusterDeployment(), testClusterProvisionWithInfraIDSet()},
+			expectError: true,
```
2uasimojo, line 151: It would be nice if we could verify that this is in fact the error we expected. But that's a latent issue, something for the backlog.

2uasimojo (commented):
Love this change. Really cleans up the logic, in addition to closing the gap as intended.

I'll leave it to Joel to stamp it.

PR #1650, opened 2022-01-13
update vendoring of x/crypto

joelddiaz:

PR #1654, opened 2022-01-18
update vendoring of x/crypto

2uasimojo:
Since cherry-picking #1650 didn't work cleanly, this is a redo of same.


2uasimojo:
/assign @joelddiaz 

Are we worried about the gcp and/or azure jobs?

joelddiaz:
looks like generic CI issues, nothing specific to hive (assuming I read that output correctly)

PR #1656, opened 2022-01-18
Update installer vendor to ad535d3.

abutcher:
/hold


pkg/controller/machinepool/gcpactuator.go
```diff
@@ -469,38 +469,20 @@ func getNetwork(remoteMachineSets []machineapi.MachineSet,
...
-	obj, gvk, err := decoder.Decode([]byte(rawExt.Raw), nil, nil)
-	if err != nil {
-		return nil, fmt.Errorf("could not decode GCP ProviderSpec: %v", err)
-	}
-	spec, ok := obj.(*machineapi.GCPMachineProviderSpec)
-	if !ok {
-		return nil, fmt.Errorf("Unexpected object: %#v", gvk)
-	}
-	return spec, nil
-}
```
abutcher, line 508: I opted to just pull in the `gcpprovider` as they have `ProviderSpecFromRawExtension(rawExtension)`.

pkg/controller/machinepool/gcpactuator.go
```diff
@@ -471,38 +469,20 @@ func getNetwork(remoteMachineSets []machineapi.MachineSet,
...
 	if err != nil {
 		logger.WithError(err).Warn("cannot decode GCPMachineProviderSpec from remote machinesets")
 		return "", "", errors.Wrap(err, "cannot decode GCPMachineProviderSpec from remote machinesets")
 	}
 
 	if len(remoteMachineProviderSpec.NetworkInterfaces) == 0 {
-		logger.Warn("remote machine do not have any network interfaces")
-		return "", "", errors.Wrap(err, "remote machine do not have any network interfaces")
+		logger.Error("remote machine does not have any network interfaces")
+		return "", "", errors.New("remote machine does not have any network interfaces")
```
abutcher, line 482: I've changed this to return an error. Previously the errors.Wrap would always be nil if we were able to decode the GCPMachineProviderSpec successfully but there were no network interfaces.

2uasimojo: Should that be a separate PR?

abutcher: The reason this needs to return an error is because `ProviderSpecFromRawExtension(rawExtension)` will give you back an empty object if your `rawExension` was `nil` which we do in our test here https://github.com/openshift/hive/blob/master/pkg/controller/machinepool/gcpactuator_test.go#L700.

2uasimojo: Got it.

2uasimojo (commented):
/lgtm

PR #1657, opened 2022-01-18
ClusterClaim: ClusterRunning cond => ClusterState status

2uasimojo:
Since #1652, we only ever assign claims to clusters that are already running, so we don't need the ClusterRunning condition anymore.
    
However, some of the other information for which that condition was being used -- namely, indicating when the cluster is deleting or nonexistent -- is still useful, though doesn't make sense to be associated with a thing called "ClusterRunning". So we add a Status.ClusterState which can have one of three values:
- `NoCluster`: Either we haven't fulfilled the claim yet, or the associated cluster has been deleted.
- `Provisioned`: The claimed cluster exists and is not being deleted (though it is not necessarily running).
- `Deleting`: The claimed cluster is in the process of being deleted (usually meaning it is deprovisioning.)

[HIVE-1743](https://issues.redhat.com/browse/HIVE-1743)

### DIFF
--- a/apis/hive/v1/clusterclaim_types.go
+++ b/apis/hive/v1/clusterclaim_types.go
@@ -16,8 +16,7 @@ type ClusterClaimSpec struct {
 	Subjects []rbacv1.Subject `json:"subjects,omitempty"`
 
 	// Namespace is the namespace containing the ClusterDeployment (name will match the namespace) of the claimed cluster.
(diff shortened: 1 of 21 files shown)


2uasimojo:
/hold

This a) is currently stacked on #1652; b) needs a [change from DPTP](https://issues.redhat.com/browse/DPTP-2720) before we can safely land it.

PR #1658, opened 2022-01-20
OSD-9688: Add regex for VPC missing install failure

blrm:
This adds a regex and test for the AWS "VPC not found" install error. This is believed to be a transient error due to AWS API slowness. 

Additional details in: https://issues.redhat.com/browse/OSD-9688


2uasimojo:
Greg acked in slack

/hold cancel

/test e2e

PR #1660, opened 2022-01-20
Leader election migration 1: ConfigMap & Lease

2uasimojo:
We were using ConfigMap-based leader election. We need to migrate to using Lease instead.

Part 1: Use ConfigMap *and* Lease so we can upgrade smoothly from a build using just ConfigMap.
Part 2: Remove ConfigMap.

This is part 1.

[HIVE-1736](https://issues.redhat.com/browse/HIVE-1736)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -60,8 +60,8 @@ import (
 )
 
 const (
-	defaultLogLevel         = "info"
-	leaderElectionConfigMap = "hive-controllers-leader"
+	defaultLogLevel        = "info"
+	leaderElectionLockName = "hive-controllers-leader"
 )
 
 type controllerSetupFunc func(manager.Manager) error
@@ -225,7 +225,7 @@ func newRootCommand() *cobra.Command {
 			if os.Getenv("HIVE_SKIP_LEADER_ELECTION") != "" {
 				run(ctx)
 			} else {
-				cmdutil.RunWithLeaderElection(ctx, cfg, hiveNSName, leaderElectionConfigMap, run)
+				cmdutil.RunWithLeaderElection(ctx, cfg, hiveNSName, leaderElectionLockName, run)
 			}
 		},
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,20 +9,15 @@ import (
 	"os"
 	"time"
 
-	"github.com/google/uuid"
 	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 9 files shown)


cmd/util/leaderelection.go
```diff
@@ -25,19 +24,26 @@ func RunWithLeaderElection(ctx context.Context, cfg *rest.Config, lockNS, lockNa
...
+		lockNS,
+		lockName,
+		kubeClient.CoreV1(),
+		kubeClient.CoordinationV1(),
+		resourcelock.ResourceLockConfig{
 			Identity: id,
 		},
+	)
+	if err != nil {
+		log.Fatal("failed to create lock for leader election config")
```
2uasimojo, line 46: `.WithError(err)`

cmd/operator/main.go
```diff
@@ -8,18 +8,13 @@ import (
...
 
-	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/kubernetes"
```
joelddiaz, line 19: If we're going to clean up this package, can we also remove those two `_` imports that appear to serve no purpose? (FWIW, they're also in the manager package too...)

cmd/manager/main.go
```diff
@@ -64,11 +60,8 @@ import (
 )
 
 const (
-	defaultLogLevel             = "info"
-	leaderElectionConfigMap     = "hive-controllers-leader"
-	leaderElectionLeaseDuration = "120s"
-	leaderElectionRenewDeadline = "90s"
-	leaderElectionRetryPeriod   = "30s"
+	defaultLogLevel         = "info"
+	leaderElectionConfigMap = "hive-controllers-leader"
```
joelddiaz: Should this be 'leaderElectionLockName' (to follow the other changes in this PR)?

2uasimojo: That's done in the third commit. (You must be looking at the second.)

cmd/operator/main.go
2uasimojo: Taking out magic `_` imports scares me, as they _sometimes_ seem to have implications for code generation and other not-runtime ballyhoo. Can I do that in a separate PR?

cmd/manager/main.go
joelddiaz: Yep. Missed it as I hoped around the commits.

PR #1665, opened 2022-01-26
Leader election migration 1: ConfigMap & Lease

2uasimojo:
We were using ConfigMap-based leader election. We need to migrate to using Lease instead.

Part 1: Use ConfigMap and Lease so we can upgrade smoothly from a build using just ConfigMap.
Part 2: Remove ConfigMap.

This is part 1.

HIVE-1736

Cherry-picked (manually, with conflicts) from #1660.


2uasimojo:
We've decided to abandon this. The plan is to leave this (phase 1) in `master`, which will be what goes into ACM 2.5. As soon as we branch `ocm-2.5`, `master` will be what eventually becomes ACM 2.6, so we can drop phase 2 in there.

PR #1668, opened 2022-01-27
ClusterPool.Spec.HibernationConfig.ResumeTimeout

2uasimojo:
Add a knob causing the clusterpool controller to consider a
ClusterDeployment broken if it spends too long resuming.

[HIVE-1746](https://issues.redhat.com/browse/HIVE-1746)

### DIFF
--- a/pkg/controller/clusterpool/collections.go
+++ b/pkg/controller/clusterpool/collections.go
@@ -213,7 +213,11 @@ type cdCollection struct {
 	byClaimName map[string]*hivev1.ClusterDeployment
 }
 
-func isBroken(cd *hivev1.ClusterDeployment) bool {
+// NOTE: This doesn't care about claimed or deleted/deleting status. That's on the caller.
+func isBroken(cd *hivev1.ClusterDeployment, pool *hivev1.ClusterPool, logger log.FieldLogger) bool {
+	////
+	// Check for ProvisionStopped
+	////
 	cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ProvisionStoppedCondition)
 	if cond == nil {
 		// Since we should be initializing conditions, this probably means the CD is super fresh.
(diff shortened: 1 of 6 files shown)


2uasimojo:
/hold

...for live test.

2uasimojo:
Live test checks out.

/hold cancel
/assign @joelddiaz 
/cc @suhanime

pkg/controller/clusterpool/collections.go
```diff
@@ -213,17 +213,57 @@ type cdCollection struct {
...
+	}
+	if !cd.Spec.Installed {
+		// Don't bother checking on a cluster that isn't installed yet.
+		return false
+	}
+	if cd.Spec.PowerState != hivev1.ClusterPowerStateRunning {
+		// If spec.powerState isn't Running, we're not resuming.
+		return false
+	}
+	if cd.Status.PowerState == hivev1.ClusterPowerStateHibernating || cd.Status.PowerState == hivev1.ClusterPowerStateRunning {
```
suhanime, line 247: I understand the intent but I can't envision a scenario which the condition based checks below won't handle - maybe remove the redundant piece?

pkg/controller/clusterpool/collections.go
```diff
@@ -213,17 +213,57 @@ type cdCollection struct {
...
+		return false
+	}
+	cond = controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
+	if cond == nil {
+		// Since we should be initializing conditions, this probably means the CD is super fresh.
+		// Don't declare it broken yet -- give it a chance to come to life.
+		return false
+	}
+	if cond.Reason != hivev1.HibernatingReasonResumingOrRunning {
+		// It would be weird if we got here, but better safe.
```
suhanime, line 259: If you choose to remove the powerstate check above, then this comment needs to be changed

suhanime (commented):
Minor nit, otherwise looks good

pkg/controller/clusterpool/collections.go
2uasimojo: The redundant piece would only be the check for `ClusterPowerStateHibernating`, right? Because the check on L258 doesn't cover `ClusterPowerStateRunning`. And given that, IMO the code reads clearer this way. (Also, nit, this check is quicker than the condition search.)

PR #1670, opened 2022-01-28
fix overly specific install log regexp for BootstrapFailed

gregsheremeta:
/assign @2uasimojo


2uasimojo (commented):
Makes sense, the alternative being a separate code for each sub-reason.

/lgtm

PR #1671, opened 2022-01-28
prefer generic InconsistentTerraformResult install log regexp

gregsheremeta:
there were two specific terraform race related regexps in the list,
but new ones are popping up every day. replace with a generic
catch-all InconsistentTerraformResult code.

/assign @2uasimojo


2uasimojo (commented):
If you say it should be thus...

...but you need to `make update-gofmt`

gregsheremeta:
> If you say it should be thus...

yeah, generic is better because the overall problem is "terraform issues" and in manifests in new ways -- we find a new one every day at this point. Today's example
https://bugzilla.redhat.com/show_bug.cgi?id=2047741

2uasimojo:
> ...but you need to `make update-gofmt`

Still this.

2uasimojo:
/lgtm

if CI is happy.

PR #1674, opened 2022-01-31
ClusterPool: Secondary sort by ns/name for running

2uasimojo:
The clusterpool controller sorts unclaimed clusters when deciding which
to make Running to satisfy `runningCount` and/or pending claims.
Generally we want FIFO behavior, so the primary sort is by
`creationTimestamp`. However, since that has a granularity of 1s, it is
possible for multiple CDs to have the same value, leading to
nondeterministic sorting on each reconcile and potential "thrashing" of
running-ness.

With this commit, we add a secondary sort by CD namespace and name to
force the sort to be deterministic.

[HIVE-1747](https://issues.redhat.com/browse/HIVE-1747)

### DIFF
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -431,11 +431,15 @@ func (r *ReconcileClusterPool) reconcileRunningClusters(
 	runningCount := int(clp.Spec.RunningCount) + extraRunning
 	// Exclude broken clusters
 	cdList := cds.Unassigned(false)
-	// Sort by age, oldest first
+	// Sort by age, oldest first, for FIFO purposes. Include secondary sort by namespace/name as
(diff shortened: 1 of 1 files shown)


2uasimojo:
NB: I couldn't figure out a reliable way to UT this. I considered a sanity check that creates two CDs with the same creationTimestamp and proves the right one gets the running state; but the current test framework doesn't currently have a way to validate *which* clusters become running, and I didn't think it was worth the effort to extend it just for this.

But I did test it live and am convinced that it worked: I created a pool with runningCount=1 and scaled it down and up until I got two CDs with the same creationTimestamp. I polled their spec.powerState with the old code and observed that it flipflopped. Installed the new code and it quit flipflopping, settling down with the alpha-smallest-named one running.

joelddiaz (commented):
/lgtm

PR #1675, opened 2022-02-01
Added NATGatewayFailed Error code

lnguyen1401:
Added this error message ( found in the install log)  to a list of known errors in OCM, so it doesn't pop up with "OCM3999 - Unknown error" code and save Primary On-Call time debugging.

Full Install Log: https://docs.google.com/document/d/1PilYlo8wyB0kAiInyIaDnb8964Y0IIf2LXgXpacaegg/edit?usp=sharing
Related Slack Thread : https://coreos.slack.com/archives/CCX9DB894/p1643661195222699
Related Jira Card: https://issues.redhat.com/browse/OSD-9846

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -124,6 +124,11 @@ data:
       installFailingMessage: AWS Subnet Does Not Exist
     # iam:CreateServiceLinkedRole is a super powerful permission that we don't give to STS clusters. We require it's done as a one-time prereq.
     # This is the error we see when the prereq step was missed.
+    - name: NATGatewayFailed 
+      searchRegexStrings:
+      - "Error waiting for NAT Gateway (.*) to become available"
+      installFailingReason: NATGatewayFailed
+      installFailingMessage: Error waiting for NAT Gateway to become available.
     - name: AWSAccessDeniedSLR
       searchRegexStrings:
       - "Error creating network Load Balancer: AccessDenied.*iam:CreateServiceLinkedRole"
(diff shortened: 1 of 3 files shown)


2uasimojo (commented):
Sigh. You got hit by a couple of things here:
- A [known bug](https://issues.redhat.com/browse/HIVE-1723) that causes codegen to generate files under github.com. This is bogus -- just remove these from your commit. (You ?_may_ be able to work around this by explicitly setting `GOPATH`.)
- A [known difference](https://issues.redhat.com/browse/HIVE-1753) in codegen from go 1.16 to 1.17 that results in some deepcopy deltas. Our CI is running under go1.16, so the `verify` test will kick those deltas out. You could work around this by using go1.16 when you `make update` -- or you could just manually remove those files from your commit.
- Not sure how `saas-objects.yaml` remained in your checkout, given that it's a temporary file we [remove](https://github.com/openshift/hive/blob/8dc97bc97ec826dbc84b1ae9c9bc872b323eb0fd/Makefile#L211-L212) in the same build rule that creates it. Anyway, that file should be removed from the commit.

And one substantive comment inline.

pkg/controller/clusterprovision/installlogmonitor_test.go
```diff
@@ -321,6 +322,26 @@ func TestParseInstallLog(t *testing.T) {
...
+				Data: map[string]string{
+					"regexes": `
+- name: NATGatewayFailed
+  searchRegexStrings:
+  - "Error waiting for NAT Gateway (.*) to become available."
+  installFailingReason:  NATGatewayFailed
+  installFailingMessage: Error waiting for NAT Gateway to become available.
+`,
+				},
+			}},
```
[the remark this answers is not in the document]
lnguyen1401: awesome thank you Eric! i'm gonna make those change !

2uasimojo (approved):
Thanks for the updates. This is now mechanically sound afaict.

I'll let @gregsheremeta add the lgtm.

2uasimojo:
Looks like you still need to gofmt the test file.

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -124,6 +124,11 @@ data:
       installFailingMessage: AWS Subnet Does Not Exist
     # iam:CreateServiceLinkedRole is a super powerful permission that we don't give to STS clusters. We require it's done as a one-time prereq.
     # This is the error we see when the prereq step was missed.
+    - name: NATGatewayFailed 
+      searchRegexStrings:
+      - "Error waiting for NAT Gateway (.*) to become available."
```
gregsheremeta, line 129: nitpick, feel free to ignore: I'd probably remove the dot at the end

lnguyen1401: i removed the full-stop just to be sure :)

gregsheremeta:
/lgtm

you're awesome, @lnguyen1401 !

PR #1676, opened 2022-02-01
Hack GOPATH for codegen

2uasimojo:
Copy/adapt [GOPATH ridiculousness](https://github.com/openshift/cloud-credential-operator/blob/aa5510254782d506f59a0578661f22a8d639dc85/hack/update-codegen.sh) from CCO to work around codegen bugs that would generate the code under github.com/ in the repo root.

Check in the resulting deepcopy deltas we've apparently been missing for a while.

[HIVE-1723](https://issues.redhat.com/browse/HIVE-1723)


akhil-rane:
from [this PR](https://github.com/kubernetes/kubernetes/pull/106964) it looks like code generation outside GOPATH should be fixed now. Maybe we should vendor in latest changes and see if that does the trick

2uasimojo:
> from [this PR](https://github.com/kubernetes/kubernetes/pull/106964) it looks like code generation outside GOPATH should be fixed now. Maybe we should vendor in latest changes and see if that does the trick

See #1677. The results are quite different:
- The deepcopy deltas seen in this PR are absent.
- #1677 has a bunch of deltas in the generated clientset files that are absent from this PR.

I don't know if any of that is significant :man_shrugging:

PR #1684, opened 2022-02-10
Bug 2050332: explicit Pattern to validate Duration fields

2uasimojo:
Due to a difference in how apiserver validates `Format=duration` vs how
apimachinery unmarshals `metav1.Duration`, using that format can allow
users to blow up some of our controllers if they provide certain values
for Duration fields (specifically: containing `d` or `w` units).
Upstream issues have been opened against both sides of this discrepancy:
- https://github.com/kubernetes/apimachinery/issues/131
- https://github.com/kubernetes/apiextensions-apiserver/issues/56

In the meantime, work around the problem by using an explicit regex
instead of the built in format validator.

Closes 2050332

### DIFF
--- a/apis/hive/v1/clusterclaim_types.go
+++ b/apis/hive/v1/clusterclaim_types.go
@@ -24,9 +24,13 @@ type ClusterClaimSpec struct {
 	// Lifetime is the maximum lifetime of the claim after it is assigned a cluster. If the claim still exists
 	// when the lifetime has elapsed, the claim will be deleted by Hive.
 	// This is a Duration value; see https://pkg.go.dev/time#ParseDuration for accepted formats.
+	// Note: due to discrepancies in validation vs parsing, we use a Pattern instead of `Format=duration`. See
+	// https://bugzilla.redhat.com/show_bug.cgi?id=2050332
+	// https://github.com/kubernetes/apimachinery/issues/131
+	// https://github.com/kubernetes/apiextensions-apiserver/issues/56
 	// +optional
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Format=duration
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
 	Lifetime *metav1.Duration `json:"lifetime,omitempty"`
 }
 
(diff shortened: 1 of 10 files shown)


2uasimojo:
I don't know how to make the bz "target 4.11.0". The only thing I see with "target" is "Target Milestone: ---", which I can't seem to edit. @akhil-rane can you help?

akhil-rane:
> I don't know how to make the bz "target 4.11.0". The only thing I see with "target" is "Target Milestone: ---", which I can't seem to edit. @akhil-rane can you help?

you need to click on `show advanced fields` and then you target release

2uasimojo:
Tested live, works. (Though I preferred the native error message.)

```
The ClusterClaim "the-claim" is invalid: spec.lifetime: Invalid value: "1q": spec.lifetime in body should match '^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$'
...
The ClusterClaim "the-claim" is invalid: spec.lifetime: Invalid value: "1w": spec.lifetime in body should match '^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$'
...
The ClusterClaim "the-claim" is invalid: spec.lifetime: Invalid value: "1d": spec.lifetime in body should match '^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$'
...
(with "1h"):
clusterclaim.hive.openshift.io/the-claim created
```
And the controllers are happy.

/assign @abutcher

2uasimojo:
@lwan-wanglin how does QE work for things like this with a BZ but no Jira?

lwan-wanglin:
> @lwan-wanglin how does QE work for things like this with a BZ but no Jira?

For hive BZ, we will move Status from ON_QA to Verified after testing. QE behavior won't influnce the hive PR merge because hive won't release within OCP.

For CCO , we now follow [non-Feature Freeze](https://docs.google.com/presentation/d/1rBU70bMWjmBbA0EUVdo9ZIqSnvvmMZU-k1NCs7ROD0w/edit#slide=id.g3d946fe0a1_0_0)  workflow, for user stories, we need QE, Docs, and PM add their respective “XXX-approved” label to the PR; for BZ, we need QE add their respective “qe-approved” label.

2uasimojo:
Manual cherry-pick: https://github.com/openshift/hive/pull/1690

PR #1614, opened 2021-11-18
Use a ConfigMap for FailedProvisionConfig

2uasimojo:
Create a ConfigMap containing FailedProvisionConfig and use it instead of the existing mechanisms:
- Environment variables for log upload
- Directly reading HiveConfig for RetryReasons

[HIVE-1692](https://issues.redhat.com/browse/HIVE-1692)

PR #1685, opened 2022-02-11
Fix provision Nil panic

abraverm:
When running the controller locally (development), it doesn't have any of the configurations that Hive operators adds when it creates the controller:
https://github.com/openshift/hive/blob/master/pkg/operator/hive/hive_controller.go#L516
[fpConfigHash](https://github.com/openshift/hive/blob/master/pkg/operator/hive/hive_controller.go#L483) -> [failedProvisionConfigMapInfo](https://github.com/openshift/hive/blob/master/pkg/operator/hive/configmap.go#L96)
This creates environment variable FAILED_PROVISION_CONFIG_FILE that [points at the created configuration](https://github.com/openshift/hive/blob/master/pkg/operator/hive/configmap.go#L100).
The bug is that when that environment variable is not defined then the [readProvisionFailedConfig](https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterprovisions.go#L573) will return nil, and that causes panic when a [provision is retried](https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterprovisions.go#L256) as it tries to access an attribute in a nil object.


2uasimojo (approved):
Yup. This was missed in e7f4b1fee / #1614. Thanks @abraverm!

akhil-rane (commented):
/lgtm

PR #1700, opened 2022-02-24
Pass ClusterName to uninstall for IBM Cloud DNS cleanup.

abutcher:
IBM Cloud uninstall requires being passed the `ClusterName` to clean up the correct DNS entries. `ClusterName` is currently not being provided resulting in **ALL** DNS entries in a CIS instance being deleted that match the uninstall `BaseDomain`. Without the `ClusterName`, the following regex matches all DNS entries for the provided `BaseDomain`: https://github.com/openshift/installer/blob/master/pkg/destroy/ibmcloud/dns.go#L31.

This PR passes `ClusterName` obtained from `cd.Spec.ClusterName` (which is a required field for `ClusterDeployment`) to uninstall.


2uasimojo:
Nice find

/lgtm

PR #1703, opened 2022-02-25
oVirt: Fix credentials and cert secret docs.

abutcher:
Correction for the oVirt credentials/certificate secrets. These match what `hiveutil` produces.


akhil-rane (commented):
/lgtm

PR #1705, opened 2022-02-28
appsre: Build schemav2 images under podman

2uasimojo:
By default podman builds images in OCI format, which is Old. The
preferred way is "Schema v2", which docker does by default. Set an
environment variable so this happens when we build our app-sre images.


2uasimojo:
Per [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1647958974295729), this might be sufficient as is. We're going to try it. A followon may be needed.

2uasimojo:
This worked.
[Before](https://quay.io/v2/app-sre/hive/manifests/sha256:1177f2334729c5809f8b0edbe96a2482d3696822cdab08995acfbc0b8c32f086):
`"mediaType":"application/vnd.oci.image.config.v1+json"`

[After](https://quay.io/v2/app-sre/hive/manifests/sha256:a2f5991b555289ea84577a44ff9e0fd9ebc64e49361431b392b7bf1d6c05199e):
`"mediaType":"application/vnd.docker.distribution.manifest.v2+json"`

PR #1706, opened 2022-02-28
IBMCloud: Deprecate CISInstanceCRN and AccountID IBM Cloud platform fields

abutcher:
[HIVE-1620](https://issues.redhat.com/browse/HIVE-1620)

`CISInstanceCRN` and `AccountID` parameters (required for cluster deprovision) can be retrieved from the IBMCloud API. Hive will no longer require them. These fields are now optional and will not be used.

/hold


2uasimojo:
This looks sane. Held for testing?

/test e2e

abutcher:
@2uasimojo Yes, also held because I'd prefer to just remove these from the API.

pkg/installmanager/installmanager.go
```diff
@@ -687,23 +688,54 @@ func cleanupFailedProvision(dynClient client.Client, cd *hivev1.ClusterDeploymen
...
+		ibmCloudAPIKey := os.Getenv(constants.IBMCloudAPIKeyEnvVar)
+		if ibmCloudAPIKey == "" {
+			return fmt.Errorf("No %s env var set, cannot proceed", constants.IBMCloudAPIKeyEnvVar)
+		}
+		ibmClient, err := ibmclient.NewClient(ibmCloudAPIKey)
+		if err != nil {
+			return errors.Wrap(err, "Unable to create IBM Cloud client")
+		}
+		// Retrieve CISInstanceCRN
+		zones, err := ibmClient.GetDNSZones(context.TODO())
```
abutcher, line 701: This is duplicated in [contrib/pkg/deprovision/ibmcloud.go](https://github.com/openshift/hive/pull/1706/files#diff-9b605f3085fc9053107d1121f6b38e00c30c59e0522bc018bcd65afe8a194d96) and should be put somewhere common for deprovision and installmanager to call.

abutcher: I added funcs for both of these to the client.

abutcher:
Discussed removal in scrum today. For backwards compat they will need to be optional in the API and then will be unused (deprecated) by hive going forward.

PR #1711, opened 2022-03-07
IBMCloud: Hiveutil fixes

abutcher:
[HIVE-1620](https://issues.redhat.com/browse/HIVE-1620)

* Remove `--resource-group-name` from `hiveutil deprovision ibmcloud`. Use provided `InfraID` as `ResourceGroupName`.
* Remove `--cis-instance-crn` and `--account-id` from `hiveutil deprovision ibmcloud` usage. These were removed in #1706 and I missed removing them here.
* Add `--cluster-name` to `hiveutil deprovision ibmcloud` usage. `--cluster-name` is a required parameter (needed for cleaning up DNS entries) should be in the usage example. 
* Fix an inconsistency with `hiveutil create-cluster` I noticed writing usage docs. `--ibm-instance-type` should also be reflected in the generated install config as should the default.

/hold

### DIFF
--- a/contrib/pkg/deprovision/ibmcloud.go
+++ b/contrib/pkg/deprovision/ibmcloud.go
@@ -18,23 +18,22 @@ import (
 
 // ibmCloudDeprovisionOptions is the set of options to deprovision an IBM Cloud cluster
 type ibmCloudDeprovisionOptions struct {
-	accountID         string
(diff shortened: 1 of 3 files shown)


abutcher:
Tested install + uninstall w/ `InfraID` as `ResourceGroupName`.

PR #1712, opened 2022-03-08
Add metrics for currently transitioning clusters

suhanime:
- Add metrics for currently transitioning clusters (that would be cleared up once they have successfully transitioned)
- Make all powerstate transition metrics optional and include a provision to opt for them via hiveConfig

/assign @2uasimojo


pkg/controller/metrics/metrics.go
```diff
@@ -197,6 +224,35 @@ type Calculator struct {
...
+					MapMetricToDuration[MetricResumingClusters] = metric.Duration.Duration
+				case hivev1.CurrentWaitingForCO:
+					metrics.Registry.MustRegister(MetricWaitingForCOClusters)
+					MapMetricToDuration[MetricWaitingForCOClusters] = metric.Duration.Duration
+				case hivev1.CumulativeHibernated:
+					metrics.Registry.MustRegister(MetricClusterHibernationTransitionSeconds)
+					MapMetricToDuration[MetricClusterHibernationTransitionSeconds] = metric.Duration.Duration
+				case hivev1.CumulativeResumed:
+					metrics.Registry.MustRegister(MetricClusterReadyTransitionSeconds)
+					MapMetricToDuration[MetricClusterReadyTransitionSeconds] = metric.Duration.Duration
```
suhanime, line 250: I tried to have a direct mapping with hiveconfig types, but golang's strict rules won't allow me to dynamically register a metric

pkg/controller/metrics/metrics.go
```diff
@@ -197,6 +224,35 @@ type Calculator struct {
 func (mc *Calculator) Start(ctx context.Context) error {
 	log.Info("started metrics calculator goroutine")
 
+	// Register optional metrics and add them to corresponding maps, so controllers logging them can access
+	// the information
+	hiveConfig := &hivev1.HiveConfig{}
```
suhanime, line 229: I wanted to move this part where we register provision metrics (when we add to manager) but I think there's no hiveconfig object available to list then

apis/hive/v1/hiveconfig_types.go
```diff
@@ -144,8 +143,45 @@ type HiveConfigSpec struct {
...
+
+const (
+	CurrentStopping     DurationMetricType = "CurrentStopping"
+	CurrentResuming     DurationMetricType = "CurrentResuming"
+	CurrentWaitingForCO DurationMetricType = "CurrentWaitingForCO"
+
+	// These metrics will not be cleared and can potentially blow up the cardinality
+	CumulativeHibernated DurationMetricType = "CumulativeHibernated"
+	CumulativeResumed    DurationMetricType = "CumulativeResumed"
+)
```
suhanime, line 183: This is purely done for the sake of mapping in metrics controller, and serves no other purpose. The real info for a use trying to set this value comes from the enum validation on DurationMetricType.

suhanime (commented):
Metrics registered if opted via HiveConfig, logged after set duration and cleared for relevant metrics

```spec:
    logLevel: debug
    metricsConfig:
      metricsWithDuration:
      - duration: 90s
        name: CurrentStopping
      - duration: 2m
        name: CurrentResuming
    targetNamespace: hive
```
image

@2uasimojo ready for review

pkg/controller/metrics/metrics.go
```diff
@@ -105,6 +106,31 @@ var (
 			Buckets: []float64{60, 90, 180, 300, 600, 1200},
 		},
 		[]string{"cluster_version", "platform", "cluster_pool_namespace", "cluster_pool_name"})
+	// MetricStoppingClusters is a prometheus metric that tracks the transition time for currently stopping clusters
+	MetricStoppingClusters = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_cluster_deployments_stopping",
```
2uasimojo, line 112: I believe metric names are supposed to have units.

...[yeah](https://prometheus.io/docs/practices/naming/)
```suggestion
			Name:    "hive_cluster_deployments_stopping_seconds",
```

pkg/controller/metrics/metrics.go
```diff
@@ -105,6 +106,31 @@ var (
...
+			Help:    "Distribution of the length of transition time for clusters currently stopping",
+			Buckets: []float64{30, 60, 180, 300, 600},
+		},
+		[]string{"cluster_deployment", "platform", "cluster_version", "cluster_pool_namespace"})
+	// MetricResumingClusters is a prometheus metric that tracks the transition time for currently resuming clusters
+	MetricResumingClusters = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_cluster_deployments_resuming",
```
2uasimojo, line 120:
```suggestion
			Name:    "hive_cluster_deployments_resuming_seconds",
```

pkg/controller/metrics/metrics.go
```diff
@@ -105,6 +106,31 @@ var (
...
+			Help:    "Distribution of the length of transition time for clusters currently resuming",
+			Buckets: []float64{60, 90, 180, 300, 600, 1200},
+		},
+		[]string{"cluster_deployment", "platform", "cluster_version", "cluster_pool_namespace"})
+	// MetricWaitingForCOClusters is a prometheus metric that tracks the transition time for clusters currently in
+	// waiting for cluster operators state
+	MetricWaitingForCOClusters = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_cluster_deployments_waiting_for_cluster_operators",
```
2uasimojo, line 129:
```suggestion
			Name:    "hive_cluster_deployments_waiting_for_cluster_operators_seconds",
```

pkg/controller/metrics/metrics.go
```diff
@@ -173,6 +199,7 @@ func Add(mgr manager.Manager) error {
 	}
 	metrics.Registry.MustRegister(newProvisioningUnderwaySecondsCollector(mgr.GetClient(), 1*time.Hour))
 	metrics.Registry.MustRegister(newProvisioningUnderwayInstallRestartsCollector(mgr.GetClient(), 1))
```
2uasimojo, line 201: Let's add a comment in here like
```suggestion
	// TODO: Make these optional & configurable via HiveConfig.Spec.MetricsConfig.
	metrics.Registry.MustRegister(newProvisioningUnderwaySecondsCollector(mgr.GetClient(), 1*time.Hour))
	metrics.Registry.MustRegister(newProvisioningUnderwayInstallRestartsCollector(mgr.GetClient(), 1))
```

pkg/controller/metrics/metrics.go
2uasimojo: Not following this comment. Let's discuss.

pkg/controller/metrics/metrics.go
```diff
@@ -105,6 +106,31 @@ var (
...
 		},
 		[]string{"cluster_version", "platform", "cluster_pool_namespace", "cluster_pool_name"})
+	// MetricStoppingClusters is a prometheus metric that tracks the transition time for currently stopping clusters
+	MetricStoppingClusters = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "hive_cluster_deployments_stopping",
+			Help:    "Distribution of the length of transition time for clusters currently stopping",
+			Buckets: []float64{30, 60, 180, 300, 600},
+		},
+		[]string{"cluster_deployment", "platform", "cluster_version", "cluster_pool_namespace"})
```
2uasimojo, line 116: We need the CD namespace as well as its name

And we need the ClusterPool name as well as its namespace.

(I could get behind making both of these a `*_namespacedname` in theory, though that doesn't match what we're doing for other metrics.)

(Same below)

pkg/controller/metrics/metrics.go
```diff
@@ -246,6 +302,48 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+					poolNS = cd.Spec.ClusterPoolRef.Namespace
+				}
+				hibernatingCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions,
+					hivev1.ClusterHibernatingCondition)
+				readyCond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions,
+					hivev1.ClusterReadyCondition)
+				if readyCond.Reason == hivev1.ReadyReasonStoppingOrHibernating &&
+					hibernatingCond.Reason != hivev1.HibernatingReasonHibernating {
+					transitionTime := time.Since(readyCond.LastTransitionTime.Time).Seconds()
+					if ShouldLogDurationMetric(MetricStoppingClusters, transitionTime) {
```
2uasimojo, line 318: The func we call could incorporate both the conditional and the observe, and be called more simply like `MaybeObserveDurationMetric(MetricStoppingClusters, time.Since(...), cd)`. That'd depend on all such metrics having the same labels... but I say since they do right now we can simplify -- break it up later if necessary.

2uasimojo (commented):
I like this @suhanime, it winds up looking pretty simple and understandable.

I don't see the code where we "clear" the metrics -- can you please point that out?

Looks like you're missing a `make update` to regen the saas template.

Other comments inline.

apis/hive/v1/hiveconfig_types.go
```diff
@@ -144,8 +143,45 @@ type HiveConfigSpec struct {
...
+	MetricsConfig *MetricsConfig `json:"metricsConfig,omitempty"`
+}
+
+type MetricsConfig struct {
+	// Optional metrics and their configurations
+	// +optional
+	MetricsWithDuration []MetricsWithDuration `json:"metricsWithDuration"`
 }
 
+// MetricsWithDuration represent metrics that report time as values,like transition seconds
```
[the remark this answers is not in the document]
2uasimojo: Also: This docstring needs to
- enumerate the metrics that are supported
- map the const names to the metric names

pkg/controller/metrics/metrics.go
2uasimojo: ...and factor out the bit that determines the list of label values

...and call :arrow_up: from a new function that encapsulates the `Delete()`s as well.

apis/hive/v1/metricsConfig/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package metricsConfig contains API Schema definitions for configurations specific to metrics controller.
+// +k8s:deepcopy-gen=package,register
+package metricsConfig
```
suhanime, line 3: Still v1 scoped, let me know if you'd like me to move these files elsewhere

apis/hive/v1/zz_generated.deepcopy.go
```diff
@@ -2402,6 +2403,11 @@ func (in *HiveConfigSpec) DeepCopyInto(out *HiveConfigSpec) {
 		*out = new(FeatureGateSelection)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.MetricsConfig != nil {
+		in, out := &in.MetricsConfig, &out.MetricsConfig
+		*out = new(metricsConfig.MetricsConfig)
+		(*in).DeepCopyInto(*out)
```
suhanime, line 2409: Cannot for the life of me figure out what small change am I missing to generate `v1/metricsConfig/zz_generated.deepcopy.go` automatically

suhanime (commented):
@2uasimojo Almost all of your suggestions have been implemented
Still need help with that deepcopy annotation missing
Kept the changes in a separate commit for easy review. Will rebase once approved

apis/hive/v1/zz_generated.deepcopy.go
2uasimojo: codegen for submodules lives in hack/update-codegen.sh. This should do the trick:
```diff
diff --git a/hack/update-codegen.sh b/hack/update-codegen.sh
index d0da2f89c..e829afffd 100755
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -19,7 +19,7 @@ GOFLAGS="" bash ${CODEGEN_PKG}/generate-groups.sh "all" \
 GOFLAGS="" bash ${CODEGEN_PKG}/generate-groups.sh "deepcopy" \
   github.com/openshift/hive/pkg/client \
   github.com/openshift/hive/apis \
-  "hive:v1/agent hive:v1/aws hive:v1/azure hive:v1/baremetal hive:v1/gcp hive:v1/none hive:v1/openstack hive:v1/ovirt hive:v1/vsphere hive:v1/ibmcloud hivecontracts:v1alpha1" \
+  "hive:v1/agent hive:v1/aws hive:v1/azure hive:v1/baremetal hive:v1/gcp hive:v1/metricsConfig hive:v1/none hive:v1/openstack hive:v1/ovirt hive:v1/vsphere hive:v1/ibmcloud hivecontracts:v1alpha1" \
   --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.go.txt \
   --trim-path-prefix github.com/openshift/hive \
   ${verify}
```

apis/hive/v1/metricsConfig/doc.go
2uasimojo: No, I think this works.

I'm not sure if there's a problem using camelCase for the package name though...

suhanime: So golang package rules suggest all lowercase - fixed it

apis/hive/v1/metricsconfig/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package metricsconfig contains API Schema definitions for configurations specific to metrics controller.
+// +k8s:deepcopy-gen=package,register
+package metricsconfig
```
2uasimojo, line 3: The other packages' doc.go files also have
```suggestion
// +k8s:conversion-gen=github.com/openshift/hive/apis/hive
package metricsconfig
```
I actually have no idea what this does or why it's there. I can't get it to behave differently with `update-codegen` either. Any ideas?

(`conversion-gen` is documented [here](https://pkg.go.dev/k8s.io/code-generator/cmd/conversion-gen) -- but we don't seem to be using it.)

apis/hive/v1/metricsconfig/durationMetrics.go
```diff
@@ -0,0 +1,39 @@
+package metricsconfig
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// MetricsWithDuration represents metrics that report time as values,like transition seconds
+// The purpose of these metrics should be to track outliers - ensure their duration is not set too low
```
2uasimojo, line 6:
```suggestion
// MetricsWithDuration represents metrics that report time as values, like transition seconds.
// The purpose of these metrics should be to track outliers - ensure their duration is not set too low.
```

apis/hive/v1/hiveconfig_types.go
```diff
@@ -144,6 +145,10 @@ type HiveConfigSpec struct {
 	// extract metrics. The operator also sets up RBAC in the TargetNamespace so that openshift
 	// prometheus in the cluster can list/access objects required to pull metrics.
 	ExportMetrics bool `json:"exportMetrics,omitempty"`
+
+	// MetricsConfig encapsulates metrics specific configurations, like opting in for certain metrics
```
2uasimojo, line 149:
```suggestion
	// MetricsConfig encapsulates metrics specific configurations, like opting in for certain metrics.
```

pkg/controller/metrics/metrics.go
```diff
@@ -655,4 +744,40 @@ func (ca *clusterAccumulator) addConditionToMap(cond hivev1.ClusterDeploymentCon
...
+	if err := json.Unmarshal(fileBytes, config); err != nil {
+		return config, err
+	}
+
+	return config, nil
+}
+
+// logDurationMetric should be used to log duration metrics
+func logDurationMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
+	poolNS := "unknown"
```
2uasimojo, line 769: Can we just leave this as the empty string for non-pool clusters?
```suggestion
	poolNS := ""
```

pkg/controller/metrics/metrics.go
```diff
@@ -201,12 +232,41 @@ func (mc *Calculator) Start(ctx context.Context) error {
 	wait.UntilWithContext(ctx, func(ctx context.Context) {
 		mcLog := log.WithField("controller", "metrics")
 		recobsrv := NewReconcileObserver(ControllerName, mcLog)
+		mConfig, err := readMetricsConfig()
+		if err != nil {
+			log.WithError(err).Error("cannot read metrics config")
+			return
```
2uasimojo, line 238: Hm, this means someone could (accidentally or deliberately) disable this whole module with a bogus HiveConfig.

Looking at the code below, I *think* you can just remove this `return` and everything will still work. (Obviously in that scenario none of the metrics will be reported.)
```suggestion
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -612,18 +611,46 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+		// Clear entry from currently resuming clusters if exists
+		deleteMetric(hivemetrics.MetricResumingClustersSeconds, cd)
 	}
 	return reconcile.Result{}, nil
 }
 
+// deleteMetric has the knowledge of labels needed to clear the metrics for currently transitioning clusters
+// Note: do not use for cumulative metrics - they have different labels and are not designed to be cleared
+func deleteMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment) {
+	poolNS := "unknown"
```
2uasimojo, line 626:
```suggestion
	poolNS := ""
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -612,18 +611,46 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+		"cluster_deployment":           cd.Name,
+		"platform":                     cd.Labels[hivev1.HiveClusterPlatformLabel],
+		"cluster_version":              cd.Labels[constants.VersionMajorMinorPatchLabel],
+		"cluster_pool_namespace":       poolNS,
+	})
+}
+
+// logMetric should be used to log cumulative metrics
+func logMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
+	poolNS, poolName := "unknown", "unknown"
```
2uasimojo, line 641:
```suggestion
	poolNS, poolName := "", ""
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -612,18 +611,46 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+			time.Since(hibernatingCondition.LastTransitionTime.Time).Seconds())
+		// Clear entry from currently resuming clusters if exists
+		deleteMetric(hivemetrics.MetricResumingClustersSeconds, cd)
 	}
 	return reconcile.Result{}, nil
 }
 
+// deleteMetric has the knowledge of labels needed to clear the metrics for currently transitioning clusters
+// Note: do not use for cumulative metrics - they have different labels and are not designed to be cleared
+func deleteMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment) {
```
2uasimojo, line 625: At a glance, `deleteMetric` and `logMetric` appear closely related. Reading the comments, I see that the former is intended for "transitioning" metrics and the latter for "cumulative" metrics. That being the case, suggest including that in the names to make it clearer.
```suggestion
func deleteTransitionMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment) {
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -612,18 +611,46 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+		"cluster_deployment_namespace": cd.Namespace,
+		"cluster_deployment":           cd.Name,
+		"platform":                     cd.Labels[hivev1.HiveClusterPlatformLabel],
+		"cluster_version":              cd.Labels[constants.VersionMajorMinorPatchLabel],
+		"cluster_pool_namespace":       poolNS,
+	})
+}
+
+// logMetric should be used to log cumulative metrics
+func logMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
```
2uasimojo, line 640: as above
```suggestion
func logCumulativeMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
```

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -612,18 +611,46 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+	poolNS, poolName := "unknown", "unknown"
+	if cd.Spec.ClusterPoolRef != nil {
+		poolNS = cd.Spec.ClusterPoolRef.Namespace
+		poolName = cd.Spec.ClusterPoolRef.PoolName
+	}
+	if hivemetrics.ShouldLogDurationMetric(metric, time) {
```
2uasimojo, line 646: Suggest putting this short-circuit at the top of the function so we're not doing extra work (albeit negligible) if we're not going to log.
```suggestion
	if !hivemetrics.ShouldLogDurationMetric(metric, time) { return }
```

2uasimojo (commented):
This is looking really good. Just a few things inline, some minor, some _really_ minor.

apis/hive/v1/metricsconfig/doc.go
suhanime: Couldn't figure out why it's not present in every api module and why it'd be needed here, but added it anyways for future purposes when this module expands and that conversion would be needed - doesn't hurt us either way

pkg/controller/hibernation/hibernation_controller.go
suhanime: Made this <none> as per our discussion

pkg/controller/hibernation/hibernation_controller.go
suhanime: <none>

pkg/controller/metrics/metrics.go
suhanime: Can't be empty so made it <none>

pkg/controller/metrics/metrics.go
```diff
@@ -339,6 +395,65 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+	if mConfig.MetricsWithDuration != nil {
+		if MapMetricToDuration == nil {
+			MapMetricToDuration = make(map[*prometheus.HistogramVec]time.Duration)
+		}
+		for _, metric := range mConfig.MetricsWithDuration {
+			switch metric.Name {
+			case metricsconfig.CurrentStopping:
+				// Metrics can only be registered once. Unregistering and registering again doesn't work well because
+				// unregistering is only meant for rare unit tests and doesn't clear all the records of that metric
+				if !existsInDurationMetricsMap(MetricStoppingClustersSeconds) {
```
suhanime, line 409: So - my pod kept on Crashing because without the check here, it was executing `MustRegister` every reconcile, which threw the `already registered` error.
There's a `metrics.Registry.Unregister` which I tried using, but that doesn't clear all the records, hence the `already registered` error wasn't cleared. So I've put in this hack to ensure we only register once.
Whenever the hiveconfig is edited, it anyways terminates the pod and starts a new instance, so I didn't need to clear any values - this hack is only necessary for the reconciles.
I've broken this entire registering part into a new function for ease of reading code.

suhanime: Have also tested with updating the existing duration in hiveconfig, as well as deleting/adding metrics - terminates pod everytime and creates a new instance - but you can still see the values of the older instance on the prometheus, so we should be good

suhanime (commented):
@2uasimojo This should be good to go now

pkg/controller/metrics/metrics.go
```diff
@@ -201,12 +232,19 @@ func (mc *Calculator) Start(ctx context.Context) error {
 	wait.UntilWithContext(ctx, func(ctx context.Context) {
 		mcLog := log.WithField("controller", "metrics")
 		recobsrv := NewReconcileObserver(ControllerName, mcLog)
+		mConfig, err := readMetricsConfig()
+		if err != nil {
+			log.WithError(err).Error("cannot read metrics config")
+		}
+		// Register optional metrics and update them in their corresponding maps, so controllers logging them can access
+		// the information
+		registerOptionalMetrics(mConfig)
```
2uasimojo, line 241: I believe this is why you're having problems with metrics trying to reregister. If you put this block outside of the `wait` loop (i.e. above L232) it should do the right thing.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -612,18 +611,47 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
...
+// logCumulativeMetric should be used to log cumulative metrics
+func logCumulativeMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
+	poolNS, poolName := "<none>", "<none>"
+	if cd.Spec.ClusterPoolRef != nil {
+		poolNS = cd.Spec.ClusterPoolRef.Namespace
+		poolName = cd.Spec.ClusterPoolRef.PoolName
+	}
+	if !hivemetrics.ShouldLogDurationMetric(metric, time) {
+		return
+	}
```
2uasimojo, line 648: nit: This block could go at the top. No reason to compute pool name/namespace if we're not gonna use 'em.

pkg/controller/metrics/metrics.go
```diff
@@ -339,6 +396,51 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+	}
+}
+
+func existsInDurationMetricsMap(metric *prometheus.HistogramVec) bool {
+	if MapMetricToDuration != nil {
+		_, ok := MapMetricToDuration[metric]
+		return ok
+	}
+	return false
+}
```
2uasimojo, line 429: Not used. Kill it dead.

pkg/controller/metrics/metrics.go
```diff
@@ -127,6 +156,9 @@ var (
 		},
 		[]string{"cluster_deployment", "namespace", "cluster_type"},
 	)
+
+	// MapMetricToDuration is a map of optional durationMetrics to their specific duration, if mentioned
+	MapMetricToDuration map[*prometheus.HistogramVec]time.Duration
```
2uasimojo, line 161: This var is only used in this package, so let's make it private.

(We could also make it an object, with a factory method to hide the fact that it's a package global, and `registerOptionalMetrics` and `ShouldLogDurationMetric` as methods on it. But le big meh.)

pkg/controller/metrics/metrics.go
```diff
@@ -339,6 +396,51 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+// optional metric has been opted into for logging, and the duration has crossed the threshold set
+func ShouldLogDurationMetric(metricToLog *prometheus.HistogramVec, timeToLog float64) bool {
+	if MapMetricToDuration != nil {
+		for metric, duration := range MapMetricToDuration {
+			if metric == metricToLog && duration.Seconds() <= timeToLog {
+				return true
+			}
+		}
+	}
+	return false
```
2uasimojo, line 441: - Shouldn't need the nil check as `registerOptionalMetrics` does a `make()`.
- Should be able to look up (O(1)) rather than iterate (O(N)).
```suggestion
	duration, ok := MapMetricToDuration[metricToLog]
	return ok && duration.Seconds() <= timeToLog
```

pkg/controller/metrics/metrics.go
```diff
@@ -655,4 +757,41 @@ func (ca *clusterAccumulator) addConditionToMap(cond hivev1.ClusterDeploymentCon
...
+
+// logDurationMetric should be used to log duration metrics
+func logDurationMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
+	poolNS := "<none>"
+	if cd.Spec.ClusterPoolRef != nil {
+		poolNS = cd.Spec.ClusterPoolRef.Namespace
+	}
+	if !ShouldLogDurationMetric(metric, time) {
+		return
+	}
```
2uasimojo, line 788: move to top of func

pkg/controller/metrics/metrics.go
```diff
@@ -655,4 +757,41 @@ func (ca *clusterAccumulator) addConditionToMap(cond hivev1.ClusterDeploymentCon
...
+	path := os.Getenv(constants.MetricsConfigFileEnvVar)
+	config := &metricsconfig.MetricsConfig{}
+	if len(path) == 0 {
+		return config, nil
+	}
+
+	fileBytes, err := ioutil.ReadFile(path)
+	if err != nil || len(fileBytes) == 0 {
+		return config, err
+	}
```
2uasimojo, line 772: This `err` will be nil if `len(fileBytes) == 0`, causing this func to "succeed" and return an empty `MetricsConfig{}`. Is that intentional? Does the file show up as zero-length if MetricsConfig is omitted from HiveConfig? Or if it's specified but empty? If not...
```suggestion
	if err != nil {
		return config, err
	}
	if len(fileBytes) == 0 {
		return config, errors.New("empty metrics config file")
	}
```

2uasimojo (commented):
Several little things that can be done in a fup.

/lgtm

One thing that might be a bug -- unhold if it isn't.

/hold

pkg/controller/metrics/metrics.go
```diff
@@ -196,6 +228,14 @@ type Calculator struct {
 // Start begins the metrics calculation loop.
 func (mc *Calculator) Start(ctx context.Context) error {
 	log.Info("started metrics calculator goroutine")
+	// Get the metrics config from hiveConfig
+	mConfig, err := readMetricsConfig()
+	if err != nil {
+		log.WithError(err).Error("cannot read metrics config")
```
2uasimojo, line 234: Per [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1651774265655569?thread_ts=1651606872.054189&cid=CE3ETN3J8), please make this fatal.

pkg/controller/metrics/metrics.go
```diff
@@ -655,4 +744,44 @@ func (ca *clusterAccumulator) addConditionToMap(cond hivev1.ClusterDeploymentCon
...
 }
 
+// readMetricsConfig reads the metrics config from the file pointed to by the MetricsConfigFileEnvVar
+// environment variable.
+func readMetricsConfig() (*metricsconfig.MetricsConfig, error) {
+	path := os.Getenv(constants.MetricsConfigFileEnvVar)
+	config := &metricsconfig.MetricsConfig{}
+	if len(path) == 0 {
+		return config, nil
+	}
```
2uasimojo, line 754: Per [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1651774265655569?thread_ts=1651606872.054189&cid=CE3ETN3J8) please make this an error. It means the hive-operator side did something wrong by not setting the env var for the hive-controllers pod.

pkg/controller/metrics/metrics.go
```diff
@@ -655,4 +744,44 @@ func (ca *clusterAccumulator) addConditionToMap(cond hivev1.ClusterDeploymentCon
...
+
+	fileBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return config, err
+	}
+	if len(fileBytes) == 0 {
+		return config, errors.New("empty metrics config file")
+	}
```
2uasimojo, line 762: Per [slack](https://coreos.slack.com/archives/CE3ETN3J8/p1651774265655569?thread_ts=1651606872.054189&cid=CE3ETN3J8), please remove this block. A zero-size file is the result of the user omitting `metricsConfig` from their `hiveConfig.spec`, which is a legitimate scenario.

2uasimojo:
/lgtm

Let's do this!

PR #1715, opened 2022-03-09
Update installer vendor to 6f7dcb2

akhil-rane:
Updating the installer vendor to a commit that includes Alibaba resources


akhil-rane:
@abutcher PTAL

PR #1723, opened 2022-03-21
Rename the "mark pool cluster for deletion" constant

2uasimojo:
This const used to be used only by the ClusterClaim controller to mark a
CD so the ClusterPool controller would delete it while observing its
`MaxConcurrent` budget. We're going to have more use cases where we need
to do something similar. So
- Make the constant more generic (not claim-specific)
- Create a utility function for setting the annotation (next to the one
that already exists for detecting it).

NB: We have to be careful about upgradeability: if the new code runs on
a CD that was annotated by the old code, it needs to still work. So we
still check for old annotation in the detection utility function.


2uasimojo:
/assign @abraverm 

cf. https://github.com/openshift/hive/pull/1672#discussion_r831575640

PR #1732, opened 2022-03-25
Leader election migration 2: Lease only

2uasimojo:
Follows up on #1660 / 7ec5c25fab0d734f9702666b1244cdd737eacee9

This is phase 2 where we move from ConfigMap+Lease to Lease only.

[HIVE-1754](https://issues.redhat.com/browse/HIVE-1754)


2uasimojo:
Tested this live, both with the legacy configmap still present and after deleting it. Works.

PR #1738, opened 2022-03-29
Update gogo/protobuf vendored dep

2uasimojo:
for CVE-2021-3121


2uasimojo:
Not sure why codegen felt the need to muck with the CRD manifests, but it's harmless so :shrug: 

/assign @abutcher

2uasimojo:
> Not sure why codegen felt the need to muck with the CRD manifests

Well, `verify` didn't like that. `make manifests` restored it. Again :shrug:

PR #1741, opened 2022-04-01
Set IBM Cloud VPC service API URLs for region

abutcher:
Region based endpoints must be set for calls to the VPC API, `ListInstances` and `CreateInstanceAction` (Start/Stop) in our client's case.
[HIVE-1801](https://issues.redhat.com/browse/HIVE-1801)


2uasimojo:
That's a weird API.

The code makes sense to look at. Trusting that it works.

/lgtm

PR #1743, opened 2022-04-05
HIVE-1802: Day 0 MachineSet Security Group Filters workaround

abutcher:
[HIVE-1802](https://issues.redhat.com/browse/HIVE-1802)


abutcher:
I am still working on adding tests to this and guard rails that tie the workaround to AWS but PTAL.
/hold

/assign @2uasimojo
/cc @dlom

pkg/installmanager/installmanager.go
```diff
@@ -766,6 +773,50 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
...
+			return nil
+		})
+		if err != nil {
+			m.log.WithError(err).Error("error finding machine pool manifests")
+			return err
+		}
+		// Modify yaml to include the additional security group
+		ops := []byte(fmt.Sprintf(`---
+- op: add
+  path: /spec/template/spec/providerSpec/value/securityGroups/0/filters/0/values/1
```
2uasimojo, line 796: As we discussed, let's see if we can append without assuming the size of the existing list via:
```suggestion
  path: /spec/template/spec/providerSpec/value/securityGroups/0/filters/0/values/-
```

pkg/installmanager/installmanager.go
```diff
@@ -766,6 +773,50 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
...
+			if err != nil {
+				m.log.WithError(err).Error("error reading manifest file")
+				return err
+			}
+			bs, err := patch.Apply(manifestBytes)
+			if err != nil {
+				m.log.WithError(err).Error("error applying patch")
+				return err
+			}
+			ioutil.WriteFile(manifest, bs, 0600)
```
2uasimojo, line 816: See above -- we should grab and use whatever the existing file mode is.

pkg/controller/utils/sa.go
```diff
@@ -51,6 +51,11 @@ var (
 			Resources: []string{"clusterprovisions", "clusterprovisions/finalizers", "clusterprovisions/status"},
 			Verbs:     []string{"get", "list", "update", "watch"},
 		},
+		{
+			APIGroups: []string{"hive.openshift.io"},
+			Resources: []string{"machinepools"},
+			Verbs:     []string{"get", "list", "update"},
```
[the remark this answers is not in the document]
abutcher: This rule is necessary in general but the SA/role won't be updated if they already exist in the namespace.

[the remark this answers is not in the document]
abutcher: When the role is created fresh with this rule it's good to go.

pkg/installmanager/installmanager.go
```diff
@@ -944,6 +995,19 @@ func (m *InstallManager) loadClusterDeployment(provision *hivev1.ClusterProvisio
 	return cd, nil
 }
 
+func (m *InstallManager) loadWorkerMachinePool(cd *hivev1.ClusterDeployment) (workerMachinePool hivev1.MachinePool, err error) {
```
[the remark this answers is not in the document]
abutcher: No reason. I'll put the vars in the returns.

pkg/installmanager/installmanager.go
```diff
@@ -766,6 +773,50 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
...
 
+	// Day 0: Configure Hive worker MachinePool manifests with additional security group when
+	// ExtraWorkerSecurityGroupAnnotation has been set in annotations for the worker MachinePool.
+	// For details, see HIVE-1802.
+	if metav1.HasAnnotation(workerMachinePool.ObjectMeta, constants.ExtraWorkerSecurityGroupAnnotation) {
+		m.log.Info("modifying worker machineset manifests")
+		// Retrieve list of worker MachinePool manifest files
+		machineSetManifests := []string{}
+		err := filepath.Walk(filepath.Join(m.WorkDir, "openshift"), func(path string, info os.FileInfo, err error) error {
+			if !info.IsDir() && strings.Contains(path, "worker-machineset") {
```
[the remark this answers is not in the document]
abutcher: The filename const comes from [here](https://github.com/openshift/installer/blob/35416442b0cdeb64b87fb23ea97b16ac3ddcb00a/pkg/asset/machines/worker.go#L73), I think it should be pretty safe to use.

abutcher:
Testing that @dlom is fully operational.

/cc @dlom

pkg/installmanager/installmanager_test.go
```diff
@@ -717,3 +722,72 @@ func Test_pasteInPullSecret(t *testing.T) {
 		})
 	}
 }
+
+func TestPatchWorkerMachineSet(t *testing.T) {
```
2uasimojo, line 726: This is a neat test, but I feel like we should be able to go higher-level and not care about the underlying implementation as much.

To facilitate that, we could mock the various ioutils to spoof walking the file system...

But I recall that's kind of a pain. I'd be okay if we just pulled out the chunk of code that starts with a MachinePool and a single file already loaded in a []byte, and ends with that []byte (possibly) having been updated to include the new filter. The way you're unmarshalling that as a MachineSet to introspect it is great.

pkg/util/yaml/yaml.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	return yamlpatch.NewNode(&iface).Container(), nil
+}
+
+func Test(c yamlpatch.Container, path, val string) (match bool, err error) {
+	defer func() {
+		if recover() != nil {
+			err = fmt.Errorf("No such path %s\n", path)
+			match = false
+		}
```
2uasimojo, line 27: Should add a comment here, since it's not obvious at a glance...
```suggestion
			// Implicit return
		}
```

pkg/installmanager/installmanager.go
```diff
@@ -944,6 +1019,19 @@ func (m *InstallManager) loadClusterDeployment(provision *hivev1.ClusterProvisio
 	return cd, nil
 }
 
+func (m *InstallManager) loadWorkerMachinePool(cd *hivev1.ClusterDeployment) (*hivev1.MachinePool, error) {
+	mpL := &hivev1.MachinePoolList{}
+	if err := m.DynamicClient.List(context.Background(), mpL, &client.ListOptions{Namespace: m.Namespace}); err != nil {
+		return nil, nil
```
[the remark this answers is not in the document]
abutcher: Yes.

pkg/installmanager/installmanager.go
```diff
@@ -788,6 +842,49 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
...
+	// installer produced yaml that could not be decoded.
+	if err != nil {
+		logger.WithError(err).Error("unable to decode manifest bytes")
+		return modifiedBytes, modified, err
+	}
+
+	// Return if file is not a worker MachineSet manifest
+	isMachineSet, _ := yamlutils.Test(c, "/kind", "MachineSet")
+	// Note: We have to escape the `/` in the label key as `~1` per https://datatracker.ietf.org/doc/html/rfc6901#section-3
+	isWorkerMachineSet, _ := yamlutils.Test(c, "/spec/template/metadata/labels/machine.openshift.io~1cluster-api-machine-type", "worker")
```
2uasimojo, line 862: nit: it would be more efficient to short-circuit (bail if the first one fails, without checking the second).

pkg/installmanager/installmanager.go
```diff
@@ -788,6 +842,49 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
...
+	// Apply patch to manifest
+	modifiedBytes, err = ops.Apply(manifestBytes)
+	if err != nil {
+		logger.WithError(err).Error("error applying patch")
+		return modifiedBytes, modified, err
+	}
+
+	if res := bytes.Compare(modifiedBytes, manifestBytes); res != 0 {
+		modified = true
+	}
```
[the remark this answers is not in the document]
abutcher: I'm adding a test for `"security group already configured"` which I would have assumed does not result in `modified = true` but it appears that if `Apply()` is successful the document will be modified. This is probably true unless your document matches exactly what `yamlpatch` would have output which won't occur in the wild?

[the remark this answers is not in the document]
2uasimojo: The way we're patching, the SG will get appended to the list even if it's already there, resulting in duplicate entries. (I don't know if that would break AWS.)

As you say, that shouldn't happen in real life because the manifest won't contain that SG initially, so it's probably not worth handling that case.

But what it does mean is that we probably don't need to bother with the Compare bit.

If you wanted to simplify further, we could make the return a `*[]byte` and `nil` means it wasn't modified. That might be stretching it though.

pkg/installmanager/installmanager.go
```diff
@@ -788,6 +843,51 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
...
+	logger.Infof("I applied it changed = %v", modified)
```
2uasimojo, line 887: leftover debug line?

abutcher: Yep

pkg/installmanager/installmanager.go
```diff
@@ -788,6 +843,46 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment) error {
 	return nil
 }
 
+// patchWorkerMachineSetManifest accepts a yaml manifest as []byte and patches the manifest to include an additional
+// security group filter if that manifest is the definition of a worker MachineSet. The security group is obtained from
+// an annotation (constants.ExtraWorkerSecurityGroupAnnotation) on the provided MachinePool and is the value of the
+// annotation (singular).
```
2uasimojo, line 849:
```suggestion
// annotation (singular).
// The first return value points to the modified manifest. It is `nil` if the manifest was not modified.
// The second return value indicates an error from which we should not proceed.
// Of note, it is not an error if `manifestBytes` represents valid yaml that is not a worker MachineSet.
```

2uasimojo:
/lgtm

Unhold when you're happy with local testing.

Thanks!

/cc @igoihman @vkareh

igoihman:
@renan-campos FYI

abutcher:
Additional security group was configured for my worker nodes in local testing of a BYO VPC install. New workers (from scaling the machinepool) also have the security group.
/unhold

PR #1744, opened 2022-04-11
AWS PrivateLink: "Spread" VPC endpoints across VPCs

2uasimojo:
Since we don't know the actual quota of endpoints per VPC, remove the arbitrary limit check, and create new endpoints in the least-used VPC. The result should be that we hit quota only when all the VPCs are full.

[HIVE-1803](https://issues.redhat.com//browse/HIVE-1803)


2uasimojo:
Working on UT.

PR #1749, opened 2022-04-25
Update ClusterPool Inventory enhancement

abraverm:
[HIVE-1367](https://issues.redhat.com//browse/HIVE-1367)


docs/enhancements/clusterpool-inventory.md
```diff
@@ -119,26 +116,32 @@ For the VSphere case, this allows the administrator to:
...
     ```
   
 ### Validation
 Webhook validation will ensure that for a clusterpool
-- if `spec.inventory.ClusterDeploymentCustomizations` is specified, it is not an empty list.
+- if `spec.inventory` is specified, it is not an empty list.
 
 Other validations include
-- If `clusterpool.spec.size` is greater than length of the inventory, we put status condition on the clusterpool saying size requirements cannot be satisfied. 
+- If `clusterpool.spec.size` is greater than length of the inventory, we change `InventoryValid` change to False on the clusterpool saying size requirements cannot be satisfied.
```
2uasimojo, line 139:
```suggestion
- If `clusterpool.spec.size` is greater than length of the inventory, we change `InventoryValid` condition to False on the clusterpool saying size requirements cannot be satisfied. 
```

but we don’t prevent the pool from consuming the inventory it has, right? So the status.size will never reach the spec.size, but the pool will still function?

2uasimojo (commented):
First pass. I haven’t correlated this against the code yet. Looking good so far though. 

@akhil-rane  care to have a glance?

docs/enhancements/clusterpool-inventory.md
```diff
@@ -119,26 +116,32 @@ For the VSphere case, this allows the administrator to:
   - op: replace
     path: metadata/name
     value: foo
+  - op: replace
```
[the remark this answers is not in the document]
abraverm: > Is this necessary even when DNS is preconfigured?

 yes, this is how VSphere/Openstack knows which IP to alllocate to which network device and machine.

> Can we deduplicate this with the section above?

 no - the section above explains the relationship of records in the DNS server to the IP addresses in the local cloud, while this section explains how this manifests in the install config and its patching.

docs/enhancements/clusterpool-inventory.md
abraverm: > but we don’t prevent the pool from consuming the inventory it has, right?

correct

> So the status.size will never reach the spec.size, but the pool will still function?

correct

abraverm:
@2uasimojo , do you want me to update this PR?

2uasimojo:
@abraverm If you are willing to continue with this, that would be great. If not, we can take over. Just let me know either way.

abraverm:
I will finish it.

PR #1750, opened 2022-04-25
dnsendpoint: Delay requeue for "not yet scraped"

2uasimojo:
To reduce thrashing, especially after a fresh deploy in a high-volume
environment, delay requeueing on `"name servers have not yet been
scraped"` for 15s.

[HIVE-1855](https://issues.redhat.com//browse/HIVE-1855)


2uasimojo:
Force merging around the freeze so we can validate that this has the desired effect before next week's prod push.

PR #1752, opened 2022-04-27
HIVE-1858: Fix decode VSphereMachineProviderSpec

abutcher:
[HIVE-1858](https://issues.redhat.com//browse/HIVE-1858)


2uasimojo:
Lovely. I tested locally, reproducing the reported error with the object from the new unit test via the old code, succeeded with the new. I like the LOC reduction; and this also lines up with how we're doing the same for GCP.

/lgtm

PR #1760, opened 2022-05-06
Metrics should allow for no conditions set

suhanime:
Since https://github.com/openshift/hive/pull/1712, metrics controller is
trying to set metrics for stopping and resuming clusters. It should
account for new cluster provisions, where cluster deployment conditions
are not initialized yet

xref: https://issues.redhat.com/browse/HIVE-1872

/assign @2uasimojo


2uasimojo:
Will manually merge this as soon as `verify` passes, as it's an urgent fix for osd.

PR #1765, opened 2022-05-11
Update hive install time metrics buckets

dlom:
Signed-off-by: Mark Old <mold@redhat.com>

xref https://issues.redhat.com/browse/HIVE-1869
xref https://issues.redhat.com/browse/HIVE-1870

### DIFF
--- a/pkg/controller/clusterdeployment/metrics.go
+++ b/pkg/controller/clusterdeployment/metrics.go
@@ -21,14 +21,14 @@ var (
 		prometheus.HistogramOpts{
 			Name:    "hive_cluster_deployment_install_job_duration_seconds",
 			Help:    "Distribution of the runtime of completed install jobs.",
-			Buckets: []float64{60, 300, 600, 1200, 1800, 2400, 3000, 3600},
+			Buckets: []float64{1800, 2400, 3000, 3600, 4500, 5400, 7200},
 		},
 	)
 	metricInstallDelaySeconds = prometheus.NewHistogram(
 		prometheus.HistogramOpts{
 			Name:    "hive_cluster_deployment_install_job_delay_seconds",
 			Help:    "Time between cluster deployment creation and creation of the job to install/provision the cluster.",
-			Buckets: []float64{30, 60, 120, 300, 600, 1200, 1800},
+			Buckets: []float64{60, 120, 180, 240, 300, 600, 1200, 1800, 2700, 3600},
 		},
 	)
 	metricImageSetDelaySeconds = prometheus.NewHistogram(


pkg/controller/clusterdeployment/metrics.go
```diff
@@ -21,14 +21,14 @@ var (
...
-			Buckets: []float64{60, 300, 600, 1200, 1800, 2400, 3000, 3600},
+			Buckets: []float64{1800, 2400, 3000, 3600, 4500, 5400, 7200},
 		},
 	)
 	metricInstallDelaySeconds = prometheus.NewHistogram(
 		prometheus.HistogramOpts{
 			Name:    "hive_cluster_deployment_install_job_delay_seconds",
 			Help:    "Time between cluster deployment creation and creation of the job to install/provision the cluster.",
-			Buckets: []float64{30, 60, 120, 300, 600, 1200, 1800},
+			Buckets: []float64{60, 120, 180, 240, 300, 600, 1200, 1800, 2700, 3600},
```
suhanime, line 31: @newtonheath Even for the 99th percentile - we don't exceed 600. I see the point of adding more buckets for <600, but why do we need more upwards of 3 times the ceiling value?
image

pkg/controller/clusterdeployment/metrics.go
```diff
@@ -21,14 +21,14 @@ var (
 		prometheus.HistogramOpts{
 			Name:    "hive_cluster_deployment_install_job_duration_seconds",
 			Help:    "Distribution of the runtime of completed install jobs.",
-			Buckets: []float64{60, 300, 600, 1200, 1800, 2400, 3000, 3600},
+			Buckets: []float64{1800, 2400, 3000, 3600, 4500, 5400, 7200},
```
suhanime, line 24: Same comments as below for this one @newtonheath - we need more buckets between 1800 and 3600, and maybe just a value below and/or above the lower and upper thresholds 
image

pkg/controller/clusterdeployment/metrics.go
newtonheath: We've had over 30 minutes for hive while I have been looking, so I'd like double that, and I threw one in between. You must be missing some data as I see this which is 19 minutes, 1140 seconds

Screen Shot 2022-05-11 at 4 27 08 PM
.

pkg/controller/clusterdeployment/metrics.go
newtonheath: Also seen installer higher than 1 hour more than 30 days ago. Again, some more detail above that can help us how bad it get above the threshold as we investigate

image

suhanime: Alright, let's push this in, provided we remove some buckets on a later date once we have a good idea of the threshold

suhanime (commented):
/lgtm

PR #1767, opened 2022-05-12
Further filter ODF security group filter by VPC ID

abutcher:
Security group names may be the same within an AWS account (belonging to different VPCs) which means filtering on the name alone is insufficient. The VPC ID is now used to further filter the correct security group when configuring the ODF security group in worker machineset manifests. The VPC ID is discovered by using the AWS client to query subnets from the worker MachinePool. BYO VPC is required when using ODF so it is expected that the worker MachinePool platform will contain subnets we can use to discover the VPC ID.

[HIVE-1874](https://issues.redhat.com//browse/HIVE-1874)

### DIFF
--- a/pkg/controller/machinepool/awsactuator.go
+++ b/pkg/controller/machinepool/awsactuator.go
@@ -227,9 +227,17 @@ func (a *AWSActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 		}
 	}
 
+	var vpcID string
+	if metav1.HasAnnotation(pool.ObjectMeta, constants.ExtraWorkerSecurityGroupAnnotation) {
+		vpcID, err = GetVPCIDForMachinePool(a.awsClient, pool)
+		if err != nil {
+			return nil, false, err
+		}
+	}
+
 	// Re-use existing AWS resources for generated MachineSets.
 	for _, ms := range installerMachineSets {
-		a.updateProviderConfig(ms, cd.Spec.ClusterMetadata.InfraID, pool)
+		a.updateProviderConfig(ms, cd.Spec.ClusterMetadata.InfraID, pool, vpcID)
 	}
 
 	return installerMachineSets, true, nil
@@ -300,7 +308,7 @@ func decodeAWSMachineProviderSpec(rawExt *runtime.RawExtension, scheme *runtime.
 // updateProviderConfig modifies values in a MachineSet's AWSMachineProviderConfig.
 // Currently we modify the AWSMachineProviderConfig IAMInstanceProfile, Subnet and SecurityGroups such that
 // the values match the worker pool originally created by the installer.
-func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, infraID string, pool *hivev1.MachinePool) {
+func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, infraID string, pool *hivev1.MachinePool, vpcID string) {
 	providerConfig := machineSet.Spec.Template.Spec.ProviderSpec.Value.Object.(*awsproviderv1beta1.AWSMachineProviderConfig)
 
 	// TODO: assumptions about pre-existing objects by name here is quite dangerous, it's already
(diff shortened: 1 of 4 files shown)


abutcher:
Verified that day0 and day2 `MachineSets` get the updated filter.

pkg/installmanager/installmanager.go
```diff
@@ -800,6 +802,42 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 	// ExtraWorkerSecurityGroupAnnotation has been set in annotations for the worker MachinePool.
 	// For details, see HIVE-1802.
 	if workerMachinePool != nil && metav1.HasAnnotation(workerMachinePool.ObjectMeta, constants.ExtraWorkerSecurityGroupAnnotation) {
+		if cd.Spec.Platform.AWS == nil {
+			return errors.New("ExtraWorkerSecurityGroup annotation cannot be confiured for non-AWS MachinePool")
```
2uasimojo, line 806:
```suggestion
			return errors.New("ExtraWorkerSecurityGroup annotation cannot be configured for non-AWS MachinePool")
```

pkg/controller/machinepool/awsactuator.go
```diff
@@ -330,7 +333,32 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
 	// NOTE: modifying the security group of an existing MachineSet will NOT result in updates to the
 	// corresponding instances in AWS and will only be configured for newly created instances.
 	if metav1.HasAnnotation(pool.ObjectMeta, constants.ExtraWorkerSecurityGroupAnnotation) {
+		// Add the security group name obtained from the ExtraWorkerSecurityGroupAnnotation
+		// annotation found on the MachinePool to the existing tag:Name filter in the worker
+		// MachineSet manifest
 		providerConfig.SecurityGroups[0].Filters[0].Values = append(providerConfig.SecurityGroups[0].Filters[0].Values, pool.Annotations[constants.ExtraWorkerSecurityGroupAnnotation])
+
+		if pool.Spec.Platform.AWS == nil {
+			return errors.New("ExtraWorkerSecurityGroup annotation cannot be confiured for non-AWS MachinePool")
```
2uasimojo, line 342:
```suggestion
			return errors.New("ExtraWorkerSecurityGroup annotation cannot be configured for non-AWS MachinePool")
```

pkg/installmanager/installmanager_test.go
```diff
@@ -854,7 +855,9 @@ spec:
 				assert.NoError(t, err, "expected to be able to decode AWSMachineProviderConfig")
 				awsMachineTemplate, _ := awsMachineTemplateObj.(*awsproviderv1beta1.AWSMachineProviderConfig)
 
-				assert.Contains(t, awsMachineTemplate.SecurityGroups[0].Filters[0].Values, "test-security-group", "expected test-security-group to be configured within AWSMachineProviderConfig")
+				assert.Contains(t, awsMachineTemplate.SecurityGroups[0].Filters[0].Values, "test-security-group", "expected test-security-group to be configured within Security Group Filters in AWSMachineProviderConfig")
+				assert.Contains(t, awsMachineTemplate.SecurityGroups[0].Filters[1].Name, "vpc-id", "expected an vpc-id named filter to be configured within Security Group Filters in AWSMachineProviderConfig")
```
2uasimojo, line 859:
```suggestion
				assert.Equals(t, awsMachineTemplate.SecurityGroups[0].Filters[1].Name, "vpc-id", "expected an vpc-id named filter to be configured within Security Group Filters in AWSMachineProviderConfig")
```
?

pkg/controller/machinepool/awsactuator_test.go
```diff
@@ -453,6 +454,39 @@ func TestAWSActuator(t *testing.T) {
 				},
 			},
 		},
+		{
```
2uasimojo, line 457: I'm going to need to stare at this test case for a minute.

Meanwhile: Do we need to add test cases for the error paths you've introduced?

2uasimojo (commented):
Partial review

pkg/controller/machinepool/awsactuator_test.go
abutcher: I could do error path tests for no subnets configured (invalid) or no subnets returned. This test case also needs to validate the filter, adding that.

abutcher: Added the validation piece to this test which is checking the filters against expected.

abutcher: Added a few error path tests and made the other pieces line up.

abutcher:
Removing hold after retesting that day0/day2 is configured in machinesets on the cluster.
/unhold

pkg/controller/machinepool/awsactuator.go
```diff
@@ -227,9 +228,30 @@ func (a *AWSActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 		}
 	}
 
+	var vpcID string
+	if metav1.HasAnnotation(pool.ObjectMeta, constants.ExtraWorkerSecurityGroupAnnotation) {
+		if len(pool.Spec.Platform.AWS.Subnets) == 0 {
```
2uasimojo, line 233: 1. I'm curious why you decided to do this out here rather than within the `updateProviderConfig` func where you've already got a branch for the annotation.
2. Can the discovery portion of this be factored out into a func that can be reused between this and the installmanager?

abutcher: 1. I have this outside `updateProviderConfig` to avoid making an API call per machine set.
2. Yes.

2uasimojo: > 1. I have this outside `updateProviderConfig` to avoid making an API call per machine set.

D'oh, I see that now, thanks.

> 2. Yes.

:+1:

pkg/controller/machinepool/awsactuator.go
```diff
@@ -562,3 +575,29 @@ func (a *AWSActuator) validateSubnets(subnets map[string]ec2.Subnet, pool *hivev
...
+	if err != nil {
+		return "", err
+	}
+	// An error should be produced if no subnets are returned. This is here as a safety net.
+	if len(subnetsOutput.Subnets) == 0 {
+		return "", errors.Errorf("DescribeSubnets unexpectedly returned no results for subnet ID %s", subnetID)
+	}
+	vpcID := *subnetsOutput.Subnets[0].VpcId
+	if vpcID == "" {
+		return "", errors.Errorf("DescribeSubnets unexpectedly returned a subnet without a VPC ID %s", subnetID)
```
2uasimojo, line 600: nit:
```suggestion
		return "", errors.Errorf("DescribeSubnets unexpectedly returned subnet %s without a VPC ID", subnetID)
```

2uasimojo (commented):
/lgtm

PR #1774, opened 2022-05-18
Fix default credentials path for alibaba used by hiveutil

akhil-rane:
Currently, default is ~/.alibaba/credentials. This PR changes
the default path to ~/.alibabacloud/credentials and makes it
consistent with the installer and ccoctl


akhil-rane:
assign @abutcher

PR #1780, opened 2022-05-26
Ensure correct time reported for transition metrics

suhanime:
Fetching cluster deployment conditions returns a pointer, and that
pointer is subject to change as other controllers update the CR, and
those conditions get sorted into a different order.

A bug was caught while testing powerstate transitioning metrics, where,
by the time we reported the hibernation transition metric, the ready
condition was pointing to some other condition of cluster deployment.
This commit ensures that before reporting the metric, we re-fetch the
relevant condition

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -368,7 +368,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		if shouldStopMachines(cd, hibernatingCondition) {
 			return r.stopMachines(cd, cdLog)
 		}
-		return r.checkClusterStopped(cd, readyCondition, false, cdLog)
+		return r.checkClusterStopped(cd, false, cdLog)
 	}
 	// If we get here, we're not supposed to be hibernating
 	if isFakeCluster {
@@ -388,7 +388,7 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	if shouldStartMachines(cd, hibernatingCondition, readyCondition) {
 		return r.startMachines(cd, cdLog)
 	}
-	return r.checkClusterRunning(cd, syncSetsApplied, cdLog, readyCondition, hibernatingCondition)
+	return r.checkClusterRunning(cd, syncSetsApplied, cdLog, readyCondition)
 }
 
 func (r *hibernationReconciler) startMachines(cd *hivev1.ClusterDeployment, logger log.FieldLogger) (reconcile.Result, error) {
@@ -451,8 +451,7 @@ func (r *hibernationReconciler) stopMachines(cd *hivev1.ClusterDeployment, logge
 	return reconcile.Result{}, err
 }
 
-func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment, readyCondition *hivev1.ClusterDeploymentCondition,
-	expectRunning bool, logger log.FieldLogger) (reconcile.Result, error) {
+func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment, expectRunning bool, logger log.FieldLogger) (reconcile.Result, error) {
 	actuator := r.getActuator(cd)
 	if actuator == nil {
 		logger.Warning("No compatible actuator found to check machine status")
@@ -493,8 +492,7 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 			return reconcile.Result{}, err
 		}
 		// logging with time since ready condition was set to StoppingOrHibernating state
-		logCumulativeMetric(hivemetrics.MetricClusterHibernationTransitionSeconds, cd,
-			time.Since(readyCondition.LastTransitionTime.Time).Seconds())
+		logCumulativeMetric(hivemetrics.MetricClusterHibernationTransitionSeconds, cd, hivev1.ClusterReadyCondition, logger)
 		// Clear entry from currently stopping and waiting for cluster operators clusters if exists
 		deleteTransitionMetric(hivemetrics.MetricStoppingClustersSeconds, cd)
 		deleteTransitionMetric(hivemetrics.MetricWaitingForCOClustersSeconds, cd)
@@ -503,7 +501,7 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 }
 
 func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment, syncSetsApplied bool, logger log.FieldLogger,
-	readyCondition *hivev1.ClusterDeploymentCondition, hibernatingCondition *hivev1.ClusterDeploymentCondition) (reconcile.Result, error) {
+	readyCondition *hivev1.ClusterDeploymentCondition) (reconcile.Result, error) {
 	actuator := r.getActuator(cd)
 	if actuator == nil {
 		logger.Warning("No compatible actuator found to check machine status")
@@ -639,8 +637,7 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
 			return reconcile.Result{}, err
 		}
 		// logging with time since hibernating condition was set to ResumingOrRunning state
-		logCumulativeMetric(hivemetrics.MetricClusterReadyTransitionSeconds, cd,
-			time.Since(hibernatingCondition.LastTransitionTime.Time).Seconds())
+		logCumulativeMetric(hivemetrics.MetricClusterReadyTransitionSeconds, cd, hivev1.ClusterHibernatingCondition, logger)
 		// Clear entry from currently resuming clusters if exists
 		deleteTransitionMetric(hivemetrics.MetricResumingClustersSeconds, cd)
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -482,7 +484,7 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 }
 
 func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment, logger log.FieldLogger,
-	readyCondition *hivev1.ClusterDeploymentCondition, hibernatingCondition *hivev1.ClusterDeploymentCondition) (reconcile.Result, error) {
+	readyCondition *hivev1.ClusterDeploymentCondition) (reconcile.Result, error) {
```
abutcher, line 487: Could the `readyCondition` parameter here suffer from the same issue?

abutcher: I see that the `reason` of the condition is checked and this is why we can ensure we're dealing with the right one.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -471,7 +471,9 @@ func (r *hibernationReconciler) checkClusterStopped(cd *hivev1.ClusterDeployment
 		if err := r.updateClusterDeploymentStatus(cd, logger); err != nil {
 			return reconcile.Result{}, err
 		}
-		// logging with time since ready condition was set to StoppingOrHibernating state
+		// Logging with time since ready condition was set to StoppingOrHibernating state
+		// Fetch ready condition again to ensure it points at the correct value
+		readyCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterReadyCondition)
```
2uasimojo, line 476: Note to self: The bug is that L467's `setCDCondition` calls `SetClusterDeploymentConditionWithChangeCheck` which calls `SortClusterDeploymentConditions`. An existing condition pointer points to some index in the array of conditions, so if that array is reordered, the condition at that index can change.

2uasimojo (commented):
@abutcher and I batted this around for a while. We really have a systemic problem here. We can fix these two cases, but it would be great if we could figure out a way to lock down the code so it's impossible (or at least harder?) to make this same mistake in the future. So far we haven't come up with anything workable. Ideas welcome.

However, to solve this specific case, it would be preferable to hide the retrieval of the condition within `logCumulativeMetric`. This would help prevent future-us from e.g. deciding `readyCondition` would be useful somewhere above wherever we're calling `setCDCondition`, moving it, and reintroducing the bug. So like `logCumulativeMetric(metric, cd, conditionName)`.

The same should be done for `shouldStopMachines`, `shouldStartMachines`, and `checkClusterRunning`, though conceivably in a separate PR.

We're also missing error checks (though this is latent): It *should* be impossible for any of our `FindClusterDeploymentCondition`s to return `nil`, since we `InitializeClusterDeploymentConditions` with exactly those conditions; but it's not very future-proof. If we added a condition and forgot to initialize it (for example) we would crash the controller with a nil pointer panic. SO it would be nice if the new version of `logCumulativeMetric` included that nil check. It could return an `error` in that case and bubble it up to Reconcile(), which would of course cause a hot loop -- that's arguably better than crashing the controller. But it may instead be possible for `logCumulativeMetric` to make a reasonable assumption/decision in that case, such as logging a warning and simply not observing the metric. (Similar philosophy for those other three methods.) The concern is how we would ultimately discover that bug to fix it. Thoughts?

Finally: We should definitely figure out a way to cover at least these two code paths with unit tests. Shouldn't be too hard to do.

Thanks!

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: Nope, it's just by luck that no path to this point runs through `setCDCondition` without immediately returning.

suhanime:
Added condition check and warning for nil ref + skip reporting metric for `logCumulativeMetric`. Toyed with the idea of implementing units for the func with a mock for `ShouldLogDuration` and then arrived at the conclusion that
-  WRT metrics, we always strive to report accurately to the best of our abilities, so skipping its reporting with a warning is definitely a better recourse than being responsible for a hot loop 
- We have already updated ClusterDeployment at this point - we only touch this code path once everything has succeeded

I think there are 2 big problems - 
- ClusterDeployment conditions getting sorted on every update but returning a pointer that is unreliable
  We can probably consider returning the entire condition instead of a pointer, which works as long as no other controller is touching that particular condition at the same time - so there are some outliers like Unreachable condition iirc. But this simple fix will definitely fix issues beyond the hibernation controller
- We update ClusterDeployment multiple times during a single reconcile sometimes
  This is another latent problem we should eventually solve - not just for reduced number of updates but also ensuring whatever information considered during a reconcile loop doesn't get outdated immediately. Although in my mind, it is acceptable to correct issues in the corresponding reconcile, I haven't completely thought this idea through

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -632,7 +630,16 @@ func deleteTransitionMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterD
...
 
 // logCumulativeMetric should be used to log cumulative metrics
-func logCumulativeMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment, time float64) {
+// Time to log is calculated as the time lapsed since the last transition time of condition mentioned
+func logCumulativeMetric(metric *prometheus.HistogramVec, cd *hivev1.ClusterDeployment,
+	conditionType hivev1.ClusterDeploymentConditionType, logger log.FieldLogger) {
+	condition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, conditionType)
+	// This shouldn't happen if conditions have been properly initialized
+	if condition == nil{
+		logger.Warningf("cannot find %s condition, logging of %s metric skipped", conditionType, &metric)
```
2uasimojo: A few things here:
- This represents a code bug for sure, so I would a) definitely make it an Error rather than a Warning; b) probably even put a loud eyecatcher in the message, e.g. `THIS IS A BUG!`.
- Isn't `&metric` a pointer to a pointer? So you'll end up with an unhelpful hex address in the message? Did you mean `*metric`?
- Prefer structured logging over embedded formats: `logger.WithField("conditionType", conditionType).WithField("metric", *metric).Error("cannot find condition, logging of metric skipped")`

2uasimojo (commented):
> implementing units for the func

I was actually talking about a unit test that validates that we do the timestamp math on the correct condition. It would be set up so that it fails without these code changes. For example, it would initialize the CD conditions in such an order that we know the Ready condition will get moved by `SortClusterDeploymentConditions`; and make the `lastTransitionTime` of all the other conditions different from that of the Ready condition so that the metric is guaranteed to be wrong if we use the wrong condition for the calculation.

> We can probably consider returning the entire condition instead of a pointer

I'm not sure if this is what you meant, but I actually really like the idea of changing the `readyCondition`/`hibernatingCondition` vars in the main loop to be `DeepCopy`s of the actual condition contents rather than a pointer into the `cd.Status.Conditions` slice. Of course, then we still need to be aware of the possibility of those copies going stale if there are code paths that update the condition in the CD. But I think that may be the lesser evil, and easier to discover, than the original bug.

> no other controller is touching that particular condition at the same time

Not worried about this. As long as we only `Get()` the CD once in a Reconcile, we're implicitly protected from asynchronous changes by `resourceVersion`.

> We update ClusterDeployment multiple times during a single reconcile sometimes

Do you mean that we update the local object multiple times; or that we actually issue `[.Status()].Update()`s multiple times in a single code path through the reconciler? As long as we're careful, the former is copacetic to reduce the number of API calls as you say. But the latter should be avoided as a best practice: any path that updates an object should immediately (ish) return; and the Reconcile should be idempotent such that a subsequent run-through will (bypass the first update, seeing it's already done, and) do the second update, etc.

2uasimojo:
...otherwise, the change looks good!

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -1270,3 +1272,138 @@ func Test_timeBeforeClusterSyncCheck(t *testing.T) {
...
+				testcd.WithCondition(hivev1.ClusterDeploymentCondition{
+					Type:               hivev1.ClusterReadyCondition,
+					Status:             corev1.ConditionFalse,
+					Reason:             hivev1.ReadyReasonStoppingOrHibernating,
+					LastTransitionTime: metav1.Time{Time: time.Now().Add(-2 * time.Hour)},
+				})).Build(),
+			cs: csBuilder.Build(),
+			setupActuator: func(actuator *mock.MockHibernationActuator, cd *hivev1.ClusterDeployment) {
+				// Simulate changing the order of cluster deployment conditions mid reconcile
+				controllerutils.SetClusterDeploymentCondition(cd.Status.Conditions,
```
suhanime, line 1320: When I debugged with this changes, setting the condition here didn't work - this condition wasn't present at all on the CD. Looking for a better way to enforce it

suhanime (commented):
@2uasimojo / @abutcher Can you help nudge me in the right direction?

suhanime:
Splitting off the bug fix from the unit tests for the situation - the latter will be addressed in https://issues.redhat.com/browse/HIVE-1947

2uasimojo (commented):
/lgtm

PR #1781, opened 2022-05-27
e2e: Save hive pod descriptions

2uasimojo:
Some e2e failures happen because the pods for one or more deployments
don't start at all. In those cases, we don't get any logs, so we can't
figure out what went wrong.

This commit captures `oc describe po` for both the operator and target
namespaces, as those descriptions often contain hints (e.g. events) as
to what went wrong.


2uasimojo:
/test e2e

infra post failed after a successful run

PR #1784, opened 2022-06-01
certificates.k8s.io/v1beta1 API deprecated in Kubernetes v1.19 and removed before v1.22

pjanouse:
This helper script was missed in https://github.com/openshift/hive/pull/1494 which updated same for the production code.


2uasimojo:
Thanks for this @pjanouse!

/assign @abutcher 

Andrew has agreed to test this via KIND, but probably won't get to it until next week.

abutcher:
Just tested with kind and the updates work for me, thank you!
/lgtm

PR #1792, opened 2022-06-10
powerstate-pause annotation: skip hibernation controller

2uasimojo:
Add a (secret, "unsupported") `hive.openshift.io/powerstate-pause`
annotation. When `"true"`, it causes the hibernation controller to
ignore the state of the cluster's machines with respect to
`cd.spec.powerState`; and set `cd.status.powerState` and the Ready and
Hibernating status conditions to `"Unknown"`.

[HIVE-1755](https://issues.redhat.com//browse/HIVE-1755)

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -195,6 +195,26 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, nil
 	}
 
+	// Secret toggle allowing consumers to e.g. perform manual maintenance on machines without the
+	// controller attempting to reconcile desired CD.Spec.PowerState.
+	if paused, err := strconv.ParseBool(cd.Annotations[constants.PowerStatePauseAnnotation]); err == nil && paused {
+		cdLog.Info("skipping reconcile of PowerState as the powerstate-pause annotation is set")
+		changed := false
+		if cd.Status.PowerState != hivev1.ClusterPowerStateUnknown {
+			changed = true
+			cd.Status.PowerState = hivev1.ClusterPowerStateUnknown
+		}
+		msg := "the powerstate-pause annotation is set"
+		changed = r.setCDCondition(cd, hivev1.ClusterHibernatingCondition, hivev1.HibernatingReasonPowerStatePaused, msg, corev1.ConditionUnknown, cdLog) ||
+			changed
+		changed = r.setCDCondition(cd, hivev1.ClusterReadyCondition, hivev1.ReadyReasonPowerStatePaused, msg, corev1.ConditionUnknown, cdLog) ||
+			changed
+		if changed {
+			return reconcile.Result{}, r.updateClusterDeploymentStatus(cd, cdLog)
+		}
+		return reconcile.Result{}, nil
+	}
+
 	hibernatingCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterHibernatingCondition)
 	readyCondition := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.ClusterReadyCondition)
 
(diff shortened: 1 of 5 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -572,7 +604,7 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
 		// Make sure we wait long enough for operators to start/settle:
 		if readyCondition.Reason == hivev1.ReadyReasonPausingForClusterOperatorsToSettle &&
 			time.Since(readyCondition.LastProbeTime.Time) < constants.ClusterOperatorSettlePause {
-			remainingPause := constants.ClusterOperatorSettlePause - time.Now().Sub(readyCondition.LastProbeTime.Time)
+			remainingPause := constants.ClusterOperatorSettlePause - time.Since(readyCondition.LastProbeTime.Time)
```
2uasimojo, line 607: NB: latent unrelated squiggly yellow line squash

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -212,6 +212,26 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
...
+		changed := false
+		if cd.Status.PowerState != hivev1.ClusterPowerStateUnknown {
+			changed = true
+			cd.Status.PowerState = hivev1.ClusterPowerStateUnknown
+		}
+		msg := "the powerstate-pause annotation is set"
+		changed = r.setCDCondition(cd, hivev1.ClusterHibernatingCondition, hivev1.HibernatingReasonPowerStatePaused, msg, corev1.ConditionUnknown, cdLog) ||
+			changed
+		changed = r.setCDCondition(cd, hivev1.ClusterReadyCondition, hivev1.ReadyReasonPowerStatePaused, msg, corev1.ConditionUnknown, cdLog) ||
+			changed
```
suhanime, line 228: These conditions with PowerStatePaused is more informative than Unsupported (if someone erroneously puts the annotation true for Unsupported clusters - it is easy for them to be mistaken) - so I would prefer we move the annotation based decision making above the checking (and setting) of unsupported clusters

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -212,6 +212,26 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, r.updateClusterDeploymentStatus(cd, cdLog)
 	}
 
+	// Secret toggle allowing consumers to e.g. perform manual maintenance on machines without the
+	// controller attempting to reconcile desired CD.Spec.PowerState.
+	if paused, err := strconv.ParseBool(cd.Annotations[constants.PowerStatePauseAnnotation]); err == nil && paused {
```
suhanime, line 217: Is someone had already applied this annotation but subsequently removes it, as per current logic, hibernation controller would attempt to run the cluster - I have not fully thought out the scenario, but we should have a way of either working with Unknown powerstate and powerstate paused condition reasons, or give an error early

pkg/controller/hibernation/hibernation_controller.go
suhanime: I saw in the subsequent commit that you are aware this will happen - I'm curious - it didn't need any changes in the shouldStartMachines logic? Maybe a unit test to simulate the scenario would help?

2uasimojo: This is Unknown, not Unsupported. 

Yes, I needed to finagle the starting logic a bit. It still “worked” without it, but it was slow and the state transitions looked odd. 

Yes, I’ll see about adding a test.

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: Yup, I ran through all off that, which is how I came up with the second commit. Again, it did work, but the timing and transitions looked odd. With the second commit it’s smoother.

pkg/controller/hibernation/hibernation_controller.go
suhanime: I meant the subsequent reconcile will set them to Unsupported [here](https://github.com/openshift/hive/blob/fa498060121355f35e620b09eb70e0ee4dc45a47/pkg/controller/hibernation/hibernation_controller.go#L201-L213)

2uasimojo: Ah, I get you now. I deliberately put the annotation check *after* the unsupported check because I want unsupported to take precedence. Does that make sense?

2uasimojo: Added a test to prove "unsupported" takes precedence over "paused".

2uasimojo: Oh. If I pause an unsupported cluster, the controller will thrash. That's not good.

2uasimojo: Fixed. Also reversed the precedence as discussed offline. And added UTs for unpausing: one to resume, one to hibernate.

PR #1793, opened 2022-06-10
Explicit permissions in operator ClusterRole

2uasimojo:
[SEC-OPS-DEP-04](https://source.redhat.com/groups/public/product-security/content/product_security_wiki/operators_secure_deployment_guide)
recommends explicitly enumerating permissions in an RBAC (Cluster)Role
rather than using wildcards (`"*"`). We believe wildcarding permissions
for groups we own is appropriate; but fix it elsewhere (which turns out
to be only one place).

[HIVE-1919](https://issues.redhat.com//browse/HIVE-1919)

### DIFF
--- a/config/operator/operator_role.yaml
+++ b/config/operator/operator_role.yaml
@@ -43,7 +43,13 @@ rules:
   - namespaces
   - persistentvolumeclaims
   verbs:
-  - "*"
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
 - apiGroups:
   - apiregistration.k8s.io
   resources:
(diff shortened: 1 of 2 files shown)


config/operator/operator_role.yaml
```diff
@@ -43,7 +43,13 @@ rules:
   - namespaces
   - persistentvolumeclaims
```
akhil-rane, line 44: Does hive operator deal with `persistentvolumeclaims`? Never heard of that. I might be wrong

2uasimojo: Opened HIVE-1920 to have a closer look at this and other things.

PR #1798, opened 2022-06-17
e2e: Bail early on ProvisionStopped

2uasimojo:
e2e was detecting ClusterDeployment condition ProvisionFailed to decide
whether to give up early. ProvisionStopped is appropriate also (or
perhaps instead?) so add it.


2uasimojo:
/hold

retest after https://github.com/openshift/hive/pull/1802 lands

PR #1799, opened 2022-06-17
e2e: increase targetNamespace timeout 30s=>3m

2uasimojo:
30s wasn't a reasonable amount of time to expect to get all the way from
Deployment to a completed reconcile loop, and we've been seeing a lot of
CI flakes as a result.


2uasimojo:
cc @abraverm -- this was probably the cause of all those flakes you started seeing.

2uasimojo:
/test e2e

but this looks like it might be real?

2uasimojo:
Force-merging this so https://github.com/openshift/hive/pull/1802 can land (and unwedge CI) without annoying flakes.

PR #1802, opened 2022-06-17
Patch installconfig generated by clusterresource builder to remove AWS metadataService field.

abutcher:
This field is not omitempty from upstream and an empty install config object cannot be unmarshalled by the installer in previous versions.


2uasimojo (commented):
/lgtm

PR #1804, opened 2022-06-21
Patch installconfig generated by clusterresource builder to remove Azure osImage field.

abutcher:
/hold


pkg/controller/machinepool/azureactuator.go
```diff
@@ -87,7 +87,7 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
 	}
 
 	var hyperVGen string
-	if osImage := pool.Spec.Platform.Azure.OSImage; osImage.Publisher != "" {
+	if osImage := pool.Spec.Platform.Azure.OSImage; osImage != nil {
```
2uasimojo, line 90:
```suggestion
	if osImage := pool.Spec.Platform.Azure.OSImage; osImage != nil && osImage.Publisher != "" {
```
right?

2uasimojo: I guess until the installer side is fixed, the effect will be the same. But the suggested change will be necessary once the installer is fixed, and doing it now will help us not forget it :)

abutcher: Updated!

2uasimojo:
/lgtm

unhold at will

PR #1806, opened 2022-06-22
Update default AWS volumeType chosen by hiveutil from gp2 to gp3.

abutcher:
https://github.com/openshift/installer/pull/5925 introduces checks which disallow our previous default of a gp2 volumeType with 100 IOPS as it was an incorrect configuration. From the linked change, IOPS may now only be set for io1, io2, & gp3 volume types.


akhil-rane (commented):
/lgtm

PR #1803, opened 2022-06-21
Update to go 1.18

dlom:

PR #1809, opened 2022-06-22
Add ClusterSyncFailingSeconds metric

suhanime:
- Add gauge "hive_clustersync_failing_seconds" that logs time since the lastTransitionTime of the clusterSync's Failed condition with the label namespaced_name as $namespace/$name
- Make it optional and log it conditionally if the time since it has been failing is more than the threshold mentioned
- Skip reporting for Unreachable clusters
- Clear the metric when the clusterSync succeeds
- Clear the metric when the clusterSync is deleted
- Clear the metric when the clusterDeplyment is deleted

/assign @2uasimojo

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -512,11 +512,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			// removed the finalizer.
 			clearDeprovisionUnderwaySecondsMetric(cd, cdLog)
 			// Clear ClusterSyncFailing metric
-			clusterSync := &hiveintv1alpha1.ClusterSync{}
-			err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync)
-			if err != nil {
-				clustersync.ClearClusterSyncFailingSecondsMetric(clusterSync, cdLog)
-			}
+			clustersync.ClearClusterSyncFailingSecondsMetric(cd.Namespace, cd.Name, cdLog)
 
 			return reconcile.Result{}, nil
 		}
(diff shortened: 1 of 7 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -505,11 +506,51 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
...
+					corev1.ConditionTrue,
+					"MissingClusterSync",
+					"ClusterSync has not yet been created",
+					controllerutils.UpdateConditionIfReasonOrMessageChange,
+				)
+			}
+			// In case the clusterSync has been deleted at this point, clear the clusterSyncFailing metric
+			clustersync.ClearClusterSyncFailingSecondsMetric(clusterSync, cdLog)
+		} else {
+			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not get ClusterSync")
```
2uasimojo, line 539: By not erroring here, we're passing an empty clustersync around. Hopefully the worst that will happen is we'll clear a metric with an empty label, which should be a no-op, but it's not future-proof.

2uasimojo (commented):
I have looked at this.

The code to declare and observe the new metric looks good ✓

However, I'm leery of the refactor, especially in such a foundational controller, especially since it does change some logic flow/order... and I don't think we actually need to do it. IIUC it's mainly for the purpose of being able to pass the clustersync around; and we only need to do that so we can scrub the metric for it; and the only information we need to do that is the namespace and name. So let's instead take advantage of the fact that the clustersync namespace and name are identical to that of the CD ([ref](https://github.com/openshift/hive/blob/d6fe109616208e47b3b436652fdde697067e281b/pkg/controller/clustersync/clustersync_controller.go#L414-L415) -- we take advantage of this elsewhere too, so it's not likely to change) and not even bother passing the clustersync around.

suhanime:
> I have looked at this.
> 
> The code to declare and observe the new metric looks good ✓
> 
> However, I'm leery of the refactor, especially in such a foundational controller, especially since it does change some logic flow/order... and I don't think we actually need to do it. IIUC it's mainly for the purpose of being able to pass the clustersync around; and we only need to do that so we can scrub the metric for it; and the only information we need to do that is the namespace and name. So let's instead take advantage of the fact that the clustersync namespace and name are identical to that of the CD ([ref](https://github.com/openshift/hive/blob/d6fe109616208e47b3b436652fdde697067e281b/pkg/controller/clustersync/clustersync_controller.go#L414-L415) -- we take advantage of this elsewhere too, so it's not likely to change) and not even bother passing the clustersync around.

This makes way more sense - fixed

pkg/controller/metrics/metrics.go
```diff
@@ -398,44 +414,57 @@ func (mc *Calculator) Start(ctx context.Context) error {
...
+	duration, ok := mapMetricToDurationHistograms[metricToLog]
+	return ok && duration.Seconds() <= timeToLog
+}
+
+// ShouldLogGaugeDurationMetric decides whether the corresponding duration metric of type gauge should be logged. It
+// first checks if that optional metric has been opted into for logging, and the duration has crossed the threshold set.
+func ShouldLogGaugeDurationMetric(metricToLog *prometheus.GaugeVec, timeToLog float64) bool {
+	duration, ok := mapMetricToDurationGauges[metricToLog]
 	return ok && duration.Seconds() <= timeToLog
 }
```
2uasimojo, line 465: NB: I think we'll be able to collapse the maps and these funcs using generics once https://github.com/openshift/hive/pull/1803 lands :)

2uasimojo (commented):
/lgtm

Thanks @suhanime!

PR #1810, opened 2022-06-24
Ensure go.mod matches apis/go.mod

2uasimojo:
Add a utility that compares statements in go.mod and apis/go.mod. If
matching module paths are found, the utility determines whether the
versions match, printing a human-readable report and exiting nonzero iff
*any* versions *do not* match.

Add a make rule, `make modcheck`, to run it.

Plumb `modcheck` into `verify`.

[HIVE-1941](https://issues.redhat.com//browse/HIVE-1941)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -320,6 +320,11 @@ lint: install-tools
 # Remove the golangci-lint from the verify until a fix is in place for permisions for writing to the /.cache directory.
 #verify: lint
 
+.PHONY: modcheck
+modcheck:
+	go run ./hack/modcheck.go
+verify: modcheck
+
 .PHONY: install-tools
 install-tools:
 	go install $(GO_MOD_FLAGS) github.com/golang/mock/mockgen
--- a/apis/go.mod
+++ b/apis/go.mod
@@ -9,9 +9,9 @@ require (
 )
 
 require (
-	github.com/go-logr/logr v1.2.0 // indirect
+	github.com/go-logr/logr v1.2.2 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
-	github.com/google/gofuzz v1.1.0 // indirect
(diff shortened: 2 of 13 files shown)


2uasimojo:
Expect this to fail `verify` on the new check. A subsequent commit in this PR will have to sync up the versions.

hack/modsync.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	return ret
+}
+
+func cmp1(root, apis map[string]string) bool {
+	insync := true
+	for path, rootver := range root {
+		apisver, ok := apis[path]
+		if !ok {
+			fmt.Printf("\t(path in root but not apis: %s)\n", path)
```
2uasimojo, line 91: This can probably go away -- too noisy.

hack/modsync.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	return insync
+}
+
+func cmp2(root, apis map[string]replacement) bool {
+	insync := true
+	for path, rootrepl := range root {
+		apisrepl, ok := apis[path]
+		if !ok {
+			fmt.Printf("\t(path in root but not apis: %s)\n", path)
```
2uasimojo, line 109: ditto

2uasimojo (commented):
docs?
CLI options like `--help`, verbosity?

abutcher:
:+1: Very cool. We have a [dependency management section in our development docs](https://github.com/openshift/hive/blob/master/docs/developing.md#dependency-management) which would probably be an okay place for docs if needed.

2uasimojo:
> We have a [dependency management section in our development docs](https://github.com/openshift/hive/blob/master/docs/developing.md#dependency-management) which would probably be an okay place for docs if needed.

Good call, done.

Added a second commit to sync the modules. This should be ready for review now.

/assign @abutcher

PR #1812, opened 2022-06-28
Updates to ibmclient authenticator.

abutcher:
IBM Cloud deprovision was broken by https://github.com/openshift/hive/pull/1764 and this PR brings the client back in line with changes in the installer's client. I attempted to consume the installer client directly but we'll need a way to instantiate the client with a secret or just the `apiKey` string in addition to needing functions related to stopping/starting/listing instances.

[HIVE-1951](https://issues.redhat.com//browse/HIVE-1951)
/assign @2uasimojo

### DIFF
--- a/pkg/ibmclient/client.go
+++ b/pkg/ibmclient/client.go
@@ -45,10 +45,11 @@ type API interface {
 
 // Client makes calls to the IBM Cloud API.
 type Client struct {
+	APIKey string
+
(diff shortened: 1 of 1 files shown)


2uasimojo:
/lgtm

See https://github.com/openshift/installer/pull/5736 and the linked bug to understand why we can't just instantiate the authenticator once.

PR #1814, opened 2022-07-01
Add clustersyncfailing metric to crd

suhanime:
While adding ClusterSyncFailingSeconds metric, adding it to the enum in
durationMetrics was missed, which didn't catch the appropriate make
targets for adding it to the crd. This commit fixes that.

/assign @2uasimojo


2uasimojo (commented):
Yah. Sorry I missed this during review.

/lgtm

PR #1815, opened 2022-07-01
Fix DNSZoneNotReady[TimedOut] hot loop

2uasimojo:
b7acaf2 / #1769 left a code path where the CD status condition
DNSZoneNotReady can thrash between two Reasons: DNSZoneNotReady and
DNSZoneNotReadyTimedOut. This pretty much defeats the intended purpose
of that effort, which was to make the latter a terminal state, whereupon
the controller should calm down and stop doing things.

This commit refactors the code to reduce the complexity, and resolves
the flawed logic.

HIVE-1881


2uasimojo:
This wfm.
/assign @abutcher

abutcher (approved):
Looks solid :+1: 
/lgtm

PR #1816, opened 2022-07-01
Don't rely on WaitingForNodes

suhanime:
While checking for cluster operators, we always assume we would have
encountered WaitingForNodes, however that is not necessary as we could
have requeued for some other reason after machines were running, and by
the time we check, the nodes could have already been ready.
Allow for the clusteroperator check to proceed for other states as well, 
like StartingMachines or WaitingForMachines.

xref: https://issues.redhat.com/browse/HIVE-1948

/assign @2uasimojo

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -591,20 +591,21 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
 		// Delicate state transitions ahead. If we've now cleared the nodes phase, transition to a pause state
 		// so we can give ClusterOperators time to get their pods running, before we check their status. This is
 		// to avoid prematurely checking and getting good status, but from before hibernation.
-		if readyCondition.Reason == hivev1.ReadyReasonWaitingForNodes {
+		switch readyCondition.Reason {
+		case hivev1.ReadyReasonStartingMachines, hivev1.ReadyReasonWaitingForMachines, hivev1.ReadyReasonWaitingForNodes:
+			// Assume that if we encounter any of the previous possible states, we are ready to proceed.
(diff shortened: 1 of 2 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -591,20 +591,21 @@ func (r *hibernationReconciler) checkClusterRunning(cd *hivev1.ClusterDeployment
 		// Delicate state transitions ahead. If we've now cleared the nodes phase, transition to a pause state
 		// so we can give ClusterOperators time to get their pods running, before we check their status. This is
 		// to avoid prematurely checking and getting good status, but from before hibernation.
-		if readyCondition.Reason == hivev1.ReadyReasonWaitingForNodes {
+		switch readyCondition.Reason {
+		case hivev1.ReadyReasonStartingMachines, hivev1.ReadyReasonWaitingForMachines, hivev1.ReadyReasonWaitingForNodes:
```
suhanime, line 595: @2uasimojo I made the changes as discussed, but one of the unit tests is not passing, and it is definitely something I wouldn't want ti change without thinking it through.
Basically, if hibernating condition reflected that syncsets were just applied, current code changes forcefully pauses for CO check instead of proceeding through

2uasimojo: Discussed in Meet™
- The SyncSetsApplied thing is a red herring, ignore
- This test *should* have the result you're observing: since we're setting up cluster operators to be ready,
  - starting from any of the three conditions in this `case` statement should cause us to go to Pausing with a 2m requeue; and
  - starting from Pausing should either cause us to go to Ready if the time stamps say >2m have elapsed; or to stay Pausing with a RequeueAfter of the remaining time.

suhanime:
@2uasimojo Ready for review
image

2uasimojo (commented):
/lgtm

We'll land it on Monday.

PR #1817, opened 2022-07-08
Azure private endpoint support

m1kola:
https://issues.redhat.com/browse/HIVE-1955

### DIFF
--- a/pkg/remoteclient/remoteclient.go
+++ b/pkg/remoteclient/remoteclient.go
@@ -4,6 +4,8 @@ package remoteclient
 
 import (
 	"context"
+	"fmt"
+	"net"
 	"time"
 
 	"github.com/pkg/errors"
@@ -49,6 +51,10 @@ type Builder interface {
 	// UseSecondaryAPIURL will use the secondary API URL. If there is an API URL override, then the initial API URL
 	// is the secondary.
 	UseSecondaryAPIURL() Builder
+}
+
+type dialer interface {
+	DialContext(ctx context.Context, network, address string) (net.Conn, error)
 }
 
 // NewBuilder creates a new Builder for creating a client to connect to the remote cluster associated with the specified
@@ -289,6 +295,14 @@ func (b *builder) RESTConfig() (*rest.Config, error) {
 		}
 	}
 
+	if override := b.cd.Spec.ControlPlaneConfig.APIServerIPOverride; override != "" {
+		dialer := &net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}
+		cfg.Dial = createDialContext(dialer, override)
+	}
+
 	return cfg, nil
 }
 
@@ -315,4 +329,19 @@ func restConfigFromSecret(kubeconfigSecret *corev1.Secret) (*rest.Config, error)
 	}
 	kubeConfig := clientcmd.NewDefaultClientConfig(*config, &clientcmd.ConfigOverrides{})
 	return kubeConfig.ClientConfig()
+}
+
+func createDialContext(d dialer, apiServerIPOverride string) func(ctx context.Context, network, address string) (net.Conn, error) {
+	return func(ctx context.Context, network, address string) (net.Conn, error) {
+		if network != "tcp" {
+			return nil, fmt.Errorf("unimplemented network %q", network)
+		}
+
+		_, port, err := net.SplitHostPort(address)
+		if err != nil {
+			return nil, err
+		}
+
+		return d.DialContext(ctx, network, apiServerIPOverride+":"+port)
+	}
 }
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -668,6 +668,12 @@ type ControlPlaneConfigSpec struct {
 	// active, Hive will use the override URL for further communications with the API server of the remote cluster.
 	// +optional
 	APIURLOverride string `json:"apiURLOverride,omitempty"`
+
+	// APIServerIPOverride is the optional override of the API server IP address.
+	// Hive will use this IP address for creating TCP connections.
+	// Port from the original API server URL will be used.
+	// +optional
+	APIServerIPOverride string `json:"apiServerIPOverride,omitempty"`
 }
 
 // ControlPlaneServingCertificateSpec specifies serving certificate settings for
--- a/config/crds/hive.openshift.io_clusterdeployments.yaml
+++ b/config/crds/hive.openshift.io_clusterdeployments.yaml
@@ -203,6 +203,12 @@ spec:
                 description: ControlPlaneConfig contains additional configuration
                   for the target cluster's control plane
                 properties:
+                  apiServerIPOverride:
+                    description: APIServerIPOverride is the optional override of the
+                      API server IP address. Hive will use this IP address for creating
+                      TCP connections. Port from the original API server URL will
+                      be used.
+                    type: string
                   apiURLOverride:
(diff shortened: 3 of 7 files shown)


2uasimojo:
Cool. So this new field is per spoke cluster, not per hub?

m1kola:
@2uasimojo yes, that's the idea. Each cluster has their own private endpoint so it has to be on per cluster basis.

m1kola:
I tested this in context of ARO and it seems to work as expected: without `apiServerEndpointIP` Hive fails to reach to the cluster. With `apiServerEndpointIP` populated Hive can talk to the cluster via private endpoint.

2uasimojo (commented):
`make update` should rebuild the saas template that's causing `verify` to fail.

But first...

I've been staring at the REST/HTTP code/docs for a while now, and I can't get my head around how Dial plays in relation to Host. My goal was really just to vet that APIServerEndpointIP is a sensible name for the new field. Because the way it's used makes it feel more like a "proxy". Or does Host not get used at all when Dial is set? Would the original code still work if instead of using APIURLOverride to set Host, we used it to do this DialContext thing? And if so, would that mean that we wouldn't need the new field, and you could just use APIURLOverride for this purpose? If not, what happens if you set both fields?

All of this ignorance makes me leery of introducing this field to the permanent API we have to support until we retire v1.

We have a couple of ways we could make this field "softer" to mitigate those concerns.
- We could wrap it in an alpha feature gate, so in order to use it, you have to explicitly opt into it in HiveConfig, and you're agreeing it might go away at any point without a version bump. (Of course we would consult with you before considering such a change.)
- We could feed it in via an annotation instead of making it a real API field. That way it's not part of the documented API, and we've got the flexibility to change/remove it if we come up with something better later on.

Thoughts?

m1kola:
@2uasimojo `verify` should be fine now. Pleaes take another look.

> I've been staring at the REST/HTTP code/docs for a while now, and I can't get my head around how Dial plays in relation to Host.

Let's imagine that we have an API server in kubeconfig set to `example.com:6443` and DNS resolves to `93.184.216.34`.

Without a custom dialer we will be making a TCP connection to `93.184.216.34` at port `6443` and sending an HTTP request with `Host: example.com` header.

With `apiServerEndpointIP` set to say `10.0.0.10` we will be using a custom dialer and making a TCP connection to `10.0.0.10` at port `6443` (original port from the kubeconfig). We still will be sending a HTTP request with `Host: example.com` header.

> Or does Host not get used at all when Dial is set?

Host will not be used in transport/dial (TCP, layer 4), but will still be used on the application layer (in the `Host` header of HTTP request, layer 7).

> Would the original code still work if instead of using APIURLOverride to set Host, we used it to do this DialContext thing? And if so, would that mean that we wouldn't need the new field, and you could just use APIURLOverride for this purpose?

We considered replacing API server hostname in kubeconfig with IP address such as `example.com:6443` -> `10.0.0.10:6443`, but it won't work because in this case we will be sending HTTP requests with `Host: 10.0.0.10` header.

I believe it will be exact the same issue if we set `APIURLOverride` to `10.0.0.10`.

> If not, what happens if you set both fields?

Given that we have the following:

* Kubeconfig has `example.com:6443` (DNS resolves to `93.184.216.34`)
* `APIURLOverride` is set to `bar.com:6443` (DNS resolves to `104.21.1.6`)

I think this is what is going to happen:
* Without `apiServerEndpointIP` field we will be making TCP connections to `104.21.1.6` at port `6443` and sending HTTP `Host: bar.com` in headers.
* With `apiServerEndpointIP` set to `10.0.0.10` we will be making TCP connections to `10.0.0.10` at port `6443` and sending HTTP `Host: bar.com` in headers.




> All of this ignorance makes me leery of introducing this field to the permanent API we have to support until we retire v1.
> 
> We have a couple of ways we could make this field "softer" to mitigate those concerns.
> 
>     * We could wrap it in an alpha feature gate, so in order to use it, you have to explicitly opt into it in HiveConfig, and you're agreeing it might go away at any point without a version bump. (Of course we would consult with you before considering such a change.)
> 
>     * We could feed it in via an annotation instead of making it a real API field. That way it's not part of the documented API, and we've got the flexibility to change/remove it if we come up with something better later on.
> 
> 
> Thoughts?

I'll leave the decision on how to support this up to you. For ARO it is important to get guarantees that support won't get dropped and that this thing is tested (minimising chances of a regression).

2uasimojo:
> Without a custom dialer we will be making a TCP connection to `93.184.216.34` at port `6443` and sending an HTTP request with `Host: example.com` header.
> 
> With `apiServerEndpointIP` set to say `10.0.0.10` we will be using a custom dialer and making a TCP connection to `10.0.0.10` at port `6443` (original port from the kubeconfig). We still will be sending a HTTP request with `Host: example.com` header.

And then what happens? Does the other end of the connection use `example.com` from that header? Does the DNS-resolved address `93.184.216.34` ever get used again?

IIUC the point is to allow the console and cluster workloads to be public, but hide the API behind a private endpoint, right? Hypothetically, would the effect here be the same as if DNS resolved console.example.com to `93.184.216.XX` and api.example.com to `10.0.0.10`?

---

Regarding unit testing, I thought I had left this comment last week, sorry:
We should be able to do similar to [this](https://github.com/openshift/hive/blob/b87d3cb0a9879d567d573c8d47d550e3ba26a026/pkg/remoteclient/remoteclient_test.go#L54) just to prove that when the override is set in the ClusterDeployment, it makes its way into the REST Config. Though without some fancy mocking, I guess the only thing we would be able to check is that Config.Dial isn't nil...

m1kola:
> And then what happens? Does the other end of the connection use `example.com` from that header? Does the DNS-resolved address `93.184.216.34` ever get used again?

There could potentially be multiple hosts at the same IP address. It is like setting up `example.com` and `bar.com` to be served by the same nginx at `10.0.0.10`. If we do `curl https://10.0.0.10/` we will end up with a request which as `Host: 10.0.0.10` and nginx will have no idea wether it is a request for `example.com` or for `bar.com`.


> IIUC the point is to allow the console and cluster workloads to be public, but hide the API behind a private endpoint, right? Hypothetically, would the effect here be the same as if DNS resolved console.example.com to `93.184.216.XX` and api.example.com to `10.0.0.10`?


We also support fully private clusters where both api server and ingress are private. To access the cluster you have to be on the same network as the cluster.

From ARO resource provider perspective we treat all the cluster as if they were private: we connect o all of them (even public ones) via private endpoint.

m1kola (commented):
> We (okay, I) would like to understand why this wouldn't be solved simply by repointing DNS for the API server URL to the desired IP

We were considering solving it with the DNS however the issue is that we have customers who:

1. Use our managed DNS `aroapp.io`
2. Who use their own custom domain.

To handle clusters with custom domains we would have to create private DNS zone on the ARO RP side and there is a [max limit of 1k private DNS zones per subscription](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#azure-dns-limits). And we group multiple regions supported by ARO RP within a single subscription. So in every ARO RP sub, we would have an upper limit of 1k clusters with custom DNS. Sooner or later we will hit this limit.

pkg/remoteclient/remoteclient.go
```diff
@@ -289,6 +295,14 @@ func (b *builder) RESTConfig() (*rest.Config, error) {
 		}
 	}
 
+	if override := b.cd.Spec.ControlPlaneConfig.APIServerIPOverride; override != "" {
+		dialer := &net.Dialer{
```
2uasimojo, line 299: nts: net.Dialer implements DialContext -- that's the prod path. We define the interface so we can swap out the impl for UT.

PR #1820, opened 2022-07-14
Generics: The Future

dlom:
xref: https://issues.redhat.com/browse/HIVE-1935

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -990,7 +990,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 					cd := testClusterDeploymentWithInitializedConditions(testClusterDeployment())
 					cd.Spec.ManageDNS = true
 					// Pretend DNSNotReady happened 4m ago
-					cond := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, hivev1.DNSNotReadyCondition)
+					cond := controllerutils.FindCondition(cd.Status.Conditions, hivev1.DNSNotReadyCondition)
 					cond.Status = corev1.ConditionTrue
 					cond.Reason = dnsNotReadyReason
 					cond.Message = "DNS Zone not yet available"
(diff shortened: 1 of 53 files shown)


apis/hive/v1/generics.go
```diff
@@ -0,0 +1,9 @@
+package v1
```
2uasimojo, line 1: nit: Something about this file name is making me leery, but I can't put my finger on it (naming a file after a language feature?) and don't have a better suggestion.

apis/hive/v1/generics.go
```diff
@@ -0,0 +1,9 @@
...
+
+type Condition interface {
+	ConditionType() ConditionType
+}
+
+type ConditionType interface {
+	AsString() string
```
2uasimojo, line 8: Out of curiosity, what happens if you do one of these (mutually exclusive) alternatives?
- Use the existng [fmt.Stringer interface](https://pkg.go.dev/fmt#Stringer)
- Nix the interface and just use `string()` inline in `FindCondition`

?

2uasimojo:
> doesn't compile yet because you haven't replaced

We crossed in the mail. Looking good. Take the Draft status off once you think it'll build, so CI can take a crack at it.

apis/hive/v1/generics.go
dlom: I forgot about the `fmt.Stringer` interface... I think I'll do something like this 

```go
type ConditionType interface {
	fmt.Stringer
}
```

so that `ConditionType` can still be used as the generic constraint on `FindCondition`.  Alternatively, `FindCondition` could take `Stringer`s instead of `ConditionType`s

apis/hive/v1/generics.go
dlom: Maybe `constraints.go`?

2uasimojo:
> @dlom: PR needs rebase.

This is gonna be a nightmare to keep rebased. Basically every PR we land is gonna cause a merge conflict :(

apis/hive/v1/generics.go
2uasimojo: I think I prefer the former. The latter would mean you would have to change every condition type to a Stringer, which isn't _awful_, but if it's gonna be an interface type anyway, I like the explicitness of `ConditionType`.

That said, what about my second bullet? Then all the condition types can just be `string` and we don't need an interface for it.

apis/hive/v1/generics.go
2uasimojo: I was thinking `conditions.go`, except that I imagine at some point we'll have generic interfacey things for other pieces. Then again, they could be in separate files when we do.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1805,7 +1805,9 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 			validate: func(c client.Client, t *testing.T) {
 				cd := getCD(c)
-				lc := controllerutils.FindClusterDeploymentCondition(cd.Status.Conditions, "ClusterImageSetNotFound")
+				// This is a hack to get Go to accept this arbitrary string as the proper type
+				dummy := hivev1.ClusterDeploymentCondition{Type: "ClusterImageSetNotFound"}
+				lc := controllerutils.FindCondition(cd.Status.Conditions, dummy.Type)
```
2uasimojo, line 1810: Rather make a const for it, no?

2uasimojo:
/lgtm

hello?

PR #1822, opened 2022-07-19
Azure private endpoint support - follow-up

m1kola:
Addresses remaining points from this comment: https://github.com/openshift/hive/pull/1817#pullrequestreview-1043467943

### DIFF
--- a/pkg/remoteclient/remoteclient.go
+++ b/pkg/remoteclient/remoteclient.go
@@ -4,7 +4,6 @@ package remoteclient
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkg/remoteclient/remoteclient.go
```diff
@@ -1,6 +1,6 @@
 package remoteclient
 
-//go:generate mockgen -source=./remoteclient.go -destination=./mock/remoteclient_generated.go -package=mock
+//go:generate mockgen -destination=./mock/remoteclient_generated.go -package=mock  github.com/openshift/hive/pkg/$GOPACKAGE Builder
```
m1kola, line 3: Had to switch from source mode to reflect mode to be able to select a specific interface.

m1kola: Switched back to the source mode and moved dialer related code into separate files as per Eric's suggestion.

2uasimojo:
/lgtm

Thanks @m1kola

PR #1823, opened 2022-07-19
Add privatelink machinepool w/ public subnets test.

abutcher:
When privatelink is enabled for a clusterDeployment, detect this and
allow public subnets to be used for machinepools.


abutcher:
/assign @2uasimojo 

I have commented out the test but left the check in the verification piece that would cause the new test to fail.

abutcher:
I reorganized this into several commits,
- Fix for a latent nil pointer dereference that happens when `expectedSubnetIDInMachineSet` is set for a test case but a subnet ID is not found on the resultant machineset.
- New failing test case.
- Make failing test case pass by commenting out `expectedSubnetIDInMachineSet` in the test along with a comment regarding `TODO` when HIVE-1805 is addressed.

2uasimojo:
lgtm so far. Leaving open for another commit with the fix.

abutcher:
Will be superseded by https://github.com/openshift/hive/pull/1984

PR #1826, opened 2022-07-20
Fix clearing of ClusterSync Failing metric 

suhanime:
ClusterSync Failing metric must be cleared when clustersync is no longer
failing, however the code path that evoked clearing of that metric in
clustersync controller when the failing condition was changed to success
was somehow never evoked.

- This commit moves the clearing logic under this criterion to the
  metrics controller. The controller assumes that if any clustersync is
not failing than the corresponding metric should be cleared.
- Since the code path in clustersync controller wasn't being evoked,
  this commit also reverts to the previous code, and moves the clearing
logic to the metrics controller, thereby ensuring we are only setting or
clearing the metric in either the metrics controller or the
clusterdeployment controller

/assign @2uasimojo

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -484,7 +484,7 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 	)
 	clusterSync.Status.SelectorSyncSets = syncStatusesForSelectorSyncSets
 
-	failing := setFailedCondition(clusterSync)
+	setFailedCondition(clusterSync)
 
 	// Set clusterSync.Status.FirstSyncSetsSuccessTime
 	syncStatuses := append(syncStatusesForSyncSets, syncStatusesForSelectorSyncSets...)
@@ -499,10 +499,6 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 			logger.WithError(err).Log(controllerutils.LogLevel(err), "could not update ClusterSync")
 			return reconcile.Result{}, err
 		}
-	}
-	// Clear the cluster sync failing metric
-	if !failing {
-		ClearClusterSyncFailingSecondsMetric(clusterSync.Namespace, clusterSync.Name, logger)
 	}
 
 	if needToDoFullReapply {
@@ -1019,8 +1015,7 @@ func doesSelectorSyncSetApplyToClusterDeployment(selectorSyncSet *hivev1.Selecto
 	return labelSelector.Matches(labels.Set(cd.Labels))
 }
 
-// setFailedCondition returns true if failed condition is set to true
-func setFailedCondition(clusterSync *hiveintv1alpha1.ClusterSync) (failing bool) {
+func setFailedCondition(clusterSync *hiveintv1alpha1.ClusterSync) {
 	status := corev1.ConditionFalse
 	reason := "Success"
 	message := "All SyncSets and SelectorSyncSets have been applied to the cluster"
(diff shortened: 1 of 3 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -502,6 +502,13 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 	}
 	// Clear the cluster sync failing metric
 	if !failing {
+		// Set the metric to zero first since clearing the metric would only stop recording it.
```
suhanime, line 505: I'd like to stress that this is not what we usually do, metrics should always be considered in conjunction with their timestamps, but I also do not see any problem with this - and we might want to consider doing so for other gauges - particularly failure reporting gauges across the codebase

2uasimojo: As we discussed in scrum, I think "usually" we're tallying a discrete event in the past (e.g. "a provision failed") whereas here we're tracking in realtime how long the clustersync *has been* in the failing state.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -502,6 +502,13 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
...
+		cond := controllerutils.FindClusterSyncCondition(origStatus.Conditions, hiveintv1alpha1.ClusterSyncFailed)
+		if cond != nil && hivemetrics.ShouldLogGaugeDurationMetric(hivemetrics.MetricClusterSyncFailingSeconds,
+			time.Since(cond.LastTransitionTime.Time).Seconds()) {
+			hivemetrics.MetricClusterSyncFailingSeconds.WithLabelValues(
+				hivemetrics.GetNameSpacedName(clusterSync.Namespace, clusterSync.Name)).Set(0)
```
2uasimojo, line 510: I think we also need to zero the metric in [this code path](https://github.com/openshift/hive/blob/e7fbba49305cab53e8e70ec433ebb3e9ac6e7cae/pkg/controller/metrics/metrics.go#L484-L489) by adding an `else` branch.

Having looked at that, I'm actually thinking we may want to get rid of handling the metric in this controller at all -- remove this whole block. Thoughts?

suhanime: `ShouldLogXDurationMetric` requires seconds as a parameter - seconds since the `lastTransitionTime`. While the clusterSync is still failing, we can make that decision from the actual status, but when we need to clear the metric and set it to zero - that's when we need the lastTransitionTime of the condition before we change it - that is only available in the clusterSync controller. Including `else` part where the metrics controller computes the metric and log it to zero regardless of whether we logged the metric for the CS or not - needlessly complicates the logic and ends up with more entries than necessary
Clearing the metric is a different case because it clears it whether or not the metric was ever logged
Besides, the test for WaitingForCO metric [shows](https://user-images.githubusercontent.com/54181650/180893909-66c85790-e42d-4b4e-927e-7eaba8dfce68.png) how the metrics controller eventually does stop logging the metric, so I do not see this as a deal breaker

2uasimojo: Oh. This is a good point. If sync succeeds, setFailedCondition (L487) will set the ClusterSyncFailed condition to False and update the LastTransitionTime to Now(). So ShouldLogGaugeDurationMetric won't trigger (unless configured to zero, which shouldn't be the case), and this branch won't be reached.

Point is, we're going to have to solve this problem regardless.

Since we're deleting it immediately, wouldn't it be okay to set it to zero even if it doesn't exist beforehand? It would take rather extreme timing for it to get scraped in that interval -- and I'm willing to accept that it would be zero in that case.

But wait again. Isn't setting to zero a no-op anyway if we're then going to delete right away? The scraper won't see that zero; it'll just see that the metric isn't there, and use the previously-reported value -- which is the whole problem we're trying to solve with this PR.

Let's sleep on this and discuss tomorrow.

suhanime:
@2uasimojo Fixed as per discussion.
Note: There was a bug in the code path for clusterSync controller, hence I moved the clearing logic to metrics controller. The execution now is much more clean and works as expected

2uasimojo (commented):
Nice. Thanks for sticking with this through all the confusion.

/lgtm

PR #1829, opened 2022-07-25
Minor cleanup (in a test) related to generics 

dlom:
xref: https://github.com/openshift/hive/pull/1820#discussion_r926123103
/assign @2uasimojo

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -1792,7 +1792,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 					cd.Spec.Provisioning.ImageSetRef = &hivev1.ClusterImageSetReference{Name: testClusterImageSetName}
 					cd.Status.Conditions = append(cd.Status.Conditions,
 						hivev1.ClusterDeploymentCondition{
-							Type:    "ClusterImageSetNotFound",
+							Type:    hivev1.ClusterImageSetNotFoundCondition,
 							Status:  corev1.ConditionFalse,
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -1805,9 +1805,9 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 			validate: func(c client.Client, t *testing.T) {
 				cd := getCD(c)
-				// This is a hack to get Go to accept this arbitrary string as the proper type
-				dummy := hivev1.ClusterDeploymentCondition{Type: "ClusterImageSetNotFound"}
-				lc := controllerutils.FindCondition(cd.Status.Conditions, dummy.Type)
+				// This is a minor hack to get Go to accept this arbitrary string as the proper type
+				const clusterImageSetNotFound hivev1.ClusterInstallConditionType = "ClusterImageSetNotFound"
```
2uasimojo, line 1809: I meant in apis/hive/v1/clusterinstall_conditions.go with the other `ClusterInstallConditionType` consts. Then there's no hackery -- this is just like any other `FindCondition` call.

2uasimojo: ...and replace the hardcoded string on L325.

Perhaps this isn't in with the other consts because it's "legacy" and therefore we don't want consumers trying to look for it? I don't see why we can't have the const though -- just make sure the docstring comment mentions it's legacy and not to use it in consuming code.

2uasimojo (commented):
Yah, just like that.

/lgtm

PR #1830, opened 2022-07-25
Report Waiting For CO metric correctly

suhanime:
In one of the previous passes when the change for reporting the
transitioning metrics was put up, the value recorded for WaitingForCO
metric was incorrectly set to report at time since the last transition
time for Ready condition instead of Hibernating condition. This commit
corrects the mistake

/assign @2uasimojo


suhanime:
I've no idea how I missed this.

Some notes about testing - 
- Can test for WaitingForCO state by injecting an incorrect CR into the spoke cluster. QEs use a CredsRequest with a secretRef to one whose namespace does not exist. CCO will be in Progressing state after because one the secrets isn't syncing. Can revert by simply deleting the CredsRequest
- Being stuck for a long time at WaitingForCO can also naturally happen when the nodes need to be upgraded

image

PS: @2uasimojo This can also help visualize what happens to a metric once we clear it - metrics controller will eventually just stop reporting it

2uasimojo (commented):
This will be covered by forthcoming UT additions you're actively working on elsewhere, yes?

/lgtm

PR #1833, opened 2022-07-26
Enable BalanceSimilarNodeGroups for the default cluster autoscaler

abutcher:
This change enables `BalanceSimilarNodeGroups` for the default cluster autoscaler. `BalanceSimilarNodeGroups` will be enabled for new cluster autoscalers created by Hive and existing cluster autoscalers will be updated to enable `BalanceSimilarNodeGroups` unless `BalanceSimilarNodeGroups` was previously disabled.

With `BalanceSimilarNodeGroups` enabled, the cluster autoscaler automatically identifies node groups with the same instance type and the same set of labels (except for automatically added zone label) and tries to keep the sizes of those node groups balanced. Ref: [Running ClusterAutoscaler with nodes in multiple zones for HA purposes](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler)

Pending discussion.
/hold

[HIVE-1976](https://issues.redhat.com//browse/HIVE-1976)

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -813,12 +814,23 @@ func (r *ReconcileMachinePool) syncClusterAutoscaler(
 		}
 	}
 	if defaultClusterAutoscaler != nil {
-		if spec := &defaultClusterAutoscaler.Spec; spec.ScaleDown == nil || !spec.ScaleDown.Enabled {
-			logger.Info("updaing cluster autoscaler")
-			if spec.ScaleDown == nil {
-				spec.ScaleDown = &autoscalingv1.ScaleDownConfig{}
-			}
+		spec := &defaultClusterAutoscaler.Spec
+		changed := false
+		if spec.ScaleDown == nil {
+			spec.ScaleDown = &autoscalingv1.ScaleDownConfig{}
+		}
+		// Ensure spec.ScaleDown.Enabled = true
+		if !spec.ScaleDown.Enabled {
 			spec.ScaleDown.Enabled = true
+			changed = true
+		}
+		// Default spec.BalanceSimilarNodeGroups to true if unset.
+		if spec.BalanceSimilarNodeGroups == nil {
+			spec.BalanceSimilarNodeGroups = pointer.BoolPtr(true)
+			changed = true
+		}
+		if changed {
+			logger.Info("updaing cluster autoscaler")
 			if err := remoteClusterAPIClient.Update(context.Background(), defaultClusterAutoscaler); err != nil {
 				logger.WithError(err).Error("could not update cluster autoscaler")
 				return err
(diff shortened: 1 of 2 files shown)


machi1990 (approved):
lgtm

Thanks @abutcher

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -813,12 +814,23 @@ func (r *ReconcileMachinePool) syncClusterAutoscaler(
...
+		if spec.ScaleDown == nil {
+			spec.ScaleDown = &autoscalingv1.ScaleDownConfig{}
+		}
+		// Ensure spec.ScaleDown.Enabled = true
+		if !spec.ScaleDown.Enabled {
 			spec.ScaleDown.Enabled = true
+			changed = true
+		}
+		// Ensure spec.BalanceSimilarNodeGroups = true.
+		if spec.BalanceSimilarNodeGroups == nil || !*spec.BalanceSimilarNodeGroups {
```
2uasimojo, line 828: I don't think we should override this if it's already set.
```suggestion
		// Default spec.BalanceSimilarNodeGroups to true if unset.
		if spec.BalanceSimilarNodeGroups == nil {
```

pkg/controller/machinepool/machinepool_controller_test.go
```diff
@@ -497,6 +497,80 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+			},
+			expectedRemoteMachineAutoscalers: []autoscalingv1beta1.MachineAutoscaler{
+				*testMachineAutoscaler("foo-12345-worker-us-east-1a", "1", 1, 2),
+				*testMachineAutoscaler("foo-12345-worker-us-east-1b", "1", 1, 2),
+				*testMachineAutoscaler("foo-12345-worker-us-east-1c", "1", 1, 1),
+			},
+			expectedRemoteClusterAutoscalers: []autoscalingv1.ClusterAutoscaler{
+				*testClusterAutoscaler("2"),
+			},
+		},
```
2uasimojo, line 573: Do we need to add logic to the test driver to assert that the field is set? I don't see anywhere we're doing that at the moment.

abutcher: The autoscaler returned from `testClusterAutoscaler()` has `BalanceSimilarNodeGroups = true` so when we test the expected clusterautoscaler below we are ensuring it is set.

2uasimojo: Cool. The additional test case (starting/staying disabled) makes this clearer as well. Thanks.

pkg/controller/machinepool/machinepool_controller_test.go
```diff
@@ -497,6 +497,115 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
+			clusterDeployment: testClusterDeployment(),
+			machinePool:       testAutoscalingMachinePool(3, 5),
+			remoteExisting: []runtime.Object{
+				testMachine("master1", "master"),
+				testMachineSet("foo-12345-worker-us-east-1a", "worker", true, 1, 0),
+				testMachineSet("foo-12345-worker-us-east-1b", "worker", true, 1, 0),
+				testMachineSet("foo-12345-worker-us-east-1c", "worker", true, 1, 0),
+				func() runtime.Object {
+					a := testClusterAutoscaler("1")
+					a.Spec.BalanceSimilarNodeGroups = pointer.BoolPtr(false)
```
2uasimojo, line 548: ✓

2uasimojo (commented):
/lgtm

2uasimojo:
/test e2e

node timeout

PR #1838, opened 2022-07-29
Make ProvisionStopped terminal (f'real this time)

2uasimojo:
ProvisionStopped is supposed to be terminal. That is, once you're in
that state, you should never leave it.

In certain circumstances, like reconciling DNSZones under managed DNS,
it was possible to reach ProvisionStopped without ever having created a
ClusterProvision. The logic that decides whether to "retry" provisioning
would happily do so in this case. The result would be that the
ProvisionStopped condition would be cleared (possibly to return later).

This commit adds an early check for the ProvisionStopped condition and
refuses to proceed if it's True.

[HIVE-1881](https://issues.redhat.com//browse/HIVE-1881)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -2130,6 +2130,32 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 			},
 		},
 		{
+			name: "ProvisionStopped already",
+			existing: []runtime.Object{
+				testInstallConfigSecret(),
+				func() runtime.Object {
+					cd := testClusterDeploymentWithInitializedConditions(testClusterDeployment())
+					cd.Status.Conditions = append(cd.Status.Conditions, hivev1.ClusterDeploymentCondition{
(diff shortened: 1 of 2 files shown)


2uasimojo:
/hold

This _still_ isn't doing the trick.

2uasimojo:
> This _still_ isn't doing the trick.

Okay, yes it is. The quirk is: if I already had a cluster that was flipflopping _before_ I deployed this change, it may continue to flipflop. But a new CD should behave properly with this change. (IMHO it's not worth chasing down and closing that gap -- reviewers, wdyt?)

/hold cancel

abutcher (commented):
Looks good to me. In the test, we're ensuring that the `ProvisionStoppedCondition` didn't flip flop and is still `ConditionTrue` :+1: 
/lgtm

abutcher:
> Okay, yes it is. The quirk is: if I already had a cluster that was flipflopping before I deployed this change, it may continue to flipflop. But a new CD should behave properly with this change. (IMHO it's not worth chasing down and closing that gap -- reviewers, wdyt?)

Noting that we discussed existing CD's flip flopping during scrum and I agree that fixing for new deployments is acceptable.

PR #1845, opened 2022-08-05
ClusterDeployment.Spec.Provisioning.ManifestsSecretRef

2uasimojo:
Add the ability to add/override installer manifests via a Secret in the
same manner as previously supported via a ConfigMap from
CD.Spec.Provisioning.ManifestsConfigMapRef.
- The two are made mutually exclusive via admission webhook.
- hiveutil is updated to generate the CD with ManifestsSecretRef instead
of ManifestsConfigMapRef when the `--manifests` arg is used, in case
those manifests contain sensitive data.
- Docs are updated to recommend this new field rather than the old,
since they (mostly) discuss using it to funnel secrets down to the
installer.

[HIVE-1794](https://issues.redhat.com//browse/HIVE-1794)


2uasimojo:
Tested end to end by specifying `--manifests` to `hiveutil create-cluster`. Embedded object ended up on the target cluster ✓

/assign @abutcher 
/cc @jharrington22 @m1kola

m1kola:
@hawkowl could you please take a look? Will it work for us?

docs/aws-sts-provisioning.md
```diff
@@ -67,7 +67,7 @@ May also soon be automated by ccoctl.
 Create a ClusterDeployment normally with the following changes:
 
   1. Create a Secret for your private service account signing key created with ccoctl key-pair above: `kubectl create secret generic bound-service-account-signing-key --from-file=bound-service-account-signing-key.key=serviceaccount-signer.private`
-  1. Create a ConfigMap for your installer manifets (credential role Secrets, Authentication config): `kubectl create configmap cluster-manifests --from-file=manifests/`
+  1. Create a Secret for your installer manifets (credential role Secrets, Authentication config): `kubectl create secret generic cluster-manifests --from-file=manifests/`
```
abutcher, line 70: Typo in the original doc.

```suggestion
  1. Create a Secret for your installer manifests (credential role Secrets, Authentication config): `kubectl create secret generic cluster-manifests --from-file=manifests/`
```

abutcher (commented):
Looks good to me. Added comment for one latent typo.

docs/aws-sts-provisioning.md
2uasimojo: Good eye. Fixed (here and one other spot).

jharrington22 (commented):
/lgtm

Thanks @2uasimojo, yes @m1kola this will satisfy our need for the cluster service principal

PR #1847, opened 2022-08-09
Quiet down "no additional log fields" message

2uasimojo:
We were overenthusiastic with this message when adding support for
additional log fields: we would emit it as a Warning even when the
annotation was absent from the CR being reconciled. Example:

```
time="2022-08-09T10:20:09.286Z" level=warning msg="failed to extract additional log fields -- ignoring" clusterPool=default/mihuang-pool-1 controller=clusterpool error="no additional log fields found" reconcileID=nskm25td
```

This commit takes that down to Debug level.

[HIVE-1966](https://issues.redhat.com//browse/HIVE-1966)

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -118,6 +118,13 @@ function save_hive_logs() {
       fi
     done
   done
+  # Let's try to save any prov/deprov pod logs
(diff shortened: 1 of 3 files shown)


2uasimojo:
I have the fix for this, but I had to repro locally to find it. Adding some extra log capture to the e2e tests for FFDC, and testing with the bug still in place. Once I've verified the logs get saved as desired, I'll upload the fix.

2uasimojo:
Should be ready to go now.

PR #1854, opened 2022-08-15
ScaleMode

2uasimojo:
Scale mode allows hive consumers to deploy multiple controller stacks and spread the object database over multiple kube API servers. See the documentation in the PR for more details.

[HIVE-1862](https://issues.redhat.com//browse/HIVE-1862)


pkg/operator/hive/hive_controller.go
```diff
@@ -230,43 +234,57 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 			}
 
-			shouldEnqueue := func(targetNS string) bool {
-				// Always enqueue all HiveConfigs if the operator ns was updated
-				if nsName == hiveOperatorNS {
-					return true
+			shouldEnqueue := func() bool {
+				if config.Spec.ScaleMode {
+					// TODO: Bleh, this is no good; if the update was to remove the label, we *do* want to reconcile.
+					// We'll need to do this with predicates.
```
2uasimojo, line 262: YOU ARE HERE

2uasimojo: ✓

2uasimojo:
/test e2e

Let's see if I at least made this backward compatible.

cmd/manager/main.go
```diff
@@ -160,16 +160,27 @@ func newRootCommand() *cobra.Command {
...
+
 				log.Info("Registering Components.")
 
 				if err := utils.SetupAdditionalCA(); err != nil {
 					log.Fatal(err)
 				}
 
 				// Setup Scheme for all resources
+				// NOTE: We're reusing the scheme for both control and data plane, even though we're
+				// referencing a disjoint subset of object types from each. It's not a// massive waste
```
2uasimojo, line 179: s,//,,

pkg/controller/utils/ownership.go
```diff
@@ -116,3 +121,77 @@ func SyncOwnerReference(owner hivev1.MetaRuntimeObject, object hivev1.MetaRuntim
 	objectLogger.Info("Successfully set owner reference using labels")
 	return nil
 }
+
```
2uasimojo, line 124: docstring the new funcs in this file

docs/scaling-hive.md
```diff
@@ -70,3 +99,78 @@ Here is an example graph from a scale test where Hive is running on an m5.xlarge
...
+- Remove `targetNamespace` if present
+- Add `scaleMode: true`
+
+E.g.
+
+```shell
+oc patch hiveconfig hive --type=json --patch '[{"op": "remove", "path": "/spec/targetNamespace"}, {"op": "add", "path": "/spec/scaleMode", "value": true}]'
+```
+
+This will cause hive to _undeploy_ the controller stack from the erstwhile `targetNamespace` and deploy it into appropriately-labeled namespaces as described below.
```
2uasimojo, line 119: "... (defaults to `hive` if unspecified) ..."

pkg/operator/hive/hive_controller.go
```diff
@@ -459,110 +493,235 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
...
+		return reconcile.Result{}, errors.New(msg)
+	}
+	return reconcile.Result{}, nil
+}
+
+// getHiveNamespaces returns a set of target namespaces into which hive controllers should be
+// deployed.
+func (r *ReconcileHiveConfig) getHiveNamespaces(config *hivev1.HiveConfig) (sets.String, error) {
+	if config.Spec.TargetNamespace != "" && config.Spec.ScaleMode {
+		return nil, fmt.Errorf("TargetNamespace and ScaleConfig are mutually exclusive")
```
2uasimojo, line 571:
```suggestion
		return nil, fmt.Errorf("TargetNamespace and ScaleMode are mutually exclusive")
```

2uasimojo:
Now with functioning hiveadmission. Alas, hypershift-specific for the moment.
TODO:
- Update design & docs to indicate hypershift specific-ness
- Investigate a hypershift enhancement to inject the hiveadmission service IP into the existing konnectivity-agent instead of rolling our own.

config/hiveadmission/konnectivity-agent.yaml
```diff
@@ -0,0 +1,107 @@
+---
```
2uasimojo, line 1: Per @csrwng, if we use a URL rather than a service ref in my APIService, we shouldn't need this konnectivity agent. He also asserts we should be able to do this without requiring any additional authz, which should mean the existing cert we're copying in (for authn) should be adequate.
Presumably we would also not need the Endpoints or the mirror Service in the HCP with this setup.

PR #1855, opened 2022-08-17
Clusterpool inventory - fix missing powerstate

abraverm:
Hibernation controller manages the CD `status.powerstate`, but without an
actuator for onprem clouds, it will never change. In case we don't need
Hibernation (runningCount == Size) then the CD is assignable if it is
Installed.


abraverm:
Hi @2uasimojo , I will add tests that will catch this issue in case of future changes, but is this the right direction?

2uasimojo:
Ugh. We really ought to figure out a way to update the PowerState for such clusters. But as you say, without an actuator, I guess we'll need something like this :(

Rather than assuming `size == runningCount` always applies to on-prem clouds, can we instead factor out the other parts of [this check](https://github.com/openshift/hive/blob/5dd1351fac13b9493d94c20146fbb5ae5a2022b3/pkg/controller/clusterpool/clusterpool_controller.go#L304) and use that in the conditional instead?

abraverm:
Like this?
```
			} else if p := pool.Spec.Platform; cd.Spec.Installed {
				alwaysRunning := p.OpenStack != nil || p.Ovirt != nil || p.VSphere != nil
				if cd.Status.PowerState == hivev1.ClusterPowerStateRunning || alwaysRunning {
					cdCol.assignable = append(cdCol.assignable, ref)
				} else {
```

2uasimojo:
Yes; but make `alwaysRunning` either a var you can pass around or a func you can call from both places.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -323,9 +324,78 @@ func TestReconcileClusterPool(t *testing.T) {
 			expectedInventoryValidStatus: corev1.ConditionFalse,
 			expectedInventoryMessage:     map[string][]string{"BrokenByCloud": {"test-cdc-1"}},
 			expectedCDCReason:            map[string]string{"test-cdc-1": hivev1.CustomizationApplyReasonBrokenCloud},
-			expectInventory:              true,
+			expectedPoolVersion:          inventoryPoolVersion,
 			expectedAssignedCDCs:         map[string]string{"test-cdc-1": "c1"},
 		},
+		{
+			name: "cp with inventory and good cdc is valid, cd created, and calimed",
```
2uasimojo, line 331: nit
```suggestion
			name: "cp with inventory and good cdc is valid, cd created, and claimed",
```

pkg/controller/clusterpool/collections.go
```diff
@@ -796,6 +796,13 @@ func (cdcs *cdcCollection) Unassign(c client.Client, cdc *hivev1.ClusterDeployme
 
 	delete(cdcs.reserved, cdc.Name)
 
+	// remove duplicates
```
2uasimojo, line 799: How does duplication happen? Wouldn't it mean the CDC was already unassigned?

It looks like it might come from [here](https://github.com/openshift/hive/pull/1855/files#diff-bbf4d3cb24b8cac87951b84ca973747902c00d4a9d662648c8e760a37569581dR999) -- the condition surrounding this dictates that the CDC is available, so why wouldn't it have been in the unassigned list in the first place? Seems like that's where we should attack this situation, rather than culling duplicates after the fact.

Is this related to the rest of the PR? If not, I would like to see it split out.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -323,9 +324,78 @@ func TestReconcileClusterPool(t *testing.T) {
...
+			existing: []runtime.Object{
+				inventoryPoolBuilder().Build(testcp.WithSize(1)),
+				testcdc.FullBuilder(testNamespace, "test-cdc-1", scheme).Build(),
+				testclaim.FullBuilder(testNamespace, "test", scheme).Build(testclaim.WithPool(testLeasePoolName)),
+				cdBuilder("c1").Build(
+					testcd.Installed(),
+					testcd.WithClusterPoolReference(testNamespace, testLeasePoolName, "test"),
```
2uasimojo, line 338: Was this one supposed to start off unassigned so we could prove that it does *not* get assigned? (Because it is not running and it is not an "always running" cloud.)

Or is that covered elsewhere? (I didn't look.)

2uasimojo (commented):
Good stuff @abraverm

pkg/controller/clusterpool/collections.go
abraverm: The duplication comes from [here](https://github.com/openshift/hive/blob/master/pkg/controller/clusterpool/collections.go#L992) (I think we are pointing at the same code). Its unrelated to this PR, just another issue I have stumbled upon. I will create another PR.

PR #1857, opened 2022-08-19
[hiveutil] Correct references to --manifests parameter.

abutcher:
There were a few places where we mention a `--manifests-dir` parameter which doesn't exist. Replace those references with the `--manifests` parameter.

/assign @2uasimojo


2uasimojo (commented):
yup, thanks

/lgtm

PR #1858, opened 2022-08-21
Fix CDC restoration of finalizers

abraverm:
When finalizer is missing in a CDC, then only add a finalizer if it is in the pool inventory

### DIFF
--- a/pkg/controller/clusterpool/collections.go
+++ b/pkg/controller/clusterpool/collections.go
@@ -881,6 +881,10 @@ func (cdcs *cdcCollection) Unassigned() []*hivev1.ClusterDeploymentCustomization
 	return cdcs.unassigned
 }
 
+func (cdcs *cdcCollection) ByName(name string) *hivev1.ClusterDeploymentCustomization {
+	return cdcs.byCDCName[name]
+}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterpool/collections.go
```diff
@@ -925,8 +925,8 @@ func (cdcs *cdcCollection) SyncClusterDeploymentCustomizationAssignments(c clien
 				}
 			}
 		} else {
-			// Ensure the finalizer is present if the CDC is not deleted, OR if it is reserved
-			if !hasFinalizer {
+			// Ensure the finalizer is present if the CDC is in this pool inventory
+			if _, ok := cdcs.byCDCName[cdc.Name]; !hasFinalizer && ok {
```
2uasimojo, line 929: nit: Since byCDCName contains pointers, you can get away with not using the `exists` form:
```suggestion
			if cdcs.byCDCName[cdc.Name] != nil && !hasFinalizer {
```

That said, the other collections have a `ByName()` method whose contract specifies that it returns `nil` if there's no such key (by simply wrapping the accessor as above). For parity, it would be nice to introduce that and use it (throughout).

2uasimojo (commented):
This looks fine, just looks like it needs to be reconciled with the other PR.

2uasimojo (commented):
/lgtm

PR #1859, opened 2022-08-22
catch install log error for terraform destroy failures

yithian:
the installer will fail if terraform can't destroy bootstrap resources,
resulting in a cluster provisioning failure


yithian:
this has been seen multiple times, most recently https://issues.redhat.com/browse/OHSS-14630

2uasimojo (commented):
/lgtm

PR #1867, opened 2022-08-30
Tech Debt: Remove pre-4.4.8 things

2uasimojo:
4.4 has been [out of support](https://access.redhat.com/support/policy/updates/openshift) for a year and a half [1], and OSD [confirms](https://coreos.slack.com/archives/CCX9DB894/p1661888186326109) a) they don't have any hangers-on; and b) if they did, they would be running at their own risk anyway.


pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -97,56 +97,6 @@ func TestReconcile(t *testing.T) {
...
-			},
-			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
-				cond, runCond := getHibernatingAndRunningConditions(cd)
-				require.NotNil(t, cond)
-				assert.Equal(t, hivev1.HibernatingReasonUnsupported, cond.Reason)
-				require.NotNil(t, runCond)
-				assert.Equal(t, hivev1.ReadyReasonRunning, runCond.Reason)
-				assert.Equal(t, hivev1.ClusterPowerStateRunning, cd.Status.PowerState)
-			},
-		},
```
suhanime, line 138: These tests actually also cover the "Unsupported" reason path - i.e., it gets set when expected without errors, and powerstate gets updated for unsupported clusters. One criterion of determining the hibernation is unsupported is the old clusterversion, while the other one is that the cloud platform (and hive) should support hibernation. I think we should keep these 2 tests, just alter them to point to clusters with unsupported platforms.
PS: A bit of history - iirc, we included one of these tests because of a bug QEs found during the initial kerfuffle of the previous refactor.

PR #1868, opened 2022-08-31
CD.Spec.Provisioning.CLIImageOverride

2uasimojo:
Allow overriding the CLI image (the one with the `oc` binary) in the
same manner as we allow overriding the Installer image.

[HIVE-2014](https://issues.redhat.com//browse/HIVE-2014)


jharrington22 (commented):
Looks good @2uasimojo we'll test in the Azure environment to confirm

PR #1869, opened 2022-09-01
Prioritize stale CD deletion for onprem clouds

abraverm:


abraverm:
Hi @2uasimojo, I would appriciate your thoughts on the subject. The main issue is resource starvation and the time that takes to deploy a cluster. I think a clusterpool option would be a better direction, but is it possible to do this in the background without blocking additional reconcile (the conditions for the original stale case)?

2uasimojo:
For starters, we can't violate `maxConcurrent` -- the total number of provision/deprovision operations allowed at a time. So at the very least that would have to be taken into account (by moving the new `case` below the one that checks `availableCurrent`).

Assuming that, would you still be interested in consuming a `maxConcurrent` slot on deprovisioning a stale cluster rather than on creating a new one, or deleting a Broken one? Why?

I need to understand more about the resource starvation you're observing. Off the cuff, I would posit that it should be resolved by tweaking existing pool params (like `maxConcurrent` itself) rather than changing the logic.

PR #1872, opened 2022-09-02
DNM: Move secret retrieval to hiveutil for AWS prov/deprov

2uasimojo:
Pods created by hive-controllers require information from secrets and
configmaps (for example, cloud credentials) in order to function. Today
these are passed into the pods from the hive-controllers side via
environment variables and volumes. Working on HIVE-1862, it's looking
like we're going to need that same business logic -- which parts of
which objects are needed in what form -- on the hiveutil side when
running in scale mode. In order to avoid duplication of that logic,
we're considering moving it *all* to the hiveutil side, even for
non-scale mode.

This experiment is a proof of this concept for provision and deprovision
paths on AWS. The cloud credentials secret contents are no longer passed
through via the pod spec. Instead, the hiveutil command itself uses a
local client to load that secret and set up the same environment
variables and files that were previously passed through on the pod.

Spike related to HIVE-1862

### DIFF
--- a/contrib/pkg/deprovision/awstagdeprovision.go
+++ b/contrib/pkg/deprovision/awstagdeprovision.go
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/hive/contrib/pkg/utils"
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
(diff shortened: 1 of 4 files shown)


2uasimojo:
/hold
experimental.
/cc @abutcher @jaybeeunix @dustman9000

2uasimojo:
This works. I'm going to do it for real via #1874.

PR #1874, opened 2022-09-06
Move secret/configmap retrieval to hiveutil for prov/deprov

2uasimojo:
Pods created by hive-controllers require information from secrets and configmaps (for example, cloud credentials) in order to function. Previously these were passed into the pods from the hive-controllers side via environment variables and volumes. In scale mode (see [HIVE-1861](https://issues.redhat.com//browse/HIVE-1861)) we're going to need that same business logic -- which parts of which objects are needed in what form -- on the hiveutil side.

In order to avoid duplication of that logic, we're moving it *all* to the hiveutil side, even for non-scale mode.

The cloud credentials secret and other secrets/configmaps are no longer passed through via the pod spec. Instead, the hiveutil command itself uses a local client to load those secrets and set up the same environment variables and files that were previously passed through on the pod.

[HIVE-2021](https://issues.redhat.com//browse/HIVE-2021)

### DIFF
--- a/contrib/pkg/utils/aws/aws.go
+++ b/contrib/pkg/utils/aws/aws.go
@@ -1,9 +1,14 @@
 package aws
 
 import (
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/hive/contrib/pkg/utils"
+	"github.com/openshift/hive/pkg/constants"
 	log "github.com/sirupsen/logrus"
 	ini "gopkg.in/ini.v1"
-	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // GetAWSCreds reads AWS credentials either from either the specified credentials file,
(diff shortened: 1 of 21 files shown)


2uasimojo:
/test e2e-azure

Not sure if that's a legit failure, but it looks like a timeout.

2uasimojo:
/test e2e

Job timed out, but seemed to be going fine up until that point. (Except that we blew ~12m trying to create a namespace -- what's that about? Certainly unrelated to the code change, which was not in play at that point.)

2uasimojo:
/test e2e-azure

Okay, that was my mistake. Ordering problem in the azure deprov command: it was validating credentials before running the thing that set them up. Should be fixed now.

2uasimojo:
/assign @abutcher 

/hold

@huangmingxia has graciously [agreed](https://issues.redhat.com/browse/HIVE-2021?focusedCommentId=20923198&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20923198) to QE this for some of the other clouds before we merge it. But code-wise it should be ready for review. Thanks!

2uasimojo:
/test e2e-azure

timeout

2uasimojo:
/assign @abutcher 

QE tested a large enough subset of this (see card) that I'm ~comfortable to land it.

contrib/pkg/utils/aws/aws.go
```diff
@@ -39,3 +44,27 @@ func GetAWSCreds(credsFile, defaultCredsFile string) (string, string, error) {
...
+
+// ConfigureCreds loads a secret designated by the environment variables CLUSTERDEPLOYMENT_NAMESPACE
+// and CREDS_SECRET_NAME and configures AWS credential environment variables and config files
+// accordingly.
+func ConfigureCreds(c client.Client) {
+	credsSecret := utils.LoadSecretOrDie(c, "CREDS_SECRET_NAME")
+	if credsSecret == nil {
+		return
+	}
+	// Should we bounce if any of the following already exist?
```
abutcher, line 56: I think this could become whack-a-mole as creds evolve per cloud if we did bounce

pkg/install/generate.go
```diff
@@ -167,248 +139,90 @@ func InstallerPodSpec(
...
+		"logs":          "/logs",
+		"installconfig": "/installconfig",
+		"pullsecret":    "/pullsecret",
 	}
 
+	var credentialRef, certificateRef string
+
 	switch {
 	case cd.Spec.Platform.AlibabaCloud != nil:
-		env = append(
```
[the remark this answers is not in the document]
abutcher: :heavy_check_mark:

pkg/install/generate.go
```diff
@@ -167,248 +139,90 @@ func InstallerPodSpec(
...
-				},
-			},
-		})
-		volumeMounts = append(volumeMounts, corev1.VolumeMount{
-			Name:      "vsphere-certificates",
-			MountPath: vsphereCloudsDir,
-		})
-		env = append(env, vSphereCredsEnvVars(cd.Spec.Platform.VSphere.CredentialsSecretRef.Name)...)
+		credentialRef, certificateRef = cd.Spec.Platform.VSphere.CredentialsSecretRef.Name, cd.Spec.Platform.VSphere.CertificatesSecretRef.Name
+		// TODO: I don't think we actually use this dir. Can we get rid of it?
```
abutcher, line 194: IIRC we use environment variables for the credentials and we can probably remove this.

abutcher (commented):
/lgtm

pkg/install/generate.go
2uasimojo: Ack. Prefer to do it in a fup jic.

PR #1878, opened 2022-09-13
Override in-cluster credentials mode

2uasimojo:
Recognize a new annotation on ClusterDeployment,
`hive.openshift.io/override-in-cluster-creds-mode`. Setting its value to a valid CredentialsMode (`Manual`, `Passthrough`, `Mint`) will cause hive to edit the install-config.yaml accordingly *after* it has already generated manifests and ignition configs.

HIVE-1989


2uasimojo:
Superseded by #1883

PR #1880, opened 2022-09-14
Add install success/fail metrics

suhanime:
Add hive_cluster_deployment_install_failure_total and
hive_cluster_deployment_install_success_total metrics, which also report
on the cluster_type, cloud platform, region, cluster version, count of
replicas/workers and the number of install attempt pertaining to the
provision attempt

xref: https://issues.redhat.com/browse/HIVE-2006

### DIFF
--- a/pkg/controller/clusterprovision/clusterprovision_controller.go
+++ b/pkg/controller/clusterprovision/clusterprovision_controller.go
@@ -3,6 +3,8 @@ package clusterprovision
 import (
 	"context"
 	"fmt"
+	"gopkg.in/yaml.v2"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -32,6 +34,7 @@ import (
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/hive/pkg/install"
 	k8slabels "github.com/openshift/hive/pkg/util/labels"
+	installertypes "github.com/openshift/installer/pkg/types"
 )
 
 const (
@@ -449,6 +452,9 @@ func (r *ReconcileClusterProvision) transitionStage(
 	if err := r.setCondition(instance, conditionType, corev1.ConditionTrue, reason, message, controllerutils.UpdateConditionAlways, pLog); err != nil {
 		return reconcile.Result{}, err
 	}
+	if stage == hivev1.ClusterProvisionStageFailed || stage == hivev1.ClusterProvisionStageComplete {
+		r.logProvisionSuccessFailureMetric(stage, instance)
+	}
 	if err := r.setStage(instance, stage, pLog); err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


suhanime:
/assign @2uasimojo 

Here are the snapshots:
For failure metric, this one represents 3 different cds over time. We are never clearing the metric - so this is what it would look like. The average over time would make sense imo, otherwise we would either have to label it with cd or report it in a discrete fashion via custom collector
image
FYI for success metric, it was a new provision - but if the same failing provision succeeded after nth attempt, the retry attempt label would reflect that
image

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -583,3 +589,53 @@ func (r *ReconcileClusterProvision) deleteInstallJob(provision *hivev1.ClusterPr
...
+		r.logger.WithError(err).Error("error getting cluster deployment")
+	} else {
+		if stage == hivev1.ClusterProvisionStageFailed {
+			metricInstallFailureSeconds.WithLabelValues(
+				hivemetrics.GetClusterDeploymentType(cd),
+				cd.Labels[hivev1.HiveClusterPlatformLabel],
+				cd.Labels[hivev1.HiveClusterRegionLabel],
+				*cd.Status.InstallVersion,
+				r.getWorkers(*cd),
+				strconv.Itoa(cd.Status.InstallRestarts),
```
2uasimojo, line 628: Should we get this from `instance.Attempt` instead? I didn't go trace back where `cd.Status.InstallRestarts` gets set -- e.g. whether it's in a different controller that runs asynchronously and therefore may be inaccurate -- but the `Attempt` in the ClusterProvision is highly likely to be accurate since it's close by and we're in the controller that owns it.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -583,3 +589,53 @@ func (r *ReconcileClusterProvision) deleteInstallJob(provision *hivev1.ClusterPr
...
+func (r *ReconcileClusterProvision) getWorkers(cd hivev1.ClusterDeployment) string {
+	icSecret := &corev1.Secret{}
+	err := r.Get(context.Background(),
+		types.NamespacedName{
+			Namespace: cd.Namespace,
+			Name:      cd.Spec.Provisioning.InstallConfigSecretRef.Name,
+		},
+		icSecret)
+	if err != nil {
+		r.logger.WithError(err).Log(controllerutils.LogLevel(err), "Error loading install config secret")
```
2uasimojo, line 603: I think this should be
```suggestion
		r.logger.WithError(err).Error("Error loading install config secret")
```
I think in other places we use `Log(LogLevel(err), ...)` so we don't emit the log for certain kinds of `err` that might be "expected" etc. But in this case we definitely want to log at error level.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -583,3 +589,53 @@ func (r *ReconcileClusterProvision) deleteInstallJob(provision *hivev1.ClusterPr
...
+		return "unknown"
+	}
+	ic := &installertypes.InstallConfig{}
+	if err := yaml.Unmarshal(icSecret.Data["install-config.yaml"], &ic); err != nil {
+		r.logger.WithError(err).Log(controllerutils.LogLevel(err), "could not unmarshal InstallConfig")
```
2uasimojo, line 608: As above
```suggestion
		r.logger.WithError(err).Error("could not unmarshal InstallConfig")
```

2uasimojo (commented):
As [discussed](https://coreos.slack.com/archives/CE3ETN3J8/p1663345349435699?thread_ts=1663269534.483899&cid=CE3ETN3J8), please make this a Histogram, similar to `metricInstallJobDuration`.

suhanime:
@2uasimojo Fixed as per your comments
I also changed "retry attempt" label to "install attempt" because I think it is easier to understand and correlate to the provisions
image

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -583,3 +589,53 @@ func (r *ReconcileClusterProvision) deleteInstallJob(provision *hivev1.ClusterPr
...
+		} else {
+			metricInstallSuccessSeconds.WithLabelValues(
+				hivemetrics.GetClusterDeploymentType(cd),
+				cd.Labels[hivev1.HiveClusterPlatformLabel],
+				cd.Labels[hivev1.HiveClusterRegionLabel],
+				*cd.Status.InstallVersion,
+				r.getWorkers(*cd),
+				strconv.Itoa(instance.Spec.Attempt),
+			).Observe(time.Since(instance.CreationTimestamp.Time).Seconds())
+		}
```
2uasimojo, line 639: nit: Could dedup this as:
```suggestion
		theMetric := metricInstallSuccessSeconds
		if stage == hivev1.ClusterProvisionStageFailed {
			theMetric = metricInstallFailureSeconds
		}
		theMetric.WithLabelValues(
			hivemetrics.GetClusterDeploymentType(cd),
			cd.Labels[hivev1.HiveClusterPlatformLabel],
			cd.Labels[hivev1.HiveClusterRegionLabel],
			*cd.Status.InstallVersion,
			r.getWorkers(*cd),
			strconv.Itoa(instance.Spec.Attempt),
		).Observe(time.Since(instance.CreationTimestamp.Time).Seconds())
```

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -583,3 +589,53 @@ func (r *ReconcileClusterProvision) deleteInstallJob(provision *hivev1.ClusterPr
...
+	}
+	return strconv.FormatInt(*ic.WorkerMachinePool().Replicas, 10)
+}
+
+func (r *ReconcileClusterProvision) logProvisionSuccessFailureMetric(
+	stage hivev1.ClusterProvisionStage, instance *hivev1.ClusterProvision) {
+	cd := &hivev1.ClusterDeployment{}
+	if err := r.Client.Get(context.Background(), types.NamespacedName{Namespace: instance.Namespace,
+		Name: instance.Spec.ClusterDeploymentRef.Name}, cd); err != nil {
+		r.logger.WithError(err).Error("error getting cluster deployment")
```
2uasimojo, line 619: nit: could save an `else`/indent below with an explicit `return` here.

2uasimojo (commented):
/lgtm

Thanks!

2uasimojo:
> "retry attempt" label to "install attempt"

To be super anal, the attempt number being zero-based would make "retry" more accurate for human eyes; but I'm fine with it :)

2uasimojo (commented):
/lgtm

PR #1883, opened 2022-09-17
Add support to override effective in-cluster credentials mode

akhil-rane:
ARO requires a different set of credentials for installer and in-cluster 
operations. This commit adds an annotation that would override 
the existing credentials mode in the install-config. The reason we 
do this hack is because in ARO we do not want to pass installer 
credentials inside the cluster. This is done by setting credentials 
mode to `Manual`. We want to instead pass the credentials 
supplied by manifestSecretRef and use the same by overriding 
credentials mode from `Manual` to `Passthrough`.

[HIVE-1989](https://issues.redhat.com//browse/HIVE-1989)

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -441,6 +441,11 @@ const (
 	// cluster. When this annotation is set to a truthy value, hive will parse the installer image URI and use its domain
 	// (everything up to the first `/`) for the CLI image, discarding whatever domain was gleaned from the release image.
 	CopyCLIImageDomainFromInstallerImage = "hive.openshift.io/cli-domain-from-installer-image"
+
+	// OverrideInClusterCredentialsModeAnnotation can be set to override credentials mode in the install config after
+	// generating installer manifests. This will be effective in-cluster credentials mode. The valid values for
+	// credentials mode are "Manual", "Mint" and "Passthrough"
+	OverrideInClusterCredentialsModeAnnotation = "hive.openshift.io/override-in-cluster-credentials-mode"
 )
 
 // GetMergedPullSecretName returns name for merged pull secret name per cluster deployment
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -90,6 +90,8 @@ const (
 	defaultPullSecretMountPath          = "/pullsecret/" + corev1.DockerConfigJsonKey
 	defaultManifestsMountPath           = "/manifests"
 	defaultHomeDir                      = "/home/hive" // Used if no HOME env var set.
+	installConfigKeyName                = "install-config"
+	clusterConfigYAML                   = "cluster-config.yaml"
 )
 
 var (
@@ -856,6 +858,13 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 			return err
 		}
 		m.log.Infof("copied %s to %s", src, dest)
+	}
+
+	if effectiveMode, ok := cd.Annotations[constants.OverrideInClusterCredentialsModeAnnotation]; ok {
+		if err := m.overrideCredentialsModeOnInstallConfig(effectiveMode); err != nil {
+			m.log.WithError(err).Error("error overriding credentials mode on install config")
+			return err
+		}
 	}
 
 	m.log.Info("running openshift-install create ignition-configs")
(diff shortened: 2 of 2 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -1689,3 +1702,48 @@ func getHomeDir() string {
...
+		return fmt.Errorf("did not find key %s in data of cluster-config configmap", installConfigKeyName)
+	}
+
+	decoder = yml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(data)), 4096)
+	installConfig := &installertypes.InstallConfig{}
+	if err := decoder.Decode(installConfig); err != nil {
+		return errors.Wrap(err, "failed to decode install config")
+	}
+
+	installConfig.CredentialsMode = installertypes.CredentialsMode(credentialsMode)
```
2uasimojo, line 1730: If we move the log line from above into this func, we could save the original value here and print both in the message. 
```go
m.log.WithField("origMode", origMode).WithField("newMode", credentialsMode).Info("override effective in-cluster credentials mode")
```
kind of thing. (Print it at the bottom of the func when we know we've succeeded.)

pkg/installmanager/installmanager.go
```diff
@@ -1689,3 +1702,48 @@ func getHomeDir() string {
...
 }
+
+func overrideCredentialsModeOnInstallConfig(manifestsDir, credentialsMode string) error {
+	clusterConfigPath := filepath.Join(manifestsDir, clusterConfigYAML)
+	clusterConfigBytes, err := ioutil.ReadFile(clusterConfigPath)
+	if err != nil {
+		return errors.Wrap(err, "error reading install config")
+	}
+
+	decoder := yml.NewYAMLOrJSONDecoder(bytes.NewReader(clusterConfigBytes), 4096)
```
2uasimojo, line 1713: Curious why we're using this new lib instead of the `yaml` one already used elsewhere in this file. We know `clusterConfigBytes` is yaml, not json, right?

2uasimojo (commented):
No functional issues, just some tidying.

pkg/constants/constants.go
```diff
@@ -441,6 +441,10 @@ const (
 	// cluster. When this annotation is set to a truthy value, hive will parse the installer image URI and use its domain
 	// (everything up to the first `/`) for the CLI image, discarding whatever domain was gleaned from the release image.
 	CopyCLIImageDomainFromInstallerImage = "hive.openshift.io/cli-domain-from-installer-image"
+
+	// HiveOverrideInClusterCredentialsModeAnnotation can be set to override credentials mode in the install config after
+	// generating installer manifests. This will be effective in-cluster credentials mode.
+	HiveOverrideInClusterCredentialsModeAnnotation = "hive.openshift.io/override-in-cluster-credentials-mode"
```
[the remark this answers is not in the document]
2uasimojo: Oh, one more thing here: Can the comment please describe what the value should look like? (One of `Manual`, `Passthrough`, `Mint` IIUC. The initial capital matters, right?)

pkg/installmanager/installmanager.go
```diff
@@ -858,6 +861,16 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 		m.log.Infof("copied %s to %s", src, dest)
 	}
 
+	if _, ok := cd.Annotations[constants.HiveOverrideInClusterCredentialsModeAnnotation]; ok {
+		effectiveMode := cd.Annotations[constants.HiveOverrideInClusterCredentialsModeAnnotation]
```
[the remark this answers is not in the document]
2uasimojo: oh yeah, we should also make sure `cd.Annotations` isn't `nil`. I think it'll pretty much never be -- both k8s and hive love to pepper a CD with annotations -- but just to be paranoid.

pkg/installmanager/installmanager.go
akhil-rane: My bad I tried to use the same one which we used in CCO. Replaced it with existing `yaml`

akhil-rane:
@2uasimojo this is ready for another review.

pkg/installmanager/installmanager.go
```diff
@@ -858,6 +860,13 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 		m.log.Infof("copied %s to %s", src, dest)
 	}
 
+	if effectiveMode, ok := cd.Annotations[constants.OverrideInClusterCredentialsModeAnnotation]; ok {
```
2uasimojo, line 863: Would still like to see a nil check here.

pkg/installmanager/installmanager.go
2uasimojo: I see the log line moved, but the old value isn't being logged. For that matter, it wouldn't be a terrible idea to have a check here like

```go
if installConfig.CredentialsMode == installertypes.CredentialsMode(credentialsMode) {
    m.log.WithField("credsMode", credentialsMode).Info("Not overriding already-set credentials mode")
    return nil
}
```

2uasimojo (commented):
/lgtm

Couple of minor things remain, but nothing crucial, and time is short; we can address them in a followon.

PR #1884, opened 2022-09-20
Revert "Explicit pod security"

abutcher:
This reverts commit a76ba2ff6cc663eb1672c7f3203439a8a81e2048.

Pods won't run with these changes on clusters with version 4.10 and below. There is no restricted SCC that allows setting `seccompProfile` `RuntimeDefault`. Reverting and investigating options for either conditionally deploying with these settings or creating an SCC that allows pods to run with this `securityContext`.


2uasimojo (commented):
/lgtm

PR #1885, opened 2022-09-21
Explicit pod security

abutcher:
This PR unreverts #1884 and removes `seccompProfile` settings from deployments. Versions prior to 4.11 do not have a restricted SCC that allows setting `seccompProfile`.

[HIVE-2023](https://issues.redhat.com/browse/HIVE-2023)


abutcher:
Rebased and cleaned up the namespace scraping a little bit. If we go this route, which I currently think we might, we may need to ensure that our PSA/UID settings jive with K8s or possibly skip configuring pod security if we're not on OpenShift.

2uasimojo:
@dlom was going through some oldymoldy to-dos and came across this, related to https://issues.redhat.com/browse/HIVE-2023

PR #1891, opened 2022-10-10
Merge Labels/Taints of MachineSets

suhanime:
When MachineSets were generated, we overrode the existing labels and taints on the remoteMachineSet to reflect what was present on the MachinePool. This commit preserves the existing labels and taints on the MachineSet as long as they do not clash with those on MachinePool.

/assign @abutcher 

xref: https://issues.redhat.com/browse/HIVE-2035

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -581,19 +581,23 @@ func (r *ReconcileMachinePool) syncMachineSets(
 					}
 				}
 
-				// Update if the labels on the remote machineset are different than the labels on the generated machineset.
+				// Merge the labels on the remote machineset, if different from the labels on the generated machineset.
 				// If the length of both labels is zero, then they match, even if one is a nil map and the other is an empty map.
-				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; (len(rl) != 0 || len(l) != 0) && !reflect.DeepEqual(rl, l) {
+				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; !reflect.DeepEqual(rl, l) {
 					msLog.WithField("desired", l).WithField("observed", rl).Info("labels out of sync")
-					rMS.Spec.Template.Spec.Labels = l
+					for key, value := range l {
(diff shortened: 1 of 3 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -581,19 +581,25 @@ func (r *ReconcileMachinePool) syncMachineSets(
 					}
 				}
 
-				// Update if the labels on the remote machineset are different than the labels on the generated machineset.
+				// Merge the labels on the remote machineset, if different from the labels on the generated machineset.
 				// If the length of both labels is zero, then they match, even if one is a nil map and the other is an empty map.
-				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; (len(rl) != 0 || len(l) != 0) && !reflect.DeepEqual(rl, l) {
+				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; len(rl) != len(l) ||
+					(len(rl) == len(l) && !reflect.DeepEqual(rl, l)) {
```
abutcher, line 587: I think we will need to test this by checking to see if the generated labels are a subset of the remote labels otherwise we will always report a difference / perform an update even if the generated labels were already present on the remote machineset. Same will apply to taints.

suhanime: So - generated labels and taints first import the same from MachinePool and then should preserve what is already present on the remote MachineSet. Do I understand that correctly?
If so - I added another test that ensures an update is not called when the labels/taints are already present - even if out of order (though reflect.DeepEqual doesn't really care about order so that was a moot point)

abutcher: Yes, we should determine that an update needs to be made if the labels we expect aren't in the existing labels.

suhanime:
@abutcher This is ready for review

pkg/controller/machinepool/machinepool_controller.go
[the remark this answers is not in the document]
suhanime: Thanks for catching it! Fixed

abutcher:
Tested that ProviderSpec labels were updated to include desired labels and were not stomped if there were pre-existing labels. :heavy_check_mark:

PR #1892, opened 2022-10-18
Get rid of OpenShift 3.11 cruft

2uasimojo:
We had some checks and tweaks for OpenShift 3.11, which has been EOL for a looong time. Clean up.

Of note: We also remove our write RBAC for CRDs. The reader perms were used to determine 3.11-ness, but are still used to determine supported contracts for ClusterInstall. The writer perms became obsolete at some point when we stopped doing
[this](https://github.com/openshift/hive/pull/385/files#diff-5b450b52ce1d65cb9d1e448269a23d1fc71a200dfa6462a5bc0a3cd2019fafa7R77-R88), but were never removed.

Of note: This partially reverts 5fecbe55 / #990, which added some yq magic to remove a field from our CRDs because 3.11 didn't like it. However, it seems controller-gen stopped generating that field anyway at some point, so this bit no longer had any effect anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,10 +105,9 @@ manifests: crd
 
 # controller-gen is adding a yaml break (---) at the beginning of each file. OLM does not like this break.
 # We use yq to strip out the yaml break by having yq replace each file with yq's formatting.
-# This also removes the spec.validation.openAPIV3Schema.type field which OpenShift 3.11 does not like.
 # $1 - CRD file
 define strip-yaml-break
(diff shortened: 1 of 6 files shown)


2uasimojo:
okay, we still need those CRD read perms for [this](https://github.com/openshift/hive/blob/4a7b257991442c7f08911462e97ca09a49f24574/pkg/operator/hive/configmap.go#L125-L128).

abutcher (approved):
Looks good to me!
/lgtm

PR #1893, opened 2022-10-19
Stop provisioning if fatal DNSZone conditions are met

dlom:
xref: https://issues.redhat.com/browse/HIVE-2036


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1630,6 +1635,26 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 		message,
 		controllerutils.UpdateConditionIfReasonOrMessageChange)
 	if dnsCondChanged {
+		if fatalCondition {
```
2uasimojo, line 1638: If we leave this nested in here, we might miss updating CDs that already exist in this state when this upgrade is deployed. To get around that, we would need to do something more like:

```go
var changed1, changed2 bool
if fatalCondition {
    cd.Status.Conditions, changed1 = controllerutils.SetClusterDeploymentConditionWithChangeCheck(...)
    cd.Status.Conditions, changed2 = controllerutils.SetClusterDeploymentConditionWithChangeCheck(...)
}
if dnsCondChanged || changed1 || changed2 {
    if err := r.Status().Update(...); err != nil { ... }
}
```

2uasimojo (commented):
I think this is good; just one little upgrade issue inline.

2uasimojo:
Thanks!

/lgtm

PR #1895, opened 2022-10-20
[docs] Add more detail to MachinePool section of usage docs.

abutcher:
Add more detail to the MachinePool section of usage docs to include information regarding MachinePool platform immutability as well as noting that a MachinePool for workers mirroring the workers defined in the install-config is not required but will result in Hive managing the worker MachineSets.

https://github.com/openshift/hive/pull/1891 should update this once `Labels` and `Taints` are merged instead of overridden.

/assign @2uasimojo 
/cc @suhanime

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -700,9 +700,17 @@ openstack:
 
 `MachinePool` is a YAML configuration by which you can create and scale worker nodes on a deployed cluster. A `MachinePool` will create `MachineSet` resources on the deployed cluster. If supported on your cloud, those MachineSets will automatically span all AZs, or you can specify an explicit list.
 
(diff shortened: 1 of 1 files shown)


2uasimojo (commented):
This looks reasonable to me.

Do we wish to reference the existing cards for the noted restrictions? Give the reader the opportunity to "vote" on them if they care to?

Do we wish to mention workaround(s) for "updating" a MachinePool? (Shenanigans with creating a new one, scaling the old one down, scaling the new one up, ...)

abutcher:
:+1: I'll add both of those suggestions.

2uasimojo:
:heart: 
/lgtm

PR #1898, opened 2022-10-27
Add regex for 'no route to host' proxy errors

tnierman:
Adds additional an `searchRegexString` to the `ProxyTimeout` reason to catch when routing misconfigurations in cluster-wide proxies cause install failures.


2uasimojo (commented):
Hi Trevor. This looks fine -- please just run `make update` to regenerate bindata and make `verify` pass.

PR #1899, opened 2022-10-28
Refine error checks

rna-afk:
Currently, the checks for BootstrapFailed are only checking for the generic error message. The installer checks to see if the API came up and if the status is set to complete
in the bootstrap config map before exiting with BootstrapFailed error and this can be checked to possibly see if the API came up before it errored out.

xref: https://issues.redhat.com/browse/HIVE-2047

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -255,16 +255,26 @@ data:
 
     # Generic OpenShift Install
     
-    - name: BootstrapFailed
-      searchRegexStrings:
-      - "Bootstrap failed to complete"
-      installFailingReason: BootstrapFailed
-      installFailingMessage: Installation Bootstrap failed to complete. Verify the networking configuration and account permissions and try again.
     - name: KubeAPIWaitTimeout
       searchRegexStrings:
       - "waiting for Kubernetes API: context deadline exceeded"
       installFailingReason: KubeAPIWaitTimeout
       installFailingMessage: Timeout waiting for the Kubernetes API to begin responding
+    - name: KubeAPIWaitFailed
+      searchRegexStrings:
+      - "Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane"
+      installFailingReason: KubeAPIWaitFailed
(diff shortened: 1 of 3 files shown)


suhanime:
/lgtm

Changes look good to me - the regex match for BootStrapFailed is made more specific instead of the catch-all during execution of the cobra command [here](https://github.com/openshift/installer/blob/dd9047c4c119e942331f702a4b7da85c60042da5/cmd/openshift-install/create.go#L132)

@2uasimojo Can you have a look too?

2uasimojo:
I have no problem with this change, though I have to take your word for it that it makes things better for SD.

Do we still need a more generic catch-all boostrap failure regex near the bottom of the pile?

You'll need to `make update` to regenerate bindata to get `verify` to pass.

rna-afk:
> Do we still need a more generic catch-all boostrap failure regex near the bottom of the pile?

Might be safe to do so I guess. I'll add it.

config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -255,16 +255,26 @@ data:
...
+      searchRegexStrings:
+      - "Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane"
+      installFailingReason: KubeAPIWaitFailed
+      installFailingMessage: Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane
+    - name: BootstrapFailed
+      searchRegexStrings:
+      - "Failed to wait for bootstrapping to complete. This error usually happens when there is a problem with control plane hosts that prevents the control plane operators from creating the control plane."
+      installFailingReason: BootstrapFailed
+      installFailingMessage: Failed to wait for bootstrapping to complete. This error usually happens when there is a problem with control plane hosts that prevents the control plane operators from creating the control plane. Verify the networking configuration and account permissions and try again.
+     - name: BootstrapFailed
```
2uasimojo, line 273: I think you have an extra space here that's causing the UT failure.

Also, I'm pretty sure you're going to need to give this a different `name` than the previous one.

2uasimojo (commented):
Please add a unit test or two that demonstrates how this change works. I.e.
- Before this delta, a given log that matches both regexes would yield BootstrapFailed; but with the change it would yield KubeAPIWaitFailed instead.
- Cover both of the bootstrap cases (with whatever their names/reasons end up being).

PR #1901, opened 2022-11-01
Add regex to catch another occurence of AWSInsufficientPermissions

mjlshen:
Per OSD-13578, we've identified another way that an install will fail on AWS due to insufficient permissions via logs like:

```
time="2022-10-14T18:40:17Z" level=error msg="Error: Error launching source instance: UnauthorizedOperation: You are not authorized to perform this operation. Encoded authorization failure message: ${encoded stuff}
```

So I added a regex search to `AWSInsufficientPermissions`

I will make a separate PR for the other failure method identified due to a blocking SCP since we will be able to give the customer a more specific error message in that case since they may have tried to give the necessary permissions, but are being blocked by an organizational SCP that they have forgotten about.


mjlshen:
Having issues running `make verify` on my M1 Mac - troubleshooting

PR #1902, opened 2022-11-01
Add support for custom http error pages

MuneebAijaz:
This PR adds support to override HttpErrorCodePages field in IngressController CR made via ClusterDeployment using Syncset. 

ref: https://issues.redhat.com/browse/HIVE-2040

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -675,6 +676,10 @@ type ClusterIngress struct {
 	// should be used for this Ingress
 	// +optional
 	ServingCertificate string `json:"servingCertificate,omitempty"`
+
+	// HttpErrorCodePages allows configuring custom HTTP error pages using the IngressController object
+	// +optional
+	HttpErrorCodePages configv1.ConfigMapNameReference `json:"httpErrorCodePages,omitempty"`
 }
 
 // ControlPlaneConfigSpec contains additional configuration settings for a target
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -679,7 +679,7 @@ type ClusterIngress struct {
 
 	// HttpErrorCodePages allows configuring custom HTTP error pages using the IngressController object
 	// +optional
-	HttpErrorCodePages configv1.ConfigMapNameReference `json:"httpErrorCodePages,omitempty"`
+	HttpErrorCodePages *configv1.ConfigMapNameReference `json:"httpErrorCodePages,omitempty"`
 }
 
 // ControlPlaneConfigSpec contains additional configuration settings for a target
--- a/pkg/controller/remoteingress/remoteingress_controller_test.go
+++ b/pkg/controller/remoteingress/remoteingress_controller_test.go
@@ -3,6 +3,7 @@ package remoteingress
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


pkg/controller/remoteingress/remoteingress_controller_test.go
```diff
@@ -362,6 +364,26 @@ func TestRemoteClusterIngressReconcile(t *testing.T) {
 				fmt.Sprintf("%s-secondCertBundle-secret", testClusterName),
 			},
 		},
+		{
+			name: "Test setting httpErrorCodePages",
+			localObjects: func() []runtime.Object {
+				objects := []runtime.Object{}
+				cd := addIngressToClusterDeployment(testClusterDeployment(), "secondingress", "moreingress.example.com", nil, nil, "", configv1.ConfigMapNameReference{})
```
2uasimojo, line 371: I suspect this was meant to be
```suggestion
				cd := addIngressToClusterDeployment(testClusterDeployment(), "secondingress", "moreingress.example.com", nil, nil, "", testHttpErrorCodePages())
```

pkg/controller/remoteingress/remoteingress_controller_test.go
```diff
@@ -362,6 +364,26 @@ func TestRemoteClusterIngressReconcile(t *testing.T) {
...
+					Status: corev1.ConditionUnknown,
+				})
+				objects = append(objects, cd)
+				return objects
+			}(),
+			expectedSyncSetIngressEntries: []SyncSetIngressEntry{
+				{
+					name:               testDefaultIngressName,
+					domain:             testIngressDomain,
+					HttpErrorCodePages: testHttpErrorCodePages(),
```
2uasimojo, line 383: ...and this isn't failing the test because we need an update to the `validateSyncSet` function. It explicitly checks each of the fields in each of the ingress controller entries; we need to add a line to check the HttpErrorCodePages.

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -675,6 +676,10 @@ type ClusterIngress struct {
 	// should be used for this Ingress
 	// +optional
 	ServingCertificate string `json:"servingCertificate,omitempty"`
+
+	// HttpErrorCodePages allows configuring custom HTTP error pages using the IngressController object
+	// +optional
+	HttpErrorCodePages configv1.ConfigMapNameReference `json:"httpErrorCodePages,omitempty"`
```
2uasimojo, line 682: I think it would be better if this was 
```suggestion
	HttpErrorCodePages *configv1.ConfigMapNameReference `json:"httpErrorCodePages,omitempty"`
```

Otherwise `omitempty` and `+optional` don't do what you expect them to; and it's also possible to break backward compatibility.

Note that if you make this change, you'll need to add nil checks wherever you reference the field.

2uasimojo (commented):
This is looking great! Just a couple of things noted inline.

pkg/controller/remoteingress/remoteingress_controller_test.go
```diff
@@ -539,13 +562,24 @@ func testRouteSelector() *metav1.LabelSelector {
...
+	if configmap != "" {
+		httpErrorCodePages := &configv1.ConfigMapNameReference{
+			Name: configmap,
+		}
+		return httpErrorCodePages
+	}
+	return nil
+}
+
+func addIngressToClusterDeployment(cd *hivev1.ClusterDeployment, ingressName, ingressDomain string, routeSelector, namespaceSelector *metav1.LabelSelector, servingCertificate string, httpErrorCodePages *configv1.ConfigMapNameReference) *hivev1.ClusterDeployment {
```
2uasimojo, line 575: This change seems a bit wasteful. Could simplify and reduce LOC count by just making `testClusterDeploymentWithHttpErrorCodePages()` explicitly add `cd.spec.Ingress[0].HttpErrorCodePages`. Or even do that inline in the test setup struct.

2uasimojo (commented):
/lgtm
/retest

Happy to re-review if you feel like making the change suggested inline (here if you catch it before it merges, or in a followon PR).

Thanks for the contribution!

PR #1903, opened 2022-11-02
Watch for ClusterSyncLease and renew if it is missing

dlom:
xref: https://issues.redhat.com/browse/HIVE-2038

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -223,6 +223,20 @@ func AddToManager(mgr manager.Manager, r *ReconcileClusterSync, concurrentReconc
 		return err
 	}
 
+	// Watch for changes to ClusterSync. These have the same name/namespace as the relevant
+	// ClusterDeployment, so when a ClusterSync watch triggers, the CD of the same name will be reconciled.
+	// When the CD reconciles, it will look up the related ClusterSync.
+	if err := c.Watch(&source.Kind{Type: &hiveintv1alpha1.ClusterSync{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
+
+	// Watch for changes to ClusterSyncLease. These have the same name/namespace as the relevant
+	// ClusterDeployment, so when a ClusterSyncLease watch triggers, the CD of the same name will be reconciled.
+	// When the CD reconciles, it will look up the related ClusterSyncLease.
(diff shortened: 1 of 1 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -417,7 +426,11 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 		ownerRef := metav1.NewControllerRef(cd, cd.GroupVersionKind())
 		ownerRef.Controller = nil
 		clusterSync.OwnerReferences = []metav1.OwnerReference{*ownerRef}
-		if err := r.Create(context.Background(), clusterSync); err != nil {
+		switch err := r.Create(context.Background(), clusterSync); {
+		case apierrors.IsAlreadyExists(err):
+			// race condition, just requeue
+			return reconcile.Result{Requeue: true}, nil
```
2uasimojo, line 432:
```suggestion
		case apierrors.IsAlreadyExists(err):
			// race condition, proceed
			break
```

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -417,13 +430,17 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
...
+		switch err := r.Create(context.Background(), clusterSync); {
+		case apierrors.IsAlreadyExists(err):
+			// race condition, just proceed
+		case err != nil:
 			logger.WithError(err).Log(controllerutils.LogLevel(err), "could not create ClusterSync")
 			return reconcile.Result{}, err
+		default: // clusterSync was created successfully
+			recobsrv.SetOutcome(hivemetrics.ReconcileOutcomeClusterSyncCreated)
+			// requeue immediately so that we reconcile soon after the ClusterSync is created
+			return reconcile.Result{Requeue: true}, nil
```
2uasimojo, line 442: This explicit requeue _should_ be redundant.
```suggestion
			// requeue immediately so that we reconcile soon after the ClusterSync is created
			return reconcile.Result{}, nil
```
but I'm good with leaving it here.

2uasimojo (commented):
/lgtm

2uasimojo:
actually

/hold

cause we talked about you looking at the logs on OSD stage before/after. Unhold once you've gotten your "before" picture.

PR #1905, opened 2022-11-03
Log ProvisionFailedTerminal metric if DNSZoneFailedToReconcile or DNSNotReadyTimedOut

suhanime:
hive_cluster_deployments_provision_failed_terminal_total should be logged everytime we set ProvisionStopped condition to true, however it was found that logging of this metric was skipped when DNSZoneFailedToReconcile and DNSNotReadyTimedOut. This commit fixes that.

xref: https://issues.redhat.com/browse/HIVE-2056

/assign @2uasimojo

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1603,6 +1603,8 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 	authenticationFailureCondition := controllerutils.FindCondition(dnsZone.Status.Conditions, hivev1.AuthenticationFailureCondition)
 	apiOptInRequiredCondition := controllerutils.FindCondition(dnsZone.Status.Conditions, hivev1.APIOptInRequiredCondition)
 	dnsErrorCondition := controllerutils.FindCondition(dnsZone.Status.Conditions, hivev1.GenericDNSErrorsCondition)
+	// Cache ProvisionStoppedCondition status, so later we can decide if ProvisionFailedTerminal metric needs to be observed.
+	provisionStoppedConditionStatus := controllerutils.FindCondition(cd.Status.Conditions, hivev1.ProvisionStoppedCondition).Status
 	var (
 		status                 corev1.ConditionStatus
 		reason, message        string
@@ -1675,6 +1677,12 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "could not update DNSNotReadyCondition")
 			return false, reconcile.Result{}, err
 		}
+	}
+
+	// Observe ProvisionFailedTerminal metric if we have set ProvisionStopped.
+	if controllerutils.FindCondition(cd.Status.Conditions, hivev1.ProvisionStoppedCondition).Status == corev1.ConditionTrue &&
+		provisionStoppedConditionStatus != corev1.ConditionTrue {
+		incProvisionFailedTerminal(cd)
 	}
 
 	// Only attempt to record the delay metric if the DNSZone is ready
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -4086,6 +4086,10 @@ func TestEnsureManagedDNSZone(t *testing.T) {
 					Type:   hivev1.DNSNotReadyCondition,
 					Status: corev1.ConditionUnknown,
 				}),
+				testclusterdeployment.WithCondition(hivev1.ClusterDeploymentCondition{
+					Type:   hivev1.ProvisionStoppedCondition,
+					Status: corev1.ConditionUnknown,
+				}),
 			),
 			expectedUpdated: true,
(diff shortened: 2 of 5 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -838,26 +838,27 @@ func (r *ReconcileClusterDeployment) reconcileInstallingClusterInstall(cd *hivev
 	return r.reconcileExistingInstallingClusterInstall(cd, logger)
 }
 
-func dnsZoneNotReadyMaybeTimedOut(cd *hivev1.ClusterDeployment, logger log.FieldLogger) (string, string, time.Duration) {
+func dnsZoneNotReadyMaybeTimedOut(cd *hivev1.ClusterDeployment, logger log.FieldLogger) (string, string, time.Duration, bool) {
 	cond := controllerutils.FindCondition(cd.Status.Conditions, hivev1.DNSNotReadyCondition)
+	var provisionStoppedChanged bool
```
2uasimojo, line 843: We could really do without this variable outside of the `timeToTimout` condition; just inline `false` everywhere except L861/878. Stylistic nit, feel free to ignore.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -838,26 +838,27 @@ func (r *ReconcileClusterDeployment) reconcileInstallingClusterInstall(cd *hivev
 	return r.reconcileExistingInstallingClusterInstall(cd, logger)
 }
 
-func dnsZoneNotReadyMaybeTimedOut(cd *hivev1.ClusterDeployment, logger log.FieldLogger) (string, string, time.Duration) {
+func dnsZoneNotReadyMaybeTimedOut(cd *hivev1.ClusterDeployment, logger log.FieldLogger) (string, string, time.Duration, bool) {
```
2uasimojo, line 841: Time to add a docstring to this func describing the inputs and outputs, noting that it has side effects on the `cd`. On that note, it sure would be nice if the bool return cleanly indicated that the cd was changed *at all* (indicating that the caller should do a Status().Update()). I get that the caller only wants to bump the metric if the `ProvisionStopped` condition has changed... but maybe we handle that a different way -- see below.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -874,13 +875,13 @@ func dnsZoneNotReadyMaybeTimedOut(cd *hivev1.ClusterDeployment, logger log.Field
 			controllerutils.UpdateConditionIfReasonOrMessageChange)
```
2uasimojo, line 875: If you go with the above suggestion (the `bool` return indicates the `cd` was updated) then we need to add a change check to this one as well, and return the `||` of the two.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1626,7 +1628,8 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 		message = dnsErrorCondition.Message
 	default:
 		status = corev1.ConditionTrue
-		reason, message, requeueAfter = dnsZoneNotReadyMaybeTimedOut(cd, cdLog)
+		// If dnsZoneNotReadyMaybeTimedOut, then ProvisionStopped should have been set to true
+		reason, message, requeueAfter, setProvisionFailedTerminalMetric = dnsZoneNotReadyMaybeTimedOut(cd, cdLog)
```
2uasimojo, line 1632: ...so the bool return here would be fed into L1664 just to decide whether to do a status update. But it would not be used to decide whether to increment the metric.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1664,6 +1668,11 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZone(cd *hivev1.ClusterDepl
 		}
 	}
 
+	// If ProvisionStopped condition was set to true, set the provision failed terminal metric
+	if (stoppedCondChanged || reason == dnsNotReadyTimedoutReason) && setProvisionFailedTerminalMetric {
```
2uasimojo, line 1672: This is really confusing. Here's a different idea:

At the beginning of this function, cache conds[ProvisionStopped].status. Then once we get here, we know we want to increment the metric if it has changed from False to True.

```go
// If we determined that the provision failed terminally for DNS reasons, the ProvisionStopped condition will have
// flipped from false to true. Note that we only want to record this metric once for such a transition -- so don't do
// that if it was true=>true.
if origProvisionStoppedStatus == ConditionFalse && nowProvisionStoppedStatus == ConditionTrue {
    incProvisionFailedTerminal(cd)
}
```

kind of thing.

2uasimojo (commented):
Will there be UT updates for this?

suhanime:
@2uasimojo Ready for review. This one is a much cleaner iteration.

Note for future me: For testing this, I used test cases [51195](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-51195) and [56011](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-56011). For DNSZoneFailedToReconcile, instead of deleting the creds, jumble them up so they are same length but invalid - so it successfully sets AuthFailure condition instead of Cloud Error. The later won't result in DNSZoneFailedToReconcile

2uasimojo (commented):
That _is_ cleaner :)

/lgtm

PR #1906, opened 2022-11-07
Add AWSDeniedBySCP failed install reason and regex

mjlshen:
Per OSD-13578, sometimes installs can fail due to a service control policy (SCP) that the customer may unknowingly be restricted by.

I chose to make this its own error message rather than bundling it into `AWSInsufficientPermissions` because the customer may have set up the roles as documented, but a different internal team may be responsible for an SCP, which isn't as obvious to them as a possible reason if we send the message: `AWS credentials are insufficient for performing cluster installation`


mjlshen:
/retest
one more time for a sanity check - seems like it's a Terraform provisioning failure in the e2e test?

abutcher (approved):
/lgtm

PR #1907, opened 2022-11-07
Bump to go 1.19

abutcher:
[HIVE-2034](https://issues.redhat.com//browse/HIVE-2034)

PR #1909, opened 2022-11-07
[OSD-13716] Deduplicate and consolidate SCACertsPullFailed error

mjlshen:
`SCACertsPullFailed` and `SCACertPullFailed` were distinct error messages, but targeting the same error.

Kind of funnily there are two main status conditions that have been observed recently that indicate this failure:
* `SCAAvailable is False` and
* `SCANotAvailable is True`

but the latter was getting overruled since it didn't match the first regex and getting caught as `BootstrapFailed` instead due to the ordering/priority in install log regexes.

This PR consolidates the two error messages into one and keeps the more generic regex so that both cases can be correctly identified. I kept `SCACertsPullFailed` because it already corresponds to the OCM error code `OCM3050`

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -7,7 +7,7 @@ data:
   regexes: |
     - name: SCACertsPullFailed
       searchRegexStrings:
-      - "SCAAvailable is False with NonHTTPError: Failed to pull SCA certs from"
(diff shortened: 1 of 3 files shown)


2uasimojo:
makes sense to me

/lgtm

Note that https://github.com/openshift/hive/pull/1907 is holding up CI here -- please wait until that merges to retest.

2uasimojo:
/test e2e

infra flake

PR #1913, opened 2022-11-15
work around for hive-1989

jianping-shu:
[HIVE-1989](https://issues.redhat.com/browse/HIVE-1989)

To override credentials secret for Azure, the following 3 fields are related to deployments and can't be nailed down until running "openshift-install create manifests"
azure_region
azure_resource_prefix
azure_resourcegroup

So a work around is to find out these fields from manifests files cluster-config.yaml and cluster-infrastructure-02-config.yml and fill into the azure override credentials secret.


pkg/installmanager/installmanager.go
```diff
@@ -1810,3 +1820,62 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
 
 	return nil
 }
+
+func (m *InstallManager) patchAzureOverrideCreds() error {
```
suhanime, line 1824: Unit tests to ensure patch is happening as expected would be a good idea

suhanime (commented):
/cc @2uasimojo

pkg/installmanager/installmanager.go
```diff
@@ -1810,3 +1820,62 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+	overrideCredsPath := filepath.Join(m.WorkDir, "manifests", overrideCredsYAML)
+	overrideCredsBytes, err := ioutil.ReadFile(overrideCredsPath)
+	if err != nil {
```
2uasimojo, line 1827: We need a check for file existence somewhere in here, right? Or is the file definitely supposed to exist by this point (whether generated by installer or provided by the user)?

pkg/installmanager/installmanager.go
```diff
@@ -1810,3 +1820,62 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+	}
+	region, err := exec.Command("bash", "-c", fmt.Sprintf("grep %s %s |awk -F ': ' '{print $2}'", "region", filepath.Join(m.WorkDir, "manifests", clusterConfigYAML))).Output()
+	if err != nil {
+		return errors.Wrap(err, "error to grep region from manifests")
+	}
+	infraName, err := exec.Command("bash", "-c", fmt.Sprintf("grep %s %s |awk -F ': ' '{print $2}'", "infrastructureName", filepath.Join(m.WorkDir, "manifests", clusterInfraConfigYAML))).Output()
+	if err != nil {
+		return errors.Wrap(err, "error to grep infrastructureName from manifests")
+	}
+	resourceGroupName, err := exec.Command("bash", "-c", fmt.Sprintf("grep %s %s |awk -F ': ' '{print $2}'", "resourceGroupName", filepath.Join(m.WorkDir, "manifests", clusterInfraConfigYAML))).Output()
```
2uasimojo, line 1845: Rather than spawning subshells, let's get these values "natively".

yamlpatch isn't the easiest to use, but here's an example of retrieving a nested value: https://go.dev/play/p/qV4eam0zsFY

pkg/installmanager/installmanager.go
```diff
@@ -1810,3 +1820,62 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+	if err != nil {
+		return errors.Wrap(err, "error to grep resourceGroupName from manifests")
+	}
+	regionBase64 := interface{}(base64.RawStdEncoding.EncodeToString(region))
+	infraNameBase64 := interface{}(base64.RawStdEncoding.EncodeToString(infraName))
+	resourceGroupNameBase64 := interface{}(base64.RawStdEncoding.EncodeToString(resourceGroupName))
```
2uasimojo, line 1851: Are these values really supposed to be base64-encoded?

2uasimojo (commented):
Thanks for this @jianping-shu!

Out of curiosity, who would QE this change? :P

pkg/installmanager/installmanager.go
jianping-shu: Yes, good idea, let me check if I can add this in installmanager_test.go

pkg/installmanager/installmanager.go
jianping-shu: Thanks for the tips! I’ll remove subshell.

pkg/installmanager/installmanager.go
jianping-shu: Yes, these are data in secret which shall be base64

pkg/installmanager/installmanager.go
```diff
@@ -1810,3 +1820,62 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+	ops := yamlpatch.Patch{
```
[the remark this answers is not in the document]
jianping-shu: Good question!
Accept is not good, since it may cause cluster resources are created in two resource groups (installer create one and image-registry uses the inputted resource group).

I am not sure if replace is good for user since user may not notice this change at all.

I prefer to report error if azure_resource_prefix/azure_resourcegroup are already supplied, region can be replaced, WDYT?

pkg/installmanager/installmanager.go
2uasimojo: Are the region, infrastructureName, and resourceGroupName also secret data? If so, why didn't they need to be decoded above? Anyway, I guess this will shake out in testing.

pkg/installmanager/installmanager.go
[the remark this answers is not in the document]
2uasimojo: I would prefer consistent behavior across all fields; but I agree with your reasoning. So let's report an error if any field is already specified *and* the value is different from whatever we're laying down.

pkg/installmanager/installmanager.go
jianping-shu: These are raw data in the manifests file so need base64 encoding before fill up.

pkg/installmanager/installmanager.go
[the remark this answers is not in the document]
jianping-shu: Sounds good.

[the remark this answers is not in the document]
jianping-shu: For simplicity, I just return error if any of azure_resource_prefix/azure_resourcegroup already exists, because the infra Id was generated randomly when running "openshift-install create manifests", so user is not able to configure it in advance. If need to handle the exist case, will make the later yamlpatch logic complicated.

pkg/installmanager/installmanager.go
jianping-shu: I included package github.com/tidwall/gjson since I found it is really simple to use, pls. let me know if it is NOT ok.

jianping-shu:
@2uasimojo With new commit, cluster installation was successful in my test. I am still checking its unit test code which is new to me :)

pkg/installmanager/installmanager.go
2uasimojo: Wait wait. I know we designed it so the two things would be used together for this use case, but empirically they're not actually related. So I still think we should add a file existence check and _not_ fail if the file isn't provided.

pkg/installmanager/installmanager.go
```diff
@@ -1803,10 +1807,93 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+
+	clusterInfraConfigJson, err := yaml.YAMLToJSONStrict(clusterInfraConfigBytes)
+	if err != nil {
+		return errors.Wrap(err, "error converting cluster-infrastructure-02-config.yml to json")
+	}
+
+	infraName := gjson.Get(string(clusterInfraConfigJson), `status.infrastructureName`).String()
+	resourceGroupName := gjson.Get(string(clusterInfraConfigJson), `status.platformStatus.azure.resourceGroupName`).String()
+	if region == "" || infraName == "" || resourceGroupName == "" {
+		return fmt.Errorf("error reading from manifests, region=%s, infrastructureName=%s, resourceGroupName=%s", region, infraName, resourceGroupName)
```
2uasimojo, line 1838: minor point: perhaps we should include the whole `clusterInfraConfigJson` in the error message as well, as it would probably make it a lot easier to figure out what the problem was. (Could be field missing, misspelled, wrong path, ...)

pkg/installmanager/installmanager.go
```diff
@@ -1803,10 +1807,93 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
 
 	m.log.WithField("newMode", credentialsMode).Info("overrode effective in-cluster credentials mode")
 
+	if cd.Spec.Platform.Azure == nil {
+		return nil
+	}
+
+	//If Azure, need fill up azure_resource_prefix and azure_resourcegroup for 99_cloud-creds-secret.yaml
+	m.log.Info("patch 99_cloud-creds-secret.yaml when platform is azure")
+	region := installConfig.Platform.Azure.Region
```
2uasimojo, line 1821: At first glance, I couldn't determine that this field would always be set; but I found that we have [webhook validation](https://github.com/openshift/hive/blob/b6d323da70b09053352d87f8ec47ff2a5be2992e/pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go#L479-L481) to make sure it is, so this is okay.

pkg/installmanager/installmanager.go
2uasimojo: I'm okay with it. We have some part of tidwall's suite in our go.sum already for some reason, so :shrug: 

@suhanime @abutcher @dlom do y'all see any concerns pulling this lib in?

pkg/installmanager/installmanager.go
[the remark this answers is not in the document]
2uasimojo: I'm fine with that for those two fields. But now I think I see your issue with the region. Since that field doesn't rely on the infra ID (or anything generated, right?) it would be totally reasonable for the user to include it in their 99_cloud-creds-secret.yaml file, and we shouldn't error if that's the case -- unless it doesn't match what they put in the cd.spec.

So, sorry to change my mind, I think it's worthwhile to split that out into a separate condition.

pkg/installmanager/installmanager.go
abutcher: No concerns, looks to be a quite popular lib.

pkg/installmanager/installmanager.go
jianping-shu: Added Test_patchAzureOverrideCreds in installmanager_test.go

pkg/installmanager/installmanager.go
jianping-shu: OK. I changed the logic, if the 99_cloud-creds-secret.yaml doesn't exist then go ahead for install

pkg/installmanager/installmanager.go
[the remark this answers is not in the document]
jianping-shu: OK. I changed the logic as conclusion.
If infraName/resourceGroupName already exist, report error
If region exists and different to the one in cd.spec, report error

pkg/installmanager/installmanager.go
```diff
@@ -1803,10 +1807,93 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+
+	//Read infrastructureName/resourceGroupName from manifests cluster-infrastructure-02-config.yml
+	clusterInfraConfigPath := filepath.Join(m.WorkDir, "manifests", clusterInfraConfigYAML)
+	clusterInfraConfigBytes, err := ioutil.ReadFile(clusterInfraConfigPath)
+	if err != nil {
+		return errors.Wrap(err, "error reading manifests cluster-infrastructure-02-config.yml")
+	}
+
+	clusterInfraConfigJson, err := yaml.YAMLToJSONStrict(clusterInfraConfigBytes)
```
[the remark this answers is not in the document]
jianping-shu: Yes, changed to YAMLToJSON

pkg/installmanager/installmanager.go
jianping-shu: ok

pkg/installmanager/installmanager.go
jianping-shu: Done

jianping-shu:
@2uasimojo PR is ready for merge, pls. let me know if further comment

pkg/installmanager/installmanager.go
```diff
@@ -1803,10 +1807,119 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+		},
+		yamlpatch.Operation{
+			Op:    "add",
+			Path:  yamlpatch.OpPath("/data/azure_resourcegroup"),
+			Value: yamlpatch.NewNode(&resourceGroupNameBase64),
+		},
+	}
+	if regionInput != "" {
+		ops = ops[1:3]
+	}
```
2uasimojo, line 1917: ah, this would mitigate the above.

IMO the code would be a bit prettier if the logic were inverted.
```go
ops := yamlpatch.Patch{ ... resource_prefix and resourcegroup operations ...}
if regionInput == "" {
    ops = append(ops, ... region operation ...)
}
```

2uasimojo (commented):
Thanks Jianping, this is getting really close. Just a couple of small tweaks.

pkg/installmanager/installmanager.go
```diff
@@ -1803,10 +1807,119 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+
+	//Return error if region exists but different to the one in cd.spec
+	if regionInput != "" {
+		regionDecode, err := base64.StdEncoding.DecodeString(regionInput)
+		if err != nil {
+			return nil, errors.Wrap(err, "error decoding data.azure_region")
+		}
+		if region != string(regionDecode) {
+			return nil, fmt.Errorf("azure_region=%s already exists in 99_cloud-creds-secret.yaml but different to region in ClusterDeployment.spec", string(regionDecode))
```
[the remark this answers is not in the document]
jianping-shu: Done

pkg/installmanager/installmanager_test.go
```diff
@@ -863,3 +863,80 @@ spec:
...
+			overrideSecretBytes: []byte(`---
+apiVersion: v1
+data:
+  azure_client_id: YWFhCg==
+`),
+			expectModified: true,
+			expectErr:      false,
+		},
+		{
+			name: "Patch applies successfully with region exists",
```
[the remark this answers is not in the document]
jianping-shu: yes, done

pkg/installmanager/installmanager.go
```diff
@@ -1803,10 +1807,119 @@ func (m *InstallManager) overrideCredentialsModeOnInstallConfig(credentialsMode
...
+	}
+
+	regionBase64 := interface{}(base64.StdEncoding.EncodeToString([]byte(region)))
+	infraNameBase64 := interface{}(base64.StdEncoding.EncodeToString([]byte(infraName)))
+	resourceGroupNameBase64 := interface{}(base64.StdEncoding.EncodeToString([]byte(resourceGroupName)))
+
+	ops := yamlpatch.Patch{
+		yamlpatch.Operation{
+			Op:    "add",
+			Path:  yamlpatch.OpPath("/data/azure_region"),
```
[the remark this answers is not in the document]
jianping-shu: Good catch. I thought "add" operation may fail for the existing node but turns out it has no problem in my testing.I'll go this way.

pkg/installmanager/installmanager_test.go
```diff
@@ -863,3 +863,80 @@ spec:
...
+			expectModified: false,
+			expectErr:      true,
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			modifiedBytes, err := patchAzureOverrideCreds([]byte(tc.overrideSecretBytes), clusterInfraConfigBytes, "centralus")
+			if tc.expectModified {
+				assert.NotNil(t, modifiedBytes, "expected credential secret to be modified")
```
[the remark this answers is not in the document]
jianping-shu: Yes, done

jianping-shu:
@2uasimojo Code updated, tests also passed with the updated case [OCP-53364](https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-53364) PTAL

pkg/installmanager/installmanager_test.go
```diff
@@ -863,3 +864,100 @@ spec:
...
+`),
+			expectModified: true,
+			expectErr:      false,
+		},
+		{
+			name: "Patch fails due to region exists but different",
+			overrideSecretBytes: []byte(`---
+apiVersion: v1
+data:
+  azure_region: ZWFzdHVzCg==//eastus
```
2uasimojo, line 911: I don't think you're actually getting a comment here, as slashes are valid in both yaml and base64. By chance, since `//eastus` happens to have 8c, the whole value will decode without error to `eastus<binary garbage>`, which will still hit the desired code path since it doesn't match `centralus`. But it would better demonstrate your intent if you put the comment outside of the byte array.

pkg/installmanager/installmanager_test.go
```diff
@@ -863,3 +864,100 @@ spec:
...
+			}
+			if tc.expectModified {
+				c, err := yamlutils.Decode(*modifiedBytes)
+				assert.NoError(t, err, "expected to be able to decode patched credential secret")
+				isRegionCorrect, _ := yamlutils.Test(c, "/data/azure_region", "Y2VudHJhbHVz")
+				assert.Equal(t, true, isRegionCorrect, "expected /data/azure_region filled correctly in patched credential secret")
+				isPrefixCorrect, _ := yamlutils.Test(c, "/data/azure_resource_prefix", "aGl2ZS1jbHVzdGVyLWc3ZnFi") //base64 for hive-cluster-g7fqb
+				assert.Equal(t, true, isPrefixCorrect, "expected /data/azure_resource_prefix filled correctly in patched credential secret")
+				isGroupCorrect, _ := yamlutils.Test(c, "/data/azure_resourcegroup", "aGl2ZS1jbHVzdGVyLWc3ZnFiLXJn") //base64 for hive-cluster-g7fqb-rg
+				assert.Equal(t, true, isGroupCorrect, "expected /data/azure_resourcegroup filled correctly in patched credential secret")
```
2uasimojo, line 959: nit: `assert.True()`

2uasimojo (commented):
Thanks Jianping.

/lgtm

I'm going to 

/hold

this in case you want to fix the couple of nits in this PR; but feel free to unhold and submit a followon instead.

pkg/installmanager/installmanager_test.go
jianping-shu: Thanks Eric, just fixed

pkg/installmanager/installmanager_test.go
jianping-shu: done

jianping-shu:
@2uasimojo Thanks for reviews! Finally it is ready :)

PR #1917, opened 2022-11-18
SCACertsPullFailed is no longer a blocking reason an install can fail

mjlshen:
Per OSD-13933, this is no longer a blocking install failure reason. Especially since it's at the top of the list, it has been causing confusion with customers. 

I don't think it has value to even point SRE too because it may confuse us if this is a "known install failure reason" when in fact it is no longer blocking a successful install, so I am removing it!

It will also be removed from OCM as well.

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -5,11 +5,6 @@ metadata:
   namespace: hive
 data:
   regexes: |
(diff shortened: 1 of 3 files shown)


suhanime:
@2uasimojo This makes sense to me, but have never removed a regex before. Stamp if you agree

2uasimojo:
Yah, shouldn't be a problem.

/lgtm

PR #1920, opened 2022-11-22
add namespace parameter to sample adoption cmd

katherinelc321:
This PR adds the namespace parameter to the sample cmd to adopt a hive cluster. During an adoption process in production we accidentally adopted a hive into the default namespace. This parameter will specify the namespace


2uasimojo (approved):
/lgtm

I'm fine with this. If you don't specify, the cluster will be adopted into your current namespace (which may or may not be `default`, pursuant to [thread](https://coreos.slack.com/archives/CE3ETN3J8/p1669123463560699)). But no harm in being explicit.

PR #1914, opened 2022-11-15
Revert "Merge pull request #1903 from dlom/HIVE-2038"

dlom:
This reverts commit 6568c2041d2a92956b430730fd599258166711f1, reversing changes made to 18e4e87e11b63c6c7231e1645bf648b214787c55.

/assign @abutcher

PR #1921, opened 2022-11-23
Syncset Testing PR #1

dlom:
This PR adds re-adds the functionality that was reverted in #1914, additionally adding some logs to better capture what's really going on in the stage environment.  There are 2 more subsequent related PRs:
* #1923 - reverts the functional changes included in this PR but retains the logging
* #1924 - reverts the non-functional logging changes included in this PR

These PRs are supersets of this PR and contain the same commit SHAs.  They should be merged in order.

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -223,6 +223,20 @@ func AddToManager(mgr manager.Manager, r *ReconcileClusterSync, concurrentReconc
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -924,16 +946,17 @@ func applyToTargetCluster(
 		return err
 	}
 
+	logger.Info("attempting to apply resource")
```
2uasimojo, line 949: Think we should put the UID/ResourceVersion in this one too? Esp the latter, I would expect it to change on apply.

PR #1922, opened 2022-11-28
Add some more 'debug' log messages for stage (revert this before next prod push)

dlom:
/assign @abutcher

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -909,6 +909,8 @@ func applyToTargetCluster(
 	applyFn func(obj []byte) (resource.ApplyResult, error),
 	logger log.FieldLogger,
 ) error {
+	logger = logger.WithField("uid", obj.GetUID()).WithField("resource_version", obj.GetResourceVersion())
+
 	startTime := time.Now()
 	labels := obj.GetLabels()
 	if labels == nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -924,16 +924,17 @@ func applyToTargetCluster(
...
 	}
 
+	logger.WithField("uid", obj.GetUID()).Info("attempting to apply resource")
 	applyResult, err := applyFn(bytes)
 	// Record the amount of time we took to apply this specific resource. When combined with the metric for duration of
 	// our kube client requests, we can get an idea how much time we're spending cpu bound vs network bound.
 	applyTime := metav1.Now().Sub(startTime).Seconds()
 	if err != nil {
-		logger.WithError(err).Warn("error applying resource")
+		logger.WithError(err).WithField("uid", obj.GetUID()).WithField("applyTime", applyTime).Warn("error applying resource")
```
abutcher, line 933: ✓ I will be interested to see what the per-resource values are for `applyTime` in stage.

abutcher (approved):
Looks good for understanding what the `applyTimes` look like. @2uasimojo for the stamp in case he has something else we could include for checking.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -909,6 +909,8 @@ func applyToTargetCluster(
 	applyFn func(obj []byte) (resource.ApplyResult, error),
 	logger log.FieldLogger,
 ) error {
+	logger = logger.WithField("uid", obj.GetUID()).WithField("resource_version", obj.GetResourceVersion())
```
2uasimojo, line 912: This will statically assign these field values at this point in time. We want to retrieve the values on the updated obj after the apply. So we need to explicitly add them with each log directive.

PR #1925, opened 2022-12-02
Add support for optional cluster type labels to cluster deployment and cluster provision metrics

suhanime:
In addition to the cluster-type label, it would be helpful for users to be able to provide their own set of labels (like STS, Private Link, Managed VPC etc) to look for while logging metrics. Allow support for such optional labels for metrics of ClusterDeployment and ClusterProvision controller

xref: https://issues.redhat.com/browse/HIVE-2045

/assign @2uasimojo


pkg/constants/constants.go
```diff
@@ -430,6 +430,18 @@ const (
...
+	STSClusterLabel = "api.openshift.com/sts"
+
+	// 	PrivateLinkClusterLabel is an optional label that can be applied to ClusterDeployments to denote it is uses
+	//	private-link. It is used as a label for some metrics.
+	PrivateLinkClusterLabel = "api.openshift.com/private-link"
+
+	// 	ManagedVPCLabel is an optional label that can be applied to ClusterDeployments to denote the VPC is managed by
+	//	SD. It is used as a label for some metrics.
+	ManagedVPCLabel = "api.openshift.com/managed-vpc"
+
```
2uasimojo, line 444: This is tickling my proprietary-ness antenna. As a rule, it's not a great look to add code that's explicitly tied to a particular consumer. Of course we add stuff at the request of consumers and for specific consumer use cases, but we usually try to design/position it so it's generically useful, at least in theory.

Ideally, we would find some way to discover and report these metric labels without using the SD-specific CD labels. E.g. we can certainly tell whether we're privatelink based on `spec.platform.aws.privateLink != nil`. Not sure about the others.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -567,7 +567,10 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 			cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error adding finalizer")
 			return reconcile.Result{}, err
 		}
-		metricClustersCreated.WithLabelValues(hivemetrics.GetClusterDeploymentType(cd)).Inc()
+		metricClustersCreated.WithLabelValues(hivemetrics.GetClusterDeploymentType(cd, hivev1.HiveClusterTypeLabel),
+			hivemetrics.GetClusterDeploymentType(cd, constants.STSClusterLabel),
+			hivemetrics.GetClusterDeploymentType(cd, constants.PrivateLinkClusterLabel),
+			hivemetrics.GetClusterDeploymentType(cd, constants.ManagedVPCLabel)).Inc()
```
2uasimojo, line 573: As mentioned elsewhere, once we get into more than one label, we should really start using `With()` rather than relying on the label order via `WithLabelValues()`.

Also as discussed, it would be cool to DRY the repetitious currying into a common function. (That way if, say, we needed to change how the label values are computed, we could do it in one place :) )

pkg/controller/metrics/metrics.go
```diff
@@ -818,6 +818,54 @@ func readMetricsConfig() (*metricsconfig.MetricsConfig, error) {
...
+	}
+	return mapClusterTypeLabelToValue, optionalClusterTypeLabels
+}
+
+// LogHistogramMetricWithOptionalLabels is used to log a histogram type metric with support for optional cluster type
+// labels. Caller should ensure there isn't an overlap between fixed and optional labels.
+func LogHistogramMetricWithOptionalLabels(metric *prometheus.HistogramVec, value float64, obj metav1.Object,
+	fixedLabels map[string]string, optionalLabels map[string]string, log log.FieldLogger) {
+	// Build a partial label vector for non-optional metrics
+	curriedVec, err := metric.CurryWith(fixedLabels)
```
suhanime, line 838: I had to go with `CurryWith` because `MustCurryWith` resulted in (rare) panics, and it is better to not log the metric and exit instead of panicking.

pkg/controller/metrics/metrics.go
```diff
@@ -818,6 +818,54 @@ func readMetricsConfig() (*metricsconfig.MetricsConfig, error) {
...
+		// The possibility of hitting an error here is low, but report it and don't observe the metric in case it happens
+		log.WithError(err).Error("error observing metric")
+	} else {
+		curriedVec.WithLabelValues().Observe(value)
+	}
+}
+
+// LogCounterMetricWithOptionalLabels is used to log a counter type metric with support for optional cluster type
+// labels. Caller should ensure there isn't an overlap between fixed and optional labels.
+func LogCounterMetricWithOptionalLabels(metric *prometheus.CounterVec, obj metav1.Object, fixedLabels map[string]string,
```
suhanime, line 853: I tried to make the logging of different types of metrics more dynamic, unfortunately the way `WithLabelValues`, `With`, `CurryWith`, `Observe` etc are implemented are different for histograms and summaries vs counters, and I couldn't find a generic interface to allow for it.

suhanime (commented):
Configs to hiveConfig can be made as following:
```
{
    "apiVersion": "hive.openshift.io/v1",
    "kind": "HiveConfig",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"hive.openshift.io/v1\",\"kind\":\"HiveConfig\",\"metadata\":{\"annotations\":{},\"name\":\"hive\"},\"spec\":{\"logLevel\":\"debug\",\"targetNamespace\":\"hive\"}}\n"
        },
        "creationTimestamp": "2022-08-31T21:48:52Z",
        "generation": 3,
        "name": "hive",
        "resourceVersion": "5436134",
        "uid": "5daecfa7-1110-4e12-aece-29b14db5356b"
    },
    "spec": {
        "logLevel": "debug",
        "metricsConfig": {
            "metricsWithClusterTypeLabels": {
                "managed_vpc": "api.openshift.com/managed-vpc",
                "private_link": "api.openshift.com/private-link",
                "sts": "api.openshift.com/sts"
            }
        },
        "targetNamespace": "hive"
    },
    "status": {
        "aggregatorClientCAHash": "369ab62e42c01c06450f6e86a9d4753c",
        "conditions": [
            {
                "lastProbeTime": "2022-12-13T17:04:24Z",
                "lastTransitionTime": "2022-12-13T17:04:24Z",
                "message": "Hive is deployed successfully",
                "reason": "DeploymentSuccess",
                "status": "True",
                "type": "Ready"
            }
        ],
        "configApplied": true,
        "observedGeneration": 3
    }
}
```

Screenshot of example metric to ensure it works well with and without optional labels being provided:
image

@2uasimojo ready for review

apis/hive/v1/metricsconfig/metricsConfig.go
```diff
@@ -4,4 +4,10 @@ type MetricsConfig struct {
 	// Optional metrics and their configurations
 	// +optional
 	MetricsWithDuration []MetricsWithDuration `json:"metricsWithDuration"`
+	// MetricsWithClusterTypeLabels is a map of cluster type to the label applied on the cluster deployment.
+	// This map is meant to be used when it would be helpful to label certain metrics with more cluster type labels.
+	// If the provided label is present on a cluster deployment, all metrics that are logged per cluster type will
+	// attempt to label with the provided cluster type as well.
+	// +optional
+	MetricsWithClusterTypeLabels map[string]string `json:"metricsWithClusterTypeLabels"`
```
2uasimojo, line 12: _bikesheds_

The `MetricsWith` prefix indicates that the field will contain a list of metrics (in the case of `MetricsWithDuration`, they're identified by name), so probably not appropriate here.

`ClusterType` isn't as precise as I would like because a) we already have a metric using a label called `cluster-type`, b) "type" is like a super vague word anyway, and c) there's nothing restricting the user to just cluster-type-ish labels -- e.g. they could use `version-major-minor-patch` (I guess you could argue that version is a "type", but see b).

I'm open to other suggestions, but I'm thinking `AdditionalClusterDeploymentLabels`. This helps indicate that the labels are coming from ClusterDeployment objects; and that they'll be included in addition to whatever baseline labels we have hardcoded. (Even if we decide to fold `hive.openshift.io/cluster-type` into these, we still have the library-ish ones like `job` and `instance`, so "Additional" is still accurate.)

Also, as discussed in slack, this docstring ought to include the caveat that labels with unbounded values, such as cluster namespaces, names, IDs, etc., should be avoided due to cardinality concerns.

Also, we need _some_ way of documenting which metrics are subject to this feature. I can't think of a great way to do this that's guaranteed not to go stale... can you? Even pointing to the code is fraught -- there's _already_ two different places we set them up.

Also, should this be `omitempty`?

As for the docstring's explanation of what this field does, I'm trying to figure out how to wordsmith it to be a bit clearer. Here's an attempt:

AdditionalClusterDeploymentLabels allows configuration of additional labels to be applied to certain metrics. The keys can be any string value suitable for a metric label (see [link to prom docs describing that format]). The values can be any ClusterDeployment label key (from `metadata.labels`). When observing an affected metric, hive will label it with the specified metric key, and copy the value from the specified ClusterDeployment label. For example, including `{"ocp_major_version": "hive.openshift.io/version-major"}` will cause affected metrics to include a label key `ocp_major_version` with the value from the `hive.openshift.io/version-major` ClusterDeployment label -- e.g. "4".
NOTE: Avoid ClusterDeployment labels whose values are unbounded, such as those representing cluster names or IDs, as these will cause your prometheus database to grow indefinitely.
[And here's where we should list or point to which metrics are affected.]

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -77,6 +77,15 @@ var (
 		// ImageSet condition initialized by cluster deployment
 		hivev1.InstallerImageResolutionFailedCondition,
 	}
+
+	// mapClusterTypeLabelToValue is a map that corresponds to MetricsWithClusterTypeLabels provided in metrics config.
+	// The key in the map is the name of the label used in defining the metric, and the value would be the cluster
+	// deployment label to look for while reporting the metric
```
2uasimojo, line 83: yeah, more like this

pkg/controller/clusterprovision/metrics.go
```diff
@@ -2,29 +2,51 @@ package clusterprovision
...
+	// Declare the metrics which allow optional labels to be added.
+	// They are defined later once the hive config has been read.
+	metricClusterProvisionsTotal *prometheus.CounterVec
+	metricInstallErrors          *prometheus.CounterVec
+
+	metricInstallFailureSeconds *prometheus.HistogramVec
+	metricInstallSuccessSeconds *prometheus.HistogramVec
+)
+
+func registerMetrics() {
```
2uasimojo, line 18: Unless there's a really good reason to do otherwise, we should make `optionalClusterTypeLabels` a param to this func rather than a global.

pkg/controller/metrics/metrics.go
```diff
@@ -744,10 +744,9 @@ func (ca *clusterAccumulator) setMetrics(total, installed, uninstalled, deprovis
 	}
 }
 
-// GetClusterDeploymentType returns the value of the hive.openshift.io/cluster-type label if set,
-// otherwise a default value.
-func GetClusterDeploymentType(obj metav1.Object) string {
-	if typeStr, ok := obj.GetLabels()[hivev1.HiveClusterTypeLabel]; ok {
+// GetClusterDeploymentType returns the value of the label if set, otherwise a default value.
+func GetClusterDeploymentType(obj metav1.Object, label string) string {
```
2uasimojo, line 748: Suggest a name more like `GetClusterDeploymentLabelValue`

pkg/controller/metrics/metrics.go
```diff
@@ -819,6 +818,54 @@ func readMetricsConfig() (*metricsconfig.MetricsConfig, error) {
...
 }
 
+// GetOptionalClusterTypeLabels reads the MetricsWithClusterTypeLabels from the metrics config and returns the same as a
+// map, as well as a slice of metric labels. The latter is helpful for defining the relevant metrics
+func GetOptionalClusterTypeLabels(mConfig *metricsconfig.MetricsConfig) (map[string]string, []string) {
+	var mapClusterTypeLabelToValue = make(map[string]string)
+	var optionalClusterTypeLabels []string
+	for k, v := range mConfig.MetricsWithClusterTypeLabels {
+		mapClusterTypeLabelToValue[k] = v
+		optionalClusterTypeLabels = append(optionalClusterTypeLabels, k)
```
2uasimojo, line 828: `range` is going to iterate over the map keys in random order, which can differ from one invocation to the next. At some point I suspect it'll bite us if `optionalClusterTypeLabels` is in a nondeterministic order. Suggest sorting it.

pkg/controller/metrics/metrics.go
2uasimojo: I don't have an objection to using this form... but this ought to be coded such that it can't panic. Were you observing panics with certain keys? certain values? Surely it wasn't just random.

pkg/controller/metrics/metrics.go
2uasimojo: Right, so this begs for an Interface that we define. And in fact, this may help us with the boggle I posed above wrt how consumers could discover which metrics are affected. To wit:

If we define an interface like

```go
type MetricWithOptionalLabels interface {
    CurryOptionalLabels(map[string]string)
}
```

...then we can define thin supertypes for the metrics we care about, e.g.

```go
type CounterMetricWithOptionalLabels struct {
    prometheus.CounterVec
}

// Assert we've implemented the method right
var _ MetricWithOptionalLabels = CounterMetricWithOptionalLabels{}

// CurryOptionalLabels implements MetricWithOptionalLabels
func (metric *CounterMetricWithOptionalLabels) CurryOptionalLabels(labels map[string]string) {
    // ... code here ...
}
```

The benefits are twofold:
1. When we declare the metrics in question, we declare them as their appropriate subtype, and then we can just call `CurryOptionalLabels()` on them.
2. Our docstring can say, "Affected metrics are instances of types that implement the MetricWithOptionalLabels interface," and consumers can go do some fancy F12ing in VSCode or whatever to figure out which ones those are.

pkg/controller/metrics/metrics.go
```diff
@@ -819,6 +818,54 @@ func readMetricsConfig() (*metricsconfig.MetricsConfig, error) {
 	return config, nil
 }
 
+// GetOptionalClusterTypeLabels reads the MetricsWithClusterTypeLabels from the metrics config and returns the same as a
+// map, as well as a slice of metric labels. The latter is helpful for defining the relevant metrics
+func GetOptionalClusterTypeLabels(mConfig *metricsconfig.MetricsConfig) (map[string]string, []string) {
```
2uasimojo, line 823: As discussed in slack, IMO we should force the user to specify `hive.openshift.io/cluster-type` explicitly if they want it. To accommodate a smoother transition, we could default to that if the config isn't used:

```go
if mConfig.MetricsWithClusterTypeLabels == nil {
    return map[string]string{"cluster_type": "hive.openshift.io/cluster-type"}, []string{"cluster_type"}
}
// ... else carry on with what you've got below
```

2uasimojo (commented):
I didn't delve into the low-level details yet -- can do that once we've agreed on the more major things discussed inline.

docs/hive-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Hive Metrics
+
+Hive publishes metrics, which can help admins to monitor the Hive operations. Most of these metrics are always published, few are [optional](#metrics-with-optional-cluster-deployment-labels).
+
+### Exporting Hive Metrics
+`HiveConfig.Spec.ExportMetrics` specifies whether the operator should enable metrics for hive controllers to be extracted for the prometheus instance running on the cluster. When set to true, `openshift.io/cluster-monitoring` label is applied and  the operator deploys ServiceMonitors so that the prometheus instances can extract metrics. The operator also sets up RBAC in the TargetNamespace so that openshift prometheus in the cluster can list/access objects required to pull metrics.
+
```
suhanime, line 22: Is this part clear about what's the default behaviour vs exporting? I wanted to incorporate the learnings from the recent slack discussions around this

pkg/controller/clusterdeployment/metrics.go
```diff
@@ -83,12 +62,67 @@ func incProvisionFailedTerminal(cd *hivev1.ClusterDeployment) {
...
+	metricProvisionFailedTerminal = hivemetrics.CounterWithDynamicLabels{
+		CounterVec: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "hive_cluster_deployments_provision_failed_terminal_total",
+			Help: "Counter incremented when a cluster provision has failed and won't be retried.",
+		},
+			append([]string{"clusterpool_namespacedname", "failure_reason"}, optionalClusterTypeLabels...)),
+		FixedLabels: map[string]string{
+			// These values must be filled before logging the metric
+			"clusterpool_namespacedname": constants.MetricLabelDefaultValue,
+			"failure_reason":             constants.MetricLabelDefaultValue,
```
suhanime, line 121: I wasn't sure if I should pass around an empty map here or do it this way. wdyt?

suhanime (commented):
@2uasimojo I think this addresses all of your comments. PTAL

pkg/controller/metrics/metrics.go
2uasimojo: Cool. Assuming there's nowhere we need to curry without observing or vice versa, that's a cleaner model anyway.

apis/hive/v1/metricsconfig/metricsConfig.go
2uasimojo: The point of `omitempty` is to avoid returning the field from etcd if it's in the schema (as understood by k8s) but not set in the database. This is to avoid compatibility problems where we retrieve such an object and try to unmarshal/use it in an older version of code where the (compiled-in) schema lacks that field. So the only way we get into trouble in our case is if we've installed the new CRD but are still running the old code -- we would panic as soon as we try to retrieve the HiveConfig object.

I'm going to say the lack of `omitempty` on `MetricsWithDuration` was an oversight rather than intentional, and we should have it on both fields.

docs/hive-metrics.md
2uasimojo: Maybe we want to defer cleaning this up until we have a better understanding. Based on the (partial) research I've just done, I think the label is only relevant if your `targetNamespace` has an `openshift-` prefix; and we lay the ServiceMonitor and RBAC stuff down as a convenience. However, based on DPTP's experience (the slack discussions you allude to) something seems to be wrong with this picture.

pkg/controller/clusterdeployment/metrics.go
2uasimojo: Not sure what you're asking. I dig the idea of `FixedLabels` and `OptionalLabels`. Though possibly `FixedLabels` should just be a slice with the keys -- what's the value of having it be a map?

pkg/controller/metrics/metricsWithDynamicLabels.go
```diff
@@ -0,0 +1,91 @@
...
+
+// GetOptionalClusterTypeLabels reads the AdditionalClusterDeploymentLabels from the metrics config and returns the same
+// as a map
+func GetOptionalClusterTypeLabels(mConfig *metricsconfig.MetricsConfig) map[string]string {
+	var mapClusterTypeLabelToValue = make(map[string]string)
+	// Currently "cluster_type" is a non-optional label for these metrics.
+	// TODO: remove as a part of https://issues.redhat.com/browse/HIVE-2077
+	if _, ok := mConfig.AdditionalClusterDeploymentLabels["cluster_type"]; !ok {
+		mapClusterTypeLabelToValue["cluster_type"] = hivev1.HiveClusterTypeLabel
+	}
```
2uasimojo, line 76: Hm.

I think I would rather add this default *only* if `AdditionalClusterDeploymentLabels` is not configured at all. E.g. what if the consumer wants to use that CD label, but name the metric label something different? They would end up with a duplicate. If they're using the new config field at all, we're not worried about backward compatibility: they can specify this explicitly.

And at that point I'm not even sure if we need to bother with HIVE-2077.

pkg/controller/metrics/metricsWithDynamicLabels.go
```diff
@@ -0,0 +1,91 @@
...
+	for k, v := range mConfig.AdditionalClusterDeploymentLabels {
+		mapClusterTypeLabelToValue[k] = v
+	}
+	return mapClusterTypeLabelToValue
+}
+
+// GetSortedKeys gets the keys of a map, sorts them and returns the same.
+func GetSortedKeys(mapToSort map[string]string) []string {
+	var keys []string
```
2uasimojo, line 85: nit: using a `make()` with `len(mapToSort)` will make this func marginally more efficient, as it won't have to reallocate `keys` on every iteration of the loop. (The compiler probably optimizes that away anyway, but...)

pkg/controller/metrics/metricsWithDynamicLabels.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+var _ metricsWithDyanamicLabels = CounterWithDynamicLabels{}
+var _ metricsWithDyanamicLabels = HistogramWithDynamicLabels{}
+
+// GetOptionalClusterTypeLabels reads the AdditionalClusterDeploymentLabels from the metrics config and returns the same
+// as a map
+func GetOptionalClusterTypeLabels(mConfig *metricsconfig.MetricsConfig) map[string]string {
+	var mapClusterTypeLabelToValue = make(map[string]string)
```
2uasimojo, line 71: nit: You're not really gaining anything by `make()`ing this map, unless you specify a size.

pkg/controller/metrics/metricsWithDynamicLabels.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type HistogramWithDynamicLabels struct {
+	*prometheus.HistogramVec
+	FixedLabels    map[string]string
+	OptionalLabels map[string]string
+}
+
+// LogMetricWithDynamicLabels observes the histogram metric with val value.
+func (h HistogramWithDynamicLabels) LogMetricWithDynamicLabels(obj metav1.Object, log log.FieldLogger, val float64) {
```
2uasimojo, line 49: Same comments as above.

2uasimojo (commented):
Partial review, will continue tomorrow.

docs/hive-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+Opt into them via `HiveConfig.Spec.MetricsConfig.AdditionalClusterDeploymentLabels`, which accepts a map of the label you would want on your metrics, to the label found on ClusterDeployment.
+For example, including `{"ocp_major_version": "hive.openshift.io/version-major"}` will cause affected metrics to include a label key ocp_major_version with the value from the `hive.openshift.io/version-major` ClusterDeployment label -- e.g. "4".
+
+Note: It is upto the cluster admins to be mindful of cardinality and ensure these labels are not too specific, like cluster id, otherwise it can negatively impact your observability system's performance
+
+### List of all Hive metrics
+
+#### ClusterDeployment controller metrics
+These metrics are logged per ClusterDeployment instance. None of these are optional.
```
2uasimojo, line 40: I _think_ I understand what you intend by "logged per ClusterDeployment instance", but I don't think we want to phrase it that way, as it implies we're labeling per-CD. Perhaps "These metrics are observed while processing ClusterDeployments."?

docs/hive-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+|    hive_cluster_deployment_imageset_job_delay_seconds    |           N            |
+|        hive_cluster_deployment_dns_delay_seconds         |           N            |
+|    hive_cluster_deployment_completed_install_restart     |           Y            |
+|          hive_cluster_deployments_created_total          |           Y            |
+|         hive_cluster_deployments_installed_total         |           Y            |
+|          hive_cluster_deployments_deleted_total          |           Y            |
+| hive_cluster_deployments_provision_failed_terminal_total |           Y            |
+
+#### ClusterProvision controller metrics
+These metrics are logged per ClusterProvision instance. If they have optional label support, ClusterDeployment will apply those labels to ClusterProvision. None of these are optional.
```
2uasimojo, line 55: Similar to above, "These metrics are observed while processing ClusterProvisions."

> ClusterDeployment will apply those labels to ClusterProvision

What does this mean? Do we get the value from the specified label on the CD or the CP? I would have expected the former, as we still have the CD object in that code flow, and we can't/shouldn't rely on the CP having all the same labels as the CD.

docs/hive-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+|                  Metric Name                  | Optional Label Support |
+|:---------------------------------------------:|:----------------------:|
+|     hive_cluster_provision_results_total      |           Y            |
+|              hive_install_errors              |           Y            |
+| hive_cluster_deployment_install_failure_total |           Y            |
+| hive_cluster_deployment_install_success_total |           Y            |
+
+#### ClusterPool controller metrics
+These metrics are logged per ClusterPool instance. None of these are optional.
```
2uasimojo, line 65: Similar, "...while processing ClusterPools."

2uasimojo (commented):
Another chunk of comments. Still not done.

pkg/controller/clusterdeployment/metrics.go
```diff
@@ -83,12 +62,67 @@ func incProvisionFailedTerminal(cd *hivev1.ClusterDeployment) {
 	if stoppedCondition != nil {
 		stoppedReason = stoppedCondition.Reason
 	}
-	metricProvisionFailedTerminal.WithLabelValues(poolNSName,
-		hivemetrics.GetClusterDeploymentType(cd),
-		stoppedReason).Inc()
+	metricProvisionFailedTerminal.FixedLabels["clusterpool_namespacedname"] = poolNSName
+	metricProvisionFailedTerminal.FixedLabels["failure_reason"] = stoppedReason
```
2uasimojo, line 66: oo, this isn't good. The metric var, and its `FixedLabel` member, are global. These assignments can race, and we can end up observing with the wrong label values.

We either need to figure out a way to pass the fixed labels into `LogMetricWithDynamicLabels`, or curry them in advance. The former could be done by adding a `prometheus.Labels` param. If the currying could be done in one line, I would say it's sixes; but if we're still determined to avoid `MustCurryWith` and you want to check errors, let's do the former.

pkg/controller/metrics/metricsWithDynamicLabels.go
```diff
@@ -0,0 +1,91 @@
...
+
+type CounterWithDynamicLabels struct {
+	*prometheus.CounterVec
+	FixedLabels    map[string]string
+	OptionalLabels map[string]string
+}
+
+// LogMetricWithDynamicLabels logs the counter metric. It simply increments the value, so last parameter isn't used.
+func (c CounterWithDynamicLabels) LogMetricWithDynamicLabels(obj metav1.Object, log log.FieldLogger, _ float64) {
+	curriedVec, err := c.CounterVec.CurryWith(c.FixedLabels)
```
2uasimojo, line 27: As noted above, we can't use whatever's in the global `FixedLabels` here -- this func needs to be invoked with the proper label values either already curried into `c` or via a param that we then curry in.

pkg/controller/clusterdeployment/metrics.go
2uasimojo: [Later] Okay, now I see how you were using the map-ness. As noted, that won't work because globals.

I think we actually end up not needing the `FixedLabels` struct member at all.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -54,25 +52,12 @@ var (
...
-	metricInstallErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Name: "hive_install_errors",
-		Help: "Counter incremented every time we observe certain errors strings in install logs.",
-	},
-		[]string{"cluster_type", "reason"},
-	)
+	// mapClusterTypeLabelToValue is a map that corresponds to MetricsWithClusterTypeLabels provided in metrics config.
+	// The key in the map is the name of the label used in defining the metric, and the value would be the cluster
+	// deployment label to look for while reporting the metric
+	mapClusterTypeLabelToValue map[string]string
```
2uasimojo, line 58: similar to above: I don't think this needs to be global

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -82,6 +67,18 @@ func Add(mgr manager.Manager) error {
...
+	mConfig, err := hivemetrics.ReadMetricsConfig()
+	if err != nil {
+		log.WithError(err).Error("error reading metrics config")
+		return err
+	}
+	mapClusterTypeLabelToValue = make(map[string]string)
+	mapClusterTypeLabelToValue = hivemetrics.GetOptionalClusterTypeLabels(mConfig)
+	// Register the metrics. This is done here to ensure we define the metrics with optional label support after we have
+	// read the hiveconfig, and we register them only once.
+	registerMetrics(mapClusterTypeLabelToValue)
```
2uasimojo, line 80: Similar to above: let's refactor this to `registerMetrics(mConfig)`.

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -379,7 +376,8 @@ func (r *ReconcileClusterProvision) reconcileSuccessfulJob(instance *hivev1.Clus
 	pLog.Info("install job succeeded")
 	result, err := r.transitionStage(instance, hivev1.ClusterProvisionStageComplete, "InstallComplete", "Install job has completed successfully", pLog)
 	if err == nil {
-		metricClusterProvisionsTotal.WithLabelValues(hivemetrics.GetClusterDeploymentType(instance), resultSuccess).Inc()
+		metricClusterProvisionsTotal.FixedLabels["result"] = resultSuccess
```
2uasimojo, line 379: As above. (I'll stop noting each one of these now.)

pkg/controller/clusterprovision/clusterprovision_controller_test.go
```diff
@@ -45,6 +45,8 @@ const (
 
 func init() {
 	log.SetLevel(log.DebugLevel)
+	// Register the metrics to avoid panics during testing
+	registerMetrics(mapClusterTypeLabelToValue)
```
2uasimojo, line 49: As above (though I didn't check which test suites require this).

2uasimojo (commented):
Okay, finally done! Sorry it took so long!

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -94,6 +94,8 @@ current-context: admin
 
 func init() {
 	log.SetLevel(log.DebugLevel)
+	// Register the metrics to avoid panics during testing
+	registerMetrics(mapClusterTypeLabelToValue)
```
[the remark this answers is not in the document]
suhanime: The metrics do not need to be registered, but they have to be defined at least - because code never checks whether a metric is registered, it always attempts to observe it, and prometheus libraries handle the rest.
I've changed the wording a bit, and I do [intend to implement the unit testing in a follow-up PR](https://issues.redhat.com/browse/HIVE-2092)

suhanime (commented):
@2uasimojo Ready for review!

pkg/controller/clusterdeployment/metrics.go
2uasimojo: Let me try to clarify:

Since `metricProvisionFailedTerminal` is a global variable, there's only one `metricsProvisionFailedTerminal.FixedLabels` map. Now imagine two threads reaching this code at the same time. (I'll use just the `clusterpool_namespacedname` fixed label to simplify the example.)
- Thread A: `FixedLabels["clusterpool_namespacedname"] = "foons/foopool"`
- Thread B: `FixedLabels["clusterpool_namespacedname"] = "barns/barpool"` <== overwrites the above <== **BAD**
- Thread B: `ObserveMetricWithDynamicLabels()` => observes with label value `"barns/barpool"`
- Thread A: `ObserveMetricWithDynamicLabels()` => *also* observes with label value `"barns/barpool"` <== **BAD**

What I'm suggesting to remedy this problem is one of the following.

### Option 1 ("pass the fixed labels in")
Change `ObserveMetricWithDynamicLabels` as follows:

```go
func (c CounterVecWithDynamicLabels) ObserveMetricWithDynamicLabels(obj metav1.Object, fixedLabels map[string]string, log log.FieldLogger, _ float64) {
	curriedVec, err := c.CounterVec.CurryWith(fixedLabels)
```

and this chunk becomes:

```go
	fixedLabels := map[string]string{
		"clusterpool_namespacedname": poolNSName,
		"failure_reason": stoppedReason,
	}
	metricProvisionFailedTerminal.ObserveMetricWithDynamicLabels(cd, fixedLabels, log, 1)
```

### ~~Option 2: ("curry them in advance")~~
(As I started to write this down, I realized it's Bad™ for a couple of reasons. Will explain below.)

Change this chunk to create a partially curried metric and call `ObserveMetricWithDynamicLabels` on that thing:
```go
	partiallyCurried, err := metricProvisionFailedTerminal.CurryWith(map[string]string{
		"clusterpool_namespacedname": poolNSName,
		"failure_reason": stoppedReason,
	})
	if err != nil {
		panic("!")
	}
	partiallyCurried.ObserveMetricWithDynamicLabels(cd, log, 1)
```

In addition to being more code than the above (since we have to error check both outside and inside the observer func), this doesn't work as written because the partially curried variable isn't of one of our `...WithDynamicLabels` types, so we would have to create one, and then how would we populate the `OptionalLabels`, blah blah messy gross. Or pass the partially curried var into the function, but now we've got two metric variables and have to be careful about which one to use, messy gross.

----

**TL;DR:** Go with Option 1.

That said, we still need to declare the fixed labels so we can register with them. So I suggest changing `FixedLabels` to either `[]string` or `sets.String` -- we only need the *keys* in the metric definition. 

Then if you like, you can add a helper function, called from each `ObserveMetricWithDynamicLabels` impl, that accepts the `fixedLabels` param passed in and validates that the keys therein are some kind of subset of the `FixedLabels` in the metric definition. Though that's probably overkill, since if they're not, `CurryWith` will error. Unless you want to enforce that `fixedLabels` contains *all* of the `FixedLabels`. Or fill in any missing ones with `MetricLabelDefaultValue`.

Does this make sense?

2uasimojo (commented):
Let me know if you want to live chat about the threading concern.

pkg/controller/metrics/metricsWithDynamicLabels.go
2uasimojo: Exactly.

I'm not worried about breaking dashboards. The dashboards, alerts, and HiveConfig are all defined in app-interface, so SD will, in one MR, either configure `cluster_type` explicitly with that exact label; or configure it with some other name and change the dashboards and alerts at the same time (unlikely).

So I'm standing by this comment.

pkg/controller/metrics/metricsWithDynamicLabels.go
```diff
@@ -0,0 +1,91 @@
...
+	// Maps cannot be sorted, however we would want to loop through them in a sorted fashion to ensure all the metric
+	// labels are added in an expected order. Get the slice of map keys first, sort them and then use it to loop over
+	// the map
+	for _, v := range GetSortedKeys(c.OptionalLabels) {
+		curriedVec, err = curriedVec.CurryWith(prometheus.Labels{v: GetLabelValue(obj, c.OptionalLabels[v])})
+	}
+	if err != nil {
+		// The possibility of hitting an error here is low, but report it and don't observe the metric in case it happens
+		log.WithError(err).Error("error observing metric")
+	} else {
```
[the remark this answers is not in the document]
2uasimojo: Ah, fair ✓

pkg/controller/clusterdeployment/metrics.go
suhanime: That's a valid concern. I liked Option 1 as well and have implemented it to the best of my abilities, however many of such metrics still have empty fixedLabels and this can still happen. As a failsafe, one of the arguments to the Observe functions is v1 object - be that cluster deployment or provision instance. Is it safe to assume we shouldn't run into this problem now?

apis/hive/v1/metricsconfig/metricsConfig.go
suhanime: I missed this in one of the previous changes - added

pkg/controller/metrics/metricsWithDynamicLabels.go
suhanime: Okay, I've changed it accordingly

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+		// The possibility of hitting an error here is low, but report it and don't observe the metric in case it happens
+		log.WithField("metric", c.CounterVec).WithError(err).Error("error observing metric")
+	} else {
+		curriedVec.WithLabelValues().Inc()
+	}
+}
+
+// verifyAllLabelsPresent ensures all fixed labels for a metric are present. If missing, it sets the corresponding label
+// to a default value
+// Note: Avoid relying on this method, as it cannot ensure the labels would get curried in the expected order
```
suhanime, line 55: I don't like adding this note - but currying accepts labels as a map which cannot be sorted in an order.  At least doing this ensures we always have all the labels (which is necessary imo - if the first ever observe is with less labels - those labels will be dropped in all the rest of the observes).
Also I ensured all of the current metric observes do not need this verification.

suhanime (commented):
@2uasimojo Ready for review!

PS: I squashed some of the commits, but I did push my changes before the squash - for ease of review

2uasimojo:
/test e2e

...while I review this

2uasimojo (commented):
This is very, very close.

I _was_ gonna say we can land it as is and address my comments in a followon... but the pointer-to-the-map thing would actually affect the API, so we have to get it right out of the gate.

apis/hive/v1/metricsconfig/metrics_config.go
```diff
@@ -0,0 +1,20 @@
...
+	// hive will label it with the specified metric key, and copy the value from the specified ClusterDeployment label.
+	// For example, including {"ocp_major_version": "hive.openshift.io/version-major"} will cause affected metrics to
+	// include a label key ocp_major_version with the value from the hive.openshift.io/version-major ClusterDeployment
+	// label -- e.g. "4".
+	// NOTE: Avoid ClusterDeployment labels whose values are unbounded, such as those representing cluster names or IDs,
+	// as these will cause your prometheus database to grow indefinitely.
+	// Affected metrics are those whose type implements the metricsWithDynamicLabels interface found in
+	// pkg/controller/metrics/metrics_with_dynamic_labels.go
+	// +optional
+	AdditionalClusterDeploymentLabels map[string]string `json:"additionalClusterDeploymentLabels,omitempty"`
```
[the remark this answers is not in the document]
suhanime: Done

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+type CounterVecWithDynamicLabels struct {
+	*prometheus.CounterVec
+	*DynamicLabels
+}
+
+// ObserveMetricWithDynamicLabels logs the counter metric. It simply increments the value, so last parameter isn't used.
+func (c CounterVecWithDynamicLabels) ObserveMetricWithDynamicLabels(obj metav1.Object, log log.FieldLogger,
+	fixedLabels map[string]string, _ float64) {
+	fixedLabels = verifyAllLabelsPresent(fixedLabels, c.FixedLabels)
+	curriedVec, err := c.CounterVec.CurryWith(fixedLabels)
```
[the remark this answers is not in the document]
suhanime: Implemented - and now we no longer need the `CurryWith` method, or needing to catch the error for the same.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+	// Maps cannot be sorted, however we would want to loop through them in a sorted fashion to ensure all the metric
+	// labels are added in an expected order. Get the slice of map keys first, sort them and then use it to loop over
+	// the map
+	labels := make(prometheus.Labels, len(c.OptionalLabels))
+	for _, v := range GetSortedKeys(c.OptionalLabels) {
```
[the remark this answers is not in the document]
suhanime: Yepp - changed to a `getKeys` function without sorting.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+// Note: Avoid relying on this method, as it cannot ensure the labels would get curried in the expected order
+func verifyAllLabelsPresent(labels map[string]string, fixedLabels []string) map[string]string {
+	if len(labels) < len(fixedLabels) {
+		for _, v := range fixedLabels {
+			if _, ok := labels[v]; !ok {
+				labels[v] = constants.MetricLabelDefaultValue
+			}
+		}
+	}
+	return labels
```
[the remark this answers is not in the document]
suhanime: Ah yes - correct

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+	} else {
+		curriedVec.WithLabelValues().Inc()
+	}
+}
+
+// verifyAllLabelsPresent ensures all fixed labels for a metric are present. If missing, it sets the corresponding label
+// to a default value
+// Note: Avoid relying on this method, as it cannot ensure the labels would get curried in the expected order
+func verifyAllLabelsPresent(labels map[string]string, fixedLabels []string) map[string]string {
+	if len(labels) < len(fixedLabels) {
```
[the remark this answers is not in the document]
suhanime: I think, since we are removing the overlaps before getting here - and making the map to the correct length, we're good.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+		log.WithField("metric", c.CounterVec).WithError(err).Error("error observing metric")
+	} else {
+		curriedVec.WithLabelValues().Inc()
+	}
+}
+
+// verifyAllLabelsPresent ensures all fixed labels for a metric are present. If missing, it sets the corresponding label
+// to a default value
+// Note: Avoid relying on this method, as it cannot ensure the labels would get curried in the expected order
+func verifyAllLabelsPresent(labels map[string]string, fixedLabels []string) map[string]string {
```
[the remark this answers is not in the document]
suhanime: renamed but kept the method just cause. Doesn't make much difference either ways

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,120 @@
...
+
+type DynamicLabels struct {
+	// FixedLabels just need to be a slice used for defining the metric.
+	FixedLabels []string
+	// OptionalLabels will be used both for defining the metric and by the Observe method
+	OptionalLabels map[string]string
+}
+
+type CounterVecWithDynamicLabels struct {
+	*prometheus.CounterVec
```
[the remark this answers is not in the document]
suhanime: done!

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func WithOptionalLabels(labels map[string]string) func(*dynamicLabels) {
+	return func(d *dynamicLabels) {
+		d.optionalLabels = labels
+	}
+}
+
+// getLabelList combines fixed and optional labels and returns the list without duplicates.
```
suhanime, line 42: While doing this, I thought about how it would be easier to have an early feedback - if a user tries to cause an overlap between these labels. However, there is not simpler way to do it, and we wouldn't want to error out because of metrics.

suhanime (commented):
@2uasimojo Ready for review!

pkg/controller/metrics/metrics.go
```diff
@@ -743,13 +743,12 @@ func (ca *clusterAccumulator) setMetrics(total, installed, uninstalled, deprovis
 	}
 }
 
-// GetClusterDeploymentType returns the value of the hive.openshift.io/cluster-type label if set,
-// otherwise a default value.
-func GetClusterDeploymentType(obj metav1.Object) string {
-	if typeStr, ok := obj.GetLabels()[hivev1.HiveClusterTypeLabel]; ok {
+// GetLabelValue returns the value of the label if set, otherwise a default value.
+func GetLabelValue(obj metav1.Object, label string) string {
+	if typeStr, ok := obj.GetLabels()[label]; ok {
```
2uasimojo, line 748: Is it okay if the cd label's value is empty? If not:
```suggestion
	if typeStr := obj.GetLabels()[label]; typeStr != "" {
```
Note that this will treat "no such label" the same as "label exists and has empty string as its value".

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	for label, value := range optionalLabels {
+		labels[label] = value
+	}
+	return labels
+}
+
+// defaultMissingLabels ensures a value is present for all corresponding labels of a metric. If missing, it sets it to a
+// default value
+func defaultMissingLabels(labelKeys []string, valueMap map[string]string) map[string]string {
```
2uasimojo, line 76: This function doesn't make a lot of sense as it's being used right now. You're passing in the full label map (fixed + optional) but GetLabelValue() has already defaulted as necessary. IMHO it makes more sense just to do the defaulting within the part of `buildLabels` that deals with the fixed labels.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+// duplicates. The resultant map can be used for observing the metric.
+func (d *dynamicLabels) buildLabels(fixedLabels map[string]string, obj metav1.Object) prometheus.Labels {
+	optionalLabels := make(map[string]string, len(d.optionalLabels))
+	for _, v := range getKeys(d.optionalLabels) {
+		optionalLabels[v] = GetLabelValue(obj, d.optionalLabels[v])
+	}
+	labels := make(map[string]string, len(d.getLabelList()))
+	for label, value := range fixedLabels {
+		labels[label] = value
+	}
```
2uasimojo, line 67: As noted below, we could get rid of `defaultMissingLabels` if we changed this to:
```suggestion
	for _, label := range d.fixedLabels {
		if value := fixedLabels[label]; value != "" {
			labels[label] = value
		} else {
			labels[label] = constants.MetricLabelDefaultValue
		}
	}
```

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+var _ metricsWithDynamicLabels = CounterVecWithDynamicLabels{}
+var _ metricsWithDynamicLabels = HistogramVecWithDynamicLabels{}
+
+// GetOptionalClusterTypeLabels reads the AdditionalClusterDeploymentLabels from the metrics config and returns the same
+// as a map
+func GetOptionalClusterTypeLabels(mConfig *metricsconfig.MetricsConfig) map[string]string {
+	var mapClusterTypeLabelToValue = make(map[string]string)
+	// Currently "cluster_type" is a non-optional label for these metrics, so if the config is empty (default), set the
+	// "cluster_type" label
+	// TODO: remove as a part of https://issues.redhat.com/browse/HIVE-2077
```
2uasimojo, line 159: I think we need to keep the defaulting forever, as we can't guarantee all consumers have cut over to using hiveconfig. Let's just get rid of this comment.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+// RegisterMetricWithDynamicLabels registers the CounterVec metric.
+func (c CounterVecWithDynamicLabels) RegisterMetricWithDynamicLabels() {
+	metrics.Registry.MustRegister(c.metric)
+}
+
+// ObserveMetricWithDynamicLabels logs the counter metric. It simply increments the value, so last parameter isn't used.
+func (c CounterVecWithDynamicLabels) ObserveMetricWithDynamicLabels(obj metav1.Object, log log.FieldLogger,
+	fixedLabels map[string]string, _ float64) {
+	labels := defaultMissingLabels(c.getLabelList(), c.buildLabels(fixedLabels, obj))
+	c.metric.With(labels).Inc()
```
2uasimojo, line 115: If you take suggestions above, this can just be
```suggestion
	c.metric.With(c.buildLabels(fixedLabels, obj)).Inc()
```
Similar below.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/hive/apis/hive/v1/metricsconfig"
+	"github.com/openshift/hive/pkg/constants"
+)
+
+// metricsWithDynamicLabels encompasses metrics that can have optional labels.
+type metricsWithDynamicLabels interface {
+	ObserveMetricWithDynamicLabels(metav1.Object, log.FieldLogger, map[string]string, float64)
+	RegisterMetricWithDynamicLabels()
```
2uasimojo, line 18: nit: it seems redundant, and takes up extra space, to include `MetricWithDynamicLabels` in these func names, considering the structs already have that in their names. (Along similar lines, I think there's actually a go linting rule that complains if you have a package named `foo` containing objects/methods/etc. named `FooWhatever`.)

2uasimojo (commented):
This is a lovely iteration. I think we're really getting somewhere with the design of the objects/methods.

Most of my comments suggest refactoring, consolidation, simplification, cleanup. However, the couple of things still holding this from merging are:
- We need to decide whether we *ever* allow the empty string as a label value. You have previously asserted that that somehow risks defining a metric with too few labels, and we agreed to always default.
- I like the idea of panicking if the user tries to configure an optional label with the same key as a fixed one.

See inline for details.

Thanks!

pkg/controller/metrics/metrics.go
suhanime: Right, so I do not remember when I discovered a problem - I distinctly remember us debating what a default value should be (like N/A or unknown). Nevertheless, I decided to track down the actual problem in the library because just 1 person remembering something like this is not ideal for future-proofing.
Okay, so `With` or `WithLabelValues` internally calls `GetMetricWith` from vec.go, and the problem comes when `getOrCreateMetricWithLabels` [creates](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/vec.go#L514) the map with size equivalent to `labels` passed instead of reading from `metric.Desc` to determine the correct size (Note - even the `extractLabelValues` only relies on labels [provided](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/vec.go#L616))

TL;DR
When the metric is created, it relies on the labels provided instead of metric definition as the dimension for creating the map. So if you provide empty string as label, it ignores it and creates the resultant metric map with less dimensions, consequently leading to always dropping some labels while observing.
Maybe `With` might not lead to such a problem - `WithLabelValues` definitely does. But if we know that happens, we don't want to treat them differently as a safety.

suhanime (commented):
@2uasimojo Ready for review

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,177 @@
...
+
+func NewCounterVecWithDynamicLabels(counterOpts *prometheus.CounterOpts,
+	labelOptions ...func(labels *dynamicLabels)) *CounterVecWithDynamicLabels {
+	counterVecMetric := &CounterVecWithDynamicLabels{
+		CounterOpts:   counterOpts,
+		dynamicLabels: &dynamicLabels{},
+	}
+	for _, o := range labelOptions {
+		o(counterVecMetric.dynamicLabels)
+	}
```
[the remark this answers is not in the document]
2uasimojo: Okay, I see. I hadn't been thinking about the options pattern. Now that I've looked at it in that light, I get the advantages. In particular, if we wanted to add fields to our fancy structure, we would have to add arguments to the constructor and then change all the calls to it. In this case, short term loss (more, and more complicated, code) but long term gain (easier to enhance in the future). If you want to put it back in a fup, I'm down. But for now this is good.

pkg/controller/metrics/metrics.go
2uasimojo: Okay; so then my comment still applies. If the label _exists_ but happens to have an empty value, this code as written will cause it to be observed with an empty value rather than our default.

Now, based on the code you pointed to, it seems to me like that would be okay, because we would actually be passing in the empty string as an argument, as opposed to omitting the argument entirely, so the number of dimensions would still be correct. And if that's true, I would argue it's the appropriate thing to do: i.e. there's a difference in behavior for the user between the CR lacking the label ("unspecified") and the label existing with an empty value (the value _is_ specified, and it's `""`).

Whichever way we decide to do it, it should be consistent. Right now the fixed labels always convert `""` to the default, but the dynamic ones don't.

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,155 @@
...
+	optionalLabels map[string]string, log log.FieldLogger) *CounterVecWithDynamicLabels {
+	counterVecMetric := &CounterVecWithDynamicLabels{
+		CounterOpts: counterOpts,
+		dynamicLabels: &dynamicLabels{
+			fixedLabels:    fixedLabels,
+			optionalLabels: optionalLabels,
+		},
+	}
+	if repeatedLabel := counterVecMetric.getRepeatedLabel(); repeatedLabel != "" {
+		log.Errorf("HiveConfig.Spec.AdditionalClusterDeploymentLabels[%q] conflicts with a fixed label for the metric %s. Please rename your label.", repeatedLabel, counterOpts.Name)
```
2uasimojo, line 90: We agreed to make this fatal, right?. (And if you turn this into a `panic`, you don't need the `log` arg.)

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -0,0 +1,155 @@
...
+func (d *dynamicLabels) getLabelList() []string {
+	var finalList []string
+	finalList = append(finalList, d.fixedLabels...)
+	for label := range d.optionalLabels {
+		finalList = append(finalList, label)
+	}
+	return finalList
+}
+
+// getRepeatedLabel detects and returns the first optional label it encounters, that overlaps with another label
```
2uasimojo, line 42: nit: This would be more useful if it returned _all_ repeated labels so the consumer doesn't have to iterate. Can be done in a fup though.

2uasimojo:
Beautiful!

/lgtm

PR #1927, opened 2022-12-02
Pause ClusterDeployment reconciliation

2uasimojo:
Add an annotation, `hive.openshift.io/reconcile-pause`. While set to a truthy value, (most) controllers dealing with ClusterDeployments will (mostly) no-op.

Note that some controllers reconcile on other objects, and update CDs as a result. Some of these things will still happen. However, the updates will not be acted upon by the relevant (other) controller(s) until the annotation is removed.

For example, the clusterpool controller will still create, delete, and update PowerState on ClusterDeployments it owns. But the respective clusters won't be provisioned, deprovisioned, or hibernated/activated because the clusterdeployment, hibernation, etc. controllers are still short circuiting due to the annotation.

HIVE-2064

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -134,6 +134,24 @@ const (
 	// the cluster, causing its machines to remain in their current state (unless acted on externally) regardless of CD.Spec.PowerState.
 	PowerStatePauseAnnotation = "hive.openshift.io/powerstate-pause"
 
+	// ReconcilePauseAnnotation is an annotation used by ClusterDeployment. If "true", (most) controllers will (mostly) ignore the CD until
+	// the annotation is cleared/non-truthy. Exceptions:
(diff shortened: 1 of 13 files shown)


2uasimojo:
I tested this live for a couple of use cases, including the installAttemptsLimit one. Seems to work as designed (and @suhanime the logs stay quiet while the annotation is in place).

Suhani & rest of the hive team (@abutcher @dlom @newtonheath), how do we feel about merging this as written? It's an annotation, not part of the API, so we can claim it's unofficial...

abutcher:
@2uasimojo Looks good to me, I remembered that [we have similar pause annotations for SyncSets & ClusterDeployments for hibernation ](https://github.com/openshift/hive/blob/206ec604183f09e34b2836aad335cc1a66234355/pkg/constants/constants.go#L130-L135) so I think this isn't too far from those implementations / use cases.

/lgtm

PR #1931, opened 2022-12-16
HIVE-2063: Add support for specifying additional AllowedPrincipal ARNs for a Private Link cluster's VPC Endpoint Service.

abutcher:
Add support for specifying `AdditonalAllowedPrincipals` (ARNs) to be configured for a Private Link cluster's VPC Endpoint Service in addition to the IAM entity used by Hive.

New field `cd.Spec.Platform.AWS.PrivateLink.AdditionalAllowedPrincipals` is made a mutable field so that Allowed Principal ARNs can be added and removed.

`AdditionalAllowedPrincipals` and the `DefaultAllowedPrincipal` will be recorded in Status within `cd.Status.Platform.AWS.PrivateLink.VPCEndpointService`. The `AdditionalAllowedPrincipals` Status entry is used to determine if we [shouldSync()](https://github.com/openshift/hive/blob/33e68f241d3c784c5362b48febea14020107dd39/pkg/controller/awsprivatelink/awsprivatelink_controller.go#L314-L355) before the normal 2hr sync that occurs in the awsprivatelink controller if the `AdditionalAllowedPrincipals` in Spec do not match those recorded in Status.

Latent: Fix typo s/desrired/desired/

[HIVE-2063](https://issues.redhat.com//browse/HIVE-2063)


pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
```diff
@@ -990,6 +990,63 @@ users:
...
+
+			m.EXPECT().GetCallerIdentity(gomock.Any()).Return(&sts.GetCallerIdentityOutput{Arn: aws.String("aws:iam:12345:hub-user")}, nil)
+			m.EXPECT().DescribeVpcEndpointServicePermissions(gomock.Any()).
+				Return(&ec2.DescribeVpcEndpointServicePermissionsOutput{
+					AllowedPrincipals: []*ec2.AllowedPrincipal{{
+						Principal: aws.String("aws:iam:12345:some-that-should-not-be-allowed"),
+					}},
+				}, nil)
+			m.EXPECT().ModifyVpcEndpointServicePermissions(&ec2.ModifyVpcEndpointServicePermissionsInput{
+				AddAllowedPrincipals:    aws.StringSlice([]string{"aws:iam:12345:another-user", "aws:iam:12345:hub-user", "aws:iam:12345:some-user"}),
```
2uasimojo, line 1042: Where is this slice getting sorted?

2uasimojo (commented):
Seems simple enough. I'm having a bit of trouble with the seeming interchangeability of the terms "Perm(ission)", "Principal", and "ARN"; but I guess someone more familiar with this area wouldn't even notice?

pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
abutcher: Good question.

patjlm:
> /cc @patjlm this is the list of ARNs that we would like Fleet Manager to append to as we discussed earlier

Cc @tiwillia

pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
abutcher: Possibly a side effect of using `sets`.

2uasimojo: Confirmed in [doc](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Byte.List) and [code](https://github.com/openshift/hive/blob/93ac39c9c1815fa71750775d3fed4cdea3ecd3dd/vendor/k8s.io/apimachinery/pkg/util/sets/string.go#L163-L175), a side effect of `List()`.

abutcher:
The VPC Endpoint Service is only modified every 2 hours based on checking a [shouldSync](https://github.com/openshift/hive/blob/33e68f241d3c784c5362b48febea14020107dd39/pkg/controller/awsprivatelink/awsprivatelink_controller.go#L314-L355) function. `shouldSync` will need to become aware of a change in the `AllowedPrincipals` to signal that an update should be made outside of the normal 2 hour sync and I'm thinking we'll need a field within `Status` to accomplish that. :construction: :construction_worker_man:

2uasimojo:
> The VPC Endpoint Service is only modified every 2 hours based on checking a [shouldSync](https://github.com/openshift/hive/blob/33e68f241d3c784c5362b48febea14020107dd39/pkg/controller/awsprivatelink/awsprivatelink_controller.go#L314-L355) function. `shouldSync` will need to become aware of a change in the `AllowedPrincipals` to signal that an update should be made outside of the normal 2 hour sync and I'm thinking we'll need a field within `Status` to accomplish that. construction construction_worker_man

Yeah, good idea: why don't we make a status field with the (full) list of applied ARNs. Didn't make a lot of sense before, but now it'll be useful two ways.

abutcher:
> Yeah, good idea: why don't we make a status field with the (full) list of applied ARNs. Didn't make a lot of sense before, but now it'll be useful two ways.

This is going to be tricky because I don't know the default ARN that gets automatically added to the VPCE Service when checking if we should sync in `shouldSync()` and would need to make an API call to determine the ARN which would eliminate the purpose of the `shouldSync()` function as a means of avoiding rate limiting since we would now be making an API call every reconcile.

I would need to know the default ARN to determine if the combined list of configured `allowedPrincipals` in spec plus the default were in status. I could check if the configured `allowedPrincipals` are in present status instead of checking if desired==current but then removals from `allowedPrincipals` wouldn't be applied until the 2 hour sync.

We'll also have to update status for every private link cluster in order to add the default ARN to the status which I assume we should do for all private link clusters and not just those that configure `allowedPrincipals`.

abutcher:
Added a commit that stores the `DefaultAllowedPrincipal` in Status. The `DefaultAllowedPrincipal` being absent from Status will not trigger a sync outside of the normal 2hr sync and will be set during normal reconcile if absent.

@2uasimojo ready for another round of review

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -695,19 +713,31 @@ func (r *ReconcileAWSPrivateLink) reconcileVPCEndpointService(awsClient *awsClie
...
-	desriredPerms := sets.NewString(aws.StringValue(stsResp.Arn))
+	// desiredPerms is the set of Allowed Principals that will be configured for the cluster's VPC Endpoint Service.
+	// desiredPerms only contains the IAM entity used by Hive (defaultARN) by default but may contain additional Allowed Principal
+	// ARNs as configured within cd.Spec.Platform.AWS.PrivateLink.AdditionalAllowedPrincipals.
+	defaultARN := aws.StringValue(stsResp.Arn)
+	desiredPerms := sets.NewString(defaultARN)
+	if allowedPrincipals := cd.Spec.Platform.AWS.PrivateLink.AdditionalAllowedPrincipals; allowedPrincipals != nil {
+		for _, allowedPrincipal := range *allowedPrincipals {
+			desiredPerms.Insert(allowedPrincipal)
+		}
```
2uasimojo, line 724: `Insert` is variadic, so you should be able to do this without the loop.
```suggestion
		desiredPerms.Insert(*allowedPrincipals...)
```

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -695,19 +713,31 @@ func (r *ReconcileAWSPrivateLink) reconcileVPCEndpointService(awsClient *awsClie
...
 			ServiceId: serviceConfig.ServiceId,
 		}
-		if added := desriredPerms.Difference(oldPerms).List(); len(added) > 0 {
+		if added := desiredPerms.Difference(oldPerms).List(); len(added) > 0 {
 			input.AddAllowedPrincipals = aws.StringSlice(added)
 		}
-		if removed := oldPerms.Difference(desriredPerms).List(); len(removed) > 0 {
+		if removed := oldPerms.Difference(desiredPerms).List(); len(removed) > 0 {
 			input.RemoveAllowedPrincipals = aws.StringSlice(removed)
 		}
```
2uasimojo, line 737: nit: you could save a negligible amount of processing/memory by waiting until inside the conditions to do `.List()`.

2uasimojo (commented):
Yah, this is looking good. A couple of minor comments inline. Assume there will be UT added before we want to make this go?

abutcher:
@2uasimojo Comments addressed, thanks! The UT changes were squished in the diff view for me but there are few tests for the ensuring that the ARNs are added to / removed from the VPC Endpoint Service as matched by the expected mock calls and a few tests to ensure that `shouldSync()` fires appropriately. Lmk if there are any tests we should add as well.

2uasimojo:
> The UT changes were squished in the diff view for me

D'oh, that was the problem. They are good, everything's good, let's do this.

/lgtm

PR #1934, opened 2023-01-10
Add status.powerState check to e2e_test.sh

lleshchi:
Test that the status.powerState matches the expected powerState both while the cluster is Running and Hibernating. This test should fail due to issue: https://issues.redhat.com/browse/HIVE-2073


lleshchi:
e2e expected test failure due to https://issues.redhat.com/browse/HIVE-2073

PR #1936, opened 2023-01-12
updates adoption steps for privatelink clusters/added configurations

tonytheleg:
- Updates adoption info to address Privatelink clusters for ROSA as well as any extra configs that may be in use
- addresses limitations of using hiveutil as it does not support privatelink setting or any extra configurations required (which we set in FedRAMP)

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -1203,11 +1203,7 @@ If the cluster you are looking to adopt is on AWS and leverages Privatelink, you
       region: us-east-1
 ```      
 
-It's also important to include any other configurations that may be in place in the original clusters' ClusterDeployment, such as `certificateBundles` settings or `controlPlaneConfig` settings as these are leveraged by certman-operator if in use. It's a good idea to compare your adoption manifest to your existing ClusterDeployment manifest. Any new secret refs added to your adoption manifest will require correlating secrets to live on the adopting cluster in the correct namespace.
-
 ### Adopting with hiveutil
-
-_NOTE: If you have any customizations outside of the sample manfiest (Privatelink, certificate configurations), the `hiveutil` method may not encompass all those._
 
 [hiveutil](hiveutil.md) is a development focused CLI tool which can be built from the hive repo. To adopt a cluster specify the following flags:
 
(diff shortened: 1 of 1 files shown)


docs/using-hive.md
```diff
@@ -1191,8 +1191,24 @@ spec:
...
+  platform:
+    aws:
+      credentialsSecretRef:
+        name: my-aws-cluster-creds
+      privateLink:
+        enabled: true
+      region: us-east-1
+```      
+
+It's also important to include any other configurations that may be in place in the original clusters' ClusterDeployment, such as `certificateBundles` settings or `controlPlaneConfig` settings as these are leveraged by certman-operator if in use. It's a good idea to compare your adoption manifest to your existing ClusterDeployment manifest. Any new secret refs added to your adoption manifest will require correlating secrets to live on the adopting cluster in the correct namespace.
```
2uasimojo, line 1206:
```suggestion
It's also important to include any other configurations that may be in place in the original cluster's ClusterDeployment, such as `certificateBundles` settings or `controlPlaneConfig` settings as these are leveraged by certman-operator if in use. It's a good idea to compare your adoption manifest to your existing ClusterDeployment manifest. Any new secret refs added to your adoption manifest will require correlating secrets to live on the adopting cluster in the correct namespace.
```

Hmm, couple things bugging me here:
- certman-operator is an OSD thing, and wouldn't be known to anyone outside of SRE-P (even OSD customers won't usually be aware of it).
- A discussion of adoption shouldn't really talk about an "existing ClusterDeployment manifest". If you have one of those, you should be using relocation instead of adoption.

Can we find a way to phrase all of this so it doesn't reference either of those things?

tonytheleg: yeah good points, as I was writing it I was worried this was too focused on OSD, and not considering OCP users who may be leveraging Hive. I could see that confusing non-OSD users. I'll work on cleaning that up a bit. Our scenario was a more unique case, and most probably won't be in that scenario.

tonytheleg: updated to only make note about privatelink as the other settings are not really covered in other places, and probably hyper specific to OSD. Keeping it simple.

2uasimojo (commented):
Sane. Thanks @tonytheleg!

@abutcher anything to add?

abutcher:
@2uasimojo :+1: looks good to me!
/lgtm

PR #1937, opened 2023-01-12
Relocate: Ignore kube-controller-manager .crt ConfigMaps

### LINKED ISSUE #1932: Hive Cluster Relocation is stuck on kube-root-ca.crt and openshift-service-ca.crt already exists
Hive Cluster Relocation is stuck on kube-root-ca.crt and openshift-service-ca.crt already exists . These configmaps are created by default when namespace created and hive clusterdeployment fails with error 

```
    - lastProbeTime: '2022-12-20T11:03:48Z'
      lastTransitionTime: '2022-12-20T10:08:51Z'
      message: >-
        failed to copy *v1.ConfigMapList: could not copy *v1.ConfigMap resource
        "kube-root-ca.crt": failed to sync existing resource: configmaps
        "kube-root-ca.crt" already exists
      reason: MoveFailed
      status: 'True'
      type: RelocationFailed
```

Is there a way to ignore these configmaps ? We are stuck in a migration , looking forward to any help !

2uasimojo:
Starting in k8s 1.21, the kube-controller-manager (and presumably OpenShift's fork thereof) creates/reconciles `kube-root-ca.crt` and `openshift-service-ca.crt` ConfigMaps in every namespace. On relocate, as we're copying objects, we'll try to copy these from the source to the destination cluster, fail because the object (with the same name) already exists, try to delete it, and then race with the kcm trying to create it again. If the relocate controller loses the race, it'll fail the relocate and try again. It might succeed eventually if it wins the race, but in at least one case (see associated issue/card) it never won.

With this commit, we ignore these two ConfigMaps, based on their names (hardcoded).

[HIVE-2080](https://issues.redhat.com//browse/HIVE-2080)

closes #1932

### DIFF
--- a/pkg/controller/clusterrelocate/clusterrelocate_controller.go
+++ b/pkg/controller/clusterrelocate/clusterrelocate_controller.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


abutcher (approved):
Looks good to me, `sets`
/lgtm

PR #1939, opened 2023-01-13
docs: clarify how to develop using podman

jstuever:
Update the create-kind-cluster script to work with podman in addition to docker. Also update the developing doc to include the steps for developing using podman.


docs/developing.md
```diff
@@ -125,7 +132,7 @@ You can leave your registry container running indefinitely. The kind cluster can
 
 ```bash
 kind delete cluster --name hive
-./hack/create-kind-cluster.sh hive
+./hack/create-kind-podman-cluster.sh hive
```
2uasimojo, line 135: similarly here, perhaps
```suggestion
# podman
./hack/create-kind-podman-cluster.sh hive
# OR docker
./hack/create-kind-cluster.sh hive
```

Do we need to reset the KUBECONFIG variable after?

docs/developing.md
```diff
@@ -80,27 +80,34 @@ A Kubernetes or OpenShift cluster is required to run Hive from source. If you wo
...
 
 If you encounter ImagePullErrors, and your kind container cannot reach your registry container, you may be experiencing problems with Fedora (at least 32, possibly earlier as well) and Docker. You can attempt to work around this by changing the FirewallBackend in the /etc/firewalld/firewalld.conf file from nftables to iptables and restarting docker. (see [this comment](https://github.com/kubernetes-sigs/kind/issues/1547#issuecomment-623756313)
 
 
 Create a local insecure registry (if one does not already exist) and then a kind cluster named 'hive' to deploy to. You can create additional clusters if desired by providing a different name argument to the script.
 
+Podman:
+```bash
+./hack/create-kind-podman-cluster.sh hive
+export KUBECONFIG=~/.kube/hive.kubeconfig
```
2uasimojo, line 93: Is setting `KUBECONFIG` necessary/advisable/possible for both tools? (If so, perhaps we can consolidate these samples.)

2uasimojo (commented):
I can only take it on faith that this works :P

docs/developing.md
2uasimojo: I do think that's a good idea. (I've never used these scripts before.)

Even better would be a single script that detects whether you've got podman or docker (preferring one if you have both -- which? probably podman) and just Does The Right Thing™. I understand if that's too much scope creep for this PR though :)

2uasimojo:
/test e2e

Looks like the DPTP AWS account is running up against quota...

2uasimojo:
> Looks like the DPTP AWS account is running up against quota...

Oh, that's not DPTP's account -- it's ours. We shouldn't be running out of load balancers. Investigating...

2uasimojo:
Found and cleaned up a bunch of stale resources. Also tried (again) to tweak our lambda bots so we hopefully start getting notifications again.

Now maybe I'll actually review this PR :)

hack/create-kind-cluster.sh
```diff
@@ -11,27 +11,75 @@ set -o errexit
...
+
+# Fail when neither podman nor docker are found.
+if [ "${PODMAN}" == "" -a "${DOCKER}" == "" ]; then
+  echo "Unable to find podman or docker."
+  exit 1;
+fi
+
+# Prefer podman and fallthrough to docker.
+container_cmd="${PODMAN:-${DOCKER}}"
+echo $"Using ${container_cmd}"
```
2uasimojo, line 26: I'm not familiar with this syntax (the `$` in front of the string). What gives?

hack/create-kind-cluster.sh
```diff
@@ -11,27 +11,75 @@ set -o errexit
...
+
+# Podman containers should default to the podman network.
+if [ "${container_cmd}" == "${PODMAN}" ]; then
+  reg_network="${reg_network:-podman}"
+fi
+
+# Use the network parameter if a network is configured.
+if [ "${reg_network}" != "" ]; then
+  reg_network_param="--network ${reg_network}"
+fi
```
2uasimojo, line 40: nit: looks like we could do without the `reg_network` var.
```suggestion
reg_network_param=''

# Podman containers should default to the podman network.
if [ "${container_cmd}" == "${PODMAN}" ]; then
  reg_network_param="--network ${reg_network}"
fi
```

2uasimojo (commented):
I'm taking it on faith that this works; my review is pretty much limited to shell syntax :)

/lgtm
/hold in case you wish to fix the nit noted inline.

hack/create-kind-cluster.sh
2uasimojo: In that case, shouldn't L30 be removed?

2uasimojo (commented):
/lgtm

sorry for the delay :ice_cube:

PR #1940, opened 2023-01-17
Remove deprecated io/ioutil usage

mjlshen:
Just noticed this while submitting an unrelated PR


2uasimojo:
Can confirm the whole ioutil library seems to be deprecated since go 1.16: https://pkg.go.dev/io/ioutil

Conversions look correct.

Thanks @mjlshen !

/lgtm

PR #1941, opened 2023-01-17
OSD-14201 Add new installer failure reason: DefaultEbsKmsKeyInsufficientPermissions

mjlshen:
This failure reason isn't very common, but currently it is being caught as `FallbackInstancesFailedToBecomeReady`, which doesn't really provide any actionable hints when troubleshooting the failed install. The AWS docs describing this failure case can be found here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/troubleshooting-launch.html#troubleshooting-launch-internal

The relevant log lines have `Client.InternalError: Client error on launch` and this means that the default EBS encryption key does not have enough permissions to allow the installer to run EC2 instances and decrypt EBS volumes.

Relates to https://issues.redhat.com/browse/OSD-14201

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -87,6 +87,11 @@ data:
         - "Error: multiple Route53Zone found"
       installFailingReason: MultipleRoute53ZonesFound
       installFailingMessage: Multiple Route53 zones found
+    - name: DefaultEbsKmsKeyInsufficientPermissions
+      searchRegexStrings:
+        - "Client.InternalError: Client error on launch"
+      installFailingReason: DefaultEbsKmsKeyInsufficientPermissions
+      installFailingMessage: Default KMS key for EBS encryption has insufficient permissions to launch EC2 instances
     - name: SimulatorThrottling
       searchRegexStrings:
       - "validate AWS credentials: checking install permissions: error simulating policy: Throttling: Rate exceeded"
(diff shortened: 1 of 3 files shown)


2uasimojo:
It feels like a stretch to assume that something as generic as "Client error on launch" always indicates the very niche "default EBS KMS key has insufficient permissions" condition. Can we maybe change the reason code to be more generic, like ClientErrorOnLaunch, and have the message suggest that the EBS thing _might_ be at fault?

mjlshen:
Oh it is a specific error based on AWS' docs: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/troubleshooting-launch.html#troubleshooting-launch-internal - if you search for `Client.InternalError: Client error on launch` you'll see it.

2uasimojo:
Right right, I read the doc; but the tone there is closer to what I'd like to see: If you see $error, check for $thing. The way this is currently written, it makes it sound like $error is always exclusively the result of $thing.

mjlshen:
Gotcha, anecdotally it has always been something wrong with the default EBS KMS key - but the sample size is admittedly just 2 haha.

I will put this on hold to do more research to make sure this is true.
/hold

mjlshen:
/unhold

I'm back :D `Client.InternalError: Client error on launch` does point to the default EBS KMS Key in all cases I'm aware of. I was able to experiment with Autoscaling groups after reading this one too: https://docs.amazonaws.cn/en_us/autoscaling/ec2/userguide/key-policy-requirements-EBS-encryption.html

PR #1946, opened 2023-01-26
Add Gp3VolumeLimitExceeded install failure reason

mjlshen:
In [OSD-14978](https://issues.redhat.com//browse/OSD-14978) we learned about a new way installs can fail.

> Why `Gp3VolumeLimitExceeded` and not `VolumeLimitExceeded`?

I would like to be able to give the customer the exact quota to increase, in this case: "L-7A658B76"

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -159,6 +159,11 @@ data:
       - "VcpuLimitExceeded"
       installFailingReason: VcpuLimitExceeded
       installFailingMessage: The install requires more vCPU capacity than your current vCPU limit
+    - name: Gp3VolumeLimitExceeded
+      searchRegexStrings:
+      - "VolumeLimitExceeded: You have exceeded your maximum gp3 storage limit"
+      installFailingReason: Gp3VolumeLimitExceeded
+      installFailingMessage: "The installation failed due to insufficient gp3 storage quota in the region (QuotaCode L-7A658B76)"
     - name: UserInitiatedShutdown
       searchRegexStrings:
       - "Error waiting for instance .* to become ready .* User initiated shutdown"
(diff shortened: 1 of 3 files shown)


config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -159,6 +159,11 @@ data:
       - "VcpuLimitExceeded"
       installFailingReason: VcpuLimitExceeded
       installFailingMessage: The install requires more vCPU capacity than your current vCPU limit
+    - name: Gp3VolumeLimitExceeded
+      searchRegexStrings:
+      - "VolumeLimitExceeded: You have exceeded your maximum gp3 storage limit"
+      installFailingReason: Gp3VolumeLimitExceeded
+      installFailingMessage: "The installation failed due to insufficient gp3 storage quota in the region (L-7A658B76)"
```
2uasimojo, line 166: Is `L-7A658B76` like a specific error code or something? Reference?

[Later] I still couldn't find this code anywhere, but based on [this article](https://aws.amazon.com/premiumsupport/knowledge-center/request-service-quota-increase-cli/) I surmise it's a "QuotaCode". Sure would feel better about this if there was an AWS document that showed it.

mjlshen: Yeah it's one of the quota codes needed when a customer goes to request a quota increase :( Similarly can't find a public doc with that code on it...

2uasimojo: And that code isn't specific to a particular region or anything like that?

Tell you what: Since we have to retest e2e anyway, how about we change this to 
```suggestion
      installFailingMessage: "The installation failed due to insufficient gp3 storage quota in the region (QuotaCode L-7A658B76)"
```

...to give the reader some idea of what that inscrutable string means?

mjlshen: Yeah I like that idea, I'll make sure that clarity gets bubbled up through OCM too

2uasimojo (commented):
/lgtm

PR #1947, opened 2023-01-26
Watch for ClusterSyncLease and renew if it is missing

dlom:
now with anti-hotlooping technology!

xref: [HIVE-2038](https://issues.redhat.com//browse/HIVE-2038)

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -223,6 +223,20 @@ func AddToManager(mgr manager.Manager, r *ReconcileClusterSync, concurrentReconc
 		return err
 	}
 
+	// Watch for changes to ClusterSync. These have the same name/namespace as the relevant
+	// ClusterDeployment, so when a ClusterSync watch triggers, the CD of the same name will be reconciled.
+	// When the CD reconciles, it will look up the related ClusterSync.
+	if err := c.Watch(&source.Kind{Type: &hiveintv1alpha1.ClusterSync{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
+
+	// Watch for changes to ClusterSyncLease. These have the same name/namespace as the relevant
+	// ClusterDeployment, so when a ClusterSyncLease watch triggers, the CD of the same name will be reconciled.
+	// When the CD reconciles, it will look up the related ClusterSyncLease.
+	if err := c.Watch(&source.Kind{Type: &hiveintv1alpha1.ClusterSyncLease{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -407,7 +421,6 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 
-	needToCreateClusterSync := false
 	clusterSync := &hiveintv1alpha1.ClusterSync{}
 	switch err := r.Get(context.Background(), request.NamespacedName, clusterSync); {
 	case apierrors.IsNotFound(err):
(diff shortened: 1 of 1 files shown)


pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -525,13 +547,23 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 		}
 	}
 
-	result := reconcile.Result{Requeue: true, RequeueAfter: r.timeUntilFullReapply(lease)}
-	if syncSetsNeedRequeue || selectorSyncSetsNeedRequeue {
+	result := reconcile.Result{Requeue: true, RequeueAfter: r.timeUntilRenew(lease)}
+	if syncSetsNeedRequeue == requeueWithDelay || selectorSyncSetsNeedRequeue == requeueWithDelay {
+		result.RequeueAfter = time.Second * time.Duration(10)
```
2uasimojo, line 552: Requeueing every 10s is still a hot loop. Where did the 10s number come from?

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -417,13 +430,17 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 		ownerRef := metav1.NewControllerRef(cd, cd.GroupVersionKind())
 		ownerRef.Controller = nil
 		clusterSync.OwnerReferences = []metav1.OwnerReference{*ownerRef}
-		if err := r.Create(context.Background(), clusterSync); err != nil {
+		switch err := r.Create(context.Background(), clusterSync); {
+		case apierrors.IsAlreadyExists(err):
+			// race condition, just proceed
```
2uasimojo, line 435: This would be strange, as we _shouldn't_ ever have multiple threads trying to reconcile the same cluster. I think we should at least log a warning if we get here.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -525,13 +547,23 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 		}
 	}
 
-	result := reconcile.Result{Requeue: true, RequeueAfter: r.timeUntilFullReapply(lease)}
-	if syncSetsNeedRequeue || selectorSyncSetsNeedRequeue {
+	result := reconcile.Result{Requeue: true, RequeueAfter: r.timeUntilRenew(lease)}
+	if syncSetsNeedRequeue == requeueWithDelay || selectorSyncSetsNeedRequeue == requeueWithDelay {
+		result.RequeueAfter = time.Second * time.Duration(10)
+	} else if syncSetsNeedRequeue == requeueImmediately || selectorSyncSetsNeedRequeue == requeueImmediately {
```
2uasimojo, line 553: I wonder if these conditions should be reversed -- i.e. if either needs to be requeued immediately, we should do that rather than delaying.

(That said, I'm still not convinced delaying is The Right Thing™.)

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -626,7 +658,7 @@ func (r *ReconcileClusterSync) applySyncSets(
 			newSyncStatus.FailureMessage = err.Error()
 		}
 		if syncSetNeedsRequeue {
-			requeue = true
+			requeue = requeueWithDelay
```
2uasimojo, line 661: So IIUC you're delaying if apply/create/update fails, but requeueing immediately if a delete fails? Can you explain the rationale?

I think if we're going to be picking and choosing, it should be based on the nature of the error... which can be hard to get right. For example, if we could detect that the syncset itself is bogus, we should arguably mark it as such and not even retry. If the failure is that the target cluster is unreachable, we should retry, probably with a delay, but what should that delay be? What about a creds failure? Perhaps we don't requeue, making the user fix their creds and force resync by deleting the lease (or waiting until the next 2h interval).

2uasimojo (commented):
I'd like to Meet™ and discuss this some more.

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -453,10 +470,15 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
...
-		logger.Info("need to reapply all syncsets")
+	needToRenew := r.timeUntilRenew(lease) <= 0
+
+	if needToCreateLease {
+		logger.Info("need to reapply all syncsets (missing lease)")
+	} else if needToRenew {
+		logger.Info("need to reapply all syncsets (time to renew)")
 	}
+
+	needToDoFullReapply := needToCreateLease || needToRenew
```
2uasimojo, line 481: Say, shouldn't this also include `|| needToCreateClusterSync`? (Which would need to be set above in the `IsNotFound` branch after Get(clusterSync) -- which was probably the original intent of the `needToCreateClusterSync` var.)

dlom: If you check the original code, nothing was ever setting it to true

2uasimojo: > If you check the original code, nothing was ever setting it to true

I know -- it should have :) See

> (Which would need to be set above in the `IsNotFound` branch after Get(clusterSync) -- which was probably the original intent of the `needToCreateClusterSync` var.)

[Later] I walked through the code and figured out how deleting the clustersync is triggering a full reapply: it's because we're using the clustersync's status -- which is empty since it just got created -- to figure out which syncsets we need to apply. So whereas visually it would be arguably more readable if we had that folded in explicitly, it would indeed be redundant. Probably worth a comment though.

2uasimojo (commented):
This looks good. I was going to complain about the lack of unit tests... but we really aren't set up to test Watch()es in UT.

Adding some syncset testing in e2e would be goodness. But that can be a separate card.

/lgtm
/hold pending @abutcher review

PR #1948, opened 2023-01-26
AWS.PrivateLink is preserved with PreserveOnDelete

jstuever:
Previously, deleting a ClusterDeployment with an AWS PrivateLink configured would always result in the deletion of the PrivateLink. This change ensures the PrivateLink is preserved when PreserveOnDelete is enabled.

https://issues.redhat.com/browse/HIVE-2087

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -741,6 +742,7 @@ users:
 		configureAWSClient func(*mock.MockClient)
 
 		hasFinalizer        bool
+		expectDeleted       bool
 		expectedAnnotations map[string]string
 		expectedStatus      *hivev1aws.PrivateLinkAccessStatus
 		expectedConditions  []hivev1.ClusterDeploymentCondition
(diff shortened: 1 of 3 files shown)


2uasimojo (commented):
It was surprisingly complicated to track down all the code paths that this affects. (Nothing to do with your code change, which is simple and elegant.) Would you please add some test cases to TestReconcile for them?

- PL enabled => disabled: *should* fully clean up, even with PoD
- CD deleted: should remove finalizer but leave the rest
- **?** During install, a provision fails; next time through we should clean up from the failed attempt regardless of PoD, right? AFAICS with PoD enabled this will "leak" resources from failed provision attempts, and I don't think that's what we want. Can you confirm?

2uasimojo:
@jstuever I'm waiting for this to go off WIP before re-reviewing, right?

2uasimojo:
> * PL enabled => disabled: _should_ fully clean up, even with PoD

As you pointed out, this is forbidden via webhook. (It's theoretically possible for the user to disable the webhook and make this happen... but they're breaking glass and I see no need to sweat supporting that.)

> * CD deleted: should remove finalizer but leave the rest

As discussed, we should be able to add UT for this, as long as a similar test _without_ PoD legitimately blows up because the cleanup stuff isn't mocked out.

Reviewing...

2uasimojo (commented):
Okay, this looks good. I've convinced myself that the path through the first cleanupRequired can't actually happen at all, and the second looks correct, except...

There's a race condition where PL is enabled, PoD is on, we fail the setup the first time through, and then the CD is deleted before we get to run again and clean up from that failed attempt.

I need to look at this again with tomorrow brain. But I suspect we'll want that deletion flow UT to prove it.

2uasimojo (commented):
/lgtm

Some nonblocking comments inline.

/hold

in case you want to address them here.

2uasimojo (commented):
/lgtm

2uasimojo:
/retest

Really gonna need to beef up that clusterpool...

2uasimojo:
Been seeing this kind of error quite a bit lately, smells like an eventual consistency problem. Not related to the PR. 

/test e2e

2uasimojo:
/test e2e-pool

which ironically timed out due to insufficient pool capacity

which should hopefully be less of a problem after hours

PR #1949, opened 2023-01-26
StringPtr deprecated in favor of String

mjlshen:
Noticed that `pointer.StringPtr` has been deprecated for `pointer.String` as well: https://pkg.go.dev/k8s.io/utils/pointer#StringPtr so just a bit of a cleanup PR while doing #1946


2uasimojo (commented):
Sure, thanks.

/lgtm

PR #1954, opened 2023-02-06
Improve PowerState status for Deprovisioned ClusterDeployments

lleshchi:
xref: https://issues.redhat.com/browse/HIVE-2134

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -526,6 +526,9 @@ const (
 	// HibernatingReasonPowerStatePaused indicates that we can't/won't discover the state of the
 	// cluster's cloud machines because the powerstate-paused annotation is set.
 	HibernatingReasonPowerStatePaused = "PowerStatePaused"
+	// HibernatingReasonClusterDeploymentDeleted indicates that a Cluster Deployment has been deleted
+	// and that the cluster is deprovisioning unless preserveOnDelete is set to true.
+	HibernatingReasonClusterDeploymentDeleted = "ClusterDeploymentDeleted"
 
 	// ReadyReasonStoppingOrHibernating is used as the reason for the Ready condition when the cluster
 	// is stopping or hibernating. Precise details are available in the Hibernating condition.
@@ -549,6 +552,9 @@ const (
 	// ReadyReasonPowerStatePaused indicates that we can't/won't discover the state of the
 	// cluster's cloud machines because the powerstate-paused annotation is set.
 	ReadyReasonPowerStatePaused = "PowerStatePaused"
+	// ReadyReasonClusterDeploymentDeleted indicates that a Cluster Deployment has been deleted
+	// and that the cluster is deprovisioning unless preserveOnDelete is set to true.
+	ReadyReasonClusterDeploymentDeleted = "ClusterDeploymentDeleted"
 )
 
 // Provisioned status condition reasons
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -178,7 +178,8 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 
 	// If cluster is already deleted, skip any processing
 	if !cd.DeletionTimestamp.IsZero() {
-		return reconcile.Result{}, nil
+		msg := "ClusterDeployment has been marked for deletion"
+		return r.setStatusStatesUnknown(cd, hivev1.HibernatingReasonClusterDeploymentDeleted, hivev1.ReadyReasonClusterDeploymentDeleted, msg, cdLog)
 	}
 
 	// Initialize cluster deployment conditions if not present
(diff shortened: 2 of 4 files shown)


apis/hive/v1/clusterdeployment_types.go
```diff
@@ -526,6 +526,9 @@ const (
 	// HibernatingReasonPowerStatePaused indicates that we can't/won't discover the state of the
 	// cluster's cloud machines because the powerstate-paused annotation is set.
 	HibernatingReasonPowerStatePaused = "PowerStatePaused"
+	// HibernatingReasonClusterDeploymentDeleted indicates that a Cluster Deployment has been deleted
+	// and that the cluster is deprovisioning unless preserveOnDelete is set to true.
+	HibernatingReasonClusterDeploymentDeleted = "ClusterDeploymentDeleted"
```
suhanime, line 531: I think for this particular code path - like `Unknown` state - it doesn't differ for Hibernating and Running conditions so I don't see a point of declaring different reasons for them...

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -178,7 +178,11 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 
 	// If cluster is already deleted, skip any processing
 	if !cd.DeletionTimestamp.IsZero() {
-		return reconcile.Result{}, nil
+		cd.Status.PowerState = hivev1.ClusterPowerStateUnknown
+		msg := "ClusterDeployment has been marked for deletion"
+		r.setCDCondition(cd, hivev1.ClusterHibernatingCondition, hivev1.HibernatingReasonClusterDeploymentDeleted, msg, corev1.ConditionUnknown, cdLog)
+		r.setCDCondition(cd, hivev1.ClusterReadyCondition, hivev1.ReadyReasonClusterDeploymentDeleted, msg, corev1.ConditionUnknown, cdLog)
+		return reconcile.Result{}, r.updateClusterDeploymentStatus(cd, cdLog)
```
suhanime, line 185: You should add a check here to see if an update is needed.
If a cluster is already marked for deletion and we have updated the status on CD in a previous reconcile, we do not need to call an `Update` again.

2uasimojo, line 185: We need to be careful not to issue the Update unless we've actually changed something. No-op updates can end up thrashing with other controllers. `setCDCondition` returns a bool indicating whether it changed anything; and you can manually check the status.powerState. Perform `updateClusterDeploymentStatus` iff any of those things are true, but not if they're all false.

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -75,13 +76,42 @@ func TestReconcile(t *testing.T) {
 		expectRequeueAfter     time.Duration
 	}{
 		{
-			name: "cluster deleted",
-			cd:   cdBuilder.GenericOptions(testgeneric.Deleted()).Options(o.shouldHibernate).Build(),
+			name: "cluster deleted with finalizer",
```
2uasimojo, line 79: tests look good ✓

pkg/controller/hibernation/hibernation_controller_test.go
```diff
@@ -75,13 +76,42 @@ func TestReconcile(t *testing.T) {
...
+			cd:   cdBuilder.GenericOptions(testgeneric.Deleted(), testgeneric.WithFinalizer(finalizer)).Options(o.shouldHibernate).Build(),
 			cs:   csBuilder.Build(),
 			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
 				cond, runCond := getHibernatingAndRunningConditions(cd)
-				assert.Nil(t, cond, "not expecting hibernating condition")
-				assert.Nil(t, runCond, "not expecting running condition")
+				require.NotNil(t, cond)
+				require.NotNil(t, runCond)
+				assert.Equal(t, corev1.ConditionUnknown, cond.Status)
```
suhanime, line 86: You can also do
```
if assert.NotNil(t, cond, "<A helpful debug msg>"){
    assert.Equal(t, <status>, cond.Status, "<A helpful debug msg>"}
```
There's also [helper functions](https://github.com/openshift/hive/blob/master/pkg/test/assert/assertions.go#L68:L108) if they make the checks simpler

apis/hive/v1/clusterdeployment_types.go
2uasimojo: Same could be said for `PowerStatePaused`.

suhanime: Ah yes - I think during the initial refactor, we decided to have different API definitions for different conditions, but we can rethink that when we refactor this controller again...

pkg/controller/hibernation/hibernation_controller.go
lleshchi: @2uasimojo Do you think it would be fair to pull this section out into a separate function, as the behavior is near identical to https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller.go#L202:L216

apis/hive/v1/clusterdeployment_types.go
lleshchi: Good point. I noticed that `"PowerStatePaused"` also differentiated between `ReadyReasonPowerStatePaused ` and `HibernatingReasonPowerStatePaused` . It does seem like a helpful way to confirm that the reason being used is applicable for the condition it's used for, but perhaps it's redundant in these cases.

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: Yup, I like.

apis/hive/v1/clusterdeployment_types.go
2uasimojo: Sometimes we'll typedef a specific string (examples near here) but in this case we haven't. I recall from playing with this that there isn't a great way to DRY it completely. I'm not gonna lose sleep over it.

lleshchi: I'm going to keep them as is for now, and either create a separate commit to remove duplicates or wait until the next refactoring. Does that sound ok?

2uasimojo (commented):
/lgtm

nice work!

PR #1956, opened 2023-02-07
Update hack/requirements.txt

2uasimojo:
...to gel with what I actually needed to `pip install` to make bundle-gen.py work in a venv.

### DIFF
--- a/hack/requirements.txt
+++ b/hack/requirements.txt
@@ -1,3 +1,3 @@
 GitPython
 PyYAML>=6.0
-yq>=4.0
+requests


dlom:
/lgtm

I must have had these all on my system already, because I haven't been using a venv when using bundle-gen to create the update PRs

2uasimojo:
> /lgtm
> 
> I must have had these all on my system already, because I haven't been using a venv when using bundle-gen to create the update PRs

Nor I... until I f'ed up my python installation (still not sure how I did that). Now I get SSL errors any time I try to do _anything_. Hence the venv.

PR #1957, opened 2023-02-08
Enable pprof for hive-clustersync-n (and every other controller)

dlom:
xref: [HIVE-2139](https://issues.redhat.com//browse/HIVE-2139)

Note: This is not a fix, this is merely to enable diagnostics so we can collect for information

/assign @2uasimojo 
/cc @abutcher


2uasimojo:
The idea is to let this roll on stage for a while and revert it before next promotion.

PR #1966, opened 2023-03-01
HIVE-2090: Add ResourcePool field to hive's MachinePool CRD

dlom:
xref: [HIVE-2090](https://issues.redhat.com//browse/HIVE-2090)

This will allow a user of hive to supply their own resource pool as part of the spec

### DIFF
--- a/apis/hive/v1/vsphere/machinepools.go
+++ b/apis/hive/v1/vsphere/machinepools.go
@@ -3,6 +3,11 @@ package vsphere
 // MachinePool stores the configuration for a machine pool installed
 // on vSphere.
 type MachinePool struct {
+	// ResourcePool is the name of the resource pool that will be used for virtual machines.
+	// If it is not present, a default value will be used.
+	// +optional
+	ResourcePool string `json:"resourcePool,omitempty"`
+
 	// NumCPUs is the total number of virtual processor cores to assign a vm.
 	NumCPUs int32 `json:"cpus"`
 
--- a/config/crds/hive.openshift.io_machinepools.yaml
+++ b/config/crds/hive.openshift.io_machinepools.yaml
@@ -511,6 +511,11 @@ spec:
                         required:
                         - diskSizeGB
                         type: object
+                      resourcePool:
+                        description: ResourcePool is the name of the resource pool
+                          that will be used for virtual machines. If it is not present,
+                          a default value will be used.
+                        type: string
                     required:
                     - coresPerSocket
                     - cpus
--- a/hack/app-sre/saas-template.yaml
+++ b/hack/app-sre/saas-template.yaml
@@ -5314,6 +5314,11 @@ objects:
                           required:
                           - diskSizeGB
                           type: object
+                        resourcePool:
+                          description: ResourcePool is the name of the resource pool
+                            that will be used for virtual machines. If it is not present,
+                            a default value will be used.
+                          type: string
                       required:
                       - coresPerSocket
                       - cpus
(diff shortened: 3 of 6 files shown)


dlom:
/unhold

I've managed to get a cluster installed in vCenter that has a custom resource pool specified.  Additionally, I've managed to create MachineSets on day 2 through hive, and they were properly assigned to the custom resource pool specified in the hive ClusterDeployment representing the cluster

2uasimojo (commented):
I need to have a closer look (which will have to be next week) but I believe we need to account for using the day 0 custom resource pool for day 2 machinesets _without_ the user needing to specify that value via the MachinePool. Otherwise I think resources can end up leaking.

2uasimojo:
Clarified in [thread](https://redhat-internal.slack.com/archives/CH06KMDRV/p1696880902020089) that the RP is not created or destroyed by installer code; and that we can indeed have arbitrary/different RPs day 2 vs day 0. So this code is the proper solution :+1: 

/lgtm

apis/hive/v1/vsphere/platform.go
```diff
@@ -27,6 +27,10 @@ type Platform struct {
 	// virtual machines.
 	Folder string `json:"folder,omitempty"`
 
+	// ResourcePool is the name of the resource pool that will be used and/or created
+	// for virtual machines.
+	ResourcePool string `json:"resourcePool,omitempty"`
```
2uasimojo, line 32: @jcpowermac points out that the upstream version of this field is `+optional`. That may not be significant since it's a non-pointer type, but might as well make it match.

Sounds like the comment could use a tweak too: nix `and/or created` since installer doesn't (can't) create RPs.

2uasimojo: Also as [discussed](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1696881945939289), we want the new field to be in MachinePool, not ClusterDeployment. Day 0, the user (optionally) specifies a RP via install-config; day 2 they (optionally) specify a (possibly different) RP via MachinePool.

2uasimojo:
> @dlom: This pull request references [HIVE-2090](https://issues.redhat.com//browse/HIVE-2090) which is a valid jira issue.
> 
> Warning: The referenced jira issue has an invalid target version for the target branch this PR targets: expected the bug to target the "4.15.0" version, but no target version was set.
> 
> In response to [this](https://github.com/openshift/hive/pull/1966):
> 
> > xref: [HIVE-2090](https://issues.redhat.com//browse/HIVE-2090)
> > This will allow a user of hive to supply their own resource pool as part of the spec
> 
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.

We need to get this switched off for hive. @bradmwilliams is that in your wheelhouse?

2uasimojo:
/lgtm
/hold

@dlom unhold at will when done testing

dlom:
Manual verification of functionality was a success

/unhold

2uasimojo:
Beauty. Don’t forget to backport to mce-2.4. @chenz4027 does this need to go back further?

bradmwilliams:
@2uasimojo that warning is in our wheelhouse.  It was added as part of: https://issues.redhat.com/browse/FDN-272.  Hive was specifically mentioned as a "repo we'd restrict", but right now we don't discriminate.  I'll update that ticket and request guidance for these restrictions.

PR #1968, opened 2023-03-06
Added hiveconfig.awsPrivateLink.SharedHostedZoneDomains

jstuever:
Added a new field to enable sharing of AWS dns zones for aws privatelink dns records. It is a list of domain names for private hosted zones that are to be shared by multiple clusters' privatelink records. They must exist in the same account as the Hive cluster and be associated with the privatelink VPC. When a cluster's domain is a subdomain of any item in the list, the privatelink record will be created in the zone of the item that most closely matches. Otherwise a new private hosted zone will be created for the cluster.

[HIVE-2062](https://issues.redhat.com//browse/HIVE-2062)

### DIFF
--- a/apis/hive/v1/hiveconfig_types.go
+++ b/apis/hive/v1/hiveconfig_types.go
@@ -192,6 +192,17 @@ type AWSPrivateLinkConfig struct {
 	// +kubebuilder:default=Alias
 	// +optional
 	DNSRecordType AWSPrivateLinkDNSRecordType `json:"dnsRecordType,omitempty"`
+
+	// SharedHostedZoneDomains is a list of domain names for private hosted
+	// zones that are to be shared by multiple clusters' privatelink
+	// records. They must exist in the same account as the Hive cluster and
+	// be associated with the privatelink VPC. When a cluster's domain is a
+	// subdomain of any item in the list, the privatelink record will be
+	// created in the zone of the item that most closely matches. Otherwise
+	// a new private hosted zone will be created for the cluster.
+	//
+	// +optional
+	SharedHostedZoneDomains []string `json:"sharedHostedZoneDomains,omitempty"`
 }
 
 // AWSPrivateLinkDNSRecordType defines what type of DNS record should be created in Private Hosted Zone
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -939,7 +939,10 @@ func (r *ReconcileAWSPrivateLink) reconcileHostedZone(awsClient *awsClient,
 	cd *hivev1.ClusterDeployment, metadata *hivev1.ClusterMetadata,
 	vpcEndpoint *ec2.VpcEndpoint, apiDomain string,
 	logger log.FieldLogger) (bool, string, error) {
-	modified, hostedZoneID, err := r.ensureHostedZone(awsClient.hub, cd, vpcEndpoint, apiDomain, logger)
+
+	// Select from SharedHostedZoneDomains; default to apiDomain.
+	hostedZoneDomain := selectDomain(apiDomain, r.controllerconfig.SharedHostedZoneDomains)
+	modified, hostedZoneID, err := r.ensureHostedZone(awsClient.hub, cd, vpcEndpoint, hostedZoneDomain, logger)
 	if err != nil {
 		logger.WithError(err).Error("error ensuring Hosted Zone was created")
 		return modified, "", err
@@ -968,6 +971,28 @@ func (r *ReconcileAWSPrivateLink) reconcileHostedZone(awsClient *awsClient,
 		return modified, "", err
 	}
 	return modified, hostedZoneID, nil
+}
+
+// selectDomain determines if the specified domain is a subdomain of any item
+// in the list and returns the most accurately matching one. Default to the
+// subdomain when no matches are found.
+func selectDomain(subdomain string, domains []string) string {
+	// The subdomain is empty, no results found.
+	if subdomain == "" {
+		return subdomain
+	}
+	bestMatch := ""
+	for _, domain := range domains {
+		// Always select the longest possible domain name that matches.
+		if len(domain) > len(bestMatch) && strings.HasSuffix(subdomain, "."+domain) {
+			bestMatch = domain
+		}
+	}
+	// No matches found; default to the subdomain
+	if bestMatch == "" {
+		return subdomain
+	}
+	return bestMatch
 }
 
 func (r *ReconcileAWSPrivateLink) recordSet(awsClient awsclient.Client, apiDomain string, vpcEndpoint *ec2.VpcEndpoint) (*route53.ResourceRecordSet, error) {
--- a/pkg/controller/awsprivatelink/cleanup.go
+++ b/pkg/controller/awsprivatelink/cleanup.go
@@ -2,6 +2,7 @@ package awsprivatelink
 
 import (
 	"context"
+	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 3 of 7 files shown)


pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -970,6 +973,28 @@ func (r *ReconcileAWSPrivateLink) reconcileHostedZone(awsClient *awsClient,
...
+// selectDomain determines if the specified domain is a subdomain of any item
+// in the list and returns the most accurately matching one. Default to the
+// subdomain when no matches are found.
+func selectDomain(subdomain string, domains []string) string {
+	// The subdomain is empty, no results found.
+	if subdomain == "" {
+		return subdomain
+	}
+	bestMatch := ""
+	for _, domain := range domains {
```
2uasimojo, line 985: I have to verify this in go playground every time: no nil check necessary here ✓

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -171,6 +176,25 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
...
+	if awsErrCodeEquals(err, "NoSuchHostedZone") {
+		return nil // no more work
+	}
+	if err != nil {
+		hzLog.WithError(err).Error("failed to get the hosted zone")
+		return err
+	}
+	for _, domain := range r.controllerconfig.SharedHostedZoneDomains {
+		if strings.EqualFold(domain, strings.TrimSuffix(aws.StringValue(zoneResp.HostedZone.Name), ".")) {
+			isSharedHostedZone = true
```
2uasimojo, line 194: nit: could `break` here

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -201,6 +234,11 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
 		}
 	}
 
+	// SharedHostedZones should not be deleted; exit here.
```
2uasimojo, line 237: > should not be deleted

...even if they're empty? I guess I can dig that.

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -136,9 +137,13 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
 		hzID = cd.Status.Platform.AWS.PrivateLink.HostedZoneID
 	}
 
+	// Figure out the API address for cluster. The error handling is only
+	// needed when the hzID is empty because then we will not be able to
+	// discover the hzID and should exit.
+	apiDomain, err := initialURL(r.Client,
+		client.ObjectKey{Namespace: cd.Namespace, Name: metadata.AdminKubeconfigSecretRef.Name})
```
2uasimojo, line 144: Consider DRYing this with the [other instance](https://github.com/openshift/hive/blob/4dfe9ecb15e1db492177e12ab1d5385b98092850/pkg/controller/awsprivatelink/awsprivatelink_controller.go#L540-L542) . (I found it kind of goofy that we were using the kubeconfig to discover the API URL; if that changes at some point we want to catch all paths at once.)

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -171,6 +176,25 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
 	}
 
 	hzLog := logger.WithField("hostedZoneID", hzID)
+
+	// Determine if hzID is one of SharedHostedZoneDomains
+	var isSharedHostedZone bool
+	zoneResp, err := awsClient.GetHostedZone(&route53.GetHostedZoneInput{
```
2uasimojo, line 182: It's kind of a pity that `findHostedZone` on L168 already gets the information we need (the Name), which would save us this call if we happened to go through that path. IDK if it's worth refactoring `findHostedZone` to return the name as well as the ID, then we can conditionally do this whole chunk only if we need it.

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -171,6 +176,25 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
...
+	}
+	if err != nil {
+		hzLog.WithError(err).Error("failed to get the hosted zone")
+		return err
+	}
+	for _, domain := range r.controllerconfig.SharedHostedZoneDomains {
+		if strings.EqualFold(domain, strings.TrimSuffix(aws.StringValue(zoneResp.HostedZone.Name), ".")) {
+			isSharedHostedZone = true
+		}
+	}
```
2uasimojo, line 196: This could probably reuse `selectDomain()`, something like:

```go
isSharedHostedZone := selectDomain(apiDomain, r.controllerconfig.SharedHostedZoneDomains) == zoneResp.HostedZone.Name
```

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -181,11 +205,20 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
 		hzLog.WithError(err).Error("failed to list the hosted zone")
 		return err
 	}
+	if isSharedHostedZone && apiDomain == "" {
+		// unable to determine specific record(s) to delete from SharedHostedZones.
+		return nil
+	}
```
2uasimojo, line 211: Could this be moved up, saving the call to `ListResourceRecordSets`?

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -2,6 +2,7 @@ package awsprivatelink
 
 import (
 	"context"
+	"strings"
```
2uasimojo, line 5: Definitely want to add UT for the changes to this file.

2uasimojo (commented):
In addition to notes inline:

In the card, we said we would add a counter metric that we increment every time we create a PHZ. This is so the consumer can alert if the `rate` is ever nonzero after the initial run, which would indicate a bug somewhere. Otherwise that bug doesn't get discovered until they hit quota again.

pkg/controller/awsprivatelink/cleanup.go
2uasimojo: I was literally talking about just this one line and the fact that it's feeding `metadata.AdminKubeconfigSecretRef.Name`, which is the goofy part. But meh.

apis/hive/v1/hiveconfig_types.go
```diff
@@ -192,6 +192,17 @@ type AWSPrivateLinkConfig struct {
 	// +kubebuilder:default=Alias
 	// +optional
 	DNSRecordType AWSPrivateLinkDNSRecordType `json:"dnsRecordType,omitempty"`
+
+	// SharedHostedZoneDomains is a list of domain names for private hosted
+	// zones that are to be shared by multiple clusters' privatelink
+	// records. They must exist in the same account as the Hive cluster and
+	// be associated with the privatelink VPC. When a cluster's domain is a
```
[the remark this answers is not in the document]
2uasimojo: Right. But the wording of this docstring implies that it's the user's responsibility to ensure that the PHZs (or the domain names?)

> exist in the same account as the Hive cluster and [are] associated with the privatelink VPC.

Whereas in fact the controller is doing the creating and associating.

Maybe it would help me to understand what the user has to prepare in their account in order to be able to add a domain to this list.

pkg/controller/awsprivatelink/cleanup.go
2uasimojo: ack

PR #1971, opened 2023-03-07
Set ulimit for buildah image build.

jstuever:
Potential fix for "too many open files"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,7 +298,7 @@ docker-dev-push: build image-hive-fedora-dev docker-push
 # Build the dev image using builah
 .PHONY: buildah-dev-build
 buildah-dev-build:
-	buildah bud -f Dockerfile --tag ${IMG}
+	buildah bud --ulimit nofile=10239:10240 -f ./Dockerfile --tag ${IMG}
 
 # Build and push the dev image with buildah
 .PHONY: buildah-dev-push
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -162,7 +162,7 @@ def build_and_push_image(registry_auth_file, image_repo, image_tag, dry_run, bui
     if build_engine == "buildah":
         build = dict(
             query="buildah images -nq {}",
-            build="buildah bud --tag {} -f ./Dockerfile",
+            build="buildah bud --ulimit nofile=10239:10240 --tag {} -f ./Dockerfile",
             push="buildah push --authfile={} {}",
         )
         registry_auth_arg = registry_auth_file


Makefile
```diff
@@ -284,7 +284,7 @@ $(addprefix generate-submodules-,$(GO_SUB_MODULES)):
 .PHONY: docker-build
```
2uasimojo, line 284: I wonder if we should also set `IMAGE_BUILD_EXTRA_FLAGS` to include this ulimit thing, for folks using the `image-hive` target. [ref](https://github.com/openshift/hive/blob/cfcd32cb8ae7a077ff8cd4244b3f0a4bc7dce4ab/vendor/github.com/openshift/build-machinery-go/make/targets/openshift/images.mk#L22)

(NB: I'm not sure whether that would actually work via imagebuilder -- would need to be tested.)

2uasimojo: Should be easy enough to make it additive, no? But even if not, I've confirmed there is nothing downstream of our invocation. (And if someone adds something later, it would be a bug on them not to make _theirs_ additive.)

2uasimojo:
/lgtm
/hold

Unhold when satisfied with local testing.

2uasimojo:
Sorry it took me like a full day to get back to this.

The [thing](https://github.com/openshift/hive/compare/db54345bc2792e966c80c911008390fdbadf3ca7..0f1f9902b8e27e6c813b7089b7ddfa16cecc1d1e) with `IMAGE_BUILD_EXTRA_FLAGS` wasn't working with `?=` because the include from build-machinery-go is [setting](https://github.com/openshift/build-machinery-go/blob/d321ffa04533d88d2fdca5bcf2d5555fdd02a0a3/make/targets/openshift/images.mk#L9) the variable to the empty string; and `?=` only [works](https://www.gnu.org/software/make/manual/html_node/Setting.html) if the variable is undefined.

So we have a couple of options that will make this work as intended:
- We can use `?=` but put the assignment _above_ the imports.
- We can leave the assignment in the same place, but use `=` instead of `?=`. That'll fully override whatever ends up getting done in the imported .mk.
- More in the spirit of how we discussed doing this in the first place, we can make it additive by using `+=`. That way if the imported .mk ever ends up putting default stuff in there, we'll preserve it along with our added stuff.

As you say, though, it may make sense to only bother with this for buildah, since it doesn't seem to have been a problem for docker...

...but now I'm confused. How does `image-hive` relate to `docker-build` and `buildah-dev-build`? It uses `imagebuilder`, which on the surface _looks_ like it's supposed to be agnostic (or more likely internally aware) of podman vs docker-ness. Do you (buildah user) run into the ulimit problem when running `make image-hive` or not?

And somewhat related: if we need to do something about this, we also need to do it in bundle-gen.py, which currently uses a `buildah` or `docker` command explicitly. Perhaps that looks like adding the ulimit option to that command... or perhaps it looks like ripping out that buildah/docker detection and branching and replacing it with `make image-hive`. (That said, we still need an engine-aware `push` in bundle-gen.py...)

2uasimojo:
Well f.

Then I guess let's make this PR add the ulimit flaggage to
- only the buildah build target in the Makefile (i.e. take it out of the docker build target)
- the [buildah build path in bundle-gen](https://github.com/openshift/hive/blob/8ea4577dd43ccdbf68d057defc7afa29cb7dff9d/hack/bundle-gen.py#L165)

2uasimojo:
...and make the commit and PR titles have "buildah" in them.

2uasimojo:
Thanks, sorry for the churn.

Taking it on faith that this works (or at least doesn't break) for buildah users.

/lgtm

PR #1974, opened 2023-03-13
Clarify private subnets requirement for AWS MachinePools in API field.

abutcher:
/assign @2uasimojo

### DIFF
--- a/apis/hive/v1/aws/machinepool.go
+++ b/apis/hive/v1/aws/machinepool.go
@@ -7,8 +7,14 @@ type MachinePoolPlatform struct {
 	Zones []string `json:"zones,omitempty"`
 
 	// Subnets is the list of subnets to which to attach the machines.
-	// There must be exactly one private subnet for each availability zone used.
-	// If public subnets are specified, there must be exactly one private and one public subnet specified for each availability zone.
+	// There must be exactly one private subnet for each availability zones used.
+	//
+	// Public subnets may be specified in addition to private subnets but public subnets will
+	// be filtered out of the subnets list. MachineSets will only use private subnets.
+	// If public subnets are specified, there must be exactly one private and one public subnet
+	// provided for each availability zone.
+	// Note that subnets will be observed to be public if they have an "kubernetes.io/role/elb"
+	// tag in AWS.
 	Subnets []string `json:"subnets,omitempty"`
 
 	// InstanceType defines the ec2 instance type.
--- a/config/crds/hive.openshift.io_machinepools.yaml
+++ b/config/crds/hive.openshift.io_machinepools.yaml
@@ -154,11 +154,16 @@ spec:
                             type: string
                         type: object
                       subnets:
-                        description: Subnets is the list of subnets to which to attach
+                        description: "Subnets is the list of subnets to which to attach
                           the machines. There must be exactly one private subnet for
-                          each availability zone used. If public subnets are specified,
-                          there must be exactly one private and one public subnet
-                          specified for each availability zone.
+                          each availability zones used. \n Public subnets may be specified
+                          in addition to private subnets but public subnets will be
+                          filtered out of the subnets list. MachineSets will only
+                          use private subnets. If public subnets are specified, there
+                          must be exactly one private and one public subnet provided
+                          for each availability zone. Note that subnets will be observed
+                          to be public if they have an \"kubernetes.io/role/elb\"
+                          tag in AWS."
                         items:
                           type: string
                         type: array
--- a/hack/app-sre/saas-template.yaml
+++ b/hack/app-sre/saas-template.yaml
@@ -4882,11 +4882,16 @@ objects:
                               type: string
                           type: object
                         subnets:
-                          description: Subnets is the list of subnets to which to
-                            attach the machines. There must be exactly one private
-                            subnet for each availability zone used. If public subnets
-                            are specified, there must be exactly one private and one
-                            public subnet specified for each availability zone.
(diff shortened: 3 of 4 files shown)


apis/hive/v1/aws/machinepool.go
```diff
@@ -7,8 +7,14 @@ type MachinePoolPlatform struct {
 	Zones []string `json:"zones,omitempty"`
 
 	// Subnets is the list of subnets to which to attach the machines.
-	// There must be exactly one private subnet for each availability zone used.
-	// If public subnets are specified, there must be exactly one private and one public subnet specified for each availability zone.
+	// There must be exactly one private subnet for each availability zones used.
+	//
+	// Public subnets may be specified in addition to private subnets but public subnets will
```
2uasimojo, line 12: Reading this without context/background: Does specifying public subnets confer any benefit or effect, or is it just extraneous and ignored?

2uasimojo: Then perhaps we can reword to make that clearer?

```
Any public subnets -- those tagged with "kubernetes.io/role/elb" in AWS -- will be ignored.
```

abutcher: > Looking at the code, I think we're crossing wires here...

Subnets are also public from Hive's POV if they have an internet gateway route as checked in [isSubnetPublic](https://github.com/openshift/hive/blob/c6bd2e9b9bf757811620ba60cf2d0bc1d0e459fd/pkg/controller/machinepool/awsactuator.go#L469-L522) but only private subnets are passed into into installer MachineSet generation as filtered out [here](https://github.com/openshift/hive/blob/c6bd2e9b9bf757811620ba60cf2d0bc1d0e459fd/pkg/controller/machinepool/awsactuator.go#L421).

2uasimojo: So like... the only thing we (hive) do with public subnets is make sure that you've either specified zero or the exact same number as private subnets? But that can't be the only effect, right? Do they somehow get passed down to the MachineSet on the target? I couldn't find where that happens.

If they don't, then why are we bothering even with the check? According to #1234, we allow them so they can be "consistent with install config"? @suhanime can you provide any context here?

It seems like unnecessary code to me. Why wouldn't we just filter them out and ignore them completely?

suhanime: I went searching for the [original ask](https://issues.redhat.com/browse/HIVE-1224), it looks like we did that for OCM so it could filter out the subnets users provide better (see [comment](https://issues.redhat.com/browse/HIVE-1224?focusedId=15332148&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15332148)). I wouldn't remove that piece of code without checking with them.

2uasimojo: Thanks for finding that! I've cross-linked that card with the PR for ease of future reference.

I'm not proposing removing the filter, which I surmise is necessary so public subnets don't make it down into the MachineSet.

Just proposing removing the extra checking that makes sure the number of public subnets is either zero or the same as the number of private ones.

suhanime: I think that was done because of Abhinav's [comment](https://github.com/openshift/hive/pull/1234#discussion_r546913617) on the PR

apis/hive/v1/aws/machinepool.go
```diff
@@ -7,8 +7,14 @@ type MachinePoolPlatform struct {
...
+	// be filtered out of the subnets list. MachineSets will only use private subnets.
+	// If public subnets are specified, there must be exactly one private and one public subnet
+	// provided for each availability zone.
+	// Note that subnets will be observed to be public if they have an "kubernetes.io/role/elb"
+	// tag in AWS.
```
2uasimojo, line 17: If we're going to keep the checking, just a couple of typographic nits:

```suggestion
	// There must be exactly one private subnet for each availability zone used.
	//
	// Public subnets may be specified in addition to private subnets but public subnets will
	// be filtered out of the subnets list. MachineSets will only use private subnets.
	// If public subnets are specified, there must be exactly one private and one public subnet
	// provided for each availability zone.
	// Note that subnets will be observed to be public if they have a "kubernetes.io/role/elb"
	// tag in AWS.
```

apis/hive/v1/aws/machinepool.go
abutcher: I'd be okay with removing the verification depending on how the installer handles public subnets; if the installer MachineSet generation doesn't accept public subnets then there is some value to us filtering them out. From a cursory scroll through it looks like [public subnets would be accepted and work](https://github.com/openshift/installer/blob/7667b46bd13a6091d8dd7791e77235adfa161935/pkg/asset/machines/aws/machinesets.go#L42) (?).

abutcher: > Zooming out, what is the significance to the user of creating a Machine "in" a public vs private subnet? Is there no use case for the former? I mean, in the past two years nobody has complained that they can't do it, so...

The machine/node would exist in a network segment that would allow outside connections if security group / firewall rules allowed.

We've gotten two asks for this, one related to us identifying subnets as public based on the tag and the other to actually allow public subnets:
- https://issues.redhat.com/browse/HIVE-1805
- https://issues.redhat.com/browse/HIVE-2196

2uasimojo: I put groom labels on the two open cards. We can discuss how we're going to execute on this tomorrow.

abutcher:
Will be superseded by #1984

PR #1975, opened 2023-03-13
Refactor bundlegen script

lleshchi:
xref: [HIVE-2145](https://issues.redhat.com/browse/HIVE-2145)


hack/version.py
```diff
@@ -6,167 +6,154 @@
...
-        if info["commit_hash"] == commit_hash:
-            channels.append(info["channel"])
+    for info in remote_mce_branch_info(hive_repo).values():
+        if info["commit_hash"] == commit.hexsha:
             # We'll only use this if we end up with one
             # entry in this list AND we're not tracking master
             maybe_prefix = info["prefix"]
 
-    if commit_hash == master_hash or (is_master_ancestor and not maybe_prefix):
+    if is_master_ancestor and not maybe_prefix:
```
lleshchi, line 127: Not sure if there is a case where `commit_hash == master_hash and maybe_prefix`?

2uasimojo: I think we have the opportunity to simplify the code so that question becomes moot.

All we want to get out of this function is a commit sha (easy) and a prefix. To get the prefix we just want to know whether we're dealing with `master` or an `mce-*` branch. If the latter, we parsed out the prefix; use it. If the former, use the default const.

We may be able to rewrite `remote_mce_branch_info` to be more generally `branch_info`, containing entries for `master` as well as `mce-*`....

2uasimojo (commented):
partial review

hack/bundle-gen.py
```diff
@@ -358,6 +366,8 @@ def open_pr(
     dest_github_org = upstream_repo.split(":")[1].split("/")[0]
     dest_github_reponame = dir_name
 
+    update_channel = CHANNEL_DEFAULT
```
2uasimojo, line 369: I'd be okay if we got rid of the `update_channel` var entirely and just hardcoded it in the two remaining spots below.

hack/version.py
```diff
@@ -31,142 +25,135 @@
...
-        prefix = maybe_prefix
+        if info["commit_hash"] == commit.hexsha:
+            maybe_prefix.append(info["prefix"])
 
-    if not channels or not prefix:
-        if _mode != "standalone":
+    if len(maybe_prefix) > 1:
+        try:
+            branch_arg = repo.active_branch.name
+        except TypeError:
```
2uasimojo, line 137: What causes this TypeError? If the active_branch is unnamed?

hack/version.py
```diff
@@ -31,142 +25,135 @@
...
 
-    # Was (a commit on) an ocm-X.Y[-mce-M.N] branch specified?
-    # This will find all such branch names and add them to the channel list.
-    # However, we'll only support >1 entry in this list if we're tracking
-    # master, because otherwise which would we use to compute the semver?
-    channels = []
-    maybe_prefix = None
+
+
+def process_current_branch(hive_repo, branch_arg):
```
2uasimojo, line 113: It looks like this is only called in one place right now, with `HEAD` as the `branch_arg`. Given that the name of the func is already `process_current_branch`, we could probably do away with `branch_arg` and just hardcode `HEAD`.

hack/version.py
```diff
@@ -31,142 +25,135 @@
...
+    """Validate and process the input (or default) branch.
```
2uasimojo, line 114: ...and then this ought to say something more like "...the current HEAD".

2uasimojo (commented):
This is looking really good. My suggestions are all around readability.

hack/version.py
lleshchi: Bear with me: Let's say you run the script and your HEAD is on another branch that you created locally, but it's two commits behind the tip of this new local branch. That commit (HEAD) also corresponds to the tip of the master branch and the tip of mce-2.2. Now, this commit does not have a corresponding branch name and it is ambigious as to which branch name version to use. This is unlikely to happen but good to handle nonetheless.

2uasimojo: How does that result in a TypeError though?

lleshchi:
```
  File "<stdin>", line 1, in <module>
  File "/home/lleshchi/.local/lib/python3.11/site-packages/git/repo/base.py", line 896, in active_branch
    return self.head.reference
           ^^^^^^^^^^^^^^^^^^^
  File "/home/lleshchi/.local/lib/python3.11/site-packages/git/refs/symbolic.py", line 309, in _get_reference
    raise TypeError("%s is a detached symbolic reference as it points to %r" % (self, sha))
TypeError: HEAD is a detached symbolic reference as it points to 'cef3cb0773f460247c60c666a33c6ec090522e8e'
```

2uasimojo: Okay, detached HEAD, got it :+1:

hack/version.py
```diff
@@ -31,142 +25,142 @@
...
-        )
+        sha = hive_repo.rev_parse(ref.name).hexsha
+        if sha not in ret:
+            ret[sha] = dict(
+                prefix={prefix},
+                name={ref.name},
+            )
+        else:
+            ret[sha]["prefix"].add(prefix)
+            ret[sha]["name"].add(ref.name)
```
2uasimojo, line 71: jfyi, [collections.defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) exists for this use case

hack/version.py
```diff
@@ -31,142 +25,142 @@
...
-        ret[ref.name] = dict(
-            prefix=prefix,
-            channel=channel,
-            commit_hash=hive_repo.rev_parse(ref.name).hexsha,
-        )
+        sha = hive_repo.rev_parse(ref.name).hexsha
+        if sha not in ret:
+            ret[sha] = dict(
+                prefix={prefix},
+                name={ref.name},
```
2uasimojo, line 67: Why are we saving the ref.name? YAGNI.

hack/version.py
```diff
@@ -31,142 +25,142 @@
...
+    # This will raise an exception if there's no such commit-ish
+    commit = hive_repo.rev_parse("HEAD")
 
-    if not channels or not prefix:
-        if _mode != "standalone":
+    prefix = None
+    maybe_prefix = []
+    info = origin_branch_info(hive_repo)
+    if commit.hexsha in info:
+        maybe_prefix = info[commit.hexsha]["prefix"]
```
2uasimojo, line 136: or
```suggestion
    maybe_prefix = origin_branch_info(hive_repo).get(commit.hexsha, {})
```

hack/version.py
```diff
@@ -31,142 +25,142 @@
...
+
+
+def process_current_branch(hive_repo):
+    """Validate and process the current HEAD.
+
+    :param hive_repo: The git.Repo for the local hive clone.
+    :return commit_hash: The canonical full-length sha of the commit corresponding to the current HEAD.
+    :return prefix: The two-digit semver prefix of the bundle version to use. If HEAD is a
+        (a descendant of) master, we'll use HIVE_VERSION_PREFIX. If not, and it names an 'mce-X.Y'
+        branch, we'll use X.Y.
```
2uasimojo, line 125:
```suggestion
        branch, we'll use X.Y. Otherwise `None` is returned.
```

2uasimojo (commented):
This looks great. A couple of minor suggestions inline.

hack/version.py
lleshchi: I was going to print an error message which included them, but then forgot to add it.

2uasimojo:
/lgtm

Thanks @lleshchi !

I'm going to fast merge this, as the remaining pending CI jobs don't touch these files.

PR #1978, opened 2023-03-14
WIP: Change bundle-gen.py to account for new bundle format

dlom:
xref: [HIVE-1658](https://issues.redhat.com//browse/HIVE-1658)

/assign @2uasimojo

### DIFF
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -12,6 +12,7 @@ import sys
 import tempfile
 import urllib3
 import yaml
+import semver
 
 import version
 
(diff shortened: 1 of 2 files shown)


hack/bundle-gen.py
```diff
@@ -200,6 +200,60 @@ def build_and_push_image(registry_auth_file, image_repo, image_tag, dry_run, bui
...
+            upstream_repo,
+        )
+        raise
+
+    versions.sort(reverse=True)
+    if len(versions) == 0:
+        # Channel not found -- no previous version.
+        return None
+
+    return versions[0]
```
2uasimojo, line 254: Couple things here:
- We shouldn't use lexicographic sort, as e.g. 1.2.100-* will be "less than" 1.2.99-*. Recommend using [semver](https://pypi.org/project/semver/), specifically [compare](https://python-semver.readthedocs.io/en/stable/usage.html#comparing-versions).
- There's really no need to save all the versions in a list -- just save the largest version seen so far.

hack/bundle-gen.py
```diff
@@ -200,6 +200,60 @@ def build_and_push_image(registry_auth_file, image_repo, image_tag, dry_run, bui
...
+
+    try:
+        versions = []
+        hive_dir = os.path.join(repo_full_path, HIVE_SUB_DIR)
+        for version in os.listdir(hive_dir):
+            annotation_yaml_path = os.path.join(hive_dir, version, "metadata", "annotations.yaml")
+            with open(annotation_yaml_path, "r") as stream:
+                annotation_yaml = yaml.load(stream, Loader=yaml.SafeLoader)
+                version_channel = annotation_yaml["annotations"]["operators.operatorframework.io.bundle.channel.default.v1"]
+                if version_channel == channel:
```
2uasimojo, line 240: I think you meant to use 
```suggestion
                if version_channel == channel_name:
```
rather than the global `channel`

2uasimojo (commented):
So IIUC this work would be in two parts:
1. Update all the packages in the upstream repos to the new format
2. Land (the eventual form of) this change to fix how we're computing the previous version.

Right?

dlom:
> Could we maybe instead maintain a metadata file that's ignored by the tooling, in which we store the current latest version? It would need to be updated as part of `open_pr`. I mean, it's basically what we're currently doing with the package.yaml file...
This could be a viable approach as well... something like `.latest_version.yaml` with an entry for each channel

2uasimojo:
> This could be a viable approach as well... something like `.latest_version.yaml` with an entry for each channel

Just so. Then we don't have to clone the whole repo, we can just curl that one file (like we're doing today for the package file).

PR #1976, opened 2023-03-13
Remove replace statement from hive/apis

mjlshen:
This is just a proposal, so this is totally up for discussion, just my thoughts:

The existing replace statement doesn't seem to be strictly needed as we can just update the dependency when we are ready with `go get -u github.com/openshift/api@release-X.YY`. Furthermore, removing it makes importing Hive's API in other modules much easier as the replaces don't need to be carried forward.

E.g. https://github.com/openshift/osdctl/blob/115f5f5b8abea2956ad73845555bcc92bd1728b1/go.mod#L172 needs to maintain a corresponding replace.

Finally, by getting the latest `release-4.13` of `openshift/api`, we also pickup the latest dependencies from upstream K8s as they iterate on `golang.org/x/net` and `golang.org/x/text`

OCP 4.13 is needed specifically things like `machineapi.GCPShieldedInstanceConfig`, `machineapi.SecureBootPolicyEnabled`, and `machineapi.SecureBootPolicyEnabled` which aren't available in <= 4.12

PR #1979, opened 2023-03-14
Fix  flaky `TestGetName`

mjlshen:
I noticed this while working on #1976 and felt it was worth a separate PR. On current latest commit (52f2706d1f70f6bfc1ac13a68b5fb80e544be387), this is quite repeatable with:

```shell
cd ./apis
# Run this a couple times in a row
go test -race -count 50 -run TestGetName ./helpers
```

There was a bug when shortName happened to be generated as a string
of length >= 246 as the test would expect that shortName + "-deploy"
did not exceed 253 characters and was not truncated. However, the 
test previously randomly generated a shortName of length [1, 253),
which causes this test to flake and fail occasionally.

This commit creates a shortBase that at maximum is the longest
possible base that has room for a hash so that it will never be
truncated.

### DIFF
--- a/apis/helpers/namer_test.go
+++ b/apis/helpers/namer_test.go
@@ -7,56 +7,75 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
+const (
+	shortSuffix = "deploy"
+	// hashLen is 10 to represent "-" + hash + "-" (hash itself is always 8 characters)
+	hashLen = 10
+)
+
 func TestGetName(t *testing.T) {
 	for i := 0; i < 10; i++ {
-		shortName := randSeq(rand.Intn(validation.DNS1123SubdomainMaxLength-1) + 1)
-		longName := randSeq(validation.DNS1123SubdomainMaxLength + rand.Intn(100))
+		// shortBase + "-" + hash will be underneath the maximum limit without any truncation
+		shortBase := randSeq(rand.Intn(validation.DNS1123SubdomainMaxLength-hashLen) + 1)
+
+		// Any usage of longBase/longSuffix will definitely need to be truncated
+		longBase := randSeq(validation.DNS1123SubdomainMaxLength + rand.Intn(100) + 1)
+		longSuffix := randSeq(validation.DNS1123SubdomainMaxLength + rand.Intn(100) + 1)
 
 		tests := []struct {
-			base, suffix, expected string
+			name, base, suffix, expected string
 		}{
 			{
-				base:     shortName,
-				suffix:   "deploy",
-				expected: shortName + "-deploy",
+				name:     "Short base and short suffix without truncation",
+				base:     shortBase,
+				suffix:   shortSuffix,
+				expected: shortBase + "-deploy",
 			},
 			{
-				base:     longName,
-				suffix:   "deploy",
-				expected: longName[:validation.DNS1123SubdomainMaxLength-16] + "-" + hash(longName) + "-deploy",
+				name:     "Long base and short suffix truncates base and inserts hash",
+				base:     longBase,
+				suffix:   shortSuffix,
+				expected: longBase[:validation.DNS1123SubdomainMaxLength-hashLen-len(shortSuffix)] + "-" + hash(longBase) + "-" + shortSuffix,
 			},
 			{
-				base:     shortName,
-				suffix:   longName,
-				expected: shortName + "-" + hash(shortName+"-"+longName),
+				name:     "Short base and long suffix inserts a hash instead of appending suffix",
+				base:     shortBase,
+				suffix:   longSuffix,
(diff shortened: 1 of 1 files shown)


apis/helpers/namer_test.go
```diff
@@ -7,57 +7,74 @@ import (
...
+	}
 
-		for _, test := range tests {
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
 			result := GetName(test.base, test.suffix, validation.DNS1123SubdomainMaxLength)
 			if result != test.expected {
-				t.Errorf("Got unexpected result. Expected: %s Got: %s", test.expected, result)
+				t.Errorf("Got unexpected result. \nExpected: %s\nGot:      %s", test.expected, result)
 			}
```
2uasimojo, line 76: [latent] This could be an `assert.Equal()`

apis/helpers/namer_test.go
```diff
@@ -7,57 +7,74 @@ import (
...
+	// hashLen is 10 to represent "-" + hash + "-" (hash itself is always 8 characters)
+	hashLen = 10
+)
+
 func TestGetName(t *testing.T) {
-	for i := 0; i < 10; i++ {
-		shortName := randSeq(rand.Intn(validation.DNS1123SubdomainMaxLength-1) + 1)
-		longName := randSeq(validation.DNS1123SubdomainMaxLength + rand.Intn(100))
+	// shortBase + "-" + hash will be underneath the maximum limit without any truncation
+	shortBase := randSeq(validation.DNS1123SubdomainMaxLength - hashLen + 1)
```
2uasimojo, line 18: Doesn't this still work if we continue to use `rand.Intn` but fix the math?
```suggestion
	shortBase := randSeq(rand.Intn(validation.DNS1123SubdomainMaxLength - hashLen) + 1)
```

(I may still be off by 1, but that's the idea.)

I gather the point is to make sure `shortBase` is never long enough to trigger hashing/truncation when used with a `shortSuffix`; and `longBase`/`longSuffix` are long enough to guarantee that that behavior _is_ triggered.

And the original randomness was there to improve coverage over many runs by using samples of varying lengths, as long as they were within the expected bounds.

So I think we still want that randomness; we just need to make sure we're guaranteeing the correct bounds.

2uasimojo (commented):
Thanks for fixing this!

apis/helpers/namer_test.go
mjlshen: Oh it definitely would - I can definitely put `rand.Intn` back to keep that behavior. 

With this approach I think I should put back the loop to generate 10 random `shortBase`s per test run as well?

apis/helpers/namer_test.go
mjlshen: `testify/assert` isn't a current dependency of the apis package and the code isn't too bloated since it's just string comparisons, so I'm thinking of leaving this be for now

apis/helpers/namer_test.go
mjlshen: Updated the commit message accordingly as well to remove the notes about removing `rand.Intn` and making it deterministic!

apis/helpers/namer_test.go
```diff
@@ -7,56 +7,75 @@ import (
...
 func TestGetName(t *testing.T) {
 	for i := 0; i < 10; i++ {
-		shortName := randSeq(rand.Intn(validation.DNS1123SubdomainMaxLength-1) + 1)
-		longName := randSeq(validation.DNS1123SubdomainMaxLength + rand.Intn(100))
+		// shortBase + "-" + hash will be underneath the maximum limit without any truncation
+		shortBase := randSeq(rand.Intn(validation.DNS1123SubdomainMaxLength-hashLen) + 1)
+
+		// Any usage of longBase/longSuffix will definitely need to be truncated
+		longBase := randSeq(validation.DNS1123SubdomainMaxLength + 100)
+		longSuffix := randSeq(validation.DNS1123SubdomainMaxLength + 100)
```
2uasimojo, line 23: These could also continue to use `rand.Intn(100)` (possibly `+ 1`?) so they're also variable. Being at minimum `validation.DNS1123SubdomainMaxLength + 1` will ensure we get truncation behavior.

2uasimojo (commented):
Looking good. I think the math in the commit message may be off -- 246 rather than 248? -- and the PR message needs to be adjusted as well.

apis/helpers/namer_test.go
mjlshen: Works for me, yeah `rand.Intn(x)` generates a range from `[0, x)`, so you're right about the +1 to guarantee that we go over `validation.DNS1123SubdomainMaxLength` haha

2uasimojo (commented):
/lgtm

Thanks @mjlshen!

PR #1980, opened 2023-03-18
ClusterPool: hold until all clusters deleted

2uasimojo:
Previously, deleting a ClusterPool would cause hive to issue Delete()s for all unassigned clusters and then remove the primary finalizer from the ClusterPool, allowing it to be garbage collected. The result was that those unassigned clusters, as well as any claimed clusters, would be "orphaned" until they were cleaned up.

With this commit, we still issue those Delete()s for unassigned clusters (claimed clusters are still not touched). But we retain the finalizer, causing the ClusterPool to remain in existence, until all of the pool's clusters -- claimed and unclaimed -- are gone.

During this process, we maintain a new status condition of Type `DeletionPossible`, whose Message includes a list of CDs pending cleanup. A couple of things to note about this condition:
- It is unset (Status `Unknown`) until the ClusterPool is `Delete()`d.
- You will _usually_ not see the Status go to `"True"` because the finalizer will be removed very quickly, causing the ClusterPool to disappear (unless there are other finalizers present).

This commit also resolves a long-standing TODO: we now honor the ClusterPool's `MaxConcurrent` setting when issuing the `Delete()`s during this cleanup process. This also takes into account any clusters which were already installing or deleting prior to the ClusterPool being deleted.

HIVE-1557


apis/hive/v1/clusterpool_types.go
```diff
@@ -218,8 +218,11 @@ const (
 	// ClusterPoolAllClustersCurrentCondition indicates whether all unassigned (installing or ready)
 	// ClusterDeployments in the pool match the current configuration of the ClusterPool.
 	ClusterPoolAllClustersCurrentCondition ClusterPoolConditionType = "AllClustersCurrent"
-	// ClusterPoolInventoryValidCondition is set to provide information on whether the cluster pool inventory is valid
+	// ClusterPoolInventoryValidCondition is set to provide information on whether the cluster pool inventory is valid.
 	ClusterPoolInventoryValidCondition ClusterPoolConditionType = "InventoryValid"
+	// ClusterPoolDeletionPossibleCondition gives information about a deleted ClusterPool which is pending cleanup.
+	// Note that it is normal for this condition to remain uninitialized until the ClusterPool is deleted.
```
suhanime, line 224: Yeah, might want to remove this comment to avoid confusion. The way you've set it up, we are always initializing this condition.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -2004,6 +2026,22 @@ func TestReconcileClusterPool(t *testing.T) {
 				}
 			}
 
+			if test.expectedDeletionPossibleStatus != "" {
+				deletionPossibleCondition := controllerutils.FindCondition(pool.Status.Conditions, hivev1.ClusterPoolDeletionPossibleCondition)
+				if assert.NotNil(t, deletionPossibleCondition, "did not find DeletionPossible condition") {
+					assert.Equal(t, test.expectedDeletionPossibleStatus, deletionPossibleCondition.Status,
+						"unexpcted DeletionPossible condition status %s", deletionPossibleCondition.Status)
```
suhanime, line 2033: typo - unexpected

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -2004,6 +2026,22 @@ func TestReconcileClusterPool(t *testing.T) {
...
+			}
+
+			if test.expectedDeletionPossibleMessage != "" {
+				deletionPossibleCondition := controllerutils.FindCondition(pool.Status.Conditions, hivev1.ClusterPoolDeletionPossibleCondition)
+				if assert.NotNil(t, deletionPossibleCondition, "did not find DeletionPossible condition") {
+					assert.Equal(t, test.expectedDeletionPossibleMessage, deletionPossibleCondition.Message,
+						"unexpected DeletionPossible condition message %s", deletionPossibleCondition.Message)
+				}
+			}
+
```
suhanime, line 2044: You can combine the logic here instead of fetching the condition twice.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -1982,7 +2004,7 @@ func TestReconcileClusterPool(t *testing.T) {
 				inventoryValidCondition := controllerutils.FindCondition(pool.Status.Conditions, hivev1.ClusterPoolInventoryValidCondition)
 				if assert.NotNil(t, inventoryValidCondition, "did not find InventoryValid condition") {
 					assert.Equal(t, test.expectedInventoryValidStatus, inventoryValidCondition.Status,
-						"unexpcted InventoryValid condition status %s", inventoryValidCondition.Message)
+						"unexpcted InventoryValid condition status %s", inventoryValidCondition.Status)
```
suhanime, line 2007: typo - unexpected

suhanime (commented):
Some minor changes needed, otherwise changes look good

pkg/controller/clusterpool/clusterpool_controller_test.go
2uasimojo: I looked at that. Because the two config fields are independent, the logic would end up being more convoluted and only slightly less repetitive :(

```suggestion
			if test.expectedDeletionPossibleStatus != "" || test.expectedDeletionPossibleMessage != "" {
				deletionPossibleCondition := controllerutils.FindCondition(pool.Status.Conditions, hivev1.ClusterPoolDeletionPossibleCondition)
				if assert.NotNil(t, deletionPossibleCondition, "did not find DeletionPossible condition") {
					if test.expectedDeletionPossibleStatus != "" {
						assert.Equal(t, test.expectedDeletionPossibleStatus, deletionPossibleCondition.Status,
							"unexpcted DeletionPossible condition status %s", deletionPossibleCondition.Status)
					}
					if test.expectedDeletionPossibleMessage != "" {
						assert.Equal(t, test.expectedDeletionPossibleMessage, deletionPossibleCondition.Message,
							"unexpected DeletionPossible condition message %s", deletionPossibleCondition.Message)
					}
				}
			}

```
Does that seem better to you? I'm okay either way.

I guess I could make the config field the whole condition and just check fields that are set... Yeah, yeah, I'll do that.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -57,6 +58,7 @@ var (
 		hivev1.ClusterPoolCapacityAvailableCondition,
 		hivev1.ClusterPoolAllClustersCurrentCondition,
 		hivev1.ClusterPoolInventoryValidCondition,
+		hivev1.ClusterPoolDeletionPossibleCondition,
```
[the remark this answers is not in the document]
2uasimojo: Yes. I think this is related to your other comment below. Resolved.

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -1089,6 +1119,36 @@ func (r *ReconcileClusterPool) setAvailableCapacityCondition(pool *hivev1.Cluste
...
+	status := corev1.ConditionTrue
+	reason := "DeletionPossible"
+	message := "No ClusterDeployments pending cleanup"
+	updateConditionCheck := controllerutils.UpdateConditionNever
+	if len(cdsPendingDeletion) > 0 {
+		status = corev1.ConditionFalse
+		reason = "ClusterDeploymentsPendingCleanup"
+		// Sort so we don't thrash the message
+		sort.Strings(cdsPendingDeletion)
+		message = fmt.Sprintf("The following ClusterDeployments are pending cleanup: %s", strings.Join(cdsPendingDeletion, ", "))
```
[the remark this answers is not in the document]
2uasimojo: Concern being overrunning the etcd limit for a single object? I'm willing to take that chance.

apis/hive/v1/clusterpool_types.go
2uasimojo: Done.

2uasimojo:
/hold cancel

Glad I did that live testing. I got many things wrong the first time around.

PR #1984, opened 2023-03-30
AWS MachinePool: Allow public and/or private subnets

2uasimojo:
Previously, via [HIVE-1224](https://issues.redhat.com//browse/HIVE-1224) / #1234 / a27ef2e, we started filtering out public subnets if they were provided in an AWS MachinePool. This led to two issues:
- Users could not create machines in public subnets at all ([HIVE-2196](https://issues.redhat.com//browse/HIVE-2196)).
- In PrivateLink, subnets are always private, even though they may appear public based on their attributes ([HIVE-1805](https://issues.redhat.com//browse/HIVE-1805)).

With this commit, we relax the restrictions such that, if the number of subnets matches the number of availability zones, we allow them whether they're public or private.

Here's a complete description of the new behavior:
- We still accept an empty Subnets list, as before.
- If the number of subnets matches the number of availability zones, we allow them, whether they're public or private. (We still validate that we were given exactly one subnet per AZ.)
- If the number of subnets is double the number of availability zones, we follow the existing behavior: filter out the ones we detect are public and pass through only the private ones.
- Any other number of subnets is an error.

[HIVE-1805](https://issues.redhat.com//browse/HIVE-1805)
[HIVE-2196](https://issues.redhat.com//browse/HIVE-2196)

### DIFF
--- a/apis/hive/v1/aws/machinepool.go
+++ b/apis/hive/v1/aws/machinepool.go
@@ -6,9 +6,14 @@ type MachinePoolPlatform struct {
 	// Zones is list of availability zones that can be used.
 	Zones []string `json:"zones,omitempty"`
 
-	// Subnets is the list of subnets to which to attach the machines.
-	// There must be exactly one private subnet for each availability zone used.
-	// If public subnets are specified, there must be exactly one private and one public subnet specified for each availability zone.
+	// Subnets is the list of IDs of subnets to which to attach the machines.
+	// There must be exactly one subnet for each availability zone used.
+	// These subnets may be public or private.
+	// As a special case, for consistency with install-config, you may specify exactly one
+	// private and one public subnet for each availability zone. In this case, the public
+	// subnets will be filtered out and only the private subnets will be used.
+	// If empty/omitted, we will look for subnets in each availability zone tagged with
+	// Name=<clusterID>-private-<az>.
 	Subnets []string `json:"subnets,omitempty"`
 
 	// InstanceType defines the ec2 instance type.
--- a/config/crds/hive.openshift.io_machinepools.yaml
+++ b/config/crds/hive.openshift.io_machinepools.yaml
@@ -172,11 +172,15 @@ spec:
                             type: string
                         type: object
                       subnets:
-                        description: Subnets is the list of subnets to which to attach
-                          the machines. There must be exactly one private subnet for
-                          each availability zone used. If public subnets are specified,
-                          there must be exactly one private and one public subnet
-                          specified for each availability zone.
+                        description: Subnets is the list of IDs of subnets to which
+                          to attach the machines. There must be exactly one subnet
+                          for each availability zone used. These subnets may be public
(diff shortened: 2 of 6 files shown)


2uasimojo:
@abutcher made a start on the code side. PTAL if you get a chance before I get too much deeper :)

@suhanime this touched a fair bit of your code from #1234, would welcome your eyes as well.

2uasimojo:
/assign @abutcher 

@abutcher I would love some tips on how to test this live, when you get a chance.

2uasimojo:
> @abutcher I would love some tips on how to test this live, when you get a chance.

https://redhat-internal.slack.com/archives/CE3ETN3J8/p1680638351592089

2uasimojo:
This tested clean. I tested permutations of:
`zones`: a subset, none, or all
`subnets`: all (corresponding to the zones) public, all (same) private, all (same) of both, and a couple of mismatches (wrong subnets for zones, one too many or one too few)

Things that seemed odd to me, given this was the first time I really played in this space:
- If you don't specify subnets, the spoke MachineSet filters them by name; but the MachinePool expects them to be specified by ID (whereupon the MachineSet filters them by ID).
- If you don't specify any subnets, hive (actually the installer code) generates a MachineSet per zone. If `zones` are not specified in this scenario, it defaults to "all of them", but you can enumerate `zones` and you get the matching private subnet for each. However, if you omit `zones` but specify `subnets` -- even all of them -- it bounces. Not sure if we should fix this.

Ready for review @abutcher.

2uasimojo:
(just a rebase)

2uasimojo:
/test e2e

Seeing this flake a lot lately:

```
admission_test.go:91: Unexpected create error: dnszonevalidators.admission.hive.openshift.io is forbidden: User "system:anonymous" cannot create resource "dnszonevalidators" in API group "admission.hive.openshift.io" at the cluster scope
```

Seems like it could be a timing problem getting RBAC deployed.

apis/hive/v1/aws/machinepool.go
```diff
@@ -7,8 +7,13 @@ type MachinePoolPlatform struct {
 	Zones []string `json:"zones,omitempty"`
 
 	// Subnets is the list of subnets to which to attach the machines.
```
2uasimojo, line 9: Should call out that it's subnet **IDs**.

abutcher:
> If you don't specify any subnets, hive (actually the installer code) generates a MachineSet per zone. If zones are not specified in this scenario, it defaults to "all of them", but you can enumerate zones and you get the matching private subnet for each. However, if you omit zones but specify subnets -- even all of them -- it bounces. Not sure if we should fix this.

We should make a card to at least bump a MachinePool with from validating webhook if we know its an invalid starting point or at most look into fixing in some way.

abutcher:
Aforementioned card https://issues.redhat.com/browse/HIVE-2227 for subnets but no zones.

PR #1988, opened 2023-03-31
New install failure reason: NoWorkerNodesReady

mjlshen:
[OSD-13511](https://issues.redhat.com//browse/OSD-13511)

Investigating AuthenticationOperatorDegraded, which is currently the 
fifth-most common install failure, can be annoying because install
failures are not caused by the Authentication operator failing; instead
it is a symptom of another failure. This PR represents a repeatable
finding where a log line indicating that 0 worker nodes have joined the 
cluster causes AuthenticationOperatorDegraded.

Returning this instead will point us in a better direction to start
investigating the install failure. If we find that it makes up all cases
of AuthenticationOperatorDegraded we can remove that failure reason.

This error code is very similar to AWSNoWorkerNodes and 
GCPNoWorkerNodes, however based on our history neither of them have
triggered for the past 90 days (and anecdotally even longer than that).
I suspect the regex is no longer valid since they were originally
created and this new regex can be their replacement.

cc @iamkirkbater @AlexVulaj

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -115,12 +115,6 @@ data:
       - "InvalidClientTokenId: The security token included in the request is invalid."
       installFailingReason: InvalidCredentials
       installFailingMessage: Credentials are invalid
-    # cf. GCPNoWorkerNodes
-    - name: AWSNoWorkerNodes
-      searchRegexStrings:
-      - "(?s)terraform-provider-aws.*Got 0 worker nodes, \\d+ master nodes"
-      installFailingReason: AWSNoWorkerNodes
-      installFailingMessage: No worker nodes could be created. Check the machine-api logs.
     - name: InvalidAWSTags
       searchRegexStrings:
       - "platform\\.aws\\.userTags.*: Invalid value:.*value contains invalid characters"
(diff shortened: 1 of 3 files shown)


config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -297,6 +285,11 @@ data:
       - "failed to initialize the cluster: Cluster operator monitoring is still updating"
       installFailingReason: MonitoringOperatorStillUpdating
       installFailingMessage: Timeout waiting for the monitoring operator to become ready
+    - name: NoWorkerNodesReady
+      searchRegexStrings:
+      - "Got 0 worker nodes, 3 master nodes.*none are schedulable or ready for ingress pods"
```
2uasimojo, line 290: I would suggest we keep the regex for the number of master nodes here. Even if OSD clusters are always three masters today, they may not be in the future; and you also don't want to miss this event if, say, one of the masters _also_ didn't come up.
```suggestion
      - "Got 0 worker nodes, \\d+ master nodes.*none are schedulable or ready for ingress pods"
```

mjlshen: Makes sense, I'll make this change. From what I understand if any of the masters fail to come up they'll be caught by `BootstrapFailed` (which gets picked up before this failure reason would be):

```
Failed to wait for bootstrapping to complete. This error usually happens when there is a problem with control plane hosts that prevents the control plane operators from creating the control plane.
```

So I _think_ this will only kick in if we ever go to >3 masters, but this covers bases in case I'm wrong about `BootstrapFailed` or if that error falls away in the future.

2uasimojo (commented):
/lgtm

PR #1992, opened 2023-04-04
Update ClusterPool status counts while deleting

2uasimojo:
ClusterPool status fields Size, Standby, and Ready were being ignored in the deletion path for the ClusterPool, and we never noticed because the ClusterPool was disappearing (see referenced card). Fix.

[HIVE-1557](https://issues.redhat.com//browse/HIVE-1557)

### DIFF
--- a/pkg/controller/clusterpool/clusterpool_controller_test.go
+++ b/pkg/controller/clusterpool/clusterpool_controller_test.go
@@ -1107,7 +1107,9 @@ func TestReconcileClusterPool(t *testing.T) {
 			// DeletionTimestamp, isn't getting garbage collected by the fake client. This is the expected result once
 			// the issue is fixed:
 			// expectedTotalClusters: 2,
-			expectedTotalClusters:   3,
+			expectedTotalClusters: 3,
+			// But this one is right
+			expectedObservedSize:    2,
 			expectedCDCurrentStatus: corev1.ConditionUnknown,
 			expectedDeletionPossibleCondition: &hivev1.ClusterPoolCondition{
 				Status:  corev1.ConditionFalse,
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterpool/collections.go
```diff
@@ -616,6 +616,7 @@ func (cds *cdCollection) Delete(c client.Client, cdName string) error {
 	cds.deleting = append(cds.deleting, cd)
 	// Remove from any of the other lists it might be in
 	removeCDsFromSlice(&cds.assignable, cdName)
+	removeCDsFromSlice(&cds.standby, cdName)
```
2uasimojo, line 619: Note to reviewers: latent, missed in #1652.

2uasimojo:
/assign @suhanime 

FUP to #1980 for a bug @jianping-shu found.

2uasimojo:
Something that caught me off guard here: The `SIZE` in the tabular output is [populated](https://github.com/openshift/hive/blob/e880f48e250fa7c246e97f1882ecd66579b51426/apis/hive/v1/clusterpool_types.go#L245) from `spec.size`, not `status.size`. So it's normal for it never to go to zero when the pool is being deleted.

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -1107,7 +1107,9 @@ func TestReconcileClusterPool(t *testing.T) {
 			// DeletionTimestamp, isn't getting garbage collected by the fake client. This is the expected result once
 			// the issue is fixed:
 			// expectedTotalClusters: 2,
-			expectedTotalClusters:   3,
+			expectedTotalClusters: 3,
```
suhanime, line 1110: Since you're adding the comment below, you can remove comment at line 1109

suhanime (commented):
There's a tiny nit but looks okay to me otherwise

pkg/controller/clusterpool/clusterpool_controller_test.go
2uasimojo: Oh, they’re for different values. The comment below is pointing out that the bug referenced above doesn’t apply to the observed size.

suhanime (commented):
/lgtm

PR #1995, opened 2023-04-11
Add new installer failure reason when the install is failing due to new S3 defaults

mjlshen:
This relates to OCPBUGS-11636


suhanime:
Skipping e2e as they wouldn't be affected and this change will help out with the bug.

PR #1997, opened 2023-04-13
Azure custom resource group support for MachinePool, hibernation, deprovision

2uasimojo:
Day 0, the installer knows how to accept a custom resource group from
the install-config and use it for the cluster's cloud resources.
However, hive previously had no visibility to that field (in the
install-config or elsewhere); so for day 2 operations including
hibernation and additional MachinePools as well as for deprovisioning,
we were using the default (<cluster-infra-id>-rg). For machine pools,
this resulted in failures in the operation. In hibernation and
deprovisioning, the operation would "succeed" but no-op. For hibernation
this meant that machines were still running (incurring the associated
costs) even though the cluster claimed it was hibernating. For
deprovisioning, this meant the cloud resources were leaked.

This commit adds a new ClusterDeployment.Spec.ClusterMetadata section
for `Azure` containing a `ResourceGroupName` field, set as follows:
1. When provisioning, we discover it from the installer metadata during
installation.
2. When retrofitting a ClusterDeployment that was adopted or created
before custom Azure resource group support, we first attempt to discover
that resource group via the Infrastructure object in the remote cluster.
3. If that fails, we try parsing the install-config referenced by
ClusterDeployment.Spec.Provisioning.InstallConfigSecretRef.

Note that we will assign the default value if we're using the installer
metadata or the install-config and it's unset there. In any other case
where we can't figure it out (e.g. we can't find the infrastructure
object; or we can't parse the install-config) we just don't set it. The
consequence of that is...

This field is now required for any of the operations that use it -- if
it is absent, those operations will fail. While we don't expect this
condition to persist, as a fallback the user can manually set the value
if necessary.

[HIVE-2204](https://issues.redhat.com//browse/HIVE-2204)


pkg/azureclient/client.go
```diff
@@ -46,7 +46,7 @@ type Client interface {
 	GetVirtualMachineSku(ctx context.Context, name, region string) (*compute.ResourceSku, error)
 
 	// Images
-	ListImagesByResourceGroup(ctx context.Context, resourgeGroupName string) (ImageListResultPage, error)
+	ListImagesByResourceGroup(ctx context.Context, resourceGroupName string) (ImageListResultPage, error)
```
2uasimojo, line 49: NTR: latent typo

pkg/controller/clusterdeployment/installconfigvalidation.go
```diff
@@ -2,9 +2,10 @@ package clusterdeployment
 
 import (
 	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
+	"sigs.k8s.io/yaml"
```
2uasimojo, line 5: NTR: yaml.v2 seems not to work properly for unmarshalling an install-config. Some (but not all) fields are unrecognized and end up unset. That makes me leery of using it at all anywhere, so a subsequent PR will purge other uses of it in this project. This PR changes just this one instance to get the feature working.

pkg/controller/hibernation/azure_actuator_test.go
```diff
@@ -22,11 +22,12 @@ import (
...
 
 const (
-	azureTestSubscription  = "1111-111-1111-111111"
-	azureTestInfraID       = "test-infra-id"
-	azureTestResourceGroup = "test-infra-id-rg"
+	azureTestSubscription = "1111-111-1111-111111"
+	azureTestInfraID      = "test-infra-id"
 )
 
+var azureTestResourceGroup = "test-infra-id-rg"
```
2uasimojo, line 29: NTR: moved from const to var so I can take its address below.

2uasimojo (commented):
/assign @abutcher

2uasimojo:
/hold

Confirmed the fix on Jose's env, with existing CDs.
Going to want to do a little more testing there:
- A new CD with an existing custom RG
- Using managed DNS if this is reasonably easy to set up
...but I'll need Jose's help.

Also want to have a look at the coverage report and possibly add some more unit tests for the new code paths.

abutcher (commented):
Looks good to me! :+1:

Edit: Noticed that we aren't plumbing the resource group into uninstall which I think makes sense because we aren't certain that the status field for the resource group is correct (?).

abutcher:
I'm slightly worried that we aren't honoring the [installer's promise to delete the resource group when specified](https://github.com/openshift/installer/blob/978b35e0c735196ecbe96688fb976c8556041dbc/pkg/types/azure/platform.go#L80-L88), "Destroying the cluster using installer will delete this resource group." We aren't cleaning up the resource group currently (via uninstall) and, unless we required the `resourceGroupName` field to be mirrored into a `ClusterDeployment` spec field at creation time, we won't be sure that we've stored the correct resource group in status.

2uasimojo:
Ah, yup, missed that!

So I think we want to consider putting the new field in Spec.ClusterMetadata instead of status.
- Like the InfraID therein, it shouldn't be messed with by the user once set.
- That makes the adoption path more comfortable, as they already have to populate known values into ClusterMetadata, and this way they don't have to muck with status in a separate step.

I wonder if we want to set it at the same time as InfraID -- when provisioning succeeds. Then we're not retrofitting existing clusters. Which, if we allow updating that field in ClusterMetadata (will have to look at webhooks...), gives the user a clean workaround, as opposed to editing a status field. But it means that in some cases where we _could_ have automatically set the correct value, we didn't, and the user would _have_ to do it by hand.

Thoughts?

abutcher:
> I wonder if we want to set it at the same time as InfraID -- when provisioning succeeds. Then we're not retrofitting existing clusters. Which, if we allow updating that field in ClusterMetadata (will have to look at webhooks...), gives the user a clean workaround, as opposed to editing a status field. But it means that in some cases where we could have automatically set the correct value, we didn't, and the user would have to do it by hand.

This is for the case where user has configured a custom resource group in the install-config but no longer has an install-config? I think that's acceptable.

Another thing I hadn't considered, and may be too heavy to check for this, is that the running cluster records the resource group in the cluster `infrastructure` object. 

Example `infrastructure` object from a cluster where I've used a custom resource group in install-config:

```
$ oc get infrastructure cluster -o yaml
apiVersion: config.openshift.io/v1
kind: Infrastructure
metadata:
...
  generation: 1
  name: cluster
...
status:
  controlPlaneTopology: HighlyAvailable
  etcdDiscoveryDomain: ""
  infrastructureName: abutcher-az-jcf5r
  infrastructureTopology: HighlyAvailable
  platform: Azure
  platformStatus:
    azure:
      cloudName: AzurePublicCloud
      networkResourceGroupName: abutchertest
      resourceGroupName: abutchertest
    type: Azure
```

2uasimojo:
> This is for the case where user has configured a custom resource group in the install-config but no longer has an install-config? I think that's acceptable.

For that case I'm okay with skipping, or with trying and getting it wrong.

I'm more concerned about the case where the install-config is still in place and unchanged, which I would expect to be by far the most common thing. With the PR as it stands, we'll correctly discover and set the rg in that case. With the proposed change, we wouldn't; we would leave it completely to the user to retrofit their own CD after the upgrade.

Now, the probability of such a scenario still isn't super high: it would have to be someone who used a custom RG but didn't care about hibernation or day 2 machinepools, and now they do.

Nice find on the infrastructure object! Of course, there are also cons with that: namely, we have to be able to remote into the cluster. I suppose we rely on that for other things -- including setting up remote MachineSets :) -- so it's a viable consideration. And it's less "vendoring dependent" than parsing the install-config (though as you can see from this PR, we were doing that anyway). But the main advantage to parsing the install-config is that we don't need the remote cluster to be accessible -- like let's say you're adopting a dead cluster just so you can clean up its cloud resources.

A big problem we have today is that the hibernation and deprovision paths both "succeed", but don't actually touch the resources in the custom rg. So you think you've shut down your machines to save money, but you actually haven't. Or you think you've destroyed your cluster, but you actually haven't. That's bad. Almost like we should put in some kind of sanity check that looks for a minimum set of resources and fails if it doesn't find them... but that feels hacky and dangerous. So I would like to minimize the probability of that happening.

Probably overthinking as usual.

2uasimojo:
Okay, I added the deprov path (cribbed/adapted from #1973) and tested on Jose's cluster, working fine.

Right now I still have the new field in status. Although ClusterMetadata _feels_ like a good place for it, we don't currently have a Platform section there. Let me noodle this over the weekend and we can discuss on Monday.

2uasimojo:
We've decided to move the new field to spec under ClusterMetadata.

2uasimojo:
[32471ab](https://github.com/openshift/hive/commit/32471ab7183f17dbf98fdb0983e4b82c29626bd2) rebase only

2uasimojo:
Top commit is just a placeholder for now, reminding next-week-me what to do and how to do it.

abutcher:
I tested the following scenarios and clusters were able to be provisioned+deprovisioned successfully. 
- No custom resource group specified in install config. Default value.
- Custom resource group specified in install config.
- Delete ClusterDeployment before cluster provisioning completed.

apis/hive/v1/azure/metadata.go
```diff
@@ -2,5 +2,10 @@ package azure
 
 // Metadata contains Azure metadata (e.g. for uninstalling the cluster).
 type Metadata struct {
+	// ResourceGroupName is the name of the resource group in which the cluster resources were created.
+	ResourceGroupName *string `json:"resourceGroupName"`
+
+	// TODO: I think this is obsolete (or was never used) -- can we remove it?
+	// Added via https://github.com/openshift/hive/pull/546
```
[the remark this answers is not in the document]
2uasimojo: ✓

2uasimojo:
yeah, had a feeling

2uasimojo:
Given the conflict, the cherry-pick should actually work after #2006 lands.

PR #1998, opened 2023-04-13
s,gopkg.in/yaml.v2,sigs.k8s.io/yaml for json-tagged structs

2uasimojo:
We recently discovered that `gopkg.in/yaml.v2` doesn't work properly in
cases where the struct being unmarshalled to has only `json` tags. An
important example is unmarshalling install-config. The reason can be
found in this blog post [1].

Swap to `sigs.k8s.io/yaml` in the couple of places where this applies.
(Notably, we're leaving alone places where the target struct has `yaml`
tags, as the interface between the two libraries is not fully
compatible.)

[1] https://web.archive.org/web/20190603050330/http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/

[HIVE-2205](https://issues.redhat.com//browse/HIVE-2205)

### DIFF
--- a/pkg/controller/clusterdeployment/installconfigvalidation.go
+++ b/pkg/controller/clusterdeployment/installconfigvalidation.go
@@ -2,7 +2,7 @@ package clusterdeployment
 
 import (
 	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
+	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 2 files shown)


2uasimojo:
> same interface

Clearly untrue.

/hold

2uasimojo:
Hello tide? 628 is closed since like Tuesday.

/hold

PR #1999, opened 2023-04-17
Beef up clusterwide proxy support

2uasimojo:
This commit rounds out our support for clusterwide proxies by doing two
things:
- Previously, we relied on an outside agent (typically OLM) to set
`HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` environment variables on the
hive-operator deployment  We would pass these values down to the
controllers and thence to workload pods such as provisioner and
deprovisioner. This commit adds logic to discover those values from the
cluster proxy object if not already set.
- Install the cluster proxy's trusted CA bundle in all provision and
deprovision pods. The ultimate source of this bundle is the ConfigMap
referenced by the proxy's spec.trustedCA.name. We take advantage of the
Cluster Network Operator's facility [1] to inject the merged CA bundle --
which includes the proxy's trusted CA bundle -- into a ConfigMap in the
appropriate namespace and mount that ConfigMap on the pods.

TODO: I tried to add a `Watch()` for `Proxy` to hive-operator (so we can
redeploy the controllers if the proxy changes) but this caused the
operator to hang on startup. Couldn't figure it out this time around.

[1] https://docs.openshift.com/container-platform/4.12/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki

[HIVE-2198](https://issues.redhat.com//browse/HIVE-2198)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
@@ -2239,7 +2239,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				func() runtime.Object {
 					cd := testClusterDeploymentWithDefaultConditions(testClusterDeploymentWithInitializedConditions(testClusterDeployment()))
 					cd.Status.InstallRestarts = 1
-					cd.Spec.InstallAttemptsLimit = pointer.Int32Ptr(2)
+					cd.Spec.InstallAttemptsLimit = pointer.Int32(2)
 					return cd
 				}(),
 				testSecret(corev1.SecretTypeDockerConfigJson, pullSecretSecret, corev1.DockerConfigJsonKey, "{}"),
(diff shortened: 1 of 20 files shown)


2uasimojo:
/hold cancel

Pending ack by @jianping-shu, but I think this might be ready to go now.

/assign @abutcher 
/cc @lleshchi

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -2239,7 +2239,7 @@ func TestClusterDeploymentReconcile(t *testing.T) {
 				func() runtime.Object {
 					cd := testClusterDeploymentWithDefaultConditions(testClusterDeploymentWithInitializedConditions(testClusterDeployment()))
 					cd.Status.InstallRestarts = 1
-					cd.Spec.InstallAttemptsLimit = pointer.Int32Ptr(2)
+					cd.Spec.InstallAttemptsLimit = pointer.Int32(2)
```
2uasimojo, line 2242: ntr: latent, resolve linter complaints throughout

PR #2000, opened 2023-04-24
Ensure hive-trusted-cabundle earlier

2uasimojo:
In [HIVE-2198](https://issues.redhat.com//browse/HIVE-2198) / #1999 / 253408ed8 we started injecting a ConfigMap containing a CA bundle for better proxy support. Alas, we were only injecting it in the provisioning path. For hives that never had the earlier code, this probably would have been fine, as the ConfigMap would continue to exist until the cluster was deleted, whereupon it could properly be mounted on the deprovisioning pods. But for spokes that existed prior to upgrading to that code, and which were never re-reconciled before they were deleted (or, conceivably, if a user deleted the ConfigMap at some point in there) the ConfigMap would not exist, resulting in mount failures attempting to bring up the deprovisioning pods.

This commit moves the ConfigMap injection earlier in the flow, ensuring it exists for both provisioning and deprovisioning paths.

[HIVE-2207](https://issues.redhat.com//browse/HIVE-2207)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -521,6 +521,11 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		}
 	}
 
+	if err := r.ensureTrustedCABundleConfigMap(cd.Namespace); err != nil {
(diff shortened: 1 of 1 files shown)


2uasimojo:
Manual merging around freeze as this is a fix that needs to get into stage before we promote.

PR #2003, opened 2023-04-26
Create Pull Policy on Installer Image During Cluster Provisioning

bennerv:
Story isn't groomed yet.  Related card / ask from ARO team: https://issues.redhat.com/browse/HIVE-2209

### DIFF
--- a/pkg/imageset/updateinstaller.go
+++ b/pkg/imageset/updateinstaller.go
@@ -182,6 +182,12 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
 		o.log.WithField("installerImage", installerImage).Info("installer image found")
 	}
 
+	installerImagePullPolicy := string(corev1.PullIfNotPresent)
+	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallerImagePullPolicyOverride != "" {
+		installerImagePullPolicy = cd.Spec.Provisioning.InstallerImagePullPolicyOverride
+		o.log.WithField("installer image pull policy", installerImagePullPolicy).Info("installer image pull policy overridden")
+	}
+
 	cliImage, err := findImageSpec(is, "cli")
 	if err != nil {
 		return errors.Wrap(err, "could not get cli image")
@@ -221,6 +227,7 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
 	cd.Status.InstallerImage = &installerImage
 	cd.Status.CLIImage = &cliImage
 	cd.Status.InstallVersion = &releaseVersion
+	cd.Status.InstallerImagePullPolicy = &installerImagePullPolicy
 	// Set InstallerImageResolutionFailedCondition to false
 	cd.Status.Conditions = controllerutils.SetClusterDeploymentCondition(
 		cd.Status.Conditions,
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -297,6 +297,11 @@ func InstallerPodSpec(
 	}
 	installerImage := *cd.Status.InstallerImage
 
+	if cd.Status.InstallerImagePullPolicy == nil {
+		return nil, fmt.Errorf("installer pull policy not resolved")
+	}
+	installerImagePullPolicy := *cd.Status.InstallerImagePullPolicy
+
 	if cd.Status.CLIImage == nil {
 		return nil, fmt.Errorf("cli image not resolved")
 	}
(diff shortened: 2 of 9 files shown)


pkg/install/generate.go
```diff
@@ -325,7 +325,7 @@ func InstallerPodSpec(
 		{
 			Name:            "cli",
 			Image:           cliImage,
-			ImagePullPolicy: corev1.PullIfNotPresent,
+			ImagePullPolicy: images.GetInstallerClusterProvisionImagePullPolicy(),
```
2uasimojo, line 328: I think you meant for this to go on L317.

pkg/controller/images/controller_images.go
```diff
@@ -47,6 +53,13 @@ func GetHiveClusterProvisionImagePullPolicy() corev1.PullPolicy {
 	return corev1.PullPolicy(envVarOrDefault(HiveClusterProvisionImagePullPolicyEnvVar, string(corev1.PullIfNotPresent)))
 }
 
+// GetInstallerClusterProvisionImagePullPolicy returns the policy to use when pulling the installer image
+// for the ClusterProvision pod.
+// Either the one specified in the environment variable or the hardcoded default.
+func GetInstallerClusterProvisionImagePullPolicy() corev1.PullPolicy {
+	return corev1.PullPolicy(envVarOrDefault(InstallerClusterProvisionImagePullPolicyEnvVar, string(corev1.PullIfNotPresent)))
```
2uasimojo, line 60: We would also need a way to get this env var down to the controller. In fact, it looks to me like the `HiveClusterProvisionImagePullPolicyEnvVar` (upon which I'm guessing you based this change) can't be used today for that same reason.

2uasimojo (commented):
FWIW, when I glanced at this I was considering adding an `InstallerImagePullPolicy` field right next to the existing [InstallerImageOverride](https://github.com/openshift/hive/blob/7cbb91212d8a79ad33c87d89d823d84ae3983f68/apis/hive/v1/clusterdeployment_types.go#L226-L229). The installmanager pod already has the ClusterDeployment, so that would be easy to read, vs having to percolate an env var down through three layers of invocation (hive-operator => hive-controllers => installer pod).

2uasimojo:
Hijacking this PR to sniff out some CI issues...

/test e2e-azure
/test e2e-gcp

pkg/imageset/updateinstaller.go
```diff
@@ -182,6 +182,12 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
 		o.log.WithField("installerImage", installerImage).Info("installer image found")
 	}
 
+	installerImagePullPolicy := string(corev1.PullIfNotPresent)
```
bennerv, line 185: This is where the default is.  I'm not sure if I should pull it up into a const to make it more visible.

pkg/imageset/updateinstaller.go
```diff
@@ -182,6 +182,12 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
...
+	if cd.Spec.Provisioning != nil && cd.Spec.Provisioning.InstallerImagePullPolicyOverride != "" {
+		installerImagePullPolicy = cd.Spec.Provisioning.InstallerImagePullPolicyOverride
+		o.log.WithField("installer image pull policy", installerImagePullPolicy).Info("installer image pull policy overridden")
```
2uasimojo, line 188: nit: I think keys for log fields are usually made without spaces so as to make the log line more easily parseable/greppable.

pkg/imageset/updateinstaller.go
2uasimojo: Shrug, if ya like, I'm fine either way.

pkg/imageset/updateinstaller.go
```diff
@@ -221,6 +227,7 @@ func (o *UpdateInstallerImageOptions) Run() (returnErr error) {
 	cd.Status.InstallerImage = &installerImage
 	cd.Status.CLIImage = &cliImage
 	cd.Status.InstallVersion = &releaseVersion
+	cd.Status.InstallerImagePullPolicy = &installerImagePullPolicy
```
2uasimojo, line 230: femtonit: might move this up next to InstallerImage @L227.

2uasimojo (commented):
This looks good. Couple of nonblocking items inline.

However, please see recent comment on the [card](https://issues.redhat.com/browse/HIVE-2209): this would not be necessary if InstallerImageOverride were specified by digest rather than by tag. Is there a good reason that's not done?

bennerv:
@2uasimojo 
>this would not be necessary if InstallerImageOverride were specified by digest rather than by tag. Is there a good reason that's not done?

I had a comment typed up last night around this.  

I'm going to see ARO's level of effort to implement this on ARO's side for production vs. this PR change.  We're going to shift to the full digest eventually, just need to confirm if we need this as a stop gap temporarily or not.

2uasimojo:
Cool. If it's just a stopgap, then I would greatly prefer it not be made part of the official API, cause we can't undo that once it's in place. It could instead be an annotation on the CD. And in that case I think we should be able to do without the status field as well, right?

bennerv:
@2uasimojo I'm going to close this as I think we can do it from our side with a similar amount of effort and it saves some code changes from hive's side as well.

PR #2005, opened 2023-05-04
Only Watch Proxy when running on OpenShift

2uasimojo:
[HIVE-2198](https://issues.redhat.com//browse/HIVE-2198) / #1999 / 253408e introduced a `Watch()` on `Proxy` so we can
update environment variables and CA bundles when it changes. But `Proxy`
is an OpenShift CRD, so that `Watch()` blows up the operator when
running on vanilla k8s (and similar).

This commit conditions that `Watch()` such that we only add it if we're
running on OpenShift.

[HIVE-2210](https://issues.redhat.com//browse/HIVE-2210)


abutcher (approved):
/lgtm

PR #2007, opened 2023-05-08
Dummy ca-bundle.crt for non-OpenShift

2uasimojo:
If we're not running on OpenShift, the magic label doesn't cause OpenShift to populate the merged CA bundle ConfigMap. This would cause the ConfigMap mount to fail, and provisions wouldn't start. This commit defaults that key to empty so the file will always exist and the mount will work. (Note that `update-ca-trust` properly ignores an empty file.)

[HIVE-2210](https://issues.redhat.com//browse/HIVE-2210)

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -509,6 +509,9 @@ const (
 
 	// TrustedCABundleDir is the path into which the merged CA bundle will be mounted.
 	TrustedCABundleDir = "/hive-trusted-cabundle"
+
+	// TrustedCABundleFile is the name of the file (and corresponding ConfigMap key) containing the merged CA bundle.
+	TrustedCABundleFile = "ca-bundle.crt"
 )
 
 // GetMergedPullSecretName returns name for merged pull secret name per cluster deployment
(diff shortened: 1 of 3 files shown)


abutcher:
Tested image [quay.io/2uasimojo/hive:2210-dummy-cabundle](https://quay.io/2uasimojo/hive:2210-dummy-cabundle) on a kind cluster and was able to create and destroy a cluster.

```
➜  hive (master) ✔ kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.24.1
Kustomize Version: v4.5.4
Server Version: v1.24.0
```

```
➜  hive (master) ✔ oc get cd
NAME           INFRAID              PLATFORM   REGION      VERSION   CLUSTERTYPE   PROVISIONSTATUS   POWERSTATE   AGE
abutchertest   abutchertest-6fzcf   aws        us-east-1   4.13.0                  Provisioned       Running      35m
```

I ran into issues due to kind running kube v1.24+ since [ServiceAccount token secrets not being automatically created ](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes) that I worked around by creating the necessary secrets. But, that's a separate issue we should resolve in `make deploy`.

For example,
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: hiveadmission-sa-token
  namespace: hive
  annotations:
    kubernetes.io/service-account.name: "hiveadmission"
type: kubernetes.io/service-account-token
```

/lgtm

2uasimojo:
We've now got https://issues.redhat.com/browse/HIVE-2219 to address those missing secrets.

PR #2008, opened 2023-05-09
[OSD-16302] Add new install failure for blocked AWS account.

bergmannf:
It can happen that AWS is blocking an account so the installation will fail.


2uasimojo:
Fine by me

/lgtm

PR #2006, opened 2023-05-05
Beef up clusterwide proxy support

2uasimojo:
This PR rounds out our support for clusterwide proxies by doing two
things:
- Previously, we relied on an outside agent (typically OLM) to set
`HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` environment variables on the
hive-operator deployment  We would pass these values down to the
controllers and thence to workload pods such as provisioner and
deprovisioner. This commit adds logic to discover those values from the
cluster proxy object if not already set.
- Install the cluster proxy's trusted CA bundle in all provision and
deprovision pods. The ultimate source of this bundle is the ConfigMap
referenced by the proxy's spec.trustedCA.name. We take advantage of the
Cluster Network Operator's facility [1] to inject the merged CA bundle --
which includes the proxy's trusted CA bundle -- into a ConfigMap in the
appropriate namespace and mount that ConfigMap on the pods.

[1] https://docs.openshift.com/container-platform/4.12/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki

[HIVE-2198](https://issues.redhat.com//browse/HIVE-2198)

PR #2011, opened 2023-05-10
Azure custom resource group support

2uasimojo:
Day 0, the installer knows how to accept a custom resource group from
the install-config and use it for the cluster's cloud resources.
However, hive previously had no visibility to that field (in the
install-config or elsewhere); so for day 2 operations including
hibernation and additional MachinePools as well as for deprovisioning,
we were using the default (<cluster-infra-id>-rg). For machine pools,
this resulted in failures in the operation. In hibernation and
deprovisioning, the operation would "succeed" but no-op. For hibernation
this meant that machines were still running (incurring the associated
costs) even though the cluster claimed it was hibernating. For
deprovisioning, this meant the cloud resources were leaked.

This commit adds a new ClusterDeployment.Spec.ClusterMetadata section
for `Azure` containing a `ResourceGroupName` field, set as follows:
1. When provisioning, we discover it from the installer metadata during
installation.
2. When retrofitting a ClusterDeployment that was adopted or created
before custom Azure resource group support, we first attempt to discover
that resource group via the Infrastructure object in the remote cluster.
3. If that fails, we try parsing the install-config referenced by
ClusterDeployment.Spec.Provisioning.InstallConfigSecretRef.

Note that we will assign the default value if we're using the installer
metadata or the install-config and it's unset there. In any other case
where we can't figure it out (e.g. we can't find the infrastructure
object; or we can't parse the install-config) we just don't set it. The
consequence of that is...

This field is now required for any of the operations that use it -- if
it is absent, those operations will fail. While we don't expect this
condition to persist, as a fallback the user can manually set the value
if necessary.

[HIVE-2204](https://issues.redhat.com//browse/HIVE-2204)


2uasimojo:
/hold

This is currently sitting atop #2006, which we should land first.

2uasimojo:
It hath been decreed that this shall not go into mce-2.3

/close

2uasimojo:
(But if this does get reopened, remember to add #2020)

PR #2012, opened 2023-05-10
Create ServiceAccount token secret for hiveadmission

2uasimojo:
Starting with k8s 1.24, service account token secrets are no longer created automatically by k8s. The hiveadmission service needs one, so this commit explicitly creates one. This should work for pre- and post-1.24.

[HIVE-2219](https://issues.redhat.com//browse/HIVE-2219)

### DIFF
--- a/config/hiveadmission/sa-token-secret.yaml
+++ b/config/hiveadmission/sa-token-secret.yaml
@@ -1,0 +1,9 @@
+# Prior to k8s 1.24, this secret was generated automatically
+apiVersion: v1
+kind: Secret
+metadata:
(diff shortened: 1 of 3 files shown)


abutcher:
Tested on a kind cluster running kube v1.24.0. The hiveadmission SA secret was created and the hiveadmission pod started. 

e2e passing must mean that creating the secret on OpenShift is copacetic. hiveadmission SA token secrets created by kube prior to v1.24.0 had a random slug like `hiveadmission-token-fwlmm` and this secret will be extra but shouldn't cause any issue based on my testing.

/lgtm

PR #2013, opened 2023-05-10
Add timestamps to messages in e2e tests

2uasimojo:
Debugging e2e failures -- especially those with a timing component -- will be easier if we have date stamps on our output messages. Make it so.

Note that this commit is only overriding the `echo` command. Output from subcommands won't be affected. That may turn out to be a problem, but it's one we can solve later.

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -1,3 +1,7 @@
+# Add timestamps to our output
+shopt -s expand_aliases
+alias echo='/bin/echo -n `date -Ins --universal`"  "; /bin/echo'
+
 ###
 # TEMPORARY workaround for https://issues.redhat.com/browse/DPTP-2871
 # The configured job timeout after isn't signaling the test script like it


2uasimojo:
Wait, what? This didn’t work at all?

/hold

2uasimojo:
> Wait, what? This didn’t work at all?

https://unix.stackexchange.com/questions/1496/why-doesnt-my-bash-script-recognize-aliases

/hold cancel

2uasimojo:
Okay, this worked ([example](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_hive/2013/pull-ci-openshift-hive-master-e2e/1656664197703929856/artifacts/e2e/test/build-log.txt))

/assign @lleshchi

2uasimojo:
Let's see if we can repro the e2e-pool flake, as that's why I did this...

/test e2e-pool

PR #2017, opened 2023-05-11
Add featureset flag to hiveutil

jstuever:
Add a featureset flag to the hiveutil command. The string value provided will be passed directly into the installconfig.


2uasimojo (commented):
Yup, straightforward enough.

/lgtm

PR #2018, opened 2023-05-11
Add azure-resource-group-name flag to hiveutil

jstuever:
Add a azure-resource-group-name flag to the hiveutil command. The string
value provided will be passed directly to the install config
platform.azure.resourceGroupName parameter.

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -179,6 +179,7 @@ type Options struct {
 	// Azure
 	AzureBaseDomainResourceGroupName string
 	AzureCloudName                   string
+	AzureResourceGroupName           string
 
 	// OpenStack
 	OpenStackCloud             string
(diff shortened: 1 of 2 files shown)


apis/hive/v1/azure/platform.go
```diff
@@ -24,6 +24,10 @@ type Platform struct {
 	// If empty, the value is equal to "AzurePublicCloud".
 	// +optional
 	CloudName CloudEnvironment `json:"cloudName,omitempty"`
+
+	// ResourceGroupName specifies a resource group where the cluster will be installed.
+	// +optional
+	ResourceGroupName string `json:"resourceGroupName,omitempty"`
```
2uasimojo, line 30: We deliberately decided not to do this when designing #1997 because we didn't want to have to validate/reconcile potential differences between a new hive API field and what the user put in the install-config. We decided the install-config is the source of truth, and in the normal provisioning path we accept the value from the installer right next to the InfraID in clusterMetadata.

2uasimojo (commented):
> * rebase on [Azure custom resource group support #2011](https://github.com/openshift/hive/pull/2011)

That's the mce-2.3 backport -- I think you want #1997 instead. And that's already landed, so you should just be able to rebase this on master and the first 10 commits will melt away :)

2uasimojo (commented):
Thanks!

/lgtm

PR #2020, opened 2023-05-12
Allow azure resource group to be set/updated in clusterMetadata

2uasimojo:
Hive controllers *and* in some cases the consumer must be able to set ClusterDeployment.spec.clusterMetadata.platform.azure.resourceGroupName. Previously our "clusterMetadata is immutable" validating webhook would (uh, sometimes??) bounce such a change. Make a special case for this sub-field.

[HIVE-2222](https://issues.redhat.com//browse/HIVE-2222)


pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook_test.go
```diff
@@ -694,6 +695,132 @@ func TestClusterDeploymentValidate(t *testing.T) {
...
+						},
+					},
+				}
+				return cd
+			}(),
+			operation:       admissionv1beta1.Update,
+			expectedAllowed: true,
+		},
+		{
+			name: "Azure update resource group name",
```
2uasimojo, line 774: These last two test cases were copy-pasted but not finished.

PR #2021, opened 2023-05-12
e2e-pool: minor fixes

2uasimojo:
- Fix a logic bug where we were redundantly hibernating a cluster in the fake pool while it was still in the pool and thus already hibernating. Now we claim a cluster, validate that it's Running by the time the claim is fulfilled, then hibernate it.
- Also add resuming the above.
- Also add deleting the claim and verifying that the claimed CD gets deleted.

### DIFF
--- a/hack/e2e-pool-test.sh
+++ b/hack/e2e-pool-test.sh
@@ -229,9 +229,11 @@ oc get clusterpool ${REAL_POOL_NAME} -o json \
 
 # Add customization to REAL POOL
 # NOTE: Use the cdcci- prefix for this cluster name so it gets caught by our leak detector.
+# NOTE: We do this after the fake pool creation so the fake pool doesn't pick up the same inventory.
 NEW_CLUSTER_NAME=cdcci-${CLUSTER_NAME#*-}
 create_customization "cdc-test" "${CLUSTER_NAMESPACE}" "${NEW_CLUSTER_NAME}"
 oc patch cp -n $CLUSTER_NAMESPACE $REAL_POOL_NAME --type=merge -p '{"spec": {"inventory": [{"name": "cdc-test"}]}}'
+
 
 wait_for_pool_to_be_ready $FAKE_POOL_NAME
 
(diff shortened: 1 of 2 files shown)


2uasimojo:
/retest

e2e was a plain ol flake.

e2e-pool looks like a (latent, unrelated) timing problem here:

```
# Modify the clusterpool and watch CDs get replaced
oc patch clusterpool $FAKE_POOL_NAME --type=merge -p '{"spec":{"imageSetRef": {"name": "fake-cis"}}}'
expect_all_clusters_current $FAKE_POOL_NAME False
oc wait --for=condition=AllClustersCurrent --timeout=10m clusterpool/$FAKE_POOL_NAME
```

I surmise that `expect_all_clusters_current` picked up the state from right before the patch update was acked by the controller. If we see this again, perhaps we want to try something more like:

```
# Modify the clusterpool and watch CDs get replaced
oc patch clusterpool $FAKE_POOL_NAME --type=merge -p '{"spec":{"imageSetRef": {"name": "fake-cis"}}}'
oc wait --for=condition=AllClustersCurrent=false --timeout=1m clusterpool/$FAKE_POOL_NAME
oc wait --for=condition=AllClustersCurrent --timeout=10m clusterpool/$FAKE_POOL_NAME
```

2uasimojo:
Same again. Must have changed the timing just enough. Let’s try the plan above. 

But just out of curiosity…

/test e2e-pool

2uasimojo:
> Same again. Must have changed the timing just enough. Let’s try the plan above.

Nope. I introduced a bug by doing this:

> (Cosmetic) Move the creation of ClusterDeploymentCustomization for the real pool out of the fake pool section.

It resulted in the fake pool getting the same inventory as the real pool, leading to thrashing and things never becoming ready. Reverted.

2uasimojo:
...but

> the plan above.

is good anyway, so I'm gonna keep it.

hack/e2e-pool-test.sh
```diff
@@ -242,21 +244,34 @@ verify_pool_cd_imagesets $FAKE_POOL_NAME $IMAGESET_NAME
 create_imageset fake-cis
 # Modify the clusterpool and watch CDs get replaced
 oc patch clusterpool $FAKE_POOL_NAME --type=merge -p '{"spec":{"imageSetRef": {"name": "fake-cis"}}}'
-expect_all_clusters_current $FAKE_POOL_NAME False
+oc wait --for=condition=AllClustersCurrent=false --timeout=1m clusterpool/$FAKE_POOL_NAME
```
lleshchi, line 247: Is it worth replacing the other call to `expect_all_clusters_current` with `oc wait`?

2uasimojo: Some of them maybe, but definitely not universally. The `oc wait` is for when we’re expecting a state transition, whereas the other is where we’re expecting the pool to already be in a particular state. Could be a separate effort to go through and audit each occurrence.

PR #2023, opened 2023-05-15
Comment: freeze ClusterAutoscaler configuration by MachinePool contro…

2uasimojo:
…ller

The MachinePool controller should not be in the business of configuring the (singleton) ClusterAutoscaler. We've got some minimal setup in the code that's considered necessary if the (any) MachinePool has Autoscaling enabled; but further configuration should be left to the user -- whether in-cluster or from the hub via syncsets.

[HIVE-2220](https://issues.redhat.com//browse/HIVE-2220)


osherdp (approved):
it makes sense to have this comment either way. LGTM

2uasimojo:
We're going a different direction via #2025 or similar. See HIVE-2220 for details.

PR #2027, opened 2023-05-18
Re-enable autoscale test for Azure/GCP

2uasimojo:
...which was disabled via #2019 / a3963cf0d.

The only thing we've changed since then is that we're now testing against OCP 4.13 instead of 4.14, which seemed to clear up flakes found after this test was disabled.


2uasimojo:
/test unit

infra flake

2uasimojo:
/test e2e-pool

May need to backport #2021

PR #2032, opened 2023-05-25
Hibernation: No more pause, hack ClusterOperators

2uasimojo:
This commit changes how we check the health of ClusterOperators when resuming from hibernation.

The problem is that we can't immediately tell by looking at a ClusterOperator whether it has been checked yet since the cluster was resumed. This is because the status conditions' `lastTransitionTime`s are only updated if/when the condition's `status` changes -- and it may not change if the CO is already healthy by the time it is inspected. Thus even if the CO's status conditions indicate it is healthy, we can't be sure that that state is not stale, from before we hibernated.

Previously, we had a hardcoded 2m wait after nodes became ready, on the assumption that this was enough time to ensure all COs had been checked.

With this commit, we get rid of that 2m wait time. Instead, we update the CO's `Available` condition to have `status="Unknown"`. This forces its owning controller to update the status and the `lastTransitionTime`, allowing us to ascertain definitively that that has happened since we started resuming (i.e. the `lastTransitionTime` of the ClusterDeployment's `Hibernating` condition).

[HIVE-2226](https://issues.redhat.com//browse/HIVE-2226)

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -91,9 +91,6 @@ const (
 	// ClusterPowerStateWaitingForNodes is used when waiting for nodes to become Ready.
 	ClusterPowerStateWaitingForNodes ClusterPowerState = "WaitingForNodes"
 
-	// ClusterPowerStatePausingForClusterOperatorsToSettle is used when pausing to let ClusterOperators start and post new status before we check it.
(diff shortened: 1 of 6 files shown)


2uasimojo:
WIP:
- Fix UT
- Live test

For grins:

/test e2e-pool

2uasimojo:
This doesn't work. Unfortunately not all COs' operators are well-behaved in the necessary way. We could fix them, but it would take years for the unfixed versions to age out of support. If we do that, it'll be a separate effort, and this work can be resurrected down the road.

PR #2035, opened 2023-05-31
Hibernation: No more pause, hack ClusterOperators

2uasimojo:
This commit changes how we check the health of ClusterOperators when resuming from hibernation.

The problem is that we can't immediately tell by looking at a ClusterOperator whether it has been checked yet since the cluster was resumed. This is because the status conditions' `lastTransitionTime`s are only updated if/when the condition's `status` changes -- and it may not change if the CO is already healthy by the time it is inspected. Thus even if the CO's status conditions indicate it is healthy, we can't be sure that that state is not stale, from before we hibernated.

Previously, we had a hardcoded 2m wait after nodes became ready, on the assumption that this was enough time to ensure all COs had been checked.

With this commit, we get rid of that 2m wait time. Instead, we update the CO's `Available` condition to have `status="Unknown"`. This forces its owning controller to update the status and the `lastTransitionTime`, allowing us to ascertain definitively that that has happened since we started resuming (i.e. the `lastTransitionTime` of the ClusterDeployment's `Hibernating` condition).

WIP: Just editing status on restart doesn't work: some operators do not pay attention to their CO CR, so they don't revert that status (unless their pods are restarted -- maybe?). Current incarnation of this commit is experimenting with hacking the status *before hibernation* as well as on resume. But we still have some timing issues to work out: In order to not hot loop the before-hibernation bit, we would only want to hack the COs whose `lastTransitionTime` is earlier than when we started hibernating (i.e. the Ready condition's `lastTransitionTime`). But that means we would have to transition that status before hacking operators, which means adding a new state to the flow. This commit is partway there, but doesn't have it right yet. (I'm becoming more and more concerned that the hibernation controller is too brittle to handle how this change is turning out -- we may just want to punt and try to work it into the ever-forthcoming PowerState controller.)

[HIVE-2226](https://issues.redhat.com//browse/HIVE-2226)

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -74,6 +74,10 @@ const (
 	// ClusterPowerStateFailedToStartMachines
 	ClusterPowerStateFailedToStartMachines ClusterPowerState = "FailedToStartMachines"
 
+	// ClusterPowerStatePreppingClusterOperators indicates we are updating ClusterOperator
+	// statuses so we can tell when they have been refreshed on resume.
+	ClusterPowerStatePreppingClusterOperators ClusterPowerState = "PreparingClusterOperators"
+
 	// ClusterPowerStateStopping indicates the cluster is transitioning
 	// from a Running state to a Hibernating state.
 	ClusterPowerStateStopping ClusterPowerState = "Stopping"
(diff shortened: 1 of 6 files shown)


2uasimojo:
Continuation of #2032

2uasimojo:
This experiment didn't work.

PR #2036, opened 2023-06-01
MachinePool: Allow Platform edits via override

2uasimojo:
Normally we forbid updates to MachinePool.Spec.Platform via an admission hook because, while we would roll out the changes to the MachineSets on the target cluster, MCO ignores those changes.

With this commit, you can add an annotation to a MachinePool
```
hive.openshift.io/override-machinepool-platform: "true"
```
to bypass this check. By doing so, you are taking responsibility for forcing rollout of any changes, e.g. by deleting the Machines on the target cluster.

[HIVE-2024](https://issues.redhat.com//browse/HIVE-2024)

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -512,6 +512,12 @@ const (
 
 	// TrustedCABundleFile is the name of the file (and corresponding ConfigMap key) containing the merged CA bundle.
 	TrustedCABundleFile = "ca-bundle.crt"
+
+	// OverrideMachinePoolPlatformAnnotation can be set to "true" on a MachinePool to bypass the validating admission
+	// hook that normally forbids modifying its Spec.Platform. By setting this annotation, you are taking responsibility
+	// for forcing rollout of such changes on the target cluster -- e.g. by deleting the Machines -- as the machine
+	// config operator will not do so.
+	OverrideMachinePoolPlatformAnnotation = "hive.openshift.io/override-machinepool-platform"
 )
 
 // GetMergedPullSecretName returns name for merged pull secret name per cluster deployment
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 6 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -829,7 +836,7 @@ func (r *ReconcileMachinePool) syncClusterAutoscaler(
 			ScaleDown: &autoscalingv1.ScaleDownConfig{
 				Enabled: true,
 			},
-			BalanceSimilarNodeGroups: pointer.BoolPtr(true),
+			BalanceSimilarNodeGroups: pointer.Bool(true),
```
2uasimojo, line 839: ntr: unrelated de-linting of deprecated `pointer.*Ptr()` throughout.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -597,6 +597,13 @@ func (r *ReconcileMachinePool) syncMachineSets(
 					objectModified = true
 				}
 
+				// Platform updates will be blocked by webhook, unless they're not.
```
2uasimojo, line 600: ntr: I am *not* double-checking the override annotation here. Open for discussion, but I think it's better to leave that decision in one place. If we got past the webhook (for any reason -- even the usual hack of deleting the VWHC) we'll effect the change.

pkg/constants/constants.go
```diff
@@ -512,6 +512,12 @@ const (
 
 	// TrustedCABundleFile is the name of the file (and corresponding ConfigMap key) containing the merged CA bundle.
 	TrustedCABundleFile = "ca-bundle.crt"
+
+	// OverrideMachinePoolPlatformAnnottaion can be set to "true" on a MachinePool to bypass the validating admission
```
lleshchi, line 516:
```suggestion
	// OverrideMachinePoolPlatformAnnotation can be set to "true" on a MachinePool to bypass the validating admission
```

lleshchi (changes requested):

pkg/constants/constants.go
2uasimojo: Wow, I even fixed this typo in the actual variable name, but I guess either I didn't use F2 or F2 didn't sweep up the comment (sometimes it does, sometimes it doesn't). Thanks!

pkg/controller/machinepool/machinepool_controller.go
abutcher: This scares me because it opens us up to any diff that would be generated by the installer's generation where we haven't been considering drift at all, now we will blast out any changes such as with a default on a big scale.

2uasimojo: ...which would mean
- short term the MachineSets and Machines are out of sync
- longer term if machines get replaced for whatever reason, they'll be different for hard-to-understand reasons
- ditto new machines when scaling up

And all of this of course only if hive has done a revendor since the Machines were created *and* the revendor pulled in some changed default from the installer code.

So really, really low probability... but ugly enough to warrant paranoia, and certainly outbalances doubling the check, which is easy.

On it.

2uasimojo: @abutcher done

PR #2044, opened 2023-06-08
e2e: Save interim manifests

2uasimojo:
It was annoying to not be able to get things like clusterdeployment manifests for e2e runs that completed successfully. Now e2e will save those manifests twice: once right after the ClusterDeployment has been created and once on cleanup. (As today, the latter will only be interesting if the test fails.)


2uasimojo:
This is sort of related to HIVE-2234 and HIVE-2235, where it's really tough to tell what version of OpenShift we're actually installing on the spokes based on the information currently left in the artifacts dir after a run. With this PR hopefully we can get that from e.g. the cluster version annotations on the CD.

PR #2045, opened 2023-06-09
Add awsprivatelink subcommand to hiveutil

fxierh:
Automate the networking setup & teardown between associate VPCs & endpoint VPCs, as well as the update of HiveConfig. 

With the new subcommand, e2e tests on PrivateLink clusters can be automated as follows:
1. Create a CloudFormation stack (not necessarily in Hive cluster's region) for an endpoint VPC and get the returned stackID. Here we can use "./hack/awsprivatelink/vpc.cf.yaml" for example. 
2. Wait until the stack is successfully created. 
3. Get VPC ID and private subnet IDs from the above stack.
4. Enable PrivateLink with "hiveutil awsprivatelink enable ..." 
5. Set up networking between the new endpoint VPC & each associate VPC with "hiveutil awsprivatelink endpointvpc add ..."
6. Create a PrivateLink cluster in the region of the endpoint VPC with "hiveutil create-cluster fxie-hive-1 --base-domain qe.devcluster.openshift.com --region endpoint-vpc-region --aws-private-link --internal"
7. Run e2e tests on the PrivateLink cluster
8. Delete the PrivateLink CD, wait for the deletion to complete
9. Tear down networking elements between the new endpoint VPC & each associate VPC with "hiveutil awsprivatelink endpointvpc remove ..."
10. Disable PrivateLink with "hiveutil awsprivatelink disable ..."
11. Delete the CloudFormation stack

Checkout https://issues.redhat.com/browse/OCPQE-14714 for details.

### DIFF
--- a/contrib/pkg/awsprivatelink/disable.go
+++ b/contrib/pkg/awsprivatelink/disable.go
@@ -1,0 +1,113 @@
+package awsprivatelink
+
+import (
+	"context"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	hiveutils "github.com/openshift/hive/contrib/pkg/utils"
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
+	operatorutils "github.com/openshift/hive/pkg/operator/hive"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+type disableOptions struct {
+	dynamicClient client.Client
+}
+
+func NewDisableAWSPrivateLinkCommand() *cobra.Command {
+	opt := &disableOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "disable",
+		Short: "Disable AWS PrivateLink",
+		Long: `Disable AWS PrivateLink:
+1) Remove Secret(s) with AWS hub account credentials created when calling "hiveutil awsprivatelink enable ..."
+2) Empty HiveConfig.spec.awsPrivateLink`,
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := opt.Complete(cmd, args); err != nil {
+				return
+			}
+			if err := opt.Validate(cmd, args); err != nil {
+				return
+			}
+			if err := opt.Run(cmd, args); err != nil {
+				return
+			}
+		},
+	}
+	return cmd
+}
+
+func (o *disableOptions) Complete(cmd *cobra.Command, args []string) error {
+	// Get controller-runtime dynamic client
+	dynamicClient, err := hiveutils.GetClient()
+	if err != nil {
+		log.WithError(err).Fatal("Failed to create controller-runtime client")
+	}
+	o.dynamicClient = dynamicClient
+
+	return nil
+}
+
+func (o *disableOptions) Validate(cmd *cobra.Command, args []string) error {
+	return nil
+}
+
+func (o *disableOptions) Run(cmd *cobra.Command, args []string) error {
+	// Get HiveConfig
+	hiveConfig := &hivev1.HiveConfig{}
+	if err := o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to get HiveConfig/hive")
+	}
+	if hiveConfig.Spec.AWSPrivateLink == nil {
+		log.Warn("AWS PrivateLink is already disabled in HiveConfig")
+	}
+	// It is assumed that no cloud resources for networking present
+	// if and only if there is either no endpoint VPC or no associated VPC.
+	if hiveConfig.Spec.AWSPrivateLink != nil &&
+		len(hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory) > 0 &&
+		len(hiveConfig.Spec.AWSPrivateLink.AssociatedVPCs) > 0 {
+		log.Fatal("HiveConfig has at least 1 associated VPC and 1 endpoint VPC specified. " +
+			"Please either remove all associated VPCs or all endpoint VPCs from it. " +
+			"Please remember to delete relevant cloud resources for networking " +
+			"(between the associated VPCs and endpoint VPCs) as you remove the VPCs." +
+			"You can remove the endpoint VPCs (and relevant cloud resources for networking) " +
+			"one by one via `hiveutil awsprivatelink endpointvpc remove ...`.")
+	}
+
+	// Delete Hub account secret(s) if present
+	hiveNS := operatorutils.GetHiveNamespace(hiveConfig)
+	hubAcctSecrets := &corev1.SecretList{}
+	if err := o.dynamicClient.List(
+		context.Background(),
+		hubAcctSecrets,
+		client.MatchingFields{"metadata.name": awsutils.PrivateLinkHubAcctCredsName},
+		client.MatchingLabels{awsutils.PrivateLinkHubAcctCredsLabel: "true"},
+		client.InNamespace(hiveNS),
+	); err != nil {
+		log.WithError(err).Error("Failed to list Hub account credentials Secrets")
+	}
+
+	for _, hubAcctSecret := range hubAcctSecrets.Items {
+		if err := o.dynamicClient.Delete(context.Background(), &hubAcctSecret); err != nil {
+			log.WithError(err).Errorf("Failed to delete Hub account credentials Secret %v", hubAcctSecret.Name)
+		} else {
+			log.Infof("Hub account credentials Secret %v deleted", hubAcctSecret.Name)
+		}
+	}
+
+	// Empty HiveConfig.spec.awsPrivateLink
+	hiveConfig.Spec.AWSPrivateLink = nil
+	if err := o.dynamicClient.Update(context.Background(), hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to update HiveConfig")
+	}
+	log.Info("HiveConfig updated")
+
+	return nil
+}
--- a/contrib/pkg/awsprivatelink/enable.go
+++ b/contrib/pkg/awsprivatelink/enable.go
@@ -1,0 +1,221 @@
+package awsprivatelink
+
+import (
+	"context"
+	"errors"
+	"os/user"
+	"path/filepath"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	configv1 "github.com/openshift/api/config/v1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	hiveutils "github.com/openshift/hive/contrib/pkg/utils"
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
+	"github.com/openshift/hive/pkg/awsclient"
+	operatorutils "github.com/openshift/hive/pkg/operator/hive"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type enableOptions struct {
+	homeDir string
+	// AWS region of the Hive cluster
+	region        string
+	infraId       string
+	dnsRecordType string
+
+	dynamicClient client.Client
+	// AWS clients in Hive cluster's region
+	awsClients awsclient.Client
+}
+
+func NewEnableAWSPrivateLinkCommand() *cobra.Command {
+	opt := &enableOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "enable",
+		Short: "Enable AWS PrivateLink",
+		Long: `Enable AWS PrivateLink:
+1) Extract AWS Hub account credentials from the environment where this command is called
+2) Create a Secret with the above credential.
+3) Add a reference to the Secret in HiveConfig.spec.awsPrivateLink.credentialsSecretRef.
+4) Add the active cluster's VPC to the list of HiveConfig.spec.awsPrivateLink.associatedVPCs`,
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := opt.Complete(cmd, args); err != nil {
+				return
+			}
+			if err := opt.Validate(cmd, args); err != nil {
+				return
+			}
+			if err := opt.Run(cmd, args); err != nil {
+				return
+			}
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&opt.dnsRecordType, "dns-record-type", "Alias", "HiveConfig.spec.awsPrivateLink.dnsRecordType")
+	return cmd
+}
+
+func (o *enableOptions) Complete(cmd *cobra.Command, args []string) error {
+	// Get controller-runtime dynamic client
+	if err := configv1.Install(scheme.Scheme); err != nil {
+		log.WithError(err).Fatal("Failed to add Openshift configv1 types to the default scheme")
+	}
+	dynamicClient, err := hiveutils.GetClient()
+	if err != nil {
+		log.WithError(err).Fatal("Failed to create controller-runtime client")
+	}
+	o.dynamicClient = dynamicClient
+
+	// Get current user's home directory
+	o.homeDir = "."
+	u, err := user.Current()
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get the current user")
+	}
+	o.homeDir = u.HomeDir
+
+	// Get AWS clients
+	o.region, o.infraId, err = o.getRegionAndInfraId()
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get region and infraID from Infrastructure/cluster")
+	}
+	log.Debugf("Found region = %v, infraId = %v", o.region, o.infraId)
+	awsClients, err := awsclient.NewClientFromSecret(nil, o.region)
+	if err != nil {
+		log.WithError(err).Fatal("Failed to create AWS clients")
+	}
+	o.awsClients = awsClients
+
+	return nil
+}
+
+func (o *enableOptions) Validate(cmd *cobra.Command, args []string) error {
+	switch o.dnsRecordType {
+	case "Alias", "ARecord":
+	default:
+		log.Fatal(`--dns-record-type must be one of "Alias", "ARecord"`)
+	}
+
+	return nil
+}
+
+func (o *enableOptions) Run(cmd *cobra.Command, args []string) error {
+	// Get HiveConfig
+	hiveConfig := &hivev1.HiveConfig{}
+	if err := o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to get HiveConfig/hive")
+	}
+	if hiveConfig.Spec.AWSPrivateLink != nil {
+		log.Fatal("AWS Private Link is already enabled. If a previous configuration attempt did not complete, " +
+			"you can clean up via `hiveutil awsprivatelink disable` and try again.")
+	}
+
+	// Get active cluster's VPC, filtering by infra-id
+	targetTagKey := "kubernetes.io/cluster/" + o.infraId
+	describeVPCsOutput, err := o.awsClients.DescribeVpcs(&ec2.DescribeVpcsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag-key"),
+				Values: []*string{aws.String(targetTagKey)},
+			},
+		},
+	})
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get VPC of the active cluster")
+	}
+	if len(describeVPCsOutput.Vpcs) == 0 {
+		log.Fatal("VPC of the active cluster not found")
+	}
+	if len(describeVPCsOutput.Vpcs) > 1 {
+		log.Fatalf("Multiple VPCs found with tag key %s, cannot determine VPC of the active cluster", targetTagKey)
+	}
+	vpcID := *describeVPCsOutput.Vpcs[0].VpcId
+	log.Debugf("Found VPC ID = %v for the active cluster", vpcID)
+
+	// Create Hub account credentials Secret
+	hiveNS := operatorutils.GetHiveNamespace(hiveConfig)
+	hubCredentialsSecret, err := o.generateAWSCredentialsSecret(hiveNS)
+	if err != nil {
+		log.WithError(err).Fatal("Failed to generate Secret with AWS credentials")
+	}
+	switch err = o.dynamicClient.Create(context.Background(), hubCredentialsSecret); {
+	case err == nil:
+		log.Infof("Secret/%s created in namespace %s", awsutils.PrivateLinkHubAcctCredsName, hiveNS)
+	case apierrors.IsAlreadyExists(err):
+		log.Warnf("Secret/%s already exists in namespace %s", awsutils.PrivateLinkHubAcctCredsName, hiveNS)
+	default:
+		log.WithError(err).Fatalf("Failed to create Secret/%s in namespace %s", awsutils.PrivateLinkHubAcctCredsName, hiveNS)
+	}
+
+	// Update HiveConfig
+	hiveConfig.Spec.AWSPrivateLink = &hivev1.AWSPrivateLinkConfig{
+		AssociatedVPCs: []hivev1.AWSAssociatedVPC{
+			{
+				AWSPrivateLinkVPC: hivev1.AWSPrivateLinkVPC{VPCID: vpcID, Region: o.region},
+			},
+		},
+		CredentialsSecretRef: corev1.LocalObjectReference{Name: hubCredentialsSecret.Name},
+		DNSRecordType:        hivev1.AWSPrivateLinkDNSRecordType(o.dnsRecordType),
+	}
+	if err = o.dynamicClient.Update(context.Background(), hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to update HiveConfig")
+	}
+	log.Info("HiveConfig updated")
+
+	return nil
+}
+
+func (o *enableOptions) getRegionAndInfraId() (string, string, error) {
+	infrastructure := &configv1.Infrastructure{}
+	if err := o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "cluster"}, infrastructure); err != nil {
+		return "", "", err
+	}
+	if infrastructure.Status.PlatformStatus == nil {
+		return "", "", errors.New("Infrastructure.status.platformStatus is empty")
+	}
+	if infrastructure.Status.PlatformStatus.AWS == nil {
+		return "", "", errors.New("Infrastructure.status.platformStatus.aws is empty")
+	}
+
+	return infrastructure.Status.PlatformStatus.AWS.Region, infrastructure.Status.InfrastructureName, nil
+}
+
+// Adapted from contrib/pkg/adm/managedns/enable.go:generateAWSCredentialsSecret()
+func (o *enableOptions) generateAWSCredentialsSecret(namespace string) (*corev1.Secret, error) {
+	defaultCredsFilePath := filepath.Join(o.homeDir, ".aws", "credentials")
+	accessKeyID, secretAccessKey, err := awsutils.GetAWSCreds("", defaultCredsFilePath)
+	if err != nil {
+		return nil, err
+	}
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: corev1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      awsutils.PrivateLinkHubAcctCredsName,
+			Namespace: namespace,
+			// Secrets without this label (e.g. the ones created and configured manually) won't be deleted
+			// when calling "hiveutil awsprivatelink disable".
+			Labels: map[string]string{awsutils.PrivateLinkHubAcctCredsLabel: "true"},
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			"aws_access_key_id":     accessKeyID,
+			"aws_secret_access_key": secretAccessKey,
+		},
+	}, nil
+}
--- a/contrib/pkg/awsprivatelink/endpointvpc/add.go
+++ b/contrib/pkg/awsprivatelink/endpointvpc/add.go
@@ -1,0 +1,418 @@
+package endpointvpc
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	hiveutils "github.com/openshift/hive/contrib/pkg/utils"
+	awsutils "github.com/openshift/hive/contrib/pkg/utils/aws"
+	"github.com/openshift/hive/pkg/awsclient"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type endpointVPCAddOptions struct {
+	hiveConfig        hivev1.HiveConfig
+	associatedVpcs    []hivev1.AWSAssociatedVPC
+	endpointVpcId     string
+	endpointVpcRegion string
+	endpointSubnetIds []string
+
+	dynamicClient      client.Client
+	endpointVpcClients awsclient.Client
+	awsClientsByRegion map[string]awsclient.Client
+}
+
+func NewEndpointVPCAddCommand() *cobra.Command {
+	opt := &endpointVPCAddOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "add vpc-id",
+		Short: "Networking setup between an endpoint VPC and each associated VPC",
+		Long: `Networking setup between an endpoint VPC and each associated VPC 
+specified in HiveConfig.spec.awsPrivateLink.associatedVPCs: 
+1) Establish VPC peering connection between the endpoint VPC and the associated VPC
+2) Add a route (to the peered VPC) in relevant route tables of the associated VPC and the endpoint VPC 
+3) Update relevant security groups of the associated VPC and the endpoint VPC to allow traffic between them
+4) Add the endpoint VPC to HiveConfig.spec.awsPrivateLink.endpointVPCInventory`,
+		Args: cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := opt.Complete(cmd, args); err != nil {
+				return
+			}
+			if err := opt.Validate(cmd, args); err != nil {
+				return
+			}
+			if err := opt.Run(cmd, args); err != nil {
+				return
+			}
+		},
+	}
+
+	regionFlag := "region"
+	subnetIdsFlag := "subnet-ids"
+
+	flags := cmd.Flags()
+	flags.StringVar(&opt.endpointVpcRegion, regionFlag, "", "AWS Region of the endpoint VPC to add")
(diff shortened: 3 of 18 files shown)


fxierh:
"hiveutil awsprivatelink associatevpc ..." subcommands could also be added in the future.

fxierh:
@jianping-shu Here we go

fxierh:
I guess e2e job is not necessary for this but emm... /retest

fxierh:
Okay, I guess I'm not allowed to run CI here

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,224 @@
...
+
+func NewEnableAWSPrivateLinkCommand() *cobra.Command {
+	opt := &enableOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "enable",
+		Short: "Enable AWS PrivateLink",
+		Long: `Enable AWS PrivateLink:
+1) Extract AWS Hub account credentials from the environment where this command is called
+2) Create a Secret with the above credential. The Secret is referred to by HiveConfig.spec.awsPrivateLink.credentialsSecretRef
```
2uasimojo, line 52: This is slightly confusing: I can't tell by reading it whether the command will add the credentialsSecretRef or if that's my responsibility. Reading the code I see it's the former, which is cool. Suggest splitting the second sentence into its own bullet and rewording slightly, e.g.
```
3) Add a reference to the Secret in HiveConfig.spec.awsPrivateLink.credentialsSecretRef.
```

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Get Hive client
+	hiveClient, err := hiveclient.NewForConfig(cfg)
+	if err != nil {
+		log.WithError(err).Fatal("Failed to create Hive clientset")
+	}
+	o.hiveClient = hiveClient
+
+	// Get Openshift CoreV1 client
+	openshiftConfigClient, err := openshiftcorev1.NewForConfig(cfg)
```
2uasimojo, line 94: When you cut over to `GetClient()`, you may need to add configv1 to the scheme [here](https://github.com/openshift/hive/blob/9dcb1a4d8acc309d69e708e57f94a98f93f59f52/contrib/pkg/utils/client.go#L20).

contrib/pkg/utils/aws/aws.go
```diff
@@ -1,16 +1,214 @@
...
+)
+
+// AssociatedVpcWithClient encapsulates an associated VPC for PrivateLink and an AWS EC2 client for it
+type AssociatedVpcWithClient struct {
+	AWSEC2Client  *ec2.EC2
+	AssociatedVpc *hivev1.AWSPrivateLinkVPC
+}
+
+// GetAWSEC2Client gets an AWS EC2 client in a certain region.
+func GetAWSEC2Client(region string) *ec2.EC2 {
```
2uasimojo, line 36: I would prefer you reuse the existing [AWS client wrapper](https://github.com/openshift/hive/blob/4dfe9ecb15e1db492177e12ab1d5385b98092850/pkg/awsclient/client.go#L60). Add more wrapper methods to it if you need.

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,224 @@
...
+	u, err := user.Current()
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get the current user's home directory")
+	}
+	o.homeDir = u.HomeDir
+
+	// Get EC2 client
+	o.region, o.infraId = o.getRegionAndInfraId()
+	log.Debugf("Found region = %v, infraId = %v", o.region, o.infraId)
+	o.EC2Client = awsutils.GetAWSEC2Client(o.region)
```
2uasimojo, line 111: As noted below, please use the existing AWS client wrapper. You should be able to init the client via [client.New()](https://github.com/openshift/hive/blob/4dfe9ecb15e1db492177e12ab1d5385b98092850/pkg/awsclient/client.go#L397) or one of its brethren in that file. In your case possibly [NewClientFromSecret()](https://github.com/openshift/hive/blob/4dfe9ecb15e1db492177e12ab1d5385b98092850/pkg/awsclient/client.go#L482) with (confusingly) a `nil` Secret.

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (o *enableOptions) Run(cmd *cobra.Command, args []string) error {
+	// Get HiveConfig, make sure AWS PrivateLink is not enabled
+	hiveConfig, err := o.hiveClient.HiveV1().HiveConfigs().Get(context.Background(), "hive", metav1.GetOptions{})
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get HiveConfig/hive")
+	}
+	if hiveConfig.Spec.AWSPrivateLink != nil {
+		log.Fatal("AWS Private Link is already enabled")
+	}
```
2uasimojo, line 134: I'm wondering if it's worth proceeding for the sake of idempotence. I.e. if something went wrong partway through the last time we invoked this, it would be nice to be able to (fix the problem and) run it again to complete the configuration.

contrib/pkg/awsprivatelink/disable.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Delete Hub account secret(s) if present
+	hiveNS := hiveConfig.Spec.TargetNamespace
+	if hiveNS == "" {
+		hiveNS = constants.DefaultHiveNamespace
+	}
+	hubAcctSecrets, err := o.kubernetesClient.CoreV1().Secrets(hiveNS).List(
+		context.Background(),
+		metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=true", awsutils.PrivateLinkHubAcctCredLabel),
```
2uasimojo, line 97: So as written, this command will only clean up Secrets created via `enable`. This is as opposed to explicitly looking for the Secret named by HiveConfig.spec.awsPrivateLink.credentialsSecretRef and deleting it. Was that an intentional decision?

2uasimojo (commented):
Partial review. Feel free to start addressing things. I'll get back to this on Tuesday (Monday is a US holiday).

contrib/pkg/awsprivatelink/disable.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if len(hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory) > 0 {
+		log.Fatal("Please remove endpoint VPCs (and the underlying cloud resources) first")
+	}
+
+	// Delete Hub account secret(s) if present
+	hiveNS := hiveConfig.Spec.TargetNamespace
+	if hiveNS == "" {
+		hiveNS = constants.DefaultHiveNamespace
+	}
```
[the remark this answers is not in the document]
fxierh: May I export this function ?

contrib/pkg/awsprivatelink/disable.go
fxierh: Yep. I was thinking of the following:
In extreme situations (e.g. "hiveutil awsprivatelink enable ..." fails midway for a few times then reruns successfully), there could be multiple such Secrets, and it would be good to delete all of them.

fxierh: Not sure if the above is a good idea.

contrib/pkg/awsprivatelink/enable.go
fxierh: I'm hesitating between the following options:
1) Abandon immediately in case HiveConfig.spec.awsPrivateLink != nil (current implementation)
2) Proceed regardless of the value of HiveConfig.spec.awsPrivateLink. In this way the existing HiveConfig.spec.awsPrivateLink is re-written.
3) Proceed regardless of the value of HiveConfig.spec.awsPrivateLink, but do not touch HiveConfig.spec.awsPrivateLink if it is non-nil (similar to the current implementation)

contrib/pkg/awsprivatelink/disable.go
```diff
@@ -0,0 +1,119 @@
...
+		log.WithError(err).Fatal("Failed to create Kubernetes clientset")
+	}
+	o.kubernetesClient = kubernetesClient
+
+	// Get Hive client
+	hiveClient, err := hiveclient.NewForConfig(cfg)
+	if err != nil {
+		log.WithError(err).Fatal("Failed to create Hive clientset")
+	}
+	o.hiveClient = hiveClient
```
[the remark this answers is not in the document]
fxierh: Okay. The code is updated.

contrib/pkg/awsprivatelink/enable.go
fxierh: Fixed.

contrib/pkg/awsprivatelink/enable.go
fxierh: Yeah, thanks for pointing that out.

contrib/pkg/utils/aws/aws.go
fxierh: Okay. The code is now updated.

fxierh:
@2uasimojo Thanks for the partial review. I made some changes. The PR is again ready for review.

contrib/pkg/awsprivatelink/enable.go
2uasimojo: I was thinking somewhere between 2 and 3: populate each field iff it is unset. That may be more work than it's worth though.

How about we leave it as is, but beef up the failure message:
```suggestion
	if hiveConfig.Spec.AWSPrivateLink != nil {
		log.Fatal("AWS Private Link is already enabled. If a previous configuration attempt did not complete, you can clean up via `hiveutil awsprivatelink disable` and try again.")
	}
```
As long as `disable` is thorough and idempotent, this advice should be sound, yes?

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,219 @@
...
+			},
+		},
+	})
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get VPC of the active cluster")
+	}
+	if len(describeVPCsOutput.Vpcs) == 0 {
+		log.Fatal("VPC of the active cluster not found")
+	}
+	vpcID := *describeVPCsOutput.Vpcs[0].VpcId
```
2uasimojo, line 141: It should be impossible for there to be more than one entry, right?

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,219 @@
...
+	log.Debugf("Found VPC ID = %v for the active cluster", vpcID)
+
+	// Create Hub account credentials Secret
+	hiveNS := hiveConfig.Spec.TargetNamespace
+	if hiveNS == "" {
+		hiveNS = constants.DefaultHiveNamespace
+	}
+	hubCredentialsSecret, err := o.generateAWSCredentialsSecret(hiveNS)
```
2uasimojo, line 149: What do you think about having a `defer` that deletes the secret if something fails beyond this point?

Or at least that prints a message suggesting that the user may wish to clean it up themselves?

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		log.WithError(err).Fatal("Failed to generate Secret with AWS credentials")
+	}
+	if err = o.dynamicClient.Create(context.Background(), hubCredentialsSecret); err != nil {
+		log.WithError(err).Fatal("Failed to create Secret with hub account credentials")
+	}
+	log.Info("Secret with hub account credentials created")
```
2uasimojo, line 156: nit: Let's have this print the ns/name of the secret.

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,219 @@
...
+	// Update HiveConfig
+	hiveConfig.Spec.AWSPrivateLink = &hivev1.AWSPrivateLinkConfig{}
+	hiveConfig.Spec.AWSPrivateLink.AssociatedVPCs = append(
+		hiveConfig.Spec.AWSPrivateLink.AssociatedVPCs,
+		hivev1.AWSAssociatedVPC{
+			AWSPrivateLinkVPC: hivev1.AWSPrivateLinkVPC{VPCID: vpcID, Region: o.region},
+		},
+	)
+	hiveConfig.Spec.AWSPrivateLink.CredentialsSecretRef = corev1.LocalObjectReference{Name: hubCredentialsSecret.Name}
+	hiveConfig.Spec.AWSPrivateLink.DNSRecordType = hivev1.AWSPrivateLinkDNSRecordType(o.dnsRecordType)
```
2uasimojo, line 167: nit: Could inline this struct spec rather than piecing it together. (Some linters will flag this -- though we probably aren't enforcing that rule.)
```suggestion
	hiveConfig.Spec.AWSPrivateLink = &hivev1.AWSPrivateLinkConfig{
		AssociatedVPCs: []hivev1.AWSAssociatedVPC{
			{
				AWSPrivateLinkVPC: hivev1.AWSPrivateLinkVPC{VPCID: vpcID, Region: o.region},
			},
		},
		CredentialsSecretRef: corev1.LocalObjectReference{Name: hubCredentialsSecret.Name},
		DNSRecordType:        hivev1.AWSPrivateLinkDNSRecordType(o.dnsRecordType),
	}
```

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,219 @@
...
+		return
+	}
+	if infrastructure.Status.PlatformStatus.AWS == nil {
+		err = errors.New("Infrastructure.status.platformStatus.aws is empty")
+		return
+	}
+
+	region = infrastructure.Status.PlatformStatus.AWS.Region
+	infraId = infrastructure.Status.InfrastructureName
+	return
```
2uasimojo, line 192: Curious why you're using implicit return value style for this one func.

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: corev1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("awsprivatelink-hub-acct-creds-%s", uuid.New().String()[:5]),
```
2uasimojo, line 208: Okay, so I think this is the crux of the "multiple secrets" issue. Why shouldn't we just have the name hardcoded and the same every time? That would make a few things easier:
- We would not risk proliferation of redundant secrets across failed `enable` attempts.
- We could do without the labeling. (Though I guess we could add the label as an extra layer of safety for the deletion, so in the case where the secret already existed before `enable`, we wouldn't delete it via `disable`.)
- If the secret `Create()` errors with `IsAlreadyExists`, we could proceed rather than failing. This would give us the added flexibility of using an existing secret -- possibly created by someone with more powerful credentials.

Thoughts?

contrib/pkg/utils/aws/aws.go
```diff
@@ -1,16 +1,227 @@
...
+// GetAssociatedVpcsWithClient extracts the associated VPCs from HiveConfig and returns a slice of AssociatedVpcWithClient.
+func GetAssociatedVpcsWithClient(hiveConfig *hivev1.HiveConfig) (associatedVpcsWithClient []*AssociatedVpcWithClient, err error) {
+	if hiveConfig.Spec.AWSPrivateLink == nil {
+		err = errors.New("AWS PrivateLink is not enabled in HiveConfig. ")
+		return
+	}
+
+	var awsClients awsclient.Client
+	for _, associatedVpc := range hiveConfig.Spec.AWSPrivateLink.AssociatedVPCs {
+		awsClients, err = awsclient.NewClientFromSecret(nil, associatedVpc.AWSPrivateLinkVPC.Region)
```
2uasimojo, line 81: Is the common case for each VPC to be in a different region? If it's possible to hit this line with the same region multiple times, I think we would be better off reusing the existing client for that region. Maybe use a local map to cache clients by region?

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,466 @@
...
+	// Get HiveConfig
+	hiveConfig := &hivev1.HiveConfig{}
+	if err = o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to get HiveConfig/hive")
+	}
+
+	// Get associated VPCs and a set of AWS clients for each of them
+	associatedVpcsWithClients, err := awsutils.GetAssociatedVpcsWithClient(hiveConfig)
+	if err != nil {
+		log.WithError(err).Fatal("Failed to retrieve associated VPCs and get AWS clients each of them")
```
2uasimojo, line 92:
```suggestion
		log.WithError(err).Fatal("Failed to retrieve associated VPCs and get AWS clients for them")
```

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+	if len(associatedVpcsWithClients) == 0 {
+		log.Warn("HiveConfig/hive does not specify any associated VPC. " +
+			"The endpoint VPC passed in as argument will still be added to HiveConfig, " +
+			"and yet no cloud resources will be set up for networking.")
+	}
+	o.associatedVpcsWithClients = associatedVpcsWithClients
+
+	// Get AWS clients for the endpoint VPC
+	awsClients, err := awsclient.NewClientFromSecret(nil, o.endpointVpcRegion)
```
2uasimojo, line 102: As below, this may end up being a redundant client if we already made one for this region.

Starting to think we should have `o.awsClientsByRegion[string]awsclient.Client` decoupled from the list of VPCs. Then if you like, you can have a little helper func like

```go
func (o *endpointVPCAddOptions) awsClientForVPC(vpc *hivev1.AWSPrivateLinkVPC) awsclient.Client {
	ret, ok := o.awsClientsByRegion[vpc.Region]
	if !ok {
		log.Fatalf("No AWS client for region %s", vpc.Region)
	}
	return ret
}
```

2uasimojo (commented):
Another partial review, still working my way through :)

contrib/pkg/awsprivatelink/enable.go
fxierh: I should have sticked to the same style. Fixed.

contrib/pkg/awsprivatelink/enable.go
fxierh: Okay, fixed.

contrib/pkg/awsprivatelink/enable.go
fxierh: Sound great. I'm going with what you suggested here. 

For disable, only secrets with the label will be deleted. So the ones configured manually will persist.

contrib/pkg/awsprivatelink/endpointvpc/add.go
fxierh: Yep. Fixed.

contrib/pkg/awsprivatelink/disable.go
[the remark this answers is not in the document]
fxierh: Fixed.

contrib/pkg/awsprivatelink/disable.go
fxierh: I've made some changes to follow your suggestion below, i.e. to generate a secret with fixed name each time, proceed if this secret already exists. Only secrets generated with the enable subcommand has the privatelink label, and only the ones with this label are deleted when calling the disable subcommand.

contrib/pkg/awsprivatelink/enable.go
fxierh: Yep. Sounds reasonable to me.

contrib/pkg/awsprivatelink/enable.go
fxierh: I've made some changes to the code. As per the current implementation, users may clean up the secret generated by the enable subcommand by calling "hiveutil awsprivatelink disable".

contrib/pkg/awsprivatelink/enable.go
fxierh: I just switched to filtering using the "kubernetes.io/cluster/INFRA-ID" key for consistency (that's mostly what I'm doing elsewhere). 

Regardless of the method of filtering, technically there could be multiple entries, as one may tag a random VPC with the same key (or give another VPC the same name, which is allowed even in the same region). That said, this is highly unlikely as I could not think of any reason for somebody to actually do this.

contrib/pkg/awsprivatelink/endpointvpc/add.go
fxierh: Thanks for pointing that out. This is now implemented as follows (for both add/remove subcommands):

- First get a map of regions where we need AWS clients. This includes regions of the associated VPCs and the endpoint VPCs. 
- Then get a set of AWS clients for each region above.

contrib/pkg/utils/aws/aws.go
fxierh: I guess it is pretty common for an associate VPC or an endpoint VPC to be in the same region as another associated VPC or endpoint VPC. So yeah, it would be great if we could re-use the same client for each region.

fxierh: Now implemented.

fxierh:
@2uasimojo Thanks for the comments. I've made some updates based on your suggestions. I guess we are good to go for another round of review :grin:.

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+}
+
+// Does not error out even if the peering connection is already established between the two VPCs.
+func setupVpcPeeringConnection(
+	associatedVpcClients, endpointVpcClients awsclient.Client,
+	associatedVpcId, endpointVpcId, endpointVpcRegion *string,
+) (*ec2.AcceptVpcPeeringConnectionOutput, error) {
+	log.Info("Setting up VPC peering connection between the associated VPC and the endpoint VPC")
+
+	createVpcPeeringConnectionOutput, err := associatedVpcClients.CreateVpcPeeringConnection(&ec2.CreateVpcPeeringConnectionInput{
```
2uasimojo, line 442: Is this idempotent? I.e. succeeds if the connection already exists?

[Later] Looks like it's an async call whose error is only non-nil if, like, the network bit fails. So :+1:

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+	o.addEndpointVpcToHiveConfig()
+
+	return nil
+}
+
+func (o *endpointVPCAddOptions) addEndpointVpcToHiveConfig() {
+	log.Infof("Adding endpoint VPC %v to HiveConfig", o.endpointVpcId)
+
+	// Get HiveConfig
+	hiveConfig := &hivev1.HiveConfig{}
```
2uasimojo, line 284: We're getting the HiveConfig in `Complete()` -- can we just add it to `o` there and reuse it here?

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+		log.WithError(err).Fatal("Failed to create controller-runtime client")
+	}
+	o.dynamicClient = dynamicClient
+
+	// Get HiveConfig
+	hiveConfig := &hivev1.HiveConfig{}
+	if err = o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to get HiveConfig/hive")
+	}
+	o.associatedVpcs = hiveConfig.Spec.AWSPrivateLink.AssociatedVPCs
```
2uasimojo, line 89: We should validate that `hiveConfig.Spec.AWSPrivateLink` is not nil here.

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+
+func (o *endpointVPCAddOptions) addEndpointVpcToHiveConfig() {
+	log.Infof("Adding endpoint VPC %v to HiveConfig", o.endpointVpcId)
+
+	// Get HiveConfig
+	hiveConfig := &hivev1.HiveConfig{}
+	if err := o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "hive"}, hiveConfig); err != nil {
+		log.WithError(err).Fatal("Failed to get HiveConfig/hive")
+	}
+	if hiveConfig.Spec.AWSPrivateLink == nil {
```
2uasimojo, line 288: We should move this check to [Complete()](https://github.com/openshift/hive/pull/2045/files#diff-9b480d80f76313c459d891f57a1aee57f8b98f610fa503f461ebc239f88e34bdR89).

contrib/pkg/utils/aws/aws.go
```diff
@@ -1,16 +1,216 @@
...
+
+	for i, endpointVpc := range inventory {
+		if vpcId == endpointVpc.AWSPrivateLinkVPC.VPCID {
+			endpointVpcIdx = i
+			found = true
+			break
+		}
+	}
+
+	return endpointVpcIdx, found
```
2uasimojo, line 58: or
```suggestion
	for i, endpointVpc := range inventory {
		if vpcId == endpointVpc.AWSPrivateLinkVPC.VPCID {
			return i, true
		}
	}
	return -1, false
```

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+		)
+		log.Debugf("Endpoint VPC added to HiveConfig")
+	}
+
+	if err := o.dynamicClient.Update(context.Background(), hiveConfig); err != nil {
+		log.WithError(err).Error("Failed to update HiveConfig/hive")
+	}
+}
+
+func addRouteToRouteTablesOfEndpointSubnets(
```
2uasimojo, line 346: At a glance, this and the below look identical except for the `Filters`. Can we DRY them?

2uasimojo (commented):
Another batch -- I got through add.go :)

contrib/pkg/awsprivatelink/enable.go
fxierh: Yep. I added an IF statement s.t. the executable errors out in case multiple VPCs are found.

contrib/pkg/awsprivatelink/disable.go
```diff
@@ -0,0 +1,112 @@
...
+			"one by one via `hiveutil awsprivatelink endpointvpc remove ...`.")
+	}
+
+	// Delete Hub account secret(s) if present
+	hiveNS := operatorutils.GetHiveNamespace(hiveConfig)
+	hubAcctSecrets := &corev1.SecretList{}
+	if err := o.dynamicClient.List(
+		context.Background(),
+		hubAcctSecrets,
+		client.MatchingLabels{awsutils.PrivateLinkHubAcctCredsLabel: "true"},
```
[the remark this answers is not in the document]
fxierh: Yep, we could. This is more accurate after all. I added an additional MatchingFields to match the Secret's name.

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	log.Info("HiveConfig updated")
+
+	return nil
+}
+
+func (o *enableOptions) getRegionAndInfraId() (string, string, error) {
+	infrastructure := &configv1.Infrastructure{}
+	if err := o.dynamicClient.Get(context.Background(), types.NamespacedName{Name: "cluster"}, infrastructure); err != nil {
+		return "", "", err
```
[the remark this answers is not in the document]
fxierh: I'm ok with returning empty strings but thanks for pointing that out.

contrib/pkg/awsprivatelink/enable.go
```diff
@@ -0,0 +1,217 @@
...
+		return "", "", errors.New("Infrastructure.status.platformStatus is empty")
+	}
+	if infrastructure.Status.PlatformStatus.AWS == nil {
+		return "", "", errors.New("Infrastructure.status.platformStatus.aws is empty")
+	}
+
+	return infrastructure.Status.PlatformStatus.AWS.Region, infrastructure.Status.InfrastructureName, nil
+}
+
+// Adapted from contrib/pkg/adm/managedns/enable.go:generateAWSCredentialsSecret()
```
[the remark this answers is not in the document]
fxierh: The function I'm referencing is different in some ways, e.g. the Secret generated has a randomized name, and no label. It also has a different receiver (it is a method of a different type). So I guess it could be complicated to use it.

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+	cmd := &cobra.Command{
+		Use:   "add vpc-id",
+		Short: "Networking setup between an endpoint VPC and each associated VPC",
+		Long: `Networking setup between an endpoint VPC and each associated VPC 
+specified in HiveConfig.spec.awsPrivateLink.associatedVPCs: 
+1) Establish VPC peering connection between the endpoint VPC and the associated VPC
+2) Add a route (to the peered VPC) in relevant route tables of the associated VPC and the endpoint VPC 
+3) Update relevant security groups of the associated VPC and the endpoint VPC to allow traffic between them
+4) Add the endpoint VPC to HiveConfig.spec.awsPrivateLink.endpointVPCInventory`,
+		Args: cobra.ExactArgs(1),
```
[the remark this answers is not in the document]
fxierh: Yep, each endpoint VPC has its own set of subnets, and usually the subnet IDs of different VPCs do not collide.

contrib/pkg/awsprivatelink/endpointvpc/add.go
fxierh: Yep. Fixed. 

I also made some changes to remove.go to reuse HiveConfig as well as endpointVpcIdx (index of the endpoint VPC to remove in the list of endpoint VPC inventory).

contrib/pkg/awsprivatelink/endpointvpc/add.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+	o.associatedVpcs = hiveConfig.Spec.AWSPrivateLink.AssociatedVPCs
+	if len(o.associatedVpcs) == 0 {
+		log.Warn("HiveConfig/hive does not specify any associated VPC. " +
+			"The endpoint VPC passed in as argument will still be added to HiveConfig, " +
+			"and yet no cloud resources will be set up for networking.")
+	}
+
+	// Get AWS clients by region
+	regions := make(map[string]bool)
```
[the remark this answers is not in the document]
fxierh: Done. 

(Looks like a Set has lower memory consumption. )

contrib/pkg/awsprivatelink/endpointvpc/add.go
fxierh: You're right, that's a pointer field. Fixed.

contrib/pkg/awsprivatelink/endpointvpc/add.go
fxierh: Done.

contrib/pkg/utils/aws/aws.go
fxierh: Go's implicit return style was used here. I switched to the explicit style later but forgot to clean this function up ... 

Fixed.

contrib/pkg/awsprivatelink/endpointvpc/add.go
fxierh: Yep. I combined the addRouteToRouteTables* functions into one. 

I also did the same for remove.go.

contrib/pkg/awsprivatelink/endpointvpc/remove.go
```diff
@@ -0,0 +1,353 @@
...
+	o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory[o.endpointVpcIdx] =
+		o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory[len(o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory)-1]
+	o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory =
+		o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory[:len(o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory)-1]
+	if err := o.dynamicClient.Update(context.Background(), &o.hiveConfig); err != nil {
+		log.WithError(err).Error("Failed to update HiveConfig")
+	}
+}
+
+func deleteVpcPeeringConnection(awsClients awsclient.Client, VpcId1, VpcId2 *string) (*string, error) {
```
2uasimojo, line 269: I noticed the first return value is not used, might as well get rid of it.

contrib/pkg/awsprivatelink/endpointvpc/remove.go
```diff
@@ -0,0 +1,353 @@
...
+func (o *endpointVPCRemoveOptions) removeEndpointVpcFromHiveConfig() {
+	log.Infof("Removing endpoint VPC %v from HiveConfig", o.endpointVpcId)
+
+	// Remove endpoint VPC from HiveConfig if necessary
+	o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory[o.endpointVpcIdx] =
+		o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory[len(o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory)-1]
+	o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory =
+		o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory[:len(o.hiveConfig.Spec.AWSPrivateLink.EndpointVPCInventory)-1]
+	if err := o.dynamicClient.Update(context.Background(), &o.hiveConfig); err != nil {
+		log.WithError(err).Error("Failed to update HiveConfig")
```
2uasimojo, line 265: Should this be fatal? Or bubble up so the caller can decide if it's fatal?

2uasimojo (commented):
Finally got to the end!

Only a couple small things left inline.

Also, I would like to see our docs updated to reflect this change. Two candidates:
- The top of [awsprivatelink.md](https://github.com/openshift/hive/blob/master/docs/awsprivatelink.md) could mention it. It should come with a caveat that hiveutil is not technically supported for general use, is provided as-is as a convenience, blah blah blah. This could just link to...
- [hiveutil.md](https://github.com/openshift/hive/blob/master/docs/hiveutil.md) could have a section with a summary and example(s). (Note that this already has a disclaimer like the one I suggested above.)

contrib/pkg/awsprivatelink/endpointvpc/remove.go
fxierh: Yep, fixed.

contrib/pkg/awsprivatelink/endpointvpc/remove.go
fxierh: Fixed: made HiveConfig update failure fatal in both add and remove.

contrib/pkg/awsprivatelink/enable.go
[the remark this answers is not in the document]
fxierh: Thanks for taking the time to look into this. Would be great to refactor in future PRs.

docs/hiveutil.md
```diff
@@ -151,6 +151,84 @@ Claim a ClusterDeployment from a [ClusterPool](./clusterpools.md):
...
+# Could be different from the active cluster's region
+export REGION=us-east-1
+export STACK=my-stack-name
+export TEMPLATE=file://hack/awsprivatelink/vpc.cf.yaml
+export AZCOUNT=3
+# Should not overlap with the CIDR of the active cluster's VPC
+export CIDR=10.1.0.0/16
+
+aws --region $REGION cloudformation create-stack --stack-name $STACK --template-body $TEMPLATE --parameters ParameterKey=AvailabilityZoneCount,ParameterValue=$AZCOUNT ParameterKey=VpcCidr,ParameterValue=$CIDR
+aws --region $REGION cloudformation wait stack-create-complete --stack-name $STACK
```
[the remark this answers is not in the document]
fxierh: Yeah, maybe something like `hiveutil awsprivatelink endpointvpc create` ?
(or some shell script in ./hack/ ?)

fxierh:
@2uasimojo I encountered something bizarre when doing some pre-merge testing. It does not seem to be related to Hive or this PR but further investigation is needed.

docs/hiveutil.md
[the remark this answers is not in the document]
2uasimojo: Sure, whatever you think best. It just seems like you did all this work so the user wouldn't have to run `aws` commands -- might as well make that complete.

2uasimojo:
> I encountered something bizarre when doing some pre-merge testing.

Okay:

/hold

You can unhold if that resolves and no updates are needed.

/lgtm

This is really great work @fxierh. Thank you for this contribution to hive!

fxierh:
@2uasimojo That's my pleasure. Thank you for reviewing this PR. It has been a rewarding experience for me.

fxierh:
The thing I encountered yesterday was that the Openshift installer can create an excessive amount (> 2 * number of nodes) of subnets when provisioning in us-east-1 (which has more AZ than others). 

I was testing to make sure that `hiveutil awsprivatelink` can cope with the situation when the Hive cluster is located in us-east-1.

PR #2048, opened 2023-06-14
Merge Labels/Taints of MachineSets

suhanime:
When MachineSets were generated, we overrode the existing labels and taints on the remoteMachineSet to reflect what was present on the MachinePool. This commit preserves the existing labels and taints on the MachineSet, and if they clash with those on MachinePool, they will be overwritten.

There was also a bug where labels and taints from generated MachineSets were not carried over to the remote MachineSet if the label map was empty - this commit also fixes that.

xref: https://issues.redhat.com/browse/HIVE-2035

/assign @abutcher

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -582,19 +582,39 @@ func (r *ReconcileMachinePool) syncMachineSets(
 					}
 				}
 
-				// Update if the labels on the remote machineset are different than the labels on the generated machineset.
-				// If the length of both labels is zero, then they match, even if one is a nil map and the other is an empty map.
-				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; (len(rl) != 0 || len(l) != 0) && !reflect.DeepEqual(rl, l) {
-					msLog.WithField("desired", l).WithField("observed", rl).Info("labels out of sync")
-					rMS.Spec.Template.Spec.Labels = l
-					objectModified = true
+				// Carry over the labels on the remote machineset from the generated machineset. Merging strategy preserves the unique labels of remote machineset.
+				if rMS.Spec.Template.Spec.Labels == nil {
+					rMS.Spec.Template.Spec.Labels = make(map[string]string)
+				}
+				for key, value := range ms.Spec.Template.Spec.Labels {
+					if rMS.Spec.Template.Spec.Labels[key] != value {
+						rMS.Spec.Template.Spec.Labels[key] = value
+						objectModified = true
+					}
 				}
 
-				// Update if the taints on the remote machineset are different than the taints on the generated machineset.
-				// If the length of both taints is zero, then they match, even if one is a nil slice and the other is an empty slice.
-				if rt, t := rMS.Spec.Template.Spec.Taints, ms.Spec.Template.Spec.Taints; (len(rt) != 0 || len(t) != 0) && !reflect.DeepEqual(rt, t) {
-					msLog.WithField("desired", t).WithField("observed", rt).Info("taints out of sync")
-					rMS.Spec.Template.Spec.Taints = t
+				// Carry over the taints on the remote machineset from the generated machineset. Merging strategy preserves the unique taints of remote machineset.
+				if rMS.Spec.Template.Spec.Taints == nil {
+					rMS.Spec.Template.Spec.Taints = []corev1.Taint{}
+				}
+				// Make a temporary map of generated MachineSet's taints for easy lookup
+				gTaintsByKey := make(map[string]*corev1.Taint)
+				for gIndex, gTaint := range ms.Spec.Template.Spec.Taints {
+					gTaintsByKey[gTaint.Key] = &ms.Spec.Template.Spec.Taints[gIndex]
+				}
+				// Go through the remote MachineSet's taints, replacing overlaps with the generated MachineSet
+				for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+					if gTaint, foundTaint := gTaintsByKey[rTaint.Key]; foundTaint {
+						if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
+							rMS.Spec.Template.Spec.Taints[rIndex] = *gTaint
+							objectModified = true
+						}
+						delete(gTaintsByKey, rTaint.Key)
+					}
+				}
+				// Any remaining taints from the temporary map are new and need to be added
+				for _, gTaint := range gTaintsByKey {
+					rMS.Spec.Template.Spec.Taints = append(rMS.Spec.Template.Spec.Taints, *gTaint)
 					objectModified = true
 				}
 
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -703,7 +703,7 @@ openstack:
 
 A MachinePool for the worker machinesets is not required. If the user creates a MachinePool for the worker MachineSets, then Hive will manage the worker MachineSets.
 
-MachinePool reconciliation is limited to updating MachineSet replicas to match the replicas configured for the MachinePool. Additionally, When `Labels` or `Taints` are configured for a MachinePool, Hive will completely override any existing `Label` or `Taints` configurations on the associated MachineSets, see [HIVE-2035](https://issues.redhat.com/browse/HIVE-2035).
+MachinePool reconciliation is limited to updating MachineSet replicas to match the replicas configured for the MachinePool. Additionally, any existing `Labels` or `Taints` on the MachineSets will be overridden if they clash with those on the MachinePool.
 
 MachinePool platform is immutable and any changes made to `MachinePool.spec.platform` are blocked by a validating webhook. The Machine Config Operator does not support updating existing machines when platform details are changed in a MachineSet and consequently Hive does not support making such changes to MachinePool platform, see [HIVE-2024](https://issues.redhat.com/browse/HIVE-2024).
 
--- a/pkg/controller/machinepool/machinepool_controller_test.go
+++ b/pkg/controller/machinepool/machinepool_controller_test.go
@@ -275,6 +275,145 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 			},
 		},
 		{
+			name:              "Merge labels and taints",
+			clusterDeployment: testClusterDeployment(),
+			machinePool: func() *hivev1.MachinePool {
+				mp := testMachinePool()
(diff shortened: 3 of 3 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -582,20 +582,36 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+							if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
+								rMS.Spec.Template.Spec.Taints[rIndex] = gTaint
+								objectModified = true
+							}
+						}
+					}
+					if !foundTaint {
+						rMS.Spec.Template.Spec.Taints = append(rMS.Spec.Template.Spec.Taints, gTaint)
+						objectModified = true
+					}
```
suhanime, line 614: This piece of code is messy but I couldn't think of another way to compare structs - especially for the case where remote has 1 (or more) taints - most struct comparisons are index-based. Do suggest if there's an alternative

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -582,20 +582,36 @@ func (r *ReconcileMachinePool) syncMachineSets(
 					}
 				}
 
-				// Update if the labels on the remote machineset are different than the labels on the generated machineset.
-				// If the length of both labels is zero, then they match, even if one is a nil map and the other is an empty map.
-				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; (len(rl) != 0 || len(l) != 0) && !reflect.DeepEqual(rl, l) {
-					msLog.WithField("desired", l).WithField("observed", rl).Info("labels out of sync")
```
suhanime, line 588: I had to get rid of this log line because I removed the deepequal comparison, but if an update is called - there would be a log entry for that.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -582,20 +582,36 @@ func (r *ReconcileMachinePool) syncMachineSets(
 					}
 				}
 
-				// Update if the labels on the remote machineset are different than the labels on the generated machineset.
-				// If the length of both labels is zero, then they match, even if one is a nil map and the other is an empty map.
-				if rl, l := rMS.Spec.Template.Spec.Labels, ms.Spec.Template.Spec.Labels; (len(rl) != 0 || len(l) != 0) && !reflect.DeepEqual(rl, l) {
```
suhanime, line 587: PS: reflect.DeepEqual was removed for both cases because it is not an effective comparison - particularly for already existing labels/taints on remote machines if they need to be preserved

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -582,20 +582,36 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+				}
+				for _, gTaint := range ms.Spec.Template.Spec.Taints {
+					foundTaint := false
+					for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+						if gTaint.Key == rTaint.Key {
+							foundTaint = true
+							if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
+								rMS.Spec.Template.Spec.Taints[rIndex] = gTaint
+								objectModified = true
+							}
```
2uasimojo, line 608:
```suggestion
							}
							break
```
No sense continuing to iterate, right?

pkg/controller/machinepool/machinepool_controller.go
2uasimojo: I have an alternative. It's mathematically more efficient, though in real life the difference will be negligible unless there's just a whole lotta taints to process. It may or may not end up being easier to read/understand/maintain.
1. Loop over `ms.Spec.Template.Spec.Taints` and construct a `map[string]Taint` keyed by each `Taint.Key`.
2. Loop over `rMS.Spec.Template.Spec.Taints`. On each iteration, look up the `Taint.Key` in the above map. If found, do your comparison, replace if necessary, and remove that entry from the map (regardless of whether you replaced or not).
3. At the end, append any remaining entries from the map.

pkg/controller/machinepool/machinepool_controller.go
suhanime: :doh missed that! Fixed

pkg/controller/machinepool/machinepool_controller.go
suhanime: So here's the problem - maps are great because you don't need to iterate over them in orderly fashion - but you can't really get index of entries - and the only way to loop through structs is in an orderly fashion via indices. So firstly, I wouldn't be able to replace it directly from the struct of remote taints in step 2 - as you suggested, and the alternative for me is to -
1.  Make a map of taint name to taint as per your step 1
2. Loop over the generated machineset's taints and update the values in the temporary map that we built in step 1
3. Get a map of keys from temporary map and sort them to rely on a sorted order (because range of map is not guaranteed to be in the same order) and then another loop to reassign values to remote MachineSet

My step 3 might not be the most optimal or well thought-out, regardless - it is still just as difficult to understand code-wise. So I've kept my current implementation...

2uasimojo: > but you can't really get index of entries

In this case, we're relying on there being no duplicates of .Key in either list, right? So .Key works as an index for the map.

> I wouldn't be able to replace it directly from the struct of remote taints in step 2

Eh? I don't understand why not.

> sort them to rely on a sorted order (because range of map is not guaranteed to be in the same order)

Why does order matter? `rMS` will stay in its original order. If there are taints left over from `ms` they'll be appended in a random and nondeterministic order, but that's okay, isn't it?

(I'm not pushing for my suggested impl, just arguing that it ought to work as I proposed it.)

2uasimojo: This delta still passes UT (though I didn't scrutinize the test suite):

```diff
diff --git a/pkg/controller/machinepool/machinepool_controller.go b/pkg/controller/machinepool/machinepool_controller.go
index 6cbef1cb8..336a7a8f7 100644
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -594,25 +594,30 @@ func (r *ReconcileMachinePool) syncMachineSets(
                                }
 
                                // Carry over the taints on the remote machineset from the generated machineset. Merging strategy preserves the unique taints of remote machineset.
+                               // First make a map of the generated MachineSet's taints:
+                               gTaintsByKey := make(map[string]*corev1.Taint)
+                               for i, gTaint := range ms.Spec.Template.Spec.Taints {
+                                       gTaintsByKey[gTaint.Key] = &ms.Spec.Template.Spec.Taints[i]
+                               }
+                               // In case the remote MachineSet's taints are empty...
                                if rMS.Spec.Template.Spec.Taints == nil {
                                        rMS.Spec.Template.Spec.Taints = []corev1.Taint{}
                                }
-                               for _, gTaint := range ms.Spec.Template.Spec.Taints {
-                                       foundTaint := false
-                                       for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
-                                               if gTaint.Key == rTaint.Key {
-                                                       foundTaint = true
-                                                       if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
-                                                               rMS.Spec.Template.Spec.Taints[rIndex] = gTaint
-                                                               objectModified = true
-                                                               break
-                                                       }
+                               // Go through the remote MachineSet's taints, replacing overlaps with the generated MachineSet
+                               // (i.e. same key but different contents).
+                               for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+                                       if gTaint, found := gTaintsByKey[rTaint.Key]; found {
+                                               if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
+                                                       rMS.Spec.Template.Spec.Taints[rIndex] = *gTaint
+                                                       objectModified = true
                                                }
+                                               delete(gTaintsByKey, rTaint.Key)
                                        }
-                                       if !foundTaint {
-                                               rMS.Spec.Template.Spec.Taints = append(rMS.Spec.Template.Spec.Taints, gTaint)
-                                               objectModified = true
-                                       }
+                               }
+                               // Any remaining taints from gTaintsByKey are new and need to be added
+                               for _, gTaint := range gTaintsByKey {
+                                       rMS.Spec.Template.Spec.Taints = append(rMS.Spec.Template.Spec.Taints, *gTaint)
+                                       objectModified = true
                                }
 
                                // Platform updates will be blocked by webhook, unless they're not.
```

suhanime: So I misunderstood and switched out the order of the structs in steps 1 and 2 - it does work and works well, and your suggestion is slightly more intuitive, so I incorporated that instead - thank you!

abutcher:
Adding a label to the worker MachinePool resulted in labels being added to the worker MachineSet where the MachineSet previously had no `spec.template.spec.metadata.labels`. Likewise, adding taints are synced :heavy_check_mark: 

```
kind: MachinePool
spec:
  clusterDeploymentRef:
    name: abutcher-test
  labels:
    testlabel1: testvalue1
  taints:
  - effect: NoSchedule
    key: reserved
    value: gpu-node
  name: worker
  platform:
    aws:
      rootVolume:
        size: 22
        type: gp2
      type: m4.xlarge
  replicas: 3
```

```
➜  ~ oc get machineset -n openshift-machine-api abutcher-test-lflmp-worker-us-east-1a -o jsonpath --template='{ .spec.template.spec.metadata.labels }'
{"testlabel1":"testvalue1"}
```

```
➜  ~ oc get machineset -n openshift-machine-api abutcher-test-lflmp-worker-us-east-1a -o jsonpath --template='{ .spec.template.spec.taints }'
[{"effect":"NoSchedule","key":"reserved","value":"gpu-node"}]
```

Adding a label to the MachineSet itself doesn't get overwritten by the labels on the MachinePool.

```
➜  ~ oc get machineset -n openshift-machine-api abutcher-test-lflmp-worker-us-east-1a -o jsonpath --template='{ .spec.template.spec.metadata.labels }'
{"existinglabel1":"existinglabel2","testlabel1":"testvalue1"}
```

Adding a taint to the MachineSet itself doesn't get overwritten by the taints on the MachinePool.

```
➜  ~ oc get machineset -n openshift-machine-api abutcher-test-lflmp-worker-us-east-1a -o jsonpath --template='{ .spec.template.spec.taints }'
[{"effect":"NoSchedule","key":"reserved","value":"gpu-node"},{"effect":"NoSchedule","key":"reserved","value":"other-node"}]
```

/lgtm

PR #2049, opened 2023-06-15
Enable Hive to delete the hive-additional-ca Secret

fxierh:
Hard-code the previously-unpopulated .apiVersion and .kind fields so that Secret/hive-additional-ca can be deleted by controller. 

See https://issues.redhat.com/browse/HIVE-2248


2uasimojo (commented):
/lgtm

PR #2050, opened 2023-06-15
Skip hive-trusted-cabundle ConfigMap creation on deleted ns

2uasimojo:
Deleting a namespace containing a ClusterDeployment and its associated artifacts results in an impossible situation: the deletion of the ClusterDeployment wants to trigger a deprovision, but we cannot create the necessary artifacts (e.g. ClusterDeprovision CR, deprovision Pod) in a namespace that is terminating. We have logic to skip deprovisioning and simply leak the cloud resources in this case.

...but we forgot to incorporate this logic for the hive-trusted-cabundle ConfigMap (introduced by https://github.com/2uasimojo/hive/commit/253408ed8ecdab5a7f3510a7f573db51008d6486 / https://github.com/openshift/hive/pull/1999), and would hot-loop attempting and failing to create it, never reaching the aforementioned "give up" logic, resulting in wedging the namespace. This commit adds the logic, resulting in the same behavior as before.

NOTE: It is still not supported / a good idea to delete a namespace containing extant ClusterDeployments, as this leaks cloud resources as described above.

[HIVE-2249](https://issues.redhat.com//browse/HIVE-2249)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -523,7 +523,7 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		}
 	}
 
-	if err := r.ensureTrustedCABundleConfigMap(cd.Namespace); err != nil {
+	if err := r.ensureTrustedCABundleConfigMap(cd, cdLog); err != nil {
 		cdLog.WithError(err).Error("Failed to create or verify the trusted CA bundle ConfigMap")
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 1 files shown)


2uasimojo:
/hold

While this still may be worth doing, I now believe it is based on a faulty premise: If the namespace is terminating, we won't be able to create the deprovision artifacts either. Before this bug was injected, we would just give up on the CD:

```
time="2023-06-15T21:45:41.51Z" level=error msg="error creating deprovision request" clusterDeployment=tmp3/testmp controller=clusterDeployment error="clusterdeprovisions.hive.openshift.io \"testmp\" is forbidden: unable to create new content in namespace tmp3 because it is being terminated" reconcileID=j4zbz9g4
time="2023-06-15T21:45:41.511Z" level=warning msg="detected a namespace deleted before deprovision request could be created, giving up on deprovision and removing finalizer" clusterDeployment=tmp3/testmp controller=clusterDeployment reconcileID=j4zbz9g4
```

...so the cloud resources would be leaked without any indication to the user.

2uasimojo:
Note: total redirect. If we detect the namespace is being deleted, we just skip the creation of the configmap and "give up". This is in line with what used to happen.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1222,6 +1222,14 @@ func (r *ReconcileClusterDeployment) ensureManagedDNSZoneDeleted(cd *hivev1.Clus
 	return false, nil
 }
 
+func (r *ReconcileClusterDeployment) namespaceDeleted(nsName string) (bool, error) {
```
lleshchi, line 1225: Add comment or rename to something like `namespaceMarkedForDeletion`

PR #2056, opened 2023-06-28
Change bundle-gen.py to account for new bundle format

lleshchi:
xref: https://issues.redhat.com/browse/HIVE-1658

### DIFF
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -7,6 +7,7 @@ import github as gh
 import json
 import os
 import requests
+import semver
 import shutil
 import subprocess
 import sys
@@ -25,7 +26,7 @@ HIVE_SUB_DIR = "operators/hive-operator"
 
 OPERATORHUB_HIVE_IMAGE_DEFAULT = "quay.io/app-sre/hive"
 
-COMMUNITY_OPERATORS_HIVE_PKG_URL = "https://raw.githubusercontent.com/redhat-openshift-ecosystem/community-operators-prod/main/operators/hive-operator/hive.package.yaml"
+COMMUNITY_OPERATORS_UPSTREAM_REPO = "git@github.com:redhat-openshift-ecosystem/community-operators-prod.git"
 
 SUBPROCESS_REDIRECT = subprocess.DEVNULL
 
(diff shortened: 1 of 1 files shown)


hack/bundle-gen.py
```diff
@@ -129,26 +130,59 @@ def get_params():
 
     return args
 
+# Traverse through version directories to detect highest version
+def get_previous_version(work_dir, channel_name):
+    upstream_repo = "git@github.com:redhat-openshift-ecosystem/community-operators-prod.git"
```
2uasimojo, line 135: We can const this (and get rid of [COMMUNITY_OPERATORS_HIVE_PKG_URL](https://github.com/openshift/hive/pull/2056/files#diff-625239517ac654a0a60ffdcc20e3a62f806a20df37b553a2ce9e700b57db8176R29), which this logic replaces)

hack/bundle-gen.py
```diff
@@ -129,26 +130,59 @@ def get_params():
...
-# get_previous_version grabs the previous hive version (without the leading `v`) from
-# COMMUNITY_OPERATORS_HIVE_PKG_URL package yaml for the provided channel_name.
-def get_previous_version(channel_name):
-    http = urllib3.PoolManager()
-    r = http.request("GET", COMMUNITY_OPERATORS_HIVE_PKG_URL)
-    pkg_yaml = yaml.load(r.data.decode("utf-8"), Loader=yaml.FullLoader)
+    repo_full_path = os.path.join(work_dir, dir_name)
+    # clone git repo
+    try:
+        git.Repo.clone_from(upstream_repo, repo_full_path)
```
2uasimojo, line 142: We're already cloning this repo in open_pr. Let's find a way to collapse and only do it once.

hack/bundle-gen.py
```diff
@@ -129,26 +130,59 @@ def get_params():
...
+    except:
+        print("Failed to checkout {}".format(upstream_branch))
+        raise
+
+    highest_version = "0.0.0"
+    try:
+        hive_dir = os.path.join(repo_full_path, HIVE_SUB_DIR)
+        for version in os.listdir(hive_dir):
+            if version == "ci.yaml":
+                continue
```
2uasimojo, line 166: I don't love this kind of hardcoding. How about instead...

hack/bundle-gen.py
```diff
@@ -129,26 +130,59 @@ def get_params():
...
+            annotation_yaml_path = os.path.join(hive_dir, version, "metadata", "annotations.yaml")
+            with open(annotation_yaml_path, "r") as stream:
```
2uasimojo, line 168: ...
2. wrap this in a try/except, skipping if it fails due to things like ENOENT, ENOTDIR, etc.

hack/bundle-gen.py
```diff
@@ -129,26 +130,59 @@ def get_params():
...
+                annotation_yaml = yaml.load(stream, Loader=yaml.SafeLoader)
+                version_channel = annotation_yaml["annotations"]["operators.operatorframework.io.bundle.channel.default.v1"]
+                if version_channel == channel_name:
+                    if semver.compare(version, highest_version) > 0:
+                        highest_version = version
     except:
```
2uasimojo, line 174: yeah, cause otherwise any anomaly in the file system will fall through to this `except` and we'll just fail to determine the prev version

hack/bundle-gen.py
```diff
@@ -129,26 +130,59 @@ def get_params():
...
-            COMMUNITY_OPERATORS_HIVE_PKG_URL,
+            upstream_repo,
         )
         raise
 
-    # Channel not found -- no previous version.
-    return None
+    if highest_version == "0.0.0":
+        # Channel not found -- no previous version.
+        return None
```
2uasimojo, line 183: I'm tempted to say this should be fatal. I mean, I guess we theoretically want this to work if we introduce a new channel... but I think the bigger risk is some bug resulting in not getting a prev_version, resulting in no `replaces`, and we don't notice. That'll mess up the upgrade graph, which is Bad™. I would rather have to hand-edit the CSV or temporarily comment out the error that one time we introduce a new channel.

(And please leave a comment explaining all of that.)

2uasimojo (commented):
Looking good :+1:

hack/bundle-gen.py
```diff
@@ -325,14 +389,17 @@ def open_pr(
...
     print()
-    print("Cloning %s" % fork_repo)
     repo_full_path = os.path.join(work_dir, dir_name)
-    # clone git repo
-    try:
-        git.Repo.clone_from(fork_repo, repo_full_path)
-    except:
-        print("Failed to clone repo {} to {}".format(fork_repo, repo_full_path))
-        raise
+    if not os.path.exists(repo_full_path):
```
2uasimojo, line 393: Ah, nice solution. My only quibble is that the connection between `dir_name` here and [above](https://github.com/openshift/hive/pull/2056/files#diff-625239517ac654a0a60ffdcc20e3a62f806a20df37b553a2ce9e700b57db8176R136) is a bit brittle. I'll say good enough for now, but worth adding a comment in one or both places referring to the other and explaining the connection.

2uasimojo:
/lgtm

Let's do this.

hack/bundle-gen.py
```diff
@@ -129,26 +130,66 @@ def get_params():
...
+
+    highest_version = "0.0.0"
+    try:
+        hive_dir = os.path.join(repo_full_path, HIVE_SUB_DIR)
+        for version in os.listdir(hive_dir):
+            annotation_yaml_path = os.path.join(hive_dir, version, "metadata", "annotations.yaml")
+            try:
+                with open(annotation_yaml_path, "r") as stream:
+                    annotation_yaml = yaml.load(stream, Loader=yaml.SafeLoader)
+                    version_channel = annotation_yaml["annotations"]["operators.operatorframework.io.bundle.channel.default.v1"]
```
2uasimojo, line 168: This appears to always be `alpha`. See #2063 for fix.

PR #2052, opened 2023-06-23
Revendor installer

lleshchi:
xref: https://issues.redhat.com/browse/HIVE-2144

Includes following revendoring updates:

    - github.com/openshift/installer v0.9.0-master.0.20230216120756-2659252502ba => v0.9.0-master.0.20230721132804-2c449439afd9
    - k8s.io/api v0.26.2 => v0.27.2
    - k8s.io/api v0.26.2 => v0.27.2
    - k8s.io/client-go v0.26.2 => v0.27.2
    - sigs.k8s.io/controller-runtime v0.14.2 => v0.15.0
    - google.golang.org/grpc v1.52.0 => v1.56.1
    - github.com/metal3-io/baremetal-operator v0.0.0-20220128094204-28771f489634 => v0.0.0-20230531194024-8dde0991ffdd
    - github.com/openshift/machine-api-operator v0.2.1-0.20220930152820-30825f121cc5 => v0.2.1-0.20230613002216-b15f199bf388

Update to installer includes:

    - Deprecated legacy VSphere platform fields
    The deprecated fields were included in pkg/clusterresource/vsphere.go
    and switched over to the new fields in pkg/controller/machinepool/vsphereactuator.go

    - Accommodate changes to function signatures:
        - func (a *AzureActuator) GenerateMachineSets
        - func (a *AWSActuator) validateSubnets
        - func (a *AWSActuator) getSubnetsByAvailabilityZone 

Update to controller-runtime v0.15.0 includes:
    
    - Change to the behavior of the fake client
    deletionTimestamp. The fake client will panic if initialized with an
    object that has a DeletionTimestamp and no finalizer, and will
    silently ignore a DeletionTimestamp on an object in a Create request,
    matching the behavior of the kube-apiserver. Tests have been rewritten
    to work with this new behavior.
    https://github.com/kubernetes-sigs/controller-runtime/pull/2316
    
    - Requirement to register a fake client
    WithStatusSubresource in order to successfully update the Status
    of an object. Due to this change, the helper function
    NewFakeClientBuilder() is added to pkg/test/fake to allow for easier
    setup of a fake client with the necessary sub-resources.
    https://github.com/kubernetes-sigs/controller-runtime/pull/2259
   
Other noteworthy changes:

    - New singleton scheme introduced to codebase.  
    The hive specific scheme is registered once, with all necessary
    packages, and used throughout the codebase, removing the need to specify
    unique schemes for each use case, and decreasing the overall amount of
    imports.
    
[...]

PR #2057, opened 2023-07-03
MachinePool: hack name of vSphere default worker MachineSet

2uasimojo:
With recent installer changes to accommodate zones in vSphere, the name of the default worker pool in OCP 4.13+ is now suffixed with a `-0` even when only one zone is in play. Hive's day 2 MachinePool logic needs to be able to generate the correct MachineSet name when talking to a 4.13+ cluster -- but also when talking to a 4.12- cluster. This commit adds in a special case to discover and account for this difference where it exists.

[HIVE-2254](https://issues.redhat.com//browse/HIVE-2254)

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -524,6 +524,26 @@ func (r *ReconcileMachinePool) ensureEnoughReplicas(
 	return nil, nil
 }
 
+// matchAndMutate decides whether gMS ("generated MachineSet") and rMS ("remote MachineSet" -- the one already extant
+// on the spoke cluster) are talking about the same machines. In certain edge cases (vsphere) this may be true when
+// the names don't match, in which case this func will mutate gMS to make the names match. HIVE-2254.
+func matchAndMutate(pool *hivev1.MachinePool, cd *hivev1.ClusterDeployment, gMS *machineapi.MachineSet, rMS machineapi.MachineSet) bool {
+	// HACK: For vsphere, the default worker pool may or may not have a `-0` suffix.
+	// - If the cluster was created on 4.12 or earlier, it won't.
+	// - If the cluster was created on 4.13 or later, it will.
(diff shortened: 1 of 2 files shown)


2uasimojo:
NTR: I tested this against 4.12 and 4.13 spokes, with the default worker pool as well as an additional one, with this code against hive master (currently vendoring 4.12 installer code) and on top of #2052 (vendoring recent master installer code).

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -524,6 +524,26 @@ func (r *ReconcileMachinePool) ensureEnoughReplicas(
...
 }
 
+// matchAndMutate decides whether gMS ("generated MachineSet") and rMS ("remote MachineSet" -- the one already extant
+// on the spoke cluster) are talking about the same machines. In certain edge cases (vsphere) this may be true when
+// the names don't match, in which case this func will mutate gMS to make the names match. HIVE-2254.
+func matchAndMutate(pool *hivev1.MachinePool, cd *hivev1.ClusterDeployment, gMS *machineapi.MachineSet, rMS machineapi.MachineSet) bool {
+	// HACK: For vsphere, the default worker pool may or may not have a `-0` suffix.
+	// - If the cluster was created on 4.12 or earlier, it won't.
+	// - If the cluster was created on 4.13 or later, it will.
+	// - If we are vendoring installer code from 4.12 or earlyer, our locally-generated MachineSets won't have the suffix.
```
abutcher, line 534:
```suggestion
	// - If we are vendoring installer code from 4.12 or earlier, our locally-generated MachineSets won't have the suffix.
```

abutcher (commented):
One nit on comment, looks good to me

pkg/controller/machinepool/machinepool_controller.go
2uasimojo: I am ashamed.

PR #2058, opened 2023-07-05
bundle-gen.py: Set the url of the origin remote unconditionally

dlom:
When opening PRs, bundle-gen will re-use one of the repositories cloned earlier to save a clone.  However, the origin of this repository points to the upstream reposistory instead of the fork repository.

/cc @lleshchi @2uasimojo 
/assign @lleshchi

### DIFF
--- a/hack/bundle-gen.py
+++ b/hack/bundle-gen.py
@@ -410,6 +410,13 @@ def open_pr(
     os.chdir(repo_full_path)
 
     repo = git.Repo(repo_full_path)
+
+    try:
+        repo.remotes.origin.set_url(fork_repo)
+    except:
+        print("Failed to set origin remote")
+        raise
+
     try:
         repo.create_remote("upstream", upstream_repo)
     except:


2uasimojo (commented):
Ah. Maybe we should just make [this](https://github.com/openshift/hive/blob/f38999660a753127d2bc2ee3d1f1ab7e00e4d2c1/hack/bundle-gen.py#L141) clone from the fork repo instead. @lleshchi wdyt?

2uasimojo:
> Ah. Maybe we should just make [this](https://github.com/openshift/hive/blob/f38999660a753127d2bc2ee3d1f1ab7e00e4d2c1/hack/bundle-gen.py#L141) clone from the fork repo instead. @lleshchi wdyt?

...except I guess we would have to be careful to fast-forward the fork in that case.

(This is all my fault -- I [suggested](https://github.com/openshift/hive/pull/2056#discussion_r1245700815) DRYing the clone.)

lleshchi:
@2uasimojo I'm in favor of the current solution as I think it is clearer to follow.

PR #2060, opened 2023-07-07
Enhance hiveutil awsprivatelink: allow the usage of an existing credentials Secret

fxierh:
Enhance hiveutil awsprivatelink subcommands by allowing the usage of an existing credentials Secret on the active cluster, e.g. Secret/aws-creds in the kube-system namespace. 

This would be useful especially in environments where the AWS environment variables are not set. 

Example usage: 
./bin/hiveutil awsprivatelink endpointvpc add vpc-07a6ad61e01bb1578 --region us-east-2 _**--creds-secret kube-system/aws-creds**_ --subnet-ids subnet-0d6731842b0943a85,subnet-0afc3944f249aea7e,subnet-0c819abeb278b3b24


fxierh:
@2uasimojo I'm back ^^

2uasimojo (commented):
This looks good.

Rather than copy/pasting the flag and processing logic in all three subcommands, why not use a `PersistentFlags()` and a `PersistentPreRun` in the `awsprivatelink` parent subcommand?

contrib/pkg/awsprivatelink/common/vars.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	CredsSecret   *corev1.Secret
+	DynamicClient client.Client
+)
```
fxierh, line 11: The CredsSecret and DynamicClient variables are defined in the `common` package (instead of the `awsprivatelink` package) to avoid import cycles like awsprivatelink -> endpointvpc -> awsprivatelink.

contrib/pkg/awsprivatelink/disable.go
```diff
@@ -16,9 +16,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type disableOptions struct {
-	dynamicClient client.Client
-}
+type disableOptions struct{}
```
fxierh, line 19: Keeping an empty struct to adhere to the same paradigm as the other subcommands. Not sure if it's worth it though.

fxierh:
@2uasimojo Good idea. I have moved the logic for getting the credentials Secret and for building the dynamic client (required for getting the Secret) to the parent command.

contrib/pkg/awsprivatelink/awsprivatelink.go
```diff
@@ -37,3 +66,34 @@ func setLogLevel(cmd *cobra.Command, args []string) {
...
 	}
 }
+
+// Get controller-runtime dynamic client
+func getDynamicClient() client.Client {
+	if err := configv1.Install(scheme.Scheme); err != nil {
+		log.WithError(err).Fatal("Failed to add Openshift configv1 types to the default scheme")
+	}
+
+	dynClient, err := hiveutils.GetClient()
```
fxierh, line 76: I feel like this dynamic client reads from the cache but not the API server directly. This leads to a small issue with etcd's optimistic locking. 

For example when I call `hiveutil awsprivatelink endpointvpc add ...` immediately after `hiveutil awsprivatelink enable` (say, in an automated test case), there's a high chance that I got something like:

`time="2023-07-06T15:32:48Z" level=fatal msg="Failed to update HiveConfig/hive" error="Operation cannot be fulfilled on hiveconfigs.hive.openshift.io \"hive\": the object has been modified; please apply your changes to the latest version and try again"`

Currently, after calling `hiveutil awsprivatelink enable` I would call wait.Poll (or something similar) on `hiveutil awsprivatelink endpointvpc add ...` which is idempotent. But is there a better solution ?

2uasimojo: You're saying you run those two commands serially/synchronously in separate processes and still see this problem? That's... not okay. There should be no way two separate processes are sharing a cache; and the only way I can think of that they're getting confused like this is if they're being assigned to different etcd replicas. IMHO that's an etcd bug.

But that doesn't help you, does it? :innocent: 

In controllers we tend to just ignore 409s like this, as they'll resolve automatically in a subsequent iteration. But in an executable context... ugh. Does it help if you, like, `sleep 5` between those two invocations?

It would be interesting to know how the `oc` command avoids this, as surely it must, or people would be screaming from the rooftops.

contrib/pkg/awsprivatelink/disable.go
2uasimojo: I kind of thought `Complete`, `Validate`, and `Run` were part of some interface, but they don't appear to be. I can't decide which pains my OCD more, having unused cruft or deviating from an established pattern. I'll leave it up to you :)

[Later] If you compose the common CredsSecret and DynamicClient in, this point becomes moot :P

contrib/pkg/awsprivatelink/common/vars.go
2uasimojo: Ack.

Though I wonder if it would look cleaner to put these in a struct that is then composed into the `*Options` of the subcommands? It would be nice to be able to avoid referring to `common.Thing` in the meat of the code.

2uasimojo (commented):
Yeah, I like this a lot better.

Comments inline, changes are optional at this point. Please squash once you're ready to merge.

contrib/pkg/awsprivatelink/awsprivatelink.go
fxierh: Dug into controller-runtime's client.New() a bit. Looks like the client actually reads and writes directly from the API server by making requests with a low-level REST client. The only thing that's being cached (as a map) is the resourceMeta (of each GVK requested once by the client), which includes a REST client, a REST mapping etc. but no actual object data.

fxierh: Yep, I ran these two commands serially like so
`cmd := exec.Command("./hiveutil", "awsprivatelink", "enable", ...)`
`out, err = cmd.CombinedOutput()`
...
`cmd = exec.Command("./hiveutil", "awsprivatelink", "endpointvpc", "add", ...)`
`out, err = cmd.CombinedOutput()`
and you are right, sleep for a few seconds help.

fxierh: Yeah this could be caused by etcd not being consistent at that moment. Not entirely sure about this but I think etcd is not strongly consistent for all nodes, as a write operation is considered successful if `quorum` number of nodes acknowledge receipt of the new data.

contrib/pkg/awsprivatelink/common/vars.go
fxierh: Thanks for pointing that out. I guess I will keep it in mind and save this to future contributions.

fxierh:
@2uasimojo Commits are squashed and this is ready to merge, thanks for the review :grin:.

PR #2062, opened 2023-07-13
Work around upstream memory leak with APIServerIPOverride

2uasimojo:
When ClusterDeployment.Spec.ControlPlaneConfig.APIServerIPOverride is in use, we experience the memory leak described by
https://github.com/kubernetes/kubernetes/issues/118703. That issue has not been resolved at the time of this writing, but describes a couple of workarounds. Try the one described in [this comment](https://github.com/kubernetes/kubernetes/issues/118703#issuecomment-1595072383):

> For now, I would probably just set the Proxy func to
> utilnet.NewProxierWithNoProxyCIDR(http.ProxyFromEnvironment) so that
> the TLS cache is skipped for your code. This isn't guaranteed behavior
> and may stop working at any time, but it will at least solve the issue
> for you for now (by restoring the old behavior).

[HIVE-2272](https://issues.redhat.com//browse/HIVE-2272)

### DIFF
--- a/pkg/remoteclient/remoteclient.go
+++ b/pkg/remoteclient/remoteclient.go
@@ -5,11 +5,13 @@ package remoteclient
 import (
 	"context"
 	"net"
+	"net/http"
 	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


2uasimojo:
image
This worked.

/assign @vincepri 
/cc @dlom

PR #2063, opened 2023-07-14
bundle-gen: Fix channel discovery for get_previous_version

2uasimojo:
We were using the wrong annotation to discover the channel of a given bundle. This resulted in `replaces` pointing to an incorrect version, and the bundles wouldn't pass CI. Fix.

[HIVE-1658](https://issues.redhat.com//browse/HIVE-1658)


2uasimojo:
Force merging; tests should not be affected by this change.

PR #2064, opened 2023-07-17
Run Hive components locally 

fxierh:
Quoting from the `usage` function:

"The script runs a Hive component (hive-opeartor or hive-controllers) locally,
simulating in-cluster environment. It records the current in-cluster Hive configurations,
scales down the selected component, then runs it locally with the same environment.
On pressing ctrl + c, the script is stopped and relevant components are scaled back up.

Warnings:
- The active cluster may be in use, in which case the script could interfere with its functioning.

Notes:
- Artifacts (in-cluster Hive configurations and log files), if produced, are stored in the --artifacts_folder
which is removed after the script ends if the --cleanup_artifacts is specified.
- Selecting hive-controllers runs all Hive controllers (including hive-clustersync) together
as a single replica.
- Endpoints of the Hive component running locally (e.g. metrics, pprof) are directly exposed on localhost.

Usage:
- ./hack/run-hive-locally.sh COMPONENT [--flag=value ...]

Parameters:
- COMPONENT: must be either $hive_component_operator or $hive_component_controllers

Optional flags:
- --hive-operator-ns=NAMESPACE, where NAMESPACE defaults to hive
- --hive-operator-log-level=LOG_LEVEL, where LOG_LEVEL defaults to debug
- --hive-controllers-log-level=LOG_LEVEL, where LOG_LEVEL defaults to debug
- --artifacts-folder=DIRECTORY, where DIRECTORY defaults to artifacts
- --preserve-artifacts
- --help"

https://issues.redhat.com/browse/OCPQE-16047

### DIFF
--- a/hack/run-hive-locally.sh
+++ b/hack/run-hive-locally.sh
@@ -1,0 +1,363 @@
+#!/usr/bin/env bash
+
+# TODO: hiveadmission support
+# TODO: run hive-controllers/hive-clustersync separately with multiple replicas each (need docker to avoid port collisions)
+
+set -e
+set -u
+set -o pipefail
+
+# Highlighted echo
+function hl_echo {
+    tput setaf 3
+    echo "$@"
+    tput sgr0
+}
+
+function usage {
+    echo "The script runs a Hive component ($hive_component_operator or $hive_component_controllers) locally
+against the active cluster, simulating in-cluster environment. It captures the in-cluster Hive
+configurations, scales down the selected component, then runs it locally with the same environment.
+On pressing ctrl + c, the script is terminated and relevant components are scaled back up.
+
+Warnings:
+- The active cluster might be in use, in which case the script could interfere with its functioning.
+
+Notes:
+- Artifacts produced are stored in the --artifacts_folder which is removed once the script ends unless
+--preserve-artifacts is specified.
+- Selecting $hive_component_controllers runs all Hive controllers (including $hive_component_clustersync) together
+as a single replica.
+- Endpoints (e.g. metrics, pprof) of the Hive component running locally are directly exposed on localhost.
+
+Usage:
+- ./hack/run-hive-locally.sh COMPONENT [--flag=value ...]
+
+Parameters:
+- COMPONENT: must be either $hive_component_operator or $hive_component_controllers
+
+Optional flags:
+- --hive-operator-ns=hive: namespace where hive-operator is installed.
+- --log-level=debug: log level of the component to run.
+- --artifacts-folder=artifacts: name of the folder that will be created for artifacts.
+- --preserve-artifacts: if specified, preserve the artifacts folder after the script ends.
+- --help"
+}
+
+function kill_background_jobs {
+    local pid
+    while read -r pid; do
+        hl_echo "Killing process $pid"
+        kill "$pid" || true
+    done < <(jobs -p)
+    wait
+}
+
+function get_envs {
+    local type_name="$1"
+    local ns="$2"
+    # Arrange output into the following format:
+    # name1 value1
+    # name2 value2
+    # ...
+    local jsonpath_query='{range .spec.template.spec.containers[0].env[*]}{@.name} {@.value}{"\n"}{end}'
+    local env
+    local value
+    local resolved_value
+
+    hl_echo "Getting environment variables from $type_name in ns/$ns"
+    while read -r env value; do
+        # Skip environment variables that are already resolved
+        if [ ${resolved_envs["$env"]+_} ]; then
+            hl_echo "Environment variable $env is already resolved, skipping"
+            continue
+        fi
+
+        # Skip irrelevant environment variables
+        if [[ $env = "CLI_CACHE_DIR" ]] || [[ $env = "HIVE_SKIP_LEADER_ELECTION" ]]; then
+            hl_echo "Environment variable $env is irrelevant, skipping"
+            continue
+        fi
+
+        # Resolve the environment variable's value
+        case "$value" in
+        /*)
+            resolved_value="./$artifacts_folder/${hc_spec_generation}$(tr / - <<< "$value").txt"
+            oc exec "$type_name" -n "$ns" -- /bin/bash -c "[ -f $value ] && cat $value || echo -n '!ENOENT!'" > "$resolved_value"
+            if [ '!ENOENT!' == "$(cat "$resolved_value")" ]; then
+                rm "$resolved_value"
+            else
+                hl_echo "File content saved to $resolved_value"
+            fi
+            ;;
+        *)
+            resolved_value="$value"
+            ;;
+        esac
+
+        resolved_envs["$env"]="$env=$resolved_value"
+        hl_echo "Environment variable $env=$resolved_value"
+    done < <(oc get "$type_name" -n "$ns" -o jsonpath="$jsonpath_query")
+}
+
+function get_envs_from {
+    local type_name="$1"
+    local ns="$2"
+    # Arrange output into the following format:
+    # ConfigMap1
+    # ConfigMap2
+    # ...
+    local jsonpath_query='{range .spec.template.spec.containers[0].envFrom[*]}{@.configMapRef.name}{"\n"}{end}'
+    # Arrange output into the following format:
+    # key1 value1
+    # key2 value2
+    # ...
+    local jq_query='to_entries[] | "\(.key) \(.value)"'
+    local cm
+    local env
+    local value
+    hl_echo "Getting envFroms from $type_name in ns/$ns"
+
+    while read -r cm; do
+        # Skip ConfigMaps that are already resolved
+        if [ ${resolved_envFrom_cms["$cm"]+_} ]; then
+            hl_echo "cm/$cm already resolved, skipping"
+            continue
+        fi
+
+        # Extract environment variables from ConfigMap
+        hl_echo "Extracting environment variables from cm/$cm in ns/$ns"
+        while read -r env value; do
+            resolved_envs["$env"]="$env=$value"
+            hl_echo "Environment variable $env=$value"
+        done < <(oc get "cm/$cm" -n "$ns" -o jsonpath="{.data}" | jq -r "$jq_query")
+
+        # Mark ConfigMap as resolved
+        resolved_envFrom_cms["$cm"]=""
+
+        # Also save ConfigMap.data for future reference
+        oc get "cm/$cm" -n "$ns" -o jsonpath="{.data}" > "./$artifacts_folder/${hc_spec_generation}-cm-$cm.txt"
+    done < <(oc get "$type_name" -n "$ns" -o jsonpath="$jsonpath_query")
+}
+
+function get_incluster_config {
+    # Get current HiveConfig.spec which will be compared against later to see if this field has changed
+    local hc_spec_file="./$artifacts_folder/${hc_spec_generation}-hiveconfig-spec.txt"
+    hiveconfig_spec="$(oc get hiveconfig hive -o jsonpath='{.spec}' | jq '.maintenanceMode = true' --sort-keys)"
+    echo "$hiveconfig_spec" > "$hc_spec_file"
+    hl_echo "Current HiveConfig.spec saved to $hc_spec_file"
+
+    # Get envs and ConfigMaps corresponding to the HiveConfig.spec obtained above
+    local type_names=("deploy/$hive_component_controllers" "sts/$hive_component_clustersync")
+    local type_name
+    for type_name in "${type_names[@]}"; do
+        hl_echo "Getting in-cluster configurations from $type_name in ns/$hive_controllers_ns"
+        get_envs "$type_name" "$hive_controllers_ns"
+        get_envs_from "$type_name" "$hive_controllers_ns"
+    done
+}
+
+function wait_until_hiveconfig_spec_changes {
+    local new_hiveconfig_spec
+    while read -r; do
+        new_hiveconfig_spec="$(oc get hiveconfig hive -o jsonpath='{.spec}' | jq)"
+        # As "oc get -w" gives spurious outputs sometimes, here we check if HiveConfig.spec is actually updated
+        if [[ "$new_hiveconfig_spec" != "$hiveconfig_spec" ]]; then
+            break
+        fi
+    done < <(oc get hiveconfig hive --watch-only --no-headers)
+
+    hc_spec_generation=$(( hc_spec_generation + 1 ))
+    hive_controllers_ns="$(oc get hiveconfig hive -o jsonpath='{.spec.targetNamespace}')"
+    hl_echo "Changes to HiveConfig.spec detected (new version on the left):"
+    hl_echo "$(diff -y <(echo "$new_hiveconfig_spec" ) <(echo "$hiveconfig_spec"))"
+    hl_echo "Current HiveConfig.spec generation = $hc_spec_generation"
+    hl_echo "Current Hive controllers namespace = $hive_controllers_ns"
+}
+
+function enable_maintenance_mode {
+    hl_echo "Enabling maintenance mode"
+    oc patch hiveconfig hive -p '{"spec": {"maintenanceMode": true}}' --type merge
+
+    hl_echo "Waiting for the controllers to be scaled down"
+    # Blocks until the existing hive-controllers is updated to have 0 replica
+    oc wait -n "$hive_controllers_ns" --for=jsonpath='{.spec.replicas}'=0 deploy/"$hive_component_controllers"
+    # Blocks until the new hive-clustersync is created and has zero replica
+    while ! oc wait -n "$hive_controllers_ns" --for=jsonpath='{.spec.replicas}'=0 sts/"$hive_component_clustersync"; do
+        sleep 1
+    done
+}
+
+function disable_maintenance_mode {
+    hl_echo "Disabling maintenance mode"
+    oc patch hiveconfig hive -p '{"spec": {"maintenanceMode": false}}' --type merge
+
+    hl_echo "Waiting for the controllers to be scaled up"
+    # Blocks until the existing hive-controllers is updated to have 1 replica
+    while ! oc wait -n "$hive_controllers_ns" --for=jsonpath='{.status.updatedReplicas}'=1 deploy/"$hive_component_controllers"; do
+        sleep 1
+    done
+    # Blocks until the new hive-clustersync is created and has the correct number of replicas
+    while ! oc wait -n "$hive_controllers_ns" --for=jsonpath='{.status.updatedReplicas}'="$hive_clustersync_replicas" sts/"$hive_component_clustersync"; do
+        sleep 1
+    done
+}
+
+function run_hive_operator_locally {
+    hl_echo "Scaling down Deployment/hive-operator"
+    oc scale -n "$hive_operator_ns" deploy/"$hive_component_operator" --replicas=0
+    oc wait -n "$hive_operator_ns" --for=jsonpath='{.spec.replicas}'=0 deploy/"$hive_component_operator"
+
+    hl_echo "Running $component locally, press ctrl + c when you are done"
+    local final_env_values=(
+        HIVE_OPERATOR_NS="$hive_operator_ns" \
+        LOG_LEVEL="$log_level"
+    )
+    env "${final_env_values[@]}" make run-operator 2>&1 | tee "./$artifacts_folder/${hive_component_operator}-logs.txt"
+}
+
+function run_hive_controllers_locally {
+    while true; do
+        # Declare (global) associative arrays
+        declare -gA resolved_envs resolved_envFrom_cms
+
+        # Get in-cluster configurations
+        get_incluster_config
+        resolved_envs["HIVE_NS"]="HIVE_NS=$hive_controllers_ns"
+        resolved_envs["HIVE_CLUSTERSYNC_POD_NAME"]="HIVE_CLUSTERSYNC_POD_NAME=hive-clustersync-0"
+
+        # Enable maintenance mode and wait for the controllers to be scaled down
+        enable_maintenance_mode
+
+        # Run hive-controllers as a background job
+        hl_echo "Running $hive_component_controllers locally, press ctrl + c when you are done"
+        # The "build" dependency of the "run" make target is only needed for the first iteration,
+        # i.e. when hc_spec_generation is zero.
+        if [ $hc_spec_generation -eq 0 ]; then
+            env "${resolved_envs[@]}" LOG_LEVEL="$log_level" make run 2>&1 |
+            tee "./$artifacts_folder/${hc_spec_generation}-${hive_component_controllers}-logs.txt" &
+        else
+            env "${resolved_envs[@]}" ./bin/manager --log-level="$log_level" 2>&1 |
+            tee "./$artifacts_folder/${hc_spec_generation}-${hive_component_controllers}-logs.txt" &
+        fi
+
+        # The controllers running locally are stopped once HiveConfig.spec changes
+        # They will be re-started with the latest configurations in the next iteration
+        wait_until_hiveconfig_spec_changes
+        kill_background_jobs
+
+        # Disable maintenance mode and wait for the controllers to be scaled up
+        # We can thus get Hive configurations with "oc exec" in the next iteration
+        disable_maintenance_mode
+        unset resolved_envs resolved_envFrom_cms
+    done
+}
+
+function restore_and_cleanup {
+    hl_echo $'\nCleaning up before exit'
+
+    # Clear trap
+    trap '' EXIT SIGINT
+
+    # Restore Hive
+    case "${component}" in
+    "$hive_component_operator")
+        hl_echo "Scaling up Deployment/hive-operator"
+        oc scale -n "$hive_operator_ns" deploy/hive-operator --replicas="$hive_operator_replicas"
+        ;;
+    "$hive_component_controllers")
+        kill_background_jobs
+        disable_maintenance_mode
+        ;;
+    esac
+
+    # Clean up artifacts
+    if [[ -d "./$artifacts_folder" ]] && [[ "$cleanup_artifacts" = "true" ]]; then
+        hl_echo "Cleaning up artifacts"
+        rm -r "$artifacts_folder"
+    fi
+
+    # Propagate the exit signal to stop script execution
+    exit 0
+}
+
+# Defaults
+hive_component_operator="hive-operator"
+hive_component_controllers="hive-controllers"
+hive_component_clustersync="hive-clustersync"
+hive_operator_ns="hive"
+log_level="debug"
+artifacts_folder="artifacts"
+cleanup_artifacts="true"
+hc_spec_generation=0
+
+# Parse parameters
+component="${1:-}"
+case "$component" in
+"$hive_component_operator"|"$hive_component_controllers")
+    shift
+    ;;
+*)
+    hl_echo "Invalid component: \"$component\"" >&2
+    usage
+    exit 1
+    ;;
+esac
+
+# Parse options
+while [ $# -gt 0 ]; do
+    case "$1" in
+    "--hive-operator-ns="*)
+        hive_operator_ns="${1#*=}"
+        ;;
+    "--log-level="*)
+        log_level="${1#*=}"
+        ;;
+    "--artifacts-folder="*)
+        artifacts_folder="${1#*=}"
+        ;;
+    "--preserve-artifacts")
+        cleanup_artifacts="false"
+        ;;
+    "--help")
+        usage
+        exit 0
+        ;;
+    *)
+        hl_echo "Invalid option: \"$1\"" >&2
+        usage
+        exit 1
+        ;;
+    esac
+    shift
+done
+
+# Exit if the artifacts folder already exists
+if [[ -d "./$artifacts_folder" ]]; then
+    hl_echo "Artifacts folder $artifacts_folder already exists, please remove it before running the script. "
+    exit 1
+fi
+
+# Queries
+hive_controllers_ns="$(oc get hiveconfig hive -o jsonpath='{.spec.targetNamespace}')"
+hive_controllers_ns="${hive_controllers_ns:-hive}"
+hive_operator_replicas="$(oc get "deploy/$hive_component_operator" -n "$hive_operator_ns" -o jsonpath='{.spec.replicas}')"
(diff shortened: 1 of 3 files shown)


fxierh:
Some thoughts: 

**Limitations of the current implementation**:

To run hive-controllers or hive-clustersync locally, hive-operator is scaled down to 0 replica as well. As a result, the ConfigMaps are not re-deployed by hive-operator, and neither are the components of Hive. In other words, nothing changes after running `oc edit hiveconfig`. 

**Potential solutions**:

Do not disable hive-operator when running hive-controllers or hive-clustersync locally. To avoid (in-cluster) Hive components getting scaled back up:
- Option 1-1: set HiveConfig.spec.controllersConfig.controllers[*].config.replicas = 0 -> **this only works for hive-clustersync but not hive-controllers**
- Option 1-2: set HiveConfig.spec.maintenanceMode = 0 -> **this scales both hive-controllers & hive-clustersync down, so they are bundled together and we'd have to run both of them locally at the same time**

Meanwhile, we watch for changes to the ConfigMaps managed by hive-operator:
- Option 2-1: use Bash's watch command to check periodically if their contents change -> **possible but not real time**
- Option 2-2: `oc rsync ...` -> for various reason (e.g. rsync command not available for remote Pod), **this does not seem to work**

Finally, re-deploy the component running locally once a ConfigMap changes:
- Option 3-1: use Bash's watch command, kill the current `make run` command and restart it (with new configuration) on ConfigMap change

**Open questions**

Is it worth implementing the solution above (e.g. option 1-2 + option 2-1 + option 3-1), which is imperfect itself ?

2uasimojo:
1. I don't mind requiring controllers+clustersync to be either both local or both in-cluster. Does this make it harder to use for debugging etc?
2. There's also `oc get ... -w` which will pop immediately any time the object is changed. You may get some spurious events this way, but you could mitigate the disruption by doing a filtered diff of the before-and-after versions of the ConfigMap.
3. Sure. This would e.g. kill the current debugging session, but that effectively simulates what happens in cluster, right?

fxierh:
@2uasimojo Finally some progress from my side 😁 . 

1. Running both of them local/in-cluster is okay for me. It is slightly simpler to check the logs this way. 
2. `oc get -w` followed by some filtering is great. I'm taking this approach now. Thanks !
3. Yeah, the current implementation (when running hive-controllers locally) get Hive configurations (ConfigMaps) -> make run (with the aforementioned configurations) in the background -> kill background jobs when HiveConfig.spec changes -> repeat

fxierh:
Running hive-clustersync locally is giving me division by zero errors when calling:
https://github.com/openshift/hive/blob/7fa8c0ba4cf77f940adf7e400d8c3c5d66594f52/pkg/controller/clustersync/clustersync_controller.go#L344
where 
https://github.com/openshift/hive/blob/7fa8c0ba4cf77f940adf7e400d8c3c5d66594f52/pkg/controller/clustersync/clustersync_controller.go#L341

Looks like when the in-cluster sts/hive-clustersync is scaled down to 0 replica, hive-clustersync won't run locally due to the above error.

2uasimojo:
> Looks like when the in-cluster sts/hive-clustersync is scaled down to 0 replica, hive-clustersync won't run locally.

Oh. Yeah.

So we either need a way to scale down controllers without scaling down clustersync; or we need to be satisfied not running clustersync at all in this mode; or we need to find a way to make clustersync believe it's running in a single replica (which is probably better for debugging anyway (unless you're trying to debug the distribution of syncsets across replicas :shrug:)).

I don't love putting test enablement in production code, but sometimes it's warranted. How about this:
```go
	logger.Debug("calculating replicas")
	replicas := int64(*sts.Spec.Replicas)
	// For test purposes, if we've scaled down clustersync so we can run locally, this will be zero; spoof it to one:
	if replicas == 0 {
		logger.Warning("ClusterSync StatefulSet has zero replicas! Hope you're running locally!")
		replicas = 1
	}
```

fxierh:
Great to be able to run all controllers locally. Thanks !

I've conducted a simple test:
1. Run all controllers locally with the script
2. ./bin/hiveutil create-cluster fxie-hive-1 --base-domain qe.devcluster.openshift.com --region us-east-2; oc create -f syncset.yaml
3. Wait for the CD to be installed. Check SyncSet, MachinePool, CD. 
The controllers (that run locally) looks healthy. 

The PR is now ready for review.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+    exit 0
+}
+
+# Configuration
+hive_component_operator="hive-operator"
+hive_component_controllers="hive-controllers"
+hive_components_to_choose_from=("$hive_component_operator" "$hive_component_controllers")
+hive_operator_ns="hive"
+hive_controllers_ns="$(oc get hiveconfig hive -o jsonpath='{.spec.targetNamespace}')"
+hive_controllers_ns="${hive_controllers_log_level:-hive}"
```
2uasimojo, line 213:
```suggestion
hive_controllers_ns="${hive_controllers_ns:-hive}"
```

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+--cleanup_artifacts
+--help"
+}
+
+function kill_background_jobs {
+    while read -r pid; do
+        hl_echo "Killing job $pid"
+        kill "$pid" || true
+    done < <(jobs -p)
+    sleep "$kill_job_grace_period_seconds"
```
2uasimojo, line 46: Are you going to send a stronger signal once this grace period passes, or are you just giving time for processes to exit? If the latter, consider using `wait` instead.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+    *)
+        hl_echo "Invalid option: $1" >&2
+        usage
+        exit 1
+        ;;
+    esac
+    shift
+done
+
+# Check if artifacts folder exists. Create it if not.
```
2uasimojo, line 253: IMHO there should always be a way to run non-interactively. I would maybe not even bother being nice here -- just fail if the artifacts folder exists, and let the user decide whether to empty, remove, or rename it.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+    hl_echo "Creating artifacts folder $artifacts_folder"
+    mkdir "$artifacts_folder"
+fi
+
+# Set trap
+trap restore_and_cleanup EXIT SIGINT
+
+# Run the selected component locally
+PS3="Which component to run locally ? "
+select component in "${hive_components_to_choose_from[@]}"; do
```
2uasimojo, line 282: Again, I believe there should be a way to be non-interactive. Why not a mandatory param to indicate which component we care about?

2uasimojo (commented):
This is really cool.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+
+function usage {
+    echo "The script runs a Hive component ($hive_component_operator or $hive_component_controllers) locally,
+simulating in-cluster environment. It records the current in-cluster Hive configurations,
+scales down the selected component, then runs it locally with the same environment.
+On pressing ctrl + c, the script is stopped and relevant components are scaled back up.
+
+Notes:
+- Artifacts (in-cluster Hive configurations and log files), if produced, are stored in the --artifacts_folder
+which is removed after the script ends if the --cleanup_artifacts is specified.
```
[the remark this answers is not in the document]
fxierh: Yep. Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+
+# Highlighted echo
+function hl_echo {
+    tput setaf 3
+    echo "$@"
+    tput sgr0
+}
+
+function usage {
+    echo "The script runs a Hive component ($hive_component_operator or $hive_component_controllers) locally,
```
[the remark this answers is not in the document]
fxierh: Okay, the introduction is updated and a warning section is added for this.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+
+Usage:
+- ./hack/run-hive-locally.sh [--flag=value ...]
+
+Supported flags:
+--hive-operator-ns=NAMESPACE
+--hive_operator_log_level=LOG_LEVEL
+--hive_controllers_log_level=LOG_LEVEL
+--artifacts_folder=DIRECTORY
+--cleanup_artifacts
```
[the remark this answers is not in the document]
fxierh: Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+
+function get_incluster_config {
+    hl_echo "Getting in-cluster Hive configurations"
+
+    resolved_env_values=()
+    local env_configmap_keys=(
+        "MANAGED_DOMAINS_FILE/hive-managed-domains/managed-domains" \
+        "AWS_PRIVATELINK_CONTROLLER_CONFIG_FILE/aws-private-link/aws-private-link" \
+        "FAILED_PROVISION_CONFIG_FILE/hive-failed-provision-config/hive-failed-provision-config" \
+        "METRICS_CONFIG_FILE/hive-metrics-config/hive-metrics-config"
```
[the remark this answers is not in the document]
fxierh: Yep. I've made some changes so the script reads (and filters) the envs and envFroms from hive-controllers and hive-clustersync using `oc exec`.

hack/run-hive-locally.sh
fxierh: Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+        hl_echo "Artifacts folder $artifacts_folder not found, creating"
+        mkdir "$artifacts_folder"
+    fi
+
+    # Resolve some necessary env's
+    local key
+    for env_configmap_key in "${env_configmap_keys[@]}"; do
+        env="${env_configmap_key%%/*}"
+        configmap="$(echo "$env_configmap_key" | cut -d '/' -f 2)"
+        key="${env_configmap_key##*/}"
```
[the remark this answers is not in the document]
fxierh: Great to know, thanks. The current version makes use of associative arrays to avoid parsing environment variables / ConfigMaps that are already read. 

(This also makes me upgrade Bash on my mac :joy:. )

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+        value="./$artifacts_folder/$configmap-${hc_spec_generation}.txt"
+
+        # Save content to file
+        oc get cm "$configmap" -n "$hive_operator_ns" -o jsonpath="{.data.$key}" > "$value" || true
```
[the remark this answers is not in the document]
fxierh: Agreed. Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+        # key1:val1
+        # key2:val2
+        # ...
+        # Then extract each pair of key/val
+        while read -r env_value; do
+            env="${env_value%:*}"
+            value="${env_value#*:}"
+            resolved_env_values+=("$env=$value")
+            hl_echo "Environment variable $env = $value"
+        done < <(oc get cm "$configmap" -n "$hive_operator_ns" -o jsonpath="{.data}" | jq -r 'to_entries[] | "\(.key):\(.value)"' || true)
```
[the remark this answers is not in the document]
fxierh: Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+
+        # Also save ConfigMap.data for future reference
+        oc get cm "$configmap" -n "$hive_operator_ns" -o jsonpath="{.data}" > \
+        "./$artifacts_folder/$configmap-${hc_spec_generation}.txt" || true
```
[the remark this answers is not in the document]
fxierh: Fixed.

hack/run-hive-locally.sh
fxierh: The latter. Fixed.

hack/run-hive-locally.sh
fxierh: Okay, current version is non-interactive, so no need for input.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+    hl_echo "Enabling maintenance mode"
+    oc patch hiveconfig hive -p '{"spec": {"maintenanceMode": true}}' --type merge
+
+    hl_echo "Waiting for the controllers to be scaled down"
+    # Blocks until the existing hive-controllers is updated to have 0 replica
+    oc wait -n "$hive_controllers_ns" --for=jsonpath='{.spec.replicas}'=0 deploy/"$hive_component_controllers"
+    # Blocks until the new hive-clustersync is created and has zero replica
+    while ! oc wait -n "$hive_controllers_ns" --for=jsonpath='{.spec.replicas}'=0 sts/"$hive_component_clustersync"; do
+        sleep 1
+    done
```
fxierh, line 191: Looks like hive-operator delete and re-create hive-clustersync when necessary (some fields are immutable). So a while loop here to avoid `oc wait` error out.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,296 @@
...
+    fi
+
+    # Restore Hive
+    case "${component}" in
+    "$hive_component_operator")
+        hl_echo "Scaling up Deployment/hive-operator"
+        oc scale -n "$hive_operator_ns" deploy/hive-operator --replicas="$hive_operator_replicas"
+        ;;
+    "$hive_component_controllers")
+        kill_background_jobs
```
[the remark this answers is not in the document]
fxierh: I'm not 100% sure about this. I think we need it here to make sure the background jobs are killed on pressing control + c, which causes wait_until_hiveconfig_spec_changes to return and starts restore_and_cleanup.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+
+function disable_maintenance_mode {
+    hl_echo "Disabling maintenance mode"
+    oc patch hiveconfig hive -p '{"spec": {"maintenanceMode": false}}' --type merge
+
+    hl_echo "Waiting for the controllers to be scaled up"
+    # Blocks until the existing hive-controllers is updated to have 1 replica
+    while ! oc wait -n "$hive_controllers_ns" --for=jsonpath='{.status.updatedReplicas}'=1 deploy/"$hive_component_controllers"; do
+        sleep 1
+    done
```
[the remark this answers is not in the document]
2uasimojo: Why use `updatedReplicas` rather than just `replicas`?

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+        "")
+            hl_echo "Environment variable $env does not have a .value, skipping"
+            continue
+            ;;
+        /*)
+            resolved_value="./$artifacts_folder/$(basename "$value")-${hc_spec_generation}.txt"
+            # Skip if the file does not exist
+            if ! oc exec "$type_name" -n "$ns" -- /bin/bash -c "cat $value" > "$resolved_value"; then
+                rm "$resolved_value"
+                continue
```
2uasimojo, line 98: There are other reasons this `oc exec` could fail, and I would think we want to blow up on most of them. Also, again it is technically possible that "file exists and is empty" behaves differently than "file does not exist". This isn't beautiful, but...
```suggestion
            oc exec "$type_name" -n "$ns" -- /bin/bash -c "[ -f $value ] && cat $value || echo -n '!ENOENT!'" > "$resolved_value"
            if [ '!ENOENT!" == "$(cat $resolved_value)" ]; then
                rm "$resolved_value"
                continue
```

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+    # Arrange output into the following format:
+    # ConfigMap1
+    # ConfigMap2
+    # ...
+    local jsonpath_query='{range .spec.template.spec.containers[0].envFrom[*]}{@.configMapRef.name}{"\n"}{end}'
+    # Arrange output into the following format:
+    # key1=value1
+    # key2=value2
+    # ...
+    local jq_query='to_entries[] | "\(.key)=\(.value)"'
```
2uasimojo, line 124: similar comment to the above about `read`ing multiple values rather than delimiting and parsing "manually".

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+            value="${env_value#*=}"
+            resolved_envs["$env"]="$env=$value"
+            hl_echo "Environment variable $env = $value"
+        done < <(oc get "cm/$cm" -n "$ns" -o jsonpath="{.data}" | jq -r "$jq_query")
+
+        # Mark ConfigMap as resolved
+        resolved_envFrom_cms["$cm"]=""
+
+        # Also save ConfigMap.data for future reference
+        oc get "cm/$cm" -n "$ns" -o jsonpath="{.data}" > "./$artifacts_folder/$cm-${hc_spec_generation}.txt"
```
2uasimojo, line 151: Similar comment about using the generation as the prefix.

Also it might not be a terrible idea to put the actual string `-cm-` in the file name to make it easy to spot that it came from a configmap?

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+        # Run hive-controllers in the background
+        hl_echo "Running $hive_component_controllers locally, press ctrl + c when you are done"
+        # Build the manager binary for the first iteration only
+        if [ $hc_spec_generation -eq 0 ]; then
+            env "${resolved_envs[@]}" LOG_LEVEL="$hive_controllers_log_level" make run 2>&1 |
+            tee "./$artifacts_folder/${hive_component_controllers}-logs-${hc_spec_generation}.txt" &
+        else
+            env "${resolved_envs[@]}" ./bin/manager --log-level="$hive_controllers_log_level" 2>&1 |
+            tee "./$artifacts_folder/${hive_component_controllers}-logs-${hc_spec_generation}.txt" &
+        fi
```
2uasimojo, line 244: Why branch here?

Oh, is it because the `build` dep actually gets rebuilt every time?

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+
+        # Re-deploy the controllers once HiveConfig.spec changes
+        wait_until_hiveconfig_spec_changes
```
2uasimojo, line 247: This seems odd. If hiveconfig hasn't changed recently, won't this hang indefinitely?

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+    hl_echo "Running $component locally, press ctrl + c when you are done"
+    local final_env_values=(
+        HIVE_OPERATOR_NS="$hive_operator_ns" \
+        LOG_LEVEL="$hive_operator_log_level"
+    )
+    env "${final_env_values[@]}" make run-operator 2>&1 | tee "./$artifacts_folder/${hive_component_operator}-logs.txt"
+}
+
+function run_hive_controllers_locally {
+    while true; do
```
2uasimojo, line 223: I don't think I understand why this loop is here. Need to look closer tomorrow...

hack/run-hive-locally.sh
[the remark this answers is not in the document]
fxierh: This is to make sure that hive-controllers has 1 replica and that replica is ready. 

(IMO updatedReplicas and readyReplicas are the same here. )

hack/run-hive-locally.sh
fxierh: Sorry for the confusion. 

This loop is meant to work as follows:
- Get the current in-cluster configurations (ConfigMaps, env vars)
- Enable maintenance mode so the in-cluster controllers are scaled down
- Run the controllers locally in the background (the `make run` and `tee` things do not hang)
- `wait_until_hiveconfig_spec_changes`, which blocks until HiveConfig.spec is updated. 
This is not a hotloop here as the while loop in `wait_until_hiveconfig_spec_changes` proceeds to the next iteration only when `oc get --watch-only` gives a new message. 
- `kill_background_jobs` if we reach here, i.e. when HiveConfig.spec changes
- Disable maintenance mode, so the in-cluster controllers are scaled up, allowing us to do `oc exec` in the next iteration when getting in-cluster configurations

hack/run-hive-locally.sh
fxierh: This is meant to hang until HiveConfig.spec changes. As a result, HiveConfig.spec changes => the program proceed to the next iteration of the while loop => get the latest in-cluster configurations => run controllers locally => repeat.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+Usage:
+- ./hack/run-hive-locally.sh COMPONENT [--flag=value ...]
+
+Parameters:
+- COMPONENT: must be either $hive_component_operator or $hive_component_controllers
+
+Optional flags:
+- --hive-operator-ns=NAMESPACE, where NAMESPACE defaults to hive
+- --hive-operator-log-level=LOG_LEVEL, where LOG_LEVEL defaults to debug
+- --hive-controllers-log-level=LOG_LEVEL, where LOG_LEVEL defaults to debug
```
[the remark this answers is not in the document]
fxierh: Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+    # ...
+    local jsonpath_query='{range .spec.template.spec.containers[0].env[*]}{@.name}={@.value}{"\n"}{end}'
+    local env_value
+    local env
+    local value
+    local resolved_value
+
+    hl_echo "Getting environment variables from $type_name in ns/$ns"
+    while read -r env_value; do
+        env="${env_value%=*}"
```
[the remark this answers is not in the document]
fxierh: Oh yeah, this is actually better. Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,365 @@
...
+
+        # Resolve the environment variable's value
+        value="${env_value#*=}"
+        case "$value" in
+        "")
+            hl_echo "Environment variable $env does not have a .value, skipping"
+            continue
+            ;;
+        /*)
+            resolved_value="./$artifacts_folder/$(basename "$value")-${hc_spec_generation}.txt"
```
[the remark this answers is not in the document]
fxierh: Fixed. Also aligned the naming of the logs file with this implementation.

hack/run-hive-locally.sh
fxierh: Fixed.

hack/run-hive-locally.sh
fxierh: Fixed. Now it looks like `0-cm-hive-controllers-config.txt` for example.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,349 @@
...
+- --help"
+}
+
+function kill_background_jobs {
+    local pid
+    while read -r pid; do
+        hl_echo "Killing process $pid"
+        kill "$pid" || true
+    done < <(jobs -p)
+    wait
```
fxierh, line 53: This might be too much detail here but I'd like to jot down some thoughts. 

`jobs -p` returns PID of the background job's process group leader. So in case of `sleep 180 | sleep 180 | sleep 180 & `, this function only kills the first sleep process, not all of them. In our case, however, this function kills the `make run` (or `./bin/manager` depending on what iteration of the while loop it is), causing the stdin to the `tee` command to be closed, so `tee` terminates once all data it has are flushed to its outputs. In other words, this function is imperfect but seem to work in our case. 

I once tried to kill the job number instead, something like `kill %n`. This does seem to work, killing all processes which belong to the job / process group. Unfortunately, it looks like something in the `make run | tee` pipeline does not handle SIGTERM properly, resulting in `make run` ending abruptly. This messes with Hive's leader election process, as the leader is killed before releasing the lease, forcing the next deployment (after HiveConfig.spec changes) to wait for a few minutes.

hack/run-hive-locally.sh
fxierh: Thanks for the suggestion. Fixed.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,349 @@
...
+            continue
+        fi
+
+        # Resolve the environment variable's value
+        case "$value" in
+        /*)
+            resolved_value="./$artifacts_folder/${hc_spec_generation}$(tr / - <<< "$value").txt"
+            oc exec "$type_name" -n "$ns" -- /bin/bash -c "[ -f $value ] && cat $value || echo -n '!ENOENT!'" > "$resolved_value"
+            if [ '!ENOENT!' == "$(cat "$resolved_value")" ]; then
+                rm "$resolved_value"
```
fxierh, line 88: Removed `continue` after this line. This way in case the file does not exist, we still have ENV=FILE where FILE does not exist. This aligns with what we have in-cluster.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,363 @@
...
+    while read -r; do
+        new_hiveconfig_spec="$(oc get hiveconfig hive -o jsonpath='{.spec}' | jq)"
+        # As "oc get -w" gives spurious outputs sometimes, here we check if HiveConfig.spec is actually updated
+        if [[ "$new_hiveconfig_spec" != "$hiveconfig_spec" ]]; then
+            break
+        fi
+    done < <(oc get hiveconfig hive --watch-only --no-headers)
+
+    hc_spec_generation=$(( hc_spec_generation + 1 ))
+    hive_controllers_ns="$(oc get hiveconfig hive -o jsonpath='{.spec.targetNamespace}')"
```
fxierh, line 171: In case the target namespace is updated

fxierh:
Hope this version is better.

hack/run-hive-locally.sh
2uasimojo: Yeah, it's a pity that `build` doesn't do the usual `make` thing and only rebuild if sources have changed :(

hack/run-hive-locally.sh
2uasimojo: Yah, cool.

May consider printing a warning about it.

hack/run-hive-locally.sh
2uasimojo: I get it now -- it was the `&`s on L238/241 that made it click for me.

So you don't have any control over the controllers getting restarted if something about the hiveconfig changes. Could that disrupt a debugging session? (I guess just... don't change the hiveconfig :P)

2uasimojo (commented):
This is looking great. A small handful of comments inline, but we're at a point where we should just land this and iterate later.

/lgtm

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,363 @@
...
+    echo "The script runs a Hive component ($hive_component_operator or $hive_component_controllers) locally
+against the active cluster, simulating in-cluster environment. It captures the in-cluster Hive
+configurations, scales down the selected component, then runs it locally with the same environment.
+On pressing ctrl + c, the script is terminated and relevant components are scaled back up.
+
+Warnings:
+- The active cluster might be in use, in which case the script could interfere with its functioning.
+
+Notes:
+- Artifacts produced are stored in the --artifacts_folder which is removed once the script ends unless
```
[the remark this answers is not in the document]
fxierh: Yep, noted. Will fix in future PR.

hack/run-hive-locally.sh
fxierh: Noted. Will fix in future PR.

hack/run-hive-locally.sh
fxierh: I guess it won't, as this aligns with the in-cluster functioning of Hive -- when HiveConfig is updated, (most often) Hive components are re-deployed to reflect the configuration change. And without rebuilding the binaries, local restarts are just slightly slower than the in-cluster re-deployment. 

But yeah, we could add a flag which controls this behavior -- whether to restart on HiveConfig.spec changes or not.

hack/run-hive-locally.sh
```diff
@@ -0,0 +1,363 @@
...
+    local value
+    local resolved_value
+
+    hl_echo "Getting environment variables from $type_name in ns/$ns"
+    while read -r env value; do
+        # Skip environment variables that are already resolved
+        if [ ${resolved_envs["$env"]+_} ]; then
+            hl_echo "Environment variable $env is already resolved, skipping"
+            continue
+        fi
```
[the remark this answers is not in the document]
fxierh: You are right. I also hesitated between running them separately (in Docker or so to avoid port collision) and together. Later I started to think that the script is meant to be a simple, quick way of debugging etc. with few dependencies, so in the end I stuck with the current implementation and added a TODO at the top of the script.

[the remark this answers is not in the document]
2uasimojo: > running them separately (in Docker or so to avoid port collision)

It's a pretty thin line between running in Docker and just running in KIND, which is already possible (or so they tell me -- I've never done it).

But yeah, I understand the issue. Hopefully it won't be a problem.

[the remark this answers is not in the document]
fxierh: This reminds me about a comparison I made between approaches for running/developing/testing Hive:
1) Openshift cluster + remote image registry (quay.io for ex.): 
Closest to production environment; 
Cluster creation is slow;
Requires an image build (so implicit requirements on platform architecture) each time the source code is updated; 
Image pushing/pulling is slow; 
This is QEs' go-to approach. 
2) KIND cluster + local image build/push/pull using `./hack/create-kind-cluster.sh`: 
Kubernetes/containerized environment; 
cluster creation is fast;
Requires an image build each time the source code is updated;  
Image pushing/pulling is fast; 
Has some requirements on memory, dependencies, setups (`./hack/hiveadmission-dev-cert.sh`) etc. 
3) Openshift + script merged by this PR: 
Runs bare components locally;
Cluster creation is slow;
Does not require image building/pushing/pulling at all;
Has minimal requirements.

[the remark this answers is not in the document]
2uasimojo: > 1. Requires an image build

Note that there's a little shortcut the dev team uses. It does depend on your host OS being fedora/RHEL:
`make image-hive-fedora-dev-base` creates the slow and seldom-changing parts (including `dnf update`) of the docker image.
Then `make docker-dev-push` knows to look for the image generated by the above and add just the hive bits on top.
(Pushing is still slow, yeah.)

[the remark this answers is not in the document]
fxierh: Would be great to save some time with this. 

-- Me on my Mac with an ARM chip :sob:

PR #2066, opened 2023-07-20
Account for duplicate taints on MachineSet

suhanime:
Taints are provided as a list, so duplicates are possible. Ensure they are all synced/updated as per the merging strategy.
Also includes a fix for occasional flakes while comparing actual and expected taints. Now we sort them first.

xref: https://issues.redhat.com/browse/HIVE-2268

/assign @2uasimojo

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -597,25 +597,21 @@ func (r *ReconcileMachinePool) syncMachineSets(
 				if rMS.Spec.Template.Spec.Taints == nil {
 					rMS.Spec.Template.Spec.Taints = []corev1.Taint{}
 				}
-				// Make a temporary map of generated MachineSet's taints for easy lookup
-				gTaintsByKey := make(map[string]*corev1.Taint)
-				for gIndex, gTaint := range ms.Spec.Template.Spec.Taints {
-					gTaintsByKey[gTaint.Key] = &ms.Spec.Template.Spec.Taints[gIndex]
-				}
-				// Go through the remote MachineSet's taints, replacing overlaps with the generated MachineSet
-				for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
-					if gTaint, foundTaint := gTaintsByKey[rTaint.Key]; foundTaint {
-						if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
-							rMS.Spec.Template.Spec.Taints[rIndex] = *gTaint
-							objectModified = true
+				for _, gTaint := range ms.Spec.Template.Spec.Taints {
+					foundTaint := false
+					for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+						if gTaint.Key == rTaint.Key {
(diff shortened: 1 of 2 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -597,25 +597,21 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
-				for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
-					if gTaint, foundTaint := gTaintsByKey[rTaint.Key]; foundTaint {
-						if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
-							rMS.Spec.Template.Spec.Taints[rIndex] = *gTaint
-							objectModified = true
+				for _, gTaint := range ms.Spec.Template.Spec.Taints {
+					foundTaint := false
+					for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+						if gTaint.Key == rTaint.Key {
+							foundTaint = true
```
2uasimojo, line 604: I dug into this a bit, and it looks like you're allowed to have multiple taints with the same key. It seems the uniqueness constraint is on key+effect.

```
(.venv) [efried@efried install]$ oc get no hive01-75tg4-worker-dlc2z -o jsonpath={.spec.taints}|jq
[
  {
    "effect": "PreferNoSchedule",
    "key": "blah",
    "value": "bar"
  },
  {
    "effect": "PreferNoSchedule",
    "key": "foo",
    "value": "bar"
  },
  {
    "effect": "NoExecute",
    "key": "foo",
    "value": "baz"
  },
  {
    "effect": "NoSchedule",
    "key": "foo",
    "value": "bar"
  }
]
(.venv) [efried@efried install]$ oc adm taint no hive01-75tg4-worker-dlc2z foo=baz:NoSchedule
error: node hive01-75tg4-worker-dlc2z already has foo taint(s) with same effect(s) and --overwrite is false
```

So we need to adjust this logic accordingly.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -597,25 +597,21 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
-				// Go through the remote MachineSet's taints, replacing overlaps with the generated MachineSet
-				for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
-					if gTaint, foundTaint := gTaintsByKey[rTaint.Key]; foundTaint {
-						if gTaint.Value != rTaint.Value || gTaint.Effect != rTaint.Effect {
-							rMS.Spec.Template.Spec.Taints[rIndex] = *gTaint
-							objectModified = true
+				for _, gTaint := range ms.Spec.Template.Spec.Taints {
+					foundTaint := false
+					for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+						if gTaint.Key == rTaint.Key {
```
2uasimojo, line 603: This should be more like
```suggestion
						if gTaint.Key == rTaint.Key && gTaint.Effect == rTaint.Effect {
```

...and subsequent changes accordingly.

PR #2069, opened 2023-07-26
MachinePool: Properly handle labels & taints

2uasimojo:
With this commit, MachinePool grows two new statusfields: `OwnedLabels` and `OwnedTaints`. These keep track of the labels and taints that were added to the remote MachineSets at the behest of this MachinePool. Conversely, this allows us to identify which ones were added on the remote side, and thus shouldn't be touched by our controller.

We also fix the logic for matching taints, which are "keyed" by the combination of the `Key` and `Effect` fields.

To summarize the behavior with this commit in play:
- Labels and taints listed in MachinePool.Spec are synced to the remote MachineSets.
- Labels and taints that are *not* in MachinePool.Spec are removed from the remote MachineSets *iff* they are present in
MachinePool.Status.Owned{Labels|Taints}.
- Upon successful syncing, we update MachinePool.Status.Owned{Labels|Taints} to match the spec.

The effect is that editing MachinePool.Spec to remove a label or taint should cause that label/taint to be removed from the remote MachineSets, BUT labels/taints that only exist on the remote MachineSets remain untouched.

[HIVE-2035](https://issues.redhat.com//browse/HIVE-2035)
[HIVE-2276](https://issues.redhat.com//browse/HIVE-2276)
[HIVE-2279](https://issues.redhat.com//browse/HIVE-2279)

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -19,6 +20,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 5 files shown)


suhanime:
Superseded by [2082](https://github.com/openshift/hive/pull/2082)

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -618,25 +638,51 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+				for _, rTaint := range rMS.Spec.Template.Spec.Taints {
+					tid := identifierForTaint(&rTaint)
+
+					// If we need to "delete" this one, skip it, but mark modified
+					if taintsToDelete.Has(tid) {
+						objectModified = true
+						continue
 					}
+
+					if gTaint, foundTaint := gTaintsByID[tid]; foundTaint && gTaint.Value != rTaint.Value {
```
2uasimojo, line 657: BUG: if the `rMS` has duplicates, we'll replace the first one, but then because we deleted it from `gTaintsByID`, subsequent dups will be added as is.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -618,25 +638,51 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
 				}
-				// Any remaining taints from the temporary map are new and need to be added
-				for _, gTaint := range gTaintsByKey {
-					rMS.Spec.Template.Spec.Taints = append(rMS.Spec.Template.Spec.Taints, *gTaint)
-					objectModified = true
+				// Anything remaining from the generated MachineSet is new
+				for _, taint := range gTaintsByID {
+					taints = append(taints, *taint)
 				}
+				sort.SliceStable(taints, func(i, j int) bool {
```
2uasimojo, line 675: POSSIBLE BUG: Does taint order matter to k8s? If so, this is not the right thing to do. We should in fact be conveying the ordering from the MachinePool to the extent that is feasible.

PR #2071, opened 2023-07-28
e2e (AWS): Allow env var to override creds file

2uasimojo:
Resolve a long-standing TODO which allows AWS environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to be used when present, even if a credentials file exists, as the gods intended.

hiveutil already uses the appropriate logic such that, if a creds file is not supplied, it looks for the env vars. We just needed to fix it so we don't pass that arg in at all if the env vars are set.

(This was an incidental requirement of [HIVE-2234](https://issues.redhat.com//browse/HIVE-2234): the AWS credentials from the hypershift installation were taking precedence for managed DNS, which then bounced because the expected hosted zone is in a different account.)

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -181,27 +181,23 @@ USE_MANAGED_DNS=${USE_MANAGED_DNS:-true}
 case "${CLOUD}" in
 "aws")
 	CREDS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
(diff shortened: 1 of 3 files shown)


2uasimojo:
Overriding freeze issue. This is test-only, so no SD impact, and I want to be able to move forward with https://github.com/openshift/release/pull/39941

PR #2075, opened 2023-08-03
Add `tar` to installer image

2uasimojo:
We use the `tar` command to package up the results of `oc adm must-gather` when install fails. It used to be included in ubi-minimal (I swear), but isn't anymore. Install it explicitly.

[HIVE-2288](https://issues.redhat.com//browse/HIVE-2288)


2uasimojo:
This change doesn't touch any CI paths, so we don't really have to wait.

PR #2076, opened 2023-08-03
Consolidate and bump default AWS instance type (m6i.xlarge)

2uasimojo:
This commit cleans up how we handle instance types in hiveutil and e2e CI.
- `hiveutil create-cluster` now accepts an `--aws-instance-type` option.
- The default instance type is now `m6i.xlarge`. It was previously `m5.xlarge` for hiveutil and `m4.large` for the MachinePool scaling tests in our e2e CI.
- This default now lives in one place, so it's easier to change later without missing things.
- To allow easier noodling via CI rehearsals, `e2e` will feed a new `$AWS_INSTANCE_TYPE` environment variable through to the new `--aws-instance-type` param for `create-cluster`.

[HIVE-2235](https://issues.redhat.com//browse/HIVE-2235)

### DIFF
--- a/contrib/pkg/clusterpool/clusterpool.go
+++ b/contrib/pkg/clusterpool/clusterpool.go
@@ -271,6 +271,16 @@ func (o *ClusterPoolOptions) generateObjects() ([]runtime.Object, error) {
 		BaseDomain: o.BaseDomain,
 	}
 
+	var awsBuilder *clusterresource.AWSCloudBuilder
+	if o.Cloud == cloudAWS {
+		awsBuilder = &clusterresource.AWSCloudBuilder{
+			Region: o.Region,
+			// TODO: CLI option for this
+			InstanceType: clusterresource.AWSInstanceTypeDefault,
+		}
+		builder.CloudBuilder = awsBuilder
+	}
+
 	if o.createCloudSecret {
 		switch o.Cloud {
 		case cloudAWS:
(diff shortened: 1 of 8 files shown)


2uasimojo:
Oh. No, that's actually broken. I think I missed a path through clusterpools somehow.

2uasimojo:
> hive-operator was in a weird hot loop

This was just the [namespace watch filter](https://github.com/openshift/hive/blob/350489bb93088e5da0596a9340d3b08c4af3c421/pkg/operator/hive/hive_controller.go#L288). We weren't actually reconciling (despite the message).

2uasimojo:
/test e2e

unrelated flake

2uasimojo:
/test coverage

known taint flake

/test e2e-pool

I have absolutely no idea what happened here. It looks like we got past the previous error, but then something weird happened with the uninstall of the ("real") pool cluster.

fxierh:
Looks like 4.14 starts to use `OVNKubernetes`. Wondering if we should generate install-config with this networkType.

2uasimojo:
This actually looks like it pretty much succeeded, but ran up against the overall 2h test timeout.

/test e2e-pool

Even if this eventually works, we're obviously skating way too close to the maximum allowable test time. I'm not even sure what we would do about that.

2uasimojo:
I'm only guessing the slowdown has something to do with having increased the instance size. (Bigger instances take longer to... provision? hibernate? resume?) I believe I found about half an hour we can trim out of the test via https://github.com/openshift/hive/pull/2080 -- will retest when that lands.

fxierh:
Sorry there might be a mix-up, I wan't referring to the e2e-pool job fail. What I meant is, in addition to the default AWS instance type (generated by hiveutil), do you think it's a good idea to also update the networkType (to align with Openshift Installer's default) ?

2uasimojo:
> Sorry there might be a mix-up, I wan't referring to the e2e-pool job fail. What I meant is, in addition to the default AWS instance type (generated by hiveutil), do you think it's a good idea to also update the networkType (to align with Openshift Installer's default) ?

Gosh, I don't know. Is there a benefit to doing so? Will it make something faster or cheaper? Is OpenShiftSDN being deprecated/removed?

jstuever:
> Gosh, I don't know. Is there a benefit to doing so? Will it make something faster or cheaper? Is OpenShiftSDN being deprecated/removed?

I'd be curious if we could remove NetworkType and let things naturally use their defaults.

2uasimojo:
> I'd be curious if we could remove NetworkType and let things naturally use their defaults.

Okay, sure: https://github.com/openshift/hive/pull/2085

Note: I'm pretty sure this is only affecting hiveutil (and thereby e2e).

PR #2078, opened 2023-08-08
Fix hiveutil awsprivatelink bug caused by unknown user

fxierh:
user.Current() errors out when the binary runs as an unknown user (no entry in /etc/passwd), which is the case when running in Openshift. 

https://github.com/golang/go/issues/38599


fxierh:
"The QE found a bug in his own backyard. "

PR #2081, opened 2023-08-08
e2e (AWS): Default BASE_DOMAIN for new CI account

2uasimojo:
Our e2e suite for AWS was setting a default base domain that (AFAIK) we don't have control over, so we were overriding it in the test config. We've recently switched to a new AWS account for these tests. Change the default base domain accordingly so we don't need to override it in the test config.

[HIVE-2259](https://issues.redhat.com//browse/HIVE-2259)


2uasimojo:
/assign @dlom 

Will allow us to resolve this TODO: [1](https://github.com/openshift/release/pull/41519/files#diff-830e25154235fbd3e0d938a97aab01217d13383d3347c90bece1ffb6da2c08d9R93), [2](https://github.com/openshift/release/pull/41519/files#diff-830e25154235fbd3e0d938a97aab01217d13383d3347c90bece1ffb6da2c08d9R131)

PR #2082, opened 2023-08-09
Properly handle MachinePool Labels and Taints

suhanime:
With implementing the merging of MachinePool Labels and Taints, we lost the ability to sync and update based on MachinePool.Spec. To fix this, we add two new statusfields to MachinePool: OwnedLabels and OwnedTaints. These keep track of the labels and taints that were added to the remote MachineSets at the behest of this MachinePool. Conversely, this allows us to identify which ones were added on the remote side, and thus shouldn't be touched by our controller.

We also fix the logic for matching taints, which are "keyed" by the combination of the Key and Effect fields.

To summarize the behavior with this commit in play:
- Labels and taints listed in MachinePool.Spec are synced to the remote MachineSets.
- Labels and taints that are *not* in MachinePool.Spec are removed from the remote MachineSets *iff* they are present in
MachinePool.Status.Owned{Labels|Taints}.
- Upon successful syncing, we update MachinePool.Status.Owned{Labels|Taints} to match the spec.

The effect is that editing MachinePool.Spec to remove a label or taint should cause that label/taint to be removed from the remote MachineSets, BUT labels/taints that only exist on the remote MachineSets remain untouched.

[HIVE-2035](https://issues.redhat.com//browse/HIVE-2035)
[HIVE-2276](https://issues.redhat.com//browse/HIVE-2276)
[HIVE-2279](https://issues.redhat.com//browse/HIVE-2279)

/assign @2uasimojo

### DIFF
--- a/apis/hive/v1/machinepool_types.go
+++ b/apis/hive/v1/machinepool_types.go
@@ -100,6 +100,26 @@ type MachinePoolStatus struct {
 	// Conditions includes more detailed status for the cluster deployment
 	// +optional
 	Conditions []MachinePoolCondition `json:"conditions,omitempty"`
+
+	// OwnedLabels lists the keys of labels this MachinePool created on the remote MachineSet.
+	// Used to identify labels to remove from the remote MachineSet when they are absent from
+	// the MachinePool's spec.labels.
+	// +optional
+	OwnedLabels []string `json:"ownedLabels,omitempty"`
+	// OwnedTaints lists identifiers of taints this MachinePool created on the remote MachineSet.
+	// Used to identify taints to remove from the remote MachineSet when they are absent from
+	// the MachinePool's spec.taints.
+	// +optional
+	OwnedTaints []TaintIdentifier `json:"ownedTaints,omitempty"`
+}
+
+// TaintIdentifier uniquely identifies a Taint. (It turns out taints are mutually exclusive by
+// key+effect, not simply by key.)
+type TaintIdentifier struct {
+	// Key matches corev1.Taint.Key.
+	Key string `json:"key:omitempty"`
+	// Effect matches corev1.Taint.Effect.
+	Effect corev1.TaintEffect `json:"effect:omitempty"`
 }
 
 // MachineSetStatus is the status of a machineset in the remote cluster.
--- a/apis/hive/v1/machinepool_types.go
+++ b/apis/hive/v1/machinepool_types.go
@@ -54,6 +54,9 @@ type MachinePoolSpec struct {
 
 	// List of taints that will be applied to the created MachineSet's MachineSpec.
 	// This list will overwrite any modifications made to Node taints on an ongoing basis.
+	// In case of duplicate entries, first encountered taint Value will be preserved,
+	// and the rest collapsed on the corresponding MachineSets.
+	// Note that taints are uniquely identified based on key+effect, not just key.
 	// +optional
 	Taints []corev1.Taint `json:"taints,omitempty"`
 }
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"reflect"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -447,8 +446,8 @@ func (r *ReconcileMachinePool) generateMachineSets(
 			ms.Spec.Template.Spec.ObjectMeta.Labels[key] = value
 		}
 
-		// Apply hive MachinePool taints to MachineSet MachineSpec.
-		ms.Spec.Template.Spec.Taints = pool.Spec.Taints
+		// Apply hive MachinePool taints to MachineSet MachineSpec. Also collapse duplicates if present.
+		ms.Spec.Template.Spec.Taints = *controllerutils.GetUniqueTaints(&pool.Spec.Taints)
 	}
 
 	logger.Infof("generated %v worker machine sets", len(generatedMachineSets))
@@ -557,7 +556,7 @@ func (r *ReconcileMachinePool) syncMachineSets(
 	taintsToDelete := sets.Set[hivev1.TaintIdentifier]{}
 	taintsToDelete.Insert(pool.Status.OwnedTaints...)
 	for _, taint := range pool.Spec.Taints {
-		taintsToDelete.Delete(identifierForTaint(&taint))
+		taintsToDelete.Delete(controllerutils.IdentifierForTaint(&taint))
 	}
 
 	// Find MachineSets that need updating/creating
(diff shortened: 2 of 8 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -617,25 +640,35 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+					foundTaint := false
+					for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+						if reflect.DeepEqual(identifierForTaint(&gTaint), identifierForTaint(&rTaint)) {
+							foundTaint = true
+							if gTaint.Value != rTaint.Value {
+								rMS.Spec.Template.Spec.Taints[rIndex].Value = gTaint.Value
+								objectModified = true
+							}
 						}
-						delete(gTaintsByKey, rTaint.Key)
```
suhanime, line 632: @2uasimojo Your code for taints was sophisticated but it still had the fatal flaw of not being able to handle duplicate entries for taints. I tried fixing it but it kept on getting too complicated, so I ditched that and opted for keeping track with a flag approach. It does require looping twice, but it is similar to the handling of labels, hence easier to read

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -637,52 +640,36 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
-					//   "owned" list, i.e. it is "remote only". Preserve it.
 					taints = append(taints, rTaint)
-					// We'll want to include any "extras" from the generated MachineSet
-					delete(gTaintsByID, tid)
 				}
-				// Anything remaining from the generated MachineSet is new
-				for _, taint := range gTaintsByID {
-					taints = append(taints, *taint)
-				}
-				sort.SliceStable(taints, func(i, j int) bool {
```
suhanime: Got rid of the sorting simply because I was lazy. Is it needed? Or just good-to-have?

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -617,25 +640,35 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
-							rMS.Spec.Template.Spec.Taints[rIndex] = *gTaint
-							objectModified = true
+				rMS.Spec.Template.Spec.Taints = taints
+
+				for _, gTaint := range ms.Spec.Template.Spec.Taints {
+					// Taints are provided as a list, so duplicate entries are possible.
+					// Use a flag to indicate if a taint was found, append if not.
+					foundTaint := false
+					for rIndex, rTaint := range rMS.Spec.Template.Spec.Taints {
+						if reflect.DeepEqual(identifierForTaint(&gTaint), identifierForTaint(&rTaint)) {
```
2uasimojo, line 660: We should not need to use DeepEqual, as TaintIdentifier is a simple struct with two scalar fields. (If we needed to use DeepEqual here, using TaintIdentifier as a map key above would be... questionable.)

pkg/controller/machinepool/machinepool_controller_test.go
```diff
@@ -380,8 +387,122 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
...
-						Key:   "test-taint",
-						Value: "test-value",
+						Key:    "test-taint",
+						Value:  "test-value",
+						Effect: "NoSchedule",
+					})
+					return ms
+				}(),
+			},
+		},
```
2uasimojo, line 397: Can we please get a test where the machinepool has duplicates? I predict that we'll end up "ignoring" (actually overwriting) all but the last one. And I think I don't mind that behavior as much as leaving (identical) duplicates in the remote MachineSet...

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -617,25 +640,35 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+							foundTaint = true
+							if gTaint.Value != rTaint.Value {
+								rMS.Spec.Template.Spec.Taints[rIndex].Value = gTaint.Value
```
2uasimojo, line 663: Another thing that's happening here: even for "owned" taints, we're preserving the order as it existed previously on the spoke. So the next bug report will be: "I can't change the order of taints by editing the MachinePool."

(I'll note that my attempt was also patently wrong on this count, as I was sorting.)

This only matters if taint order matters in the first place. I'm not sure it does. Research needed.

2uasimojo (commented):
My main concern here is still around duplicate taints. As currently written:
- If there are duplicates in the MP, we'll end up using the value for the last one. I'm actually okay with this (but it needs UT).
- If there are duplicates on the remote side, we'll end up with the same number of identical (same value) duplicates. I can't think of a scenario where this would be useful. Even if the order in which the taints are processed comes into play (more on this below), the same effect could be achieved by preserving only the [first or last, depending how they are consumed] one. (It would be up to the spoke user to decide which in this case.)

There's also the concern about ordering. Does taint order matter? Educated guess: it does not, since they are effectively ANDed/cumulative: the presence of any matching taint will trigger the effect; and the most strict effect will be honored in case of duplicate key+values. But we should confirm this.

pkg/controller/machinepool/machinepool_controller.go
2uasimojo: Missed this; but addressed it in my comments anyway :)

pkg/controller/machinepool/machinepool_controller.go
suhanime: Done. The only DeepEqual now wrt Taints in the Reconcile Test where expected and actual taints are compared.

pkg/controller/machinepool/machinepool_controller_test.go
suhanime: Done. Also ensured the expected entries are collapsed.

pkg/controller/machinepool/machinepool_controller.go
suhanime: Taint order doesn't matter as far as I can see, for the corresponding node should tolerate all taints from the corresponding Machine (though I'm still very green wrt taints and tolerations). That being said, I have removed the sorting of labels and taints - so the user is aware about which taints are collapsed.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -601,29 +623,28 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+				taintMap := make(map[hivev1.TaintIdentifier]corev1.Taint)
+				for _, rTaint := range rMS.Spec.Template.Spec.Taints {
+					if taintsToDelete.Has(controllerutils.IdentifierForTaint(&rTaint)) {
+						// This entry would be deleted from the final list
+						objectModified = true
+						continue
+					}
+					// Preserve the taint entry first encountered.
+					if !controllerutils.TaintExists(taintMap, &rTaint) {
+						taintMap[controllerutils.IdentifierForTaint(&rTaint)] = rTaint
```
2uasimojo, line 636: Hm, I think if we *don't* add this entry -- because it's a dup -- we need to set `objectModified`.
```suggestion
						taintMap[controllerutils.IdentifierForTaint(&rTaint)] = rTaint
					} else {
						// Skip this taint, which effectively means we're removing a duplicate.
						objectModified = true
```

We should have UT for this path: I.e. when the only "modification" is removing a dup from the rMS, make sure we're doing an `Update()`. As currently written, I don't think we will.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -601,29 +623,28 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
 					}
 				}
-				// Any remaining taints from the temporary map are new and need to be added
-				for _, gTaint := range gTaintsByKey {
-					rMS.Spec.Template.Spec.Taints = append(rMS.Spec.Template.Spec.Taints, *gTaint)
-					objectModified = true
+				for _, gTaint := range ms.Spec.Template.Spec.Taints {
+					// Generated MachineSet will not have duplicate taints, because we have collapsed duplicates when we generated them.
+					if !controllerutils.TaintExists(taintMap, &gTaint) ||
+						(controllerutils.TaintExists(taintMap, &gTaint) && taintMap[controllerutils.IdentifierForTaint(&gTaint)] != gTaint) {
```
2uasimojo, line 642: The `TaintExists` here is redundant, isn't it?
```suggestion
						taintMap[controllerutils.IdentifierForTaint(&gTaint)] != gTaint {
```

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -601,29 +623,28 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+						taintMap[controllerutils.IdentifierForTaint(&gTaint)] = gTaint
+						objectModified = true
+					}
 				}
+				rMS.Spec.Template.Spec.Taints = *controllerutils.ListFromTaintMap(&taintMap)
```
2uasimojo, line 647: So yeah, see below: even though the logic behind `Update()` is based on `objectModified` and not `DeepEqual`, I feel like it would be a weird UX to see the taints get shuffled randomly every time there's a change.

But meh. Let's defer that to a future PR.

apis/hive/v1/machinepool_types.go
```diff
@@ -54,6 +54,8 @@ type MachinePoolSpec struct {
 
 	// List of taints that will be applied to the created MachineSet's MachineSpec.
 	// This list will overwrite any modifications made to Node taints on an ongoing basis.
+	// In case of duplicate entries, first encountered taint Value will be preserved,
+	// and the rest collapsed on the corresponding MachineSets
```
2uasimojo, line 58: It wouldn't hurt to add
```suggestion
	// and the rest collapsed on the corresponding MachineSets.
	// Note that taints are uniquely identified based on key+effect, not just key.
```

pkg/controller/utils/taints.go
```diff
@@ -0,0 +1,46 @@
...
+		// This means that when we encounter duplicates, the value of the first entry encountered will be preserved.
+		if !TaintExists(uniqueTaints, &taint) {
+			uniqueTaints[IdentifierForTaint(&taint)] = taint
+		}
+	}
+	return ListFromTaintMap(&uniqueTaints)
+}
+
+// ListFromTaintMap simply compiles a list of taints from the corresponding map
+func ListFromTaintMap(taintMap *map[hivev1.TaintIdentifier]corev1.Taint) *[]corev1.Taint {
```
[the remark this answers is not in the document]
suhanime: So I thought about it, but really, we cannot guarantee any order as long as a map is involved. If fact, even multiple append/updates will result in k8s returning the list with values added as they encounter the request - so the list itself could also be in different order.
I have taken care to reword so we say first entry "encountered" will be preserved, not the topmost.
As far as the order changing arbitrarily, we can at the least guarantee CR will only be updated when necessary.

suhanime (commented):
@2uasimojo this should be ready to go in now.

pkg/controller/utils/taints.go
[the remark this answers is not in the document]
2uasimojo: We can guarantee ordering as long as we _only use maps for dup checking_. Roughly:

```
// Used to detect dups
uniques := map[TaintIdentifier]Taint{}
// Built up in order, first by generated, then by remote
taints := []Taint{}
// MachinePool takes precedence, so loop it first
for taint in gMS.Taints:
  id4t := IdentifierForTaint(taint)
  // We shouldn't have to consult taintsToDelete in this loop
  // because they'll be absent from gMS.Taints anyway.
  if _, ok := uniques[id4t]; ok:
    // Dup being squashed out
    objectModified = true
  else:
    // Note that this may be adding a new taint, but we don't
    // know that yet for purposes of objectModified; we'll figure
    // it out later.
    taints = append(taints, taint)
    uniques[id4t] = taint

// Now process remotes, differently.
// This set will be used to decide if the above loop *added* taints.
origRemotes := Set[TaintIdentifier]{}
for taint in rMS.Taints:
  id4t := IdentifierForTaint(taint)
  origRemotes.Add(id4t)
  if dup, ok := uniques[id4t]; ok:
    // This may be a modification (values differ) *or* a no-op
    // (values are the same)
    if dup.Value != taint.Value:
      objectModified = true
  elif _, ok := taintsToDelete[id4t]; ok:
    // Owned taint removed from MP
    objectModified = true
  else:
    taints = append(taints, taint)
    // Ensure dups that are only in the rMS get squashed too
    uniques[id4t] = taint

// At this point, if uniques contains any keys that weren't in the
// original rMS, it means we added taints from the MP.
for k := range uniques:
  if !origRemotes.Has(k):
    objectModified = true
    break
```

PR #2083, opened 2023-08-09
e2e-pool: Plumb in BASE_DOMAIN

2uasimojo:
hiveutil still defaults the base domain to new-installer.openshift.com, which is probably what we want since we (hive engineers) are the ones invoking it by hand, using the hive-team cluster as a hub, and will thus prefer less typing.

But now that we're switching to a new AWS account for CI, we want the AWS-based e2e suites to use the new base domain. We recognize and default a `$BASE_DOMAIN` variable for e2e itself, but weren't yet plumbing it through to e2e-pool. Fix.

[HIVE-2259](https://issues.redhat.com//browse/HIVE-2259)

### DIFF
--- a/hack/e2e-pool-test.sh
+++ b/hack/e2e-pool-test.sh
@@ -218,6 +218,7 @@ go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" clusterpool create-pool \
   ${REGION_ARG} \
 	${CREDS_FILE_ARG} \
 	--pull-secret-file="${PULL_SECRET_FILE}" \
+	--base-domain="${CLUSTER_DOMAIN}" \
   --image-set "${IMAGESET_NAME}" \
   --region us-east-1 \
   --size 0 \


hack/e2e-pool-test.sh
```diff
@@ -218,6 +218,7 @@ go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" clusterpool create-pool \
   ${REGION_ARG} \
 	${CREDS_FILE_ARG} \
 	--pull-secret-file="${PULL_SECRET_FILE}" \
+	--base-domain="${CLUSTER_DOMAIN}" \
```
2uasimojo, line 221: ntr: CLUSTER_DOMAIN is set based on BASE_DOMAIN depending whether we're [using managed DNS](https://github.com/openshift/hive/blob/5bbb1928ca9960e9641794d19035a7d8bdc6aac4/hack/e2e-common.sh#L219) or [not](https://github.com/openshift/hive/blob/5bbb1928ca9960e9641794d19035a7d8bdc6aac4/hack/e2e-common.sh#L234)

2uasimojo:
Well the plumbing is working.

But somehow we're enabling managed DNS, despite explicitly disabling it in the [test config](https://github.com/openshift/release/blob/a3ca5e1feda13d29abc85cc0501a33b76945e500/ci-operator/config/openshift/hive/openshift-hive-master.yaml#L152).

Actually, now that I look at it, we should be `export`ing that sucker. Fix on the way...

https://github.com/openshift/release/pull/42172

2uasimojo:
Of note: it doesn't look like we have a way to make clusterpool clusters use managed DNS. I'm not completely sure there's a use case for that in the first place, but I thought it was interesting.

2uasimojo:
> Actually, now that I look at it, we should be `export`ing that sucker. Fix on the way...

✓

test e2e-pool

2uasimojo:
also

/hold cancel

2uasimojo:
Okay, I'm sick of this -- infra flake on cleanup after the test succeeded.

/override ci/prow/e2e-pool

PR #2084, opened 2023-08-09
Add (and use) full version label to ClusterDeployment

2uasimojo:
Our clusterversion controller creates labels on the ClusterDeployment based on the spoke cluster's `clusterversion version` object's `status.desired.version` string. However, previously we were using `semver` to parse that string and reporting only the major, minor, and patch numerals, e.g.

```
      "hive.openshift.io/version-major": "4",
      "hive.openshift.io/version-major-minor": "4.14",
      "hive.openshift.io/version-major-minor-patch": "4.14.0"
```

In cases where the clusterversion is pre-release, this lacks the pre-release information.

This commit adds a new label containing the full, un-parsed version string from clusterversion, e.g.

```
      "hive.openshift.io/version": "4.14.0-ec.4"
```

We also cut over all the internal code paths that used to use the major-minor-patch label for version checking, as the full string will always be more accurate, particularly when determining whether we're before or after a version containing specific functionality.

The other external-facing effect is in the tabular output of `oc get cd`, which now reports the full version string rather than the major-minor-patch string. For GA-level clusters, this will be identical; it will only differ (being more precise) for pre-release clusters.

[HIVE-2293](https://issues.redhat.com//browse/HIVE-2293)


config/crds/hive.openshift.io_clusterdeployments.yaml
```diff
@@ -26,7 +26,7 @@ spec:
     - jsonPath: .metadata.labels.hive\.openshift\.io/cluster-region
       name: Region
       type: string
-    - jsonPath: .metadata.labels.hive\.openshift\.io/version-major-minor-patch
```
suhanime, line 29: Are we completely sure no one uses this label before we remove it? From what I understand, the practice is to keep both and let the user decide which one they want to use

2uasimojo: We’re not removing the label, just changing which label we use in `oc get cd`

2uasimojo: To expand a bit: 99% of clusters (and 100% of prod clusters, since they should only be using GAed releases) will have the same value here before & after.

2uasimojo:
/test e2e-pool

claim deletion hung: improving artifact gathering via #2086.

suhanime (commented):
/lgtm

PR #2088, opened 2023-08-10
e2e: unset exit-on-error during cleanup

2uasimojo:
We want cleanup to be best-effort, and we don't want failures therein to blow the whole test up if it already succeeded otherwise. `set +e` in all exit traps.

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -161,7 +161,7 @@ function save_hive_logs() {
 }
 # The consumer of this lib can set up its own exit trap, but this basic one will at least help
 # debug e.g. problems from `make deploy` and managed DNS setup.
-trap 'kill %1; save_hive_logs' EXIT
+trap 'set +e; kill %1; save_hive_logs' EXIT
 
 # Install Hive
 IMG="${HIVE_IMAGE}" make deploy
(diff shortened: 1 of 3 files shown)


2uasimojo:
/assign @dlom 

...for [this flake](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hive/2085/pull-ci-openshift-hive-master-e2e-pool/1689717061896376320)

hack/e2e-common.sh
```diff
@@ -161,7 +161,7 @@ function save_hive_logs() {
 }
 # The consumer of this lib can set up its own exit trap, but this basic one will at least help
 # debug e.g. problems from `make deploy` and managed DNS setup.
-trap 'kill %1; save_hive_logs' EXIT
+trap 'set +e; kill %1; save_hive_logs' EXIT
```
dlom, line 164: Shouldn't it be `set -e`? I've never seen it with a + before

2uasimojo: `set -e` turns error-on-exit *on*; `set +e` turns it *off*.

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html <== see 3rd paragraph from the bottom.

PR #2095, opened 2023-08-22
Add and use Reachable() in (remote client) Builder

2uasimojo:
With the controller-runtime upgrade from 8cfdccaec, our remote client `Builder` interface's `Build()` method got lazier. Before, if the remote client was unreachable, `Build()` itself would fail. Now it won't fail until the resulting client is actually used. This resulted in our "unreachable" controller incorrectly marking the ClusterDeployment's APIURLOverride as reachable when it wasn't, which had the ripple effect of making the clustersync controller fail early by trying to use that URL rather than falling back to the default.

This was exposed in situations where the DNS for the APIURLOverride was being configured via a syncset: the clustersync controller tried to use a URL that wouldn't resolve, so it couldn't sync the resource that would make that DNS work.

To fix, this commit adds a new `Reachable()` method to the `Builder` interface. This internally builds the client and then uses it to query the `/apis` of the remote client. We replace our used of `Build()` with this method in the "unreachable" controller to get back to the previous behavior.

[HIVE-2144](https://issues.redhat.com//browse/HIVE-2144)

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -91,6 +91,8 @@ fi
 export CLUSTER_NAME="${CLUSTER_NAME:-hive-$(uuidgen | tr '[:upper:]' '[:lower:]')}"
 
 echo "Creating cluster deployment"
+# Add a bogus API URL override to validate that our unreachable controller correctly
+# falls back to the default API URL when the override is unreachable.
 go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" create-cluster "${CLUSTER_NAME}" \
 	--cloud="${CLOUD}" \
 	${CREDS_FILE_ARG} \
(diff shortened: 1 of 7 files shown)


2uasimojo:
/hold for testing
/assign @abutcher 
/cc @lleshchi 

It would be pretty cool if we could add e2e for this... but I don't know how we would do that.

2uasimojo:
> It would be pretty cool if we could add e2e for this... but I don't know how we would do that.

I guess we could just always have our e2e set the APIURLOverride to something bogus. IIUC everything will work exactly as before because we'll always fall back to the default. Thoughts?

2uasimojo:
> have our e2e set the APIURLOverride to something bogus.

This worked in the lab, so I added it here (via a separate commit so we can yank it easily if need be).

abutcher (commented):
Looks good to me! The bogus override e2e addition is a great idea

2uasimojo:
/hold cancel

We're as satisfied with the fix as we can be without testing it in OSD staging. Will merge manually once CI passes. Then we plan to roll it out to one stage shard to test.

PR #2100, opened 2023-08-29
AWS deprovision: Plumb HostedZoneRole for shared VPC

2uasimojo:
The shared VPC flow added the need for the installer to use a HostedZoneRole to manage certain DNS records. This is provided at installation time via the install-config, and is included in the metadata.json produced by the installer when provisioning completes.

The uninstall process requires this HostedZoneRole during deprovisioning to perform the corresponding cleanup -- without it, those records are leaked. So we need to plumb that datum through from the metadata.json to the uninstall process.

To make this more general-purpose for adoption and disaster recovery, and for retrofitting existing clusters that were created before this commit, we also attempt to pull the HostedZoneRole out of the install-config if it is unset.

NOTE: We will refuse to proceed with deprovision if this field is not set. Once this change lands, we should audit existing clusters to make sure they all get retrofitted; and manually retrofit any others (with the empty string).

[HIVE-2297](https://issues.redhat.com//browse/HIVE-2297)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1794,10 +1801,16 @@ func generateDeprovision(cd *hivev1.ClusterDeployment) (*hivev1.ClusterDeprovisi
 
 	switch {
 	case cd.Spec.Platform.AWS != nil:
+		// If we haven't discovered the HostedZoneRole yet, fail
+		hzrp := controllerutils.AWSHostedZoneRole(cd)
```
jharrington22, line 1805: This role wont exit for all clusters, only those installing with the shared VPC. If I understand this correctly this will block all clusters deprovisions where the hosted zone role doesn't exist?

2uasimojo:
NTR: This mostly follows the precedent established by #1997 / #2020.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: This will only block deprovisions where the hosted zone role is unset **in the `cd.spec.clusterMetadata`**. We set that value in one of two code paths:
- When provision completes, we pull it out of the metadata.json laid down by the installer.
- For retrofitting existing clusters, we pull it out of the install-config.

In both of those installer structures, the field is of type `string`, not `*string` so it will always "exist".

However, `cd.spec.clusterMetadata.aws.hostedZoneRole` is of type `*string`, and we're distinguishing among three possible permutations:
- `nil` means "unset" -- i.e. we have never set the clusterMetadata field. This is the condition where we block deprovision.
- non-`nil`, empty string: this will be the case for legacy/non-shared-VPC clusters. This is a valid value and will *pass* the check on this line, allowing deprovision.
- non-`nil`, nonempty string: this is the case for shared-VPC clusters. This is valid, will pass this check, and will also pass that value down into the deprovisioner.

2uasimojo:
Oh, btw, I do intend to add UT for this in a fup. For right now it's urgent that we get this landed and tested live.

2uasimojo:
D'oh, forgot to `git add` new files, rookie mistake :eyeroll:

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -26,6 +26,7 @@ import (
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivev1aws "github.com/openshift/hive/apis/hive/v1/aws"
+	"github.com/openshift/hive/apis/hive/v1/aws"
```
abutcher, line 29: `make update-gofmt` wants to sort this for `make verify`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
jharrington22: Ack this make sense

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2241,6 +2254,99 @@ func (r *ReconcileClusterDeployment) addOwnershipToSecret(cd *hivev1.ClusterDepl
...
+
+	if ic.Platform.AWS == nil {
+		// This log might be redundant with one from ValidateInstallConfig
+		log.Warn("No AWS platform section in install-config")
+		return false
+	}
+
+	hzr := ic.Platform.AWS.HostedZoneRole
+	// It's okay if this is the empty string.
+	log.WithField("hostedZoneRole", hzr).Info("Found AWS HostedZoneRole in install-config")
```
jharrington22, line 2345: If this cluster is non shared VPC wont we log the empty string saying the hosted zone role "" was found?

2uasimojo: Yup.

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
2uasimojo: Sigh.

For the sake of expediency, Imma let e2e* finish before I push this change; then if `verify` passes I'll mash the merge button, wfy?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: I should elaborate: That's useful information, particularly in this code path where we're retrofitting existing clusters.

jharrington22: I was thinking that might be confusing to someone debugging issues. Not sure if we want to have two messages one where its found (not an empty string) and one where the role is not set, though might be confusing with the `nil` case. Up to you.

2uasimojo:
e2e-pool: quota flake.
verify: fixed by latest push.

We're going to merge this manually assuming `lgtm`, as we're reasonably confident in the fix and it'll take some time to resolve the quota issue.

abutcher:
e2e logged:

```
time="2023-08-29T20:08:07.097Z" level=info msg="Found AWS HostedZoneRole in ClusterMetadata" hostedZoneRole=
```

PR #2101, opened 2023-08-30
Revert "AWS deprovision: Plumb HostedZoneRole for shared VPC"

abutcher:
Reverts openshift/hive#2100


abutcher:
Pinning stage and integration for the time being instead.

PR #2102, opened 2023-08-30
Adding Temporary debug logs to vendored installer destroy code

patrickdillon:
In case we want to add some debug statements to installer destroy code for [HIVE-2297](https://issues.redhat.com//browse/HIVE-2297)


2uasimojo:
/cc @abutcher 

This does two things:
- **Disable the "don't change clusterMetadata once set" webhook validation.**

This should allow us to run this code in stage without breaking cluster service. That breakage was because CS a) vendors the old CRD, and b) is accidentally "editing" out the new platform.aws field via the ol' [round trip bug](https://groups.google.com/a/redhat.com/g/aos-devel/c/xs96y-oQdDA/m/tmEkFkPcAQAJ). What should happen now is that CS will remove the new field, but hive will put it right back.

This change will cause us to fail `unit` and `coverage` CI because we haven't updated the unit test cases accordingly.

- **Add some debug messages to vendored installer code.**

This will hopefully give us some breadcrumbs to track *why* the DNS cleanup code is not being executed.

Because this is hacked into vendored code, it will break `verify`.

The plan is to ignore the CI failures and just merge this. AppSRE will build an image tagged at commit X, which we will explicitly deploy to stg/int via sha.

2uasimojo:
CI failures are as expected. Merging...

PR #2103, opened 2023-08-31
AWS deprovision: Plumb ClusterDomain

2uasimojo:
ClusterUninstaller.ClusterDomain has existed for a long time, but was never used before the advent of shared VPC. In this new flow, if it is absent, DNS entries in the private hosted zone are not cleaned up in the destroy flow. This commit adds a new CLI parameter, `--cluster-domain`, to `hiveutil aws-tag-deprovision`. It is plumbed through by adding `spec.BaseDomain` to the ClusterDeprovision CRD and constructing the value by joining that to the `spec.ClusterName`.

TODO: Our platform-specific ClusterDeprovision sub-structs for Alibaba and IBMCloud each have a `BaseDomain` field. We should transition those to using this new platform-agnostic field -- but later. See HIVE-2298.

HIVE-2297

### DIFF
--- a/apis/hive/v1/clusterdeprovision_types.go
+++ b/apis/hive/v1/clusterdeprovision_types.go
@@ -20,6 +20,9 @@ type ClusterDeprovisionSpec struct {
 	// cluster is useful.
 	ClusterName string `json:"clusterName,omitempty"`
 
+	// BaseDomain is the DNS base domain.
+	BaseDomain string `json:"baseDomain,omitempty"`
+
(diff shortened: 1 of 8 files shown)


2uasimojo:
As this builds on #2100, which includes a couple of CI-breaking deltas, we're going to need to merge this manually.

jharrington22:
LGTM

2uasimojo:
`coverage`, `unit`, and `verify` failz are as expected from #2102.

/override ci/prow/coverage ci/prow/unit ci/prow/verify

PR #2106, opened 2023-09-01
Fix grafana dashboard

suhanime:
When exporting the JSON of an existing dashboard, it was chosen with an option `Export externally`, which has made it incompatible with the predefined variables that are define in the app-interface repo.


suhanime:
Wouldn't affect any tests. Force merging.

PR #2109, opened 2023-09-06
Revert builder.Unreachable() and add reachable check to builder.Build()

lleshchi:
With the controller-runtime upgrade from 8cfdcca, our remote client
`Builder` interface's `Build()` method got lazier. Before, if the remote
client was unreachable, `Build()` itself would fail. Now it won't fail
until the resulting client is actually used. This resulted in our
"unreachable" controller incorrectly marking the ClusterDeployment's
APIURLOverride as reachable when it wasn't, which had the ripple effect
of making the clustersync controller fail early by trying to use that
URL rather than falling back to the default.

This was exposed in situations where the DNS for the APIURLOverride was
being configured via a syncset: the clustersync controller tried to use
a URL that wouldn't resolve, so it couldn't sync the resource that would
make that DNS work.

Initially, to fix this issue, a `Reachable()` method was added to the `Builder`,
but on further inspection, there are many cases where uses of the `Build()`
method result in a client being falsely marked as reachable.

To fix, this PR reverts the `Reachable()` commit 576401b8d
and moves the reachability check to the `Build()` method of the `Builder` interface.
This internally builds the client and then uses it to query the `/apis` of the remote client.

xref: [HIVE-2300](https://issues.redhat.com//browse/HIVE-2300)

### DIFF
--- a/pkg/controller/unreachable/unreachable_controller.go
+++ b/pkg/controller/unreachable/unreachable_controller.go
@@ -198,7 +198,7 @@ func (r *ReconcileRemoteMachineSet) Reconcile(ctx context.Context, request recon
 	updateUnreachable := true
 	var primaryErr error
 	// Attempt to connect to the remote cluster using the preferred API URL.
-	primaryErr = remoteClientBuilder.UsePrimaryAPIURL().Reachable()
+	_, primaryErr = remoteClientBuilder.UsePrimaryAPIURL().Build()
 	if primaryErr != nil {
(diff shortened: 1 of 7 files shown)


lleshchi:
Need to add to kubeconfigBuilder as well

2uasimojo (commented):
This looks great. Just one thing: can you add a comment above the new checks explaining what they're doing? So future-we don't have to rely on `git blame`.

2uasimojo:
/test e2e-azure

Ignore me, drive by.

/test coverage

infra flake

PR #2115, opened 2023-09-14
MachinePool (GCP): Add NetworkProjectID 

2uasimojo:
Add NetworkProjectID to the GCP Platform section of MachinePool and plumb it through to the generated MachineSets. This is in support of XPN (shared VPC).

[HIVE-2310](https://issues.redhat.com//browse/HIVE-2310)

### DIFF
--- a/apis/hive/v1/gcp/machinepools.go
+++ b/apis/hive/v1/gcp/machinepools.go
@@ -13,6 +13,11 @@ type MachinePool struct {
 	//
 	// +optional
 	OSDisk OSDisk `json:"osDisk"`
+
+	// NetworkProjectID specifies which project the network and subnets exist in when
+	// they are not in the main ProjectID.
+	// +optional
+	NetworkProjectID string `json:"networkProjectID,omitempty"`
 }
 
 // OSDisk defines the disk for machines on GCP.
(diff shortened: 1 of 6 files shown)


apis/hive/v1/gcp/machinepools.go
```diff
@@ -13,6 +13,18 @@ type MachinePool struct {
...
 	// +optional
 	OSDisk OSDisk `json:"osDisk"`
+
+	// NetworkProjectID specifies which project the network and subnets exist in when
+	// they are not in the main ProjectID.
+	// +optional
+	NetworkProjectID string `json:"networkProjectID,omitempty"`
+
+	// ServiceAccount is the email of a gcp service account to be used for shared
+	// vpn installations. The provided service account will be attached to nodes
```
abutcher, line 23:
```suggestion
	// vpc installations. The provided service account will be attached to nodes
```

abutcher (changes requested):
Plumbing looks good to me. Added one nit on a comment.

apis/hive/v1/gcp/machinepools.go
2uasimojo: I did notice this -- but copied from the upstream struct so I left it alone :shrug:

2uasimojo:
/hold

- testing
- not sure if we want to *add* or *replace* the service account...

apis/hive/v1/gcp/machinepools.go
2uasimojo: No longer applicable -- ServiceAccount isn't part of the feature anymore.

jnpacker:
ACK. projectID on the networkInterface is what OSD GCP Shared VPC needs for the Infra nodes.

PR #2114, opened 2023-09-13
Register OpenStackProviderConfig/v1alpha1 with master scheme

2uasimojo:
This was missed in #2052 / 26a0b814.

[HIVE-2308](https://issues.redhat.com//browse/HIVE-2308)

PR #2117, opened 2023-09-18
OpenstackProviderSpec moved to machine v1alpha1

2uasimojo:
Since the recent revendor (#2052 / 26a0b81), OpenstackProviderSpec (embedded in MachineSets under OpenStack) stopped being able to unmarshal. This is because its schema moved from
github.com/openshift/cluster-api-provider-openstack to github.com/openshift/api/machine/v1alpha1 and the revendor picked up this move from the installer code: see
https://github.com/openshift/installer/pull/6382

This commit moves our references accordingly to match that upstream code.

Quirk 1: Strangely, this one type is in o/api/machine v1alpha1 even though everything else in o/api/machine is in v1beta1.

Quirk 2: o/api/machine v1alpha1's registration methods don't actually register the OpenstackProviderSpec type. (Why??) So we have to register it explicitly, as the installer does.

This more or less reverts the prior attempt at #2114 / 1f1ec2c0, which turned out to be completely redundant (the added registration call ended up in the same place as the one on the previous line).

[HIVE-2308](https://issues.redhat.com//browse/HIVE-2308)


2uasimojo:
> I want to add some UT for this niche

Well, that turns out not to be possible. The [existing UT](https://github.com/openshift/hive/blob/43376d8aac4adf84cd25b746344948d7fcbd8507/pkg/controller/machinepool/openstackactuator_test.go#L21) for this code path is disabled, noted as broken, because the installer's MachineSets() func tries to [call into OpenStack](https://github.com/openshift/hive/blob/3c31b4d7a61919ee62daf4eca4fc53951226b06d/vendor/github.com/openshift/installer/pkg/asset/machines/openstack/machinesets.go#L39).

We can't mock any part of this because there are no interfaces.

We can't bypass it and call [generateProviderSpec](https://github.com/openshift/hive/blob/3c31b4d7a61919ee62daf4eca4fc53951226b06d/vendor/github.com/openshift/installer/pkg/asset/machines/openstack/machinesets.go#L61) directly because it's private.

And that's as low as we could go while still achieving the goal of matching the GroupVersion of the returned OpenstackProviderSpec.

Le sigh.

/hold cancel

PR #2118, opened 2023-09-19
Add powerState label to metricClusterDeploymentsTotal

lleshchi:
xref: https://issues.redhat.com/browse/HIVE-2313

Add label in order to count number of CDs with a given powerState.

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -33,7 +33,7 @@ var (
 	metricClusterDeploymentsTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "hive_cluster_deployments",
 		Help: "Total number of cluster deployments.",
-	}, []string{"cluster_type", "age_lt"})
+	}, []string{"cluster_type", "age_lt", "power_state"})
 	metricClusterDeploymentsInstalledTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "hive_cluster_deployments_installed",
 		Help: "Total number of cluster deployments that are successfully installed.",
@@ -495,8 +495,8 @@ type clusterAccumulator struct {
 	ageFilter    string
 	ageFilterDur time.Duration
 
-	// total maps cluster type to counter.
-	total map[string]int
+	// total maps powerState to cluster type to counter.
+	total map[string]map[string]int
 
 	// deprovisioning maps a "greater than" duration string (i.e. 8h) to
 	// cluster type to counter. Specify 0h if you want a bucket for the smallest duration.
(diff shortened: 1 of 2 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -645,8 +650,18 @@ func (ca *clusterAccumulator) processCluster(cd *hivev1.ClusterDeployment) {
...
-		total.WithLabelValues(k, ca.ageFilter).Set(float64(v))
+		for clusterType := range ca.clusterTypesSet {
+			if count, ok := v[clusterType]; ok {
+				total.WithLabelValues(clusterType, ca.ageFilter, string(k)).Set(float64(count))
+			} else {
+				// We need to potentially clear out old cluster types no longer showing in the list.
+				// This will work so long as there is at least one cluster of that type still remaining
+				// in hive somewhere.
+				total.WithLabelValues(clusterType, ca.ageFilter, string(k)).Set(float64(0))
+			}
```
2uasimojo, line 661: I see where you're following the pattern of the existing metrics, but I think we might not need this conditional, since a map returns the zero value of the member type when the key doesn't exist.
```suggestion
			total.WithLabelValues(clusterType, ca.ageFilter, string(k)).Set(float64(v[clusterType]))
```

That said, there's no need to "fix" it here -- may as well roll them all up as tech debt for later.

It's also vaguely disturbing that, if we delete the last cluster of a given `clusterType`, its count will remain whatever the last reported value was. But again, that's latent and need not be addressed here.

2uasimojo (commented):
Seems sane to me.

/lgtm

but

/hold

as I'd like @suhanime to have a gander as well.

pkg/controller/metrics/metrics.go
suhanime: +1 to Eric's suggested change, but the reason is - [this](https://github.com/openshift/hive/blob/master/pkg/controller/metrics/metrics.go#L193:L196) ensures [Start method](https://github.com/openshift/hive/blob/master/pkg/controller/metrics/metrics.go#L220) is run every 2 minutes, so the values are going to be reset anyways :)

pkg/controller/metrics/metrics.go
```diff
@@ -599,10 +603,11 @@ func (ca *clusterAccumulator) processCluster(cd *hivev1.ClusterDeployment) {
 	}
 
 	clusterType := GetLabelValue(cd, hivev1.HiveClusterTypeLabel)
-	ca.ensureClusterTypeBuckets(clusterType)
+	powerState := cd.Status.PowerState
```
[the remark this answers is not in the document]
2uasimojo: Ah yes, thanks for catching that. As you noted below, we should map empty string to the default const.

There are no pointer types in the cd.status.powerState path, so we shouldn't need to check for `nil` anywhere.

/lgtm cancel

2uasimojo:
Yah good.

/lgtm

PR #2126, opened 2023-10-06
Fix TaintIdentifier fields in applyconfiguration

2uasimojo:
Still don't know what applyconfiguration is or does, but `make update` made this change that looks like it should have been swept up in #2112.

### DIFF
--- a/pkg/client/applyconfiguration/hive/v1/taintidentifier.go
+++ b/pkg/client/applyconfiguration/hive/v1/taintidentifier.go
@@ -9,8 +9,8 @@ import (
 // TaintIdentifierApplyConfiguration represents an declarative configuration of the TaintIdentifier type for use
 // with apply.
 type TaintIdentifierApplyConfiguration struct {
-	Key    *string         `json:"key:omitempty,omitempty"`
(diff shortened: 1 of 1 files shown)


suhanime:
/lgtm

I can't believe I forgot to run the update. Maybe we should add it to verify...

This needs to be backported to the MCE branch too.

2uasimojo:
> I can't believe I forgot to run the update. Maybe we should add it to verify...

Agree; I would have expected it to be caught up in that automatically.

> This needs to be backported to the MCE branch too.

Yah.

2uasimojo:
> > I can't believe I forgot to run the update. Maybe we should add it to verify...
> 
> Agree; I would have expected it to be caught up in that automatically.

See #2131 

> > This needs to be backported to the MCE branch too.
> 
> Yah.

Will do that along with the above once it merges.

2uasimojo:
> > > This needs to be backported to the MCE branch too.
> > 
> > 
> > Yah.
> 
> Will do that along with the above once it merges.

Covered by https://github.com/openshift/hive/pull/2134 (which doesn't technically backport this PR, but makes everything copacetic via a new-and-improved `make update vendor`).

PR #2127, opened 2023-10-06
clustersync: fix pod security

2uasimojo:
Get rid of annoying log messages and potential future inability to stand up clustersync pods.


pjanouse:
Guys, it seems this MR introduced an issue reported as https://github.com/openshift/hive/issues/2145.

PR #2129, opened 2023-10-10
DNM: Test e2e-gcp required-ness

2uasimojo:
We're trying to make e2e-gcp (and -azure) mandatory iff they are triggered: https://github.com/openshift/release/pull/44171

This commit does two things:
- No-op change to a GCP-specific source file to cause e2e-gcp to be triggered.
- Change to postinstall that should make e2e-gcp fail

The idea will be to `lgtm` this PR to demonstrate that it doesn't merge!

(Then we'll want to watch a normal PR that we _do_ want to merge and make sure that not running e2e-gcp at all doesn't block it.)

### DIFF
--- a/pkg/controller/machinepool/gcpactuator.go
+++ b/pkg/controller/machinepool/gcpactuator.go
@@ -222,7 +222,7 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 			return nil, false, errors.Wrap(err, "compute pool not providing list of zones and failed to fetch list of zones")
 		}
 		if len(zones) == 0 {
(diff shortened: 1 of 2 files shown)


2uasimojo:
except dangit, my change didn't cause e2e-gcp to fail. Let's try again...

2uasimojo:
Good, e2e-gcp failure as triggered. To save time & resources, and since the test result isn't actually relevant, let's do this:

/override ci/prow/e2e

and

/assign @dlom 

Again, what we're hoping is that prow will *not* merge this once it has `/lgtm` because e2e-gcp should be mandatory.

If it does merge, we'll just have to revert it quick :innocent:

2uasimojo:
> tide Pending — Not mergeable. Job ci/prow/e2e-gcp has not succeeded.

Score.

PR #2130, opened 2023-10-10
Ensure stale cluster deployment metrics are not reported

lleshchi:
[HIVE-2313](https://issues.redhat.com/browse/HIVE-2313)

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -242,6 +242,8 @@ func (mc *Calculator) Start(ctx context.Context) error {
 		if err != nil {
 			log.WithError(err).Error("error listing cluster deployments")
 		} else {
+			// Reset metric on each pass to prevent reporting stale metrics when count drops to zero
+			metricClusterDeploymentsTotal.Reset()
 			accumulator, err := newClusterAccumulator(infinity, []string{"0h", "1h", "2h", "8h", "24h", "72h"})
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/metrics/metrics.go
```diff
@@ -242,6 +242,7 @@ func (mc *Calculator) Start(ctx context.Context) error {
 		if err != nil {
 			log.WithError(err).Error("error listing cluster deployments")
 		} else {
+			metricClusterDeploymentsTotal.Reset()
```
suhanime, line 245: Include a note on why we're doing this. And do any other clusterAccumulator metrics look like they might need it? 

Explainer:
Most of the clusterAccumulator part was written at a time when we did not require QE to ensure metrics stuff works as expected, so even though they caught the bug with your changes, it might also apply to other metrics that follow the same code path.

2uasimojo: > Include a note on why we're doing this.

+1

> And do any other clusterAccumulator metrics look like they might need it?

Agree, but let's do that via a separate card & PR please.

lleshchi: > And do any other clusterAccumulator metrics look like they might need it?

I started evaluating this yesterday and will create a card for it. So far, I have found at least one other metric which also needs to be cleared.

2uasimojo (commented):
/lgtm

Thanks for the work here (most of which is not reflected by the size of this PR!)

PR #2131, opened 2023-10-10
Fix `make {update|verify}[-codegen]`

2uasimojo:
First:

`verify-codegen.sh` was a thin wrapper around `update-codegen.sh` that
passed a `--verify-only` flag to the generators. But generators wasn't
the only thing in `update-codegen.sh`, which led to inconsistencies.
This commit changes the strategy entirely:

We're getting rid of the `verify-codegen.sh` wrapper. Instead
`verify-codegen` will run `update-codegen` and then check if anything
changed. (This is similar to how `verify-vendor` works.)

Second:

Apparently the generators invoked by `update-codegen.sh` used to put the
generated files under vendor/, so we had an extra line to copy them back
to the non-vendor/ side. At some point this seems to have changed such
that the generators put the files in non-vendor/. So that copy was
actually _undoing_ what the generators had just done.

We're getting rid of that copy-back entirely. So now, `make
update-codegen` shouldn't touch the vendor/ directory; and `verify` will
catch discrepancies here (i.e. you forgot to run `make vendor`) via
`verify-vendor` rather than in `verify-codegen` itself (which wasn't
working anyway).

Third:

`make update-codegen` wouldn't fail if something in it went wrong. Add
`-e` to `update-codegen.sh` so it does. (This is _almost_ totally
unrelated to the rest of what's going on here.)

Finally:

Check in the changed files that are now properly generated.


2uasimojo:
In putting this together, I made the first commit, then ran `make verify` to prove that it would fail. Checked in the files changed thereby -- which were just the non-vendor/ ones! -- and ran `make verify` again to prove that it would _still_ fail. Ran `make vendor` to get them all copied down to vendor/, checked those in, and then `make verify` ran cleanly.

PR #2132, opened 2023-10-11
Add labels to ClusterSyncFailing metric

suhanime:
- Add label that reports the status of unreachable condition for the corresponding ClusterDeployment
- Add optional label support
  - Use the existing implementation for dynamic labels wherever possible
  - MustNewConstMetric used to report this metric via custom collector does not accept a map of label-value pairs. So supply a slice of values for the corresponding labels instead. To achieve this, maintaining the order of labels used while defining the metric is important.
    - Add a wrapper struct around dynamic labels which, in addition to the fixed and optional labels, also stores the labelList
    - Add buildLabelSlice method which is logically similar to buildLabels, but instead of reporting a map of labels and values, uses the labelList and returns a slice of label values to be used while observing the metric.
  - Add some unit tests
- Some minor corrections to get rid of the yellow squiggly lines.

xref: https://issues.redhat.com/browse/HIVE-2286

/assign @2uasimojo

### DIFF
--- a/pkg/controller/metrics/custom_collectors.go
+++ b/pkg/controller/metrics/custom_collectors.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -33,8 +34,8 @@ var (
 type provisioningUnderwayCollector struct {
 	client client.Client
 
-	// minDuration, when non-zero, is the minimum duration afer which clusters provisioning
-	// will start becomming part of this metric. When set to zero, all clusters provisioning
+	// minDuration, when non-zero, is the minimum duration after which clusters provisioning
+	// will start becoming part of this metric. When set to zero, all clusters provisioning
 	// will be included in the metric.
 	minDuration time.Duration
 
@@ -43,7 +44,7 @@ type provisioningUnderwayCollector struct {
 	metricClusterDeploymentProvisionUnderwaySeconds *prometheus.Desc
 }
 
-// collects the metrics for provisioningUnderwayCollector
+// Collect collects the metrics for provisioningUnderwayCollector
 func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 	ccLog := log.WithField("controller", "metrics")
 	ccLog.Info("calculating provisioning underway metrics across all ClusterDeployments")
@@ -132,7 +133,7 @@ type provisioningUnderwayInstallRestartsCollector struct {
 	metricClusterDeploymentProvisionUnderwayInstallRestarts *prometheus.Desc
 }
 
-// collects the metrics for provisioningUnderwayInstallRestartsCollector
+// Collect collects the metrics for provisioningUnderwayInstallRestartsCollector
 func (cc provisioningUnderwayInstallRestartsCollector) Collect(ch chan<- prometheus.Metric) {
 	ccLog := log.WithField("controller", "metrics")
 	ccLog.Info("calculating provisioning underway install restarts metrics across all ClusterDeployments")
@@ -249,7 +250,7 @@ type deprovisioningUnderwayCollector struct {
 	metricClusterDeploymentDeprovisionUnderwaySeconds *prometheus.Desc
 }
 
-// collects the metrics for deprovisioningUnderwayCollector
+// Collect collects the metrics for deprovisioningUnderwayCollector
 func (cc deprovisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 	ccLog := log.WithField("controller", "metrics")
 	ccLog.Info("calculating deprovisioning underway metrics across all ClusterDeployments")
@@ -306,7 +307,7 @@ func newDeprovisioningUnderwaySecondsCollector(client client.Client) prometheus.
 type clusterSyncFailingCollector struct {
 	client client.Client
 
-	// minDuration, when non-zero, is the minimum duration afer which clustersync failure
+	// minDuration, when non-zero, is the minimum duration after which clustersync failure
 	// will start becoming part of this metric. When set to zero, all clustersync failures
 	// will be included in the metric.
 	minDuration time.Duration
@@ -314,9 +315,14 @@ type clusterSyncFailingCollector struct {
 	// metricClusterSyncFailingSeconds is a prometheus metric for the number of seconds
 	// between the start of a failing clustersync and now.
 	metricClusterSyncFailingSeconds *prometheus.Desc
+
+	// dynamicLabels is a collection of fixed (mandatory) and optional labels for the clusterSyncFailing metric.
+	// It also has labelList which stores the order of labels as defined in Desc. MustNewConstMetric doesn't
+	// accept a map of labels and values, so maintaining the order is needed while observing the metric.
+	dynamicLabels dynamicLabelsForCustomCollectors
 }
 
-// collects the metrics for custerSyncFailingCollector
+// Collect collects the metrics for custerSyncFailingCollector
 func (cc clusterSyncFailingCollector) Collect(ch chan<- prometheus.Metric) {
 	ccLog := log.WithField("controller", "metrics")
 	ccLog.Info("calculating cluster sync failing seconds metrics")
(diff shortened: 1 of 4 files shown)


pkg/controller/metrics/custom_collectors.go
```diff
@@ -328,19 +334,40 @@ func (cc clusterSyncFailingCollector) Collect(ch chan<- prometheus.Metric) {
...
+			for _, cd := range clusterDeployments.Items {
+				// Expect to find the corresponding ClusterDeployment. In the rare case it got garbage collected by the time
+				// we fetched the list of ClusterSyncs, do not report the metric.
+				if cd.Namespace == cs.Namespace {
+					// Ignore uninstalled ClusterDeployments as it wouldn't have all conditions initialized.
+					if unreachableCondition := controllerutils.FindCondition(cd.Status.Conditions, hivev1.UnreachableCondition); unreachableCondition != nil {
+						fixedLabels := make(map[string]string, len(cc.dynamicLabels.fixedLabels))
+						fixedLabels["namespaced_name"] = cs.Namespace + "/" + cs.Name
+						fixedLabels["unreachable"] = string(unreachableCondition.Status)
+						labelValues := cc.dynamicLabels.buildLabelSlice(fixedLabels, &cd)
```
2uasimojo, line 357: So yeah, this is uncomfortable, because we're relying on two disparate bits of code agreeing on something (the exact set of dynamic label keys) without enforcement. If we have a typo in one place or another, things could go weird, possibly in a difficult-to-notice-or-debug way.

I'll spare you where my brain went in the interim, but I think we could close the gap sufficiently by adding logic to `buildLabelSlice` to blow up if the input map contains an unrecognized key. Thoughts?

pkg/controller/metrics/custom_collectors_test.go
```diff
@@ -691,32 +696,104 @@ func TestClusterSyncCollector(t *testing.T) {
...
 			},
-			min:       0 * time.Hour,
-			expected1: []string{"namespaced_name = test-namespace/test-name"},
+			min:            0 * time.Hour,
+			optionalLabels: map[string]string{},
+			expected1:      []string{"namespaced_name = test-namespace/test-name unreachable = True"},
+			expected2:      []string(nil),
+		},
+		{
+			name: "skip reporting when ClusterDeployment not found",
```
2uasimojo, line 741: Let's add one for when the CD exists, but has no `unreachable` condition.

pkg/controller/metrics/custom_collectors.go
```diff
@@ -328,19 +334,40 @@ func (cc clusterSyncFailingCollector) Collect(ch chan<- prometheus.Metric) {
...
+							ch <- prometheus.MustNewConstMetric(
+								cc.metricClusterSyncFailingSeconds,
+								prometheus.GaugeValue,
+								seconds,
+								labelValues...,
+							)
+						}
+					}
+				}
+				continue
```
2uasimojo, line 370: redundant?

pkg/controller/metrics/custom_collectors.go
suhanime: I'm not fond of panicking while reporting a metric. I have tweaked the logic a little bit, so we will report the default value for all the labels not set in the calling method, and essentially ignore the other values that could be there. And we have the unit tests to ensure code works as expected. Does that seem acceptable?

pkg/controller/metrics/custom_collectors.go
```diff
@@ -328,19 +334,40 @@ func (cc clusterSyncFailingCollector) Collect(ch chan<- prometheus.Metric) {
...
+						// check if duration crosses the threshold
+						if cc.minDuration.Seconds() <= seconds {
+							ch <- prometheus.MustNewConstMetric(
+								cc.metricClusterSyncFailingSeconds,
+								prometheus.GaugeValue,
+								seconds,
+								labelValues...,
+							)
+						}
+					}
```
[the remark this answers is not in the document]
suhanime: Made the relevant changes.

pkg/controller/metrics/custom_collectors.go
suhanime: A mistake actually, it was supposed to be `break`, to stop the looping through all the remaining cluster deployments once the appropriate one has already been found.

pkg/controller/metrics/custom_collectors_test.go
suhanime: Added, along with ones for the clusterdeployment not found.

suhanime (commented):
@2uasimojo this is ready for another pass

pkg/controller/metrics/metrics_with_dynamic_labels.go
```diff
@@ -73,6 +82,30 @@ func (d *dynamicLabels) buildLabels(fixedLabels map[string]string, obj metav1.Ob
...
+// that can be used for observing metrics that cannot accept a map of labels and values.
+// The resultant slice will follow the same order as the labelList to ensure
+// correct values are reported for their corresponding labels.
+func (d *dynamicLabelsForCustomCollectors) buildLabelSlice(fixedLabels map[string]string, obj metav1.Object) []string {
+	labelValues := make([]string, len(d.labelList))
+
+	for index, label := range d.labelList {
+		// Add values for fixed labels
+		if index < len(fixedLabels) {
+			if value, ok := fixedLabels[label]; ok {
```
2uasimojo, line 95: I don't feel like this really helped us. In fact, it's a step backward, because now a legitimate label with empty string as the value will show up as empty string rather than `unspecified`. (Let's have a test case for that, please.)

I agree panicking for an unknown label key is not a good look. (If we could do it at register time, I'd be all for it.)

So then I can no longer
> spare you where my brain went in the interim

:innocent: 

Let's define a `type` for `fixedLabels` with explicit setters for each fixed label; and one getter that produces the `labelValues` slice.

pkg/controller/metrics/custom_collectors.go
```diff
@@ -328,18 +334,40 @@ func (cc clusterSyncFailingCollector) Collect(ch chan<- prometheus.Metric) {
...
+			var cdRef hivev1.ClusterDeployment
+			for _, cd := range clusterDeployments.Items {
+				if cd.Namespace == cs.Namespace {
+					cdRef = cd
+					break
+				}
+			}
+			fixedLabels := make(map[string]string, len(cc.dynamicLabels.fixedLabels))
+			fixedLabels["namespaced_name"] = cs.Namespace + "/" + cs.Name
+			if &cdRef != nil {
```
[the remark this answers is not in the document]
suhanime: - You were correct - I had 2 mistakes here, first was comparing the address instead of the value, and second was to compare the struct directly to nil. After a deep-dive of Laws of Reflection, I believe I have the correct nil/empty comparison now 
- Incorrect, `cdRef` is declared everytime we loop over a cs - aka, it is a temporary variable sparing the length of 1 iteration of clusterSync
- Cool - fixed.
- True, that's why I break

See - usually I would agree, but we're already storing all the clusterdeployments. Do I want to build another map of essentially the same information (I see how it would have less memory because of string and pointers, but it would still have O(n) extra storage), or would I have to loop through them a few times - which is equivalent to the number of failing clustersyncs? I think I prefer the latter.

pkg/controller/metrics/metrics_with_dynamic_labels.go
suhanime: Good catch. Fixed it.
fixed and dynamic labels here are basically encapsulated version of what's already present for other metrics which have dynamic labels but are not reported via a custom collectors. Having explicit setters would require touching that interface, and I'm not sure if I'm prepared to go there as a part of this PR. The comparison I fixed mimics the present [buildLabels](https://github.com/openshift/hive/blob/master/pkg/controller/metrics/metrics_with_dynamic_labels.go#L64). I think it should be enough. Also, fwiw, the possibility of the current implementation reaching a stage where any of the fixedLabels for failing clustersyncs is passed to this method with empty string set - is none.

pkg/controller/metrics/custom_collectors.go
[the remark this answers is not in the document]
2uasimojo: > After a deep-dive of Laws of Reflection, I believe I have the correct nil/empty comparison now

Ack, but see new inline comment.

> * Incorrect, `cdRef` is declared everytime we loop over a cs - aka, it is a temporary variable sparing the length of 1 iteration of clusterSync

Ah, right, my mistake.

> * True, that's why I break
> 
> See - usually I would agree, but we're already storing all the clusterdeployments. Do I want to build another map of essentially the same information (I see how it would have less memory because of string and pointers, but it would still have O(n) extra storage), or would I have to loop through them a few times - which is equivalent to the number of failing clustersyncs? I think I prefer the latter.

As you say, it's not a lot of additional storage if the values are pointers. It's a tradeoff between memory and CPU. Digging in a bit:
In OSD prod, a typical "ns/name" will be in the neighborhood of 64 bytes. A pointer is 8. Let's say for the sake of discussion that the map struct overhead rounds it out to 100. An OSD prod shard has ~600 clusters. So we're talking about building a struct on the order of 60KB via ~600 string copies and pointer assignments.

Now, I did miss that the outer loop is already filtered to clustersyncs that are failing. On p02 right now that's 44/532. Assuming we have to search on average half of the CD list to find a match, that's ~12000 loop iterations with string compares.

So in our little hypothetical, cutting the processing by a factor of ~20 costs us ~60KB. Which is more expensive, the apples or the oranges? :shrug: 

----

But we can actually do better than both:
- Iterate over the ClusterSyncs once, generating a `map[ns/name]{labelValues, seconds}`-ish.
- Iterate over the CDs once, looking for ns/name matches to the above (each of which is O(1) because map) and filling in `labelValues[unreachable]`
- Iterate over the map, pushing the metric to prom (and filling in `unreachable=unspecified` if the CD wasn't found in the previous loop)

This is O(1) -- more specifically {#CSes + #CDs + #failingCSes} -- and in our hypothetical costs something like 4KB of extra memory.

pkg/controller/metrics/metrics_with_dynamic_labels.go
2uasimojo: I'm okay leaving this to a fup. But I think it would be good to do, to future-proof against developer mistakes.

2uasimojo:
/lgtm

Let's make the performance improvement in a followon PR.

I think similar opportunities may be available in the other custom collectors...

PR #2140, opened 2023-10-17
Annotation to hold uninstall pod

2uasimojo:
To destroy a cluster, you delete the ClusterDeployment. Hive responds as follows:
- Create a ClusterDeprovision
- Create a Job
- ...which creates a Pod
- ...which has a Container
- ...which runs the installer's destroy code

Under normal circumstances, when the destroy process finishes, the Pod gets garbage collected very quickly -- often before logs can even finish streaming, if they're being watched. This can make debugging uninstall issues difficult.

With this commit, we add support for an annotation on ClusterDeployment:

`hive.openshift.io/hold-uninstall-pod: true`

This annotation is copied to the ClusterDeprovision, and when hive sees it there, it will add that key (`hive.openshift.io/hold-uninstall-pod`) as a finalizer to the uninstall pod. This will cause it to remain in `Terminating` state until the finalizer is removed, *which must be done "manually" by the consumer*.

[HIVE-2309](https://issues.redhat.com//browse/HIVE-2309)

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -523,6 +523,11 @@ const (
 	// for forcing rollout of such changes on the target cluster -- e.g. by deleting the Machines -- as the machine
 	// config operator will not do so.
 	OverrideMachinePoolPlatformAnnotation = "hive.openshift.io/override-machinepool-platform"
+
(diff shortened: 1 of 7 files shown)


2uasimojo:
We're tabling this effort for now. See card for details.

PR #2144, opened 2023-10-17
golang.org/x/net v0.17.0

2uasimojo:
CVE-2023-44487

[HIVE-2337](https://issues.redhat.com//browse/HIVE-2337)


2uasimojo:
Okay, so clearly there's more to this than I thought based on the original reports. Once done scanning dependencies, snyk looks for stuff in our repo (including the vendor dir). We're going to need a way to disable/bypass/silence at least some of these -- e.g. "hardcoded credentials" in test suites are not an actual problem.

But as far as this PR goes, it resolves the vulns in x/net, so it should be good to land.

PR #2151, opened 2023-10-20
GCP deprovision: Plumb NetworkProjectID for shared VPC

2uasimojo:
The shared VPC flow added the need for the installer to use a NetworkProjectID to manage certain resources. This is provided at installation time via the install-config, and is included in the metadata.json produced by the installer when provisioning completes.

The uninstall process requires this NetworkProjectID during deprovisioning to perform the corresponding cleanup -- without it, those resources are leaked. So we need to plumb that datum through from the metadata.json to the uninstall process.

To make this more general-purpose for adoption and disaster recovery, and for retrofitting existing clusters that were created before this commit, we also attempt to pull the NetworkProjectID out of the install-config if it is unset.

NOTE: We will refuse to proceed with deprovision if this field is not set. Once this change lands, we should audit existing clusters to make sure they all get retrofitted; and manually retrofit any others (with the empty string).

[HIVE-2334](https://issues.redhat.com//browse/HIVE-2334)


2uasimojo:
/assign @lleshchi 

This very closely follows the example of #2100.

pkg/controller/clusterdeployment/clusterprovisions.go
```diff
@@ -367,6 +368,38 @@ func setAzureResourceGroupFromMetadata(cd *hivev1.ClusterDeployment, cm *hivev1.
...
+	if err := json.Unmarshal(pmjson, &im); err != nil {
+		logger.WithError(err).Error("Could not unmarshal ClusterMetadata!")
+		return
+	}
+
+	if im.GCP == nil {
+		logger.Warn("ClusterMetadata unexpectedly has no GCP section")
+		return
+	}
+	npid := im.GCP.NetworkProjectID
```
lleshchi, line 389: Does this section require a check for whether `im.GCP.NetworkProjectID` is not nil, which would be different from an empty string? (Assuming it's not necessary but would like to know why)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2466,6 +2467,62 @@ func (r *ReconcileClusterDeployment) discoverAzureResourceGroup(cd *hivev1.Clust
...
+// discoverGCPNetworkProjectID is intended to retrofit GCP clusters which were a) adopted, or b) created before
+// support for shared VPC was added. It attempts to discover the network project ID for the cluster by parsing
+// the install-config.
+// No op if platform is not GCP.
+// No op if the field is already set (we don't validate that it is correct).
+// This is best effort. If things go wrong, we log but do not fail out. THIS WILL BLOCK DEPROVISIONING. If
+// this happens, the field must be set manually.
+// The return indicates whether the field was modified. If `true`, the caller is responsible for pushing
+// an Update() back to the server.
+func (r *ReconcileClusterDeployment) discoverGCPNetworkProjectID(cd *hivev1.ClusterDeployment, log log.FieldLogger) bool {
```
lleshchi, line 2479: Where is this function used?

pkg/controller/clusterdeployment/clusterprovisions.go
2uasimojo: Is the field in installertypes.ClusterMetadata a pointer? I thought it was a string type.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: Did I forget the bit where it gets called? :sheepish:

pkg/controller/clusterdeployment/clusterprovisions.go
lleshchi: Oh, you're right. Makes sense.

2uasimojo: [ref](https://github.com/openshift/hive/blob/a9aaf72f8d1849aa5b5f2e3d6e137e93d9fbbdb5/vendor/github.com/openshift/installer/pkg/types/gcp/platform.go#L26)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: Fixed.

PR #2154, opened 2023-10-24
Add --network-project-id param to GCP destroyer command

2uasimojo:
This was *also* missed in #2151.

[HIVE-2334](https://issues.redhat.com//browse/HIVE-2334)


2uasimojo:
/test e2e-gcp

...since this actually changes the destroyer CLI for GCP. (In e2e we'll end up passing in the empty string, but we need to prove that works, as it'll be the default/common case anyway.)

2uasimojo:
/test e2e-gcp

infra flake

PR #2157, opened 2023-10-25
Minimal install mode (don't pull release or oc images)

2uasimojo:
Certain environments, like ARO:
- Always use InstallerImageOverride
- Don't care to must-gather spokes on install failures

The latter is the only thing we use the `oc` command for. And the release image is only used to discover the URI for the installer image and the `oc` image.

To reduce burdens like ensuring those images are available and secure, this commit introduces a (back door, not-officially-supported) "minimal install mode". Activate it by doing two things:

1. Set ClusterDeployment annotation `hive.openshift.io/minimal-install-mode: True`
2. Set `ClusterDeployment.Spec.Provisioning.InstallerImageOverride`

This will cause hive to wend its way through the provisioning flow without pulling the release or `oc` images. It will use the InstallerImageOverride as the source of the `openshift-install` binary.

NOTE: This introduces a few limitations/oddities:
- You cannot use release image verification. (There's no release image to verify.)
- Failed installs will not attempt to must-gather the spoke, even if you have the rest of the install log capture stuff configured.
- `ClusterDeployment.Status.CLIImage` will be set to `"<NONE>"`.
- `ClusterDeployment.Status.InstallVersion` will not be set. Consequently...
- The `cluster_version` label on the `hive_cluster_deployment_install_failure_total` and `hive_cluster_deployment_install_success_total` metrics will be `"unspecified"` for these clusters.

[HIVE-2208](https://issues.redhat.com//browse/HIVE-2208)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -972,6 +972,37 @@ const (
 func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDeployment, releaseImage string, cdLog log.FieldLogger) (*reconcile.Result, error) {
 	areImagesResolved := cd.Status.InstallerImage != nil && cd.Status.CLIImage != nil
 
+	// In minimal install mode, don't even run the image resolver pod -- just trust the InstallerImageOverride
+	if minimal, err := strconv.ParseBool(cd.Annotations[constants.MinimalInstallModeAnnotation]); err == nil && minimal {
+		changed1, changed2 := false, false
+		if !areImagesResolved {
+			// "resolve" the images via installerImageOverride
+			installerImage := cd.Spec.Provisioning.InstallerImageOverride
+			if installerImage == "" {
+				return &reconcile.Result{}, r.updateCondition(
+					cd, hivev1.InstallImagesNotResolvedCondition, corev1.ConditionTrue,
+					"MinimalInstallModeWithoutImageOverride",
+					"InstallerImageOverride is required when using minimal-install-mode annotation", cdLog)
+			}
+			cd.Status.InstallerImage = &installerImage
+			cd.Status.CLIImage = pointer.String("<NONE>")
+			changed1 = true
+		}
+		cd.Status.Conditions, changed2 = controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+			cd.Status.Conditions,
+			hivev1.InstallImagesNotResolvedCondition,
+			corev1.ConditionFalse,
+			imagesResolvedReason,
+			imagesResolvedMsg,
+			controllerutils.UpdateConditionIfReasonOrMessageChange,
+		)
+		var err error
+		if changed1 || changed2 {
+			err = r.statusUpdate(cd, cdLog)
+		}
+		return nil, err
+	}
+
 	jobKey := client.ObjectKey{Namespace: cd.Namespace, Name: imageset.GetImageSetJobName(cd.Name)}
 	jobLog := cdLog.WithField("job", jobKey.Name)
 
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -523,6 +523,11 @@ const (
 	// for forcing rollout of such changes on the target cluster -- e.g. by deleting the Machines -- as the machine
 	// config operator will not do so.
 	OverrideMachinePoolPlatformAnnotation = "hive.openshift.io/override-machinepool-platform"
+
+	// MinimalInstallModeAnnotation, if set to "true" on a ClusterDeployment along with InstallerImageOverride, asks hive
+	// to avoid downloading the release and oc images at all -- only the (overridden) installer image will be pulled.
+	// Side effects include: a) You can't use a release image verifier; b) We won't try to must-gather on the spoke.
+	MinimalInstallModeAnnotation = "hive.openshift.io/minimal-install-mode"
 )
 
 // GetMergedPullSecretName returns name for merged pull secret name per cluster deployment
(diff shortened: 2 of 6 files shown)


2uasimojo:
/hold

for testing

2uasimojo (commented):
/assign @suhanime

2uasimojo:
Got an offline ack from @bennerv 

/hold cancel

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -972,6 +972,28 @@ const (
...
+					"MinimalInstallModeWithoutImageOverride",
+					"InstallerImageOverride is required when using minimal-install-mode annotation", cdLog)
+			}
+			cd.Status.InstallerImage = &installerImage
+			cd.Status.CLIImage = pointer.String("<NONE>")
+			// We could count on InstallImagesNotResolvedCondition being True and let updateCondition sweep up this
+			// Status().Update(), but just in case...
+			if err := r.statusUpdate(cd, cdLog); err != nil {
+				return nil, err
+			}
```
suhanime, line 992: I don't think you need to call a status update here at all, since while exiting the function you're passing the cd object to set the condition, and that method always calls a status update.

2uasimojo: Yeah, I flip flopped a bit on that. The condition setter only updates if something changes in the condition. Which it *should* any time we’re in this code path. But I didn’t want to rely on it.

suhanime: Ah I missed that detail. Cool.

2uasimojo:
Actually

/hold

you’re right, there’s a failure mode here that I’d like to close: namely if the second update fails, we’ll never update the condition.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: So to dig in a bit...

Since `r.updateCondition()` internally decides whether to `Status().Update()` based (only!) on whether the condition changed, I couldn't use the usual pattern of tracking a `changed` var or three unless I wanted to change the call spec (and all 17 existing refs) or write some other func (which would be substantially identical except for that).

So my next thought was to pre-update the condition in such a way that `r.updateCondition()` would be guaranteed to change it -- i.e. force the status, reason, or message to something bogus. Obviously this is hacky, but I had to actually start writing it to realize *how* hacky :disappointed:.

However, I dislike the noted failure mode (if the first update succeeds but the second fails, we'll never hit this code path again, and the status condition will stay wrong forever) enough that I'm reconsidering the horrible hack.

...unless you have other ideas?

suhanime: These options are very hacky and at that point, we are making it confusing. `updateCondition` was written because we update so many cd conditions within the controller that it wraps the logic of setting the condition and calling the update to make it simpler for the caller Reconcile to safely exit. Since this is a special case, I recommend you directly call the controllerutils setter and explicitly update here as you exit.

2uasimojo: Yeah, that's fair. Will fix. Thanks!

2uasimojo: Done.

2uasimojo:
> you’re right, there’s a failure mode here that I’d like to close: namely if the second update fails, we’ll never update the condition.

Fixed!

/hold cancel

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -972,6 +972,40 @@ const (
...
+		}
+		conditions, changed2 := controllerutils.SetClusterDeploymentConditionWithChangeCheck(
+			cd.Status.Conditions,
+			hivev1.InstallImagesNotResolvedCondition,
+			corev1.ConditionFalse,
+			imagesResolvedReason,
+			imagesResolvedMsg,
+			controllerutils.UpdateConditionIfReasonOrMessageChange,
+		)
+		if changed2 {
```
suhanime, line 999: `SetClusterDeploymentConditionWithChangeCheck` is smart enough to take care of this, you can directly set cd.Status.Conditions at L991.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -972,6 +972,40 @@ const (
 func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDeployment, releaseImage string, cdLog log.FieldLogger) (*reconcile.Result, error) {
 	areImagesResolved := cd.Status.InstallerImage != nil && cd.Status.CLIImage != nil
 
+	// In minimal install mode, don't even run the image resolver pod -- just trust the InstallerImageOverride
+	if minimal, err := strconv.ParseBool(cd.Annotations[constants.MinimalInstallModeAnnotation]); err == nil && minimal {
+		changed1 := false
+		if !areImagesResolved {
```
suhanime, line 978: I recommend you add a unit test for clearing the existing `InstallImagesNotResolvedCondition` once the correct variables are set.

2uasimojo: Something other than [this](https://github.com/openshift/hive/pull/2157/files#diff-b9685aa0bf796d1db13613fe411ed287146777107405bcc935ebd1d45d64ea47R626-R632)?

2uasimojo: Oh, you mean if the variables are somehow set but the condition is not? Can do...

2uasimojo: ...except that's not going to work with the code as written, is it?

2uasimojo: Done. Good call.

2uasimojo:
/hold

Ben reports a NPE, investigating.

2uasimojo:
> Ben reports a NPE, investigating.

Sorted

/hold cancel

suhanime (commented):
/lgtm

PR #2160, opened 2023-10-27
Spoof Sleep() for installmanager_test

2uasimojo:
installmanager_test takes an annoyingly, unnecessarily long time because there are several `time.Sleep()`s in the installmanager code.

This commit moves the sleep function to the InstallManager struct so we can fake it out with a no-op in the test suite.


2uasimojo:
btw, before:
```
ok  	github.com/openshift/hive/pkg/installmanager	17.654s
```
after:
```
ok  	github.com/openshift/hive/pkg/installmanager	0.383s
```

PR #2173, opened 2023-11-27
Reset metrics on each pass of metrics controller

lleshchi:
[HIVE-2326](https://issues.redhat.com/browse/HIVE-2326)


2uasimojo:
Seems sane to me, but I would like Suhani to have a look -- in particular to validate that this makes sense for *all* of these metrics.

/assign @suhanime

PR #2175, opened 2023-11-30
Remove SyncSetInstance cleanup code

2uasimojo:
Remove the code to clean up legacy SyncSetInstance CRs and CRD, which were supplanted by ClusterSync via #1093 ~3y ago.

The consequence is that, if a customer upgrades from a pre-ClusterSync version of hive to a version at/after this commit, the SyncSetInstance CRs and CRD will be orphaned, unnecessarily cluttering etcd.

It has been long enough that we consider the probability of this to be quite low. If it does happen, there is a simple workaround: manually delete the CRs and CRD.

Why do this at all? I'm glad you asked. Since the SyncSetInstance CRD existed in the now-deprecated v1beta1 apiVersion of apiextensions.k8s.io, our API call to discover it was showing up on customers' radar and making them uncomfortable. At this point the risk of orphaning resources is the lesser evil.

[HIVE-2025](https://issues.redhat.com//browse/HIVE-2025)

### DIFF
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -17,9 +17,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


dlom:
I love it when I see a +0 PR

/lgtm

PR #2178, opened 2023-12-21
Read-only root file system

2uasimojo:
Add `readOnlyRootFilesystem` to our Deployments and StatefulSets. Why? Cause it's more securer.

One thing I needed to do to make this work was to add a volume for `/tmp`, which I guess is by default part of `/`. This resulted in an actual problem in hive-controllers creating a Session for AWS... but I did it for all the things just in case.

This scares me a little bit. For one thing, any time we mess with `securityContext`s we seem to break our friends running hive under vanilla k8s. But also, who knows what corner cases are expecting to be able to write somewhere in the root file system? We won't find out if, like, trying to use additional certs under OpenStack is a problem until some customer tries it.

[HIVE-2350](https://issues.redhat.com//browse/HIVE-2350)

### DIFF
--- a/config/clustersync/statefulset.yaml
+++ b/config/clustersync/statefulset.yaml
@@ -54,6 +54,14 @@ spec:
               fieldPath: metadata.name
         - name: HIVE_SKIP_LEADER_ELECTION
           value: "true"
+        - name: TMPDIR
+          value: /tmp
+        volumeMounts:
+        - name: tmp
+          mountPath: /tmp
+        securityContext:
+          privileged: false
+          readOnlyRootFilesystem: true
         livenessProbe:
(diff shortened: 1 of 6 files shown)


2uasimojo:
/test security

just to make sure we didn't pick up any others before I fix this

2uasimojo:
Glad I checked. Resolving the two new vulns via HIVE-2398 / #2180.

ngraham20 (approved):
/lgtm

2uasimojo:
/hold cancel

:crossed_fingers:

2uasimojo:
/override ci/prow/security

Same low-sev, not-yet-fixed vuln we've been tracking.

PR #2179, opened 2024-01-05
Reconcile MachineSets using MachinePool labels and AZs

lleshchi:
The MachinePool controller reconciles MachineSets by
generating the necessary MachineSets for a given MachinePool and
matching generated MachineSets with existing remote MachineSets to
determine what updates to MachineSet state should occur. Historically,
the matching of MachineSets was determined by comparing MachineSet
names. This process results in some edge cases, especially now that
VSphere has introduced prefixes to MachineSet names to represent
availability zones, which resulted in mismatching of MachineSet names
across hive upgrades, where the hive vendored installer code that
creates the generated MachineSets does not match the version which
created the remote MachineSets. This resulted in odd behavior where
duplicates of already existing MachineSets were being generated with
only a prefix discrepancy.

The solution to this mismatch dilemma is twofold:

For Day 0, the installmanager matches MachineSets to their
corresponding MachinePool and patches the MachineSet manifests with a
label indicating the name of the MachinePool they belong to.

For Day 2, the MachinePool controller checks the remote MachineSets for
a hive MachinePool label. If the label matches the current MachinePool
on which the controller is reconciling, the controller will check the
availability zone of the MachineSet (as there can be multiple
MachineSets per MachinePool with differing AZs), and if the AZ matches,
the two MachineSets are considered matched. This removes any need to
check or depend on the name of the remote MachineSets.

Important note: If a MachineSet does not have a MachinePool label, the
MachinePool controller *will not* manage or consider the MachineSet.
For manual creation or adoption of remote MachineSets, the
MachinePool label must be added manually to ensure management by Hive.

[HIVE-2254](https://issues.redhat.com/browse/HIVE-2254 )


2uasimojo (commented):
Oh, look, draft comments. Sorry about the delay.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -512,24 +518,64 @@ func (r *ReconcileMachinePool) ensureEnoughReplicas(
...
+	rSpec := rMS.Spec.Template.Spec
+	gSpec := gMS.Spec.Template.Spec
+	var err error
+	if rSpec.ProviderSpec.Value == nil || gSpec.ProviderSpec.Value == nil {
+		return false, errors.New("provider spec is nil")
+	}
+	if rSpec.ProviderSpec.Value.Raw == nil {
+		rSpec.ProviderSpec.Value.Raw, err = json.Marshal(rMS.Spec.Template.Spec.ProviderSpec.Value.Object)
+		if err != nil {
+			logger.WithError(err).Error("unable to marshal remote ms provider spec object to raw value")
```
2uasimojo, line 532: Any reason not to make this fatal right away? IIUC as written it'll ultimately fail around L545... but that might be a less scrutable error.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -512,24 +518,64 @@ func (r *ReconcileMachinePool) ensureEnoughReplicas(
...
+		}
+	}
+	if gSpec.ProviderSpec.Value.Raw == nil {
+		gSpec.ProviderSpec.Value.Raw, err = json.Marshal(gMS.Spec.Template.Spec.ProviderSpec.Value.Object)
+		if err != nil {
+			logger.WithError(err).Error("unable to marshal generate ms provider spec object to raw value")
```
2uasimojo, line 538: ditto

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -563,10 +609,18 @@ func (r *ReconcileMachinePool) syncMachineSets(
 	for i, ms := range generatedMachineSets {
 		found := false
 		for _, rMS := range remoteMachineSets.Items {
-			if matchAndMutate(pool, cd, ms, rMS) {
-				found = true
+			var err error
+			found, err = matchMachineSets(ms, rMS, logger)
+			if err != nil {
+				logger.WithError(err).Errorf("unable to match MachineSets %s and %s", ms.Name, rMS.Name)
+				continue // FIXME is this the correct behavior?
```
2uasimojo, line 616: Great question. Not all errors are created equal.

If we fail to match a rMS that we _should_ have matched, we'll end up creating a dup, and that's Bad™. In that case we would ideally want to fail the whole mpool reconcile.

BUT

If we fail the mpool due to an error parsing an _irrelevant_ rMS, it's kind of a DoS. (IOW a bad remote mset can bring down the whole mpool controller.) Ideally we would ignore that one and `continue`.

Like, ugh, but how about we figure a way to distinguish "labels matched, and the error happened after that". We can be fairly confident that label matching won't error (missing gMS label scenario notwithstanding), and it's reasonable to use that as a criterion for failing _this_ mpool reconcile.

So, let's give `matchMachineSets` a third return, `labelsMatched bool`, and use that to decide whether to error (`true`) or `continue` (`false`). With a nice code comment explaining the rationale.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -563,10 +609,18 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+			}
+			if found {
+				logger.Infof("Found matching remote MachineSet %s for generated MachineSet %s", rMS.Name, ms.Name)
 				objectModified := false
 				objectMetaModified := false
+				// Rename generated MachineSet if there is a name mismatch
+				ms.Name = rMS.Name
```
2uasimojo, line 623: Is it possible for `rMS.Name` to end up colliding with a _different_ generated mset that we'll see in another iteration, either before or after this one? I.e. renaming will result in two msets in our list with the same name?

I'm not gonna try to think through this yet, just noting it down for now.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -698,7 +752,33 @@ func (r *ReconcileMachinePool) syncMachineSets(
 	for _, ms := range machineSetsToCreate {
 		logger.WithField("machineset", ms.Name).Info("creating machineset")
 		if err := remoteClusterAPIClient.Create(context.Background(), ms); err != nil {
-			logger.WithError(err).Error("unable to create machine set")
+			// Check error message
+			message := "Unable to create MachineSet"
+			if apierrors.IsAlreadyExists(err) {
+				rMS := machineapi.MachineSet{}
+				remoteClusterAPIClient.Get(context.Background(),
```
2uasimojo, line 759: This needs an error check

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -698,7 +752,33 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+					types.NamespacedName{
+						Name:      ms.Name,
+						Namespace: ms.Namespace,
+					},
+					&machineapi.MachineSet{})
+
+				// FIXME condense these into one?
+				if rMS.Labels[constants.HiveManagedLabel] != "true" {
+					logger.WithError(err).Errorf("%s. A MachineSet with the same name but without a hive managed label already exists."+
+						" To have hive manage this MachineSet, add the label \"hive.openshift.io/managed\" with the value \"true\" to the MachineSet.", message)
```
2uasimojo, line 769: Oh, no, let's not put adoption instructions into messages. We don't even know if we're going to officially support adoption, or to what level. For the moment, we want to keep it in our back pocket as a workaround for individual cases.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -698,7 +752,33 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
+					logger.WithError(err).Errorf("%s. A MachineSet with the same name but without a matching MachinePool label already exists."+
+						" To have Machine Pool \"%s\" manage this MachineSet, add the label \"hive.openshift.io/machine-pool\" with the value \"%s\" to the MachineSet.", message, pool.Spec.Name, pool.Spec.Name)
+				}
+				if match, err := matchFailureDomains(ms, rMS, logger); err != nil && !match {
+					logger.WithError(err).Errorf("%s. A MachineSet with the same name but with a different failure domain already exists."+
+						" To have Machine Pool \"%s\" manage this MachineSet, ensure that the failure domain matches", message, pool.Spec.Name)
+				}
+			} else {
+				logger.WithError(err).Error(message)
+			}
```
2uasimojo, line 781: Okay, I get what you're trying to do with this section -- provide useful detail about how we ended up here. But I think it's probably too much effort and too fiddly to be worth it. Dumping the error ought to be sufficient: whoever's debugging ought to be able to go grab and scrutinize the objects themselves.

So I would say just leave the log on L780 and trash the rest.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -755,6 +835,7 @@ func (r *ReconcileMachinePool) syncMachineAutoscalers(
 			found := false
 			for _, rMA := range remoteMachineAutoscalers.Items {
 				if ms.Name == rMA.Name {
+
```
2uasimojo, line 838: whitespace damage

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -861,6 +942,7 @@ func (r *ReconcileMachinePool) syncClusterAutoscaler(
 	logger log.FieldLogger,
 ) error {
 	if pool.Spec.Autoscaling == nil {
+		logger.Infof("Cluster Autoscaler pool.Spec.Autoscaling is nil. mp: %#v", pool)
```
2uasimojo, line 945: Debug?

2uasimojo (commented):
Partial review. Couple points that'll end up changing behavior.

pkg/installmanager/installmanager.go
```diff
@@ -864,6 +865,58 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
...
+	mapPoolsByType := map[string]string{}
+	mpL := &hivev1.MachinePoolList{}
+	if err := m.DynamicClient.List(context.Background(), mpL, &client.ListOptions{Namespace: m.Namespace}); err != nil {
+		return err
+	}
+	for _, pool := range mpL.Items {
+		if pool.Spec.ClusterDeploymentRef.Name == cd.Name {
+			mapPoolsByType[pool.Spec.Name] = pool.Name
+		}
+	}
```
2uasimojo, line 879: Redundant `List` call. Let's collapse this chunk with the guts of `loadWorkerMachinePool()`. Make that func instead `loadMachinePools()` and produce `map[string]*MachinePool`; replace the `workerMachinePool` arg into this func with that map; then below this chunk you can do `workerMachinePool := mapPoolsByType["worker"]` to leave the rest working as it does today. `patchLabelMachineSetManifsts` will need a trivial refactor to extract the `pool.Name` from the map value.

pkg/installmanager/installmanager.go
```diff
@@ -864,6 +865,58 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
...
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
+			m.log.WithError(err).Error("error finding machine pool manifests")
+			return err
+		}
+	}
```
2uasimojo, line 918: Yeah, let's totally DRY this section with the next. With very little refactoring, we can correctly call both `patch*` funcs from within a single `WalkDir` looper.

pkg/installmanager/installmanager.go
```diff
@@ -950,6 +1003,70 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
...
 
+func patchLabelMachineSetManifest(manifestBytes []byte, poolsByType map[string]string, logger log.FieldLogger) (*[]byte, error) {
+	manifestBytesJson, err := yaml.YAMLToJSON(manifestBytes)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error converting yaml to json")
+	}
+
+	if err != nil {
+		return nil, errors.Wrap(err, "Error converting yaml to json")
+	}
```
2uasimojo, line 1014: dup

pkg/installmanager/installmanager.go
```diff
@@ -950,6 +1003,70 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
...
+	c, err := yamlutils.Decode(manifestBytes)
```
2uasimojo, line 1015: We're getting this `c` _just_ for the `yamlutils.Test` on L1024? We can't do that test with gjson and get rid of this conversion entirely?

pkg/installmanager/installmanager.go
```diff
@@ -950,6 +1003,70 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 	return nil
 }
 
+func patchLabelMachineSetManifest(manifestBytes []byte, poolsByType map[string]string, logger log.FieldLogger) (*[]byte, error) {
```
2uasimojo, line 1006: I want to see this DRYed with the other patch func eventually. But we can't do that until we've cut both over to using gjson/sjson. So for now please just add a TODO comment and reference the card you created for that cutover.

2uasimojo (commented):
Done with installmanager.go

I can hit installmanager_test.go next, but will wait for mpool controller tests until next rev.

pkg/installmanager/installmanager_test.go
```diff
@@ -865,6 +866,205 @@ spec:
...
+			scheme := scheme.GetScheme()
+			codecFactory := serializer.NewCodecFactory(scheme)
+			decoder := codecFactory.UniversalDecoder(machineapi.SchemeGroupVersion)
+
+			if tc.expectModified {
+				machineSetObj, _, err := decoder.Decode(*modifiedBytes, nil, nil)
+				assert.NoError(t, err, "expected to be able to decode MachineSet yaml")
+				machineSet, _ := machineSetObj.(*machineapi.MachineSet)
+
+				assert.Equal(t, pool.Spec.Name, machineSet.ObjectMeta.Labels[machinePoolNameLabel], "expected label %s to be %s", machinePoolNameLabel, pool.Spec.Name)
```
2uasimojo, line 1055: Also check the hive managed label?

pkg/installmanager/installmanager_test.go
```diff
@@ -865,6 +866,205 @@ spec:
...
+				if tc.testExistingLabels != nil {
+					for k, v := range tc.testExistingLabels {
+						assert.Equal(t, v, machineSet.ObjectMeta.Labels[k], "expected label %s to be %s", k, v)
+					}
+				}
+			}
+
+		})
+	}
+}
```
2uasimojo, line 1066: Good tests. I feel like if we were going past 80/20 we would also
- run this suite with multiple MPs and/or msets present, e.g. to make sure there isn't bleed-over
- cover the WalkDir bit

but this works for now.

test/e2e/postinstall/machinesets/infra_test.go
```diff
@@ -272,17 +272,20 @@ func TestAutoscalingMachinePool(t *testing.T) {
...
 		case apierrors.IsNotFound(err):
 			t.Log("waiting for Hive to create cluster autoscaler")
 			time.Sleep(10 * time.Second)
 		case err != nil:
 			t.Fatalf("could not get the cluster autoscaler: %v", err)
+		default:
+			t.Log("found cluster autoscaler")
+			break
 		}
 	}
```
2uasimojo, line 286: Since we're in here, could we add a check for the two mset labels somewhere?

2uasimojo:
/test e2e

Looks like we hit a regional limit in AWS :cry:

pkg/installmanager/installmanager.go
```diff
@@ -894,27 +907,37 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 				m.log.WithError(err).Error("error reading manifest file")
 				return err
 			}
-
-			modifiedManifestBytes, err := patchWorkerMachineSetManifest(manifestBytes, workerMachinePool, vpcID, m.log)
+			if addSecurityGroup {
+				modifiedBytes, err := patchWorkerMachineSetManifest(manifestBytes, workerMachinePool, vpcID, m.log)
```
2uasimojo, line 911: nts: `workerMachinePool` is guaranteed not to be `nil` here because `addSecurityGroup` would be false per L878.

pkg/installmanager/installmanager.go
```diff
@@ -950,14 +973,99 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, workerMach
 	return nil
 }
 
+// TODO: Combine this with patchWorkerMachineSetManifest once HIVE-2397 is completed
```
2uasimojo, line 976: ...and then you could inline this func.

Also, I think this comment was intended to go with the next func.

2uasimojo (commented):
/lgtm

Some small refactors noted inline, but let's address those in a followon.

PR #2181, opened 2024-01-09
Managed DNS is mutex with User-Configured DNS

2uasimojo:
Add a note in the docs that hive's managed DNS is mutually exclusive with installer's user-configured DNS.

[HIVE-1812](https://issues.redhat.com//browse/HIVE-1812)


2uasimojo:
/hold

We should at least wait until the [enhancement](https://github.com/openshift/enhancements/pull/1468) merges, if not until the feature is GA in installer.

PR #2182, opened 2024-01-10
Enhancement for metricsConfig redesign

suhanime:
xref: https://issues.redhat.com/browse/HIVE-2344

/assign @2uasimojo

### DIFF
--- a/docs/enhancements/metricsConfig_redesign.md
+++ b/docs/enhancements/metricsConfig_redesign.md
@@ -1,0 +1,180 @@
+<table>
+<tr>
+<th>Title</th>
+<th>Author</th>
+<th colspan="7">Reviewers</th>
+<th>Approver</th>
+</tr>
+<tr>
+<td>HiveConfig.spec.metricsConfig redesign</td>
+<td>@suhanime</td>
+<td>@tzvatot</td>
+<td>@bmeng</td>
+<td>@dustman9000</td>
+<td>@dofinn</td>
+<td>@janboll</td>
+<td>@berenss</td>
+<td>@hongkailiu</td>
+<td>@2uasimojo</td>
+</tr>
+</table>
+
+# HiveConfig.spec.metricsConfig redesign
+
+[HIVE-2344](https://issues.redhat.com/browse/HIVE-2344)
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Current implementations](#current-implementations)
+    - [Duration based threshold](#duration-based-threshold)
+    - [Additional label support](#additional-label-support)
+  - [Goals](#goals)
+  - [Non Goals](#non-goals)
+- [Proposal](#proposal)
+  - [metricsToReport](#metricstoreport)
+    - [metricNames](#metricnames)
+    - [minimumDuration](#minimumduration)
+    - [additionalClusterDeploymentLabels](#additionalclusterdeploymentlabels)
+    - [clusterDeploymentLabelSelector](#clusterdeploymentlabelselector)
+  - [Implementation Details / Notes](#implementation-details--notes)
+    - [Deprecation](#deprecation)
+    - [Failure Modes](#failure-modes) 
+  - [Risks and Mitigations](#risks-and-mitigations)
+
+## Summary
+As a Hive user interested in consuming the metrics it publishes, I would like to be able to configure the metrics as per my needs. 
+This includes the ability to choose which metrics are published, reduce the amount of observations for certain metrics unless it meets configured conditions, and the ability to add labels to the reported metrics for grouping and filtering purposes.
+
+## Motivation
+Hive administrators tend to use metrics for monitoring, alerting and troubleshooting.
+
+For monitoring and alerting, customers often want to apply their own filters/labels to the metrics, especially when 1 hive instance is used to manage different kinds of clusters, so they could tailor their queries and SLOs per their needs. For example, private-link clusters might take longer to install, or not wanting alerts to be thrown for clusters in limited support.
+These labels or filters are unique to their use case, so instead of codifying them for relevant metrics, the changes brought on by this enhancement would allow the admins to configure them via their hiveConfig.
+
+Troubleshooting involves identifying individual objects (clusters, syncsets, ...) via CRs and logs. In order to minimize the cognitive load of tracking down the offending object, there have been countless asks over the years to allow for applying object names as labels to the metrics.
+Prometheus metrics are designed for aggregation - identifying individual contributors to a metric is a known difficult use case because every label on a metric results in a whole new time series in the database. Labels with unbounded cardinality -- like cluster ID -- can result in unbounded growth and eventual dysfunction of the Prom DB. Thus, it is generally not recommended to label metrics with unique identifiers (like cluster IDs).
+
+This has been the major motivation for this enhancement. With the help of established baselines, admins can, via configuration, 
+- Enumerate exactly which metrics are to be reported.
+- Limit reporting metrics unless they exceed some defined boundary.
+- Label metrics arbitrarily, including with unique identifiers such as cluster IDs.
+- Employ LabelSelectors to group or limit clusters for which metrics are reported.
+
+### Current implementations
+Hive [publishes many metrics](../hive_metrics.md) that can be used for observations and alerting. Presently we offer 2 major customizations for some metrics:
+
+#### Duration based threshold
+Certain duration based metrics are used for alerting. While making them as specific as possible to quickly flag an issue with a cluster, it is also necessary to be mindful of the cardinality.
+It is possible to configure a threshold duration, so that metric wouldn't be logged until the value to be reported matches or exceeds the configured duration.
+It is important to call out these metrics are optional - hive doesn't report them by default.
+
+- **Pros for this design**:
+  - Threshold can be configured as per the needs of the alerts, so they're reported when needed, and there's less likelihood of these alerts needing to be silenced.
+  - The customization has a 1-1 relationship with the metric itself.
+  - These metrics happen to be optional, and not reported by default. 
+- **Cons we would like to overcome**: Given how we have designed it, raw/whole metric name is not used in the hiveConfig - instead, a corresponding camelCase key is hardcoded to refer to this metric. We would want to move away from such design, and implement a different - less confusing strategy that can be expandable.
+
+#### Additional label support
+Hive can manage a lot of clusters, so we usually avoid labelling any metric with cluster name or namespace to keep the cardinality in check. So, in addition to the labels hive reports for a metric, some admins might want extra labels.
+For example, if admins can identify that a metric corresponds to a "managed" cluster, they can apply effective filters for their observability and can observe trends across only managed clusters, and also customize their alerts.
+
+- **Pros for this design**: The process of applying the labels to clusterDeployment, configuring those labels in hiveConfig to be reported with the metric, and using them as filters for alerts and grafana queries - is all in the hands of the admins/users and does not require any code change.
+It also puts the onus of cardinality on the admins.
+- **Cons we would like to overcome**: 
+  - The way the code is written right now - we have a hardcoded hack for cluster-type label as mandatory if no additional labels are configured, and, most importantly, any additional labels configured will be applied to all the metrics which have optional label support, even though they might not be needed for all.
+  We're essentially worsening the cardinality this way, and it makes adding this support to more metrics tricky. 
+  - There's also the caveat of parsing through fixed and optional labels - ensuring there's no overlap and maintaining the order for certain metrics which do not accept map of label-value while observing, are some complications that currently exist and probably will not be fixed as a part of this enhancement.
+
+
+### Goals
+- Make all Hive metrics optional. By this design, no metric is to be reported by default unless it is explicitly configured in HiveConfig.
+- Allow configuring additional labels, duration based threshold and expression matching support for all metrics that are related to cluster deployment, via HiveConfig.
+- Enforce naming convention for duration based metrics - all metrics that report duration and allow customization for duration-based threshold must be named `*_seconds`.
+- Move away from throwing a panic if metricsConfig fails validation, update HiveConfig.status.Conditions instead.
+
+### Non-Goals
+Here are some special call-outs pertaining to what will be out-of-scope for this proposal, however this proposal is designed to support future expansion, so we may work on these in the future.
+- Non-clusterDeployment related metrics - like metrics reported by HiveOperator, or metrics that do not have the related clusterDeployment available while reporting (for ex, some syncSet, selectorSyncSet and clusterPool metrics) will not support additional label and expression matching. 
+- Adding labels only refers to the labels propagated from the clusterDeployment, any other strings as labels cannot be configured this way.
+
+
+## Proposal
+Allow hiveConfig.Spec.metricsConfig to look like
+```
+hiveConfig:
+  spec:
+    metricsConfig:
+      metricsToReport:
+      - metricNames:
+        - hive_foo_histogram
+        - hive_foo_gauge
+        minimumDuration: 10m
+        clusterDeploymentLabelSelector:
+          matchLabels:
+            hive.openshift.io/aro-snowflake: "true"
+          matchExpressions:
+          - key: hive.openshift.io/limited-support
+            operator: NotIn
+            values: 
+            - "true"
+          - key: hive.openshift.io/limited-support
+            operator: NotExists
+        additionalClusterDeploymentLabels:
+          prom_label_name: hive.openshift.io/cd-label-key
+     - metricNames:
+       - hive_foo_counter
+```
+### metricsToReport
+We would now default to not reporting any metrics unless they have been explicitly listed out under `metricsToReport`
+Within `metricsToReport`, there needs to be a non-empty list of hive metrics following an entry for `metricNames`, and there can be optional entries for customizations `minimumDuration`, `clusterDeploymentLabelSelector` and `additionalClusterDeploymentLabels`.
+`metricsToReport` is a list, each element of which requests reporting and configures filtering and customizations for the metrics provided in its `metricNames` list.
+
+#### metricNames
+`metricNames` would be a non-empty list of hive metrics that need to be reported, and can have optional customizations.
+All customizations listed for an entry of `metricToReport` will apply to all the metrics provided in that entry's `metricNames` list. This allows for grouping filters for a list of relevant metrics.
+A metric name must appear at most once across all `metricsToReport[].metricNames[]` in order to avoid ambiguity.
+Implementation will be adapting the existing setup for optional metrics. However, instead of a shorthand camelCase key that is used for current `hiveconfig.spec.metricsConfig.metricsWithDuration`, we would now need the full metric name listed under `metricNames`.
+In the example above, hive would only log `hive_foo_counter`, `hive_foo_gauge` and `hive_foo_histogram` metrics.
+
+#### minimumDuration
+Deprecate the current implementation of hiveConfig.spec.metricsConfig.metricsWithDuration, and change it to be reported as metricsConfig.metricsToReport.metricNames[].minimumDuration. The implementation of using the duration as a threshold before we report the metric stays the same.
+In the example above, `hive_foo_histogram` and `hive_foo_gauge` will only be logged if the value reported for them exceeds 10 minutes.
+
+#### additionalClusterDeploymentLabels
+Deprecate current implementation of hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels and change it to be reported as metricsConfig.metricsToReport.metricNames[].additionalClusterDeploymentLabels. Its implementation will not change.
+In the example above, `hive_foo_histogram` and `hive_foo_gauge` would report an additional label `prom_label_name`, its value corresponding to the value of `hive.openshift.io/cd-label-key` label on the corresponding clusterDeployment.
+
+#### clusterDeploymentLabelSelector
+This would be a new feature, of type [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), and we'd use the LabelSelector.MatchLabels and/or LabelSelector.MatchExpressions to match the conditions in order to decide if a metric should be reported.
+This encapsulates slightly more advanced filter logic over the existing clusterDeployment labels. In the example above, `hive_foo_histogram` and `hive_foo_gauge` metrics will only be reported for the clusterDeployments that are labelled with aro-snowflake and not in limited support.
+
+### Implementation Details / Notes
+
+- All the options configured for a `metricsToReport` entry will work in tandem with each other for each metric listed in its `metricNames`. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration.
+- This design allows for grouping of metrics for which you would want to define particular customizations. Grouping allows hiveConfig to be parsable. To avoid ambiguity, we will not allow multiple entries for any hive metric within `metricsToReport`.
+- We would be removing the fixed labels that can be propagated from the clusterDeployment, i.e.`cluster_type`, `platform` and `version`. This means, if admins want these labels for the metrics, they will have to define them via  metricsConfig.metricsToReport[].additionalClusterDeploymentLabels. This would only be a breaking change if the admins suddenly cut over from deprecated/old way of metricsConfig to the new metricsToReport, sufficient documentation should help though.
+- We would be renaming `hive_cluster_deployment_install_failure_total` and `hive_cluster_deployment_install_success_total` to ensure it ends with `seconds`, as it is a duration-based metric. Note that this will result in cloning the existing metric for the deprecation period, because if `metricsToReport` isn't present in the hiveConfig, then the _old_ way has to still work.
+- We would no longer panic when the metricsConfig fails validation, instead we would be updating hiveConfig.Status.Conditions[ReadyCondition] to false, with appropriate reason and message. See [Failure Modes](#failure-modes) for more details.
+
+#### Deprecation
+
+In order to implement this change, we will have to deprecate the existing hiveConfig.spec.metricsConfig.metricsWithDuration and hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+As a way to handle the deprecation, we will assert that in a hiveConfig, you can either provide the new configurations (aka `metricsToReport`) or the old way of hiveConfig.spec.metricsConfig.metricsWithDuration and hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels. If there's an empty metricsConfig, we would default to _old_ behaviour for the period of deprecation.
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only. The old design will log `hive_cluster_deployment_install_failure_total` and `hive_cluster_deployment_install_success_total`, in the new design, they will be renamed to `hive_cluster_deployment_install_failure_total_seconds` and `hive_cluster_deployment_install_success_total_seconds`. 
+
+#### Failure modes
+
+These situations will result in hive-operator failing to deploy the controllers. It will set hiveConfig.Status.Conditions[ReadyCondition] to `"False"` with an appropriate error message.
+-  `metricsToReport[].metricNames[$name]` doesn't exist as a metric.
+- `metricNames` lists a metric without duration but `minimumDuration` is specified in the same `metricsToReport` entry.
+- the same metric is mentioned more than once across all `metricsToReport[].metricNames[]`.
+- additionalClusterDeploymentLabels key conflict with an existing fixed label key for that metric.
+- if deprecated methods are used along with the new metricsToReport. You can either choose the _old_ way or the _new_ way.
+
+
+### Risks and Mitigations
+The biggest risk is the sheer number of metrics that are going to be affected. This would require thorough testing to ensure no metric changes its behaviour unexpectedly.
+
+It can get confusing for consumers to know the fixed labels each metric has. Update the [hive_metrics](https://github.com/openshift/hive/blob/master/docs/hive_metrics.md) doc to list out all labels and keep it up-to-date. We may, in the future, attempt to autogenerate these docs -- see [Hive-2413](https://issues.redhat.com/browse/HIVE-2413).
+
+Also, the more customizations there are, the longer hiveConfig will be. We have allowed for grouping of metrics for a set of customizations as a countermeasure.


docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+- There are other fixed labels (like platform and version), which are directed fed from the clusterdeployment labels. Ideally we would want to feed them via hiveconfig, but we might need an exception for the cluster deployment labels that are applied by hive. 
+- All the options configured for a metric within clusterDeploymentRelatedMetrics, will work in tandem with each other. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration
+- Failure modes:
+  These situations will cause a panic().
+  - clusterDeploymentRelatedMetrics[].name doesn't exist as a metric
+  - clusterDeploymentRelatedMetrics[].name is a metric without duration but minimumDuration was specified
+  - additionalClusterDeploymentLabels conflict with existing fixed label for that metric
+
+### Risks and Mitigations
```
suhanime, line 112: Mike (and Toni) reviewed this recently with me, and they raised some concerns about the deprecation of existing implementations and how it would work. It should be easy to detect if `metricsConfig.clusterDeploymentRelatedMetrics` is present or not - and that can dictate the "old" path vs "new" path. Would that be enough? Or should we decide and document more info about it?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### additionalClusterDeploymentLabels
+Deprecate current implementation of hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels and change it to be reported as metricsConfig.clusterDeploymentRelatedMetrics[$name].additionalClusterDeploymentLabels. It's implementation will not change.
+
+#### clusterDeploymentLabelSelector
+This would be a new feature, of type [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), and we'd use the LabelSelector.MatchLabels and/or LabelSelector.MatchExpressions to match the conditions in order to decide if a metric should be reported.
+This encapsulates slightly more advanced filter logic over the existing clusterdeployment labels. In the example above, hive_foo_counter metric will only be reported for the cluster deployments that are labelled with aro-snowflake, not in limited support, and has the metric value > 10m.
+
+### Implementation Details / Notes
+- In order to implement this change, we will have to deprecate the existing hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.`
```
2uasimojo, line 101: Would you mind elaborating a little bit on the plan for deprecation? Are we planning to support both mechanisms for $time? Will they be mutually exclusive for that period, resulting in a hard error (operator crashes) if both are configured? Etc.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+- Current configuration of using a shorthand camelCase key to refer to a corresponding metric while configuring hiveconfig.spec.metricsConfig.metricsWithDuration will be removed, and users will have to use the full metric name in the proposed configuration.
```
2uasimojo, line 102: The name in question is in a different location in the config in the new design vs old, right? The new design will use the full metric name instead; but this implies that the old thing will get switched over right away, which I don't think is what you intended.

Assuming there's a dual-support deprecation period (as commented above), we would continue to support the camelCase names _only_ in the old, deprecated sections of the config, until those sections are fully removed.

Right?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+- Additionally, we would also want to remove `cluster_type` as a fixed label for all the metrics that do report it now. This would be a breaking change as the consumers who want `cluster_type` will have to configure it via metricsConfig.clusterDeploymentRelatedMetrics[].additionalClusterDeploymentLabels for all the metrics they need it for.
```
2uasimojo, line 103: Just to clarify: We will be able to detect whether the old or new config mechanism is in play (esp if we make them mutually exclusive as suggested above), so it wouldn't be crazy to leave the `cluster_type` fixed label for the old way, but remove it for the new, right?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+- There are other fixed labels (like platform and version), which are directed fed from the clusterdeployment labels. Ideally we would want to feed them via hiveconfig, but we might need an exception for the cluster deployment labels that are applied by hive.
```
2uasimojo, line 104: Again here let's make a commitment. Are we going to put platform and version in by default as fixed labels, or leave them completely to the user? I'm inclined to say the latter, but open to discussion.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+- Additionally, we would also want to remove `cluster_type` as a fixed label for all the metrics that do report it now. This would be a breaking change as the consumers who want `cluster_type` will have to configure it via metricsConfig.clusterDeploymentRelatedMetrics[].additionalClusterDeploymentLabels for all the metrics they need it for.
+- There are other fixed labels (like platform and version), which are directed fed from the clusterdeployment labels. Ideally we would want to feed them via hiveconfig, but we might need an exception for the cluster deployment labels that are applied by hive. 
+- All the options configured for a metric within clusterDeploymentRelatedMetrics, will work in tandem with each other. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration
+- Failure modes:
+  These situations will cause a panic().
+  - clusterDeploymentRelatedMetrics[].name doesn't exist as a metric
```
2uasimojo, line 108: I'm a little leery of this. It makes it hard e.g. for a consumer to share a hiveconfig across heterogeneous deployments with different versions of hive (think ACM).

I do feel like we discussed this, but I can't remember what we landed on, or why. Did we talk about having status conditions instead?

If I'm retreading ground here, it means the reasoning isn't obvious, which means we should have the alternatives discussed in the doc. (Inline or a whole separate "Alternatives" section -- whatever you prefer.)

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+  - clusterDeploymentRelatedMetrics[].name is a metric without duration but minimumDuration was specified
+  - additionalClusterDeploymentLabels conflict with existing fixed label for that metric
```
2uasimojo, line 110: Unless we decide to have zero fixed labels.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Oh, look, great minds and all that. TL;DR: Yes. See above.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Risks and Mitigations
+The biggest risk is the sheer number of metrics that are going to be affected.
+
+Many metrics related to cluster deployment which do not currently have support for additionalClusterDeploymentLabels will now have that support. This might create an avenue for worsening the cardinality, however, that would be the responsibility of Hive admin.
+
+It can get confusing for consumers to know the fixed labels each metric has.Update the [hive_metrics](https://github.com/openshift/hive/blob/master/docs/hive_metrics.md) doc to list out all labels and keep it up-to-date.
```
2uasimojo, line 117: ...or have zero fixed labels. The obvious disadvantage being that the user has to configure a given label on each and every one if they want it across the board.

I anticipate SD reflexively asserting that they'll want labels like platform and version for every metric. But I think if we challenge that assertion, they're pretty likely to walk it back.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+
+Also, the more customizations there are, the longer hiveConfig will be. Currently, we view this as an acceptable change.
```
2uasimojo, line 119: Mm, this is a good point. There's a limit on the size of a k8s object. We should probably do a little bit of back-of-napkin math to see roughly what order of magnitude this config could be for scenarios where e.g. every metric gets configured for platform and version labels.

Today on hivep02ue1 (all prod shards should be identical, or near enough as makes no difference):
```
$ oc get hiveconfig hive -o yaml | wc -c
77278
```

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+  - [Non Goals](#non-goals)
+- [Proposal](#proposal)
+  - [clusterDeploymentRelatedMetrics](#clusterdeploymentrelatedmetrics)
+    - [minimumDuration](#minimumduration)
+    - [additionalClusterDeploymentLabels](#additionalclusterdeploymentlabels)
+    - [clusterDeploymentLabelSelector](#clusterdeploymentlabelselector)
+  - [Implementation Details / Notes](#implementation-details--notes)
+  - [Risks and Mitigations](#risks-and-mitigations)
+
+## Summary
```
2uasimojo, line 20: Let's get a "Reviewers" section. I want to get acks from _at least_ SRE-P (the likes of Naveen/Dustin/Christoph) and OCM (Elad? Priya?) before we proceed with code changes. AppSRE (Patrick/Maor) ACM (maybe start with Gurney, who can delegate) and DPTP (Bear) should also have the opportunity to provide feedback.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,119 @@
...
+- There are other fixed labels (like platform and version), which are directed fed from the clusterdeployment labels. Ideally we would want to feed them via hiveconfig, but we might need an exception for the cluster deployment labels that are applied by hive. 
+- All the options configured for a metric within clusterDeploymentRelatedMetrics, will work in tandem with each other. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration
+- Failure modes:
+  These situations will cause a panic().
+  - clusterDeploymentRelatedMetrics[].name doesn't exist as a metric
+  - clusterDeploymentRelatedMetrics[].name is a metric without duration but minimumDuration was specified
+  - additionalClusterDeploymentLabels conflict with existing fixed label for that metric
+
+### Risks and Mitigations
+The biggest risk is the sheer number of metrics that are going to be affected.
```
2uasimojo, line 113: Yeah, it should be possible to get an exact count, right? Let's drop that in the doc. (And feed it into the napkin math suggested below.)

docs/enhancements/metricsConfig_redesign.md
2uasimojo, line 110: Add panic for mutually exclusive old & new configs, if we go that route.

Also, did we decide all new failure modes are panics? Nothing that'll end up in a status condition instead? Or just be logged?

2uasimojo (commented):
Good stuff @suhanime, thank you.

docs/enhancements/metricsConfig_redesign.md
suhanime: Uh, I don't understand this comment well enough. I personally think it is better to only refer to the full-name of metric in the hiveConfig. It has to be the hiveConfig because we register metrics way before we parse the conditions of hiveOperator - and registration requires us to know all the labels a metric will report. We can only register a metric once during a pod's lifetime. That's why hiveconfig seemed as a better option, and we were okay with panics (because these panics happen within seconds - at most 2 minutes of an hiveConfig edit).
We can consider manifests or volumeMounts or something - the pod has to restart anyways. Do you have any alternatives?

docs/enhancements/metricsConfig_redesign.md
suhanime: Not possible for every metric - see below.

suhanime: All panics are related to registration of metrics, so it is better if they are panics since they happen early enough.

docs/enhancements/metricsConfig_redesign.md
suhanime: Also - I hope uses see the hassle of mentioning multiple metrics in hiveConfig as enough of a deterrent to not apply additionalLabels to all metrics.

suhanime (commented):
@2uasimojo ready for another pass

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+- Current configuration of using a shorthand camelCase key to refer to a corresponding metric while configuring hiveconfig.spec.metricsConfig.metricsWithDuration will be removed, and users will have to use the full metric name in the proposed configuration.
+- Presently, the metrics that support hiveconfig.spec.metricsConfig.metricsWithDuration are optional, so they will not be logged unless they're specified in HiveConfig. After this redesign, the metrics that are currently optional will continue to be optional - that behaviour will not change.
+  For example, after the redesign, hive_cluster_deployments_running_transition_seconds will be reported if and only if metricsConfig.clusterDeploymentRelatedMetrics.metrics[hive_cluster_deployments_running_transition_seconds].minimumDuration is not empty, however hive_cluster_deployment_install_job_duration_seconds, which is a non-optional metric that would support minimumDuration, will be reported even if metricsConfig.clusterDeploymentRelatedMetrics.metrics[hive_cluster_deployment_install_job_duration_seconds].minimumDuration is not specified.
```
suhanime, line 131: I forgot to specify that optional metrics will continue to remain optional. Hope this explanation isn't confusing, and metrics doc does specify if a metric is optional or not.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Sorry, my comment was around whether we should actually panic when we encounter an unrecognized metric name in hiveconfig.

I don't claim to remember our discussion from last year perfectly... but maybe we decided panic was the lesser evil. For instance, it would catch a typo which might otherwise go unnoticed until too late. Worth the tradeoff of having to make sure you're only referencing metrics available on the hive version running on _this_ hub.

So, I'm just asking that that rationale be explained a little bit. A sentence or two is plenty.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Ahh, okay, the clarifying point is that the optional labels are always coming from ClusterDeployment.Labels.

What about if we said we will never have a fixed label that corresponds to a CD label? Does that work?

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Thanks for doing the numbers. Given what you've computed, I'm really not worried about exceeding the k8s object size limit, even if we extend this to every metric we produce, and even if we add more later.

I'm also not worried about hiveconfig being awkward for humans to read. That ship has sailed, what with all the privatelink jazz in there. I'm fine with the user doing jq/yq wizardry if they really need something they can't just do a quick text search for.

That said, I do like your idea of being able to group filters by supplying a list of metric names. Let me take a closer look at the proposal and get back to you.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+## Proposal
+Allow HiveConfig.Spec.metricsConfig to look like
+```
+HiveConfig:
+  spec:
+    metricsConfig:
+      clusterDeploymentRelatedMetrics:
+      - metrics:
```
2uasimojo, line 92: nit/bikeshedding: maybe something more explicit like `metricNames`?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+        - hive_foo_counter
+        - hive_foo_gauge
+        minimumDuration: 10m
```
2uasimojo, line 95: We may mention this elsewhere, but
- Non-duration-based metrics are allowed to be configured with the new hotness, right? And/but...
- If you specify this `minimumDuration` knob for a non-duration-based metric, we'll panic, right?

IOW, non-duration-based metrics can still be label-filtered and get additional labels; they just can't be restricted by duration.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Implementation Details / Notes
+
+- All the options configured for a metric within clusterDeploymentRelatedMetrics, will work in tandem with each other. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration
+- This design allows for grouping of metrics for which you would want to define particular customizations. Grouping allows hiveconfig to be parsable, but implementation of this model would require an efficient solution to validate there are no ambiguous entries for a metric, and storing it within the code for parsing purposes.
+- There needs to be clear logic defining which metrics can have `minimumDuration` versus which ones can't. Perhaps if a metric has `seconds` in the name can be the decider.
```
2uasimojo, line 128: Are you talking about clear logic for the user or for the code?

In code we can have a separate interface that only duration-based metrics would implement; and we can exploit that interface in some way to enforce that the `minimumDuration` knob is only specified for those metrics.

For the UX, it would indeed be nice if we could make sure there's some naming convention... but only if we find that's already true. I don't think we want to rename existing metrics.

Worst case, the UX is simply a) docs; and b) that hive-operator panics if they mess it up :shrug:. I'm okay with that.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Well now hold on.

You're saying that metrics that are *currently* not optional simply because we haven't (yet) plumbed them with minimumDuration ==> these will now _support_ minimumDuration, but will differ from metrics that are *currently* optional in that they _will_ be reported if they are not mentioned in the config??

Naw.

This redesign is an opportunity to make the overall picture -- as in, what a brand new user would see -- consistent and sensible. Reporting some metrics by default and not others because we made a seemingly-arbitrary decision -- or worse, for historical reasons growing out of our interactions with one customer (OSD) -- is not a good look. We should decide on one default behavior across the board. And personally, I'm in favor of reporting *nothing* by default. YAGNI principle.

Now the ugly bit is going to be how to handle the deprecation period. Unfortunately, we're going to have to bake in logic like
```
if hiveconfig has the old metrics config section
  then {these metrics} are reported by default but {these metrics} are not
else if hiveconfig has the new metrics config section
  then report all _and only_ metrics that are mentioned in the config, iff their filters say aye.
else if hiveconfig lacks either section
  then, ummm, I guess use the old behavior during the deprecation period? I can be convinced either way here.
else (hiveconfig has both sections)
  panic
```

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+  For example, after the redesign, hive_cluster_deployments_running_transition_seconds will be reported if and only if metricsConfig.clusterDeploymentRelatedMetrics.metrics[hive_cluster_deployments_running_transition_seconds].minimumDuration is not empty, however hive_cluster_deployment_install_job_duration_seconds, which is a non-optional metric that would support minimumDuration, will be reported even if metricsConfig.clusterDeploymentRelatedMetrics.metrics[hive_cluster_deployment_install_job_duration_seconds].minimumDuration is not specified.
+- Additionally, we would be removing the fixed labels that can be propagated from the cluster deployment, i.e.`cluster_type`, `platform` and `version`. This means, if admins want these labels for the metrics, they will have to define them via  metricsConfig.clusterDeploymentRelatedMetrics[].additionalClusterDeploymentLabels. This would only be a breaking change if the admins suddenly cut over from deprecated/old way of metricsConfig to the new clusterDeploymentRelatedMetrics, sufficient documentation should help though.
```
2uasimojo, line 132: Yeah, nice ✓

docs/enhancements/metricsConfig_redesign.md
2uasimojo: [Later] I see now where you've said exactly this, cool ✓

docs/enhancements/metricsConfig_redesign.md
2uasimojo: [Later] Having seen it, I still like it, and don't see any obvious gotchas. Let's see what other reviewers think, but I'm inclined to go with it.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+In order to implement this change, we will have to deprecate the existing hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+As a way to handle the deprecation, we will assert that in a hiveconfig, you can either provide the new configurations (aka `clusterDeploymentRelatedMetrics`) or the old way of hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only.
+
+#### Failure modes
+
+These situations will cause a panic().
+- clusterDeploymentRelatedMetrics.metrics[].name doesn't exist as a metric
+- clusterDeploymentRelatedMetrics.metrics[].name is a metric without duration but minimumDuration was specified.
+- a metric is defined across multiple groupings and the customizations defined contradict. For ex, following hiveConfig will cause panic because hive will be confused whether to parse `minimumDuration` for `hive_foo_seconds` to be 10m or 1h.
```
2uasimojo, line 145: I would be stricter here and not even attempt to detect (non-)contradictions. If a metric is mentioned in two places, panic unequivocally.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+        clusterDeploymentRelatedMetrics:
+        - metrics:
+          - hive_foo_seconds
+          minimumDuration: 10m
+        - metrics:
+          - hive_foo_seconds
+          - hive_foo_gauge
+          minimumDuration: 1h
+  ```
+- additionalClusterDeploymentLabels conflict with existing fixed label for that metric
```
2uasimojo, line 159: Since we're trashing CD-label-related fixed metrics, this failure mode is actually: if they specify a prom label _key_ that conflicts with an existing fixed label _key_, right? Doesn't really matter where (either) _value_ comes from.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+- Additionally, we would be removing the fixed labels that can be propagated from the cluster deployment, i.e.`cluster_type`, `platform` and `version`. This means, if admins want these labels for the metrics, they will have to define them via  metricsConfig.clusterDeploymentRelatedMetrics[].additionalClusterDeploymentLabels. This would only be a breaking change if the admins suddenly cut over from deprecated/old way of metricsConfig to the new clusterDeploymentRelatedMetrics, sufficient documentation should help though.
+
+#### Deprecation
+
+In order to implement this change, we will have to deprecate the existing hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+As a way to handle the deprecation, we will assert that in a hiveconfig, you can either provide the new configurations (aka `clusterDeploymentRelatedMetrics`) or the old way of hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only.
+
```
2uasimojo, line 139: As mentioned above, since the two modes differ in several default behaviors, we need to say what happens if *neither* config is given in hiveconfig. And per above, my starting bid is "old behavior". I could maybe be talked out of it if there's a compelling argument on the other side. Thoughts?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+- additionalClusterDeploymentLabels conflict with existing fixed label for that metric
+- if deprecated methods are used along with the new clusterDeploymentRelatedMetrics. You can either choose the _old_ way or the _new_ way.
+
+
+### Risks and Mitigations
+The biggest risk is the sheer number of metrics that are going to be affected. Roughly 29 metrics that we report right now can be classified as related to clusterDeployment.
+
+Moreover, 8 of those metrics do not have `minimumDuration` support right now will be able to have that customization, and we would need ideally a non-hardcoded way  of classifying them as supported or not.
+
+About 19 metrics related to cluster deployment which do not currently have support for `additionalClusterDeploymentLabels` will now have that support. This might create an avenue for worsening the cardinality, however, that would be the responsibility of Hive admin.
```
2uasimojo, line 168: I think this is the crux: If, as suggested above, we do _not_ report a metric under the new mode unless it is explicitly mentioned in the config, then a) we are putting _all_ of the responsibility on the user, as you say; but also b) since we report _nothing_ by default, the number of time series will be minimized by default, which is goodness.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+          - hive_foo_seconds
+          - hive_foo_gauge
+          minimumDuration: 1h
+  ```
+- additionalClusterDeploymentLabels conflict with existing fixed label for that metric
+- if deprecated methods are used along with the new clusterDeploymentRelatedMetrics. You can either choose the _old_ way or the _new_ way.
+
+
+### Risks and Mitigations
+The biggest risk is the sheer number of metrics that are going to be affected. Roughly 29 metrics that we report right now can be classified as related to clusterDeployment.
```
2uasimojo, line 164: Not to be pedantic, but why is this a risk? Because of the amount of code we're going to touch? The disruption to the user when they cut over? It shouldn't pose a risk from a design/architecture or execution standpoint.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,172 @@
...
+
+Moreover, 8 of those metrics do not have `minimumDuration` support right now will be able to have that customization, and we would need ideally a non-hardcoded way  of classifying them as supported or not.
+
+About 19 metrics related to cluster deployment which do not currently have support for `additionalClusterDeploymentLabels` will now have that support. This might create an avenue for worsening the cardinality, however, that would be the responsibility of Hive admin.
+
+It can get confusing for consumers to know the fixed labels each metric has. Update the [hive_metrics](https://github.com/openshift/hive/blob/master/docs/hive_metrics.md) doc to list out all labels and keep it up-to-date.
```
2uasimojo, line 170: I know I mentioned this before as a pie-in-the-sky possibility: can we say we'll at least _try_ to investigate having the docs generated from code so they're guaranteed to stay up to date? Even if it ends up being a card on the backlog forever...

2uasimojo (commented):
This is looking great. I think the only crucial issues at this stage are
- Deciding whether to make _all_ affected metrics off by default
- Deciding on default behavior (if neither metrics config is provided) during the deprecation period.

bmeng:
With a read through the design doc, I do not think there would be any issue from SREP side.

docs/enhancements/metricsConfig_redesign.md
tzvatot, line 1: Reading through this, I have several questions:
1. Is there/should there be an ADR/DDR for this? I was working on https://docs.google.com/document/d/1lYK_S3LCt-gCwV9d4Y5Oe2FeBfQ_6C_EyudCRtOrw3k/edit#heading=h.bupciudrwmna which kinds of overlap
2. What will be the frequency of updating the metrics? CS is resonciling on the entire fleet, so we should be minded of the performance penalty of getting a load of metrics update simultaniosly. 
3. While this may solve the hive domain, we still have the same challange with ACM/Hypershift. It would be better to have a single way of reporting metrics (a new CRD?) that we can potentially re-use in the HCP domain. Having diverged solutions (hive, HCP, and whatever comes next) slows down development process.
4. Added a comment on the jira ticket referring on the examples provided there: https://issues.redhat.com/browse/HIVE-2344?focusedId=24065859&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24065859

2uasimojo: Hi @tzvatot. Thank you for reviewing! TL;DR we suspect this enhancement to be of little interest to CS, but wanted to get your eyes on it just in case :)
1. I don't believe there's overlap with that ADR, which IIUC is looking to get *away* from using metrics to gather per-cluster data. That's appropriate IMO: metrics should aggregate trends, not report on individual clusters. That way lies cardinality hell.
2. We update the metrics continuously. Some are updated realtime as we reconcile objects that affect them. Others are updated periodically (/2m) as we poll those objects. In all cases there is a certain amount of delay built in based on the scrape interval of the prometheus service. All standard prometheus patterns I believe.
3. I can't see a world where it makes sense to report _metrics_ in a CRD, nor where it makes sense for two different services (hive and hypershift) to attempt to converge on design of specific metrics. I suspect we're conflating use cases here. It makes sense to converge on reporting _metadata_ as described by the ADR you linked -- but that's a different thing entirely.
4. Ack.

docs/enhancements/metricsConfig_redesign.md
suhanime: Technically true, added.

docs/enhancements/metricsConfig_redesign.md
suhanime: I think it is a risk to touch too many metrics, you have to ensure no metric changes its behaviour unexpectedly - this would require thorough testing from both the development and QE standpoint. I've added some clarification, though I don't mind dropping this from the section if you insist.

docs/enhancements/metricsConfig_redesign.md
suhanime: I don't think we should be changing whether a metric is always reported or optional as a part of this enhancement, but I'll keep this comment unresolved for now, so I can address it if that changes.

suhanime (commented):
@2uasimojo ready for another pass.
I think we still differ on whether all metrics should be made optional or not, and I personally would like to handle that separately. Let's discuss some more on that.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Leaving aside whether we change which metrics are reported (by default), the default labels _are_ different between the old and new, so IMO it's important to call out explicitly which behavior you get if you leave the config empty.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Those would be in wholly separate deployments (different HiveConfig instances on different clusters). Same argument as "<name> doesn't exist as a metric": a metric might exist in one version of hive and not another, so I can't count on using the same HiveConfig on two different deployments.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Deprecation
+
+In order to implement this change, we will have to deprecate the existing hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+As a way to handle the deprecation, we will assert that in a hiveconfig, you can either provide the new configurations (aka `clusterDeploymentRelatedMetrics`) or the old way of hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only.
+
+#### Failure modes
+
+These situations will cause a panic().
```
2uasimojo, line 131: I thought through this a bit and I'm not sure `panic` is the right reaction. Rather, I think we want a status condition on HiveConfig that indicates that the config is invalid, and refuse to deploy the controllers, and hive-operator reconcile returns `nil` so it doesn't hot-loop, but waits to be poked by a hiveconfig edit. That way, the remedy is `oc edit hiveconfig hive` rather than
```
oc logs -n hive deploy/hive-operator
<find the problem>
oc edit hiveconfig hive
oc delete po -n hive hive-operator-xyz-abc  # to restart the service
```

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Nice, thanks.

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Copyout from DM brain dump:

IIUC currently a *subset* of duration metrics are optional, and the mechanism for opting into or out of them is to specify a minimum duration below which they will not be reported.

Trying to understand this concept from the [doc](https://github.com/openshift/hive/blob/master/docs/hive_metrics.md) is now triggering my pedantic button. "Optional label support" is not the same thing as a metric being optional. A metric being optional implies that it is not reported by default, but must be opted into. Labels are an attribute of a metric, and whether a particular label is included on a reported metric is orthogonal to whether the metric itself is reported or not.

The thing we're discussing (I think) is the default behavior of a duration-based metric that is not mentioned in hiveconfig: do we report it (for all values of the duration) or not? Put another way, is the default minimumDuration zero or infinity? (Which might be a reasonable way to structure the code btw.)

And my position is: I want to be looking at the metrics picture after this redesign as if it were brand new, in isolation. And in such a view, I don't feel it makes sense to have an arbitrary subset of duration metrics that are reported by default, and the rest are not reported by default.

I say "arbitrary" because, while I respect the notion of identifying some metrics that are "critical" or "important" or whatever, IMO that's a user-specific thing. Any decision we make may _mostly_ make sense to _most_ consumers, but there will be a handful where they'll be like, "oh, **this** one is off by default, why??"

Now, I get where we have an issue if we say all duration metrics are off by default, but the rest are on by default. Same kind of seemingly-arbitrary distinction (though this one is at least a little crisper)

I'm thus inclined to say all metrics should be opt-in. This could be a separate section of metricsConfig with a simple list of metric names. (Do we require you to duplicate the cd-related ones in that list and in clusterDeploymentRelatedMetrics? Open to discussion.)

> This has been the trend that I have observed across different services also.

I'm not aware of any other service that has opt-in metrics. Are you?

suhanime: Summary of our discussion from DM:
- Make all metrics opt-in as a part of this enhancement (Does this apply only to clusterDeploymentRelatedMetrics or all metrics?)
- Rename `metricNames` to `metricsToReport`. Within clusterDeploymentRelatedMetrics, only `metricsToReport` will be mandatory, all other customizations are optional.
- Basically, if you're using the new config (aka clusterDeploymentRelatedMetrics), you need to specify all the hive metrics you want reported, hive will not report any metric by default. The list of all metrics to be reported sounds easy enough to collate - say from the dashboard (grafana, amplitude etc) config/deployment files, and from the alerting rules.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Deprecation
+
+In order to implement this change, we will have to deprecate the existing hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+As a way to handle the deprecation, we will assert that in a hiveconfig, you can either provide the new configurations (aka `metricsToReport`) or the old way of hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels. If there's an empty metricsConfig, we would default to _old_ behaviour for the period of deprecation.
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only.
+
+#### Failure modes
+
+These situations will result in metrics not getting registered and hiveConfig.Status.Conditions[ReadyCondition] will be set to false:.
```
suhanime, line 144: Okay so I looked at your suggestion for registering the metric in hiveOperator reconcile, but it wouldn't work correctly because we would need to share the registry with which we're registering the metric. I'm still unsure if we can make it work, and then there's separate registration for HiveOperator and Hive controller metrics - so I didn't specify the implementation details here, just the result.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,169 @@
...
+          - hive_foo_seconds
+          - hive_foo_gauge
+          minimumDuration: 1h
+  ```
+- additionalClusterDeploymentLabels key conflict with an existing fixed label key for that metric
+- if deprecated methods are used along with the new metricsToReport. You can either choose the _old_ way or the _new_ way.
+
+
+### Risks and Mitigations
+The biggest risk is the sheer number of metrics that are going to be affected. This would require thorough testing to ensure no metric changes its behaviour unexpectedly.
```
suhanime, line 165: I removed the calculation for the total metrics affected. All duration-based metrics should support minimumDuration, regardless of whether they're related to cluster deployment or not, but the other 2 customizations can only be supported if we have a cluster deployment available. I decided to let those implementation details be clarified in the hive docs for metrics once we're done with these changes.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,169 @@
...
+As a way to handle the deprecation, we will assert that in a hiveconfig, you can either provide the new configurations (aka `metricsToReport`) or the old way of hiveconfig.spec.metricsConfig.metricsWithDuration and hiveconfig.spec.metricsConfig.additionalClusterDeploymentLabels. If there's an empty metricsConfig, we would default to _old_ behaviour for the period of deprecation.
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only.
+
+#### Failure modes
+
+These situations will result in metrics not getting registered and hiveConfig.Status.Conditions[ReadyCondition] will be set to false:.
+- metricsToReport.metricNames[].<name> doesn't exist as a metric
+- metricsToReport.metricNames[].<name> is a metric without duration but minimumDuration was specified.
+- a metric is defined across multiple groupings and the customizations defined contradict. For ex, following hiveConfig will cause panic because hive will be confused whether to parse `minimumDuration` for `hive_foo_seconds` to be 10m or 1h. Please note, we might end up allowing only 1 entry of hive metric in the hiveConfig.
```
suhanime, line 147: I hear your suggestion for allowing only 1 entry for a metric - I want to keep this avenue open and see how it pans out during implementation.

docs/enhancements/metricsConfig_redesign.md
suhanime: If it matters, now 17 metrics will support minimumDuration, 6 of them already have it so 11 more than we already do. Rest of the calcs stay the same - 19 more will have additional label support and 29 will support label selection. 
52 total metrics that we can report as of now, so 46 metrics of them will now become optional.
We originally wanted the math because we were worried about exceeding size of hiveconfig, since that's no longer an issue - no point calling it out.

suhanime (commented):
@2uasimojo Hopefully this is it?

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Well, hold on, what harm would there actually be in registering them if we never report them? Doesn't the /metrics endpoint only show time series with values, or am I remembering that wrong?

2uasimojo: [Later] We could also [create a separate registry](https://github.com/openshift/hive/blob/3de1784bb5a21e967581c22a44420013afa30afa/vendor/github.com/prometheus/client_golang/prometheus/registry.go#L67) just for this purpose, and throw it away.

2uasimojo: Re "metrics not getting registered", I think we'll actually end up just... not deploying the controllers at all, right? Since hive-operator will be erroring out of its reconcile loop? So indeed metrics wouldn't get registered... because the thing that would be registering them isn't running :)

docs/enhancements/metricsConfig_redesign.md
2uasimojo: As written you're suggesting that we would allow duplicates if they're effectively identical. What possible reason could there be for allowing that? It can only be confusing to the user looking at HiveConfig; and it makes our job harder in the code (how do you determine "effectively identical"?) So I say we take a hard stance here. Duplicates => error.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Goals
+- Make all Hive metrics optional. By this design, no metric is to be reported by default unless it is explicitly configured in HiveConfig.
+- Allow configuring additional labels, duration based threshold and expression matching support for all metrics that are related to cluster deployment, via HiveConfig.
+- Enforce naming convention for duration based metrics - all metrics that report duration and allow customization for duration-based threshold must be named `*_seconds`.
+- Move away from throwing a panic if metricsConfig fails validation, update HiveConfig.status.Conditions instead
+
+### Non-Goals
+- Non-cluster deployment related metrics - like metrics reported by HiveOperator, or metrics that do not have the related cluster deployment available while reporting (for ex, some syncset, selectorsyncset and clusterpool metrics) will not support additional label and expression matching. 
+- Adding labels only refers to the labels propagated from the Cluster Deployment, any other strings as labels cannot be configured this way.
```
2uasimojo, line 80: Not sure exactly how/where to put this, but I'd like to mention that, while the non-goals are _currently_ non-goals, it is a goal that, if we *did* decide to do those things in the future, it would be easy to expand this design to accommodate them. In fact, it may be worth calling out in the `Current implementations` section that a con of the old design that we would like to overcome is its non-expandability.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,169 @@
...
+        prom_label_name: hive.openshift.io/cd-label-key
+   - metricsToReport:
+     - hive_foo_histogram
+```
+### metricsToReport
+We would now default to not reporting any metrics unless they have been explicitly listed out under `metricsToReport`.
+There needs to be a non-empty list of hive metrics following an entry for `metricsToReport`, and there can be optional entries for customizations `minimumDuration`, `clusterDeploymentLabelSelector` and `additionalClusterDeploymentLabels`.
+All customizations listed for an entry of `metricsToReport` will apply to all the metrics provided in the corresponding list. This allows for grouping filters for a list of relevant metrics.
+Multiple entries for `metricsToReport` are allowed.
+Implementation will be adapting the existing setup for optional metrics. However, instead of a shorthand CamelCase key that is used for current `hiveconfig.spec.metricsConfig.metricsWithDuration`, we would now need the full metric name listed under `metricsToReport`.
```
2uasimojo, line 113:
```suggestion
Implementation will be adapting the existing setup for optional metrics. However, instead of a shorthand camelCase key that is used for current `hiveconfig.spec.metricsConfig.metricsWithDuration`, we would now need the full metric name listed under `metricsToReport`.
```

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,169 @@
...
+- This design allows for grouping of metrics for which you would want to define particular customizations. Grouping allows hiveconfig to be parsable, but implementation of this model would require an efficient solution to validate there are no ambiguous entries for a metric, and storing it within the code for parsing purposes. We might end up enforcing only 1 entry for a metric in the HiveConfig.
+- We would be removing the fixed labels that can be propagated from the cluster deployment, i.e.`cluster_type`, `platform` and `version`. This means, if admins want these labels for the metrics, they will have to define them via  metricsConfig.metricsToReport[].additionalClusterDeploymentLabels. This would only be a breaking change if the admins suddenly cut over from deprecated/old way of metricsConfig to the new metricsToReport, sufficient documentation should help though.
+- We would be renaming `hive_cluster_deployment_install_failure_total` and `hive_cluster_deployment_install_success_total` to ensure it ends with `seconds`, as it is a duration-based metric.
```
2uasimojo, line 133: This is concerning, as it's an _actual_ breaking change. Everything else will be invisible to the user when we deploy the new build, giving them the power to choose when they cut over. But this will silently break existing graphs/alerts.

Much as I hate to do it, let's clone the existing metrics instead of outright replacing them. Then we can remove the old copies after the deprecation period is over. Glad it's only two :P

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Can you please link that card here? ("We may in the future attempt to auto-generate these docs -- see [HIVE-2413](https://issues.redhat.com/browse/HIVE-2413)")

2uasimojo (commented):
Yup, this is getting real close. Only really one substantive change (schema compat discrepancy I missed when I suggested the current hiveconfig shape).

docs/enhancements/metricsConfig_redesign.md
suhanime: I had not mentioned so explicitly, but it would be cloning anyways because old way and new way will exist side-by-side until the deprecation period is over. Also, remember, the new way makes all metrics to be optional - so the user will need to specify the metric in hiveConfig - they will notice it early on. Included a note.
(I'm also glad it's only 2. I think I added these metrics a while ago actually, and the only reason we missed it at that time is because our naming conventions for metrics are not documented - just somehow implemented in code. That's why I want to correct it by documenting it, so it is easy for future maintenance)

docs/enhancements/metricsConfig_redesign.md
suhanime: Yeah, you're right. Let's not allow duplicate entries.

docs/enhancements/metricsConfig_redesign.md
suhanime: I meant, I have never actually implemented registering the metric in one controller and reporting it via another - given that a lot of these implementation details are handled by the metrics library, we would have to try it to know it. Regardless of how we choose to implement it (yeah, I'm making this a future-me problem), the result will be the hiveConfig ReadyCondition getting set to false, so that's the only thing I'm choosing to mention in the enhancement.

docs/enhancements/metricsConfig_redesign.md
suhanime: done

suhanime (commented):
@2uasimojo Final Pass?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### clusterDeploymentLabelSelector
+This would be a new feature, of type [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), and we'd use the LabelSelector.MatchLabels and/or LabelSelector.MatchExpressions to match the conditions in order to decide if a metric should be reported.
+This encapsulates slightly more advanced filter logic over the existing clusterDeployment labels. In the example above, `hive_foo_counter` and `hive_foo_gauge` metrics will only be reported for the clusterDeployments that are labelled with aro-snowflake and not in limited support.
+
+### Implementation Details / Notes
+
+- All the options configured for a metric within metricNames, will work in tandem with each other. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration
```
2uasimojo, line 137: Again, the reader will probably "get it", but this still isn't quiiite right. Let me see if I can reword...

```suggestion
`metricsToReport` is a list, each element of which requests reporting and configures filtering and customizations for the metrics provided in its `metricNames` list.

#### metricNames
`metricNames` would be a non-empty list of hive metrics that need to be reported, and can have optional customizations.
All customizations listed for an entry of `metricToReport` will apply to all the metrics provided in that entry's `metricNames` list. This allows for grouping filters for a list of relevant metrics.
A metric name must appear at most once across all `metricsToReport[].metricNames[]` in order to avoid ambiguity.
Implementation will be adapting the existing setup for optional metrics. However, instead of a shorthand camelCase key that is used for current `hiveconfig.spec.metricsConfig.metricsWithDuration`, we would now need the full metric name listed under `metricNames`.
In the example above, hive would only log `hive_foo_counter`, `hive_foo_gauge` and `hive_foo_histogram` metrics.

#### minimumDuration
Deprecate the current implementation of hiveConfig.spec.metricsConfig.metricsWithDuration, and change it to be reported as metricsConfig.metricsToReport[].minimumDuration. The implementation of using the duration as a threshold before we report the metric stays the same.
In the example above, `hive_foo_counter` and `hive_foo_gauge` will only be logged if the value reported for them exceeds 10 minutes.

#### additionalClusterDeploymentLabels
Deprecate current implementation of hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels and change it to be reported as metricsConfig.metricsToReport[].additionalClusterDeploymentLabels. Its implementation will not change.
In the example above, `hive_foo_counter` and `hive_foo_gauge` would report an additional label `prom_label_name`, its value corresponding to the value of `hive.openshift.io/cd-label-key` label on the corresponding clusterDeployment.

#### clusterDeploymentLabelSelector
This would be a new feature, of type [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), and we'd use the LabelSelector.MatchLabels and/or LabelSelector.MatchExpressions to match the conditions in order to decide if a metric should be reported.
This encapsulates slightly more advanced filter logic over the existing clusterDeployment labels. In the example above, `hive_foo_counter` and `hive_foo_gauge` metrics will only be reported for the clusterDeployments that are labelled with aro-snowflake and not in limited support.

### Implementation Details / Notes

- All the options configured for a `metricsToReport` entry will work in tandem with each other for each metric listed in its `metricNames`. For ex, if all possible options are specified for a metric, then that metric will be reported with the additional labels as per additionalClusterDeploymentLabels, and will be reported only if it matches the labels and/or expressions as per clusterDeploymentLabelSelector and if the duration to be reported exceeds the minimumDuration
```

docs/enhancements/metricsConfig_redesign.md
2uasimojo: Ah, okay, this may help:

We would be registering the metrics in hive-operator _only_ as a means to validate that they work properly.

We will *also* need to register them in the controllers. Those are separate processes running in separate pods, so they can't share one registration.

(But they can share source code, and should, since the registrations themselves should be identical in both places.)

```suggestion
These situations will result in hive-operator failing to deploy the controllers. It will set hiveConfig.Status.Conditions[ReadyCondition] to `"False"` with an appropriate error message.
```

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,164 @@
...
+Note that only the old design will support the mapping of camelCase key to metric for minimumDuration, the new way will require full metric names only. The old design will log `hive_cluster_deployment_install_failure_total` and `hive_cluster_deployment_install_success_total`, in the new design, they will be renamed to `hive_cluster_deployment_install_failure_total_seconds` and `hive_cluster_deployment_install_success_total_seconds`. 
+
+#### Failure modes
+
+These situations will result in metrics not getting registered and hiveConfig.Status.Conditions[ReadyCondition] will be set to false:.
+- metricsToReport.metricNames[$name] doesn't exist as a metric.
+- metricsToReport.metricNames[$name] is a metric without duration but minimumDuration was specified.
+- there are multiple entries of metricsToReport.metricNames[$name].
```
2uasimojo, line 154:
```suggestion
- `metricsToReport[].metricNames[$name]` doesn't exist as a metric.
- `metricNames` lists a metric without duration but `minimumDuration` is specified in the same `metricsToReport` entry..
- the same metric is mentioned more than once across all `metricsToReport[].metricNames[]`.
```

2uasimojo (commented):
If you agree with my suggestions, we can open this up for broader review now.

suhanime:
@2uasimojo Implemented all your suggestions and rebased, this is ready for a broader review.

@bmeng, @tzvatot  and @hongkailiu Can you please review this again? We made some changes, the biggest one is that we have decided to make all metrics optional (aka hive will not report any metric by default post the deprecation period).

@berenss Is it okay to assume ACM doesn't use Hive metrics, hence it won't be affected by these changes?

@maorfr Would you be able to help find an appropriate reviewer from AppSRE?

2uasimojo:
Thanks @suhanime! This lgtm to open for broader review.

suhanime:
@dustman9000 @patjlm @janboll Added you to the Reviewers. Please review at your earlier convenience.

@berenss and @tzvatot confirmed no impact to ACM and OCM

Waiting for the go-ahead from SREP, App-SRE and @hongkailiu

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,166 @@
...
+- We would no longer panic when the metricsConfig fails validation, instead we would be updating hiveConfig.Status.Conditions[ReadyCondition] to false, with appropriate reason and message. See [Failure Modes](#failure-modes) for more details.
+
+#### Deprecation
+
+In order to implement this change, we will have to deprecate the existing hiveConfig.spec.metricsConfig.metricsWithDuration and hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels.
+As a way to handle the deprecation, we will assert that in a hiveConfig, you can either provide the new configurations (aka `metricsToReport`) or the old way of hiveConfig.spec.metricsConfig.metricsWithDuration and hiveConfig.spec.metricsConfig.additionalClusterDeploymentLabels. If there's an empty metricsConfig, we would default to _old_ behaviour for the period of deprecation.
```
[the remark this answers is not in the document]
suhanime: - I suppose conversion is not the write word, I'm just stating you can either provide the new `metricsToReport` option in the config, or follow whatever is implemented right now, to avoid ambiguity.
- Correct, but I also wanted to point out that all the metrics will become optional with `metricsToReport`. So when you change your HiveConfig to use the new options, you would also have to compile a list of all the hive metrics you use or need, and you'll have to configure them in the hiveConfig, otherwise hive will not report them.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,166 @@
...
+    - [minimumDuration](#minimumduration)
+    - [additionalClusterDeploymentLabels](#additionalclusterdeploymentlabels)
+    - [clusterDeploymentLabelSelector](#clusterdeploymentlabelselector)
+  - [Implementation Details / Notes](#implementation-details--notes)
+    - [Deprecation](#deprecation)
+    - [Failure Modes](#failure-modes) 
+  - [Risks and Mitigations](#risks-and-mitigations)
+
+## Summary
+As a Hive user interested in customizing metrics, I would like to specify customizations for individual metrics.
```
dofinn, line 45: Problem statement is not entirely clear to me. It seems like there a multiple issues simmering regarding "how to effectively operate hive" and they are all trying to be solved with how hive produces metrics. 

I think for an effective metrics solution, we need a concrete operational narrative that defines the role of metrics within it. Currently metrics are being proposed as solutions to:
- observability
- alerting
- troubleshooting (inferring this)

IMO, stake holder expectations need to defined and agreed upon; including those that are not RH SRE. Then we can assess the solution against the problem statement.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,166 @@
...
+All customizations listed for an entry of `metricToReport` will apply to all the metrics provided in that entry's `metricNames` list. This allows for grouping filters for a list of relevant metrics.
+A metric name must appear at most once across all `metricsToReport[].metricNames[]` in order to avoid ambiguity.
+Implementation will be adapting the existing setup for optional metrics. However, instead of a shorthand camelCase key that is used for current `hiveconfig.spec.metricsConfig.metricsWithDuration`, we would now need the full metric name listed under `metricNames`.
+In the example above, hive would only log `hive_foo_counter`, `hive_foo_gauge` and `hive_foo_histogram` metrics.
+
+#### minimumDuration
+Deprecate the current implementation of hiveConfig.spec.metricsConfig.metricsWithDuration, and change it to be reported as metricsConfig.metricsToReport.metricNames[].minimumDuration. The implementation of using the duration as a threshold before we report the metric stays the same.
+In the example above, `hive_foo_counter` and `hive_foo_gauge` will only be logged if the value reported for them exceeds 10 minutes.
```
janboll, line 127: Isn't a counter just being incremented? Which value do you compare here?

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,166 @@
...
+
+These situations will result in hive-operator failing to deploy the controllers. It will set hiveConfig.Status.Conditions[ReadyCondition] to `"False"` with an appropriate error message.
+-  `metricsToReport[].metricNames[$name]` doesn't exist as a metric.
+- `metricNames` lists a metric without duration but `minimumDuration` is specified in the same `metricsToReport` entry.
+- the same metric is mentioned more than once across all `metricsToReport[].metricNames[]`.
+- additionalClusterDeploymentLabels key conflict with an existing fixed label key for that metric.
+- if deprecated methods are used along with the new metricsToReport. You can either choose the _old_ way or the _new_ way.
+
+
+### Risks and Mitigations
```
janboll, line 161: Do you have processes or tools in mind, to prevent acidentialy enabling a lot of metrics?
Or asked in another way, can you predict the impact/outcome of a config change?

docs/enhancements/metricsConfig_redesign.md
suhanime: I changed the summary to be more precise - does it help?
I can see your point about having concrete operational narrative around metrics, but to be clear, hive's job is to publish the metrics, and it is upto the consumer how they want to use them. You can very well use the same metric for both observability and alerting. I don't think prometheus and metrics are the solutions for troubleshooting, though you can definitely use them to notice the patterns.
I would also want to point out - as a part of this enhancement, we're just changing some internal code to make things easier - we're not introducing any new metrics. We've also gotten it reviewed from all major consumers of hive.

docs/enhancements/metricsConfig_redesign.md
```diff
@@ -0,0 +1,180 @@
...
+  - [Implementation Details / Notes](#implementation-details--notes)
+    - [Deprecation](#deprecation)
+    - [Failure Modes](#failure-modes) 
+  - [Risks and Mitigations](#risks-and-mitigations)
+
+## Summary
+As a Hive user interested in consuming the metrics it publishes, I would like to be able to configure the metrics as per my needs. 
+This includes the ability to choose which metrics are published, reduce the amount of observations for certain metrics unless it meets configured conditions, and the ability to add labels to the reported metrics for grouping and filtering purposes.
+
+## Motivation
```
2uasimojo, line 48: Excellent update here, thanks!

2uasimojo:
/lgtm

We may find little things to tweak/add as we implement, but I think this is ready to land. Nice work!

PR #2183, opened 2024-01-12
fix vendor is not put back and failed on MAC env after run update-codegen.sh

zhiweiyin318:
fix 2 parts in update.codegen.sh:
1. change to gsed on MAC env
2. fix the trap after exit


2uasimojo (commented):
Thanks for this!

(Taking your word for the mac thing...)

PR #2188, opened 2024-01-17
Add Fixed Labels to metrics doc

suhanime:
It can be confusing for consumers wanting to add labels to metrics with optional label support without being aware of the fixed labels already reported by those metrics, this commit documents the same to resolve the problem.

There's also the fix for cluster sync failing metric - recently it allows additional labels but document update was pending.


2uasimojo (commented):
Call me lazy, but I'm going to trust that you got all of these right.

Sure would be neat to figure out a way to generate these tables from code.

/lgtm

PR #2194, opened 2024-01-29
HIVE-2410: Add config for controller memory requests and limits

dlom:
xref: [HIVE-2410](https://issues.redhat.com//browse/HIVE-2410)
/assign @2uasimojo

### DIFF
--- a/apis/hive/v1/hiveconfig_types.go
+++ b/apis/hive/v1/hiveconfig_types.go
@@ -528,7 +528,13 @@ type ControllerConfig struct {
 	// Replicas specifies the number of replicas the specific controller pod should use.
 	// This is ONLY for controllers that have been split out into their own pods.
 	// This is ignored for all others.
+	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+	// Resources describes the compute resource requirements of the controller container.
+	// This is ONLY for controllers that have been split out into their own pods.
+	// This is ignored for all others.
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
 }
 
 // +kubebuilder:validation:Enum=clusterDeployment;clusterrelocate;clusterstate;clusterversion;controlPlaneCerts;dnsendpoint;dnszone;remoteingress;remotemachineset;machinepool;syncidentityprovider;unreachable;velerobackup;clusterprovision;clusterDeprovision;clusterpool;clusterpoolnamespace;hibernation;clusterclaim;metrics;clustersync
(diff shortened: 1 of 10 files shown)


dlom:
Side note: I renamed the `hiveContainer` variable because the name seemed inaccurate

apis/hive/v1/hiveconfig_types.go
```diff
@@ -529,6 +529,16 @@ type ControllerConfig struct {
...
+	// MemoryRequest specifies the memory request for the container that the controller is living in.
+	// The format for this string is the same as that specified in the container spec memory requests (e.g. "512Mi").
+	// This is ONLY for controllers that have been split out into their own pods.
+	// This is ignored for all others.
+	MemoryRequest *string `json:"memoryRequest,omitempty"`
+	// MemoryLimit specifies the memory limit for the container that the controller is living in.
+	// The format for this string is the same as that specified in the container spec memory limits (e.g. "512Mi").
+	// This is ONLY for controllers that have been split out into their own pods.
+	// This is ignored for all others.
+	MemoryLimit *string `json:"memoryLimit,omitempty"`
```
2uasimojo, line 541: Rather than bespoke fields that are limited to memory and also need to be parsed with logic duplicated from upstream, can we instead just make a `Resources` field of type corev1.ResourceRequirements, just like what's [defined](https://github.com/openshift/hive/blob/e21e524f15a2334c06aea4f4221c47e2a612cc78/vendor/k8s.io/api/core/v1/types.go#L2503) on Container itself? Then we can just copy the whole thing into the container spec and be done. Also makes it more natural-feeling for the user.

2uasimojo (commented):
Good start, everything is in the right place.

Please see comment inline.

`verify` fail is because codegen needs to be rerun (whenever you're done). `make update` *should* handle it.

2uasimojo:
The code change lgtm.

Waiting to make sure `verify` is happy.

Also discussing adding some e2e. If we can get that done quickish, we'll include it, otherwise we can do it in a fup.

The `security` fail is the same low-sev, no-fix-available-yet vuln we've been tracking for a couple of weeks. Can be overridden when we're ready to merge.

2uasimojo:
/test e2e

looks like a timing flake laying down RBAC

2uasimojo:
/lgtm
/override ci/prow/security
/hold

Unhold once live sniff test checks out.

dlom:
Sniff test looks good:

```bash
$ oc get statefulset -n hive hive-clustersync -o yaml | yq '.spec.template.spec.containers[0].resources'
limits:
  memory: 4Gi
requests:
  memory: 2Gi
$ oc get hiveconfig hive -o yaml | yq '.spec.controllersConfig.controllers[0].config.resources'
limits:
  memory: 4096Mi
requests:
  memory: 2048Mi
```

PR #2204, opened 2024-02-07
Reconcile MachineSets using MachinePool labels and AZs

lleshchi:
Manual backport of #2179 

The MachinePool controller reconciles MachineSets by generating the necessary MachineSets for a given MachinePool and matching generated MachineSets with existing remote MachineSets to determine what updates to MachineSet state should occur. Historically, the matching of MachineSets was determined by comparing MachineSet names. This process results in some edge cases, especially now that VSphere has introduced suffixes to MachineSet names to represent availability zones, which resulted in mismatching of MachineSet names across hive upgrades, where the hive vendored installer code that creates the generated MachineSets does not match the version which created the remote MachineSets. This resulted in odd behavior where duplicates of already existing MachineSets were being generated with only a suffix discrepancy.

The solution to this mismat dilemma is twofold:

For Day 0, the installmanager matches MachineSets to their corresponding MachinePool and patches the MachineSet manifests with a label indicating the name of the MachinePool they belong to.

For Day 2, the MachinePool controller checks the remote MachineSets for a hive MachinePool label. If the label matches the current MachinePool on which the controller is reconciling, the controller will check the availability zone of the MachineSet (as there can be multiple MachineSets per MachinePool with differing AZs), and if the AZ matches, the two MachineSets are considered matched. This removes any need to check or depend on the name of the remote MachineSets.

Important note: If a MachineSet does not have a MachinePool label, the MachinePool controller *will not* manage or consider the MachineSet. For manual creation or adoption of remote MachineSets, the MachinePool label must be added manually to ensure management by Hive.

[HIVE-2254](https://issues.redhat.com//browse/HIVE-2254)


2uasimojo:
/lgtm

Hell, bot??

PR #2193, opened 2024-01-22
Bump to k8s 0.29.1

2uasimojo:
[HIVE-2351](https://issues.redhat.com//browse/HIVE-2351)

PR #2206, opened 2024-02-09
Copy ClusterVersion.Status into ClusterDeployment

2uasimojo:
Add a new field, ClusterDeployment.Status.ClusterVersionStatus.

At the behest of an (undocumented, unsupported) annotation, `hive.openshift.io/sync-clusterversion-status: "true"` (note: string `"true"` since it's an annotation) our clusterversion controller (the same one that updates the `hive.openshift.io/version*` labels) will populate it with a wholesale copy of the spoke's ClusterVersion.Status field.

NOTE: This requires that our vendor of openshift/api be kept up to date, as e.g. new enum values for `capabilities` will break us.

[HIVE-2366](https://issues.redhat.com//browse/HIVE-2366)

### DIFF
--- a/apis/hive/v1/clusterdeployment_types.go
+++ b/apis/hive/v1/clusterdeployment_types.go
@@ -381,6 +381,12 @@ type ClusterDeploymentStatus struct {
 	// perform the installation.
 	// +optional
 	Platform *PlatformStatus `json:"platformStatus,omitempty"`
+
+	// ClusterVersionStatus is a wholesale copy of the Status section of the spoke cluster's
+	// `clusterversion version` object. This is not officially supported, and is only populated
+	// on request.
+	// +optional
+	ClusterVersionStatus *configv1.ClusterVersionStatus `json:"clusterVersionStatus,omitempty"`
 }
 
 // ClusterDeploymentCondition contains details for the current condition of a cluster deployment
--- a/pkg/controller/clusterversion/clusterversion_controller.go
+++ b/pkg/controller/clusterversion/clusterversion_controller.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	diff "github.com/google/go-cmp/cmp"
 	openshiftapiv1 "github.com/openshift/api/config/v1"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
@@ -167,7 +168,13 @@ func (r *ReconcileClusterVersion) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, err
 	}
 
+	// This Update()s CD iff necessary
 	if err := r.updateClusterVersionMetadata(cd, clusterVersion, cdLog); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// This Update()s CD.Status iff necessary
+	if err := r.syncClusterVersionStatus(cd, clusterVersion, cdLog); err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -236,6 +243,38 @@ func (r *ReconcileClusterVersion) updateClusterVersionMetadata(cd *hivev1.Cluste
 
 	if err := r.Update(context.TODO(), cd); err != nil {
 		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error updating cluster deployment metadata")
+		return err
+	}
+	return nil
+}
+
+func (r *ReconcileClusterVersion) syncClusterVersionStatus(cd *hivev1.ClusterDeployment, clusterVersion *openshiftapiv1.ClusterVersion, cdLog log.FieldLogger) error {
+	if populate, err := strconv.ParseBool(cd.Annotations[constants.SyncClusterVersionStatusAnnotation]); err == nil && populate {
+		logFields := log.Fields{}
+		if cd.Status.ClusterVersionStatus == nil {
+			logFields["reason"] = "previously unset"
+		} else {
+			delta := diff.Diff(*cd.Status.ClusterVersionStatus, clusterVersion.Status)
+			if delta == "" {
+				cdLog.Debug("clusterversion status has not changed, nothing to update")
+				return nil
+			}
+			logFields["reason"] = "changed"
+			logFields["diff"] = delta
+		}
+		cdLog.WithFields(logFields).Debug("updating clusterversion status")
+		cd.Status.ClusterVersionStatus = &clusterVersion.Status
+	} else {
+		// If this got switched off, clear the field so it doesn't go stale
+		if cd.Status.ClusterVersionStatus == nil {
+			return nil
+		}
+		cdLog.Debug("clearing clusterversion status")
+		cd.Status.ClusterVersionStatus = nil
+	}
+
+	if err := r.Status().Update(context.TODO(), cd); err != nil {
+		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "error updating cluster deployment's clusterVersionStatus")
 		return err
 	}
 	return nil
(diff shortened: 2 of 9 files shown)


2uasimojo:
/hold

- This currently represents a hybrid approach, where we've added an actual API field (`ClusterDeployment.Status.ClusterVersionStatus`) for the content, but are using an annotation (`hive.openshift.io/sync-clusterversion-status`) to activate it. Design review is needed to decide whether we want to do that, or both in the API, or both in annotations.
- The current approach also types our new field by importing the entire ClusterVersionStatus sub-object schema. This is in contrast to [hypershift's approach](https://github.com/openshift/hypershift/blob/d66aae7cde24d9e4e8787b832a1d4acb74cc0848/control-plane-operator/hostedclusterconfigoperator/controllers/hcpstatus/hcpstatus.go#L91), where they are hand-picking certain fields to copy in, and "converting" some of them to locally-defined types. A third alternative is to blat the content down as a JSON blob (in a RawExtension if we stick with the API approach; or as an ugly escaped string if in an annotation, as these things seem to be done "normally").
- As currently written this is busted for newer OCP versions of the spoke because breaking changes were made to ClusterVersion.Status in openshift/api somewhere between what we [currently vendor](https://github.com/openshift/hive/blob/43cd29ff7c9242d070d5a7ba8923fbcb5c954295/go.mod#L37) (`20230720094506-afcbe27aec7c`) and 4.15.0-rc.2 (I haven't narrowed it down further than that yet). Upgrading o/api pulls k8s deps bumps to 0.29.1, which for us also drags in a new revendor of openshift/installer. In other words, it stacks this behind https://github.com/openshift/hive/pull/2193 which itself has a chain of prereqs. That's assuming we continue with the current design of importing the entire ClusterVersionStatus schema. Any of the other approaches -- annotation, RawExtension, or piecemeal copyin -- would be able to ignore discrepancies like this.
- We'll want to add some UT for this, wherever it lands.

/cc @vkareh @cblecker @jewzaam @LalatenduMohanty @gvanderpotte @wanghaoran1988 @wking 
cf. https://docs.google.com/document/d/1hgMiDYN9W60BEIzYCSiu09uV4CrD_cCCZ8As2m7Br1s/edit

pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -167,10 +168,16 @@ func (r *ReconcileClusterVersion) Reconcile(ctx context.Context, request reconci
...
 	}
 
+	// This Update()s CD iff necessary
 	if err := r.updateClusterVersionMetadata(cd, clusterVersion, cdLog); err != nil {
 		return reconcile.Result{}, err
 	}
 
+	// This Update()s CD.Status iff necessary
+	if err := r.syncClusterVersionStatus(cd, clusterVersion, cdLog); err != nil {
+		return reconcile.Result{}, err
```
wking, line 178: if we fail here via the `error parsing ... annotation` branch or some such, we'd be bailing out before the `requeueAfter` portion.  There are other existing examples of that error-handling approach in this function, but I'm curious about how the system recovers from those failures.  Is there some higher-level logic that launches a "maybe they've fixed the busted annotation value" recovery attempt, even if we don't ask for a requeue explicitly here?

2uasimojo: Returning a non-nil error automatically requeues the request immediately. We do indeed have this same pattern throughout (in _all_ of our controllers), but we do make the effort to treat "intervention required" errors differently so they don't just hopelessly thrash into exponential backoff. It would make sense to do so here. In fact, specifically for `ParseBool()`, [we](https://github.com/openshift/hive/blob/08e49876b54fb77879d4362e4339df68318f4d93/pkg/controller/argocdregister/argocdregister_controller.go#L156) [treat](https://github.com/openshift/hive/blob/602f884d6bbb14547ffb5c99aaed9566ee5aa307/pkg/controller/awsprivatelink/awsprivatelink_controller.go#L169) [unparseable](https://github.com/openshift/hive/blob/14a8640f00100fd7c565606863ff28b5c287c0f2/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L156) [as](https://github.com/openshift/hive/blob/14a8640f00100fd7c565606863ff28b5c287c0f2/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L363) ["false"](https://github.com/openshift/hive/blob/08e49876b54fb77879d4362e4339df68318f4d93/pkg/controller/controlplanecerts/controlplanecerts_controller.go#L148)  [_almost_](https://github.com/openshift/hive/blob/14a8640f00100fd7c565606863ff28b5c287c0f2/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1058)  [everywhere](https://github.com/openshift/hive/blob/14a8640f00100fd7c565606863ff28b5c287c0f2/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L2297) [we](https://github.com/openshift/hive/blob/9615c77cd786dd82079a9dacd60b64c882f56327/pkg/controller/clusterrelocate/clusterrelocate_controller.go#L182) [use](https://github.com/openshift/hive/blob/08e49876b54fb77879d4362e4339df68318f4d93/pkg/controller/clusterstate/clusterstate_controller.go#L125) [it](https://github.com/openshift/hive/blob/a8c3d1cea206ca48ba7d7250041686a4261802ec/pkg/controller/clusterversion/clusterversion_controller.go#L133). (The only [exception](https://github.com/openshift/hive/blob/7d9bf8a09eccdf70176cda14243e1d35c17d77d6/pkg/controller/clusterdeprovision/clusterdeprovision_controller.go#L91) I can find would actually crash on controller startup, and that one is [controlled](https://github.com/openshift/hive/blob/7d9bf8a09eccdf70176cda14243e1d35c17d77d6/pkg/operator/hive/hive.go#L232-L239) by hive-operator, so that should be impossible.)

TL;DR: You're right to call this out, and we should do the same here. Will fix.

2uasimojo: Oh, and

> Is there some higher-level logic that launches a "maybe they've fixed the busted annotation value" recovery attempt, even if we don't ask for a requeue explicitly here?

this is now moot, but fixing the busted annotation would [trigger our Watch()](https://github.com/openshift/hive/blob/ffd8a176e8085d592650e62504b1895be0f09103/pkg/controller/clusterversion/clusterversion_controller.go#L87) and run the reconciler.

wking: Thanks for the context!  And the approach in your new ffd8a176e808 that treats failed parsing the same as `false` makes sense to me too.  Feel free to mark this thread resolved (I don't have the ability to do that in this repo/pull myself).

apis/hive/v1/clusterdeployment_types.go
```diff
@@ -381,6 +381,12 @@ type ClusterDeploymentStatus struct {
 	// perform the installation.
 	// +optional
 	Platform *PlatformStatus `json:"platformStatus,omitempty"`
+
+	// ClusterVersionStatus is a wholesale copy of the Status section of the spoke cluster's
+	// `clusterversion version` object. This is not officially supported, and is only populated
+	// on request.
+	// +optional
+	ClusterVersionStatus *configv1.ClusterVersionStatus `json:"clusterVersionStatus,omitempty"`
```
lucasponce, line 389: In case it's relevant to compare with Hypershift API

https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L1787

I'm not familiar with the motivation to create a struct rather than fetching the whole object.

lucasponce:
/lgtm

Minor comment about the decision that took Hypershift to copy the struct instead to reuse the type from openshift api.

But PR looks good, thanks Eric.

apis/hive/v1/clusterdeployment_types.go
2uasimojo: Yup, we definitely considered doing it that way. Ultimately we landed on this approach mainly because it makes it so much easier to carry changes forward: we just have to revendor o/api. That's especially easy for hive as we don't operate under the same long release cadence as OCP itself. The only serious concern was CR space, but we convinced ourselves with live data that we're nowhere near the danger zone in that regard.

PR #2207, opened 2024-02-12
Adding a new reason to AWSDeniedBySCP install failure

Nikokolas3270:


2uasimojo:
This needs to get into bindata -- please do a `make update`.

Also, we could add a separate explicit unit test for this path. But meh, up to you, I won't hold up for it.

Nikokolas3270:
> This needs to get into bindata -- please do a `make update`.
> 
> Also, we could add a separate explicit unit test for this path. But meh, up to you, I won't hold up for it.

Thanks for your feedback @2uasimojo:
- I regenerated the `bindata` file
- Also I added some unittest :)

2uasimojo (commented):
Thanks!

/lgtm

PR #2208, opened 2024-02-12
Nix and stop generating informers, listers

2uasimojo:
We haven't gone and confirmed this, but we think that k8s.io/code-generator's `generate-groups.sh` when invoked with `"all"` used to generate fewer things, and as they added things over the years, we ended up generating them and checking them into our repository even though we didn't need them and never used them.

`informers` and `listers` are two of these things.

This commit gets rid of the code those generators generated, and switches our invocation of `generate-groups.sh` to explicitly list the generators we intend to invoke, now omitting those two.

[HIVE-2417](https://issues.redhat.com//browse/HIVE-2417)


2uasimojo:
Note: this is only part of HIVE-2417. I intend to see if we can get rid of clientset and applyconfiguration as well. But since we're actually using those in code, I'll have to cut over to our usual client that we use everywhere else, which will be a slightly deeper effort, and carry slightly more risk, so I wanted to do it separately.

PR #2212, opened 2024-02-19
Remove uses of yaml-patch

lleshchi:
The [yaml-patch](https://github.com/krishicks/yaml-patch) package is now abandoned. Replaced with [gjson](https://github.com/tidwall/gjson) and [sjson](https://github.com/tidwall/sjson) in appropriate places.

[HIVE-2397](https://issues.redhat.com/browse/HIVE-2397)

### DIFF
--- a/pkg/clusterresource/builder.go
+++ b/pkg/clusterresource/builder.go
@@ -5,12 +5,12 @@ import (
 	"time"
 
 	"github.com/ghodss/yaml"
-	yamlpatch "github.com/krishicks/yaml-patch"
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/installer/pkg/ipnet"
 	installertypes "github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/validate"
 	"github.com/pkg/errors"
+	"github.com/tidwall/sjson"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -419,17 +419,21 @@ func (o *Builder) generateInstallConfigSecret() (*corev1.Secret, error) {
 	// Remove metadataService field from machinepool platform within installconfig.
 	// TODO: Remove this once https://bugzilla.redhat.com/show_bug.cgi?id=2098299 has been addressed.
 	if installConfig.Platform.AWS != nil {
-		ops := yamlpatch.Patch{
-			yamlpatch.Operation{
-				Op:   "remove",
-				Path: yamlpatch.OpPath("/compute/0/platform/aws/metadataService"),
-			},
-			yamlpatch.Operation{
-				Op:   "remove",
-				Path: yamlpatch.OpPath("/controlPlane/platform/aws/metadataService"),
-			},
+		d, err = yaml.YAMLToJSON(d)
+		if err != nil {
+			return nil, errors.Wrap(err, "error converting install-config.yaml to JSON")
 		}
-		modifiedBytes, err := ops.Apply(d)
+		d, err = sjson.DeleteBytes(d, `compute.0.platform.aws.metadataService`)
+		if err != nil {
+			return nil, errors.Wrap(err, "error removing metadataService field from install-config.yaml")
+		}
+		d, err = sjson.DeleteBytes(d, `controlPlane.platform.aws.metadataService`)
+		if err != nil {
+			return nil, errors.Wrap(err, "error removing metadataService field from install-config.yaml")
+		}
+
+		// Convert back to yaml
+		modifiedBytes, err := yaml.JSONToYAML(d)
 		if err != nil {
 			return nil, errors.Wrap(err, "error patching install-config.yaml to remove metadataService field")
 		}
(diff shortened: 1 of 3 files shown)


pkg/clusterresource/builder.go
```diff
@@ -419,17 +419,21 @@ func (o *Builder) generateInstallConfigSecret() (*corev1.Secret, error) {
...
+		if err != nil {
+			return nil, errors.Wrap(err, "error converting install-config.yaml to JSON")
+		}
+		d, err = sjson.DeleteBytes(d, `compute.0.platform.aws.metadataService`)
+		if err != nil {
+			return nil, errors.Wrap(err, "error removing metadataService field from install-config.yaml")
+		}
+		d, err = sjson.DeleteBytes(d, `controlPlane.platform.aws.metadataService`)
+		if err != nil {
+			return nil, errors.Wrap(err, "error removing metadataService field from install-config.yaml")
```
2uasimojo, line 432: These error messages should be distinguishable so we know which path broke.
```suggestion
			return nil, errors.Wrap(err, "error removing metadataService field from compute section of install-config.yaml")
		}
		d, err = sjson.DeleteBytes(d, `controlPlane.platform.aws.metadataService`)
		if err != nil {
			return nil, errors.Wrap(err, "error removing metadataService field from controlPlane section of install-config.yaml")
```

pkg/clusterresource/builder.go
```diff
@@ -438,17 +442,19 @@ func (o *Builder) generateInstallConfigSecret() (*corev1.Secret, error) {
...
+		if err != nil {
+			return nil, errors.Wrap(err, "error converting install-config.yaml to JSON")
+		}
+		d, err = sjson.DeleteBytes(d, `compute.0.platform.azure.osImage`)
+		if err != nil {
+			return nil, errors.Wrap(err, "error removing osImage field from install-config.yaml")
+		}
+		d, err = sjson.DeleteBytes(d, `controlPlane.platform.azure.osImage`)
+		if err != nil {
+			return nil, errors.Wrap(err, "error removing osImage field from install-config.yaml")
```
2uasimojo, line 455: Likewise
```suggestion
			return nil, errors.Wrap(err, "error removing osImage field from compute section of install-config.yaml")
		}
		d, err = sjson.DeleteBytes(d, `controlPlane.platform.azure.osImage`)
		if err != nil {
			return nil, errors.Wrap(err, "error removing osImage field from controlPlane section of install-config.yaml")
```

pkg/installmanager/installmanager.go
```diff
@@ -853,7 +851,7 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, mapPoolsBy
 	// For details, see HIVE-1802.
 	// Configure Hive MachineSet manifests with a label for the MachinePool name
 	var awsClient awsclient.Client
-	var vpcID string
+	vpcID := ""
 	addSecurityGroup := (workerMachinePool != nil) && (metav1.HasAnnotation(workerMachinePool.ObjectMeta, constants.ExtraWorkerSecurityGroupAnnotation))
```
2uasimojo, line 855: nit: Seems like we could get rid of the `addSecurityGroup` bool and just use the non-empty-ness of `vpcID` to indicate same. It would be one less arg to pass into the already-cluttered patchMachineSetManifest func.

Similarly, I feel like we could get rid of `addMPLabel`. Inline `len(mapPoolsByType) > 0` on L874 and then just let the `exists` on L1014 filter from there.

pkg/installmanager/installmanager.go
```diff
@@ -952,7 +933,6 @@ func (m *InstallManager) generateAssets(cd *hivev1.ClusterDeployment, mapPoolsBy
 	return nil
 }
 
-// TODO: Combine this with patchWorkerMachineSetManifest once HIVE-2397 is completed
 func writeModifiedBytesToFile(modifiedManifestBytes []byte, d fs.DirEntry, path string, logger log.FieldLogger) error {
 	if modifiedManifestBytes != nil {
```
2uasimojo, line 937: nit: could invert this condition with a `return` and save some indent.

2uasimojo (commented):
This seems sane so far. Most comments along the lines of, "Since we're in here refactoring anyway..."

pkg/installmanager/installmanager.go
```diff
@@ -969,60 +949,93 @@ func writeModifiedBytesToFile(modifiedManifestBytes []byte, d fs.DirEntry, path
...
+				"name":   "vpc-id",
+				"values": []string{vpcID},
+			}
+
+			// Add the security group name obtained from the ExtraWorkerSecurityGroupAnnotation
+			// annotation found on the MachinePool to the existing tag:Name filter in the worker
+			// MachineSet manifest
+			securityGroupFilterValue := pool.Annotations[constants.ExtraWorkerSecurityGroupAnnotation]
+
+			manifestBytesJson, err = sjson.SetBytes(manifestBytesJson, `spec.template.spec.providerSpec.value.securityGroups.0.filters.0.values.0`, securityGroupFilterValue)
```
2uasimojo, line 990: Eh, I think this may be wrong: it is _replacing_ the existing value, whereas the original code is _adding_.

We start with
```yaml
          securityGroups:
          - filters:
            - name: tag:Name
              values:
              - test-czcpt-worker-sg
```

and we want this line to result in
```yaml
          securityGroups:
          - filters:
            - name: tag:Name
              values:
              - test-czcpt-worker-sg
              - value-from-annotation
```

but I believe it's instead ending up as
```yaml
          securityGroups:
          - filters:
            - name: tag:Name
              values:
              - value-from-annotation
```

Need to confirm via UT. Existing tests are not sufficient: [these additions](https://github.com/openshift/hive/pull/2251/commits/6de60ba5d9af04001021d516145f63268adadebb#diff-e9eea65e1b515787afedc1c1e6f448892582a3f8604589b56e4773c3d8880a4f) (which are part of the needed rebase) get partway there, but per [this comment](https://github.com/openshift/hive/pull/2251/commits/6de60ba5d9af04001021d516145f63268adadebb#r1588428431) we still need to add a check for the above.

2uasimojo:
/remove-lifecycle rotten

@lleshchi bump

2uasimojo:
> /hold After completing revisions, realized clusterpool controller and validating webhooks were not included in the original changes. Working on those now. We can wait for that work to be completed before merging this PR or create a seperate PR for those updates.

I'm fine doing it in a separate PR.

I'll review this tomorrow.

2uasimojo:
> > /hold After completing revisions, realized clusterpool controller and validating webhooks were not included in the original changes. Working on those now. We can wait for that work to be completed before merging this PR or create a seperate PR for those updates.
> 
> I'm fine doing it in a separate PR.

I take it back. I don't want to commit to this until we understand what it's going to look like for ClusterDeploymentCustomizations, which accept a very yamlpatch-ish object notation.

2uasimojo:
> I take it back. I don't want to commit to this until we understand what it's going to look like for ClusterDeploymentCustomizations, which accept a very yamlpatch-ish object notation.

Summary of offline discussion: we're going to pivot to https://github.com/evanphx/json-patch instead of sjson. The delta will be much smaller, especially when we get to ClusterDeploymentCustomizations, and thus more provably correct/backward-compatible.

2uasimojo:
Closed in favor of #2498

PR #2214, opened 2024-02-20
DNM: PoC Spot Instances for masters and workers

2uasimojo:
[HIVE-2419](https://issues.redhat.com//browse/HIVE-2419)

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -17,7 +17,6 @@ import (
 	"strings"
 	"time"
 
-	yamlpatch "github.com/krishicks/yaml-patch"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -76,7 +75,6 @@ import (
 	"github.com/openshift/hive/pkg/resource"
 	k8slabels "github.com/openshift/hive/pkg/util/labels"
 	"github.com/openshift/hive/pkg/util/scheme"
-	yamlutils "github.com/openshift/hive/pkg/util/yaml"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


2uasimojo:
Oh, yeah, definitely don't merge this as is.

/hold

pkg/installmanager/installmanager.go
```diff
@@ -964,11 +964,37 @@ func patchMachineSetManifest(manifestBytes []byte, poolsByType map[string]*hivev
...
+	default:
+		goto normal
+	}
+	manifestBytesJson, err = sjson.SetBytes(manifestBytesJson, path+`spec.providerSpec.value.spotMarketOptions`, map[string]string{})
+	if err != nil {
+		logger.WithError(err).Errorf("error adding spot-ness to %s manifests", kind)
+		return nil, err
+	}
+	modified = true
+	goto modified
```
2uasimojo, line 990: Oh, if MachineSet, we want to fall through. This is what's causing the unit test failures. Not too worried about it for this PoC... but I think it might result in duplicate MachineSets being created because we'll come up with the default worker MachineSets unlabeled for hive ownership, causing our MachinePool to make moar. Except it'll make them with conflicting names? So maybe it'll error? Bleh, we'll see.

2uasimojo: Yeah, checked it in my test bed, the mp controller pukes:

```
time="2024-02-20T21:04:57.036Z" level=error msg="unable to create machine set" controller=machinepool error="machinesets.machine.openshift.io \"efried-spot-poc-lqs5b-worker-us-east-1a\" already exists" machinePool=sstest/efried-spot-poc-worker reconcileID=j9zz79sh
```

which means the machine scaling bits of e2e won't work. Fixing...

2uasimojo:
/test e2e

Let's get a few runs in and see if we can get it to flake.

PR #2227, opened 2024-03-11
Automatically apply area/unit-tests label to PRs

celebdor:
This change will have prow apply an area/unit-tests label to any PR that changes a unit test file inside the "pkg" dir.

### DIFF
--- a/pkg/OWNERS
+++ b/pkg/OWNERS
@@ -1,0 +1,4 @@
+filters:
+  "_test\\.go$":
+     labels:
+     - area/unit-tests


2uasimojo:
Please, what is the intended purpose of this label?

2uasimojo:
> Please, what is the intended purpose of this label?

QE has asked to be notified of PRs that touch unit tests. This will allow them to do so at will via a query.

/lgtm

2uasimojo:
@celebdor I think you may need to add this label to https://github.com/openshift/release/blob/master/core-services/prow/02_config/_labels.yaml for this to work.

PR #2229, opened 2024-03-11
Delete MachineAutoscalers that would get MaxReplicas==0

2uasimojo:
MachineAutoscalers are not allowed to have MaxReplicas==0.

PR #2215 / bce2d472 partially fixed the scenario where a MachinePool's autoscaling maxReplicas is less than the number of AZs by causing such MAs not to be *created*. However, when reducing the MachinePool's maxReplicas to below the number of AZs, we were still trying to update *existing* MAs to have MaxReplicas==0. This commit adjusts the logic to delete them instead.

[HIVE-2415](https://issues.redhat.com//browse/HIVE-2415)

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -843,7 +843,7 @@ func (r *ReconcileMachinePool) syncMachineAutoscalers(
 		for i, ms := range machineSets {
 			minReplicas, maxReplicas := getMinMaxReplicasForMachineSet(pool, machineSets, i)
 			found := false
-			for _, rMA := range remoteMachineAutoscalers.Items {
+			for j, rMA := range remoteMachineAutoscalers.Items {
 				if ms.Name == rMA.Name {
 					found = true
 					objectModified := false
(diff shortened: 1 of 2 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -853,25 +853,29 @@ func (r *ReconcileMachinePool) syncMachineAutoscalers(
 						maLog.WithField("desired", minReplicas).
 							WithField("observed", rMA.Spec.MinReplicas).
 							Info("min replicas out of sync")
-						rMA.Spec.MinReplicas = minReplicas
+						remoteMachineAutoscalers.Items[j].Spec.MinReplicas = minReplicas
```
2uasimojo, line 856: Note to reviewers: Had to change this block to operate on the original (it was previously operating on the loop variable, which was an ephemeral copy scoped to this block) so we can detect maxReplicas==0 below in the deletion discovery path.

pkg/controller/machinepool/machinepool_controller_test.go
```diff
@@ -1294,8 +1326,8 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 					for _, rMS := range rMSL.Items {
 						if eMS.Name == rMS.Name {
 							found = true
-							assert.Equal(t, *eMS.Spec.Replicas, *rMS.Spec.Replicas, "Replicas")
-							assert.Equal(t, eMS.Generation, rMS.Generation, "Generation")
+							assert.Equal(t, *eMS.Spec.Replicas, *rMS.Spec.Replicas, "Replicas for %s", rMS.Name)
+							assert.Equal(t, eMS.Generation, rMS.Generation, "Generation for %s", rMS.Name)
```
2uasimojo, line 1330: ntr: mostly-unrelated change, makes debugging easier.

2uasimojo:
/override ci/prow/security

addressed by #2231

PR #2243, opened 2024-03-14
OCM-6713 | feat: allow user defined tags on machinepool creation

gdbranco:
[OCM-6713](https://issues.redhat.com//browse/OCM-6713)

Create AWS tags for machine pool on cluster creation time (day 1) and post-cluster creation time (day 2) on ROSA Classic.  Modification of AWS tags are out-of-scope.

Hive to support additional map of AWS tags on MachinePools
These are AWS user-defined / custom tags, not to be confused with node labels

## Implementation Strategy
The strategy opted for this PR is to merge the user tags with ClusterDeployment.Spec.Platform.AWS.UserTags, where the machine pool level field takes precedence when keys collide.

### DIFF
--- a/apis/hive/v1/aws/machinepool.go
+++ b/apis/hive/v1/aws/machinepool.go
@@ -36,6 +36,12 @@ type MachinePoolPlatform struct {
 	//
 	// +optional
 	AdditionalSecurityGroupIDs []string `json:"additionalSecurityGroupIDs,omitempty"`
+
+	// UserTags contains the user defined tags to be supplied for the ec2 instance.
+	// Note that these will be merged with ClusterDeployment.Spec.Platform.AWS.UserTags, with
+	// this field taking precedence when keys collide.
+	// +optional
+	UserTags map[string]string `json:"userTags,omitempty"`
 }
 
 // SpotMarketOptions defines the options available to a user when configuring
--- a/apis/hive/v1/aws/zz_generated.deepcopy.go
+++ b/apis/hive/v1/aws/zz_generated.deepcopy.go
@@ -82,6 +82,13 @@ func (in *MachinePoolPlatform) DeepCopyInto(out *MachinePoolPlatform) {
 		*out = make([]string, len(*in))
 		copy(*out, *in)
 	}
+	if in.UserTags != nil {
+		in, out := &in.UserTags, &out.UserTags
+		*out = make(map[string]string, len(*in))
+		for key, val := range *in {
+			(*out)[key] = val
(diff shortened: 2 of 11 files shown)


apis/hive/v1/aws/machinepool.go
```diff
@@ -36,6 +36,10 @@ type MachinePoolPlatform struct {
 	//
 	// +optional
 	AdditionalSecurityGroupIDs []string `json:"additionalSecurityGroupIDs,omitempty"`
+
+	// Tags contains the user defined tags to be supplied for the ec2 instance
```
2uasimojo, line 40:
```suggestion
	// Tags contains the user defined tags to be supplied for the ec2 instance.
	// Note that these will be merged with ClusterDeployment.Spec.Platform.AWS.UserTags, with
	// this field taking precedence when keys collide.
```

2uasimojo (commented):
This looks great. Just one minor request to beef up the docstring on the new field. (Don't forget to regenerate.)

Also, would you please expand the commit message and PR description to include a high-level summary of the design points in the card? That'll make it easier for future-us to find later.

Thanks!

2uasimojo:
@lleshchi would you mind having a look at this as well please?

lleshchi:
Looks good to me as well!

2uasimojo:
Thanks @gdbranco 

/lgtm

2uasimojo:
/lgtm cancel

looks like this needs a `make update vendor`

gdbranco:
@2uasimojo Resolved. Thanks!

2uasimojo:
I wonder why the new field *isn't* making it into deepcopy. Let me poke at it this afternoon. I'll really try to have this in by EOD.

2uasimojo:
I dorked around with this and I think the problem is that the PR as it currently stands has the deepcopy delta in vendor/github.com/openshift/hive/apis/hive/v1/aws/zz_generated.deepcopy.go but not in apis/hive/v1/aws/zz_generated.deepcopy.go

It's a quirk of `verify` that it's complaining about the vendor/ one being extra, rather than complaining about the non-vendor/ one being missing.

But when I download this PR and `make update vendor` locally, the non-vendor/ one gets added in, which I think is where we want to end up.

I'm trying it with #2247. Meanwhile, feel free to do the same to this PR, and also squash commits for prettiness :)

2uasimojo:
> I'm trying it with #2247.

This seems to have worked. See if you can get the same result; and please also squash commits.

PR #2245, opened 2024-03-20
Draft msets subnet filters

lleshchi:


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -572,16 +576,69 @@ func matchFailureDomains(gMS *machineapi.MachineSet, rMS machineapi.MachineSet,
...
+			return true, subnets, nil
+		}
+	}
+
+	return fd1.Equal(fd2), nil, nil
+}
+
+func getSubnetIDByFilter(subnet *v1.AWSResourceReference, actuator AWSActuator) ([]*ec2.Subnet, error) {
+	for _, filter := range *subnet.Filters {
+		if filter.Name == "tag:Name" { // FIXME use all filters or just this one?
```
2uasimojo, line 616: All filters. The point is to find the same subnet(s) that AWS would when fed this filter from the Machine itself. So just `req := &ec2.DescribeSubnetsInput{Filters: subnet.Filters}` and go. No outer loop.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -572,16 +576,69 @@ func matchFailureDomains(gMS *machineapi.MachineSet, rMS machineapi.MachineSet,
...
+			req := &ec2.DescribeSubnetsInput{
+				Filters: []*ec2.Filter{
+					{
+						Name:   aws.String("tag:Name"),
+						Values: aws.StringSlice(filter.Values),
+					},
+				},
+			}
+			resp, err := actuator.awsClient.DescribeSubnets(req)
+			if err != nil {
```
2uasimojo, line 626: nts: Noodle whether we should special-case 404 here. Thinking AWS equiv of "this MachineSet is in a different ResourceGroup". Not sure off the top whether we support that for AWS.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -572,16 +576,69 @@ func matchFailureDomains(gMS *machineapi.MachineSet, rMS machineapi.MachineSet,
...
+				return nil, err
+			}
+			if len(resp.Subnets) == 0 {
+				return nil, fmt.Errorf("no subnets found for filter %v", filter)
+			}
+			// compare subnets to generated machineset
+			// FIXME do I compare all subnets in list? How does this work
+			if len(resp.Subnets) > 1 {
+				return nil, fmt.Errorf("multiple subnets found for filter %v", filter)
+			}
```
2uasimojo, line 636: Yeah, gosh, I really don't know. Might be worth asking someone in MAPI-land if we're expecting exactly one subnet, and should error if we find otherwise. Is the public/private thing still in play?

2uasimojo (commented):
Lots going on here. But I think the approach we discussed the other day is still going to make this tighter and less messy. It was something like:
- Convert the remote machineset list into a slice of some new struct where we can keep track of whether we've looked up and modified the subnet.
- Now there's no need to do that lookup in the matcher: they should just be comparable like we were expecting in the first place.
- Fold in the bool from the scrubbed slice when deciding whether to post an Update() to the remote MachineSet.

As for how to shunt that slice around, have a look at how the GCPActuator is [determining](https://github.com/openshift/hive/blob/61457bbe0fe5f975b835dd2c0ceb1d5bb7d34e5f/pkg/controller/machinepool/gcpactuator.go#L94) `network` and `subnet`. The point is that we have the luxury of
- [Passing](https://github.com/openshift/hive/blob/8d2b041e42754b51313f976aaff04bae62157683/pkg/controller/machinepool/machinepool_controller.go#L1177) anything we want -- in this case `remoteMachineSets` -- into the actuator builder
- Putting anything we want into the platform-specific actuator struct. In this case perhaps it's the wrapped list of remote msets.

Then when it comes time to use that thing from the main mp controller, just make sure you've wrapped the appropriate functionality in an actuator interface func, as suggested inline here. So only that func -- and not the main mp controller logic -- needs to know that that field exists and how to use it.

pkg/controller/machinepool/awsactuator.go
```diff
@@ -363,14 +364,47 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
...
+					CIDR: aws.StringValue(subnet.CidrBlock),
+				} // FIXME confirm
+
+				if subnets[zone].ID == "" {
+					return nil, errors.Errorf("empty subnet id found for zone %s", zone)
+				}
+			}
+		}
+		// Not legal to have no subnets
+		if len(subnets) < 0 {
```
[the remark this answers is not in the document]
lleshchi: Typo :)

pkg/controller/machinepool/machinepool_controller.go
2uasimojo: > No outer loop.

Okay, so those filters are different types internally (scalar vs pointer types in the leaves of the structs) so you will still have to have the outer loop to convert.

2uasimojo:
We're going to call this issue fixed via #2279. We can resurrect this solution in the future if it becomes necessary.

PR #2246, opened 2024-03-20
HIVE-2399: Reuse restConfigClientGetter when creating cmdutil.Factory

dlom:
xref: [HIVE-2399](https://issues.redhat.com//browse/HIVE-2399)

(copied and formatted explanation from the comments in the code)

[HIVE-2399](https://issues.redhat.com//browse/HIVE-2399): In the cluster sync controller, `getRESTConfigFactory` is called once per reconcile loop. Previously, a `restConfigClientGetter` was created for every reconcile, used to apply (or rather, attempt to apply) various synced objects, and then thrown away at the end of the reconcile pass.

The `restConfigClientGetter` (or more accurately, the  `cmdutil.Factory` that wraps it) is used by the cluster sync controller (via the resource `helper`) to apply the various objects to the target cluster during the syncing process. During the application process (see `apply.go`), the apply command is set up (aptly named `setupApplyCommand()`) using the factory returned by the `getRESTConfigFactory()` function. Within this setup, the `ToRESTMapper()` function of the `restConfigClientGetter` is invoked. While the mapper returned is a deferred variant, eventually the inner `NewDiscoveryRESTMapper()` function is invoked, which is **very expensive** and allocates a **_huge_** amount of memory over time if repeatedly called in a hot loop. A graphic demonstrating these allocations is attached the the [HIVE-2399](https://issues.redhat.com//browse/HIVE-2399) card.

The solution is twofold:
1) store the deferred mapper on the `restConfigClientGetter` so that it can be reused
2) cache the `restConfigClientGetter` for each rest config and reuse it for future reconciles

/assign @2uasimojo 
/hold

### DIFF
--- a/pkg/controller/clustersync/clustersync_controller.go
+++ b/pkg/controller/clustersync/clustersync_controller.go
@@ -183,13 +183,16 @@ func resourceHelperBuilderFunc(
 		return resource.NewFakeHelper(logger), nil
 	}
 
-	restConfig, err := remoteClusterAPIClientBuilderFunc(cd).RESTConfig()
+	restConfig, secretVersion, err := remoteClusterAPIClientBuilderFunc(cd).RESTConfigAndSecretVersion()
 	if err != nil {
 		logger.WithError(err).Error("unable to get REST config")
 		return nil, err
 	}
 
-	return resource.NewHelperFromRESTConfig(restConfig, logger)
+	if cd.Spec.ClusterMetadata == nil || len(cd.Spec.ClusterMetadata.ClusterID) == 0 {
+		return nil, fmt.Errorf("missing clusterID for cluster %s/%s", cd.Namespace, cd.Name)
+	}
+	return resource.NewHelper(logger, resource.WithRESTConfig(restConfig), resource.WithCacheKeyAndVersion(cd.Spec.ClusterMetadata.ClusterID, secretVersion))
 }
 
 // AddToManager adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -406,7 +409,32 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 
 	if cd.DeletionTimestamp != nil {
 		logger.Debug("cluster is being deleted")
+
+		if controllerutils.HasFinalizer(cd, hiveintv1alpha1.FinalizerCachePurge) {
+			if cd.Spec.ClusterMetadata == nil || len(cd.Spec.ClusterMetadata.ClusterID) == 0 {
+				logger.
+					WithField("cluster_name", cd.Name).
+					WithField("cluster_namespace", cd.Namespace).
+					Warn("missing clusterID for cluster, cache entry may dangle")
+			} else {
+				resource.RemoveRestConfigClientGetterCacheEntry(logger, cd.Spec.ClusterMetadata.ClusterID)
+			}
+
+			logger.Debugf("removing cache purge finalizer")
+			controllerutils.DeleteFinalizer(cd, hiveintv1alpha1.FinalizerCachePurge)
+			return reconcile.Result{}, r.Update(context.TODO(), cd)
+		}
+
 		return reconcile.Result{}, nil
+	} else {
+		if !controllerutils.HasFinalizer(cd, hiveintv1alpha1.FinalizerCachePurge) {
+			logger.Debug("adding cache purge finalizer")
+			controllerutils.AddFinalizer(cd, hiveintv1alpha1.FinalizerCachePurge)
+			if err = r.Update(context.TODO(), cd); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{Requeue: true}, nil
+		}
 	}
 
 	if !cd.Spec.Installed {
--- a/pkg/resource/restconfig_factory.go
+++ b/pkg/resource/restconfig_factory.go
@@ -2,13 +2,42 @@ package resource
 
 import (
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
+	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sync"
 )
+
+// HIVE-2399: In the cluster sync controller, getRESTConfigFactory is called once per
+// reconcile loop. Previously, a restConfigClientGetter was created for every reconcile,
+// used to apply (or rather, attempt to apply) various synced objects, and then thrown
+// away at the end of the reconcile pass.
+//
+// The restConfigClientGetter (or more accurately, the  cmdutil.Factory that wraps it) is
+// used by the cluster sync controller (via this helper) to apply the various objects to
+// the target cluster during the syncing process. During the application process (see apply.go),
+// the apply command is set up (aptly named setupApplyCommand()) using the factory returned
+// by the getRESTConfigFactory() below. Within this setup, the ToRESTMapper() function of the
+// restConfigClientGetter is invoked. While the mapper returned is a deferred variant (restmapper.DeferredDiscoveryRESTMapper),
+// eventually the inner NewDiscoveryRESTMapper() function is invoked, which is very expensive
+// and allocates a huge amount of memory over time if repeatedly called in a hot loop. A graphic
+// demonstrating these allocations is attached the HIVE-2399 card.
+//
+// The solution is twofold:
+// 1) store the deferred mapper on the restConfigClientGetter so that it can be reused
+// 2) cache the restConfigClientGetter for each rest config and reuse it for future reconciles
+// The type of this map is effectively map[string]*restConfigClientGetter
+var restConfigClientGetterCache sync.Map
+
+func RemoveRestConfigClientGetterCacheEntry(logger log.FieldLogger, key string) {
+	logger.WithField("getter-cache-key", key).Debug("removing client config cache entry")
+
+	restConfigClientGetterCache.Delete(key)
+}
 
 func (r *helper) getRESTConfigFactory(namespace string) (cmdutil.Factory, error) {
 	if r.metricsEnabled {
@@ -17,8 +46,47 @@ func (r *helper) getRESTConfigFactory(namespace string) (cmdutil.Factory, error)
 		controllerutils.AddControllerMetricsTransportWrapper(cfg, r.controllerName, false)
 		r.restConfig = cfg
 	}
-	r.logger.WithField("cache-dir", r.cacheDir).Debug("creating cmdutil.Factory from REST client config and cache directory")
-	f := cmdutil.NewFactory(&restConfigClientGetter{restConfig: r.restConfig, cacheDir: r.cacheDir, namespace: namespace})
+
+	// If we're not using the cache, this will be the getter used. If we are using the cache,
+	// this value will be stored if the cache misses.
+	freshRestClientGetter := &restConfigClientGetter{
+		restConfig: r.restConfig,
+		cacheDir:   r.cacheDir,
+		namespace:  namespace,
+		logger:     r.logger,
+		version:    r.restConfigClientGetterVersion,
+	}
+	restClientGetter := freshRestClientGetter
+	reused := false
+
+	// HIVE-2399: reuse the same restConfigClientGetter if a cache key is provided
+	if len(r.restConfigClientGetterKey) != 0 {
+		// retrieve a cached getter, or store one if none exists for the given cache key
+		cachedRestClientGetterAny, loaded := restConfigClientGetterCache.LoadOrStore(r.restConfigClientGetterKey, freshRestClientGetter)
+
+		// if we didn't store a new getter, we check to see if we can reuse it:
+		if loaded {
+			cachedRestClientGetter := cachedRestClientGetterAny.(*restConfigClientGetter)
+
+			if cachedRestClientGetter.version == r.restConfigClientGetterVersion {
+				// if the getter has the same version, we reuse it
+				restClientGetter = cachedRestClientGetter
+				reused = true
+			} else {
+				// otherwise, update it with the fresh one
+				restConfigClientGetterCache.Store(r.restConfigClientGetterKey, freshRestClientGetter)
+			}
+		}
+	}
+
+	r.logger.
+		WithField("cache-dir", r.cacheDir).
+		WithField("getter-cache-key", r.restConfigClientGetterKey).
+		WithField("reused-getter", reused).
+		WithField("getter-version", restClientGetter.version).
+		Debug("creating cmdutil.Factory from REST client config and cache directory")
+
+	f := cmdutil.NewFactory(restClientGetter)
 	return f, nil
 }
 
@@ -26,6 +94,13 @@ type restConfigClientGetter struct {
 	restConfig *rest.Config
 	cacheDir   string
 	namespace  string
+	logger     log.FieldLogger
+
+	// HIVE-2399: since we are reusing the restConfigClientGetter, it may be accessed concurrently
+	mu              sync.Mutex
+	mapper          *restmapper.DeferredDiscoveryRESTMapper
+	discoveryClient discovery.CachedDiscoveryInterface
+	version         string
 }
 
 // ToRESTConfig returns restconfig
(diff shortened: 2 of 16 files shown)


pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,71 @@ package resource
...
 		// Copy the possibly shared restConfig reference and add a metrics wrapper.
 		cfg := rest.CopyConfig(r.restConfig)
 		controllerutils.AddControllerMetricsTransportWrapper(cfg, r.controllerName, false)
 		r.restConfig = cfg
 	}
-	r.logger.WithField("cache-dir", r.cacheDir).Debug("creating cmdutil.Factory from REST client config and cache directory")
-	f := cmdutil.NewFactory(&restConfigClientGetter{restConfig: r.restConfig, cacheDir: r.cacheDir, namespace: namespace})
+
+	// HIVE-2399: reuse the same restConfigClientGetter for each restConfig
+	restClientGetter, loaded := restConfigClientGetterCache.LoadOrStore(r.restConfig, &restConfigClientGetter{
```
2uasimojo, line 44: I'm pretty interested/concerned about how a rest.Config is used as a map key. Especially since some of the members are funcs. I guess as long as the funcs are static (not inlines/lambdas) they'll always hash the same? Except at least one of the funcs (cfg.Dial) looks like it's a lambda...

My point is: how do we know we're actually reusing entries?

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,71 @@ package resource
...
+// by the getRESTConfigFactory() below. Within this setup, the ToRESTMapper() function of the
+// restConfigClientGetter is invoked. While the mapper returned is a deferred variant,
+// eventually the inner NewDiscoveryRESTMapper() function is invoked, which is very expensive
+// and allocates a huge amount of memory over time if repeatedly called in a hot loop. A graphic
+// demonstrating these allocations is attached the the HIVE-2399 card.
+//
+// The solution is twofold:
+// 1) store the deferred mapper on the restConfigClientGetter so that it can be reused
+// 2) cache the restConfigClientGetter for each rest config and reuse it for future reconciles
+var restConfigClientGetterCache sync.Map
```
2uasimojo, line 33: I'm not done reading yet, but unless we implement some kind of manual garbage collection, this cache will still be able to leak memory, as it will never remove an entry.

IIUC reconciling on a given spoke will require an entry, yes? But it could even be worse than that because if e.g. creds are rotated, that would result in a _different_ rest.Config for the same spoke?

And even if we managed to remove a "stale" entry, due to the nature of the underlying problem, it wouldn't mean we actually scrubbed the memory it consumed internally?

:arrow_up: partially-formed thoughts, discussion needed.

2uasimojo (commented):
Definitely going to need to Meet™ to talk this through. Agenda:
- What kind of field testing have you done to validate that cache entries are truly being reused?
- How are we going to deal with stale cache entries?

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,71 @@ package resource
...
+	restClientGetter, loaded := restConfigClientGetterCache.LoadOrStore(r.restConfig, &restConfigClientGetter{
+		restConfig: r.restConfig,
+		cacheDir:   r.cacheDir,
+		namespace:  namespace,
+		logger:     r.logger,
+	})
+
+	r.logger.
+		WithField("cache-dir", r.cacheDir).
+		WithField("reused-getter", !loaded).
```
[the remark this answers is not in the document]
dlom: Yes, I had renamed the field without updating the variable

pkg/resource/restconfig_factory.go
dlom: Long story short: it doesn't (and I want to discuss further)

2uasimojo:
Plan:
- Use something more deterministic for the cache key, like cluster ID. (Mark discovered that we weren't getting cache hits, probably because the rest.Config has pointers and stuff in it.)
- Figure out whether & how the above can result in a cache hit that returns an invalid rest.Config (e.g. if certs were rotated), and what that failure looks like. The idea being to decide how to react -- probably clear that cache entry and bubble up the error so we rebuild the rest.Config on the next iteration.
- Think of a way to clear stale cache entries for deleted clusters. Ideas we brainstormed:
  - Add a lastUsedTimestamp to cache entries. Have a garbage collector periodically sweep the cache looking for entries older than $threshold (probably something like 2x the reconcile interval).
  - Add a finalizer to CD or ClusterSync so our reconciler is guaranteed an opportunity to run when the CD is deleted. (Or plug into the existing clustersync controller-specific finalizer if there is one.) Have it clear that cluster's cache entry before removing the finalizer.
  - Periodically, like O(days), just dump the whole cache.

Orthogonal: We also talked about adding a feature: a (configurable, probably via hiveconfig=>controllersconfig) int threshold number of failures one [S]SS is allowed to have before we'll quit trying to sync it. Add fields to the clustersync entry for each [S]SS tracking its resourceVersion and the number of failures. Reset the count when (the sync succeeds or) the resourceVersion changes. Thus "I've fixed the problem, please try again" happens for free when the user edits the [S]SS. If the fix was somewhere other than in the [S]SS, you have to edit it anyway, e.g. adding a bogus annotation, to trigger us to try again. Or even delete the ClusterSync, if you want to use a big hammer. Mark to open a separate card for this.

pkg/resource/restconfig_factory.go
```diff
@@ -1,31 +1,138 @@
...
+var cacheCleanMu sync.Mutex
+var lastCacheClean time.Time
+var cacheCleanInterval = time.Hour * 1
+var cacheEntryMaxAge = time.Hour * 8
+
+func cleanRestConfigClientGetterCache(logger log.FieldLogger) {
+	cacheCleanMu.Lock()
+
+	now := time.Now()
+	if now.Before(lastCacheClean.Add(cacheCleanInterval)) {
```
2uasimojo, line 49: So you're using this to keep this func an O(1) no-op the majority of the time. It's nice in that it's simple, but it's a tad hacky.

An alternative -- more complex but arguably more correcter -- is to make the cleanRestConfigClientGetterCache() func run fully async in its own queue, like the nameserver scraper. The queueing/polling behavior for that guy is defined [here](https://github.com/openshift/hive/blob/84c194aeb89c027a731fc1653081b1ca323b7a9d/pkg/controller/dnsendpoint/nameserverscraper.go#L211); and that Start function is implicitly run by the controller manager by having been Add()ed to it like [this](https://github.com/openshift/hive/blob/84c194aeb89c027a731fc1653081b1ca323b7a9d/pkg/controller/dnsendpoint/dnsendpoint_controller.go#L134).

Definitely not insisting on changing this, but wanted to put it in your head.

pkg/resource/restconfig_factory.go
```diff
@@ -1,31 +1,138 @@
...
+// The solution is twofold:
+// 1) store the deferred mapper on the restConfigClientGetter so that it can be reused
+// 2) cache the restConfigClientGetter for each rest config and reuse it for future reconciles
+// The type of this map is effectively map[string]*restConfigClientGetter
+var restConfigClientGetterCache sync.Map
+
+var cacheCleanMu sync.Mutex
+var lastCacheClean time.Time
+var cacheCleanInterval = time.Hour * 1
+var cacheEntryMaxAge = time.Hour * 8
```
2uasimojo, line 43: I guess we'll see how expensive the cleaner is, but this seems more frequent than we really need.

We may want to consider making one or both of these tunable. Initially while we're PoCing we could do it via env vars funneled from the hive-operator Deployment to the controllers. Iff we find out we actually want to keep the tunability, HiveConfig.

pkg/resource/restconfig_factory.go
```diff
@@ -63,3 +181,25 @@ func (r *restConfigClientGetter) ToRawKubeConfigLoader() clientcmd.ClientConfig
...
+}
+
+// validateRestConfig returns nil iff the rest config is valid and working.
+func validateRestConfig(config *rest.Config) error {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	// perform a basic action (listing namespaces) to verify connection
+	_, err = clientset.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{})
```
2uasimojo, line 203: ee, this could be pretty expensive. If we are gonna keep a test query, I think we want it to be something cheaper, like maybe grabbing the `infrastructure` object (non-namespaced singleton named `cluster`).

But I don't love the idea of doing a test query on the spoke every time we pull a cache entry. We talked about letting the normal reconcile fail organically, detecting that that failure had to do with the cache entry being stale, and using that to invalidate it. Did you explore that option? One possible impl approach that comes to mind would involve wrapping the clustersync Reconcile func to do that discovery/invalidation before bubbling up the error.

Also, we talked about only invalidating the cache entry for _certain_ errors. You were going to figure out what it looks like when certs expire... which I _think_ is the only way this happens? Because e.g. for some temporary network glitch, as currently written we would end up thrashing, creating fresh getters, resulting in the original leak, right?

pkg/resource/restconfig_factory.go
```diff
@@ -1,31 +1,138 @@
...
 	return f, nil
 }
 
 type restConfigClientGetter struct {
 	restConfig *rest.Config
 	cacheDir   string
 	namespace  string
+	logger     log.FieldLogger
+
+	// HIVE-2399: since we are reusing the restConfigClientGetter, it may used concurrently
```
2uasimojo, line 131:
```suggestion
	// HIVE-2399: since we are reusing the restConfigClientGetter, it may be accessed concurrently
```

pkg/resource/restconfig_factory.go
```diff
@@ -41,16 +148,27 @@ func (r *restConfigClientGetter) ToDiscoveryClient() (discovery.CachedDiscoveryI
...
+		if err != nil {
+			r.mu.Unlock()
+			return nil, err
+		}
+		r.mapper = restmapper.NewDeferredDiscoveryRESTMapper(discoveryClient)
+		r.discoveryClient = discoveryClient
+	} else {
+		r.logger.Debugf("reusing discovery client and mapper for ToRESTMapper()")
 	}
+	r.mu.Unlock()
```
2uasimojo, line 165: Use `defer r.mu.Unlock()` after L152 instead. If you're really wanting to keep the mutex window tight, wrap L152-164 in an anonymous `func(){}()`.

dlom:
@2uasimojo Finalizer code is good, cache entries are properly removed when the clustersync object is destroyed

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,115 @@ package resource
...
+var restConfigClientGetterCache sync.Map
+
+var cacheCleanMu sync.Mutex
+var lastCacheClean time.Time
+var cacheCleanInterval = time.Hour * 4
+var cacheEntryMaxAge = time.Hour * 24
+
+func init() {
+	// TODO: override cache duration variables from env
+}
```
2uasimojo, line 44: YAGNI. We can put a TODO comment on the vars themselves, but no need to have an empty `init` hanging around before we need it.

[Later] Except those vars aren't even used anymore, right?

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,115 @@ package resource
...
+
+	restConfigClientGetterCache.Delete(key)
+}
+
 func (r *helper) getRESTConfigFactory(namespace string) (cmdutil.Factory, error) {
 	if r.metricsEnabled {
 		// Copy the possibly shared restConfig reference and add a metrics wrapper.
 		cfg := rest.CopyConfig(r.restConfig)
 		controllerutils.AddControllerMetricsTransportWrapper(cfg, r.controllerName, false)
 		r.restConfig = cfg
```
2uasimojo, line 57: Eh, is this creating extra copies of something?

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,115 @@ package resource
...
 	}
-	r.logger.WithField("cache-dir", r.cacheDir).Debug("creating cmdutil.Factory from REST client config and cache directory")
-	f := cmdutil.NewFactory(&restConfigClientGetter{restConfig: r.restConfig, cacheDir: r.cacheDir, namespace: namespace})
+
+	// If we're not using the cache, this will be the getter used. If we are using the cache,
+	// this value will be stored if the cache misses.
+	freshRestClientGetter := &restConfigClientGetter{
+		restConfig: r.restConfig,
+		cacheDir:   r.cacheDir,
```
2uasimojo, line 64: I'm curious what this cacheDir is for. Does its function overlap with the caching we're implementing here?

[Later] Looks like it might be for schema, which would be cool (and hopefully unrelated??)

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,115 @@ package resource
...
+// The solution is twofold:
+// 1) store the deferred mapper on the restConfigClientGetter so that it can be reused
+// 2) cache the restConfigClientGetter for each rest config and reuse it for future reconciles
+// The type of this map is effectively map[string]*restConfigClientGetter
+var restConfigClientGetterCache sync.Map
+
+var cacheCleanMu sync.Mutex
+var lastCacheClean time.Time
+var cacheCleanInterval = time.Hour * 4
+var cacheEntryMaxAge = time.Hour * 24
```
2uasimojo, line 40: looks like these are no longer used

pkg/resource/restconfig_factory.go
```diff
@@ -2,30 +2,115 @@ package resource
...
+		if loaded {
+			cachedRestClientGetter := cachedRestClientGetterAny.(*restConfigClientGetter)
+
+			if cachedRestClientGetter.version == r.restConfigClientGetterVersion {
+				// if the getter has the same version, we reuse it
+				restClientGetter = cachedRestClientGetter
+				reused = true
+			} else {
+				// otherwise, update it with the fresh one
+				restConfigClientGetterCache.Store(r.restConfigClientGetterKey, freshRestClientGetter)
```
2uasimojo, line 87: Do these two cache ops need to be wrapped in a mutex? Or... maybe they're under the mutex [below](https://github.com/openshift/hive/pull/2246/files#diff-f52f3cee6c548137a153544d536b82bf4e67d80ad1a10546830b4433752ff038R129-R130)? /me goes hunting for call stack...

2uasimojo (commented):
A few of these comments are open questions...

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -406,7 +410,28 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
...
+			logger.WithError(err).Log(controllerutils.LogLevel(err), "could not get ClusterSync")
+			return reconcile.Result{}, err
+		}
+
+		if clusterSync.DeletionTimestamp != nil {
+			logger.Debug("clustersync is also being deleted")
+			return reconcile.Result{}, r.cleanUpClusterSync(logger, clusterSync, cd)
+		} else {
+			logger.Debug("delete clustersync too")
+			if err := r.Delete(context.Background(), clusterSync); err != nil {
```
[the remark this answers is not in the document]
dlom: But if the clustersync is only deleted after the clusterdeployment is fully deleted, then the reconcile loop for the clustersync spins forever with "clusterdeployment not found"

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -454,6 +479,20 @@ func (r *ReconcileClusterSync) Reconcile(ctx context.Context, request reconcile.
 
 	clusterSync.Status.ControlledByReplica = &ordinal
 
+	if clusterSync.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
dlom: It also presents a way to force-clear the cache entry, though I suppose similar functionality already exists (when updating the admin secret)

pkg/controller/clustersync/clustersync_controller.go
```diff
@@ -1176,3 +1215,25 @@ func (r *ReconcileClusterSync) timeUntilRenew(lease *hiveintv1alpha1.ClusterSync
...
+		resource.RemoveRestConfigClientGetterCacheEntry(logger, cd.Spec.ClusterMetadata.ClusterID)
+
+		logger.Debugf("removing clustersync finalizer")
+		return r.removeClusterSyncFinalizer(cs)
+	}
+	return nil
+}
+
+func (r *ReconcileClusterSync) addClusterSyncFinalizer(cs *hiveintv1alpha1.ClusterSync) error {
+	cs = cs.DeepCopy()
```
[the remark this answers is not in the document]
dlom: Probably, yeah, I copied it straight from the clusterdeployment controller https://github.com/openshift/hive/blob/0d5507f91935701146f3615c990941f24bd42fe1/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1467-L1471

pkg/resource/restconfig_factory.go
dlom: It was there before, I just kept it there.  No idea what it does

pkg/controller/clustersync/clustersync_controller.go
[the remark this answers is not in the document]
2uasimojo: Have you seen this happen in practice, without forcing deletions? It _should_ be impossible, since the CD won't be garbage collected until its referents (including this one) are garbage collected.

pkg/controller/clustersync/clustersync_controller.go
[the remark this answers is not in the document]
2uasimojo: Right.

Up to you if you want to keep this chunk; but if you do, probably worth adding a comment summarizing the above so future-we know we thought about it.

pkg/controller/clustersync/clustersync_controller.go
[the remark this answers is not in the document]
2uasimojo: I'm gonna say let's get rid of it. A CD is not huge, but it's not nothing, and there doesn't seem to be a benefit to doing this.

pkg/controller/clustersync/clustersync_controller.go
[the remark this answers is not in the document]
dlom: Yes, it turns out that by default kubernetes will delete the CD first before the CS ("background" cascade) as opposed to what you think would be logical (the CS being deleted before the CD, "foreground" cascade).  Luckily we can just add a special finalizer to the CD as described here: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference.BlockOwnerDeletion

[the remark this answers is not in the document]
dlom: @2uasimojo ok so it's basically impossible to use this built in finalizer (and therefore ensure that the clustersync is cleaned up) unless you explicitely pass `--cascade=foreground` to `oc delete`, which is not something we want to require always.  Therefore, I think there are two options:
1. keep the codepath to manually delete the CD (as it stands currently)
2. store enough information on the clustersync object so that it doesn't require the CD to still exist in order to clean itself up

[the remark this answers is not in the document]
dlom: To be clear: Adding the finalizer manually to the CD, and then deleting the CD with `oc delete` will cause k8s to [_explicitly remove the finalizer for us_](https://github.com/kubernetes/kubernetes/blob/f30a87d517c538f22e1d84b4d7c497f57f7b760a/pkg/registry/core/namespace/storage/storage.go#L212) (again, because the default mode for deleting an object is `background`)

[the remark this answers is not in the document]
dlom: TL;DR: `blockOwnerDeletion` means nothing unless you supply the right option when deleting the owner

2uasimojo:
/lgtm

pending testing, unWIP/unhold when you're happy

dlom:
/retitle HIVE-2399: Reuse restConfigClientGetter when creating cmdutil.Factory

PR #2247, opened 2024-03-20
tmp: OCM-6713: AWS MachinePool UserTags

2uasimojo:
Temporary PR to see whether deepcopy mod in non-vendor dir makes this pass `verify`


2uasimojo:
This worked :shrug: 

TODO: Figure out why `verify` will pass without regenerating deepcopy, as long as you refrain from generating both of them.

PR #2259, opened 2024-04-16
Include linux/amd64 nodeSelector by default

2uasimojo:
Today we only build the hive image for linux/amd64. Attempting to run it in a multi-arch cluster without adding a nodeSelector can result in "exec format error" if any component (hive-operator, hive-controllers, hive-clustersync, hiveadmission) lands on e.g. an ARM node.

We provide a sample hive-operator Deployment which gets built into the operatorhub CSV and the OSD SaaS template, and may also be used as a starting point by other consumers. Absent this PR:
- For OLM-based installations, it can be provided via Subscription.config.nodeSelector.
- Bespoke consumers can add nodeSelector manually if desired.
- OSD would need to do... something else; but this is not currently a problem we've been asked to fix.

This PR adds a linux/amd64 nodeSelector to the sample hive-operator Deployment manifest, which will result in that nodeSelector being included in the OperatorHub bundle and the OSD SaaS deployment by default.

[HIVE-2487](https://issues.redhat.com//browse/HIVE-2487)

### DIFF
--- a/config/operator/operator_deployment.yaml
+++ b/config/operator/operator_deployment.yaml
@@ -26,6 +26,9 @@ spec:
         control-plane: hive-operator
         controller-tools.k8s.io: "1.0"
     spec:
+      nodeSelector:
+        kubernetes.io/arch: amd64
+        kubernetes.io/os: linux
       serviceAccountName: hive-operator
       volumes:
       - name: kubectl-cache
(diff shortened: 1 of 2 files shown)


2uasimojo:
/hold

Not sure we want to do this -- see discussion in card.

2uasimojo:
With HIVE-2496 / #2269, we have full support for feeding tolerations and nodeSelectors down from the operator to all relevant pods, giving the hive deployer the power to pin to linux/x86 explicitly. No need to do it by default.

PR #2261, opened 2024-04-17
HIVE-2531: Add podman build target

celebdor:
Even though they are related, podman is much more common than buildah nowadays. It makes sense to have a target for podman for rootless, daemonless builds

TODO: Add Jira reference


jstuever (approved):
/lgtm
/approve

PR #2269, opened 2024-04-24
Copy nodeSelector, tolerations into Jobs

2uasimojo:
We already copy `nodeSelector` and `tolerations` from hive-operator into the various controllers. But we weren't copying them into the Jobs for imageset, provision, and uninstall. This could result in their Pods e.g. being scheduled to Nodes that can't run them; or being totally unschedulable.

With this commit, we:
- Add the namespace and name of the running pod as env vars on hive-controllers.
- Use those env vars to retrieve the hive-controllers Pod spec.
- Copy the nodeSelector and tolerations from that Pod spec down into the (PodSpecs of the) Jobs for imageset, provision, and uninstall.

[HIVE-2496](https://issues.redhat.com//browse/HIVE-2496)

### DIFF
--- a/config/controllers/deployment.yaml
+++ b/config/controllers/deployment.yaml
@@ -47,6 +47,14 @@ spec:
         - name: tmp
           mountPath: /tmp
         env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
         - name: CLI_CACHE_DIR
           value: /var/cache/kubectl
         - name: HIVE_NS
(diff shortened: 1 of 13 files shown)


2uasimojo:
- `verify`: I forgot to regenerate bindata.
- `e2e*`: due to :arrow_up: the controller manifest doesn't have the new env vars, so it (correctly) [hits the new error path](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hive/2269/pull-ci-openshift-hive-master-e2e/1783271844216836096/artifacts/e2e/test/artifacts/hive-controllers.log) when starting.

Good red path test case ✓
Fixing...

2uasimojo:
So now we're failing because the cache hasn't started. Guess I need to add a retry loop? Starting to seem awful hacky...

2uasimojo:
> So now we're failing because the cache hasn't started. Guess I need to add a retry loop? Starting to seem awful hacky...

I'm instead doing the lookup in the reconcile loop, having made the new fields pointers so I can tell if I've already done that, so I can make sure I only do it once.

2uasimojo:
Tested this locally, works.
/assign @suhanime 
/cc @hongkailiu @bear-redhat

PR #2253, opened 2024-04-02
MachinePools/OpenStack: Fix nil pointer exception

2uasimojo:
...when MachinePool.Spec.Replicas is unset, as it is when we're doing autoscaling.

The fix was [upstream](https://github.com/openshift/installer/pull/8227) so this PR just revendors to pick it up, and adds unit tests to prove it works.

[HIVE-2476](https://issues.redhat.com//browse/HIVE-2476)

PR #2251, opened 2024-03-28
Revendor installer, repair OpenStack MachinePool UT

2uasimojo:
* Revendor installer to pick up https://github.com/openshift/installer/pull/8187 and https://github.com/openshift/installer/pull/8209
* Refactor OpenStack MachinePool code to use the new code path for trunk support discovery and accommodate the new signature of the `MachineSets()` generator
* Repair long-broken OpenStack MachinePool unit tests, which were previously blocked by the above
* Accommodate https://github.com/openshift/installer/pull/8006 which is also brought in by the revendor

A subsequent PR will use this to
* Expand unit tests to reproduce the error in the referenced card
* Fix it.

Part of [HIVE-2476](https://issues.redhat.com//browse/HIVE-2476)

PR #2274, opened 2024-04-30
MachinePools/OpenStack: Fix nil pointer exception

2uasimojo:
This is *not* a cherry-pick of #2253 / (c6b37ee6 & f59f327d). That solution revendored installer to pick up the fix from upstream. In older branches, this would have dragged in too many dependencies, so we instead fix it "locally" with an explicit nil check.

Note also that the original fix added unit tests. We can't do that here either because the new tests rely on the OpenStack UT suite being un-broken [1], which again relied on upstream changes [2][3] we can't pull into older branches.

[1] https://github.com/openshift/hive/pull/2251
[2] https://github.com/openshift/installer/pull/8187
[3] https://github.com/openshift/installer/pull/8209

[HIVE-2476](https://issues.redhat.com//browse/HIVE-2476)

### DIFF
--- a/pkg/controller/machinepool/openstackactuator.go
+++ b/pkg/controller/machinepool/openstackactuator.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


2uasimojo:
Thanks @suhanime!

/hold for testing

2uasimojo:
/remove-lifecycle stale

/hold cancel

/override "Red Hat Konflux / hive-mce-25-on-pull-request"

The segfault issue has been addressed; the autoscaling problem will be investigated separately (HIVE-2590).

2uasimojo:
/override ci/prow/security

#2387 

/override "Red Hat Konflux / hive-mce-25-on-pull-request"

PR #2276, opened 2024-05-07
Add alternate regex for TerraformFailedToDeleteResources

abyrne55:
Per OSD-22828, sometimes an install can fail if Terraform can't properly cleanup after itself (e.g., because the user attached the bootstrap node's security group to an unrelated network interface). While TerraformFailedToDeleteResources already covered this scenario, it seems that Terraform gives a slightly different error message in some cases, tripping up the existing regex. This PR adds an additional regex to TerraformFailedToDeleteResources that should recognize the alternative error message.


2uasimojo:
This is going to prod right now via https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/105713

PR #2278, opened 2024-05-14
MachinePools/AWS: Match by subnet ID only

2uasimojo:
The MachinePool controller compares generated MachineSets with existing remote MachineSets to figure out which to add, update, or delete.

We used to do this matching based on naming conventions, but those have been changing upstream, so we recently ([HIVE-2254](https://issues.redhat.com//browse/HIVE-2254) / #2179) changed to matching by label + failure domain. The latter is extracted using code from control-plane-machine-set.

Unfortunately, in AWS, the same failure domain can contain a subnet represented in three different ways:
- ID (canonical)
- ARN (canonical; not used)
- Filters (ambiguous)

If the generated and remote MachineSets differ in either the way the failure domain subnet is represented (e.g. ID vs Filters) *or* both are using Filters that differ despite resolving to the same subnet, our existing algorithm will fail to match them, and the reconcile will fail quietly (remote MachineSets are not updated, local MachinePool status is not changed).

With this commit, we disambiguate by ensuring both sides represent their failure domain subnets by ID. If the remote MachineSet uses filters, we update it (MAPI won't act on that change, but it should be functionally a no-op) so we won't have to look it up again next time. *TODO:* We should also figure out a way to cache the subnet IDs for the MachinePool for the same reason -- perhaps by updating the MachinePool itself.

### DIFF
--- a/pkg/controller/machinepool/awsactuator.go
+++ b/pkg/controller/machinepool/awsactuator.go
@@ -18,6 +18,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	configv1 "github.com/openshift/api/config/v1"
+	v1 "github.com/openshift/api/machine/v1"
 	machineapi "github.com/openshift/api/machine/v1beta1"
 	icaws "github.com/openshift/installer/pkg/asset/installconfig/aws"
 	installaws "github.com/openshift/installer/pkg/asset/machines/aws"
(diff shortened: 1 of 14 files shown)


pkg/controller/machinepool/awsactuator.go
```diff
@@ -46,7 +49,62 @@ var (
...
+		if err != nil {
+			return nil, err
+		}
+
+		if aws.StringValue(fd1.AWS().Subnet.ID) == "" {
+			subnets, err := getSubnetIDByFilter(fd1.AWS().Subnet, *a)
+			if err != nil {
+				return nil, err
+			}
+			subnets, err = a.filterPublicSubnets(subnets, pool)
```
2uasimojo, line 90: This seems like it will break in validateSubnets, which makes sure we have exactly one public subnet per AZ in the pool, because here we're passing in just the one subnet we gleaned from the one remote MachineSet.

2uasimojo:
This is just a rebase of #2245, which gets rid of the bottom commit that already merged.

But I'm going to close it in favor of #2279, and maybe Leah will come back to 2245 later.

PR #2279, opened 2024-05-14
MachinePool/AWS: Match MachineSets by AZ only (not subnet)

2uasimojo:
The MachinePool controller compares generated MachineSets with existing remote MachineSets to figure out which to add, update, or delete.

We used to do this matching based on naming conventions, but those have been changing upstream, so we recently ([HIVE-2254](https://issues.redhat.com//browse/HIVE-2254) / #2179) changed to matching by label + failure domain. The latter is extracted using code from control-plane-machine-set.

Unfortunately, in AWS, the same failure domain can contain a subnet represented in three different ways:

- ID (canonical)
- ARN (canonical; not used)
- Filters (ambiguous)

If the generated and remote MachineSets differ in either the way the failure domain subnet is represented (e.g. ID vs Filters) or both are using Filters that differ despite resolving to the same subnet, our existing algorithm will fail to match them, and the reconcile will fail quietly (remote MachineSets are not updated, local MachinePool status is not changed).

Prior to [HIVE-2254](https://issues.redhat.com//browse/HIVE-2254), the name matching relied only on the availability zone (e.g. `us-east-1a`). So with this commit, we special-case the matching for AWS to ignore the subnet piece of the failure domain and just match by that AZ.

(A future effort may reinstate matching by subnet in an unambiguous way by always resolving both sides to IDs.)

This commit also resolves a latent bug whereby we were replacing by-name subnet filters with just the legacy terraform name (`${clusterID}-private-${az}`), which would have broken in newer CAPI-based installs. We now include both the legacy name and `${clusterID}-subnet-private-${az}` which is used by CAPI-based installs.

### DIFF
--- a/apis/hive/v1/aws/machinepool.go
+++ b/apis/hive/v1/aws/machinepool.go
@@ -13,7 +13,8 @@ type MachinePoolPlatform struct {
 	// private and one public subnet for each availability zone. In this case, the public
 	// subnets will be filtered out and only the private subnets will be used.
 	// If empty/omitted, we will look for subnets in each availability zone tagged with
-	// Name=<clusterID>-private-<az>.
+	// Name=<clusterID>-private-<az> (legacy terraform) or <clusterID>-subnet-private-<az>
+	// (CAPA).
 	Subnets []string `json:"subnets,omitempty"`
 
 	// InstanceType defines the ec2 instance type.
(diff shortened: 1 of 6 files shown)


2uasimojo:
/test e2e

rbac timing flake (I think) that I've seen before

2uasimojo:
/cherry-pick mce-2.6 mce-2.5 mce-2.4 mce-2.3 mce-2.2

Let's try out this new cherry-pick feature :)

vishnuchalla:
Thanks for this fix.

PR #2284, opened 2024-05-17
Test konflux regex

2uasimojo:


2uasimojo:
Oh. Well of course that's not going to work: the regex doesn't exclude `.tekton` :eyeroll:

PR #2292, opened 2024-05-22
ACM-11805: Use Entitled builds for Konflux

celebdor:
As part of Konflux enablement, we should be using containers from registry.redhat.com.


Containerfile
```diff
@@ -0,0 +1,46 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.21 as builder
+RUN mkdir -p /go/src/github.com/openshift/hive
+WORKDIR /go/src/github.com/openshift/hive
+COPY . .
+RUN make build
```
celebdor, line 5: python GitPython is missing in the image.

celebdor:
The build passed locally. I think the konflux failure is a flake

celebdor:
/hold

Konflux cluster config changes needed

PR #2299, opened 2024-06-03
Nil check for InstalledTimeStamp missing

suhanime:
Within the hibernateAfter code path, if the check for InstalledTimeStamp takes place before the clusterDeployment controller reconcile has had a chance to update it (for ex, for adopted clusters - aka the clusterdeployments created with installed field set to true but the InstalledTimeStamp nil, we set it to creationTimeStamp), the code would panic. This commit adds a nil check for the same to skip computing the installedSince. The logic for calculating if Hibernation should happen tried to find the latest one amongst installedSince, claimedSince, runningSince, and should work fine with the other values present.

xref: https://issues.redhat.com/browse/HIVE-2486


2uasimojo (commented):
/lgtm

PR #2309, opened 2024-06-14
Plumb HiveConfig logLevel into controller manager

2uasimojo:
[HIVE-2535](https://issues.redhat.com//browse/HIVE-2535)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -154,11 +154,13 @@ func newRootCommand() *cobra.Command {
 			defer cancel()
 
 			run := func(ctx context.Context) {
+				mgrLogger := log.StandardLogger()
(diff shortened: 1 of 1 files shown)


2uasimojo:
This doesn't really seem to have done anything.

2uasimojo:
> This doesn't really seem to have done anything.

I think it may be because the [logger itself](https://github.com/openshift/hive/blob/df3db863884a7b366f478eea47b29b8b434597ad/pkg/util/logrus/logr.go#L16) only does Info and Error. And that's our util wrapper -- which I think is there to convert upstream Info to Debug -- but the upstream thing we're wrapping _also_ only has Info and Error. Not sure where to go from here :shrug:

PR #2310, opened 2024-06-14
Make MachinePool poll interval tunable

2uasimojo:
In addition to `Watch()`ing the local MachinePool and ClusterDeployment objects, the machinepool controller also sets up a poll. Previously the interval was hardcoded to 30m. With this commit, we retain that as the default, but allow it to be tuned via a new field: `HiveConfig.Spec.MachinePoolPollInterval`.

This is a Duration string -- e.g. `"30m"`. If set to zero or less (e.g. `"0m"`) we won't poll at all.

[HIVE-2536](https://issues.redhat.com//browse/HIVE-2536)

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -56,6 +56,7 @@ const (
 	machinePoolNameLabel       = "hive.openshift.io/machine-pool"
 	finalizer                  = "hive.openshift.io/remotemachineset"
 	masterMachineLabelSelector = "machine.openshift.io/cluster-api-machine-type=master"
+	defaultPollInterval        = 30 * time.Minute
 )
 
 var (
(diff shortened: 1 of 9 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -128,10 +129,23 @@ func Add(mgr manager.Manager) error {
...
+		pollInterval, err = time.ParseDuration(envPollInterval)
+		if err != nil {
+			log.WithError(err).WithField("reapplyInterval", envPollInterval).Errorf("unable to parse %s", constants.SyncSetReapplyIntervalEnvVar)
+			return err
+		}
+	}
+
+	// Periodically watch MachinePools for syncing status from external clusters -- but allow this to be
+	// disabled via HiveConfig.Spec.MachinePoolPollInterval <= 0
+	if pollInterval >= 0 {
```
2uasimojo, line 144: nope

2uasimojo (commented):
/hold

2uasimojo:
Actually

/hold

I just want to make sure we're not getting a default poll from c-r, will let this run for a couple hours with our poll disabled.

But still reviewable.

2uasimojo:
/hold cancel
Test checks out.

/override ci/prow/verify

Unrelated work is changing build containers, and it looks like we moved to one without python.

@joelanford suggested some tweaks, which I'll address separately.

@suhanime ready to go!

2uasimojo:
> @2uasimojo: 2uasimojo unauthorized: /override is restricted to Repo administrators, approvers in top level OWNERS file, and the following github teams:openshift: openshift-release-oversight openshift-staff-engineers.

[approvers in top level OWNERS file](https://github.com/openshift/hive/blob/master/OWNERS#L2), eh?

/override ci/prow/verify

2uasimojo:
TODO: Make hive-operator reroll hive-controllers when this changes in HiveConfig!

PR #2312, opened 2024-06-14
Redeploy controllers when machinePoolPollInterval changes

2uasimojo:
The knob added via #2310, HiveConfig.Spec.MachinePoolPollInterval, wasn't automatically redeploying hive-controllers when it changed. Now it will.

[HIVE-2536](https://issues.redhat.com//browse/HIVE-2536)


2uasimojo:
/hold

This doesn't seem to be needed, will confirm why and then close.

2uasimojo:
Yeah, we don't need this because the hiveconfig change results in a diff in the hive-controllers Deployment spec (specifically the container env), so it'll get reapplied automatically. It's only for config options that get passed down outside of the Deployment that we need to do special things to force redeploy.

PR #2318, opened 2024-06-19
HIVE-2400: fix obsolete python references

celebdor:
New builder images do not alias python to python3


celebdor:
Both this and openshift/release#53396 need to go in (other changes might be needed to. Working on unblocking the branch

PR #2320, opened 2024-06-20
HIVE-2400: Run installmanager binary in install container for fips compatibility

celebdor:
For FIPS compliance, Hiveutil needs to be able to run installer code against the same RHEL (8 or 9) against which it was built.


2uasimojo:
/lgtm

🤞

PR #2260, opened 2024-04-17
HIVE-2548: Run installmanager binary in install container for fips compatibility

lleshchi:
[HIVE-2400](https://issues.redhat.com/browse/HIVE-2400)
[HIVE-2548](https://issues.redhat.com//browse/HIVE-2548)

PR #2322, opened 2024-06-21
Fix AWS STS for RHEL8/9 transition

2uasimojo:
In #2260 we changed how we're invoking the openshift-install binary.

Before: Copy openshift-install into the hive container and run it via `/usr/bin/hiveutil install-manager`

After: Copy hiveutil into the installer container and run installer via `/output/hiveutil.rhel$VER install-manager`

What we missed was that, for STS flows, we inject an AWS credentials file containing a `credential_process` that invoked `/usr/bin/hiveutil install-manager aws-credentials` -- but `hiveutil` no longer lives there.

Fix.

[HIVE-2400](https://issues.redhat.com//browse/HIVE-2400)


2uasimojo:
/test periodic-images

Mysterious error message, may or may not be related to recent dockerfile changes.

PR #2323, opened 2024-06-21
HIVE-2400: credential_process run right hiveutil

celebdor:
Keeping how we copy, let's just do a symlink so that it finds the right executable


celebdor:
Duplicates #2322 
/close

PR #2342, opened 2024-07-04
HIVE-2558: Drop codegen k8s.io/code-generator

celebdor:
generate-groups.sh was deprecated in k8s.io/code-generator 1.29 and deleted in 1.30. Instead, the usage was replaced with:

- kube_codegen.sh library's kube::codegen::gen_helpers

The issue with the above is:

- It doesn't allow to specify just running deepcopy and instead adds as well code conversion and defaulter code generation.
- While deepcopy-gen and conversion-gen can deal well with the multiple input paths that we have, conversion-gen wants to work at the package level, which means it ends up wrongly processing the vendored k8s deepcopy directives. This in turn means that we are forced to use the kubernetes header files to respect their copyright, but ends up adding the kubernetes copyright to openshift generated code.
- gen_helpers enforces usage of --output-base matching the import name, which means that we had to workaround with a symlink of ${REPO_ROOT}/github.com/openshift/ to "${REPO_ROOT}..".
- vendored generated deepcopy got deleted and `make vendor` had to pick up the tab.

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -1,36 +1,29 @@
-#!/bin/bash -xe
+#!/bin/bash
 
-SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
-CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../../../k8s.io/code-generator)}
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
 
-# use gsed for MAC env
-SED_CMD=sed
-if [[ `uname` == 'Darwin' ]]; then
-  SED_CMD=gsed
-fi
+REPO_ROOT=$(realpath $(dirname ${BASH_SOURCE[0]})/..)
 
-# HACK: For some reason this script is not executable.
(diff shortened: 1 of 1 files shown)


hack/update-codegen.sh
```diff
@@ -1,36 +1,29 @@
...
-if [[ `uname` == 'Darwin' ]]; then
-  SED_CMD=gsed
-fi
+REPO_ROOT=$(realpath $(dirname ${BASH_SOURCE[0]})/..)
 
-# HACK: For some reason this script is not executable.
-${SED_CMD} -i 's,^exec \(.*/generate-internal-groups.sh\),bash \1,g' ${CODEGEN_PKG}/generate-groups.sh
-# ...but we have to put it back, or `verify` will puke.
-trap "git checkout ${CODEGEN_PKG}/generate-groups.sh" EXIT
+GO111MODULE=on go install k8s.io/code-generator/cmd/deepcopy-gen@release-1.29
```
2uasimojo, line 10: I ran into the same failure with `go install <web URI>` in https://github.com/openshift/hive/pull/2481

Can we try:
- make sure we have code-generator vendored at the appropriate version in go.mod, and naked-imported in pkg/dependencymagnet/go.doc
- Use `go install $(shell realpath vendor/k8s.io/code-generator/cmd/deepcopy-gen` here to build the tool from the vendor dir

?

2uasimojo:
/remove-lifecycle stale

@celebdor lmk if you want me to pick this up.

2uasimojo:
Looking at the current tip of `master`, this seems to have already been done elsewhere.

PR #2348, opened 2024-07-08
HIVE-2571: fixing build not to end up erroring out on too many open files

celebdor:
This prevents the hiveutil go compilation from ending up in too many open files error in both podman, buildah and Konflux.


2uasimojo:
/override ci/prow/security

Addressed via #2364 

/lgtm

(Guessing we'll have to override konflux too, but will wait for it to fail.)

PR #2352, opened 2024-07-10
fix(build): update to the fixed clamav image

celebdor:
The upcoming OpenShift Pipelines which will be deployed shortly has stricter validations on Pipelines and Task manifests. Clamav would fail the new validations. This commit should make us ready.

Fixes: #[HIVE-2563](https://issues.redhat.com//browse/HIVE-2563)


2uasimojo:
Oh, I guess we also need this:

/approve

PR #2354, opened 2024-07-15
fix(controller-runtime): fix metricsprovider wiring

celebdor:
Upstream controller-runtime does not provide a way to pass a metricsprovider when creating a Controller. This results in metrics/workqueue not being exported.


2uasimojo:
Yes, this gels with what we discussed :+1:

2uasimojo:
We expect `/verify` to fail since this is a direct edit of vendored code.

2uasimojo:
/payload-job-with-prs periodic-ci-openshift-hive-master-periodic-e2e-gcp-weekly openshift/api#1958 openshift/installer#8723

2uasimojo:
We went with a different temporary fix (#2355). This one may still help inform what we want the upstream fix to look like.

PR #2356, opened 2024-07-17
Forbid kubeconfig exec

2uasimojo:
Users with write access to the admin kubeconfig Secret for a given ClusterDeployment should not be able to execute arbitrary code in the privileged environment in which we run the controllers that use those Secrets. Funnel all code paths that load such Secrets through a validator to ensure that the AuthInfos[].Exec path is not used.

[HIVE-2485](https://issues.redhat.com//browse/HIVE-2485)

### DIFF
--- a/pkg/controller/argocdregister/argocdregister_controller.go
+++ b/pkg/controller/argocdregister/argocdregister_controller.go
@@ -211,7 +211,7 @@ func (r *ArgoCDRegisterController) reconcileCluster(cd *hivev1.ClusterDeployment
 		return reconcile.Result{}, nil
 	}
 
-	kubeConfigSecretName := cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name
+	kubeConfigSecretName := cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name // HIVE-2485 ✓
 	argoCDServerConfigBytes, err := r.generateArgoCDServerConfig(kubeConfigSecretName, cd.Namespace, argoCDNamespace, cdLog)
 	if err != nil {
 		return reconcile.Result{}, err
@@ -320,7 +320,7 @@ func (r *ArgoCDRegisterController) loadArgoCDServiceAccountToken(argoCDNamespace
 }
 
 func (r *ArgoCDRegisterController) generateArgoCDServerConfig(kubeconfigSecretName, kubeConfigSecretNamespace, argoCDNamespace string, cdLog log.FieldLogger) ([]byte, error) {
-	kubeconfig, err := r.loadSecretData(kubeconfigSecretName, kubeConfigSecretNamespace, adminKubeConfigKey)
+	kubeconfig, err := controllerutils.LoadSecretData(r.Client, kubeconfigSecretName, kubeConfigSecretNamespace, adminKubeConfigKey)
 	if err != nil {
 		cdLog.WithError(err).Error("unable to load cluster admin kubeconfig")
 		return nil, err
@@ -373,19 +373,6 @@ func tlsClientConfigBuilderFunc(kubeConfig clientcmd.ClientConfig, cdLog log.Fie
 	}
 
 	return tlsClientConfig, nil
-}
-
-func (r *ArgoCDRegisterController) loadSecretData(secretName, namespace, dataKey string) (string, error) {
-	s := &corev1.Secret{}
-	err := r.Get(context.TODO(), types.NamespacedName{Name: secretName, Namespace: namespace}, s)
-	if err != nil {
-		return "", err
-	}
-	retStr, ok := s.Data[dataKey]
-	if !ok {
-		return "", fmt.Errorf("secret %s did not contain key %s", secretName, dataKey)
-	}
-	return string(retStr), nil
 }
 
 // getPredictableSecretName generates a unique secret name by hashing the server API URL,
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -25,8 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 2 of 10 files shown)


pkg/controller/argocdregister/argocdregister_controller.go
```diff
@@ -211,7 +211,7 @@ func (r *ArgoCDRegisterController) reconcileCluster(cd *hivev1.ClusterDeployment
 		return reconcile.Result{}, nil
 	}
 
-	kubeConfigSecretName := cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name
+	kubeConfigSecretName := cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name // HIVE-2485 ✓
```
suhanime, line 214: Is this comment intentional?

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -301,7 +299,13 @@ func (r *ReconcileAWSPrivateLink) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, nil
 	}
 
-	return r.reconcilePrivateLink(cd, &hivev1.ClusterMetadata{InfraID: *cp.Spec.InfraID, AdminKubeconfigSecretRef: *cp.Spec.AdminKubeconfigSecretRef}, logger)
+	return r.reconcilePrivateLink(
+		cd,
+		&hivev1.ClusterMetadata{
+			InfraID:                  *cp.Spec.InfraID,
+			AdminKubeconfigSecretRef: *cp.Spec.AdminKubeconfigSecretRef, // HIVE-2585: via ClusterMetadata
```
suhanime, line 306: I don't have perms to view this issue but just wanted to ensure this comment is intentional

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -533,7 +537,7 @@ func (r *ReconcileAWSPrivateLink) reconcilePrivateLink(cd *hivev1.ClusterDeploym
 
 	// Figure out the API address for cluster.
 	apiDomain, err := initialURL(r.Client,
-		client.ObjectKey{Namespace: cd.Namespace, Name: clusterMetadata.AdminKubeconfigSecretRef.Name})
+		client.ObjectKey{Namespace: cd.Namespace, Name: clusterMetadata.AdminKubeconfigSecretRef.Name}) // HIVE-2485 ✓
```
suhanime, line 540: This too

pkg/controller/utils/secrets.go
```diff
@@ -2,23 +2,76 @@ package utils
...
+		}
+	}
+	return string(retBytes), nil
+}
+
+// RestConfigFromSecret accepts a Secret containing `kubeconfig` and/or `raw-kubeconfig` keys
+// and returns a rest.Config loaded therefrom.
+// If tryRaw is true, we will look for `raw-kubeconfig` first and use it if present, falling
+// back to `kubeconfig` otherwise.
+// The error return is non-nil if:
```
suhanime, line 41:
```suggestion
// The error returned is non-nil if:
```

suhanime (commented):
/lgtm

@2uasimojo Approving this right now because it is an important/urgent change, but there are many commented entries from your code-walkthrough for these changes, which I do not think are necessary to be in the code base. Can we remove them in a fup?
I also wanted to bring up the fact that we're not validating the kubeconfig when we're adopting it, which I think we should.
Also, do we need to document this anywhere?

2uasimojo:

Thanks @suhanime, I’ll respond to your comments shortly. TL:DR all intentional, but perhaps not necessary.

pkg/controller/awsprivatelink/awsprivatelink_controller.go
2uasimojo: > I don't have perms to view this issue

You do now. But that does put rather an ick on the posterity argument. Then again, hopefully someone whacking moles in this area later would be given access :shrug:

pkg/controller/utils/secrets.go
2uasimojo: This was intentionally a noun ("the error return variable"?) but I can change it in a fup if it bugs ya :)

2uasimojo:
> I also wanted to bring up the fact that we're not validating the kubeconfig when we're adopting it, which I think we should.

Yes, good eye. I did come across those code paths. I deliberately left them alone because they won't do damage unless they're _used_, and we should be covering all those cases. (It's also not super clear what the error path would look like in that case. Webhook validation can't be relied upon, so while we could write such to reject CD creation, it wouldn't be an effective barrier. So then we would have to "allow" the adoption but... put a status condition on the CD? Etc.)

> Also, do we need to document this anywhere?

I don't think so. The normal use case is for our provisioner to lay down the Secret, and then for hive to use it. The user should really never need to modify it (or even look at it) unless they're rotating KAS certs on the spoke -- and then they're just updating the cert fields.

pkg/controller/awsprivatelink/awsprivatelink_controller.go
2uasimojo: D'oh, the comment has a typo here.
```suggestion
			AdminKubeconfigSecretRef: *cp.Spec.AdminKubeconfigSecretRef, // HIVE-2485: via ClusterMetadata
```

pkg/controller/argocdregister/argocdregister_controller.go
suhanime: You don't have to scrub them per-se, but maybe label them something other than a ticket that's inaccessible to many? Or - how about we include a note where AdminKubeconfigSecretRef is defined, and all someone has to do is `Find usages` on that constant to see where we need to change/debug? And if that constant is just fetched and not used, we can include a note there as well. Essentially decouple the ticket and its specific use-case from here's-where-we-actually-use-kubeconfig-to-connect-to-the-cluster.

pkg/controller/awsprivatelink/awsprivatelink_controller.go
suhanime: I didn't catch it earlier but yeah, this is a typo - 2485 not 2585

pkg/controller/utils/secrets.go
suhanime: Eh - either makes sense to me so ignore

suhanime:
> It's also not super clear what the error path would look like in that case. Webhook validation can't be relied upon, so while we could write such to reject CD creation, it wouldn't be an effective barrier. So then we would have to "allow" the adoption but... put a status condition on the CD? Etc.

I think we shouldn't create a CD and throw an error. We should let them know this kubeconfig isn't going to work early on. Or - we can create a CD but atleast throw a warning in the log. Setting AuthenticationFailure condition seems too premature at the adoption stage.

2uasimojo:
> > It's also not super clear what the error path would look like in that case. Webhook validation can't be relied upon, so while we could write such to reject CD creation, it wouldn't be an effective barrier. So then we would have to "allow" the adoption but... put a status condition on the CD? Etc.
> 
> I think we shouldn't create a CD and throw an error. We should let them know this kubeconfig isn't going to work early on. Or - we can create a CD but atleast throw a warning in the log. Setting AuthenticationFailure condition seems too premature at the adoption stage.

> I think we shouldn't create a CD

Wait, are we talking about adoption through hiveutil? Yeah, I'm totally not going to sweat that code path. I thought you were talking about adding logic to adoption-specific paths in the controller.

Remember hiveutil is positioned as an internal dev tool, not for public use. If someone is going to break that seal, ultimately they're just using it as a convenience to generate yamls so they don't have to write them by hand... but either way ultimately they're creating CRs, which we can't stop them from doing (webhooks notwithstanding).

PR #2360, opened 2024-07-18
HIVE-2565: hiveutil awsprivatelink to be compatbile with CAPI based installer

jstuever:
In the terraform based installer, the worker security group was named INFRA_ID-worker-sg. However, when the installer transitioned to the CAPI based installer, the worker security group name became INFRA_ID-node. As a result, the hiveutil awsprivatelink endpoint command failed to find the security group.

This change adds the new name in addition to the old one so that both formats will be found when searching for the worker security group.

### DIFF
--- a/contrib/pkg/utils/aws/aws.go
+++ b/contrib/pkg/utils/aws/aws.go
@@ -182,7 +182,7 @@ func GetWorkerSGFromVpcId(awsClients awsclient.Client, vpcId *string) (string, e
 		Filters: []*ec2.Filter{
 			{
 				Name:   aws.String("tag:Name"),
-				Values: []*string{aws.String(infraID + "-worker-sg")},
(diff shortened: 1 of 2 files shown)


2uasimojo:
Thanks @jstuever!

Should we update [this doc](https://github.com/openshift/hive/blob/f4b08f4faa76ae512862d797af825d9aaffdbb37/docs/awsprivatelink.md?plain=1#L412) as well?

jstuever:
> Should we update [this doc](https://github.com/openshift/hive/blob/f4b08f4faa76ae512862d797af825d9aaffdbb37/docs/awsprivatelink.md?plain=1#L412) as well?

Good catch. [Done](https://github.com/openshift/hive/compare/f4b08f4faa76ae512862d797af825d9aaffdbb37..ff28487bea5d1e0974a1172021d237b6993b6eb3)

2uasimojo (commented):
/lgtm

/override ci/prow/security

PR #2364, opened 2024-07-19
fix(snyk): Ignore go-client misreporting

celebdor:
Due to the chaotic go-client history with tags, snyk misidentifies our go-client version as not meeting the minimum version containing a couple of fixes. This commit introduces entries in .snyk to ignore them.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -6,3 +6,10 @@ exclude:
     - vendor/**
     - apis/vendor/**
     - "**/*_test.go"
+ignore:
+    'SNYK-GOLANG-K8SIOCLIENTGOUTILJSONPATH-7540854':
+     - '* > k8s.io/client-go/util/jsonpath':
+        reason: 'Snyk mistakenly identifies v1.20.0-alpha.1 as newer than v0.29.7. client-go has a complicated history of versioning. Presumably, v1.20.0-alpha.1 referred to kubernetes-v1.20. kubernetes tags have since been prefaced by "kubernetes", whereas actual client-go versions resumed numbering from 0.x'
+    'SNYK-GOLANG-K8SIOCLIENTGOTRANSPORT-7538822':
+     - '* > k8s.io/client-go/transport':
+        reason: 'Snyk mistakenly identifies v1.17.0-alpha.1 as newer than v0.29.7. client-go has a complicated history of versioning. Presumably, v1.20.0-alpha.1 referred to kubernetes-v1.20. kubernetes tags have since been prefaced by "kubernetes", whereas actual client-go versions resumed numbering from 0.x'


.snyk
```diff
@@ -6,3 +6,10 @@ exclude:
     - vendor/**
     - apis/vendor/**
     - "**/*_test.go"
+ignore:
+    'SNYK-GOLANG-K8SIOCLIENTGOUTILJSONPATH-7540854':
+     - '* > k8s.io/client-go/util/jsonpath':
+        reason: 'Snyk mistakent identifies v1.20.0-alpha.1 as newer than v0.29.7. client-go has a complicated history of versioning. Presumably, v1.20.0-alpha.1 referred to kubernetes-v1.20. kubernetes tags have since been prefaced by "kubernetes", whereas actual client-go versions resumed numbering from 0.x'
```
2uasimojo, line 12:
```suggestion
        reason: 'Snyk mistakenly identifies v1.20.0-alpha.1 as newer than v0.29.7. client-go has a complicated history of versioning. Presumably, v1.20.0-alpha.1 referred to kubernetes-v1.20. kubernetes tags have since been prefaced by "kubernetes", whereas actual client-go versions resumed numbering from 0.x'
```

.snyk
```diff
@@ -6,3 +6,10 @@ exclude:
...
+    'SNYK-GOLANG-K8SIOCLIENTGOTRANSPORT-7538822':
+     - '* > k8s.io/client-go/transport':
+        reason: 'Snyk mistakent identifies v1.17.0-alpha.1 as newer than v0.29.7. client-go has a complicated history of versioning. Presumably, v1.20.0-alpha.1 referred to kubernetes-v1.20. kubernetes tags have since been prefaced by "kubernetes", whereas actual client-go versions resumed numbering from 0.x'
```
2uasimojo, line 15:
```suggestion
        reason: 'Snyk mistakenly identifies v1.17.0-alpha.1 as newer than v0.29.7. client-go has a complicated history of versioning. Presumably, v1.20.0-alpha.1 referred to kubernetes-v1.20. kubernetes tags have since been prefaced by "kubernetes", whereas actual client-go versions resumed numbering from 0.x'
```

2uasimojo (commented):
I won't hold this up on the typos, but I do have a question:

So do these vulns actually exist? Are they addressed yet in an actually-newer version of client-go? If not, how will we know when they are, if we're `ignore`ing these vulns?

celebdor:
> So do these vulns actually exist? Are they addressed yet in an actually-newer version of client-go? If not, how will we know when they are, if we're ignoreing these vulns?

As I tried to convey in the ignore messages, snyk indicates that they were fixed in v1.17.0-alpha.1 and v1.20.0-alpha.1 . These tags are from 2019. It's in the past. The problem is that these tags presumably refer to https://github.com/kubernetes/client-go/tree/kubernetes-1.17.0-alpha.1 and https://github.com/kubernetes/client-go/tree/kubernetes-1.20.0-alpha.1 which are waaaaaay older than v0.29.2, but snyk doesn't know that

2uasimojo:
Right right, but the vulns are still being flagged, which presumably means the vulns do exist; it’s just that snyk is giving bad info about where the fix can be found.

Or is it that these are super old vulns that were fixed long ago, and snyk thinks they are present in the current version because its semver is “smaller” than the versions where the fixes were introduced?

celebdor:
> Or is it that these are super old vulns that were fixed long ago, and snyk thinks they are present in the current version because its semver is “smaller” than the versions where the fixes were introduced?

The latter is my interpretation

2uasimojo:
Ack, with offline discussions. Let's get this cleared out.

/lgtm

/override "Red Hat Konflux"

2uasimojo:
I'm going to mash this through as it already passed the only test it could affect (`ci/prow/security`) and I'd like to clear the queue.

PR #2366, opened 2024-07-19
AWS: Forbid case-insensitive `credential_process`

2uasimojo:
A previous commit (#2306 / 13ea4f47) put in checks to forbid the use of `credential_process` in AWS config/credentials files. It turns out that AWS accepts this key case-insensitively, so this commit updates our checks accordingly.

[HIVE-2485](https://issues.redhat.com//browse/HIVE-2485)

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


2uasimojo:
/test security

Fix has merged for this 'un.

Once e2e and e2e-pool pass the first time, I'm just going to force merge this. The things that have merged in the interim are orthogonal and we don't need to burn retests.

PR #2395, opened 2024-08-01
Build faster and leaner: podman-dev-build uses local go cache

2uasimojo:
If a go cache is configured (per `go env GOCACHE`), mount it into the container when running `make podman-dev-build`. This should speed up the build, and also reduce the size of interim container layers.

NOTE: This currently only affects `podman-dev-build`. If you're using docker, or the real `image-hive` target, this doesn't help you.

PR #2398, opened 2024-08-02
HIVE-2584: drop unused Fedora-dev

celebdor:
Fedora-dev builds fell into lack of usage and maintenance. We'll look to improve the main Dockerfile for non-entitled non-registered builds instead.


2uasimojo:
/lgtm

These files are outdated as currently committed.

At this point we have speedy builds via podman with recent changes and #2395 once it lands. If there is demand for actual-docker dev building, we can work something up along similar lines.

PR #2402, opened 2024-08-05
Add python dep 'semver' to hack/requirements.txt

dlom:
When I tried to run hack/bundle-gen.py from within venve this week, I encountered the following error message:

```bash
$ ./bundle-gen.py
  File "./bundle-gen.py", line 10, in <module>
    import semver
ModuleNotFoundError: No module named 'semver'
```

Adding 'semver' to the requirement.txt and re-running pip install fixed the problem

### DIFF
--- a/hack/requirements.txt
+++ b/hack/requirements.txt
@@ -1,3 +1,4 @@
 GitPython
+semver
 PyYAML>=6.0
 requests


2uasimojo (commented):
/lgtm

PR #2416, opened 2024-08-12
fieldManager: "hive-$controllername"

2uasimojo:
When debugging CR updates, e.g. via `--show-managed-fields`, hive used to annoyingly show `manager: manager`, which is not helpful.

With this commit, we've tried to make sure all updates are driven by a fieldManager with a `hive` prefix:
- For many paths, the prefix will be `hiveN` where `N` is an integer unique to the code path that's configuring the fieldManager. We did this to make it easier to trace these paths in the code (which can otherwise be Very Hard™).
- `hiveutil` commands will show up as either `hiveutil-$subcommand` or `hiveN-util-$subcommand`, depending which path they go through to create their client.

If you find cases that still show up as `manager` -- or some other unhelpful string -- please open a bug!

[HIVE-1744](https://issues.redhat.com//browse/HIVE-1744)

**Note to reviewers:**
There's some weird stuff going on in this PR. Because of the way hive has grown organically, and at the hands of several different cadres of developers, the code paths that touch k8s objects do so in a dizzying variety of ways, and many of those have different ways of populating the value that shows up in the `manager` field.
- controller-runtime's client.Client allows you to set it up for all subsequent operations via a `WithFieldOwner` (**not** `WithFieldManager` :roll_eyes:) modifier.
- client-go's write-y REST functions accept a `FieldManager` field in their respective `*Options` arguments.
- In one case we're importing code from the kubectl CLI and using a path which currently only accepts the field manager via argv.
- Some paths may or may not default to using the `user-agent` header if the field manager is not explicitly set.

In conclusion:
- This PR is a scattered hodgepodge.
- Some of the changes may be redundant (though harmless).
- I easily could have missed some paths.

### DIFF
--- a/docs/syncset.md
+++ b/docs/syncset.md
@@ -36,7 +36,7 @@ spec:
 
   resourceApplyMode: Upsert
 
-  enableResourceTemplates  : false
+  enableResourceTemplates: false
 
   resources:
   - apiVersion: user.openshift.io/v1
@@ -54,7 +54,7 @@ spec:
     patch: |-
       { "data": { "foo": "new-bar" } }
     patchType: merge
-    
+
   secretMappings:
   - sourceRef:
       name: ad-bind-password
--- a/pkg/controller/utils/clientwrapper.go
+++ b/pkg/controller/utils/clientwrapper.go
@@ -74,7 +74,7 @@ func NewClientWithMetricsOrDie(mgr manager.Manager, ctrlrName hivev1.ControllerN
 		log.WithError(err).Fatal("unable to initialize metrics wrapped client")
 	}
 
-	return dc
+	return client.WithFieldOwner(dc, "hive1-"+string(ctrlrName))
 }
 
 // AddControllerMetricsTransportWrapper adds a transport wrapper to the given rest config which
--- a/pkg/operator/hive/hive_controller.go
+++ b/pkg/operator/hive/hive_controller.go
@@ -148,6 +148,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+	// This client isn't currently used for write-y things, but future-proof it:
+	tempClient = client.WithFieldOwner(tempClient, hiveOperatorDeploymentName)
 
 	hiveOperatorNS := os.Getenv(HiveOperatorNamespaceEnvVar)
 	r.(*ReconcileHiveConfig).hiveOperatorNamespace = hiveOperatorNS
@@ -394,7 +396,7 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	h, err := resource.NewHelperFromRESTConfig(r.restConfig, hLog)
+	h, err := resource.NewHelperFromRESTConfig(r.restConfig, "operator", hLog)
 	if err != nil {
 		hLog.WithError(err).Error("error creating resource helper")
 		instance.Status.Conditions = util.SetHiveConfigCondition(instance.Status.Conditions, hivev1.HiveReadyCondition, corev1.ConditionFalse, "ErrorCreatingResourceHelper", err.Error())
@@ -740,7 +742,9 @@ func (r *ReconcileHiveConfig) updateHiveConfigStatus(origHiveConfig, newHiveConf
 	}
 	var usNewHiveConfig unstructured.Unstructured
 	usNewHiveConfig.SetUnstructuredContent(content)
-	usUpdated, err := hiveConfigClient.UpdateStatus(context.TODO(), &usNewHiveConfig, metav1.UpdateOptions{})
+	usUpdated, err := hiveConfigClient.UpdateStatus(context.TODO(), &usNewHiveConfig, metav1.UpdateOptions{
+		FieldManager: hiveOperatorDeploymentName,
+	})
 	if err != nil {
 		logger.WithError(err).Error("failed to update HiveConfig status")
 		return err
--- a/pkg/resource/helper.go
+++ b/pkg/resource/helper.go
@@ -66,11 +66,12 @@ func (r *helper) cacheOpenAPISchema() error {
 }
 
 // NewHelperFromRESTConfig returns a new object that allows apply and patch operations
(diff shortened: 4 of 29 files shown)


2uasimojo:
/hold

This doesn't work at all.

2uasimojo:
> /hold
> 
> This doesn't work at all.

Now it does (mostly, except as noted in the description).

/hold cancel

2uasimojo:
/retest

GCP: cloud flaked
Konflux: quay outage

docs/syncset.md
2uasimojo, line 1: ntr: unrelated latent whitespace cleanup

2uasimojo:
/test e2e-azure

Rate limiting again. We may need to talk to DPP about this?

2uasimojo:
> Rate limiting again. We may need to talk to DPP about this?

Eh, having started a conversation with DPP, I went back and tried to find other examples of this rate limiting failure... and couldn't. I must be remembering something else :shrug:

pkg/operator/hive/hive_controller.go
```diff
@@ -110,6 +110,11 @@ func mapToHiveConfig[T client.Object](r reconcile.Reconciler, src string) func(c
 	}
 }
 
+// TODO: This may or may not be setting fieldManager.
```
suhanime, line 113: I prefer the `todo` line to have at least an action item, so anyone who wants to work on it in the future has a clear sense of what to implement in order to remove this line.

pkg/resource/helper.go
```diff
@@ -84,7 +85,8 @@ func NewHelperWithMetricsFromRESTConfig(restConfig *rest.Config, controllerName
 		metricsEnabled: true,
 		controllerName: controllerName,
 		cacheDir:       getCacheDir(logger),
-		restConfig:     restConfig,
+		// TODO: This may or may not set fieldManager
```
suhanime, line 88: Same as above

pkg/resource/patch.go
```diff
@@ -50,7 +51,11 @@ func (r *helper) Patch(name types.NamespacedName, kind, apiVersion string, patch
 }
 
 func (r *helper) setupPatchCommand(name, kind, apiVersion, patchType string, f cmdutil.Factory, patch string, ioStreams genericclioptions.IOStreams) (*kcmdpatch.PatchOptions, error) {
-
+	// This is bizarre and I don't know why it works, but it's the only way I could figure out
+	// to set the `manager` properly. HIVE-1744.
+	os.Args = []string{
```
suhanime, line 56: This one gave me much headache trying to find _some_ documentation around it, and I still don't like it - what if this implementation is very different on different os, and it causes bugs? Other places are harmless, but I really hope this one works as expected - atleast on unix systems.

suhanime (commented):
/lgtm
/hold

Holding for Mark/Leah to review if they want.

pkg/resource/patch.go
2uasimojo: Seriously, this was trial and error. I get why the kubectl stuff we're importing would want to slurp from argv. What I can't figure out is why argv doesn't need to look like I expected. In particular, I initially tried
```go
os.Args = []string{
	"--field-manager", "hive8-"+string(r.controllerName),
}
```
...but ended up with `manager: --field-manager`!

I'm not worried about platform compatibility as this will always be running on a POSIX-compliant *ix.

pkg/operator/hive/hive_controller.go
2uasimojo: I looked deeper and am _pretty_ sure none of our paths will fall back to using user-agent, so I reverted this and its callers.

pkg/resource/helper.go
2uasimojo: Same as above.

2uasimojo:
/test e2e-azure

OSCI workflow setup flake

pkg/resource/patch.go
suhanime: The most related [PR](https://github.com/golang/go/commit/96b79bd3208078835d6a1b9b8c7050554eccc734) I could find (which seriously isn't much) at least affirms that argv[1] cannot be empty when the eventual syscall happens, and when it is, they force it to be set as the parent function (the caller) before executing the child.
So I suppose we at least aren't causing a faux pas.

pkg/controller/utils/clientwrapper.go
```diff
@@ -74,7 +74,7 @@ func NewClientWithMetricsOrDie(mgr manager.Manager, ctrlrName hivev1.ControllerN
 		log.WithError(err).Fatal("unable to initialize metrics wrapped client")
 	}
 
-	return dc
+	return client.WithFieldOwner(dc, "hive1-"+string(ctrlrName))
```
suhanime, line 77: I didn't mention this earlier but I'm glad you were able to find these paths and number them for debugging. I suppose some day we should try and streamline the ways we touch k8s objects - we might even succeed at lessening the vendored deps.

suhanime (commented):
/lgtm

pkg/controller/utils/clientwrapper.go
2uasimojo: Totally agree it would be great to consolidate to a much smaller set of k8s-client-y code.

Off the top, it seems like the major categories we have are:
- Controllers talking to hub
- Controllers talking to spoke
- CLIs (imageset, provision & deprovision pods; other standalone `hiveutil`z) talking to hub

I can see why these should go through different paths high in the call stack, but once it comes to actually creating the kube client from a config, I'm not sure there's any reason for them to be divergent.

That said, we would want to be _very_ thorough in testing the critical paths -- like clustersync! -- if we swap those guts out. It's hard to justify replacing foundational pieces of components that are at this point so well hardened and proven.

dlom: Hardened, proven, and entirely opaque... I agree and think a better future is possible

dlom:
/lgtm

@2uasimojo unhold at will

2uasimojo:
> > Rate limiting again. We may need to talk to DPP about this?
> 
> Eh, having started a conversation with DPP, I went back and tried to find other examples of this rate limiting failure... and couldn't. I must be remembering something else 🤷

Re this, we're tracking the rate limiting via OCPBUGS-38609, and we did find at least one other instance. (It's not related to this PR.)

PR #2417, opened 2024-08-15
ClusterPool: Label owned ClusterDeployments

2uasimojo:
Since ClusterPools create their ClusterDeployments in unique namespaces, it could be tricky to find them -- you would have to do some variant of `oc get cd -A | grep $clusterpoolName` or some fancy jq'ing.

With this commit, we add two labels to ClusterDeployments created by a ClusterPool:
- `hive.openshift.io/clusterpool-namespace`
- `hive.openshift.io/clusterpool-name`

...with the obvious values. With this, you should be able to find all CDs for a pool unambiguously with a query like:

`oc get cd -A -l hive.openshift.io/clusterpool-namespace=poolns,hive.openshift.io/clusterpool-name=poolname`

Often you'll only have one pool in a namespace, or your pool name will be unique in your cluster, so you'll be able to get away with just one of those filters.

Note that upgrading your hive to/past this commit will not automatically label existing ClusterPool-owned ClusterDeployments. If this is crucial:
- Scale your pool down to zero
- Delete all existing ClusterClaims
- Scale your pool back to the desired size

...or manually label them :)

[HIVE-2597](https://issues.redhat.com//browse/HIVE-2597)

### DIFF
--- a/pkg/controller/clusterpool/clusterpool_controller.go
+++ b/pkg/controller/clusterpool/clusterpool_controller.go
@@ -41,6 +41,8 @@ import (
 const (
 	ControllerName                  = hivev1.ClusterpoolControllerName
 	finalizer                       = "hive.openshift.io/clusters"
+	clusterPoolNamespaceLabelKey    = "hive.openshift.io/clusterpool-namespace"
+	clusterPoolNameLabelKey         = "hive.openshift.io/clusterpool-name"
 	imageSetDependent               = "cluster image set"
 	pullSecretDependent             = "pull secret"
 	credentialsSecretDependent      = "credentials secret"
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -781,7 +791,7 @@ func (r *ReconcileClusterPool) createCluster(
 		}
 	}
 
-	if err := r.patchInstallConfig(clp, cd, secret, cdcs, logger); err != nil {
+	if err := r.patchInstallConfig(clp, cd, secret, cdcs); err != nil {
```
2uasimojo, line 794: ntr: unrelated cleanup: latent unused param (yellow squiggly line in vscode)

pkg/controller/clusterpool/clusterpool_controller_test.go
```diff
@@ -1314,6 +1314,10 @@ func TestReconcileClusterPool(t *testing.T) {
 			expectedTotalClusters:              1,
 			expectedObservedSize:               0,
 			expectedObservedReady:              0,
+			// Wedge in a check for the default labels. We're doing this on a test case that ends
+			// up with at least one cluster, but where we don't prepopulate any (which we would
+			// have to label explicitly -- see note in the test driver).
+			expectedLabels: map[string]string{},
```
2uasimojo, line 1320: ntr: existing test case [above](https://github.com/openshift/hive/pull/2417/files#diff-eeaa0ab295d9f13e4b97378aee0241071403161c3932b437df0fb65f5629d4c5R812) where explicit ClusterPool.Spec.Labels are provided -- we verify that the defaults are added as well.

PR #2420, opened 2024-08-16
MachinePool API: MachineLabels

2uasimojo:
The existing MachinePool.Spec.Labels field populates the resulting MachineSets' *MachineSpec*, resulting in the specified labels ending up on *Nodes*.

This commit adds MachinePool.Spec.MachineLabels, which populates the *MachineTemplateSpec*, resulting in the specified labels ending up on *Machines*. But note that we will ignore labels that conflict with those generated by installer code:
- machine.openshift.io/cluster-api-cluster
- machine.openshift.io/cluster-api-machineset

We also add MachinePool.Status.OwnedMachineLabels, corresponding to the existing OwnedLabels, to help us merge values correctly.

[HIVE-2320](https://issues.redhat.com//browse/HIVE-2320)

### DIFF
--- a/apis/hive/v1/machinepool_types.go
+++ b/apis/hive/v1/machinepool_types.go
@@ -46,10 +46,19 @@ type MachinePoolSpec struct {
 	Platform MachinePoolPlatform `json:"platform"`
 
 	// Map of label string keys and values that will be applied to the created MachineSet's
-	// MachineSpec. This list will overwrite any modifications made to Node labels on an
-	// ongoing basis.
+	// MachineSpec. This affects the labels that will end up on the *Nodes* (in contrast with
+	// the MachineLabels field). This list will overwrite any modifications made to Node labels
+	// on an ongoing basis.
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
+
+	// Map of label string keys and values that will be applied to the created MachineSet's
+	// MachineTemplateSpec. This affects the labels that will end up on the *Machines* (in
+	// contrast with the Labels field). This list will overwrite any modifications made to
+	// Machine labels on an ongoing basis. Note: We ignore entries that conflict with
+	// generated labels.
+	// +optional
+	MachineLabels map[string]string `json:"machineLabels,omitempty"`
 
 	// List of taints that will be applied to the created MachineSet's MachineSpec.
 	// This list will overwrite any modifications made to Node taints on an ongoing basis.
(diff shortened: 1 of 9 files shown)


2uasimojo:
TODO: UT

2uasimojo:
/retest

/assign @suhanime 

This works. Though I _thought_ the label updates were supposed to be reconciled immediately by MAPI, but that wasn't the case in my testing (including for the existing node labels).

apis/hive/v1/machinepool_types.go
```diff
@@ -46,11 +46,19 @@ type MachinePoolSpec struct {
 	Platform MachinePoolPlatform `json:"platform"`
 
 	// Map of label string keys and values that will be applied to the created MachineSet's
-	// MachineSpec. This list will overwrite any modifications made to Node labels on an
-	// ongoing basis.
+	// MachineSpec. This affects the labels that will end up on the *Nodes* (in contrast with
+	// the MachineLabels field).This list will overwrite any modifications made to Node labels
```
2uasimojo, line 50:
```suggestion
	// the MachineLabels field). This list will overwrite any modifications made to Node labels
```

:roll_eyes: 
If an update is needed...

apis/hive/v1/machinepool_types.go
```diff
@@ -46,11 +46,19 @@ type MachinePoolSpec struct {
...
+	// on an ongoing basis.
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
 
+	// Map of label string keys and values that will be applied to the created MachineSet's
+	// MachineTemplateSpec. This affects the labels that will end up on the *Machines* (in
+	// contrast with the Labels field).This list will overwrite any modifications made to
```
2uasimojo, line 57: "

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -743,23 +753,34 @@ func (r *ReconcileMachinePool) syncMachineSets(
...
-				// Carry over the labels on the remote machineset from the generated machineset. Merging strategy preserves the unique labels of remote machineset.
-				if rMS.Spec.Template.Spec.Labels == nil {
-					rMS.Spec.Template.Spec.Labels = make(map[string]string)
-				}
-				// First delete the labels that need to be deleted, then sync the remoteMachineSet labels with that of the generatedMachineSet.
-				for key := range rMS.Spec.Template.Spec.Labels {
-					if labelsToDelete.Has(key) {
-						// Safe to delete while iterating as long as we're deleting the key for this iteration.
-						delete(rMS.Spec.Template.Spec.Labels, key)
-						objectMetaModified = true
```
2uasimojo, line 755: @suhanime I noticed this path used `objectMetaModified` whereas L761 used `objectModified`. They both seem to have the same effect on the outcome, so I consolidated on the latter in my update. Any reason they were different before?

2uasimojo:
/test e2e

Starting to wonder if it's not a flake. But I forgot to add machinepool log capture when I sharded the controller :(

pkg/controller/machinepool/machinepool_controller.go
suhanime: I went looking, and the only difference I could see was for [resourcemerge.EnsureObjectMeta](https://github.com/openshift/hive/blob/master/pkg/controller/machinepool/machinepool_controller.go#L710). I don't know if we should be changing that - but I'll leave that upto you.

suhanime (commented):
@2uasimojo I didn't have a chance to look at this PR too closely, but everything seems okay - at least it just follows the footsteps of MachineSet Labels code - apart from the tests and the part where you change the objectMetaModified. I'm not sure about the latter, but everything else LGTM.

pkg/controller/machinepool/machinepool_controller.go
2uasimojo: Okay, yeah, it really doesn't look like there's much need for both vars at all, but in any case that the use of the `Meta` one in this context was most likely just a mistake (though a benign one). Let's roll with it :)

2uasimojo:
Going to rebase this to pick up #2423 so I can check the machinepool logs to see what's going on here.

2uasimojo:
Okey dokey:

```
time="2024-08-20T22:49:58.831Z" level=error msg="unable to create machine set" controller=machinepool error="admission webhook \"validation.machineset.machine.openshift.io\" denied the request: spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`" machinePool=cluster-test/hiveci-66c51341-974-infra reconcileID=8f56xqnk
```

/hold

2uasimojo:
I hadn't considered that installer code generates the MachineSet with a couple of labels populated (see updated commit message) and was clobbering them. This led me to a latent bug whereby, if we were matching a MachineSet with a different name, we would have updated that label to have the wrong value. Both things are now fixed.

/cc @lleshchi

2uasimojo:

PR #2422, opened 2024-08-19
HIVE-2599: Add mkdocs for Red Hat Developer Hub Hive App

celebdor:
This file will allow Red Hat Developer Hub to link to our in-repo documentation.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,0 +1,23 @@
+docs_dir: ./docs
+site_name: Hive Documentation
+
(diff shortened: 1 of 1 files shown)


2uasimojo:
- Where will this show up?
- Having this in hardcopy, it seems like it will quickly diverge as we update the actual docs and forget to update this thing. Do we have plans to add a way to e.g. generate this content from the directory contents, and enforce that it's up to date via `verify`?

PR #2425, opened 2024-08-22
Remove existing Secret files in (deprovision) Pod

2uasimojo:
For Pods with `restartPolicy: OnFailure`, a failed container may be rerun in the same Pod, which will reuse the same file system as the initial run. When we project Secrets (for credentials, certs, etc) to directories in such containers, those writes can fail the second time around because the file already exists. Fix by removing the file, if it exists, before we write it.

Note that at the time of this commit, this only affects deprovision pods:
- imageset pods don't use ProjectToDir
- provision pods have `restartPolicy: Never`

HIVE-2604

### DIFF
--- a/contrib/pkg/utils/generic.go
+++ b/contrib/pkg/utils/generic.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


dlom:
/lgtm

I don't envy the backport process for this...

2uasimojo:

> I don't envy the backport process for this...

Shouldn't be so bad, though I will probably have to one of them manually.

/cherry-pick mce-2.6 mce-2.5 mce-2.4 mce-2.3

PR #2400, opened 2024-08-02
Hive 2485/mce 2.4: Backport AssumeRole, credential_process, and kubeconfig exec fixes

2uasimojo:
Almost perfectly clean backport of #2391. Just imports in pkg/install/generate.go needed cleanup.

[HIVE-2485](https://issues.redhat.com//browse/HIVE-2485)

PR #2387, opened 2024-07-31
Bump azidentity

2uasimojo:
https://security.snyk.io/vuln/SNYK-GOLANG-GITHUBCOMAZUREAZURESDKFORGOSDKAZIDENTITY-7246767

Manual cherry-pick of #2319 / #2308.

[HIVE-2532](https://issues.redhat.com//browse/HIVE-2532)

PR #2430, opened 2024-08-27
Remove existing Secret files in (deprovision) Pod

2uasimojo:
For Pods with `restartPolicy: OnFailure`, a failed container may be rerun in the same Pod, which will reuse the same file system as the initial run. When we project Secrets (for credentials, certs, etc) to directories in such containers, those writes can fail the second time around because the file already exists. Fix by removing the file, if it exists, before we write it.

Note that at the time of this commit, this only affects deprovision pods:
- imageset pods don't use ProjectToDir
- provision pods have `restartPolicy: Never`

[HIVE-2604](https://issues.redhat.com//browse/HIVE-2604)

(cherry picked from commit 4af190c2db19aecd7f9b07e8d926882e99dcfd79)

Conflicts:
	contrib/pkg/utils/generic.go

### DIFF
--- a/contrib/pkg/utils/generic.go
+++ b/contrib/pkg/utils/generic.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


2uasimojo:
/assign @dlom 

Manual backport, just had to get around a var name change.

2uasimojo:
/override ci/prow/security

Backport of #2387, if that happens.

2uasimojo:
Closing in favor of successful bot cherry-pick via #2431 since #2400 landed.

PR #2434, opened 2024-08-28
GOCACHE dir may not exist

2uasimojo:
...if this is the first build in this file system.

In this case, `podman` will freak out when passed that directory as a volume mount.

So check the directory and only use it if it exists.


2uasimojo:

PR #2436, opened 2024-08-29
Revendor installer: OpenStack >=1 MachineSet

2uasimojo:
We got unlucky when we revendored installer and picked up OCPBUGS-17064, which causes the MachineSets generator to produce zero MachineSets when the input specifies zero Replicas, as it does when we're autoscaling.

Revendor to the (current) tip of the 4.14 branch to pick up the fix.

[HIVE-2590](https://issues.redhat.com//browse/HIVE-2590)


2uasimojo:
/override ci/prow/security

#2387 (which might possibly work now that we've revendored installer here??)

2uasimojo:
/cherry-pick mce-2.4

I have a feeling we'll conflict in the vendor dirs, but it's worth a try :)

dlom:
/lgtm

Vendoring the entire installer sucks and I wish there was a better way :/

2uasimojo:
Manually applied to mce-2.4 via #2438.

PR #2438, opened 2024-09-03
Revendor installer: OpenStack >=1 MachineSet

2uasimojo:
We got unlucky when we revendored installer and picked up OCPBUGS-17064, which causes the MachineSets generator to produce zero MachineSets when the input specifies zero Replicas, as it does when we're autoscaling.

Revendor to the (current) tip of the 4.14 branch to pick up the fix.

[HIVE-2590](https://issues.redhat.com//browse/HIVE-2590)

Manual redo of #2436


2uasimojo:
/override ci/prow/security

#2387 backport

2uasimojo:
/assign @dlom 

Local test on openstack checked out ✓

PR #2444, opened 2024-09-09
OCPBUGS-41370: PrivateLink AWS hub actuator to select a vpc in the hub account.

jstuever:
Prior to this change, the AWS hub actuator used the first VPC in the list of associatedVPCs when creating the DNS hosted zone. However, if that VPC was in a different account, this caused an error due to permissions.

This change modifies the behavior so that it selects the EndPoint VPC when on AWS. Otherwise, it selects the first VPC that is in the same account. It also modifies the deletion behavior to only look for the hosted zone in VPCs that use the same account. As a result, at least one item must be added to the list of associatedVPCs that either has no credentails specified, or uses the same secret as the primary account.

### DIFF
--- a/pkg/controller/privatelink/actuator/awsactuator/awshubactuator.go
+++ b/pkg/controller/privatelink/actuator/awsactuator/awshubactuator.go
@@ -212,7 +212,12 @@ func (a *AWSHubActuator) ensureHostedZone(cd *hivev1.ClusterDeployment, metadata
 
 	hzID, err := a.findHostedZone(associatedVPCs, apiDomain)
 	if err != nil && errors.Is(err, errNoHostedZoneFoundForVPC) {
-		newHzID, err := a.createHostedZone(&associatedVPCs[0], apiDomain)
+		selectedVPC, err := a.selectHostedZoneVPC(cd, metadata, logger)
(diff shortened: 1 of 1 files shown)


pkg/controller/privatelink/actuator/awsactuator/awshubactuator.go
```diff
@@ -567,27 +576,71 @@ func (a *AWSHubActuator) getAssociatedVPCs(
...
+			return selectedVPC, errors.Wrap(err, "error getting Endpoint VPC")
 		}
 
-		associatedVPCs = append(associatedVPCs, endpointVPC)
+		if endpointVPC.VPCID == "" {
+			return selectedVPC, errors.New("unable to select Endpoint VPC: Endpoint not found")
+		}
+
+		return endpointVPC, nil
 	}
```
2uasimojo, line 629: I feel like there's an opportunity to DRY some of this -- though I can see why you didn't.

Perhaps `getEndpointVPC()`'s return could indicate whether `VPCEndpointID` was unset in the status?

2uasimojo (commented):
Would be pretty nice to get some UT going for this stuff...

/lgtm

PR #2445, opened 2024-09-10
Hibernation controller documentation update

lleshchi:
[HIVE-2215](https://issues.redhat.com/browse/HIVE-2215)


docs/hibernating-clusters.md
```diff
@@ -143,7 +310,14 @@ we should follow similar checks as the [cluster machine approver](https://github
 to ensure we are not introducing an additional security exposure.
 
 #### Resuming from a Hibernating State
-When a cluster is hibernated, the unreachable controller should properly set the unreachable condition on
-the cluster once it stops responding. This will cause other controllers like the remotemachineset controller to
-stop trying to reconcile the cluster. Once the cluster deployment resumes, the unreachable controller should
-set it back to reachable and syncing of hive controllers should resume.
+The unreachable controller tracks the reachability of a cluster and is responsible for updating the ClusterDeploymentConditionType accordingly.
+
+Once a cluster hibernates and stops responding, the unreachable controller sets the ClusterDeploymentConditionType to the UnreachableCondition:
```
2uasimojo, line 315:
```suggestion
The unreachable controller tracks the reachability of a cluster and is responsible for updating the `Unreachable` status condition accordingly.

Once a cluster hibernates and stops responding, the unreachable controller sets the `Unreachable` condition's `Status` to `True`.
```

...and I don't think we need the code paste.

docs/hibernating-clusters.md
```diff
@@ -143,7 +310,14 @@ we should follow similar checks as the [cluster machine approver](https://github
...
+
+```go
+const (
+	// UnreachableCondition indicates that Hive is unable to establish an API connection to the remote cluster.
+	UnreachableCondition ClusterDeploymentConditionType = "Unreachable"
+)
+```
+Once the cluster resumes, the unreachable controller unsets the condition and the hive controllers are free to resume syncing.
```
2uasimojo, line 323: sets the condition to `False`?

2uasimojo (commented):
I would like to punt scrutiny of the flow diagram to @suhanime 

Otherwise, I like where this is going. In general I would prefer to see more explanations/examples and less code copy/paste.

suhanime:
> I would like to punt scrutiny of the flow diagram to @suhanime

So, I have only reviewed the flow diagram - I like the design, but have some comments:
- I do not think `ResumingOrRunning` is a valid powerState in the flow (ShouldStartMachines > no - notes). I think you got confused at [this line
](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller.go#L881) Logically, you can ignore that state as far as powerState is concerned since we already check the ready condition before hitting this point, the other reasons are correct - 2 of them are error states
- you don't have to eliminate SyncSetsNotApplied - it simply means the cluster is new and the clustersync's first success time is nil, and we do eventually [requeue](https://github.com/openshift/hive/blob/master/pkg/controller/hibernation/hibernation_controller.go#L329:L347) if we need to hibernate in that case. I recall you talked about eliminating that flow path as it sounds confusing, so I don't mind if you still decide to drop it.
- Did you also eliminate the setting the workflow of WaitingForMachines -> WaitingForNodes -> WaitingForClusterOperatorsToSettle -> Running/Ready? Many times the clusters get stuck here, if users want to refer to these docs - or the developer working on it wants to quickly check where they should be looking, it makes sense to include these paths. Infact, if I'm reading this correctly, setting of hibernating (without hibernateafter) or running/ready is missing.

2uasimojo:
/lgtm

for the text changes.

/hold

for @suhanime to ack the diagram changes.

suhanime:
/hold cancel

This is awesome!

PR #2447, opened 2024-09-10
MachinePool GCP: fix zone filter

2uasimojo:
When `MachinePool.Spec.Platform.GCP.Zones` does not explicitly enumerate the zones to use, we attempt to default to all zones in the configured region. However, this query was using an overly-broad regex which matched any zone whose region was a *substring* of the configured region. This broke when GCP added regions `europe-west10` and `europe-west12`, which our filter would retrieve when the cluster was in region `europe-west1`.

The zone list response includes the region as a full URI (e.g. `https://www.googleapis.com/compute/v1/projects/openshift-hive-dev/regions/europe-west1`). Fix our regex to match our full region string, anchored by the last `/` and the end of the string.

[HIVE-2610](https://issues.redhat.com//browse/HIVE-2610)

### DIFF
--- a/pkg/controller/machinepool/gcpactuator.go
+++ b/pkg/controller/machinepool/gcpactuator.go
@@ -246,8 +246,15 @@ func (a *GCPActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *hi
 func (a *GCPActuator) getZones(region string) ([]string, error) {
 	zones := []string{}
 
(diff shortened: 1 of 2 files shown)


2uasimojo:
/hold

This ain't right yet.

2uasimojo:
/hold cancel

Sorted.

2uasimojo:
/override "Red Hat Konflux / hive-operator-enterprise-contract"

2uasimojo:

PR #2449, opened 2024-09-11
Azure: make creds case-insensitive

2uasimojo:
Installer accepts azure creds (osServicePrincipal.json) with case-insensitive keys, but previously we were parsing them into a generic map and extracting the values by exact (case-sensitive) keys.

This commit switches from generic map to the installer-provided Credentials struct. This makes the unmarshalling case-insensitive.

[ACM-14248](https://issues.redhat.com//browse/ACM-14248)


2uasimojo:
Tested live, works.

PR #2454, opened 2024-09-12
OCPBUGS-41797: PrivateLink: GCP PSC to use global endpoints

jstuever:
Enabling AllowPscGlobalAccess so that the endpoint is available from all of the regions as is necessary when the endpoint exists in a region other than the one a VPN is connected to.


2uasimojo:
The code path only affects PSC, which we'll get a chance to validate live.

/lgtm

PR #2459, opened 2024-09-16
ACM-13064: ClusterInstall: Idempotence improvements

2uasimojo:
When reconciling a ClusterInstall, we may need to update both the spec and the status of the ClusterDeployment, and those need to be done in separate operations (because subresource). If both succeed, great. If the first fails, fine -- we'll requeue and try the whole thing again. But if the first succeeds and the second fails, we need to make sure we're in a state where the subsequent reconcile will still retry it.

Prior to this commit, we were doing the spec update first. In a normal sequence, this would set cd.Spec.Installed to true, which would activate a branch on the subsequent reconcile that would skip the routine that would have retried the status update.

This commit swaps the order -- we do the status update first -- based on the fact that the affected fields have no effect on the controller logic flow elsewhere, thus allowing subsequent reconciles to hit the correct routine again and perform the spec update to complete the contract.

But we also had to make the above (status fields having no effect on the controller flow) be true, prompting several small updates to tighten the logic and ensure updates were made only and always when needed.

[ACM-13064](https://issues.redhat.com//browse/ACM-13064)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterinstalls.go
+++ b/pkg/controller/clusterdeployment/clusterinstalls.go
@@ -30,6 +30,7 @@ func (r *ReconcileClusterDeployment) reconcileExistingInstallingClusterInstall(c
 		Version: ref.Version,
 		Kind:    ref.Kind,
 	}
+	var updated bool
 
 	logger = logger.WithField("clusterinstall", ref.Name).WithField("gvk", gvk)
 	logger.Debug("reconciling existing clusterinstall")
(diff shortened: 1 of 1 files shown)


2uasimojo:
/hold

Still working through this.

2uasimojo:
/assign @dlom 
/cc @suhanime 

While working on this, I discovered that Mark had most recently touched the code in this area, so reassigning.

2uasimojo:
fmr: [thread](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1726508640199429) where this was developed and tested live.

PR #2460, opened 2024-09-17
MachinePool: GCP: Custom UserTags

2uasimojo:
Add a new API field, MachinePool.Spec.GCP.UserTags, and plumb it into the GCP MachineSet generator. Results end up in the GCP's providerSpec.ResourceManagerTags (per upstream installer code at the time of this commit).

NOTE: This commit breaks ground on a new way of enhancing hive's APIs based on the installer: We are including installer as a dependency in the `apis/` submodule for the first time to pull in the `UserTag` type by reference rather than copying it as we have done in the past. The dependency tooling seems smart enough to avoid pulling all of installer's transitive dependencies into the `apis/` submodule. We must be vigilant that this remains true going forward.


pkg/controller/machinepool/gcpactuator.go
```diff
@@ -85,13 +85,15 @@ func NewGCPActuator(
...
+	var imageID string
+	imageID, err = getGCPImageID(masterMachine, pool, logger)
 	if err != nil {
 		logger.WithError(err).Error("error getting image ID from master machine")
 		return nil, err
 	}
 
-	network, subnet, err := getNetwork(remoteMachineSets, scheme, logger)
+	var network, subnet string
+	network, subnet, err = getNetwork(remoteMachineSets, logger)
```
2uasimojo, line 96: ntr: unrelated delinting, remove unused `scheme` param.

PR #2463, opened 2024-09-19
MachinePool: GCP: Custom UserTags

2uasimojo:
Add a new API field, MachinePool.Spec.GCP.UserTags, and plumb it into the GCP MachineSet generator. Results end up in the GCP's providerSpec.ResourceManagerTags (per upstream installer code at the time of this commit).

NOTE: This commit breaks ground on a new way of enhancing hive's APIs based on the installer: We are including installer as a dependency in the `apis/` submodule for the first time to pull in the `UserTag` type by reference rather than copying it as we have done in the past. The dependency tooling seems smart enough to avoid pulling all of installer's transitive dependencies into the `apis/` submodule. We must be vigilant that this remains true going forward.

(cherry picked from commit c7581516b99535b743844d12fa574828d5ea53b7)


2uasimojo:
/assign @jstuever 

Had to do this manually to get around conflicts with #2453 (which looks like it won't be backportable).

PR #2467, opened 2024-09-23
ClusterVersion: tunable poll interval

2uasimojo:
Add HiveConfig.Spec.ClusterVersionPollInterval, a time.Duration value which (plus random jitter of 0-1s) will be used to requeue ClusterDeployments for the clusterversion controller, which is responsible for setting `hive.openshift.io/version*` labels on ClusterDeployment objects based on the ClusterVersion object on the spoke.

If unset or zero, this is disabled, resulting in the legacy behavior: we will reconcile on controller startup, ClusterDeployment changes, or (I think) the default controller-runtime interval of 1000s.

[HIVE-2619](https://issues.redhat.com//browse/HIVE-2619)

### DIFF
--- a/pkg/controller/clusterversion/clusterversion_controller.go
+++ b/pkg/controller/clusterversion/clusterversion_controller.go
@@ -3,7 +3,10 @@ package clusterversion
 import (
 	"context"
 	"fmt"
+	"math/rand"
+	"os"
 	"strconv"
+	"time"
 
 	"github.com/blang/semver/v4"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -155,8 +158,21 @@ func (r *ReconcileClusterVersion) Reconcile(ctx context.Context, request reconci
...
+	var requeueAfter time.Duration
+	if envPollInterval := os.Getenv(constants.ClusterVersionPollIntervalEnvVar); len(envPollInterval) > 0 {
+		var err error
+		requeueAfter, err = time.ParseDuration(envPollInterval)
+		if err != nil {
+			cdLog.WithError(err).WithField("pollInterval", envPollInterval).Errorf("unable to parse %s, disabling poll interval", constants.ClusterVersionPollIntervalEnvVar)
+			os.Setenv(constants.ClusterVersionPollIntervalEnvVar, "")
+		}
+		// Add a random fraction of a second jitter to reduce bunching
+		requeueAfter += time.Duration(rand.Float64() * float64(time.Second))
```
suhanime, line 170: If unset or zero, you're defaulting to legacy behaviour. Shouldn't we also default to legacy behaviour when envPollInterval cannot be parsed? This code path will result in requeues within a fraction of second - you might want to add an exit if err != nil right above this

2uasimojo: Hm, you're correct that I missed this, and I'll fix it.

Not returning an error was deliberate, as I'm effectively switching off this entire branch by `Setenv()`ing the var to empty, and saw no reason to trigger an immediate requeue in that case. However, as currently written, we'll requeue once, after a random fraction of 1s -- and on that iteration we'll end up with `RequeueAfter: 0`.

I think just adding `else` will get the behavior we want. Agree?

2uasimojo: Eh, I also think I have the wrong behavior if the var is explicitly set to zero -- in that case we'll always requeue in a random fraction of a second!

2uasimojo:
Good eye @suhanime, thanks! Ready for re-review.

2uasimojo:

PR #2468, opened 2024-09-26
HIVE-2622: GCP PrivateLink to use pre-existing psc subnet.

jstuever:
Prior to this change, GCP PrivateLink created a subnet in the cluster VPC to be used by the service attachment. To do so, it queried the VPC network from the api and used the returned selfLink for creating the subnet. However, when using BYO VPCs, the network name is not guaranteed to match the name created by the installer, so the privatelink controller would fail to find it. Furthermore, in a Shared VPC configuration, Hive is unable to create the subnet in the host project.

This change adds new fields to the clusterDeployment spec for GCP PrivateLink configuration to enable a pre-existing subnet to be configured. In order to accommodate Shared VPC, it also enables the configuration of the project name. As a result, GCP PrivateLink can now deploy using BYO VPC in addition to Shared VPC configurations.


pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
```diff
@@ -145,45 +149,61 @@ func (a *GCPLinkActuator) Reconcile(cd *hivev1.ClusterDeployment, metadata *hive
...
-	logger.Debug("reconciling Service Attachment Subnet")
-	subnetModified, subnet, err := a.ensureServiceAttachmentSubnet(cd, metadata, network)
-	if err != nil {
-		logger.WithError(err).Error("failed to reconcile the Service Attachment Subnet")
-
-		err := conditions.SetErrConditionWithRetry(*a.client, cd, "ServiceAttachmentSubnetReconcileFailed", errors.New(controllerutils.ErrorScrub(err)), logger)
+	var subnet *compute.Subnetwork
+	// Use a preexisting subnet if configured.
+	if subnetName := getServiceAttachmentSubnetExistingName(cd); subnetName != "" {
+		subnet, err = a.gcpClientSpoke.GetSubnet(subnetName, cd.Spec.Platform.GCP.Region, getServiceAttachmentSubnetExistingProject(cd))
```
2uasimojo, line 155: nit: debug log saying we're using existing subnet?

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
```diff
@@ -145,45 +149,61 @@ func (a *GCPLinkActuator) Reconcile(cd *hivev1.ClusterDeployment, metadata *hive
...
-	if firewallModified {
-		err := conditions.SetReadyConditionWithRetry(*a.client, cd, corev1.ConditionFalse,
-			"ReconciledServiceAttachmentFirewall",
-			"reconciled the Service Attachment Firewall",
-			logger)
+		if subnetModified {
+			err := conditions.SetReadyConditionWithRetry(*a.client, cd, corev1.ConditionFalse,
+				"ReconciledServiceAttachmentSubnet",
+				"reconciled the Service Attachment Subnet",
+				logger)
```
2uasimojo, line 181: Should be we be tweaking this condition in the other code path? Like "Validated existing service attachment subnet" kind of thing?

2uasimojo (commented):
Seems sane to me. I think @dlom is the official reviewer. Couple of minor thoughts inline.

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
jstuever: It exists a few lines lower (159) using values from the retrieved subnet. Though, that value should be equal to the subnetName and could be moved higher up. Would you prefer that?

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
```diff
@@ -299,17 +321,42 @@ func (a *GCPLinkActuator) ensureServiceAttachmentSubnet(cd *hivev1.ClusterDeploy
...
+		cd.Spec.Platform.GCP != nil &&
+		cd.Spec.Platform.GCP.PrivateServiceConnect != nil &&
+		cd.Spec.Platform.GCP.PrivateServiceConnect.ServiceAttachment != nil &&
+		cd.Spec.Platform.GCP.PrivateServiceConnect.ServiceAttachment.Subnet != nil &&
+		cd.Spec.Platform.GCP.PrivateServiceConnect.ServiceAttachment.Subnet.Existing != nil {
+
+		return cd.Spec.Platform.GCP.PrivateServiceConnect.ServiceAttachment.Subnet.Existing.Project
 	}
-	return cidr
+	return ""
```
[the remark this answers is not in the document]
dlom: It's a shame that such a hack is required to write concise code

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
jstuever: IMHO: no due to the fact that we are also not populating a value for service attachment subnet in the cd.status. We didn't really reconcile it so much as validate the input prior to using it.

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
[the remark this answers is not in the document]
jstuever: Noted. I wasn't aware of this hack. I can do this if there arises a need to make another change (and therefore do another 3hr manual testing cycle on the PR).

dlom:
/lgtm

Will QE look at this at some point?

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
2uasimojo: Ack. I was thinking more as a way to show progress, but I'm okay with it as is.

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
2uasimojo: oh, no, I just missed it, it's cool.

pkg/controller/privatelink/actuator/gcpactuator/gcplinkactuator.go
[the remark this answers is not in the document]
2uasimojo: > I wasn't aware of this trick.

Heh, me neither, wasn't even sure it would work. I had thought of it in passing every time I saw this pattern, but finally decided to try it today. (Those playground links are mine.)

> It's a shame that such a hack is required to write concise code

It's not super different from try/except (or try/catch, or however it's spelled in your preferred language). Though definitely has a different feel, which takes some getting used to. Like channels and goroutines.

I'm curious how you would do it in python or perl or java or whatever?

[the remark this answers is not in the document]
jstuever: In my experience, most languages where null pointer exceptions are possible would require what I did here by walking the chain.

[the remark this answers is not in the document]
2uasimojo: Or for example

```python
try:
  return thing.field.otherfield.whatever
except NullPointerException:
  return default
```

The only improvement here is being specific about what error we expect, allowing any other error to continue to panic. It may be possible for golang's `recover()` to be more specific similarly -- haven't played with it.

jstuever:

PR #2470, opened 2024-09-26
clusterversion controller: crash on unparseable poll interval

2uasimojo:
We recently (#2467) added a tunable poll interval for the clusterversion controller. However, we were being too nice about unparseable values, ignoring them and reverting to the default behavior. With this change, we'll crash the controller instead. The pod will go into Error (or CrashLoopBackoff eventually) with a message telling you what you did wrong.

[HIVE-2619](https://issues.redhat.com//browse/HIVE-2619)

### DIFF
--- a/pkg/controller/clusterversion/clusterversion_controller.go
+++ b/pkg/controller/clusterversion/clusterversion_controller.go
@@ -13,7 +13,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


2uasimojo:

pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -41,24 +40,33 @@ const (
...
 		logger.WithError(err).Error("could not get controller configurations")
 		return err
 	}
-	return AddToManager(mgr, NewReconciler(mgr, clientRateLimiter), concurrentReconciles, queueRateLimiter)
-}
 
-// NewReconciler returns a new reconcile.Reconciler
-func NewReconciler(mgr manager.Manager, rateLimiter flowcontrol.RateLimiter) reconcile.Reconciler {
+	var pollInterval time.Duration
+	if envPollInterval := os.Getenv(constants.ClusterVersionPollIntervalEnvVar); len(envPollInterval) > 0 {
```
suhanime, line 51: I missed this earlier but this is great - we should be reading the env earlier when we add to manager so we error out early and not when the reconcile is called

suhanime (commented):
/lgtm

PR #2472, opened 2024-09-26
No-op revendor installer

2uasimojo:
This is a special revendor.

A prior commit (https://github.com/openshift/hive/pull/2460 / https://github.com/2uasimojo/hive/commit/c7581516b99535b743844d12fa574828d5ea53b7) included installer as a dependency in our apis/ package for the first time. This was causing problems for downstream consumers seemingly because the pseudo-version generated by `go mod`  represented a years-old tag with a recent commit, resulting in excessive churn in the go sum database -- see the referenced bug report.

Now a tag has been created at that commit to make this pseudo-version less gappy, hopefully eliminating this problem for downstreams.

However, when I tried to revendor to that tag, go replaced it with a pseudo-version... and then complained that we shouldn't use a pseudo-version when there's a tag available!! (Shakes fist.)

So this commit revendors to the *next* commit, which should mean that a pseudo-version is the only way to reference it. The delta on that commit is null for vendoring purposes (the changed file is not one that makes its way into the vendor/ directory) so this is just a bump of pseudo-versions.

OCPBUGS-42448

### DIFF
--- a/apis/go.mod
+++ b/apis/go.mod
@@ -11,7 +11,7 @@ require (
 	// !WARNING! We intend this to pull in *only* pkg/types. Bringing in anything
 	// else could cause a dependency explosion, which we must avoid in this apis/
 	// package for the sake of our downstream consumers.
-	github.com/openshift/installer v0.9.0-master.0.20240912103450-dcf8320c8c4e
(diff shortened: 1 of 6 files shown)


2uasimojo:

PR #2473, opened 2024-09-27
No-op revendor installer

2uasimojo:
This is a special revendor.

A prior commit (#2463 / d8ebe585) included installer as a dependency in our apis/ package for the first time. This was causing problems for downstream consumers seemingly because the pseudo-version generated by `go mod` represented a years-old tag with a recent commit, resulting in excessive churn in the go sum database -- see the referenced bug report.

Now a tag has been created at that commit to make this pseudo-version less gappy, hopefully eliminating this problem for downstreams.

So this commit revendors to that tag, which computes to the same commit, so there is no actual change in vendoring -- only to go module metadata.

OCPBUGS-42448

### DIFF
--- a/apis/go.mod
+++ b/apis/go.mod
@@ -11,7 +11,7 @@ require (
 	// !WARNING! We intend this to pull in *only* pkg/types. Bringing in anything
 	// else could cause a dependency explosion, which we must avoid in this apis/
 	// package for the sake of our downstream consumers.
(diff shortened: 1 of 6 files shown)


patrickdillon:
/test security

I added v0.90.99 to the failing CVE fix version. Let's see if snyk picks it up.

PR #2475, opened 2024-09-30
Remove installer dep from apis/

2uasimojo:
In a prior change (#2460 / c7581516) we tried using a `UserTag` type imported from openshift/installer, creating a dependency on same for downstream consumers of our apis/ package. This is causing problems downstream with the checksum database, so for the sake of unblocking those downstreams for now, we're pulling that out and copy/pasting the type definition instead (like we have always done in the past with types of this ilk).

We hope to reinstate the dependency in the future, as it would improve the maintainability of hive.

OCPBUGS-42448


jstuever (approved):
/lgtm

2uasimojo:
/override "Red Hat Konflux / hive-mce-27-on-pull-request" "Red Hat Konflux / hive-operator-enterprise-contract / hive"

PR #2476, opened 2024-09-30
modcheck -f phase 1: fix `require`s

2uasimojo:
Our modcheck tool looks at the go.mod files in the repository root and the apis/ directory, and fails if any entries in the latter exist in the former at a different version. Heretofore, such failures would have to be addressed by manually updating the apis/go.mod to sync those versions.

With this change, we add a `-f` flag which asks modcheck to `f`ix discrepancies it encounters.

In this first phase, we add the ability for modcheck to fix `require` entries, as these are by far the most common discrepancies we encounter.

*Further work is required to make the tool fix `exclude` and `replace` entries.*

The tool will succeed (exit zero) if the files are in sync when it completes -- whether because they started that way, or because it fixed everything it encountered.

The tool will fail (exit nonzero) if it encounters a processing error; or if the files are out of sync when it completes -- even if *partially* fixed.

Put another way: barring errors, the exit code is the same as if modcheck were run again without `-f`.

[HIVE-2624](https://issues.redhat.com//browse/HIVE-2624)

### DIFF
--- a/hack/modcheck.go
+++ b/hack/modcheck.go
@@ -7,16 +7,64 @@ import (
 	"golang.org/x/mod/modfile"
 )
 
+const (
+	rootpath = "go.mod"
+	apispath = "apis/go.mod"
+)
+
+// TODO: un-global
+var doFix bool
+
 func main() {
-	rootgmf := readGoMod("go.mod")
-	apisgmf := readGoMod("apis/go.mod")
-	insync := true
-	insync = cmp1("require", mapRequire(rootgmf.Require), mapRequire(apisgmf.Require)) && insync
-	insync = cmp1("exclude", mapExclude(rootgmf.Exclude), mapExclude(apisgmf.Exclude)) && insync
-	insync = cmp2("replace", mapReplace(rootgmf.Replace), mapReplace(apisgmf.Replace)) && insync
+	// TODO: Actual arg parsing
+	if len(os.Args) > 1 && os.Args[1] == "-f" {
+		doFix = true
+	}
+	rootgmf := readGoMod(rootpath)
+	apisgmf := readGoMod(apispath)
+	needWrite := false
+	insync, err := processRequire(*apisgmf, mapRequire(rootgmf.Require), mapRequire(apisgmf.Require))
+	if err != nil {
+		// processRequire() printed the error
+		os.Exit(2)
+	}
+	if doFix && !insync {
+		needWrite = true
+		// *Now* the files are in sync. This informs the exit code, which should be "success"
+		// if we fully fixed the file. (May still be "failure" if other mismatches are found.)
+		insync = true
+	}
+
+	// TODO: Make these respond to doFix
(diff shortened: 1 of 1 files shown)


2uasimojo:

hack/modcheck.go
```diff
@@ -7,16 +7,64 @@ import (
...
+		b, err := apisgmf.Format()
+		if err != nil {
+			fmt.Printf("Couldn't format modified %s: %s\n", apispath, err)
+			os.Exit(2)
+		}
+		if err = os.WriteFile(apispath, b, 0); err != nil {
+			fmt.Printf("Failed to write modified %s: %s\n", apispath, err)
+			os.Exit(2)
+		}
+		fmt.Printf("\tDone")
```
2uasimojo, line 54:
```suggestion
		fmt.Printf("\tDone\n")
```

hack/modcheck.go
```diff
@@ -7,16 +7,64 @@ import (
...
+	}
+
 	if insync {
+		fmt.Printf("%s is in sync\n", apispath)
+		if needWrite {
+			fmt.Printf("\t(after fixing)")
```
2uasimojo, line 60:
```suggestion
			fmt.Printf("\t(after fixing)\n")
```

hack/modcheck.go
```diff
@@ -7,16 +7,64 @@ import (
...
+		}
 		os.Exit(0)
 	} else {
+		fmt.Printf("\n%s is out of sync\n", apispath)
+		if needWrite {
+			fmt.Printf("\t(despite partial fixing)")
```
2uasimojo, line 66:
```suggestion
			fmt.Printf("\t(despite partial fixing)\n")
```

hack/modcheck.go
```diff
@@ -85,7 +133,40 @@ func mapReplace(theList []*modfile.Replace) map[string]replacement {
...
+		}
+		if rootver == apisver {
+			// For a future "verbose mode":
+			// fmt.Printf("\tOK %s %s\n", path, rootver)
+		} else {
+			fmt.Printf("XX require %s: root(%s) apis(%s)\n", path, rootver, apisver)
+			insync = false
+			if doFix {
+				if err := apisfile.DropRequire(path); err != nil {
+					fmt.Printf("Error dropping requirement for %s: %s", path, err)
```
2uasimojo, line 155:
```suggestion
					fmt.Printf("Error dropping requirement for %s: %s\n", path, err)
```

hack/modcheck.go
```diff
@@ -85,7 +133,40 @@ func mapReplace(theList []*modfile.Replace) map[string]replacement {
...
+					return false, err
+				}
+				if err := apisfile.AddRequire(path, rootver); err != nil {
+					fmt.Printf("Error adding requirement for %s: %s", path, err)
```
2uasimojo, line 159:
```suggestion
					fmt.Printf("Error adding requirement for %s: %s\n", path, err)
```

2uasimojo (commented):
spoiled by logger

2uasimojo:

PR #2477, opened 2024-09-30
Upgrade deps to k8s 1.31

2uasimojo:
NOTE: At this level, k8s.io/apis started using an empty `groupName` annotation, which broke the version of controller-gen we were using to generate CRDs. This was fixed via
https://github.com/kubernetes-sigs/controller-tools/pull/1000 which became available in controller-tools v0.16, so this includes an upgrade to v0.16.1.

NOTE: This also prompted an upgrade to controller-runtime v0.19.0.

[HIVE-2616](https://issues.redhat.com//browse/HIVE-2616)

PR #2481, opened 2024-10-03
controller-gen: use vendor/, not build-machinery-go

2uasimojo:
The build-machinery-go project is suffering from bit rot, which is making it difficult to upgrade the version of controller-gen we use to generate CRDs, which we're going to need to do in order to upgrade k8s dependencies (see [HIVE-2616](https://issues.redhat.com//browse/HIVE-2616)).

With this change, we stop using the piece of build-machinery-go that deals with controller-gen, and instead install it directly from our vendor directory.

[HIVE-2626](https://issues.redhat.com//browse/HIVE-2626)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ endif
 # Include the library makefile
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
-	targets/openshift/controller-gen.mk \
 	targets/openshift/yq.mk \
 	targets/openshift/bindata.mk \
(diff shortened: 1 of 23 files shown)


2uasimojo:

2uasimojo:
These changes only affected CRDs, and those in a way that is irrelevant to operations. (Confirmed by passing e2e-pool.)

This is blocking HIVE-2616 / #2477, which is time-sensitive, so rather than wait for e2e, and for everything to retest (since #2476 merged, which doesn't affect the payload or CRDs at all), I'm going to merge manually.

PR #2483, opened 2024-10-03
OCPBUGS-42740: PrivateLink Actuators should return errors on failure.

jstuever:
Prior to this change, the GCP Link actuator failed to return an error when a failure was experienced while creating certian cloud resources. This is because the actuators were overriding the value of the err varliabe prior to using it in a following return statement. As a result, the controller would assume the link actuator was successful and continue calling the hub actuator. This caused, the PrivateLinkFailed condidtion would show an error regarding the failed provisioning of DNS resources instead of the underlying issue in the link actuator.

This change fixes the problem by placing the secondary assignment of the err variable into the corrisponding if block, thereby maintaining the origional value of the variable. This causes the final return to act as expected by condition of the origional error. It also ensures the origional error is always returned while logging any condition status update errorrs inline.

https://issues.redhat.com/browse/OCPBUGS-42740


jstuever:
/cc @2uasimojo
re prior discussions

2uasimojo:

PR #2480, opened 2024-10-02
HIVE-2440: Add doc for PrivateLink Controller

jstuever:

PR #2488, opened 2024-10-08
controller-gen: proper building and pathing

2uasimojo:
In #2481 / 6bb17298 we started getting the controller-gen tool by compiling and installing it from the vendor directory. But the method used there assumed that the resulting binary would end up in the user's `$PATH`, which is not necessarily the case.

With this change, we:
- Use `go list` to discover where the controller-gen binary will be installed; and use this full path when invoking it.
- Treat the controller-gen binary as a proper `make` build target: make it dependent on its source so that it will only be rebuilt if necessary.

[HIVE-2626](https://issues.redhat.com//browse/HIVE-2626)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,15 +119,17 @@ define patch-crd-yq
 
 endef
 
-.PHONY: ensure-controller-gen
-ensure-controller-gen:
-	go install $(shell realpath vendor/sigs.k8s.io/controller-tools/cmd/controller-gen)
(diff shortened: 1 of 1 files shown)


2uasimojo:

2uasimojo:
Manual merge. We don't need to retest this after rebasing on a [doc PR](https://github.com/openshift/hive/pull/2480).

PR #2490, opened 2024-10-08
Rewrite hack/modchecker.go to be more generic

dlom:
Due to limitations in the generics system implemented by go, a small hack is employed to allow us to work generically with foreign types (logical pieces of the go.mod file).  Lots of documentation is written inline

/assign @2uasimojo

### DIFF
--- a/hack/modcheck.go
+++ b/hack/modcheck.go
@@ -2,74 +2,17 @@ package main
 
 import (
 	"fmt"
-	"os"
-
 	"golang.org/x/mod/modfile"
+	"golang.org/x/mod/module"
+	"os"
 )
 
 const (
-	rootpath = "go.mod"
-	apispath = "apis/go.mod"
+	rootPath = "go.mod"
+	apisPath = "apis/go.mod"
 )
 
-// TODO: un-global
-var doFix bool
-
-func main() {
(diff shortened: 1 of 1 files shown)


dlom:
/retitle HIVE-2629: Rewrite hack/modchecker.go to be more generic

2uasimojo (commented):
This is great, @dlom. It's sad that the language ties our hands in ways that make the resulting code _have_ to have awkward constructs, but you've worked into that admirably.

Before we merge this, I would like to see the commit message and PR summary updated to reflect all the things that have changed. You didn't just rewrite for generics; you also implemented fixers for excludes and replaces.

But on that note: I tried to get the tool to do anything with excludes and couldn't. Do you have a test case for that?

...which leads me to wonder what we would actually _expect_ the behavior to be in that regard. Would we want to make sure that, if any exclude with a matching path is given, all the excludes for that path are identical in both files? That probably doesn't make sense given that the root file's go.sum can (and is even quite likely to) contain more versions of any given lib, and thus be subject to more exclusions of that lib.

We don't necessarily have to answer that question here; but whatever we decide, we should call it out in the code comments and the commit/PR messaging.

PR #2495, opened 2024-10-17
Hibernation: Requeue without error when ClusterSync doesn't exist yet

2uasimojo:
When installation finishes, the clustersync and hibernation controllers race to start their business. But the latter needs to make use of the ClusterSync object. When the clustersync controller hasn't created it yet, the hibernation controller needs to requeue and wait for it.

Previously we were just letting the 404 from Get(ClusterSync) to bubble up; but this counts against the reconciler error budget for which alerts may be triggered. Since this is an expected scenario, we change this code path to do a non-error immediate requeue instead.

[HIVE-2611](https://issues.redhat.com//browse/HIVE-2611)

### DIFF
--- a/pkg/controller/hibernation/hibernation_controller.go
+++ b/pkg/controller/hibernation/hibernation_controller.go
@@ -231,7 +231,11 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	clusterSync := &hiveintv1alpha1.ClusterSync{}
 	if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync); err != nil {
 		// This may be NotFound, which means the clustersync controller hasn't created the ClusterSync yet.
-		// Fail and requeue to wait for it to exist.
+		// It does so shortly after the cluster is marked as installed. Requeue immediately to "poll" for it.
+		if apierrors.IsNotFound(err) {
+			cdLog.Info("clustersync does not exist yet; requeueing")
+			return reconcile.Result{Requeue: true}, nil
+		}
 		return reconcile.Result{}, fmt.Errorf("could not get ClusterSync: %v", err)
 	}
 	syncSetsApplied := clusterSync.Status.FirstSuccessTime != nil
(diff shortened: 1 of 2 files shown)


pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -231,7 +231,11 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
 	clusterSync := &hiveintv1alpha1.ClusterSync{}
 	if err := r.Get(context.Background(), types.NamespacedName{Namespace: cd.Namespace, Name: cd.Name}, clusterSync); err != nil {
 		// This may be NotFound, which means the clustersync controller hasn't created the ClusterSync yet.
-		// Fail and requeue to wait for it to exist.
+		// It does so shortly after the cluster is marked as installed. Requeue immediately to "poll" for it.
```
suhanime, line 234: :heavy_check_mark: Error should be thrown by ClusterSync controller (it creates/attempts to create and requeue) instead of this one.

pkg/controller/hibernation/hibernation_controller.go
```diff
@@ -231,7 +231,11 @@ func (r *hibernationReconciler) Reconcile(ctx context.Context, request reconcile
...
+		if apierrors.IsNotFound(err) {
+			cdLog.Info("clustersync does not exist yet; requeueing")
+			return reconcile.Result{Requeue: true}, nil
```
suhanime, line 237: Should be include a time-check here? ClusterSync controller will throw the error, but this controller will continuously requeue until someone fixes it - so how about we requeue if it has been less than 10 min since cd was created, or error out?

suhanime (commented):
Other things look okay, but I'm wondering if maybe we should requeue conditionally - let me know what you think @2uasimojo

pkg/controller/hibernation/hibernation_controller.go
2uasimojo: I think I understand: you're talking about a scenario where the clustersync controller *can't* create the ClusterSync for some reason, so this would thrash indefinitely? That's a good thought; I was only thinking about the brief race condition right after the cluster finishes installing.

However, I think we're probably okay. The "immediate" requeue is still [subject to rate limiting](https://github.com/openshift/hive/blob/5b4fe1dce2677d80c0a8c1237eb909aeaa74b740/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L328) so it'll back off.

Unless there's some compelling reason to do otherwise, I'd like to leave that logic upstream and avoid trying to "outsmart" it here. Thoughts?

suhanime (commented):
/lgtm

PR #2496, opened 2024-10-18
Use Duration Pattern for syncset, machinepool, clusterversion polls

2uasimojo:
Today our backstop for invalid (i.e. not parseable as Duration) values for HiveConfig.Spec.
- SyncSetReapplyInterval
- MachinePoolPollInterval
- ClusterVersionPollInterval

...is to crash the respective controllers.

With this change, we add schema validation, which should cause invalid values to be rejected by the kube API server before they even get into etcd. (We leave the backstops in place for invalid values in pre-existing deployments.)

[HIVE-2635](https://issues.redhat.com//browse/HIVE-2635)

### DIFF
--- a/apis/hive/v1/hiveconfig_types.go
+++ b/apis/hive/v1/hiveconfig_types.go
@@ -64,17 +64,41 @@ type HiveConfigSpec struct {
 	// SyncSetReapplyInterval is a string duration indicating how much time must pass before SyncSet resources
 	// will be reapplied.
 	// The default reapply interval is two hours.
+	// This is a Duration value; see https://pkg.go.dev/time#ParseDuration for accepted formats.
+	// Note: due to discrepancies in validation vs parsing, we use a Pattern instead of `Format=duration`. See
+	// https://bugzilla.redhat.com/show_bug.cgi?id=2050332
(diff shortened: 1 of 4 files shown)


suhanime (commented):
I wonder why the type is set to string instead of metav1.Duration like in [duration metrics](https://github.com/openshift/hive/blob/master/apis/hive/v1/metricsconfig/duration_metrics.go#L17) - can't change that in API now but the validation pattern matches that one so I'm going to assume it works as expected.

/lgtm

2uasimojo:
> I wonder why the type is set to string instead of metav1.Duration like in [duration metrics](https://github.com/openshift/hive/blob/master/apis/hive/v1/metricsconfig/duration_metrics.go#L17) - can't change that in API now but the validation pattern matches that one so I'm going to assume it works as expected.

Yeah, have a look at the bug reports in the comments -- it turns out that the default validation for the Duration type does not match what ParseDuration expects!

Thanks for the review.

suhanime:
> it turns out that the default validation for the Duration type does not match what ParseDuration expects!
Ah! I need to remember to change the type of Duration metrics when we switch them over to metricsConfig.metricsToReport.metricNames[].minimumDuration - thanks!

PR #2498, opened 2024-10-22
HIVE-2397: Nix yaml-patch

2uasimojo:
[HIVE-2397](https://issues.redhat.com//browse/HIVE-2397):

Remove uses of `github.com/krishicks/yaml-patch`, which is unmaintained.

Replace by wrapping `gopkg.in/evanphx/json-patch.v4` with a new function to:
- Convert YAML to JSON
- Convert `PatchEntity` -- see below -- to a patch suitable to...
- ...use json-patch to patch the JSON
- Convert back to YAML

This allowed us to convert most of the code previously using yaml-patch with relatively little disruption to the code flow.

In anticipation of future work, we decided to reuse `PatchEntity`. This is a hive API type originally designed to patch install-config templates for ClusterPool clusters; but it is suitable for this purpose since it represents [RFC6902](https://datatracker.ietf.org/doc/html/rfc6902) patches. However...

[HIVE-2214](https://issues.redhat.com//browse/HIVE-2214):

In this effort, it became necessary to add support for complex (non-string) types to `PatchEntity`. We did this by adding a new member, `ValueJSON`, which accepts a string representing encoded JSON. `Value` can still be used for string types.

Also:
In passing, we noticed that PatchEntity.Op, which is restricted to a discrete set of values per the RFC, was being validated by webhook. We've changed this to use enum schema validation instead.

[HIVE-2397](https://issues.redhat.com//browse/HIVE-2397)
[HIVE-2214](https://issues.redhat.com//browse/HIVE-2214)


pkg/clusterresource/builder.go
```diff
@@ -444,17 +445,17 @@ func (o *Builder) generateInstallConfigSecret() (*corev1.Secret, error) {
...
 			},
-			yamlpatch.Operation{
+			{
 				Op:   "remove",
-				Path: yamlpatch.OpPath("/controlPlane/platform/azure/osImage"),
+				Path: "/controlPlane/platform/azure/osImage",
 			},
 		}
-		modifiedBytes, err := ops.Apply(d)
+		modifiedBytes, err := yamlpatch.ApplyPatches(d, ops)
```
2uasimojo, line 458: Note that we're passing in JSON here (which is technically valid YAML) and getting back YAML.

apis/hive/v1/clusterdeploymentcustomization_types.go
```diff
@@ -54,9 +57,50 @@ type PatchEntity struct {
...
-	// +required
-	Value string `json:"value"`
+	// Value is the *string* value to be used in the operation. For more complex values, use
+	// ValueJSON.
+	// +optional
+	Value string `json:"value,omitempty"`
+	// ValueJSON is a string representing a JSON object to be used in the operation. As such,
+	// internal quotes must be escaped. If nonempty, Value is ignored.
+	// +optional
+	ValueJSON string `json:"valueJSON,omitempty"`
```
2uasimojo, line 67: Should we make this `[]byte` rather than `string`?

pkg/validating-webhooks/hive/v1/clusterdeploymentcustomization_validating_admission_hook.go
```diff
@@ -271,15 +270,15 @@ func validateInstallConfigPatches(path *field.Path, patches []hivev1.PatchEntity
 	return allErrs
 }
 
-func isValidOP(op yamlpatch.Op) bool {
+func isValidOP(op string) bool {
```
2uasimojo, line 273: Is this still necessary now that we're doing `kubebuilder:validation:Enum`?

2uasimojo: Nope. Removing.

pkg/installmanager/installmanager_test.go
```diff
@@ -120,15 +116,15 @@ func TestInstallManager(t *testing.T) {
...
+			failedProvisionUpdate:  ptr.To(int32(0)),
 			expectKubeconfigSecret: true,
 			expectPasswordSecret:   true,
 			expectError:            true,
 		},
 		{
 			name:                          "failed cluster provision log update", // a non-fatal error
 			existing:                      []runtime.Object{testClusterDeployment(), testClusterProvision()},
-			failedProvisionUpdate:         pointer.Int32Ptr(1),
+			failedProvisionUpdate:         ptr.To(int32(1)),
```
2uasimojo, line 127: ntr: latent: replace deprecated

pkg/installmanager/installmanager_test.go
```diff
@@ -84,10 +84,6 @@ func init() {
 	log.SetLevel(log.DebugLevel)
 }
 
-func dummySleep(d time.Duration) {
```
2uasimojo, line 87: ntr: unrelated; inline trivial func

apis/hive/v1/clusterdeploymentcustomization_types.go
```diff
@@ -54,9 +57,50 @@ type PatchEntity struct {
...
+	var val string
+	// Prefer ValueJSON
+	if len(pe.ValueJSON) != 0 {
+		// ValueJSON will be a raw JSON object in the patch; don't quote it.
+		val = string(pe.ValueJSON)
+	} else {
+		// Value will be a JSON string value in the patch; quote it.
+		val = fmt.Sprintf("%q", pe.Value)
+	}
+	// Is this overkill? Should we just assemble the whole thing and let jsonpatch figure it out?
```
dlom, line 82: 20 short lines of logic that probably wont grow, seems fine to me

go.mod
```diff
@@ -130,7 +129,10 @@ require (
 	sigs.k8s.io/cluster-api-provider-gcp v1.7.1-0.20240724153512-c3b8b533143c // indirect
 )
 
-require github.com/gophercloud/utils/v2 v2.0.0-20240701101423-2401526caee5
+require (
```
dlom, line 132: any reason to sequester these in their own require block?

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -836,18 +836,7 @@ func (r *ReconcileClusterPool) patchInstallConfig(clp *hivev1.ClusterPool, cd *h
...
-		newPatch = append(newPatch, yamlpatch.Operation{
-			Op:    yamlpatch.Op(patch.Op),
-			Path:  yamlpatch.OpPath(patch.Path),
-			From:  yamlpatch.OpPath(patch.From),
-			Value: yamlpatch.NewNode(&value),
-		})
-	}
-
-	installConfig, err := newPatch.Apply([]byte(secret.StringData["install-config.yaml"]))
+	installConfig, err := yamlpatch.ApplyPatches([]byte(secret.StringData["install-config.yaml"]), cdc.Spec.InstallConfigPatches)
```
dlom, line 839: love it

pkg/clusterresource/builder.go
dlom: seems like ApplyPatches immediately turns it into JSON anyways

dlom: Just ironic, no change required

apis/hive/v1/clusterdeploymentcustomization_types.go
```diff
@@ -54,9 +57,50 @@ type PatchEntity struct {
...
+	Value string `json:"value,omitempty"`
+	// ValueJSON is a string representing a JSON object to be used in the operation. As such,
+	// internal quotes must be escaped. If nonempty, Value is ignored.
+	// +optional
+	ValueJSON []byte `json:"valueJSON,omitempty"`
+}
+
+// Encode returns a string representation of the RFC6902 patching operation represented by the
+// PatchEntity.
+func (pe *PatchEntity) Encode() string {
```
dlom, line 72: Does this function need the be exposed in our public API? It could just be another short standalone util function in util/yaml.go (`func encodePatchEntity(pe *PatchEntity) string`)

go.mod
2uasimojo: `go mod tidy` did it -- I didn't touch the file by hand.

apis/hive/v1/clusterdeploymentcustomization_types.go
2uasimojo: It probably doesn't *need* to be exposed, no; but I did so because it's simple, doesn't pull in any extra deps, and could facilitate downstreams making use of PatchEntity for their own RFC6902ing.

apis/hive/v1/clusterdeploymentcustomization_types.go
dlom: string is fine I'd say

PR #2502, opened 2024-10-29
Add ClusterDeploymentLabelSelector support

suhanime:
Adds the labelSelector customization for some metrics. This feature can be opted into via HiveConfig, and the metrics, for which this customization is configured, will only be reported if the label query matches the labels on the related clusterdeployment.

To get around cyclic import cycle, the definition and registration of all supported metrics from clusterDeployment, clusterProvision and clusterDeprovision controller have been moved to metrics controller.

[Hive-2285](https://issues.redhat.com/browse/HIVE-2285)


pkg/operator/hive/hive_controller.go
```diff
@@ -513,6 +514,15 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
...
 
+	// Read the metricsConfig section and fetch clusterDeploymentLabelSelectors, error out if it fails validation
+	err = controllerMetrics.GetClusterDeploymentLabelSelectors(hLog, origHiveConfig.Spec.MetricsConfig)
+	if err != nil {
+		hLog.WithError(err).Error("error in metricsConfig")
+		instance.Status.Conditions = util.SetHiveConfigCondition(instance.Status.Conditions, hivev1.HiveReadyCondition, corev1.ConditionFalse, "ErrorInMetricsConfig", err.Error())
+		r.updateHiveConfigStatus(origHiveConfig, instance, hLog, false)
+		return reconcile.Result{}, err
+	}
+
```
suhanime, line 525: Is this location/incorporation correct? My testing hasn't yielded fruitful results - happy to move this to metrics controller and panic if error, and tackle this feature later if time is an issue

pkg/controller/metrics/custom_collectors.go
```diff
@@ -54,7 +54,7 @@ func (cc provisioningUnderwayCollector) Collect(ch chan<- prometheus.Metric) {
 	clusterDeployments := &hivev1.ClusterDeploymentList{}
 	err := cc.client.List(context.Background(), clusterDeployments)
 	if err != nil {
-		log.WithError(err).Error("error listing cluster deployments")
+		ccLog.WithError(err).Error("error listing cluster deployments")
```
suhanime, line 57: We were initializing ccLog but using the one without the field set to metrics controller - corrected this in the entire file

pkg/controller/metrics/metrics.go
```diff
@@ -155,6 +155,115 @@ var (
 	// mapMetricToDurationGauges is a map of optional durationMetrics of type Gauge to their specific duration, if
 	// mentioned
 	mapMetricToDurationGauges map[*prometheus.GaugeVec]time.Duration
+
+	// Metrics reported by ClusterDeployment controller
+
+	MetricInstallJobDuration = prometheus.NewHistogramVec(
```
suhanime, line 161: You might notice, I changed the type of some of the metrics, especially ones declared as Histogram or HistogramOpts, for consistency.

pkg/controller/clusterdeployment/metrics.go
suhanime, line 1: Yeah - couldn't think of any other way to avoid the import cycles. Eventual plan is to move all metrics' definitions to metrics/metrics_to_report, and streamline their types, handle registration and encapsulate observation.

suhanime (commented):
/hold 

must verify hiveoperator codepath works correctly

/assign @2uasimojo

PR #2504, opened 2024-10-31
Don't generate long-lived token for hiveadmission on OpenShift

2uasimojo:
HIVE-2219 / [https://github.com/openshift/hive/pull/2012](#2012) / 7f27836e started injecting a hiveadmission-sa-token Secret into the targetNamespace because k8s, which used to create it automatically, stopped doing so in 1.24. In OpenShift, this Secret was/is unnecessary, but for simplicity we did not differentiate.

Apparently this long-lived token is not desirable in OpenShift, so we're moving its deployment to a non-OpenShift-only code path.

Once this commit is deployed in your OpenShift environment, you should be able to delete the Secret and it will not be restored.

ACM-14645


jstuever (approved):
/lgtm

PR #2506, opened 2024-11-04
Azure: Pass BaseDomainResourceGroupName to destroyer

2uasimojo:
If an Azure spoke was provisioned with a custom
BaseDomainResourceGroupName, DNS records were being leaked when destroying because we weren't passing that information through to the deprovisioner.

This change adds a new API field,
ClusterDeployment.Spec.ClusterMetadata.Platform.Azure.BaseDomainResourceGroupName (and a corresponding passthrough field in the ClusterDeprovision API). When set, we will pass this value through to the deprovisioner.

*It is the user's responsibility to set this field to the same value as in the install-config.yaml.* We will not try to validate or sync the two. This is because we have no reliable way to discover the value from a running cluster.

HIVE-2651


2uasimojo:
I just discovered that the BaseDomainResourceGroupName is not only already [extant](https://github.com/openshift/hive/blob/011916b2246abcb8bbd11a6b590acaff0de000c8/vendor/github.com/openshift/hive/apis/hive/v1/azure/platform.go#L20) in the API, it is [required](https://github.com/openshift/hive/blob/16b131d1a6445060298ae6ffa8dc09b62da3c5ef/pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go#L472-L473). So this PR is doing far more than it needs to do. Will start over...

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -3394,6 +3400,7 @@ platform:
 	}
 
 	for _, test := range tests {
+		if test.name != "Get Azure ResourceGroupName and BaseDomainResourceGroupName from provision metadata" {continue}
```
2uasimojo, line 3403: x

2uasimojo:
Closing in favor of #2507

PR #2512, opened 2024-11-08
HIVE-2656: Various fixes related to hive developer experience (dx)

dlom:
`.gitignore`: add venv so I stop accidentally adding it with `git add hack/`

`hack/*.sh`: use `oc` instead of `kubectl`

`Makefile`: add target for using podman to push to registry (local registry running in podman has no security setup, hence ignore TLS)

/assign @2uasimojo 

card: [HIVE-2656](https://issues.redhat.com//browse/HIVE-2656)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 bin/**
 **/__debug_bin
 **/__pycache__
+**/venv
 
 # Test binary, build with `go test -c`
 *.test
--- a/Makefile
+++ b/Makefile
@@ -317,6 +317,10 @@ buildah-dev-build:
 podman-dev-build:
 	podman build --tag ${IMG} $(GOCACHE_VOL_ARG) -f ./Dockerfile .
 
+.PHONY: podman-dev-push
+podman-dev-push: podman-dev-build
+	podman push --tls-verify=false ${IMG}
+
 # Build and push the dev image with buildah
 .PHONY: buildah-dev-push
 buildah-dev-push: buildah-dev-build
--- a/hack/create-service-account-secrets.sh
+++ b/hack/create-service-account-secrets.sh
@@ -5,7 +5,7 @@
 # - hiveadmission
 # - hive-controllers
 #
-cat <<EOF | kubectl apply -f -
+cat <<EOF | oc apply -f -
 apiVersion: v1
 kind: List
 items:
--- a/hack/hiveadmission-dev-cert.sh
+++ b/hack/hiveadmission-dev-cert.sh
@@ -24,7 +24,7 @@ cat <<EOF | cfssl genkey - | cfssljson -bare server
 }
 EOF
 
-cat <<EOF | kubectl apply -f -
+cat <<EOF | oc apply -f -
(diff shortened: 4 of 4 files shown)


dlom:
/retitle HIVE-2656: Various fixes related to hive developer experience (dx)

Makefile
```diff
@@ -317,6 +317,10 @@ buildah-dev-build:
 podman-dev-build:
 	podman build --tag ${IMG} $(GOCACHE_VOL_ARG) -f ./Dockerfile .
 
+.PHONY: podman-dev-push
+podman-dev-push: podman-dev-build
+	podman push --tls-verify=false ${IMG}
```
2uasimojo, line 322: Definitely +1 to having this target.

I don't think we should turn off TLS by default though. Other options:
- Make a `PODMAN_PUSH_FLAGS` var so you can `make PODMAN_PUSH_FLAGS=--tls-verify=false podman-dev-push` when your registry is "insecure" (or whatever).
- Configure your personal insecure registries per the man page:

> If not specified, TLS verification is used unless the target registry is listed as an insecure registry in containers-registries.conf(5)

.gitignore
```diff
@@ -7,6 +7,7 @@
 bin/**
 **/__debug_bin
 **/__pycache__
+**/venv
```
2uasimojo, line 10: I don't hate this, but it assumes everyone uses the same name for their python venv as you.

You can set up personal gitignores in several ways. For example, my ~/.gitconfig contains:

```
[core]
	excludesfile = /home/efried/.gitignore
```

...and that file contains:

```
.vscode/
*.orig
.sw[a-p]
.*.sw[a-p]
```

(The idea being that not everyone necessarily uses vscode -- though I see we've got that in here on L33 -- or vim.)

hack/create-service-account-secrets.sh
```diff
@@ -5,7 +5,7 @@
 # - hiveadmission
 # - hive-controllers
 #
-cat <<EOF | kubectl apply -f -
+cat <<EOF | oc apply -f -
```
2uasimojo, line 8: -1

This script is explicitly for folks setting up hive on non-openshifts.

And most of the time, people have them (sym)linked to each other anyway.

```
efried@efried-thinkpadp16vgen1:~/go/src/github.com/openshift/hive$ ll -i ~/bin/oc ~/bin/kubectl
2567104 -rwxr-xr-x. 2 efried efried 160441224 Jul 18 16:42 /home/efried/bin/kubectl*
2567104 -rwxr-xr-x. 2 efried efried 160441224 Jul 18 16:42 /home/efried/bin/oc*
```

hack/hiveadmission-dev-cert.sh
```diff
@@ -24,7 +24,7 @@ cat <<EOF | cfssl genkey - | cfssljson -bare server
 }
 EOF
 
-cat <<EOF | kubectl apply -f -
+cat <<EOF | oc apply -f -
```
2uasimojo, line 27: ditto (and below)

PR #2513, opened 2024-11-08
hive-controllers: configurable Resources

2uasimojo:
Prior to this change,
HiveConfig.Spec.ControllersConfig.Default.Resources was ignored. Now it is used to *replace* the default Resources for the hive-controllers Deployment (the one that runs all the controllers not sharded out into StatefulSets).

A quirk of the implementation is that, if you specify this field, you don't get the default values at all, even if you don't specify those values in your customization. For example, with Default.Resources omitted, you get:

```yaml
resources:
  requests:
    cpu: 50m
    memory: 512Mi
```

But if your HiveConfig says:

```yaml
resources:
  requests:
    cpu: 100m
    # memory not specified!
```

you get exactly that; you do *not* get:

```yaml
resources:
  requests:
    cpu: 100m
    memory: 512Mi
```

[HIVE-2650](https://issues.redhat.com//browse/HIVE-2650)

### DIFF
--- a/apis/hive/v1/hiveconfig_types.go
+++ b/apis/hive/v1/hiveconfig_types.go
@@ -605,8 +605,6 @@ type ControllerConfig struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 	// Resources describes the compute resource requirements of the controller container.
-	// This is ONLY for controllers that have been split out into their own pods.
-	// This is ignored for all others.
 	// +optional
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
 }
@@ -680,6 +678,8 @@ type ControllersConfig struct {
 	// default for client burst is 10
 	// default for queue qps is 10
 	// default for queue burst is 100
+	// default resource requests are 50m CPU, 512Mi memory
+	// There are no resource limits by default
 	// +optional
 	Default *ControllerConfig `json:"default,omitempty"`
 	// Controllers contains a list of configurations for different controllers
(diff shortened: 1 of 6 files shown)


apis/hive/v1/hiveconfig_types.go
```diff
@@ -605,8 +605,6 @@ type ControllerConfig struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 	// Resources describes the compute resource requirements of the controller container.
-	// This is ONLY for controllers that have been split out into their own pods.
-	// This is ignored for all others.
```
dlom, line 609: Is this no longer true?

2uasimojo: nope

pkg/util/scheme/scheme.go
```diff
@@ -78,3 +78,33 @@ func init() {
...
+//		nestedVal = *localObj.Some.Nested.Field.With.Pointers
+//	}
+//
+// Note: This will return the result of fallback() on *any* panic of accessor(). So while it's
+// intended to trap nil pointer dereferences in nested fields, it could be (ab)used for other
+// purposes.
+//
+// Why is this in the scheme package? Because generally you want to use it to access a nested
+// field in a scheme-based kube object :shrug:
+func SafeGetNestedField[T any](accessor func() T, fallback T) (ret T) {
```
dlom, line 106: Maybe this can also be used here? https://github.com/openshift/hive/blob/00f84cc1aaf6e0068588013c4b57dcddeaf420af/pkg/controller/utils/clusterdeployment.go#L90

2uasimojo: Yeah; and all the places we have the annoying nil check conditional chain pattern.

2uasimojo:
Ahem, this functionality already existed via HiveConfig.Spec.DeploymentConfig.

PR #2516, opened 2024-11-09
BREAKING: make `PatchEntity.ValueJSON` a `string`

2uasimojo:
`PatchEntity.ValueJSON` was introduced via #2498 / 23fc954a7 as type `[]byte` because that seems to be how other components want to accept raw JSON input. What I didn't realize was that, in the context of a k8s type, that means values in that field must be base64-encoded. This results in a poor UX:

```yaml
      patches:
      - op: add
        path: metadata/labels
        valueJSON: eyJlZnJpZWQub3BlbnNoaWZ0LmlvL2ZvbyI6ICJiYXIifQo=
```

This change switches the field to be of type `string` for better readability:

```yaml
      patches:
      - op: add
        path: metadata/labels
        valueJSON: |
          {"efried.openshift.io/foo": "bar"}
```

We aren't concerned about excluding binary inputs since these patches are always destined for textual yaml manifests.

**THIS IS TECHNICALLY A BREAKING CHANGE**

If a consumer happened to take advantage of `ValueJSON` in the small handful of commits since it was introduced, they will have to manually migrate any affected CDCs. We're considering this risk to be minimal, and the impact negligible.

[HIVE-2214](https://issues.redhat.com//browse/HIVE-2214)


2uasimojo:
/override ci/prow/e2e-pool

Actual test succeeded; flaked in post.

:roll_eyes:

2uasimojo:

PR #2519, opened 2024-11-14
Allow settting vCenter cluster per Machine Pool

lukaspj:
Inspiration for incoming ticket

### DIFF
--- a/apis/hive/v1/vsphere/machinepools.go
+++ b/apis/hive/v1/vsphere/machinepools.go
@@ -8,6 +8,11 @@ type MachinePool struct {
 	// +optional
 	ResourcePool string `json:"resourcePool,omitempty"`
 
+	// Cluster is the name of the cluster virtual machines will be cloned into.
+	// If it is not present, the value in the ClusterDeployment will be used.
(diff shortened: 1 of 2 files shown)


2uasimojo:
- Is it a common/frequent thing to rename the vcenter cluster?
- Assuming the OCP cluster uses a single failure domain (hive does not yet support multiple), is it possible for worker machines to be deployed across different vcenter clusters for the same OCP cluster?

Assuming the answer to both is "no", I think it makes sense in this case to update the Cluster in the ClusterDeployment configuration to match what was done to the vCenter. (This requires working around our webhook, which is not something we want you to do without your support agent's guidance.)

lukaspj:
It’s an exceptional change that is not expected to happen again but evidently, we can’t be sure.

we generally only have one vCenter cluster per vSphere server, the failure domains are handled transparently by our compute team at a vm level

lukaspj:
We got instructions from Red Hat support portal on how to patch the field on our ClusterDeployment and it sounds like we would prefer not changing the behavior here so I closed the PR.

PR #2523, opened 2024-11-22
OSD-26538: Grant hive permission to patch networks.config.openshift.io

joshbranham:
As part of https://issues.redhat.com/browse/OSD-26538, Hive will need RBAC to be able to patch a ClusterDeployment's Network in reaction to a SelectorSyncSet patch.

Do we need any other verbs for the operator to acton this, my guess is no, but let me know if so.

Related to https://github.com/openshift/managed-cluster-config/pull/2281

### DIFF
--- a/config/operator/operator_role.yaml
+++ b/config/operator/operator_role.yaml
@@ -235,3 +235,10 @@ rules:
   - get
   - list
   - watch
+# Required to patch the Network resource to initiate SDN to OVN migration via a SelectorSyncSet
+- apiGroups:
+    - config.openshift.io
+  resources:
+    - networks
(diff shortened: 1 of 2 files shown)


joshbranham:
I actually don't think this approach is going to work based on our requirements so closing this for now

2uasimojo:
I haven't fully grokked the issue, but I think this isn't a valid approach because the RBAC touched by this PR affects hive's permissions on the *hub* (management) cluster. IIUC the network patch operation needs to happen on the *spoke* (managed) cluster. Hive already has `admin` permissions there, so no RBAC changes should be necessary.

PR #2524, opened 2024-11-26
syncset: support templates in `patches`

2uasimojo:
[HIVE-1748](https://issues.redhat.com//browse/HIVE-1748) / #2192 added the ability to flip a switch in a [Selector]SyncSet that causes hive to process golang [text/template](https://pkg.go.dev/text/template)s in leaf values of objects in the [S]SS's `resources` section.

This does a similar thing for JSON patch strings in the `patches` section. For example:

```yaml
kind: SyncSet
metadata:
  name: paramss
spec:
  ...
  enablePatchTemplates: true
  ...

  patches:
  - apiVersion: v1
    kind: ConfigMap
    name: foo
    namespace: default
    patch: |-
      {"metadata": {"annotations": {"test": "{{ fromCDLabel "hive.openshift.io/cluster-platform" }}" }}}
    patchType: merge
```

Note that the data type of `patches[].patch` is `string`, so internal quotes must be escaped somehow (in the above example, it's via `|-`). But the value of the patch _once the template is rendered_ must be valid JSON, e.g. with double quotes surrounding string values. This can make the quoting in the unrendered patch look syntactically odd.

Enables [OSD-26538](https://issues.redhat.com//browse/OSD-26538) by allowing (almost)
[this](https://github.com/openshift/managed-cluster-config/blob/7b43b8354123bd1c7ee529109228be3da811b67b/deploy/osd-sdn-to-ovn-migration/network-operator.patch.yaml#L12-L14) to work.

[HIVE-2669](https://issues.redhat.com//browse/HIVE-2669)

### DIFF
--- a/apis/hive/v1/syncset_types.go
+++ b/apis/hive/v1/syncset_types.go
@@ -238,6 +238,16 @@ type SyncSetCommonSpec struct {
 	// empty string is interpolated if there are no labels, or if the indicated key does not exist.
 	// Note that this only works in values (not e.g. map keys) that are of type string.
 	EnableResourceTemplates bool `json:"enableResourceTemplates,omitempty"`
+
+	// EnablePatchTemplates, if True, causes hive to honor golang text/templates in Patches[].Patch
+	// strings. While the standard syntax is supported, it won't do you a whole lot of good as the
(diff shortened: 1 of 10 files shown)


2uasimojo:

PR #2534, opened 2024-12-19
GCP: Hibernation: Support SSDs

2uasimojo:
It would seem as though, when stopping an instance with SSDs, one is required to specify whether their contents should be preserved or discarded, despite what the documentation [1] says about defaulting that behavior to the latter. This will result in a failure to stop instances, with the following message appearing in the ClusterDeployment's Hibernating condition:

```
Error 400: VM has a Local SSD attached but
an undefined value for `discard-local-ssd`. If using gcloud, please add `--discard-local-ssd=false`
or `--discard-local-ssd=true` to your command.
```

This change adds
ClusterDeployment.Spec.Platform.GCP.DiscardLocalSsdOnHibernate, a _pointer_ to a bool that we will pass through to the GCP API if provided. The default behavior is as before: when the field is not populated, we will not pass the option to the GCP API at all, which continues to be fine for non-SSD-having VMs.

For usability purposes, we allow this field to be edited (most spec fields are immutable, enforced by validating webhook).

[1] https://cloud.google.com/compute/docs/disks/local-ssd#stop_instance

[HIVE-2693](https://issues.redhat.com//browse/HIVE-2693)

### DIFF
--- a/contrib/pkg/createcluster/create.go
+++ b/contrib/pkg/createcluster/create.go
@@ -180,6 +180,15 @@ type Options struct {
 	AzureCloudName                   string
 	AzureResourceGroupName           string
 
+	// GCP
+	// This field is used by the cobra flag...
+	DiscardLocalSsdOnHibernate bool
+	// ...but we need a way to differentiate among `true` (discard), `false` (preserve), and `nil`
+	// (user did not specify, so we won't send the option through the GCP API at all). Cobra gives
+	// us a way to ask, after the fact, whether the field was "Changed"; we'll use that in the code
+	// to set this (lowercase) option so we can trigger the correct behavior when talking to GCP.
+	discardLocalSsdOnHibernate *bool
+
 	// OpenStack
 	OpenStackCloud             string
 	OpenStackExternalNetwork   string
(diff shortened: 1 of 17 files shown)


2uasimojo:
I give it about a 50/50 chance that passing the new option in for non-SSD-havin VMs will splode the GCP API :roll_eyes:

2uasimojo:
Security vuln is new, will be addressed by https://issues.redhat.com/browse/HIVE-2711

2uasimojo:
> I give it about a 50/50 chance that passing the new option in for non-SSD-havin VMs will splode the GCP API 🙄

Awesome, we won this one.

contrib/pkg/createcluster/create.go
```diff
@@ -180,6 +180,11 @@ type Options struct {
 	AzureCloudName                   string
 	AzureResourceGroupName           string
 
+	// GCP
+	DiscardLocalSsdOnHibernate bool
+	// And because we only want to pass this through if explicitly provided by the user...
+	discardLocalSsdOnHibernate *bool
+
```
suhanime, line 187: Uh, why is the private var necessary? Or maybe an explanatory note on how they differ would help here, or at the comment on Line 504

pkg/controller/hibernation/gcp_actuator_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	compute "google.golang.org/api/compute/v1"
+	"k8s.io/utils/ptr"
```
suhanime, line 12: Nit - club this together with the k8s dep below

suhanime (commented):
@2uasimojo Functionally the changes look okay, I do have 1 question about the createcluster changes.

2uasimojo:

PR #2548, opened 2025-01-21
HIVE-2752: PrivateLink: add control-plane network tags to gcp psc

jstuever:
Prior to this change, the gcp privatelink firewall rule was targeting the <infraid>-master network tag. However, in 4.16 the installer changed to using <infraid>-control-plane. This change adds the newer tag to the firewall rule.


2uasimojo (commented):
/lgtm

PR #2549, opened 2025-01-21
Replace go-restful v2.15=>2.16 in go.sum

2uasimojo:
...for CVE-2022-1996, which gets flagged based on seeing 2.15 in go.sum, even though none of the actual code is imported at all.

[ARO-14549](https://issues.redhat.com//browse/ARO-14549)


fahlmant (commented):
/lgtm

PR #2550, opened 2025-01-23
[WIP] MGMT-19741: Add Nutanix support

eliorerz:
/cc @eliorerz


apis/vendor/github.com/openshift/api/machine/v1/types_alibabaprovider.go
2uasimojo, line 1: Self: It's weird that this (and others for platforms hive doesn't support) have been pulled in via this PR. Need to investigate further.

2uasimojo (commented):
Sorry, I didn't make it very far today. I'll keep hammering away at it next week.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -1259,6 +1259,10 @@ func (r *ReconcileMachinePool) createActuator(
 	logger log.FieldLogger,
 ) (Actuator, error) {
 	switch {
+	case cd.Spec.Platform.Nutanix != nil:
```
2uasimojo, line 1262: nit: can we put this `case` in its alpha-ordered spot in the `switch`?

pkg/controller/machinepool/nutanixactuator.go
```diff
@@ -0,0 +1,118 @@
...
+func NewNutanixActuator(client client.Client, scheme *runtime.Scheme, logger log.FieldLogger) (*NutanixActuator, error) {
+	//nutanixClient, err := nutanixclient.NewClient()
+	//if err != nil {
+	//	return nil, errors.Wrap(err, "failed to create Nutanix session")
+	//}
+
+	actuator := &NutanixActuator{
+		client: client,
+		scheme: scheme,
+		//nutanixClient: nutanixClient,
```
2uasimojo, line 39: Let's talk about whether you're actually going to need this. In general, the major cloud bit -- managing VMs -- is done way down the line by MAO, using the cloud creds on the spoke, operating on the MachineSets we generate. Exceptions exist: for example, in [GCP](https://github.com/openshift/hive/blob/e4b80d0c1e6a97443a18594ff42fbf85651a6171/pkg/controller/machinepool/gcpactuator.go#L275) and [IBMCloud](https://github.com/openshift/hive/blob/cb1d134550919670e4307c671e52fb923d0d1925/pkg/controller/machinepool/ibmcloudactuator.go#L82) we need to discover zone names. But by default I would say YAGNI.

pkg/controller/machinepool/nutanixactuator.go
```diff
@@ -0,0 +1,118 @@
...
+
+	installerMachineSets, err := installvsphere.MachineSets(
+		cd.Spec.ClusterMetadata.InfraID,
+		ic,
+		computePool,
+		// With FailureDomains[0].Topology.Template set, this param is ignored.
+		"HIVE_BUG!",
+		workerRole,
+		workerUserDataName,
+	)
```
2uasimojo, line 111: copypasta, looks like this still needs some work.

When we get down to it though, I think we should take the nutanix installer team up on their offer to help out by reviewing this func.

The general principle is that our code here needs to dummy up the inputs to the platform-specific MachineSets generator. We can not rely on having access to the original install-config.yaml for the cluster, which is sometimes annoying and awkward, forcing us to either assume certain values or duplicate them in the (hive) MachinePool CRD. Since we're making this actuator from scratch, we should take the opportunity to evaluate those things carefully; else we can end up painting ourselves into a corner where e.g. we have to support bogus/deprecated/obsolete fields "forever".

pkg/install/generate.go
2uasimojo, line 1: This file will want code for the uninstaller as well. Examples for other platforms are referenced in a `switch` starting around L519. And that'll correspond to a `hiveutil deprovision nutanix` subcommand, with code under contrib/pkg/deprovision, plugged into deprovision.go.

contrib/pkg/createcluster/create.go
```diff
@@ -829,6 +840,72 @@ func (o *Options) GenerateObjects() ([]runtime.Object, error) {
...
 		builder.CloudBuilder = ibmCloudProvider
+
+	case cloudNutanix:
+		// Validate and extract required credentials and configuration
+		nutanixUsername := os.Getenv(constants.NutanixUsernameEnvVar)
+		if nutanixUsername == "" {
+			return nil, fmt.Errorf("no %s env var set, cannot proceed", constants.NutanixUsernameEnvVar)
+		}
+
+		nutanixPassword := os.Getenv(constants.NutanixPasswordEnvVar)
```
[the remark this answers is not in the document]
2uasimojo: Yeah, it is ✓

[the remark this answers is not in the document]
eliorerz: I'm not familiar with one but we can store that in a JSON or yaml formats. It will be easy to read and to maintain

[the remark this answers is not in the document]
2uasimojo: > I'm not asking you to invent a config file format, but does nutanix already have one that it supports?

Let me rephrase: Please *don't* invent a config file format :P

But if there's one nutanix already accepts, let's use it.

eliorerz:
Implemented on https://github.com/openshift/hive/pull/2573

PR #2551, opened 2025-01-27
automated development environment for hive

msu8:
Scripts for setting up a development environment for hive. Everything is automated by running a makefile "Makefile.environment".
The setup is rootless and OS-independent as it only uses binaries that are downloaded into a .tmp dir in the repo's root.

The process is described in the developing.md doc, with additional information on how to set up VScode for debugging.

It currently only works on amd64(x86) arches, but the idea would be to try to make it run natively on other arches, like arm/aarch64.


abraverm:
ping

PR #2552, opened 2025-01-28
Eliminate openshift/custom-resource-status dependency

2uasimojo:
openshift/custom-resource-status has been bit-rotting, and its functionality is available upstream in apimachinery. Replace.

NOTE: This entails swapping out the type of
ClusterDeploymentCustomization.Status.Conditions from c-r-s to apimachinery/.../meta/v1. While these types are slightly different, we were only using the common subset of their fields, and those fields *should* marshal/unmarshal the same.

[HIVE-2754](https://issues.redhat.com//browse/HIVE-2754)


2uasimojo:
/hold cancel

QE signed off ✓

/assign @suhanime

PR #2556, opened 2025-02-04
HIVE-2764: Upstream builds for operatorhub

celebdor:
This adds a Makefile target that relies on ubi8, ubi9 and centos stream to provide an upstream image that is fit for OperatorHub.


2uasimojo:
/lgtm

Successfully ran e2e locally on the unentitled image built with this PR.

PR #2557, opened 2025-02-05
Document metrics FAQ

suhanime:
We get frequent queries about which hive metric someone can use to trace a cluster that is causing an alert to fire. So this commit adds a small addition to metrics documentation to document our stance regarding the same.

This was added in the hive metrics document because metrics users are more likely to read there, instead of general hive FAQs

/assign @2uasimojo


2uasimojo (commented):
:heart:

PR #2572, opened 2025-02-17
feat(ci): Build container images without python deps

celebdor:
This change allows building the container images without the python dependencies that are getting harder to get by.
It keeps the pre-existing version2.py so that it can be imported from the python bundle generation code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,9 @@ endif
 # we don't tag versions. Override using the same versioning we apply to OperatorHub builds:
 # v{major}.{minor}.{commitcount}-{sha}
 # Note that building against a local commit may result in {major}.{minor} being rendered as
-# `UnknownBranch`. However, the {commitcount} and {sha} should still be accurate.
-SOURCE_GIT_TAG := $(shell export HOME=$(HOME); python3 -m ensurepip >&2; python3 -mpip --no-cache install --user gitpython pyyaml >&2; hack/version2.py)
+# 0.0.x-y if it is an `UnknownBranch`. However, the {commitcount} and {sha}
+# should still be accurate.
+SOURCE_GIT_TAG := $(shell hack/version2.sh)
 
 BINDATA_INPUTS :=./config/sharded_controllers/... ./config/hiveadmission/... ./config/controllers/... ./config/rbac/... ./config/configmaps/...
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -76,8 +76,9 @@ endif
 # we don't tag versions. Override using the same versioning we apply to OperatorHub builds:
 # v{major}.{minor}.{commitcount}-{sha}
 # Note that building against a local commit may result in {major}.{minor} being rendered as
-# `UnknownBranch`. However, the {commitcount} and {sha} should still be accurate.
-SOURCE_GIT_TAG := $(shell export HOME=$(HOME); python3 -m ensurepip >&2; python3 -mpip --no-cache install --user gitpython pyyaml >&2; hack/version2.py)
+# 0.0.x-y if it is an `UnknownBranch`. However, the {commitcount} and {sha}
+# should still be accurate.
+SOURCE_GIT_TAG := $(shell hack/version2.sh)
```
2uasimojo, line 81: Love this. It was super annoying waiting for pip every time the Makefile loaded up.

However, this may have been the only thing ensuring that `pyyaml` was installed when running [generate-saas-template.py](https://github.com/openshift/hive/blob/00b13a6365cc4107aababc93067b2794d92e0339/Makefile#L243). I'm going to look into porting that to bash/yq as well, but in the meantime, can we add the `ensurepip` and `pyyaml` business inside that rule? (At least then it will only run when that rule is invoked.)

hack/version2.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+declare -r UNKNOWN_BRANCH_PREFIX="0.0"
+declare -r MASTER_BRANCH_PREFIX="1.2"
+
+function prefix_from_branch() {
+	declare -r branch_name="$1"
+
+	if [[ "$branch_name" =~ ^([^/]+/)?mce-([[:digit:]]+.[[:digit:]]+) ]]; then
```
2uasimojo, line 8: I've [asked](https://redhat-internal.slack.com/archives/C012L9Y9C79/p1739828571006479) and ACM/MCE don't seem to be using bundle-gen anymore. Given that, and the fact that we don't build OperatorHub bundles for MCE branches anymore, we should be able to eliminate this condition.

PR #2582, opened 2025-02-24
HIVE-2781: nix terminateWhenFilesChange()

2uasimojo:
We used to mount credentials Secrets on deprovision pods. When we were doing that, if the Secret changed, the data in the mounted directory would get updated dynamically by kube. We had code in place to terminate the deprovision command in that scenario, since it would presumably fail 401 the next time it needed to create a cloud client.

Via [HIVE-2021](https://issues.redhat.com//browse/HIVE-2021) / #1874 / 0f4b1de7 we stopped with the mounting, and started loading the Secrets explicitly from the deprovision pods. Although we're still laying the data down in the same kind of directory structure, it is no longer possible for the values to change dynamically. This renders the aforementioned code obsolete.

So we're ripping it out.

[HIVE-2781](https://issues.redhat.com//browse/HIVE-2781)


dlom:
-66

/lgtm

2uasimojo:
nice, this also nixes a couple of vendored files

@dlom please restamp.

dlom:
-376

/lgtm

PR #2583, opened 2025-02-26
HIVE-2785: Remove obsolete version check

2uasimojo:
Remove logic associated with checking whether the spoke cluster's version is >=4.4.8, since that's way out of service.


2uasimojo:
/override ci/prow/security

Will be addressed via HIVE-2786.

PR #2541, opened 2025-01-13
HIVE-2270: proper e2e support for vSphere

dlom:
xref: [HIVE-2270](https://issues.redhat.com//browse/HIVE-2270)

corresponding CI PR: https://github.com/openshift/release/pull/58886

PR #2590, opened 2025-02-28
Update vSphere e2e parameter configuration

huangmingxia:
1. Set managed DNS to false.  
   Currently, managed DNS only supports AWS, GCP, and Azure clouds.  

2. This function `get_vips()` uses `echo` to obtain the value, and the `echo` output includes a [timestamp](https://github.com/openshift/hive/blob/master/hack/e2e-common.sh#L3-L4).

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -223,6 +223,7 @@ case "${CLOUD}" in
 	;;
 "vsphere")
   BASE_DOMAIN="${BASE_DOMAIN:-vmc.devcluster.openshift.com}"
+  USE_MANAGED_DNS=false
   if [ -z "$NETWORK_NAME" ]; then
     echo "Variable 'NETWORK_NAME' not set."
     exit 1
@@ -334,5 +335,5 @@ function get_vips() {
   fi
 
   vip=$(sed -n "${idx}p" "$vips")
-  echo "$vip"
+  printf "%s" "$vip"
 }


hack/e2e-common.sh
```diff
@@ -334,5 +335,5 @@ function get_vips() {
   fi
 
   vip=$(sed -n "${idx}p" "$vips")
-  echo "$vip"
+  printf "%s" "$vip"
```
2uasimojo, line 338: nit: the format string is redundant
```suggestion
  printf "$vip"
```

...but for that matter, so is the temporary variable. We could just consolidate with the previous line:
```suggestion
  sed -n "${idx}p" "$vips"
```

2uasimojo (commented):
I believe Mark is still working on this stuff.

dlom:
This works for me.  As soon as vsphere CI is working again we can merge this (I'll hold a local copy in my branch)

huangmingxia:
> This works for me. As soon as vsphere CI is working again we can merge this (I'll hold a local copy in my branch)

Sounds good!

2uasimojo:
@dlom have you already incorporated this into #2541?

dlom:
Yes, this PR can be closed.  Thank you @huangmingxia

huangmingxia:
Hi @dlom  My pleasure! 
Sorry it took me so long to reply! Your message got buried in my overflowing inbox until I cleaned it up just now :)

PR #2594, opened 2025-03-03
HIVE-2791: MachinePool/OpenStack: Remove obsolete version check

2uasimojo:
We were checking for OCP versions >=4.7.x to determine whether zero minimum replicas are allowed for autoscaling. That version is long out of service. Remove the obsolete code.

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/blang/semver/v4"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -1374,38 +1373,13 @@ func getClusterVersion(cd *hivev1.ClusterDeployment) (string, error) {
 }
 
 func platformAllowsZeroAutoscalingMinReplicas(cd *hivev1.ClusterDeployment) bool {
-	// Since 4.5, AWS, Azure, and GCP allow zero-sized minReplicas for autoscaling
-	if cd.Spec.Platform.AWS != nil || cd.Spec.Platform.Azure != nil || cd.Spec.Platform.GCP != nil {
+	// Zero-sized minReplicas for autoscaling are allowed since OCP:
+	// - 4.5 for AWS, Azure, and GCP
+	// - 4.7 for OpenStack
```
2uasimojo, line 1378: musing: other platforms may have been enabled since this was last touched, may be worth looking into.

PR #2596, opened 2025-03-06
HIVE-2475: Add comment for machinepool override annotation

suhanime:


suhanime:
@djoshy Please review

pkg/constants/constants.go
```diff
@@ -525,6 +525,7 @@ const (
 	// hook that normally forbids modifying its Spec.Platform. By setting this annotation, you are taking responsibility
 	// for forcing rollout of such changes on the target cluster -- e.g. by deleting the Machines -- as the machine
 	// config operator will not do so.
+	// Not to be used when custom boot images are in play, can result in thrashing.
```
djoshy, line 528:
```suggestion
	// Not to be used when boot images updates are in play, can result in thrashing.
```
Minor rephrase as "custom boot images" is another MCO feature :sweat_smile:

PR #2611, opened 2025-03-17
HIVE-2803: Revendor installer Q12025

eliorerz:
This PR update go packages needed for adding Nutanix provider

/cc @2uasimojo


2uasimojo:
/lgtm

...assuming CI is happy.

2uasimojo:
/override ci/prow/e2e-vsphere

This test isn't working yet in general.

The security fail is spurious -- I believe it has to do with the fact that the installer tag we're using sorts lower than the release branch containing the fix, even though it's chronologically newer by ~3.5y. But before we just override it or add a .snyk ignore (which _all_ installer downstreams would have to do as well), let's chat with @patrickdillon and see if we would rather use a different tag naming strategy.

2uasimojo:
/lgtm

though it looks like e2e-pool may time out again...

2uasimojo:
/override "Red Hat Konflux"

That silly thing should never have run in the first place -- it was triggered by me messing around with retesting the _actual_ konflux job.

e2e-vsphere shouldn't be required/blocking, but let's see...

2uasimojo:
/override ci/prow/e2e-vsphere

I may have to mash the merge button if that konflux thing is still in the way.

2uasimojo:
Eh, that spurious konflux job seems to have wedged this PR pretty good. I'm just gonna force merge.

PR #2612, opened 2025-03-17
HIVE-2804: Scrub username and password fields on metadata.json 

dlom:
xref: [HIVE-2804](https://issues.redhat.com//browse/HIVE-2804)

/assign @2uasimojo

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -57,6 +57,7 @@ import (
 	installertypesovirt "github.com/openshift/installer/pkg/types/ovirt"
 	installertypesvsphere "github.com/openshift/installer/pkg/types/vsphere"
 
+	jsoniter "github.com/json-iterator/go"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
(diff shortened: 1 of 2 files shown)


pkg/installmanager/installmanager.go
```diff
@@ -432,6 +433,12 @@ func (m *InstallManager) Run() error {
 		m.log.WithError(err).Error("error reading cluster metadata")
 		return errors.Wrap(err, "error reading cluster metadata")
 	}
+	scrubbedMetadataBytes, err := scrubMetadataJSON(metadataBytes)
+	if err != nil {
+		m.log.WithError(err).Error("error cleaning up metadata")
+		scrubbedMetadataBytes = []byte("{}")
```
2uasimojo, line 439: Oh, I just noticed we're not actually erroring out here. I think we want to. That's the lesser evil vs. laying down an empty metadata.json, which will result in problems later in the flow from the methods that rely on it.

2uasimojo:
/test all

prow seems ill today

dlom:
> @2uasimojo: No presubmit jobs available for openshift/hive@master

something seems off

PR #2613, opened 2025-03-17
HIVE-2805: Ignore a couple of snyk vulns

2uasimojo:
Add a couple of ignores to .snyk for things that rely on difficult backports of transitive dependencies.


2uasimojo:
Let's see if we can restore prow by turning it off and on again...

2uasimojo:
Skipping CI, as this didn't change any code.

PR #2623, opened 2025-03-19
DNM: Does Draft PR run [periodic-]images?

2uasimojo:
[DPTP-4372](https://issues.redhat.com//browse/DPTP-4372)


2uasimojo:
From the card: we're not actually *running* those jobs here. For some reason prow puts them in as expected (but not the other required jobs??) but doesn't actually trigger them.

PR #2624, opened 2025-03-20
DNM: Test prow ignores

2uasimojo:
- [DPTP-4372](https://issues.redhat.com//browse/DPTP-4372): While this is in draft state, we shouldn't run -- or even "expect" -- any prow jobs.
- [HIVE-2808](https://issues.redhat.com//browse/HIVE-2808): Main CI jobs shouldn't run for .tekton/-only changes.

### DIFF
--- a/.tekton/hive-pull-request.yaml
+++ b/.tekton/hive-pull-request.yaml
@@ -1,3 +1,4 @@
+test: true
 apiVersion: tekton.dev/v1
 kind: PipelineRun
 metadata:


2uasimojo:
Hm, those "Expected" jobs are still showing up. 0 for 1.

On to the next; un-drafting...

2uasimojo:
Okay, declaring success on HIVE-2808.

No joy on the other. But as long as we know it's not actually burning resources, I guess it's okay.

PR #2625, opened 2025-03-21
DPTP-4372: Test prow ignores again

2uasimojo:
It has now been long enough that branch protection rules should have been refreshed. Testing that `images` and `periodic-images` don't show up as "Expected" anymore for draft PRs.


2uasimojo:
Score, that seems to have worked.

PR #2639, opened 2025-03-28
HIVE-2819: Lift upgradeable condition from CVO to cluster deployment label

AlexVulaj:
This PR lifts the `Upgradeable` status from the ClusterVersion into a new `hive.openshift.io/minor-version-upgrade-unavailable` label on the ClusterDeployment.

Higher level tools can consume this message to warn users about minor version upgrades that CVO would reject with a reason.

### DIFF
--- a/pkg/controller/clusterversion/clusterversion_controller.go
+++ b/pkg/controller/clusterversion/clusterversion_controller.go
@@ -1,6 +1,7 @@
 package clusterversion
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"math/rand"
@@ -209,6 +210,22 @@ func (r *ReconcileClusterVersion) updateClusterVersionLabels(cd *hivev1.ClusterD
 		delete(cd.Labels, constants.VersionMajorMinorLabel)
 		delete(cd.Labels, constants.VersionMajorMinorPatchLabel)
 		changed = changed || origLen != len(cd.Labels)
+	}
+
+	upgradeableCondition := ""
+	for _, condition := range clusterVersion.Status.Conditions {
+		if condition.Type == openshiftapiv1.OperatorUpgradeable {
+			if condition.Status != openshiftapiv1.ConditionTrue {
+				upgradeableCondition = cmp.Or(condition.Message, fmt.Sprintf("%s: %s", condition.Type, condition.Status))
+			}
+			break
+		}
+	}
+	changed = changed || upgradeableCondition != cd.Labels[constants.MinorVersionUpgradeUnavailable]
+	if upgradeableCondition == "" {
+		delete(cd.Labels, constants.MinorVersionUpgradeUnavailable)
+	} else {
+		cd.Labels[constants.MinorVersionUpgradeUnavailable] = upgradeableCondition
 	}
 
 	if !changed {
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -357,6 +357,10 @@ const (
 	// in the form "[MAJOR].[MINOR].[PATCH]".
 	VersionMajorMinorPatchLabel = "hive.openshift.io/version-major-minor-patch"
 
+	// MinorVersionUpgradeUnavailable is a label applied to ClusterDeployments to show concerns about upgrades to the
+	// next minor version. i.e 4.y -> 4.(y+1)
+	MinorVersionUpgradeUnavailable = "hive.openshift.io/minor-version-upgrade-unavailable"
+
 	// OvirtCredentialsDir is the directory containing Ovirt credentials files.
 	OvirtCredentialsDir = "/.ovirt"
 
--- a/pkg/controller/clusterversion/clusterversion_controller_test.go
+++ b/pkg/controller/clusterversion/clusterversion_controller_test.go
@@ -45,11 +45,12 @@ func init() {
 func TestClusterVersionReconcile(t *testing.T) {
 
 	tests := []struct {
-		name         string
-		existing     []runtime.Object
-		noRemoteCall bool
-		expectError  bool
-		validate     func(*testing.T, *hivev1.ClusterDeployment)
+		name                 string
+		existing             []runtime.Object
+		clusterVersionStatus configv1.ClusterVersionStatus
+		noRemoteCall         bool
+		expectError          bool
+		validate             func(*testing.T, *hivev1.ClusterDeployment)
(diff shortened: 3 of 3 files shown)


wking (approved):
I'm not a Hive approver, but looks good to me; thanks!

/lgtm

wking:
Colon delimiter to help the Jira-linking bots:

/retitle HIVE-2819: Lift upgradeable condition from CVO to cluster deployment label

2uasimojo (commented):
Please add some unit testing for this. We don't need much. Some helpful references:
- How I added UT in a [previous PR](https://github.com/openshift/hive/pull/2084/files#diff-d62226d287da17341c857e45e5a322048193bd1b9aeeda78616c82247dec9ca6R76) in this area.
- The [dummy ClusterVersion object](https://github.com/openshift/hive/blob/26a0b814475020cb6e2e4e46ccdf421bef2c1b1a/pkg/controller/clusterversion/clusterversion_controller_test.go#L196) you'll have to enhance with some status conditions to make your test work. (You may have to do some refactoring to get both positive and negative code paths covered, since currently this thing is hardcoded and buried inside another object.)

I also want to point out that this controller is by default only [Watch()ing ClusterDeployment](https://github.com/openshift/hive/blob/08e49876b54fb77879d4362e4339df68318f4d93/pkg/controller/clusterversion/clusterversion_controller.go#L85) (on the hub), which means there may be a nontrivial delay between when the spoke ClusterVersion object changes and when the controller runs to update the hub CD labels. I believe @hlipsig struggled with this for ARO, resulting in https://issues.redhat.com//browse/HIVE-2619, which you may end up finding handy when you go to use this thing.

pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -211,6 +212,19 @@ func (r *ReconcileClusterVersion) updateClusterVersionLabels(cd *hivev1.ClusterD
 		changed = changed || origLen != len(cd.Labels)
 	}
 
+	upgradeableCondition := ""
+	for _, condition := range clusterVersion.Status.Conditions {
+		if condition.Type == openshiftapiv1.OperatorUpgradeable && condition.Status != openshiftapiv1.ConditionTrue {
+			upgradeableCondition = cmp.Or(condition.Message, fmt.Sprintf("%s: %s", condition.Type, condition.Status))
```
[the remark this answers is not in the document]
wking: The CVO does have [some throttling][1], but with a lower bound of 15s in the `minOnFailedPreconditions` case, it may be so little you don't care ;).  There's also an `ignoreThrottlePeriod` knob in the CVO from openshift/cluster-version-operator@b6b7345d6675cac0a1f3ddd8b1e24eba8d94f6e5, but at the moment that's still only used for the sync-on-CVO-container-exit, which should be rare.

That's not a lot of hard numbers, and without a full survey of ClusterOperator maintainers, it's hard to claim exhaustive coverage.  But anecdotally, I'm not aware of folks including mutable, high-churn strings in their messages, and I could see folks filing bugs against churny components to ask them to calm down.  I could also see something defensive in Hive about "we realize this could churn, and are not interested in trying to keep up", and adding some kind of per-cluster throttling/back-off to avoid the thrashing you're concerned about.

[1]: https://github.com/openshift/cluster-version-operator/blob/ddb11b36c14413496af944c22fdadc599766fd51/pkg/cvo/upgradeable.go#L51-L65C68

pkg/controller/clusterversion/clusterversion_controller.go
```diff
@@ -1,6 +1,7 @@
 package clusterversion
```
wking, line 1: I don't want the main comment-stream to get too noisy with the latency issue, so pulling [this][1]:

> I also want to point out that this controller is by default only [Watch()ing ClusterDeployment](https://github.com/openshift/hive/blob/08e49876b54fb77879d4362e4339df68318f4d93/pkg/controller/clusterversion/clusterversion_controller.go#L85) (on the hub), which means there may be a nontrivial delay between when the spoke ClusterVersion object changes and when the controller runs to update the hub CD labels.

out tot this random, unrelated line of code to give it a dedicated thread.

I'm personally not concerned about the latency here, because most of the issues that `Upgradeable` complains about are long-running, slow-changing issues (like "[you're on SDN; migrate to OVN to access 4.17][2]").  So ~hours stale is likely to be still accurate in many cases.

And when we miss with a false negative (stale ClusterDeployment data said the update was ok, but turned out ClusterVersion had moved to be `Upgradeable=False`), it's not terrible.  The user could request an update, and the cluster-version operator would reject the request with whatever the `Upgradeable=False` message was.  So the cluster is still safe, it's just a bit more of a rug-pull UX.  Having fresher data would improve the UX, but would not increase cluster safety.

When we miss with a false positive (stale ClusterDeployment data said the update was blocked, but turned out ClusterVersion had moved to be `Upgradeable=True` or unset the `Upgradeable` condition), it's not terrible either.  The user's access to the next 4.(y+1) is delayed by an hour or two until the ClusterDeployment catches up.  But it's just a feature update, patch updates pulling in bugfixes and CVEs would not be impacted.  And users who want to avoid any risk of delay could just get their `Upgradeable` ducks lined up more than an hour before they were hoping to launch the update (hopefully nobody is actually trying to cut it  that close).

The gap here is flappy issues like `PoolUpdating`, which I dropped in 4.19 (openshift/machine-config-operator#4760), and I'm happy to backport that (and fixes to any other flappy `Upgradeable` conditions, although I can't think of any offhand) to older 4.y, if stale ClusterDeployment UX impacts turn out to be an issue.

[1]: https://github.com/openshift/hive/pull/2639#pullrequestreview-2726750171
[2]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/release_notes/ocp-4-16-release-notes#ocp-4-16-networking-openshift-sdn-upgrade-block_release-notes

2uasimojo: Cool. Note that a dummy CD update (like an annotation) could be used to force a resync. That's a thing you can't do through OCM today (that I know of), but would be trivial to implement.

2uasimojo:
The tests look great, thanks @AlexVulaj.

/lgtm
/retest hive-on-pull-request

2uasimojo:
Bah, konflux seems down. Since we're not actually using it yet...

PR #2661, opened 2025-04-17
fix(HIVE-2838): privatelink-controller: paginate retrieved vpc endpoints

typeid:
**Issue:**
https://issues.redhat.com/browse/HIVE-2838

When there are many VPC endpoints associated with a cluster, AWS may paginate the response to `DescribeVpcEndpoints`. In some cases, AWS returns an empty first page (no `VpcEndpoints`) with a NextToken, which is highly unintuitive.

Example returned data:
```
{
  NextToken: "eyJ2IjoiM<snip>="
}
```

Hive currently does not handle pagination in this code path. As a result:
- we falsely assume no VPC endpoint exists
- we create a new endpoint unnecessarily

This results in leaked VPC endpoints. We believe the reason for having so many VPCE in the first place is another bug ([HIVE-2837](https://issues.redhat.com/browse/HIVE-2837)), where rejected connections are leaked by hive.

**Proposed fix:**
This PR adds proper pagination support when describing VPC endpoints.

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller.go
@@ -855,22 +855,37 @@ func (r *ReconcileAWSPrivateLink) reconcileVPCEndpoint(awsClient *awsClient,
 	endpointLog := logger.WithField("tag:key", aws.StringValue(tag.Name)).WithField("tag:value", aws.StringValueSlice(tag.Values))
 
 	var vpcEndpoint *ec2.VpcEndpoint
-	resp, err := awsClient.hub.DescribeVpcEndpoints(&ec2.DescribeVpcEndpointsInput{
+
+	input := &ec2.DescribeVpcEndpointsInput{
 		Filters: []*ec2.Filter{tag},
-	})
-	if err != nil {
-		endpointLog.WithError(err).Error("error getting VPC Endpoint")
-		return modified, nil, err
 	}
-	if len(resp.VpcEndpoints) == 0 {
+
+	var allVpcEndpoints []*ec2.VpcEndpoint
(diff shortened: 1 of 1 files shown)


2uasimojo:
Thanks for the fix! That's crazy that the AWS API will send back an empty first page. It's probably a bug on their end. But it makes sense to handle it as you have.

I would like to see a unit test or two that exercises this quirk.

2uasimojo:
> I would like to see a unit test or two that exercises this quirk.

We'll do this in a followon for the sake of expediency.

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -855,22 +855,37 @@ func (r *ReconcileAWSPrivateLink) reconcileVPCEndpoint(awsClient *awsClient,
...
+
+	var allVpcEndpoints []*ec2.VpcEndpoint
+	for {
+		resp, err := awsClient.hub.DescribeVpcEndpoints(input)
+		if err != nil {
+			endpointLog.WithError(err).Error("error getting VPC Endpoint")
+			return modified, nil, err
+		}
+
+		allVpcEndpoints = append(allVpcEndpoints, resp.VpcEndpoints...)
```
2uasimojo, line 871: For the current fix, I'm not sure that we need to keep looping until we get *all* the VPCEs. We only care about one of them, so we could `break` as soon as we get a nonempty page. Otherwise we're potentially incurring a bunch of extra network calls & bytes. (Does throttling get applied to page requests??)

However, this may also be where cleanup would need to happen. Again, since we only want one VPCE for this filter, we know that we want to scrub `allVpcEndpoints[1:]`.

PR #2663, opened 2025-04-17
HIVE-2837: PrivateLink: Clean up zombie VPCEndpoints

2uasimojo:
VPCEndpoints can end up in non-"available" states (e.g. "rejected"). We'll try to create a new one, but we weren't previously scrubbing these stale entries, so they were leaked.

Unit tests also cover the change introduced via HIVE-2838 / #2661.


typeid (commented):
Thanks for the quick cleanup implementation @2uasimojo!

One major thing to note is that we _not only_ found endpoints leaked in the `rejected` state, but also in the `available` state. As an example, the issue described in https://github.com/openshift/hive/pull/2661: the leaked endpoints would still be connected to the service, thus they would be in the 'available' state. 

To define whether a VPCE in the "accepted" state was actually in use during the incident, we checked the respective HostedZones for clusters to find the VPCE referenced there. We considered every VPCE that isn't part of the HostedZone as 'leaked'. 

While we did fix the main cause for the leaks with https://github.com/openshift/hive/pull/2661 but I suspect we still have a few more leaks. Therefore, it might make sense to build in something like the above logic to also handle leaked VPCE in good states. 

One more example I can think of in which we would leak VPCE in a good state in [the current creation logic](https://github.com/openshift/hive/blob/44c3beacca130510a33938be20e2fb3496a4c965/pkg/controller/awsprivatelink/awsprivatelink_controller.go#L863-L886): 
1) Initial reconciliation loop describes VPCE, finds none.
2) VPCE creation is called and runs into a timeout (e.g. consider AWS slowness resulting in the VPCE creation only succeeding after 3 minutes - there's no timing guarantees from AWS here).
3) Second reconciliation loop describes VPCE again a minute after the initial reconciliation, no VPCE is found.
4) Second reconciliation loop creates another VPCE
...

We could either handle all these eventual consistency cases rigorously, but it might get quite complicated as the logic is currently blocking for the whole privatelink reconciler to my understanding. The easier choice would be to also clean up the leaked VPCE in a good state as mentioned above.

2uasimojo:
@typeid thanks for the analysis. If I've done it right, the logic in this PR will clean up all n-1 VPCEs regardless of their State. I've [tweaked the unit test](https://github.com/openshift/hive/compare/781db6c8a0319873d56a63b4eaf270d125ac4121..f7c3bed1c9b90e3c5cef3346e32849c3a332dbb2) to prove that. Please let me know if I've missed your concern.

I believe we're still operating on the assumption that VPCEs are returned to us in a deterministic/consistent order (we always pick the first "good" one). I don't know how much effort it's worth going through to hedge against that not being true. Thoughts?

typeid:
@2uasimojo Thanks clarifying and adding a unit test for it. 

I'm a bit concerned about picking the first "good one" vs the actual good one in use (referenced in the HostedZone). What happens if:
- we try to delete the VPCE currently referenced in the HostedZone? Will AWS even allow that, or does the dns record need to be deleted first?
- we end up deleting the VPCE currently referenced in the HostedZone. Will the reconciliation logic update the HostedZone's dns record to contain the first found VPCE? I'm assuming it would, as this sounds like a fairly normal reconciliation case. 

Note that we also found another bug where clusters had up to 8 total HostedZones instead of 1, which is another bug that we opened. Could this end up biting us here?

2uasimojo:
> * we try to delete the VPCE currently referenced in the HostedZone? Will AWS even allow that, or does the dns record need to be deleted first?

I don't know this answer. Maybe @jstuever does. The [docs](https://api.github.com/repos/kubernetes-sigs/kustomize/releases) don't say (or if they do it's in a way I don't understand). However, our input to the HZ is only the VPC ID, not the VPCE in any way (more on this below), which would lead me to believe they're not that tightly coupled.

> * we end up deleting the VPCE currently referenced in the HostedZone. Will the reconciliation logic update the HostedZone's dns record to contain the first found VPCE? I'm assuming it would, as this sounds like a fairly normal reconciliation case.

Yes.

> Note that we also found another bug where clusters had up to 8 total HostedZones instead of 1, which is another bug that we opened. Could this end up biting us here?

Most likely that's an artifact of the original HIVE-2838 bug.
- Look for existing VPCE, don't find it due to first-page-empty issue.
- Create (extra) VPCE.
- Carry on reconciling to create the HZ for it.

Note that this path would only create a new HZ if the new VPCE is in a different VPC. Which it's likely to be for the first $total_num_vpcs occurrences of the bug, since we attempt to "spread" (vs "pack") VPCEs across VPCs. If you still have some of these zombies around, you could confirm that they're all attached to different VPCs.

Now, this _does_ imply a leak in the new cleanup code: we would ideally be scrubbing the HZs for all the VPCs other than the one associated with our anointed VPCE. Let me look into what it would take to close that gap...

2uasimojo:
So I think I could add code to scrub the extra HZs (though it wouldn't be pretty). But as I was looking around it occurred to me: what if the user creates HZs -- or VPCEs for that matter -- out of band? Presumably we wouldn't want to be scrubbing those. Makes me wonder whether this whole PR is invalid.

2uasimojo:
Talked this over with @jstuever (thanks for the consult!) and we have a path forward: I need to search for all HZs that are
- associated with the cluster's apiDomain
- in VPCs that are configured in the inventory in HiveConfig...
- ...but *not* the one in the VPC associated with the anointed VPCE. (I suppose if I see more than one of these, I should destroy all but one of them?)

typeid:
@2uasimojo it looks good to me, thanks!

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -955,6 +982,11 @@ func (r *ReconcileAWSPrivateLink) reconcileHostedZone(awsClient *awsClient,
 		return modified, "", err
 	}
 
+	if err := r.scrubHostedZones(awsClient.hub, apiDomain, cd.Spec.Platform.AWS.Region, hostedZoneID, logger); err != nil {
+		logger.WithError(err).Error("error ensuring Hosted Zone was created")
```
2uasimojo, line 986: Bleh, copypasta

/hold

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -990,8 +1022,7 @@ func (r *ReconcileAWSPrivateLink) recordSet(awsClient awsclient.Client, apiDomai
 		rSet.TTL = aws.Int64(10)
 
 		// get the ips from the elastic networking interfaces attached to the VPC endpoint
-		enis := vpcEndpoint.NetworkInterfaceIds
-		if len(enis) == 0 {
+		if len(vpcEndpoint.NetworkInterfaceIds) == 0 {
```
2uasimojo, line 1025: ntr: latent: extraneous local var.

pkg/controller/awsprivatelink/awsprivatelink_controller.go
2uasimojo:
```suggestion
		logger.WithError(err).Error("failed to scrub redundant Hosted Zones")
```

pkg/controller/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -860,42 +869,60 @@ func (r *ReconcileAWSPrivateLink) reconcileVPCEndpoint(awsClient *awsClient,
...
 		if err != nil {
 			endpointLog.WithError(err).Error("error getting VPC Endpoint")
-			return modified, nil, err
+			return endpointReconciled, nil, err
 		}
 
-		allVpcEndpoints = append(allVpcEndpoints, resp.VpcEndpoints...)
+		// HIVE-2838: First page may be empty even when there are results!
+		if len(resp.VpcEndpoints) != 0 {
+			for _, vpce := range resp.VpcEndpoints {
```
jstuever, line 883: Nit: Do we need the additional if wrapper? I thought `range` would simply skip when empty?

2uasimojo: Right you are. I feel like in earlier versions of go, an uninitialized list variable or struct member would throw a nil pointer panic under `range`. But I just verified with 1.23 and it's fine.

PR #2668, opened 2025-04-29
HIVE-2837: Revert AWSPrivateLink cleanup code

2uasimojo:
The reverted commits were attempting to introduce code to clean up stale resources from AWSPrivateLink reconciles. This turned out to be too fragile, error-prone, and expensive to be worth the benefit we would derive from it.


2uasimojo:
I wonder if

/ok-to-test

will actually have an effect on this spurious "Red Hat Konflux" job that seems to spring into existence when I attempt to retest the _real_ konflux job, hive-on-pull-request...

2uasimojo:
Argh, it started **all** the jobs over!

PR #2676, opened 2025-05-05
HIVE-2792: Allow hive util to not pass any PC certificate

eliorerz:
- Allow installation without certificate on hiveutil.
- Fix case where the certificate is empty and the controller fail validating that the certificate.


2uasimojo (commented):
Sane.

/lgtm

2uasimojo:
/lgtm cancel

Looks like unit tests need updating

eliorerz:
> Looks like unit tests need updating

Fixed, and added another UT for catching this case

PR #2177, opened 2023-12-08
Add install log regexes for cert-related network failures

2uasimojo:
These were showing up as `FallbackInsvalidInstallConfig` with a generic message. Make them more specific/scrutable/actionable.

[HIVE-2360](https://issues.redhat.com//browse/HIVE-2360)

PR #2679, opened 2025-05-07
OSD-29996: Handle TagLimitExceeded failures for AWS subnets during install

joshbranham:
We recently saw a cluster fail to install on AWS where at least one subnet had reached the maximum number of tags (50) and therefore the install could not proceed.

The exact log from the installer was:
```
time=\"2025-05-06T11:12:26Z\" level=fatal msg=\"failed to fetch Cluster: failed to generate asset \\"Cluster\\": could not add tags to subnets: TagLimitExceeded: The resultant tag set must not have more than 50 user tags.\n\tstatus code: 400, request id: b097f663-0c2e-4403-9efa-73ca2ca3051c\"
```

This adds a handling for `could not add tags to subnets: TagLimitExceeded` which should cover this specific scenario.

### DIFF
--- a/config/configmaps/install-log-regexes-configmap.yaml
+++ b/config/configmaps/install-log-regexes-configmap.yaml
@@ -46,6 +46,11 @@ data:
       - "InvalidSubnet: Not enough IP space available in"
       installFailingReason: AWSSubnetInsufficientIPSpace
       installFailingMessage: Insufficient IP space available in subnet
+    - name: AWSSubnetTagLimitExceeded
+      searchRegexStrings:
+      - "could not add tags to subnets: TagLimitExceeded"
+      installFailingReason: AWSSubnetTagLimitExceeded
+      installFailingMessage: AWS Subnet exceeds the maximum number of tags allowed (50)
     - name: MissingPublicSubnetForZone
       searchRegexStrings:
       - "No public subnet provided for zone"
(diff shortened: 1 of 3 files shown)


config/configmaps/install-log-regexes-configmap.yaml
```diff
@@ -46,6 +46,11 @@ data:
       - "InvalidSubnet: Not enough IP space available in"
       installFailingReason: AWSSubnetInsufficientIPSpace
       installFailingMessage: Insufficient IP space available in subnet
+    - name: AWSSubnetTagLimitExceeded
+      searchRegexStrings:
+      - "could not add tags to subnets: TagLimitExceeded"
+      installFailingReason: AWSSubnetTagLimitExceeded
+      installFailingMessage: AWS Subnet exceeds the maximum number of tags allowed (50)
```
2uasimojo, line 53: How future-proof is this message? Will it always be 50? Is the limit tunable (like a quota)?

2uasimojo (commented):
Cool.

- We typically include UT when we make additions like this. [Example](https://github.com/openshift/hive/pull/2177).
- You'll need to regenerate bindata. `make generate` or `make update` should do it I think.
- One minor question inline.

2uasimojo:
> The exact log from Hive was:

nit: that's technically from installer

config/configmaps/install-log-regexes-configmap.yaml
joshbranham: The tag limit is not tunable

2uasimojo (commented):
/lgtm

PR #2682, opened 2025-05-15
HIVE-2847: AWSPrivateLink: Use ListHostedZonesByName

2uasimojo:
ListHostedZonesByVPC is problematic because
- We have to know which VPC(s) to look in, which isn't always easy
- In order to find an existing HZ, we would have to iterate over many VPCs.

Using ListHostedZonesByName should assuage both of these issues by providing a way to find our HZ with a single call.


2uasimojo:
Live testing is green. UT cutover is done. This should be ready to go.

/assign @jstuever

jstuever (approved):
/lgtm

PR #2688, opened 2025-05-28
Konflux: Enable multi-arch builds for push

dlom:
This will also hopefully retrigger the push job successfully once merged

/assign @2uasimojo


2uasimojo:
We should do this to the presubmit copy also, no?

dlom:
In theory the presubmit job will be single arch to save on compute power, and the multi-arch build will only happen on code that makes it into the master branch

2uasimojo (commented):
/lgtm

PR #2692, opened 2025-05-30
HIVE-2862: Remove python dependency from makefile

dlom:
This change allows building the container images without the python dependencies that are getting harder to get by.

xref: HIVE-2862

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,9 @@ endif
 # we don't tag versions. Override using the same versioning we apply to OperatorHub builds:
 # v{major}.{minor}.{commitcount}-{sha}
 # Note that building against a local commit may result in {major}.{minor} being rendered as
-# `UnknownBranch`. However, the {commitcount} and {sha} should still be accurate.
-SOURCE_GIT_TAG := $(shell export HOME=$(HOME); python3 -m ensurepip >&2; python3 -mpip --no-cache install --user gitpython pyyaml >&2; hack/version2.py)
+# 0.0.x-y if it is an `UnknownBranch`. However, the {commitcount} and {sha}
+# should still be accurate.
+SOURCE_GIT_TAG := $(shell hack/version2.sh)
 
 BINDATA_INPUTS :=./config/sharded_controllers/... ./config/hiveadmission/... ./config/controllers/... ./config/rbac/... ./config/configmaps/...
 $(call add-bindata,operator,$(BINDATA_INPUTS),,assets,pkg/operator/assets/bindata.go)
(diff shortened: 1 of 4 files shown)


hack/version2.sh
```diff
@@ -0,0 +1,25 @@
...
+		echo "$UNKNOWN_BRANCH_PREFIX"
+	fi
+}
+
+function commit_count() {
+	git rev-list --count "$(git rev-list --max-parents=0 HEAD)..HEAD"
+}
+
+function short_sha() {
+	git rev-parse --short HEAD
```
2uasimojo, line 22: To be compatible with the existing functionality, this should be 
```suggestion
	git rev-parse --short=7 HEAD
```

hack/version2.sh
```diff
@@ -0,0 +1,25 @@
...
+}
+
+echo "$(prefix_from_branch "$(git rev-parse --abbrev-ref HEAD)").$(commit_count)-$(short_sha)"
```
2uasimojo, line 25: To be compatible with the existing functionality, we need to include the `v` prefix, so either here via 
```suggestion
echo v"$(prefix_from_branch "$(git rev-parse --abbrev-ref HEAD)").$(commit_count)-$(short_sha)"
```

...or in the Makefile as noted above.

Makefile
```diff
@@ -76,8 +76,9 @@ endif
 # we don't tag versions. Override using the same versioning we apply to OperatorHub builds:
 # v{major}.{minor}.{commitcount}-{sha}
 # Note that building against a local commit may result in {major}.{minor} being rendered as
-# `UnknownBranch`. However, the {commitcount} and {sha} should still be accurate.
-SOURCE_GIT_TAG := $(shell export HOME=$(HOME); python3 -m ensurepip >&2; python3 -mpip --no-cache install --user gitpython pyyaml >&2; hack/version2.py)
+# 0.0.x-y if it is an `UnknownBranch`. However, the {commitcount} and {sha}
+# should still be accurate.
+SOURCE_GIT_TAG := $(shell hack/version2.sh)
```
2uasimojo, line 81: To be compatible with the existing functionality, this needs to include the `v` prefix, either by prepending it here
```suggestion
SOURCE_GIT_TAG := v$(shell hack/version2.sh)
```
...or by including it in the output of version2.sh as noted below.

hack/version2.sh
2uasimojo, line 1: We should be able to delete version2.py as part of this PR, yes?

2uasimojo:
Looks like you're either going to have to install pyyaml as part of `build-app-sre-template`, or port generate-saas-template.py to shell. The latter _might_ be possible via yq, which we definitely already install for other reasons.

2uasimojo:
@dlom any chance of resurrecting this?

/remove-lifecycle rotten

PR #2700, opened 2025-06-10
HIVE-2870: hiveadmission: proper authn to the KAS

2uasimojo:
This change corrects the hiveadmission service's authentication to the hub's kube-apiserver by injecting the correct CA bundle into the ValidatingWebhookConfigurations (VWHCs) when running on OpenShift.

Previously we were relying on the `service.beta.openshift.io/inject-cabundle` annotation. It turns out this injects the *wrong* bundle for VWHCs (though, accidentally, the right one for the aggregated APIService). Our webhooks still worked because the KAS has special handling, ignoring the VWHC's caBundle entirely, for clients using the aggregated API service, which hiveadmission does. However, OpenShift's kube-apiserver-operator (KASO) has a webhook validator that lacks this special handling (see OCPBUGS-50978) and thus generates error logs. While these errors are harmless and do not reflect any functional problem, they can be annoying/distracting for customers.

Modern (all currently-supported) versions of OpenShift create ConfigMaps in every namespace containing the service CA cert, which is needed by the APIService to authn to the KAS, and the kube root CA cert, which is needed in the VWHCs by KASO (though not by the KAS) to validate the webhooks. So this change:
- Removes the `inject-cabundle` annotation from the APIService and all VWHCs.
- Adds a code path for OpenShift clusters to look up these ConfigMaps and inject the respective certs into the respective objects explicitly.

The logic for non-OpenShift clusters remains the same.

Related-To: OCPBUGS-50978

### DIFF
--- a/pkg/operator/hive/hive_controller.go
+++ b/pkg/operator/hive/hive_controller.go
@@ -324,23 +324,24 @@ var _ reconcile.Reconciler = &ReconcileHiveConfig{}
 
 // ReconcileHiveConfig reconciles a Hive object
 type ReconcileHiveConfig struct {
-	scheme                            *runtime.Scheme
-	kubeClient                        kubernetes.Interface
-	discoveryClient                   discovery.DiscoveryInterface
-	dynamicClient                     dynamic.Interface
-	restConfig                        *rest.Config
-	hiveImage                         string
-	hiveOperatorNamespace             string
-	hiveImagePullPolicy               corev1.PullPolicy
-	nodeSelector                      map[string]string
-	tolerations                       []corev1.Toleration
-	syncAggregatorCA                  bool
-	managedConfigCMLister             corev1listers.ConfigMapLister
-	ctrlr                             controller.Controller
-	hiveSecretLister                  corev1listers.SecretLister
-	secretWatchEstablishedInNamespace string
-	mgr                               manager.Manager
-	isOpenShift                       bool
+	scheme                               *runtime.Scheme
+	kubeClient                           kubernetes.Interface
+	discoveryClient                      discovery.DiscoveryInterface
+	dynamicClient                        dynamic.Interface
+	restConfig                           *rest.Config
+	hiveImage                            string
+	hiveOperatorNamespace                string
+	hiveImagePullPolicy                  corev1.PullPolicy
+	nodeSelector                         map[string]string
+	tolerations                          []corev1.Toleration
+	syncAggregatorCA                     bool
+	managedConfigCMLister                corev1listers.ConfigMapLister
+	ctrlr                                controller.Controller
+	hiveSecretLister                     corev1listers.SecretLister
+	configMapWatchEstablishedInNamespace string
+	secretWatchEstablishedInNamespace    string
+	mgr                                  manager.Manager
+	isOpenShift                          bool
 }
 
 // Reconcile reads that state of the cluster for a Hive object and makes changes based on the state read
@@ -370,6 +371,7 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
 			hLog.Debug("HiveConfig not found, deleted?")
+			r.configMapWatchEstablishedInNamespace = ""
 			r.secretWatchEstablishedInNamespace = ""
 			return reconcile.Result{}, nil
 		}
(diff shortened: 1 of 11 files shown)


2uasimojo:
cc @jacobsee

jacobsee:
Thanks @2uasimojo!
/lgtm

2uasimojo:
> /retest-required

Hi @jacobsee. This ain't close to right in its current form. I've made progress in my sandbox, but gotten sidetracked by other matters for several days. Will hopefully update soon.

/hold
/lgtm cancel

2uasimojo:
/hold cancel

This checks out on OpenShift and KIND. Ready for review again.

pkg/operator/hive/hive_controller.go
```diff
@@ -390,6 +392,12 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if err := r.establishConfigMapWatch(hLog, hiveNSName); err != nil {
+		instance.Status.Conditions = util.SetHiveConfigCondition(instance.Status.Conditions, hivev1.HiveReadyCondition, corev1.ConditionFalse, "ErrorEstablishingConfigMapWatch", err.Error())
+		r.updateHiveConfigStatus(origHiveConfig, instance, hLog, false)
+		return reconcile.Result{}, err
+	}
+
```
jstuever: Is there a particular reason to add this to `reconcile` as opposed to `add` like the other watch use cases?

2uasimojo: Excellent question. To be honest, I copypasta'ed from the secret watch without really thinking about it.

But now having thought about it, I believe it's because we want to limit it to the currently-configured `targetNamespace`, which can change at runtime. The configmaps do get created in all namespaces, so in theory I could just pick an arbitrary one... but I would have to be guaranteed that it would always exist. Is there such a namespace? `default`? Any of the `openshift-*` ones (see below)? Is there one we're certain to have the right RBAC on? Is it worth relying on all of that for the amount of simplification it buys us, which isn't nothing, but isn't a _whole_ lot?

...but also, what I *should* have done is conditioned this on openshift-ness, since we're only using the configmaps there. I don't love the loose coupling, but it's the right thing to do. Let me rip that right quick...

2uasimojo: Done.

jstuever: You mean, like [this](https://github.com/openshift/hive/blob/03a48a83ccbbc1992131296228979f54c2e8fef8/pkg/operator/hive/hive_controller.go#L154-L156)?
```
	hiveOperatorNS := os.Getenv(HiveOperatorNamespaceEnvVar)
	r.(*ReconcileHiveConfig).hiveOperatorNamespace = hiveOperatorNS
	log.Infof("hive operator NS: %s", hiveOperatorNS)
```

2uasimojo: That env var is set when the pod is created and doesn't change at runtime. That's where hive-operator is running, which isn't necessarily the same as where the controllers are running.

But perhaps you're saying that I could rely on the cert ConfigMaps existing in *that* namespace... which is a fair point, assuming that the contents aren't specific to the namespace... which they shouldn't be, right??

jstuever: I was actually assuming the controllers run in the same namespace as the operator. It hadn't occurred to me that they could be run from another.

I think you are onto something, though. If that cert is the same in all namespaces, then it should be possible to use the one landed in the operator namespace.

With all of this in mind, I am happy to go with what you have now. I'll wait to lgm to give you an opportunity to decide which way you want to go.

2uasimojo: > It hadn't occurred to me that they could be run from another.

TBH I'm not sure *why* it's a thing, but HiveConfig.Spec.TargetNamespace lets you send the controllers to a different ns. Maybe it's an RBAC thing: you (theoretically) need different powers to deploy hive itself than to manage its controllers?

Anyway...

I tried to find confirmation that the cert stuff is the same in all namespaces, and I think the [answer](https://x.com/i/grok?conversation=1937611203787845963) is, "yes unless you've tried really hard to make it otherwise, which would comprise an unsupported misconfiguration."

That said, I think I would like to work on that in a followon... where I would also like to investigate getting rid of the k8s-vs-openshift discrepancies, since I _think_ we can count on kube-root-ca.crt existing in 1.20 (1.21?) and higher, which we know is kosher for OpenShift support, but should also be a reasonable minimum for non-OpenShift clusters.

So yeah, let's go ahead and land this, and I'll open a separate card for :arrow_up: 

Thanks!

2uasimojo:
/retest-required

ClusterPool dry

pkg/operator/hive/hive_controller.go
2uasimojo: > separate card

https://issues.redhat.com/browse/HIVE-2874

PR #2650, opened 2025-04-01
HIVE-2819: Use annotation instead of label for upgradeable

AlexVulaj:
This is a follow up to https://github.com/openshift/hive/pull/2639 , but moving the value to the cluster deployment annotations instead of labels. This change is due to an issue found with certain allowed characters for labels that do not affect annotations.

For example, the following error was seen testing the labels from the previous PR:
```
time="2025-04-01T15:52:28.453Z" level=error msg="Reconciler error" controller=clusterversion-controller error="...a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]"
```

PR #2705, opened 2025-06-16
HIVE-2871 - Always show the upgradeable annotation.

AlexVulaj:
Follow up to [HIVE-2819](https://issues.redhat.com//browse/HIVE-2819) - always show the upgradeable annotation. If the annotation is present with an empty string, this means that the cluster is upgradeable.

### DIFF
--- a/pkg/controller/clusterversion/clusterversion_controller_test.go
+++ b/pkg/controller/clusterversion/clusterversion_controller_test.go
@@ -89,7 +89,7 @@ func TestClusterVersionReconcile(t *testing.T) {
 				Status: configv1.ConditionTrue,
 			}),
 			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
-				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected version major-minor-patch label")
+				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected value for annotation hive.openshift.io/minor-version-upgrade-unavailable")
 			},
 		},
 		{
@@ -103,7 +103,7 @@ func TestClusterVersionReconcile(t *testing.T) {
 				Status: configv1.ConditionFalse,
 			}),
 			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
-				assert.Equal(t, "Upgradeable: False", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected version major-minor-patch label")
+				assert.Equal(t, "Upgradeable: False", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected value for annotation hive.openshift.io/minor-version-upgrade-unavailable")
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterversion/clusterversion_controller_test.go
```diff
@@ -89,7 +89,7 @@ func TestClusterVersionReconcile(t *testing.T) {
 				Status: configv1.ConditionTrue,
 			}),
 			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
-				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected version major-minor-patch label")
+				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "Unexpected value for label hive.openshift.io/minor-version-upgrade-unavailable")
```
wking: while we're touching this string, we probably want to fix "label" -> "annotation" to catch up with #2650.  And also [here][1], where the annotation constant is defined.

[1]: https://github.com/openshift/hive/blob/8796c4f5340c05584051b201e258fd19dfd92234/pkg/constants/constants.go#L364

wking: Hmm, and I guess we should also make up our minds about whether the annotation constant could change or not.  Some folks use Go variables/constants in unit tests, because they think it might change, and they want the unit tests to auto-pivot without needing manual intervention (like your `cd.Annotations[constants.MinorVersionUpgradeUnavailable]`).  Some folks hard-code the string in the unit test to remind themselves (via failing unit tests) that the policy is to not rename the annotation (or whatever) once it's gone live (like your `Unexpected value for label hive.openshift.io/minor-version-upgrade-unavailable`).  But hard-coding the string in the error context while using the Go variable/constant in the test itself seems like it won't fail on a rename, but might give confusing results by referring to the stale previous name if it fails on some post-rename change.

wking: Something like this for the "name can change, and we want the unit tests to auto-pivot" world:

```suggestion
				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], fmt.Sprintf("Unexpected value for annotation %s", constants.MinorVersionUpgradeUnavailable)) 
```

2uasimojo: Would have to dig into why this test was working before (as opposed to having to assert that there was no such key) but it seems accurate in its new form :shrug:

2uasimojo (commented):
I agree with Trevor at least that the label=>annotation change should be made, but since that's just in unit tests, I'm happy to let this land as is and see that change in a fup.

/lgtm

...but I'll

/hold

so you can control that cadence however you want (re-push and I'll re-stamp; or unhold and fup).

pkg/controller/clusterversion/clusterversion_controller_test.go
AlexVulaj: The cd.Annotations map gets set to a non-nil value in the functional part of the code. In Go, trying to pull a map entry by a key that doesn't exist will return the 0 value of that map's value type. For example:
```
myMap := map[string]string{}
foo := myMap["foo"]
```
in this case, `foo` will be an empty string.

The old test probably should've been asserting instead that the key didn't exist instead of looking at the value.

AlexVulaj: I updated the string and decided to keep with the const for now.

PR #2706, opened 2025-07-01
HIVE-2861: Fix residual Nutanix resources after cluster deletion due to incorrect casting of Prism Central port

eliorerz:
When deleting Nutanix clusters, residual Nutanix resources such as virtual machines and images were not being properly cleaned up. This occurred due to an incorrect casting of the `cd.Spec.Platform.Nutanix.PrismCentral.Port` field from `int32` to `string` using a direct string() cast.

/cc @2uasimojo

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -673,6 +673,8 @@ func (m *InstallManager) cleanupFailedInstall(cd *hivev1.ClusterDeployment, infr
 
 func cleanupFailedProvision(dynClient client.Client, cd *hivev1.ClusterDeployment, infraID string, logger log.FieldLogger) error {
 	var uninstaller providers.Destroyer
+
+	logger.Info("starting a cleanup failed provision for ClusterDeployment with infraID: ", infraID)
 	switch {
 	case cd.Spec.Platform.AWS != nil:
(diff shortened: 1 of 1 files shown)


2uasimojo:
/retest hive-mce-29-on-pull-request

[#itn-2025-00153](https://redhat.enterprise.slack.com/archives/C093XRTHLJW)

dlom:
@eliorerz can you rebase this PR? I've merged a change to reduce pre-merge test times

eliorerz:
> Nice find. I had no idea `string(i)` wouldn't work! (Just verified via go playground -- it always reports empty string. You would think that would never be the desired behavior, and they would either change it or disallow it via compiler/linting 🤷)
> 
> I sure would like to see some unit test for this, but it looks like it would require a refactor of `cleanupFailedProvision()` to separate building the destroyer from running it. I'm writing a tech debt card to look into doing that (for all platforms): https://issues.redhat.com/browse/HIVE-2876
> 
> /lgtm

Yeah, it took me a while to find it.
Regarding the UT, as you said I wanted to add some unittests for that case but could not do that without refactoring the `cleanupFailedProvision` method.

dlom:
re-lgtming, as my request for rebase knocked off Eric's

/lgtm

PR #2707, opened 2025-07-02
HIVE-2871 | (fix): Update the clusterversion controller to handle the case where the upgradeable annotation doesn't exist and needs to be re-populated

AlexVulaj:
Currently, the controller will not recognize that the "upgradeable" annotation is missing and requires repopulation. This change treats a missing annotation as a required change that needs to be updated during reconcilation.

### DIFF
--- a/pkg/controller/clusterversion/clusterversion_controller_test.go
+++ b/pkg/controller/clusterversion/clusterversion_controller_test.go
@@ -81,82 +81,175 @@ func TestClusterVersionReconcile(t *testing.T) {
 		{
 			name: "upgradeable condition true",
 			existing: []runtime.Object{
-				testClusterDeployment(),
+				testClusterDeploymentWithLabelsAndAnnotations(map[string]string{
+					constants.VersionLabel:                "1.2.3",
+					constants.VersionMajorLabel:           "1",
+					constants.VersionMajorMinorLabel:      "1.2",
+					constants.VersionMajorMinorPatchLabel: "1.2.3",
+				}, map[string]string{constants.MinorVersionUpgradeUnavailable: "Can't upgrade"}),
 				testKubeconfigSecret(),
 			},
-			clusterVersionStatus: testRemoteClusterVersionStatus(configv1.ClusterOperatorStatusCondition{
-				Type:   configv1.OperatorUpgradeable,
-				Status: configv1.ConditionTrue,
-			}),
+			clusterVersionStatus: testRemoteClusterVersionStatusWithVersion("1.2.3",
+				configv1.ClusterOperatorStatusCondition{
+					Type:   configv1.OperatorUpgradeable,
+					Status: configv1.ConditionTrue,
+				}),
 			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
-				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected value for annotation hive.openshift.io/minor-version-upgrade-unavailable")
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterversion/clusterversion_controller_test.go
```diff
@@ -106,6 +106,35 @@ func TestClusterVersionReconcile(t *testing.T) {
...
+			existing: []runtime.Object{
+				testClusterDeploymentWithoutAnnotations(),
+				testKubeconfigSecret(),
+			},
+			clusterVersionStatus: testRemoteClusterVersionStatus(configv1.ClusterOperatorStatusCondition{
+				Type:   configv1.OperatorUpgradeable,
+				Status: configv1.ConditionTrue,
+			}),
+			validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
+				assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected value for annotation hive.openshift.io/minor-version-upgrade-unavailable")
```
2uasimojo: This still passes with the controller change reverted.

Also, I believe it's identical to the case two slots above it (`"upgradeable condition true"`).

Passing in this form makes sense. We're getting hit by the same behavior: the nil value of the annotation is the empty string.

What _doesn't_ make sense is why this doesn't fix it:

```diff
diff --git a/pkg/controller/clusterversion/clusterversion_controller_test.go b/pkg/controller/clusterversion/clusterversion_controller_test.go
index b1286722f..d18a4b232 100644
--- a/pkg/controller/clusterversion/clusterversion_controller_test.go
+++ b/pkg/controller/clusterversion/clusterversion_controller_test.go
@@ -89,7 +89,9 @@ func TestClusterVersionReconcile(t *testing.T) {
                                Status: configv1.ConditionTrue,
                        }),
                        validate: func(t *testing.T, cd *hivev1.ClusterDeployment) {
-                               assert.Equal(t, "", cd.Annotations[constants.MinorVersionUpgradeUnavailable], "unexpected value for annotation hive.openshift.io/minor-version-upgrade-unavailable")
+                               val, ok := cd.Annotations[constants.MinorVersionUpgradeUnavailable]
+                               assert.True(t, ok, "Expected annotation to be present")
+                               assert.Equal(t, "", val, "unexpected value for annotation hive.openshift.io/minor-version-upgrade-unavailable")
                        },
                },
                {
```

2uasimojo: Sorry:

> Also, I believe it's identical to the case two slots above it (`"upgradeable condition true"`).

This is not so; yours starts off with no annotations :+1: 

And the delta I pasted above was originally applied to the wrong test case... but when I apply it to your new one it still passes with the old controller code...

AlexVulaj: It looks like "changed" is getting set to true before the part we care about in those specific tests, let me quickly change that up.

2uasimojo: Okay, I think it's because we need to set up the seed CD with the semver labels already set. Otherwise the function will detect a change and populate `upgradeable` along with the labels.

AlexVulaj: Correct - I'm going to get around this by making sure the tests around this annotation have a matching Desired version from the CV status as well as what's on the CD label. That should isolate our tests around the annotation alone.

AlexVulaj: Alright - should be all good now!

PR #2708, opened 2025-07-02
Align RPM versions with the ELS image

dhaiducek:
With the ELS image in use, there are old versions of RPMs lingering while the lockfile specifies new versions. Using the EUS repo aligns the versions.

Slack thread: https://redhat-internal.slack.com/archives/C06TJJ3E0MU/p1751287431351829


2uasimojo:
We _think_ this may have broken HCM, which is now suffering from CrashLoopBackoff with `exec /opt/services/hive-operator: exec format error`. I'm going to open up a revert to get things moving.

PR #2710, opened 2025-07-07
Revert "Align RPM versions with the ELS image"

2uasimojo:
Reverts openshift/hive#2708

We think this may be causing an arch conflict in HCM. Reverting while we figure out how to fix it properly.


2uasimojo:
This turned out to be (probably) unrelated to the build issue.

PR #2714, opened 2025-07-16
HIVE-2860: Fix `ensureEnoughReplicas` to properly clear `NotEnoughReplicas` condition 

eliorerz:
Fix `NotEnoughReplicas` condition not being cleared when switching from autoscaling to fixed replicas.
The `ensureEnoughReplicas` function wasn't properly clearing the `NotEnoughReplicas` condition when a `MachinePool` transitioned from autoscaling mode to having sufficient fixed replicas. 
This fix ensures the condition is cleared when autoscaling is disabled and replica count meets requirements.
Includes unit tests to verify proper condition management across autoscaling transitions.

/cc @2uasimojo

### DIFF
--- a/pkg/controller/machinepool/machinepool_controller.go
+++ b/pkg/controller/machinepool/machinepool_controller.go
@@ -619,29 +619,41 @@ func (r *ReconcileMachinePool) ensureEnoughReplicas(
 	cd *hivev1.ClusterDeployment,
 	logger log.FieldLogger,
 ) (*reconcile.Result, error) {
-	if pool.Spec.Autoscaling == nil {
-		return nil, nil
-	}
-	if pool.Spec.Autoscaling.MinReplicas < int32(len(generatedMachineSets)) && !platformAllowsZeroAutoscalingMinReplicas(cd) {
-		logger.WithField("machinesets", len(generatedMachineSets)).
-			WithField("minReplicas", pool.Spec.Autoscaling.MinReplicas).
-			Warning("when auto-scaling, the MachinePool must have at least one replica for each MachineSet")
+	if pool.Spec.Autoscaling != nil {
+		// Handle autoscaling case: check if we have enough replicas
+		if pool.Spec.Autoscaling.MinReplicas < int32(len(generatedMachineSets)) && !platformAllowsZeroAutoscalingMinReplicas(cd) {
+			logger.WithField("machinesets", len(generatedMachineSets)).
+				WithField("minReplicas", pool.Spec.Autoscaling.MinReplicas).
+				Warning("when auto-scaling, the MachinePool must have at least one replica for each MachineSet")
+			err := r.updateCondition(
+				pool,
+				hivev1.NotEnoughReplicasMachinePoolCondition,
+				corev1.ConditionTrue,
+				"MinReplicasTooSmall",
+				fmt.Sprintf("When auto-scaling, the MachinePool must have at least one replica for each MachineSet. The minReplicas must be at least %d", len(generatedMachineSets)),
+				logger,
+			)
+			return &reconcile.Result{}, err
+		}
+		// Autoscaling enabled and we have enough replicas
 		err := r.updateCondition(
 			pool,
 			hivev1.NotEnoughReplicasMachinePoolCondition,
-			corev1.ConditionTrue,
-			"MinReplicasTooSmall",
-			fmt.Sprintf("When auto-scaling, the MachinePool must have at least one replica for each MachineSet. The minReplicas must be at least %d", len(generatedMachineSets)),
+			corev1.ConditionFalse,
+			"EnoughReplicas",
+			"The MachinePool has sufficient replicas for each MachineSet",
 			logger,
 		)
-		return &reconcile.Result{}, err
+		return nil, err
 	}
+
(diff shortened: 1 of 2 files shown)


huangmingxia:
Pre-merge test passed.
1. Installed Hive using `quay.io/mihuang/hive:2860-ensureEnoughReplicas`.
2. Created a MachinePool with autoscaling enabled. [ok]
3. Updated minReplicas to 2 (less than the number of failure domains, numOfFD = 3): MachinePool entered the NotEnoughReplicas condition.
```
spec:
  autoscaling:
    maxReplicas: 4
    minReplicas: 2
  clusterDeploymentRef:
    name: sgao-sts-zone
  name: infra
  platform:
    nutanix:
      bootType: Legacy
      coresPerSocket: 2
      cpus: 4
      failureDomains:
      - failure_domain_1
      - failure_domain_2
      - failure_domain_3
      memoryMiB: 16348
      osDisk:
        diskSizeGiB: 120
status:
  conditions:
  - lastProbeTime: "2025-07-16T10:02:23Z"
    lastTransitionTime: "2025-07-16T10:02:23Z"
    message: When auto-scaling, the MachinePool must have at least one replica for
      each MachineSet. The minReplicas must be at least 3
```
4. Updated minReplicas to 5 to meet the requirement. The NotEnoughReplicas condition was cleared. After logging into the spoke cluster, the corresponding MachineSets/Machines were created or removed accordingly.
  ```
  spec:
    autoscaling:
      maxReplicas: 6
      minReplicas: 5
    clusterDeploymentRef:
      name: sgao-sts-zone
    name: infra
    platform:
      nutanix:
        bootType: Legacy
        coresPerSocket: 2
        cpus: 4
        failureDomains:
        - failure_domain_1
        - failure_domain_2
        - failure_domain_3
        memoryMiB: 16348
        osDisk:
          diskSizeGiB: 120
  status:
    conditions:
    - lastProbeTime: "2025-07-16T10:03:09Z"
      lastTransitionTime: "2025-07-16T10:03:09Z"
      message: The MachinePool has sufficient replicas for each MachineSet
      reason: EnoughReplicas
      status: "False"
      type: NotEnoughReplicas
  ```
5. When NotEnoughReplicas was true, removing the autoscaling field and setting replicas = 3 cleared the NotEnoughReplicas condition.

```
 replicas: 2
status:
  conditions:
  - lastProbeTime: "2025-07-16T10:12:14Z"
    lastTransitionTime: "2025-07-16T10:12:14Z"
    message: The MachinePool has sufficient replicas for each MachineSet
    reason: EnoughReplicas
    status: "False"
    type: NotEnoughReplicas
```

pkg/controller/machinepool/machinepool_controller_test.go
```diff
@@ -1350,6 +1351,64 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 				testMachineSet("foo-12345-worker", "worker", true, 1, 0),
 			},
 		},
+		{
+			name:              "NotEnoughReplicas condition cleared when switching from autoscaling to fixed replicas",
```
2uasimojo: Is this test case providing value beyond the new suite you've introduced below?

pkg/controller/machinepool/machinepool_controller_test.go
```diff
@@ -1428,6 +1487,16 @@ func TestRemoteMachineSetReconcile(t *testing.T) {
 				return
 			}
 
+			// Special validation for NotEnoughReplicas condition clearing test
+			if test.name == "NotEnoughReplicas condition cleared when switching from autoscaling to fixed replicas" {
```
2uasimojo: Eh, it seems icky to call out a specific test by name here.

I think we can do better, particularly in light of the comment I made on the code side.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -620,7 +620,16 @@ func (r *ReconcileMachinePool) ensureEnoughReplicas(
 	logger log.FieldLogger,
 ) (*reconcile.Result, error) {
 	if pool.Spec.Autoscaling == nil {
-		return nil, nil
+		// When autoscaling is disabled, clear the NotEnoughReplicas condition if it exists
+		err := r.updateCondition(
```
2uasimojo, line 639: I was going to start by complaining that this could be consolidated with L648-55 below by inverting the condition and wrapping it around L634-47...

...but upon closer inspection, I think we want to do something different. As written, this will result in inconsistent behavior:
- I create my MP without autoscaling. This condition shows up in its Initialized form.
- I enable autoscaling. This condition gets set to one of the known values.
- I disable autoscaling. As written, the condition will always get set to the "EnoughReplicas" state.

Remember that k8s tries to be declarative rather than procedural. In other words, we're not looking for "when switching autoscaling off, X should happen," but rather, "whenever autoscaling is off, the condition should be in X state."

TL;DR: I think we should return the condition to the Initialized state here.

This also makes unit testing easier: you can simply assert that the condition is Initialized any time autoscaling is off in the input config.

eliorerz: Good point! Fixed both issues:

1. Now sets condition to ConditionUnknown/Initialized when autoscaling is off.
2. Inverted the logic as you suggested 

Tests are simpler now too :)

PR #2715, opened 2025-07-16
Add quay.io image pull secret to app-sre service account

dlom:
Corresponding app-interface MR: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/150878 

Relevant docs: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/sop/make-registry-private.md

### DIFF
--- a/config/controllers/hive_controllers_serviceaccount.yaml
+++ b/config/controllers/hive_controllers_serviceaccount.yaml
@@ -3,3 +3,6 @@ kind: ServiceAccount
 metadata:
   name: hive-controllers
   namespace: hive
+imagePullSecrets:
+# TODO: don't hardcode this in this file.  As far as I can tell, this is harmless if the secret doesn't exist
+- name: quay.io
--- a/config/hiveadmission/service-account.yaml
+++ b/config/hiveadmission/service-account.yaml
@@ -4,3 +4,6 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: hiveadmission
+imagePullSecrets:
+# TODO: don't hardcode this in this file.  As far as I can tell, this is harmless if the secret doesn't exist
+- name: quay.io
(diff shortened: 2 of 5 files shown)


dlom:
Nevermind, I think this PR may be unnecessary

2uasimojo:
- Looks like we need a different token for the prod repo
- Are we supposed to use the workloads repo for presubmit and the prod repo for post?

2uasimojo:
> Nevermind, I think this PR may be unnecessary

Mm, I thought I remembered the process being that we would copy (only) the appropriate images from workloads to prod after they were built. (After "signing" them or something??) The way this is written it looks like we would just be using the prod repo for everything.

dlom:
/retitle Add quay.io image pull secret to app-sre service account

dlom:
@2uasimojo I've repurposed this PR.  It's ready for a proper review now

2uasimojo (commented):
I don't understand what's happening here, but apparently neither does anyone else, so we're basically just iterating by throwing stuff at the wall and seeing what sticks.

/lgtm

2uasimojo:
I think you're going to need to hit the admission SA as well.

config/controllers/hive_controllers_serviceaccount.yaml
```diff
@@ -3,3 +3,6 @@ kind: ServiceAccount
 metadata:
   name: hive-controllers
   namespace: hive
+imagePullSecrets:
+# TODO: don't hardcode this in this file.  As far as I can tell, this is harmless if the secret doesn't exist
+- name: quay.io
```
2uasimojo, line 8: I don't _think_ the missing newline will affect anything...

2uasimojo (commented):
/lgtm

PR #2721, opened 2025-07-24
HIVE-2887: Cursor: refactor pkg/operator/util/apply.go

2uasimojo:
Cursor's take on consolidating all the Apply*() functions. Compare to PR #2720.

See [docs/cursor/cursor_consolidate_apply_functions_in_a.md](https://github.com/openshift/hive/blob/bebb47e64986672c94052ea738a4842e88c0c367/docs/cursor/cursor_consolidate_apply_functions_in_a.md) for the
prompts/responses.

### DIFF
--- a/docs/cursor/cursor_consolidate_apply_functions_in_a.md
+++ b/docs/cursor/cursor_consolidate_apply_functions_in_a.md
@@ -1,0 +1,2048 @@
+# Consolidate Apply functions in apply.go
+_Exported on 7/28/2025 at 16:08:17 CDT from Cursor (1.2.4)_
+
+---
+
+**User**
+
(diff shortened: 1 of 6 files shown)


2uasimojo:
oh yeah

/hold

we have no intention of merging this.

2uasimojo:
I tried asking cursor explicitly to use the functional options pattern rather than a config object parameter, and the result was substantially the same -- it used a variadic arg to pass in functions that set values on a config object!

I don't hate the config object approach. I don't have a strong argument for either, really.

For the inputs, I might actually prefer the type switch to the way I've done it with converter funcs. The only advantage to the latter would be the ability to extend the input types without modifying the function -- but that would only matter if the function was intended for external use, which it ain't.

Anyway, thank you cursor, your work here is done.

/close

PR #2722, opened 2025-07-28
OpenStack Hibernation

msu8:
Adds hibernation functionality for OpenStack clusters.

Hibernation is based on restoring instances from snapshots, freeing up all of the project's resources that were taken by the instances. Replicating true hibernation in OpenStack is difficult to impossible for us - e.g. shelving in OpenStack exists, however it does not release project's resources, as it depends on how OpenStack is configured. This solution is universally compatible.

After a ClusterPool is created and claimed 

`$ oc patch cd $cd -n $cd_ns --type='merge' -p $'spec:\n powerState: Hibernating'`

Hibernation:
1. Find all cluster instances by infraID prefix 
2. Create snapshots of each instance
3. Collect and save all instance attributes to a secret (CD ns):
   - Flavor IDs, port IDs, security groups
   - Instance metadata, networking config
   - Snapshot IDs for restoration
4. Delete the actual instances (this frees up project's resources)

`$ oc patch cd $cd -n $cd_ns --type='merge' -p $'spec:\n powerState: Running'`

Restoration
1. Load saved config from secret
2. Validate project has enough resources
3. Recreate instances from snapshots
4. Wait for instances to become ACTIVE
5. Clean up hibernation secret

This approach still has many quirks, especially the time it takes to save the snapshots and restore the instances, with the instances going through different states - the code is setup to work around those, frequently update and check the actual state of the instances.

Looking forward to feedback and more suggestions.


2uasimojo:
/ok-to-test

I'll start having a look at this now.

About how long does each side of the operation take?

docs/hibernating-clusters.md
```diff
@@ -345,4 +346,22 @@ The unreachable controller tracks the reachability of a cluster and is responsib
...
+Restoration recreates instances with identical attributes to their original state.
+
+Due to the specific nature of the snapshot based implementation, the hibernation and restoration process inherently requires more time to finish compared to other cloud providers.
+
+Once the ClusterPool is created and the cluster is claimed, the workflow is similar to other implementations:
+
+```bash
+$ oc patch cd $cd -n $cd_ns --type='merge' -p $'spec:\n powerState: Hibernating'
+$ oc patch cd $cd -n $cd_ns --type='merge' -p $'spec:\n powerState: Running'
+```
```
2uasimojo: I think this chunk isn't needed.
- Hibernation is used automatically for unclaimed clusters in ClusterPools whenever `size > runningCount`.
- Hibernation isn't specific to ClusterPools -- it can be used for standalone ClusterDeployments.
- These patch commands are already listed above (~L35).

2uasimojo:
I'd like to have a discussion about the different options available here. Specifically the relative resource savings provided by shelving vs. something like suspend, traded off with the time to perform the hibernate and resume operations.

It has been a minute, so I asked [grok](https://x.com/i/grok/share/kgbwvaJGzl1oFYms9gPFMXxdw) to refresh my memory. IIUC shelving:
- depends on a glance setup (is that pretty much universal with all OpenStack deployments?)
- frees up *all* the infra resources (CPU, memory, network, storage) at the expense of big blobs in glance
- takes a long time because it's pushing/pulling instance disk images across the network

In contrast, suspend:
- saves RAM state to the compute node for each instance (i.e. a lot less data, and local vs over the network)
- frees CPU but retains memory and local storage reservations
- still stops I/O

The architectural raison d'être of hibernation in a cloud scenario is to reduce infra cost while making it faster to resume than to provision. In AWS, the former is ~2/3 savings; and the latter is about an order of magnitude (~4-5m vs 45m). In particular, although it may be a side benefit in some setups, overcommitting is *not* a primary goal. That is, we're not trying to provide a way to create more clusters than you have resources for, on the theory that you would only ever be running a subset of them. The controls we provide on ClusterPools -- `size`/`runningCount`, `maxSize` and `maxConcurrent` -- are more designed for quota and throttling than for overcommitting. (It feels like you could plan for saturation based on `maxSize - (size - runningCount)`, but that doesn't account for in-flight provisions that happen when a cluster is claimed from the pool -- those clusters are only hibernated once they have finished provisioning.) And importantly, hibernation in AWS entails stopping instances, but we're not deleting them. We're also not touching other infra pieces, which would still contribute to overall quota usage.

Putting this in context with shelve vs suspend: shelving results in a 100% resource savings (glance aside) at the expense of slow hibernate/resume, with no real guarantee of overcommit capability. Whereas suspend is much quicker (right?) at the cost of effectively enforcing no overcommit.

In conclusion, I feel like suspend aligns more closely with the philosophy for which hibernation was conceived. What are your thoughts? I would like to get @abraverm's input as well.

(BTW, it would have been neat to have this discussion, e.g. via an enhancement proposal, before you sunk all this work into the implementation...)

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+	// Security Groups
+	GetServerSecurityGroups(ctx context.Context, serverID string) ([]string, error)
+
+	// Project resources
+	GetComputeQuotas(ctx context.Context) (*ComputeQuotas, error)
+	GetComputeUsage(ctx context.Context) (*ComputeUsage, error)
+	GetFlavorDetails(ctx context.Context, flavorID string) (*FlavorDetails, error)
+}
+
+type Network struct {
```
2uasimojo: Can't we import these typedefs from gophercloud rather than shimming them?

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+
+	// Convert custom options to satisfy GopherCloud
+	createOpts := &servers.CreateOpts{
+		Name:           opts.Name,
+		ImageRef:       opts.ImageRef,
+		FlavorRef:      opts.FlavorRef,
+		Networks:       networks,
+		SecurityGroups: opts.SecurityGroups,
+		Metadata:       opts.Metadata,
+	}
```
2uasimojo: ...like here. What's the advantage of having our own `ServerCreateOpts` that we simply map to `servers.CreateOpts` in code, vs letting the caller construct the `servers.CreateOpts` out of effectively the same data?

2uasimojo:
Re build errors: it looks like you need to do `go mod vendor; make update vendor`. (We really should have it so the latter incorporates the former, but it's harder than it sounds to get everything done in the right order :( )

You'll have to explicitly check in the new vendored files from gophercloud.

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+	allPages, err := servers.List(c.computeClient, opts).AllPages()
+	if err != nil {
+		return nil, err
+	}
+	return servers.ExtractServers(allPages)
+}
+
+func (c *openstackClient) GetServer(ctx context.Context, serverID string) (*servers.Server, error) {
+	server, err := servers.Get(c.computeClient, serverID).Extract()
+	return server, err
```
2uasimojo: nit, could just be 
```suggestion
	return servers.Get(c.computeClient, serverID).Extract()
```

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+		return nil, fmt.Errorf("failed to create server: %w", err)
+	}
+
+	return server, nil
+}
+
+// Image methods
+func (c *openstackClient) GetImage(ctx context.Context, imageID string) (*images.Image, error) {
+	image, err := images.Get(c.imageClient, imageID).Extract()
+	return image, err
```
2uasimojo: nit:
```suggestion
	return images.Get(c.imageClient, imageID).Extract()
```

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+}
+
+// Ports
+func (c *openstackClient) ListPorts(ctx context.Context) ([]Port, error) {
+	allPages, err := ports.List(c.networkClient, nil).AllPages()
+	if err != nil {
+		return nil, err
+	}
+
+	portList, err := ports.ExtractPorts(allPages)
```
2uasimojo: Another example: why would we need to repackage the resulting `[]ports.Port` into our own `[]Port` rather than just 
```suggestion
	return ports.ExtractPorts(allPages)
```
?

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+	net := networkList[0]
+	return &Network{
+		ID:     net.ID,
+		Name:   net.Name,
+		Status: net.Status,
+	}, nil
+}
+
+// Get the security group names for a specific server
+func (c *openstackClient) GetServerSecurityGroups(ctx context.Context, serverID string) ([]string, error) {
```
2uasimojo: If it is truly desirable for this func to return a list of names rather than a list of security groups, suggest naming it
```suggestion
func (c *openstackClient) GetServerSecurityGroupNames(ctx context.Context, serverID string) ([]string, error) {
```
instead.

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+	if creds.UserDomainID != "" {
+		authOpts.DomainID = creds.UserDomainID
+	}
+	if creds.ProjectDomainID != "" {
+		authOpts.DomainID = creds.ProjectDomainID
+	}
+	if creds.ProjectDomainName != "" {
+		authOpts.DomainName = creds.ProjectDomainName
+	}
```
2uasimojo: I think all of these conditionals are redundant. If the `creds.*` field is the empty string, you're just avoiding setting the corresponding `authOpts.* field to the empty string... which is already its default (nil) value.

...but again, this is only relevant if we continue to see the need for a shim struct rather than just letting the caller use the upstream `gophercloud.AuthOptions` directly.

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,513 @@
...
+		Name:  flavor.Name,
+		VCPUs: flavor.VCPUs,
+		RAM:   flavor.RAM,
+		Disk:  flavor.Disk,
+	}, nil
+}
+
+// Create a client wrapper object for interacting with OpenStack.
+func NewClientFromSecret(secret *corev1.Secret) (Client, error) {
+	cloudsYaml, ok := secret.Data["clouds.yaml"]
```
2uasimojo: nit: we have a [const](https://github.com/openshift/hive/blob/2c213e9891ef54307e4521782b5dbba3848a5a69/pkg/constants/constants.go#L234) for this.

abraverm:
> I'd like to have a discussion about the different options available here. Specifically the relative resource savings provided by shelving vs. something like suspend, traded off with the time to perform the hibernate and resume operations.
> 
> It has been a minute, so I asked [grok](https://x.com/i/grok/share/kgbwvaJGzl1oFYms9gPFMXxdw) to refresh my memory. IIUC shelving:
> 
> * depends on a glance setup (is that pretty much universal with all OpenStack deployments?)
> * frees up _all_ the infra resources (CPU, memory, network, storage) at the expense of big blobs in glance
> * takes a long time because it's pushing/pulling instance disk images across the network
> 
> In contrast, suspend:
> 
> * saves RAM state to the compute node for each instance (i.e. a lot less data, and local vs over the network)
> * frees CPU but retains memory and local storage reservations
> * still stops I/O
> 
> The architectural raison d'être of hibernation in a cloud scenario is to reduce infra cost while making it faster to resume than to provision. In AWS, the former is ~2/3 savings; and the latter is about an order of magnitude (~4-5m vs 45m). In particular, although it may be a side benefit in some setups, overcommitting is _not_ a primary goal. That is, we're not trying to provide a way to create more clusters than you have resources for, on the theory that you would only ever be running a subset of them. The controls we provide on ClusterPools -- `size`/`runningCount`, `maxSize` and `maxConcurrent` -- are more designed for quota and throttling than for overcommitting. (It feels like you could plan for saturation based on `maxSize - (size - runningCount)`, but that doesn't account for in-flight provisions that happen when a cluster is claimed from the pool -- those clusters are only hibernated once they have finished provisioning.) And importantly, hibernation in AWS entails stopping instances, but we're not deleting them. We're also not touching other infra pieces, which would still contribute to overall quota usage.
> 
> Putting this in context with shelve vs suspend: shelving results in a 100% resource savings (glance aside) at the expense of slow hibernate/resume, with no real guarantee of overcommit capability. Whereas suspend is much quicker (right?) at the cost of effectively enforcing no overcommit.
> 
> In conclusion, I feel like suspend aligns more closely with the philosophy for which hibernation was conceived. What are your thoughts? I would like to get @abraverm's input as well.
> 

The main purpose of hibernation in on premise clouds is to increase number of clusters, as the resources are limited and disk is cheaper than memory and CPU. That is having more clusters in hibernation, ready to be started and save the time it takes to deploy a new cluster (about ~10 min vs 1 hour). In Openstack each project (tenant) usually has a resources quota. Suspending an instance (VM)  **doesn't** reduce the resource usage. Shelving could reduce the resource uage but it requires additional configuration - [Quota usage from placement](https://docs.openstack.org/nova/latest/admin/quotas.html#quota-usage-from-placement), which in our case caused problem and we had to revert. This is why we ended up with this design.

> (BTW, it would have been neat to have this discussion, e.g. via an enhancement proposal, before you sunk all this work into the implementation...)

Sure thing.

2uasimojo:
> (about ~10 min vs 1 hour).

If this is a good heuristic for the timing of the shelving/unshelving process, then I'm fine with it. I was just envisioning trying to shove a couple TB of disk images across a network to a storage device and thinking that might take long enough to kill the value vs spinning up a new cluster. Presumably the speed is dependent on having glance connected to particular types of storage -- more direct-attached-y and less network-y? If so, that seems like something worth calling out in the doc section you're adding here.

> The main purpose of hibernation in on premise clouds is to increase number of clusters

I'll defer to your experience on this point, but I'll reiterate that hive seems to have been designed (before my time) such that overcommit capability was not a primary goal. Let me elaborate on the example I mentioned above:

> (It feels like you could plan for saturation based on maxSize - (size - runningCount), but that doesn't account for in-flight provisions that happen when a cluster is claimed from the pool -- those clusters are only hibernated once they have finished provisioning.)

So simplistically let's say you have 1 CPU per cluster and a quota of 5 CPU. You think you can overcommit with the following settings:
```yaml
size: 7
runningCount: 2
maxSize: 10
maxConcurrent: 5
```

- `maxSize - size = 10 - 7 = 3` -- this is the max number of claimed clusters at once.
- Balanced with `runningCount = 2`, that's your 5 total running clusters to saturate your CPU quota.
- `size - runningCount = 7 - 2 = 5` -- this is your overcommit, which is okay because they're hibernating and don't affect quota.
- `maxConcurrent = 5` is how you're able to populate the pool initially: provisioning clusters are effectively "running" for quota purposes; you need this throttle so the pool waits until the first cluster finishes and hibernates before starting the sixth provision, etc.

Now consider the steady state of a populated pool with no claimed clusters:
==> 5 hibernating, 2 running.
Three claims come in.
We satisfy two of them with the already-running clusters, and resume one of the hibernating ones for the third claim.
==> 4 hibernating, 3 running+claimed.
Now we try to replenish the pool. We need to add 3 clusters.
==> 4 hibernating, 3 running+claimed, 3 provisioning.
That's 3+3=6 clusters contributing to quota, so one of the new ones will fail to provision.

I'm not saying you shouldn't try to play this game; just that, even when you're pretty sure you've got the math right, there are non-crazy scenarios that will still blow up, and it's because the controller logic wasn't designed with overcommit in mind.

> [Quota usage from placement](https://docs.openstack.org/nova/latest/admin/quotas.html#quota-usage-from-placement)

Interesting side note, this text looks eerily familiar. I was a Placement core while I was working in upstream OpenStack, and it's possible I had a hand in authoring this, or at least reviewing it :)

abraverm:
>If this is a good heuristic for the timing of the shelving/unshelving process, then I'm fine with it. I was just 
envisioning trying to shove a couple TB of disk images across a network to a storage device and thinking that might take long enough to kill the value vs spinning up a new cluster. Presumably the speed is dependent on having glance connected to particular types of storage -- more direct-attached-y and less network-y? If so, that seems like something worth calling out in the doc section you're adding here.

IIRC the snapshots were actully smaller than expected, that is not 100% copy of a disk.

@msu8, please add info in the doc about the expected hibernation storage usage for normal deployment.

> I'm not saying you shouldn't try to play this game; just that, even when you're pretty sure you've got the math right, there are non-crazy scenarios that will still blow up, and it's because the controller logic wasn't designed with overcommit in mind.

We could add some queue for on-prem cloud credentials that would decide if to deploy a clusters or restore from hibernation, but lets call it future work and avoid over-engineering this feature.

@msu8, could you add a warning in the documenation about how Hive would try restore a cluster from hibernation and how it doesn't solves a competing claims and clusterpools?

msu8:
Really appreciate your initial look at this @2uasimojo  , and thanks for further clarifying some stuff @abraverm , I'm occupied with other projects right now, so I'll try to address the above issues as soon as possible (early next week).

msu8:
Hi, I tried to address the feedback from the first review and shortened the client considerably, as well as hopefully fixed the vendor issue.

I was stuck a little bit in overengineering the actuator code - I added a two-phase snapshot deletion to free up as much resources as possible. But I’m not sure whether this is the correct way to do this, so I’m looking forward to more feedback.

The issue essentially is that if you create a basic cp (3 masters, 3 workers), the worker snapshots cannot be deleted, only the master instances can. This means that the “old” worker instances will be cleaned up in the next hibernation phase and new snapshots will be created.

However I’m not too confident on the *correctness* of this solution or whether this is even necessary, as this will leave out the possibility to manually recreate the master instance; if that’s even desirable. At the same time, this minimizes the resource use even more (although the snapshots don’t take away from the project resources - they need to take away from somewhere).

Also, I did not add more to the documents as I feel there’ll be more work and changes on the doc side of things.

abraverm:
> The issue essentially is that if you create a basic cp (3 masters, 3 workers), the worker snapshots cannot be deleted, only the master instances can. This means that the “old” worker instances will be cleaned up in the next hibernation phase and new snapshots will be created.

Having two snapshots for a short time is not ideal but there is nothing you can do about it AFAIK. I think you should document this so the user won't be surprised that during hibernation there is a surge of disk usage on Openstack.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -360,10 +340,18 @@ func (a *openstackActuator) validateRestoreResources(openstackClient openstackcl
...
 
+	// Handle usage data (can be nil)
+	var instancesUsed, coresUsed, ramUsed int
+	if usage != nil {
+		instancesUsed = len(usage.ServerUsages)
+		coresUsed = int(usage.TotalVCPUsUsage)
+		ramUsed = int(usage.TotalMemoryMBUsage)
+	}
+
 	// Check available resources
```
2uasimojo: We generally don't bother with preflight checks for quota usage. The reason is that many factors can affect quotas asynchronously: other clusters may be provisioning/deprovisioning, hibernating/resuming, etc. -- which may or may not even be happening/visible on *this* hub. Point being that you're not seriously reducing the probability of hitting a quota failure by doing this check.

I won't veto it if you really want to keep it, but I see it being mostly a waste of code, including the associated maintenance etc.

pkg/openstackclient/client.go
```diff
@@ -28,53 +30,24 @@ type Client interface {
 	GetServer(ctx context.Context, serverID string) (*servers.Server, error)
```
2uasimojo, line 28: At a glance (heh) this interface looks a lot better now, thank you!

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -105,41 +83,44 @@ func (a *openstackActuator) StopMachines(cd *hivev1.ClusterDeployment, hiveClien
...
 
-	// Validate instance states before snapshotting
-	err = a.validateInstanceStates(openstackClient, matchingServers, logger)
-	if err != nil {
+	if err := a.validateInstanceStates(openstackClient, matchingServers, logger); err != nil {
 		return err
 	}
 
-	// Create snapshots for each instance
+	// 3.
```
2uasimojo: There doesn't seem to be a `2.` :innocent:

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -199,9 +180,10 @@ func (a *openstackActuator) createSnapshots(openstackClient openstackclient.Clie
 	for i, server := range servers {
 		logger.Infof("creating snapshot %d/%d for instance %s", i+1, len(servers), server.Name)
 
-		snapshotID, err := openstackClient.CreateServerSnapshot(ctx, server.ID, server.Name)
+		snapshotName := fmt.Sprintf("%s-hibernation-%s", server.Name, time.Now().UTC().Format("20060102-150405"))
```
2uasimojo: Running `time.Now()` inside of the loop means we're likely to end up with different suffixes for snapshots for the different servers for the same hibernation event. Is that the desired outcome? I would have thought it would be better for debuggability if they matched.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -401,13 +389,13 @@ func (a *openstackActuator) calculateResourceRequirements(openstackClient openst
...
 	// Track unique flavors to avoid duplicate API calls
-	flavorCache := make(map[string]*openstackclient.FlavorDetails)
+	flavorCache := make(map[string]*flavors.Flavor) // Changed type
 
 	for _, instance := range instances {
 		requirements.Instances++
 
 		// Get flavor details
-		var flavor *openstackclient.FlavorDetails
+		var flavor *flavors.Flavor // Changed type
```
2uasimojo: Were these `Changed type` comments supposed to be permanent?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -1000,3 +959,107 @@ func (a *openstackActuator) getNetworkIDForCluster(openstackClient openstackclie
...
+	if len(allSnapshots) == 0 {
+		logger.Info("no hibernation snapshots found")
+		return nil
+	}
+
+	// Create map of current snapshot IDs for exclusion
+	currentIDs := make(map[string]bool)
+	for _, id := range currentSnapshotIDs {
+		currentIDs[id] = true
+	}
```
2uasimojo: Looks like you may want to use [sets](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets) from apimachinery. Lots of examples in the hive project.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -1000,3 +959,107 @@ func (a *openstackActuator) getNetworkIDForCluster(openstackClient openstackclie
...
+	existingSnapshots, err := openstackClient.ListImages(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list snapshots: %v", err)
+	}
+
+	var hibernationSnapshots []images.Image
+
+	// Filter snapshots
+	for _, snapshot := range existingSnapshots {
+		if strings.Contains(snapshot.Name, infraID) && strings.Contains(snapshot.Name, "hibernation") {
```
2uasimojo: Is this matching the names generated [here](https://github.com/openshift/hive/pull/2722/commits/cd33a4852803fa0e7c4338ec555aae242f27e680#diff-6cad44ef62c3cf232ce411645cbdda6162f22ce9c3c265dd318f616a61e40843R183)? Is an `infraID` here the same as a `server.Name` there?

Perhaps it makes sense to make two (or more) funcs for snapshot names -- one to generate, another to match -- and put them right next to each other to make it easy for future-us to find, read, and correlate them.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -1000,3 +959,107 @@ func (a *openstackActuator) getNetworkIDForCluster(openstackClient openstackclie
...
+	logger.Infof("old snapshot cleanup completed: %d/%d snapshots deleted", successCount, len(oldSnapshots))
+	return nil
+}
+
+// Helper function to find all hibernation snapshots
+func (a *openstackActuator) findAllHibernationSnapshots(openstackClient openstackclient.Client, infraID string, logger log.FieldLogger) ([]images.Image, error) {
+	ctx := context.Background()
+
+	// List all images
+	existingSnapshots, err := openstackClient.ListImages(ctx, nil)
```
2uasimojo: Looks like an opportunity to make use of the `ListOpts` param to make `ListImages()` return only the relevant images, avoiding the need to filter below. You *could* implement a custom `ListOptsBuilder` for that... but it might be cooler if you tagged the snapshots with the infra ID (and/or a static `hibernation` string) when created; and then you could use the native [Tags](https://github.com/openshift/hive/blob/cd33a4852803fa0e7c4338ec555aae242f27e680/vendor/github.com/gophercloud/gophercloud/openstack/imageservice/v2/images/requests.go#L77) filter here.

[Later] Ugh, I just found out that you can't tag a snapshot on creation; you have to do it afterward. That's terrible, and pretty much kills the above idea, because we can't count on not getting interrupted between those two operations. We could declare failure if that happens; or we could make the whole thing idempotent; but then we still have to have logic to look up the images based on name so we can delete or tag orphans.

[Later later] GDI they really don't want to make it easy to prefilter these results, do they? I'm just worried about scaling here. An `Image` is a biggish struct, and there could potentially be quite a lot of them in a given cloud at a given time, and using no `ListOpts` and then `AllPages()` is guaranteed to give you all of them, which could be a nontrivial amount of memory.

I'm out of time for today, but lmk if you have any ideas for how to make this more efficient.

2uasimojo (commented):
Partial.

pkg/controller/hibernation/openstack_actuator.go
abraverm: @msu8 , I recommend to scratch this part considering the impact.

pkg/openstackclient/client.go
2uasimojo: [Later] I looked a bit deeper, still wanting to cut out some of the intermediary structs, see below.

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,393 @@
...
+	AuthURL           string `yaml:"auth_url"`
+	Username          string `yaml:"username"`
+	Password          string `yaml:"password"`
+	ProjectID         string `yaml:"project_id"`
+	ProjectName       string `yaml:"project_name"`
+	UserDomainName    string `yaml:"user_domain_name"`
+	ProjectDomainName string `yaml:"project_domain_name"`
+	UserDomainID      string `yaml:"user_domain_id"`
+	ProjectDomainID   string `yaml:"project_domain_id"`
+}
```
2uasimojo: Is there a reason we can't get all of this from upstream?

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,393 @@
...
+
+type openstackClient struct {
+	provider      *gophercloud.ProviderClient
+	computeClient *gophercloud.ServiceClient
+	imageClient   *gophercloud.ServiceClient
+	networkClient *gophercloud.ServiceClient
+	credentials   *Credentials
+}
+
```
2uasimojo, line 60: Good practice to include a type validation statement like

```go
var _ Client = &openstackClient{}
```

This will go red if something changes such that `openstackClient` doesn't satisfy the `Client` interface.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+	NetworkID          string            `json:"networkID"`
+	OpenshiftClusterID string            `json:"openshiftClusterID"`
+	Metadata           map[string]string `json:"metadata"`
+}
+
+// Return servers that match the infraID prefix
+func (a *openstackActuator) findInstancesByPrefix(openstackClient openstackclient.Client, prefix string) ([]ServerInfo, error) {
+	ctx := context.Background()
+
+	servers, err := openstackClient.ListServers(ctx, nil)
```
2uasimojo: Okay, so here I think we probably *can* make this more efficient by taking advantage of `opts`. Worth checking with the installer team, but *usually* they tag cloud resources associated with an OpenShift cluster (to make it easier to find them when deprovisioning). For example, AWS resources for an OpenShift cluster are all tagged with `kubernetes.io/cluster/$infraID=owned`.

And, nit, consider spelling it `findInstancesByInfraID(..., infraID string)`

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+
+	logger.Infof("found %d instances still running", len(notStopped))
+	return false, notStopped, nil
+}
+
+// basic server information
+type ServerInfo struct {
+	ID   string
+	Name string
+}
```
2uasimojo: I think I get why you did this -- `Server` itself is kind of big to be lugging around -- but since it seems unavoidable that we're going to be slurping the whole struct across the wire anyway, the only difference it makes is when it gets garbage collected. The duration of one reconcile loop should be short enough that we're not gaining anything by releasing it "early". As long as you're passing around `*Server`, you'll avoid making copies, and then IMO it makes sense to get rid of this type.

...unless I missed somewhere we're saving this struct to persistent storage?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+
+type OpenStackInstanceConfig struct {
```
2uasimojo: A comment on this struct would be nice. It looks like (a list of) it gets marshaled into the hibernation config Secret.

I'm a little bit worried about the versioning of this struct. I haven't thought it all the way through yet, though...

[Later] I'm again wondering why we're inventing our own struct here rather than using the native types returned by the SDK. It seems like `Server` would have most of what we need in order to recreate the instance.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+		logger.Warnf("could not check existing instances: %v", err)
+		existingServers = []ServerInfo{} // Assume none exist
+	}
+
+	// Figure out what we need to create
+	instancesToCreate := a.findMissingInstances(instances, existingServers, logger)
+
+	if len(instancesToCreate) == 0 {
+		logger.Info("all instances already exist - restoration complete")
+		return a.deleteHibernationConfigSecret(cd, hiveClient, logger)
```
2uasimojo: Idempotence warning! Based on the logic below, part of the flow includes:
- Making sure all the instances are active
- Cleaning up the snapshots

As written, if you hit this line, those bits will end up being skipped. Yes, there are flows where they would be redundant... but there are also flows where they wouldn't be :)

There are a number of ways this could be addressed, but an easy one would be to make `createMissingInstances()` no-op "succeed" if given an empty `instancesToCreate`. (It looks like it would already behave that way.)

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+	return a.deleteHibernationConfigSecret(cd, hiveClient, logger)
+}
+
+// Create missing instances during restoration
+func (a *openstackActuator) createMissingInstances(openstackClient openstackclient.Client, instancesToCreate []OpenStackInstanceConfig, logger log.FieldLogger) error {
+	ctx := context.Background()
+	var errors []string
+
+	for i, instance := range instancesToCreate {
+		logger.Infof("creating missing instance %d/%d: %s", i+1, len(instancesToCreate), instance.Name)
```
2uasimojo: nit: prefer "structured logging" -- i.e. using `WithField()`/`WithFields()` for variables rather than embedding them in the string.

(Throughout)

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+			errors = append(errors, errorMsg)
+			logger.Error(errorMsg)
+			continue
+		}
+
+		logger.Infof("created instance %s (ID: %s)", instance.Name, newServer.ID)
+	}
+
+	if len(errors) > 0 {
+		return fmt.Errorf("some instance creation failed: %s", strings.Join(errors, "; "))
```
2uasimojo: Consider using apimachinery's [errors.Aggregate](https://github.com/openshift/hive/blob/f83eba3c8ab4b203e75dfc582d61d4fedf977cd1/vendor/k8s.io/apimachinery/pkg/util/errors/errors.go#L35).

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+		_, err := openstackClient.GetImage(ctx, instance.SnapshotID)
+		if err != nil {
+			errorMsg := fmt.Sprintf("snapshot %s not found for %s: %v", instance.SnapshotID, instance.Name, err)
+			errors = append(errors, errorMsg)
+			logger.Error(errorMsg)
+			continue
+		}
+
+		// Build server creation options with complete metadata
+		createOpts := &servers.CreateOpts{
```
2uasimojo, line 544: Noticed this doesn't include `Tags`. Again, I may be assuming too much about how the installer creates/uses them, but if I'm right, and we recreate the instances without the appropriate tags, then we'll subsequently fail to destroy them.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+
+	if len(snapshotIDs) != len(matchingServers) {
+		return fmt.Errorf("mismatch between servers (%d) and snapshot IDs (%d)", len(matchingServers), len(snapshotIDs))
+	}
+
+	// Get InfraID
+	infraID := cd.Spec.ClusterMetadata.InfraID
+
+	// Get shared configuration
+	networkID, err := a.getNetworkIDForCluster(openstackClient, infraID)
```
2uasimojo, line 666: Is there a reason we can't do lookups like this on the resume side instead?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+	// Get all ports
+	allPorts, err := openstackClient.ListPorts(ctx)
+	if err != nil {
+		return fmt.Errorf("error listing ports: %v", err)
+	}
+
+	// Build configuration for each instance
+	var instanceConfigs []OpenStackInstanceConfig
+	for i, serverInfo := range matchingServers {
+		serverDetails, err := openstackClient.GetServer(ctx, serverInfo.ID)
```
2uasimojo: Pursuant to [comment](https://github.com/openshift/hive/pull/2722/files#diff-6cad44ef62c3cf232ce411645cbdda6162f22ce9c3c265dd318f616a61e40843R672), if we were passing around `*Server` instead of our reduced `ServerInfo`, this call would be unnecessary.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+				break
+			}
+		}
+
+		if portID == "" {
+			return fmt.Errorf("no port found for instance %s", serverInfo.Name)
+		}
+
+		// Get security groups
+		secGroups, err := openstackClient.GetServerSecurityGroupNames(ctx, serverInfo.ID)
```
2uasimojo: Do these security groups persist when the instances are deleted? Can they still be looked up by server ID at that point? My intent: Either we need to recreate the SGs as part of the resume process; or as noted above we would be able to do this lookup on the resume side instead of here.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+
+func init() {
+	RegisterActuator(&openstackActuator{openstackClientFn: getOpenStackClient})
+}
+
+// openstackActuator implements HibernationActuator for OpenStack
+type openstackActuator struct {
+	openstackClientFn func(*hivev1.ClusterDeployment, client.Client, log.FieldLogger) (openstackclient.Client, error)
+}
+
```
2uasimojo, line 34: Consider a type assertion

```go
var _ HibernationActuator = &openstackActuator{}
```

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func init() {
+	RegisterActuator(&openstackActuator{openstackClientFn: getOpenStackClient})
+}
+
+// openstackActuator implements HibernationActuator for OpenStack
+type openstackActuator struct {
+	openstackClientFn func(*hivev1.ClusterDeployment, client.Client, log.FieldLogger) (openstackclient.Client, error)
```
2uasimojo, line 32: Just an idea: rather than passing the client around as a param to every helper func, we could cache it in the actuator struct. It would go something like:

```diff
diff --git a/pkg/controller/hibernation/openstack_actuator.go b/pkg/controller/hibernation/openstack_actuator.go
index 50fbeeb7d..623d6efba 100644
--- a/pkg/controller/hibernation/openstack_actuator.go
+++ b/pkg/controller/hibernation/openstack_actuator.go
@@ -28,6 +28,16 @@ func init() {
 // openstackActuator implements HibernationActuator for OpenStack
 type openstackActuator struct {
        openstackClientFn func(*hivev1.ClusterDeployment, client.Client, log.FieldLogger) (openstackclient.Client, error)
+       openstackClient      openstackclient.Client
+}
+
+func (a *openstackActuator) cacheClient(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) error {
+       if a.openstackClient != nil {
+               return nil
+       }
+       var err error
+       a.openstackClient, err = a.openstackClientFn(cd, c, logger)
+       return err
 }
 
 // Create API client
@@ -63,7 +73,7 @@ func (a *openstackActuator) StopMachines(cd *hivev1.ClusterDeployment, hiveClien
        logger = logger.WithField("cloud", "openstack")
        logger.Info("stopping machines and creating snapshots")
 
-       openstackClient, err := a.openstackClientFn(cd, hiveClient, logger)
+       err := a.cacheClient(cd, hiveClient, logger)
        if err != nil {
                return fmt.Errorf("failed to create OpenStack client: %v", err)
        }
@@ -71,7 +81,7 @@ func (a *openstackActuator) StopMachines(cd *hivev1.ClusterDeployment, hiveClien
        infraID := cd.Spec.ClusterMetadata.InfraID
 
        // 1.
-       matchingServers, err := a.findInstancesByPrefix(openstackClient, infraID)
+       matchingServers, err := a.findInstancesByPrefix(infraID)
        if err != nil {
                return fmt.Errorf("error finding instances: %v", err)
        }
@@ -684,10 +694,10 @@ type OpenStackInstanceConfig struct {
 }
 
 // Return servers that match the infraID prefix
-func (a *openstackActuator) findInstancesByPrefix(openstackClient openstackclient.Client, prefix string) ([]ServerInfo, error) {
+func (a *openstackActuator) findInstancesByPrefix(prefix string) ([]ServerInfo, error) {
        ctx := context.Background()
 
-       servers, err := openstackClient.ListServers(ctx, nil)
+       servers, err := a.openstackClient.ListServers(ctx, nil)
        if err != nil {
                return nil, fmt.Errorf("error listing servers: %v", err)
        }
```

...and similar changes in the other helper funcs.

The tradeoff is tighter coupling: the helpers rely on that cache func having been called, or they'll panic. I'm not sure if it's worth the characters saved by eliminating that extra param :shrug: 

(I would have liked for all the helpers to call something more like `a.getClient().DoWhatever(...)`, and have `getClient()` do the nil check and cache the client if necessary... but then there's no good way to handle the error from the client creation :( )

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1065 @@
...
+	}
+
+	if len(matchingServers) == 0 {
+		logger.Info("no instances found - cluster already hibernated")
+		return nil
+	}
+
+	logger.Infof("found %d instances to hibernate", len(matchingServers))
+
+	if err := a.validateInstanceStates(openstackClient, matchingServers, logger); err != nil {
```
2uasimojo: Again here, I'm not sure it helps us to prevalidate. Since things can happen asynchronously, we would still need to have an error path for when we try to do the real thing and it fails because an instance is "invalid".

Also for idempotence, if we are interrupted partway through, is it possible to hit the next iteration and encounter one or more instances legitimately deleting, because we already snapshotted them last time? [Later] Hm, not as the algorithm is currently written... but I don't think that flow is idempotent...

This is sending me down a mental rabbit hole. Is it possible to somehow "freeze" or "lock" an instance so that it can't be touched by another process while we're working? If so, how would *we* be able to touch it in the aforementioned second-run-through scenario?

This question is actually relevant in terms of losing work as well. Scenario: we create a snapshot while some process is running on the instance. The process continues after our snapshot finishes, but before we delete the instance. When we restore, we'll go back to that old state, having lost whatever progress was made in the interim. Depending on the workload, that may or may not be recoverable.

Can we [Pause()](https://github.com/openshift/hive/blob/c392ca38fb489267cd7bfbdb3c5a76cc36163686/vendor/github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers/requests.go#L1200) the instances as the first step?

2uasimojo (commented):
Partial.

(Sorry, forgot to post this last night; and I've got to pause reviewing this for other priorities today.)

msu8:
Hi, thanks a lot for the amazing feedback and sorry that the following commit took so much tim - some things were a bit difficult for me and I unfortunately had to focus on other work.

I think I addressed all of the feedback except for a couple off points that I think aren't as necessary or would add friction/slowdown - I'll as additional questions monday and will elaborate further. Sort of time pressed.

- Removed quota validation
- Fixed timestamp consistency
- Streamlined snapshot fetching with using already known names
- Removed redundant structs
- Add client creation from upstream clientconfig
- Added instance tagging
  --   {
      "name": "rd-ssegal-01-xczfk-worker-0-q4d76",
      "flavor": "66502c57-663a-4e51-82f0-bdbd9522ace0",
      "portID": "34f4f1f2-37b6-47a1-8558-98239b2e535b",
      "snapshotID": "04a3aaaf-8338-4b30-96de-2e1775a97c7f",
      "snapshotName": "rd-ssegal-01-xczfk-worker-0-q4d76-hibernation-20250912-155926",
      "securityGroups": [
        "rd-ssegal-01-xczfk-worker"
      ],
      "clusterID": "rd-ssegal-01-xczfk",
      "networkID": "f544f0d0-5051-4fd8-acdc-9edcad84f916",
      "openshiftClusterID": "rd-ssegal-01-xczfk",
      "metadata": {
        "Name": "rd-ssegal-01-xczfk-worker",
        "openshiftClusterID": "rd-ssegal-01-xczfk"
      },
      "tags": [ <------
        "cluster-api-provider-openstack",
        "openshift-machine-api-rd-ssegal-01-xczfk",
        "openshiftClusterID=rd-ssegal-01-xczfk"
      ]
    },
- Added pausing/unpausing -> specificially requires a microversion to be specified otherwise it doesn't work
- Other smaller changes like renaming, structure, comments, logging
- Updated docs

msu8:
Hello @2uasimojo, have you had a chance to look at the latest commit?

2uasimojo:
I have been back-burnering this review, possibly because of this

> I'll as additional questions monday and will elaborate further.

and the assumption you would want to fix CI first... or possibly just because it's been consistently usurped by other, noisier work :)

I'll start another pass.

docs/hibernating-clusters.md
```diff
@@ -345,4 +346,17 @@ The unreachable controller tracks the reachability of a cluster and is responsib
 
 Once a cluster hibernates and stops responding, the unreachable controller sets the `Unreachable` condition's `Status` to `True`.
 
-Once the cluster resumes, the unreachable controller set the condition `False` and the hive controllers are free to resume syncing.
\ No newline at end of file
+Once the cluster resumes, the unreachable controller set the condition `False` and the hive controllers are free to resume syncing.
```
2uasimojo: (Latent)
```suggestion
Once the cluster resumes, the unreachable controller sets the condition `False` and the hive controllers are free to resume syncing.
```

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+// openstackActuator implements HibernationActuator for OpenStack
+type openstackActuator struct {
+	openstackClientFn func(*hivev1.ClusterDeployment, client.Client, log.FieldLogger) (openstackclient.Client, error)
+}
+
+var _ HibernationActuator = &openstackActuator{}
+
+// Create API client
+func getOpenStackClient(cd *hivev1.ClusterDeployment, c client.Client, logger log.FieldLogger) (openstackclient.Client, error) {
+	ctx := context.Background()
```
2uasimojo: Nothing wrong with this, but in this project we usually seem to provide the context param inline. Particularly in this case since we're only using it once.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+	secretName := cd.Spec.Platform.OpenStack.CredentialsSecretRef.Name
+	secretNamespace := cd.Namespace
+
+	secret := &corev1.Secret{}
+	err := c.Get(ctx, types.NamespacedName{
+		Name:      secretName,
+		Namespace: secretNamespace,
+	}, secret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get credentials secret %s/%s: %v", secretNamespace, secretName, err)
```
2uasimojo: Prefer `errors.Wrapf()` (throughout).

msu8:
> I have been back-burnering this review, possibly because of this
> 
> > I'll as additional questions monday and will elaborate further.
> 
> and the assumption you would want to fix CI first... or possibly just because it's been consistently usurped by other, noisier work :)
> 
> I'll start another pass.

Oh no, I just saw the dependency issue in the CI logs. Sorry, messed up on my local, I'll let you know once it's sorted out, feel free to focus on the noisier work until I've sorted this out.

go.mod
```diff
@@ -151,10 +151,13 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.30.3 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sts v1.34.0
 	github.com/aws/smithy-go v1.22.4
+	github.com/gophercloud/gophercloud v1.14.1
```
2uasimojo, line 154: nit: Any chance we could group all the gophercloud deps together?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+		serverLogger := logger.WithFields(log.Fields{
+			"instance": server.Name,
+			"status":   server.Status,
+		})
+
+		serverLogger.Info("instance status")
+
+		// Check for deleting states that would cause conflicts
+		if strings.Contains(strings.ToLower(server.Status), "delet") {
+			return fmt.Errorf("cannot hibernate: instance %s is being deleted by another process", server.Name)
```
2uasimojo: As written this will only report the first instance that's in a bad state. Do we maybe want to accumulate these errors so we can report them all at once? E.g. via an [Aggregate](https://github.com/openshift/hive/blob/f83eba3c8ab4b203e75dfc582d61d4fedf977cd1/vendor/k8s.io/apimachinery/pkg/util/errors/errors.go#L46).

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+	for _, server := range servers {
+		serverLogger := logger.WithFields(log.Fields{
+			"instance": server.Name,
+			"status":   server.Status,
+		})
+
+		serverLogger.Info("instance status")
+
+		// Check for deleting states that would cause conflicts
+		if strings.Contains(strings.ToLower(server.Status), "delet") {
```
2uasimojo, line 317: I'm guessing we could be deleting instances as part of a machineset scaling activity, right? That could be user-driven (e.g. via hive's MachinePools) or via the autoscaler. In such cases, do we really want to block the hibernation, or simply ignore that instance?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+		})
+
+		serverLogger.Info("pausing instance for snapshot consistency")
+
+		err := openstackClient.PauseServer(ctx, server.ID)
+		if err != nil {
+			// If pause fails, try to unpause any previously paused instances
+			if i > 0 {
+				logger.Warn("pause failed - attempting to unpause previously paused instances")
+				a.unpauseInstances(openstackClient, servers[:i], logger)
```
2uasimojo: I'd really like to aim for a more idempotent design. I looked at our other actuators to confirm: none of them try to recover from an in-band failure like this. (Granted none of them have multiple steps associated with hibernating an instance, but still.)

The preferred architecture is to bubble up an error immediately and have the controller designed so the next pass through attempts to pick up where the previous left off. If that flow gets stuck, you should be able to flip the desired state and have the _resume_ pick up at the appropriate point and move toward that desired state.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+// Unpause instances
+func (a *openstackActuator) unpauseInstances(openstackClient openstackclient.Client, servers []*servers.Server, logger log.FieldLogger) {
+	ctx := context.Background()
+
+	for _, server := range servers {
+		serverLogger := logger.WithField("instance", server.Name)
+		serverLogger.Info("unpausing instance")
+
+		err := openstackClient.UnpauseServer(ctx, server.ID)
+		if err != nil {
```
2uasimojo: Heh, I noticed you aren't trying to re-pause if an unpause fails. Exercise for the reader why that would be Bad™ :innocent:

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+	if err := a.pauseInstances(openstackClient, matchingServers, logger); err != nil {
+		return fmt.Errorf("failed to pause instances: %v", err)
+	}
+
+	// 4. Create snapshots
+	snapshotIDs, snapshotNames, err := a.createSnapshots(openstackClient, matchingServers, logger)
+	if err != nil {
+		// If snapshot creation fails, unpause
+		logger.Warn("snapshot creation failed - unpausing instances")
+		a.unpauseInstances(openstackClient, matchingServers, logger)
```
2uasimojo: I get that you're trying to "unwind" here to clean up what's been done previously. This pattern gets awkward fast, as you noticed by the fact that you're having to `unpause` in three different branches here, and should arguably be doing the same for the other error conditions.

IF we wanted to use this architecture, a better approach might be to push cleanup funcs onto a stack consumed by a `defer cleanup()`.

HOWEVER, as mentioned elsewhere, the k8s controller paradigm is to make the actual state match the declared state, which in this case is "Hibernating". If something goes wrong, it is up to the next pass to redrive the operation from where it left off the last time, which could be _anywhere_.  (This is in contrast to a more classical operational approach which is like, "Do operation X, and if operation X fails, try to leave the system in a good state if possible".)

In this case, let's say for example that the third of five instances failed to `pause` because of a network glitch that persists long enough for (some of?!) the `unpause`s to fail. The next reconcile will still be attempting to make the actual state "Hibernating", which means it'll get here and try to make all the instances paused, which means it needs to handle the case where _any number of instances_ are already(/still) paused. And since we're going to do that anyway, immediately (modulo backoff if we hot loop, see below), there's really no reason to attempt to revert in-band like this. (In fact, the worst case is if it _works_ -- then we end up unpausing instances we're going to immediately pause again!)

Looking at error conditions from a more zoomed-out perspective (that is, across however many reconcile iterations it takes to get to a "terminal" state -- one that persists until something changes), we typically want to end up in one of two places:
- Hot loop, which will result in exponential backoff. The actual state is "dirty": not consistent with the desired state and not whatever it was before the current desired state was asserted. This is most appropriate for error conditions that are temporary, and/or whose remedy we can't (reasonably) detect. Like in the network glitch example: maybe we're waiting for someone to plug a cable back in; but we're not going to get notified when that happens, so we effectively poll.
- Set a status condition and "exit" (return from the reconcile without requeueing). Again, the actual state remains "dirty". This is good for when the error condition might be remedied by a user action that we can detect. For example, if we start getting creds failures, and we're `Watch()`ing the Secret housing those creds, we can safely exit, since we'll be triggered again when that Secret is updated.

In summary: I would like to see the flows in this actuator designed with idempotence / eventual consistency / drive-toward-desired-state in mind.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+	OpenshiftClusterID string            `json:"openshiftClusterID"`
+	Metadata           map[string]string `json:"metadata"`
+	Tags               []string          `json:"tags"`
+}
+
+// Return servers that match the infraID prefix
+func (a *openstackActuator) findInstancesByInfraID(openstackClient openstackclient.Client, prefix string) ([]*servers.Server, error) {
+	ctx := context.Background()
+
+	allServers, err := openstackClient.ListServers(ctx, nil)
```
2uasimojo: I think I mentioned this [previously](https://github.com/openshift/hive/pull/2722#discussion_r2274304095): can we prefilter using `Tags`?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1091 @@
...
+	ctx := context.Background()
+
+	for i, server := range servers {
+		logger.WithFields(log.Fields{
+			"current":  i + 1,
+			"total":    len(servers),
+			"instance": server.Name,
+		}).Info("deleting instance")
+
+		err := openstackClient.DeleteServer(ctx, server.ID)
```
2uasimojo: Do you know if this, and the other write-y API operations, are idempotent? That is, what happens if you `DeleteServer()` a server that's already deleting/deleted? This will still work if it errors, but it'll result in at least one (error log and) additional reconcile while we wait for the server to disappear from the list returned by `findInstancesByInfraID()`. In which case we might, depending how difficult it is, try to detect and ignore those errors.

2uasimojo (commented):
I'm grinding my way through this very slowly. I'll leave this week's partial review now, as I'm going on PTO until next Tuesday.

The main thing I'd like to see addressed is idempotence. I wrote a lot of words about it inline. Please let me know if anything is unclear, or if you would like to set up a live discussion.

pkg/controller/hibernation/openstack_actuator.go
msu8: Ok, I think I understand, good explanation.

I'll try to commit something near the end of this week, just got back from vacation and moved right into a new team, hopefully I'll have enough time. Worst case Monday.

pkg/controller/hibernation/openstack_actuator.go
msu8: I'm not sure whether this wouldn't add unnecessary friction, i.e. I'd love to use the already fetched information as much as possible and not to que anything else.

OR

I may not be understanding this correctly - but we have the infraid that is exclusive, what would be the benfit of using other parameter, instead of the infraid?

2uasimojo: I'm suggesting that we can use the `ListOpts` param to ListServers() to filter the results on the server side, thus reducing our resource usage (network traffic, client-side filtering). Again, I'm not positive how (or even if!) servers are tagged by installer, but they are for other platforms. So my suggestion would be to do something like:
```suggestion
	allServers, err := openstackClient.ListServers(ctx, &servers.ListOpts{
		// I would rename `prefix` to `infraID`
		Tags: prefix,
	})
```
...and then you could get rid of LL615-20.

msu8:
I though to have commit monday but I'm waiting to regain access to my old team's openstack instance to test my code, lost it after my team change...

msu8:
Hello, finally pushed new changes.

- idempotency should be addressed
- aggregating errors and general error handling fixed too as per the feedback, no rollbacks
- With regards to prefiltering with tags, I'm fetching the instances by the tags everywhere it's possible, however some functions still require to look up the instances via the prefix, this is because the tagging (after restoration) happens after the instances are created and all the other options/metadata are added. Adding the "tag" in openstack is unfortunately done individually and isn't an "option" when the instance is created.

waiting for the jobs to finish.

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,307 @@
...
+		ProjectName:       cloud.AuthInfo.ProjectName,
+		UserDomainName:    cloud.AuthInfo.UserDomainName,
+		ProjectDomainName: cloud.AuthInfo.ProjectDomainName,
+	}
+
+	opts := &clientconfig.ClientOpts{
+		AuthInfo:     authInfo,
+		RegionName:   cloud.RegionName,
+		EndpointType: cloud.Interface,
+	}
```
2uasimojo, line 276: I'm curious why this can't be
```suggestion
	opts := &clientconfig.ClientOpts{
		AuthInfo:     cloud.AuthInfo,
		RegionName:   cloud.RegionName,
		EndpointType: cloud.Interface,
	}
```

Does it have something to do with needing to scrub out the other fields of `cloud.AuthInfo`?

pkg/openstackclient/client.go
```diff
@@ -0,0 +1,307 @@
...
+	// Get the first cloud (or specific cloud name)
+	var cloud *clientconfig.Cloud
+	for _, c := range clouds.Clouds {
+		cloud = &c
+		break
+	}
+
+	if cloud == nil {
+		return nil, fmt.Errorf("no cloud configuration found in clouds.yaml")
+	}
```
2uasimojo, line 260: No change requested here, just wanted to note this interesting pattern, new to my eyes. I'm used to seeing either
```suggestion
	// Get the first cloud (or specific cloud name)
	var cloud *clientconfig.Cloud
	if len(clouds.Clouds) == 0 {
		return nil, fmt.Errorf("no cloud configuration found in clouds.yaml")
	}
	cloud = &clouds.Clouds[0]
```

or
```suggestion
	// Get the first cloud (or specific cloud name)
	var cloud *clientconfig.Cloud
	found := false
	for _, c := range clouds.Clouds {
		cloud = &c
		found = true
		break
	}

	if !found {
		return nil, fmt.Errorf("no cloud configuration found in clouds.yaml")
	}
```

Your way is certainly more elegant than the latter; and I never liked the hardcoded `[0]` of the former.

(I'm going to go out on a limb and say you're a python dev, and were really wishing you had `for`/`else` here :P )

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to delete hibernation config secret")
+	}
+
+	logger.Infof("deleted hibernation config secret %s", secretName)
+	return nil
+}
+
+// Cleanup old snapshots
+func (a *openstackActuator) cleanupOldSnapshots(cd *hivev1.ClusterDeployment, hiveClient client.Client, openstackClient openstackclient.Client, infraID string, currentSnapshots map[string]string, logger log.FieldLogger) error {
```
2uasimojo, line 889: Unused params
```suggestion
func (a *openstackActuator) cleanupOldSnapshots(openstackClient openstackclient.Client, infraID string, currentSnapshots map[string]string, logger log.FieldLogger) error {
```

pkg/controller/hibernation/openstack_actuator.go
2uasimojo: :arrow_up: QE test case: hibernate, resume, destroy; make sure we don't leak instances.

It looks like instances could wind up untagged if we error (unrecoverably) between `createMissingInstances` and `restoreInstanceTags`. So the [question](https://redhat-internal.slack.com/archives/CH98TDJUD/p1762469658651789) is: does the destroyer search by tags exclusively, or does it use prefixes instead/also?

(Later) It does use tags exclusively, so this risk does exist. However, I feel it's acceptable as long as this method idempotently continues to attempt tagging if it fails initially. Would also like to see a comment explaining this hole.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+}
+
+// Return true if this actuator can handle the given ClusterDeployment
+func (a *openstackActuator) CanHandle(cd *hivev1.ClusterDeployment) bool {
+	return cd.Spec.Platform.OpenStack != nil
+}
+
+// StopMachines
+func (a *openstackActuator) StopMachines(cd *hivev1.ClusterDeployment, hiveClient client.Client, logger log.FieldLogger) error {
+	logger = logger.WithField("cloud", "openstack")
```
2uasimojo, line 80: Meh, this may just be noise. If we're looking at these logs, chances are we already know this is an openstack cluster. (Not going to hold up the PR for this though.)

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+
+	if successCount > 0 {
+		logger.Infof("deleted %d restoration snapshots", successCount)
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+// Using tags to filter the instances
+func (a *openstackActuator) findInstancesByInfraID(openstackClient openstackclient.Client, infraID string) ([]*servers.Server, error) {
```
2uasimojo, line 966: Since both `findInstances...()` funcs are finding instances by infra ID, suggest renaming this one:
```suggestion
func (a *openstackActuator) findInstancesByTag(openstackClient openstackclient.Client, infraID string) ([]*servers.Server, error) {
```

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+
+	// Wait for all instances to be active
+	if err := a.waitForAllInstancesToBeActive(openstackClient, infraID, len(instances), logger); err != nil {
+		return errors.Wrap(err, "not all instances are active yet")
+	}
+
+	// Restore tags after instances are active
+	if err := a.restoreInstanceTags(openstackClient, instances, logger); err != nil {
+		logger.Warnf("failed to restore some tags: %v", err)
+	}
```
2uasimojo, line 220: Since, as discussed elsewhere, we've got this window where servers don't have tags, we should do whatever we can to make that window as small as possible. At this point in the flow, that would mean instead of waiting for all instances, then tagging all instances, we could tag each instance as it becomes active.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+		}
+	}
+}
+
+// Restore instance tags
+func (a *openstackActuator) restoreInstanceTags(openstackClient openstackclient.Client, instances []OpenStackInstanceConfig, logger log.FieldLogger) error {
+	var errs []error
+
+	for _, instance := range instances {
+		if len(instance.Tags) == 0 {
```
2uasimojo, line 624: This should never happen, right? We should at least have the installer-generated tags?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+
+// openstackActuator implements HibernationActuator for OpenStack
+type openstackActuator struct {
+	openstackClientFn func(*hivev1.ClusterDeployment, client.Client, log.FieldLogger) (openstackclient.Client, error)
+}
+
+var _ HibernationActuator = &openstackActuator{}
+
+// OpenStackInstanceConfig stores instance configuration for restoration
+type OpenStackInstanceConfig struct {
```
2uasimojo, line 38: I think "instance" and "server" are the same thing, right? Can we be consistent about the naming? Like, this would be OpenStackServerConfig.

And wherever we're currently passing them around in variables called `instance`, call them `serverConfig` instead (i.e. be explicit that it's the config, not the actual server object).

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+
+	if len(matchingServers) == 0 {
+		logger.Info("no instances found - cluster already hibernated")
+		return nil
+	}
+
+	logger.WithField("count", len(matchingServers)).Info("found instances to hibernate")
+
+	// Step 2: Validate instance states
+	if err := a.validateInstanceStates(matchingServers, logger); err != nil {
```
2uasimojo, line 104: Discussed previously: I think we're not getting value from this validation -- let's scrap it.

pkg/controller/hibernation/openstack_actuator.go
2uasimojo, line 38: No reason for this to be exported (capitalized), is there?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+			continue
+		}
+
+		if current.Status == "PAUSED" {
+			serverLogger.Info("instance already paused")
+			continue
+		}
+
+		serverLogger.Info("pausing instance")
+		if err := openstackClient.PauseServer(context.Background(), server.ID); err != nil {
```
2uasimojo, line 354: Is this synchronous?

If not, we need a way to make sure we don't proceed to the next step (`createSnapshots()`) untill all servers are paused.

And if so, I fear for the performance of this algorithm. We would benefit from _making_ it asynchronous (on our end, via goroutines)... and then still adding logic to ensure all servers are paused before proceeding.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+
+	for _, server := range servers {
+		snapshotName := fmt.Sprintf("%s-hibernation-%s", server.Name, timestamp)
+		serverLogger := logger.WithFields(log.Fields{
+			"instance":      server.Name,
+			"snapshot_name": snapshotName,
+		})
+
+		// Check if snapshot already exists
+		existing, err := a.findSnapshotByName(openstackClient, snapshotName)
```
2uasimojo, line 381: How would this ever find an existing snapshot, given you've constructed the name with a `Now()` timestamp?

...but I'm thinking about the problem you're trying to solve, which presumably is that you don't want to accidentally use a snapshot from a previous hibernation that didn't get scrubbed by `cleanupOldSnapshots()` for whatever reason.

And for idempotence, we can't *not* tie a snapshot to an individual hibernation event.

What we need is a way to persist a marker that's guaranteed to be unique to every hibernation event. A timestamp would work as long as we generate it only once per hibernation event vs every reconcile as written. The problem is that StopMachines has to have a way to know the marker is for *this* event. We can't even use the CD's `generation`, as nothing's stopping the CD getting updated while the hibernation is ongoing.

My best idea: use the `lastTransitionTime` of the `Ready` status condition.

(Background: the hibernation controller maintains two status fields during the hibernate and resume processes: [Ready](https://github.com/openshift/hive/blob/fb278f83591b508c097679b39d0704103be006d4/vendor/github.com/openshift/hive/apis/hive/v1/clusterdeployment_types.go#L468) and [Hibernating](https://github.com/openshift/hive/blob/fb278f83591b508c097679b39d0704103be006d4/vendor/github.com/openshift/hive/apis/hive/v1/clusterdeployment_types.go#L464). `Ready` flips once at the beginning of hibernate, and `Hibernating` updates several times to track sub-steps in the process. When resuming, it's the opposite: `Hibernating` flips at the beginning and `Ready` tracks the sub-steps.)

Whatever we choose, we would have to save it in a CR field as soon as we generate it. And for DR purposes, it shouldn't be in `status`. For the sake of brainstorming this design, let's say it'll be `cd.spec.platform.openstack.hibernationSnapshotSuffix`.

- I'm thinking `cleanupOldSnapshots()` is responsible for clearing the field once *all snapshots* are gone.
- At the beginning of `StopMachines()`:
   - If the field is already set, make sure it matches the current `lastTransitionTime` of the `Ready` condition.
      - If it doesn't match, call `cleanupOldSnapshots()` and requeue. (I'm not sure it's possible for this to occur under normal circumstances... but maybe if you resume in mid-hibernate, and then re-hibernate??)
      - If it does match, it means we've re-entered the current hibernation event; proceed.
   - If the field is unset, it means we're starting a hibernation event. Cool, proceed.
- In `StopMachines()`, (maybe?) wait until after all machines are stopped and...
      - cull the desired value from the `Ready` condition
      - set it in the CD.spec
      - requeue.
- Now when `createSnapshots()` is called from `StopMachines()` it can confidently use the value of that field to construct the names of the snapshots, and idempotently create any snapshots not already extant.
- Now `StartMachines()` knows which snapshots to use. (Though I guess you're saving the snapshot IDs in the config object.)
- At the end of `StartMachines()` -- possibly even before we allow that func to declare success -- we call `cleanupOldSnapshots()`, which deletes the snapshots *and* clears the field.

WDYT?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+		return err
+	}
+
+	// Step 8: Wait for cleanup
+	if err := a.waitForInstanceCleanup(openstackClient, infraID, logger); err != nil {
+		return err
+	}
+
+	// Step 9: Best-effort cleanup of old snapshots
+	if err := a.cleanupOldSnapshots(cd, hiveClient, openstackClient, infraID, snapshotMapping, logger); err != nil {
```
2uasimojo, line 140: Seems like an odd place for this. Suggest we call it at the end of `StartMachines()`, which is where we stop needing the snapshots (right?).

Maybe also call it right before `createSnapshots()` above to make sure we're starting clean?

But see also my [other comment](https://github.com/openshift/hive/pull/2722/changes#r2756164210) around snapshot naming.

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+		if snapshot.Name == snapshotName {
+			return &snapshot, nil
+		}
+	}
+
+	return nil, errors.New("snapshot not found")
+}
+
+// Helper utilities
+func isNotFoundError(err error) bool {
```
2uasimojo, line 1051: I noticed you've used the [upstream func](https://github.com/openshift/hive/blob/22034ff39aac4711c0703f55f39609106e652c26/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L528) in several places -- is there a reason it doesn't work all the time?

pkg/controller/hibernation/openstack_actuator.go
```diff
@@ -0,0 +1,1053 @@
...
+	}
+
+	// Restore tags after instances are active
+	if err := a.restoreInstanceTags(openstackClient, instances, logger); err != nil {
+		logger.Warnf("failed to restore some tags: %v", err)
+	}
+
+	// Cleanup hibernation snapshots
+	logger.Info("cleaning up hibernation snapshots after successful restoration")
+	if err := a.cleanupRestorationSnapshots(openstackClient, instances, logger); err != nil {
```
2uasimojo, line 224: If we go with the algorithm described [here](https://github.com/openshift/hive/pull/2722/changes#r2756164210), we can probably scrap this special-purpose scrubber.

2uasimojo (commented):
With apologies again for the ridiculous amount of time between reviews....

2uasimojo:
The players involved are aware of the following, but for visibility:

@msu8, thank you very much for all your hard work on this feature, and for your patience with the review cycle. Unfortunately we've been finding that the architectural challenges outweigh the amount of engineering resource we're collectively able to bring to bear to get it over the finish line. Hence we're going to shelve it for now.

If anyone out there has been burning for this feature, please leave a comment. If nothing else, it will help us gauge demand.

PR #2725, opened 2025-07-31
HIVE-2891: Make hive able to run as a private image

dlom:

/assign @2uasimojo 

(previous PR #2724 referenced wrong branch)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -319,11 +319,8 @@ type ReconcileClusterDeployment struct {
 
 	protectedDelete bool
 
-	// nodeSelector is copied from the hive-controllers pod and must be included in any Jobs we create from here.
-	nodeSelector *map[string]string
-
-	// tolerations is copied from the hive-controllers pod and must be included in any Jobs we create from here.
-	tolerations *[]corev1.Toleration
+	// sharedPodConfig is copied from the hive-controllers pod and must be included in any Jobs we create from here.
+	sharedPodConfig *controllerutils.SharedPodConfig
 }
 
 // Reconcile reads that state of the cluster for a ClusterDeployment object and makes changes based on the state read
@@ -331,16 +328,15 @@ type ReconcileClusterDeployment struct {
 func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reconcile.Request) (result reconcile.Result, returnErr error) {
 	cdLog := controllerutils.BuildControllerLogger(ControllerName, "clusterDeployment", request.NamespacedName)
 
-	// Discover scheduling settings from the controller. We would like to do this in NewReconciler,
+	// Discover settings from the controller. We would like to do this in NewReconciler,
 	// but we can't count on the cache having been started at that point.
-	if r.nodeSelector == nil || r.tolerations == nil {
-		thisPod, err := controllerutils.GetThisPod(r)
+	if r.sharedPodConfig == nil {
+		sharedPodConfig, err := controllerutils.ReadSharedConfigFromThisPod(r)
 		if err != nil {
-			cdLog.WithError(err).Error("Failed to retrieve the running pod")
+			cdLog.WithError(err).Error("error reading shared pod config")
 			return reconcile.Result{}, err
 		}
-		r.nodeSelector = &thisPod.Spec.NodeSelector
-		r.tolerations = &thisPod.Spec.Tolerations
+		r.sharedPodConfig = sharedPodConfig
 	}
 
 	cdLog.Info("reconciling cluster deployment")
@@ -1104,8 +1100,8 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 			os.Getenv("HTTP_PROXY"),
 			os.Getenv("HTTPS_PROXY"),
 			os.Getenv("NO_PROXY"),
-			*r.nodeSelector,
-			*r.tolerations)
+			*r.sharedPodConfig,
+		)
 
 		cdLog.WithField("derivedObject", job.Name).Debug("Setting labels on derived object")
 		job.Labels = k8slabels.AddLabel(job.Labels, constants.ClusterDeploymentNameLabel, cd.Name)
--- a/pkg/controller/utils/podconfig.go
+++ b/pkg/controller/utils/podconfig.go
@@ -1,0 +1,46 @@
+package utils
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SharedPodConfig is initially created from the hive-operator pod and must be included in any Jobs or Pods
+// we create from downstream.
+type SharedPodConfig struct {
+	NodeSelector     map[string]string
+	Tolerations      []corev1.Toleration
+	ImagePullSecrets []corev1.LocalObjectReference
+}
+
+func ReadSharedConfigFromThisPod(cl client.Client) (*SharedPodConfig, error) {
+	thisPod, err := getThisPod(cl)
+	if err != nil {
+		return nil, err
+	}
+
+	return &SharedPodConfig{
+		NodeSelector:     thisPod.Spec.NodeSelector,
+		Tolerations:      thisPod.Spec.Tolerations,
+		ImagePullSecrets: thisPod.Spec.ImagePullSecrets,
+	}, nil
+}
+
+// getThisPod gets this pod.  If this function is useful somewhere else, make it public again
+func getThisPod(cl client.Client) (*corev1.Pod, error) {
+	podName := os.Getenv("POD_NAME")
+	podNamespace := os.Getenv("POD_NAMESPACE")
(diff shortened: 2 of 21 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -333,14 +336,15 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
 
 	// Discover scheduling settings from the controller. We would like to do this in NewReconciler,
```
2uasimojo: nit, the new thing isn't a scheduling setting, so perhaps 
```suggestion
	// Discover settings from the controller. We would like to do this in NewReconciler,
```

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -333,14 +336,15 @@ func (r *ReconcileClusterDeployment) Reconcile(ctx context.Context, request reco
...
-	if r.nodeSelector == nil || r.tolerations == nil {
+	if r.nodeSelector == nil || r.tolerations == nil || r.imagePullSecrets == nil {
 		thisPod, err := controllerutils.GetThisPod(r)
 		if err != nil {
 			cdLog.WithError(err).Error("Failed to retrieve the running pod")
 			return reconcile.Result{}, err
 		}
 		r.nodeSelector = &thisPod.Spec.NodeSelector
 		r.tolerations = &thisPod.Spec.Tolerations
+		r.imagePullSecrets = &thisPod.Spec.ImagePullSecrets
```
2uasimojo: So, wait, do the pull secrets get funneled automatically from the ServiceAccount into the pod spec by... whatever thingy processes the Deployment?

[Later] Ah, [yup](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#verify-that-imagepullsecrets-are-set-for-new-pods), cool. We should definitely leave comments to this effect wherever future-we would end up scratching our heads over this. Include the above link in said comments.

Umm, can we take advantage of this for the imageset/provision/deprovision pods? What SA do they run under?

[Later again] Looks like imageset & installer both run under the `cluster-installer` SA, configured via [SetupClusterInstallServiceAccount()](https://github.com/openshift/hive/blob/8dfac7acf50063ab5233cc1d81fd05841b393ed6/pkg/controller/utils/sa.go#L77); and uninstaller runs under `cluster-uninstaller`, configured via [SetupClusterUninstallServiceAccount()](https://github.com/openshift/hive/blob/8dfac7acf50063ab5233cc1d81fd05841b393ed6/pkg/controller/utils/sa.go#L96).

I don't know that there's any advantage to switching. And in fact the advantage to doing it the way you're doing it is that we don't have to add continuous reconciliation the SAs (currently they're only created, not updated). If the name of the pull secret changes, the user will have to change it in the hive-operator Deployment, which will re-roll all the controllers, and any new pods would pick up the controller's value, vs picking up the possibly-stale value from the SA.

But now knowing that these two possibilities exist, I wonder if we should inject the imagePullSecrets directly into the hive-controllers/hiveadmission pod specs rather than via their SAs. By the above logic, it won't make any functional difference; but it might make the code easier to understand. Thoughts?

pkg/operator/hive/hive_controller.go
```diff
@@ -306,6 +306,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		tolerations := operatorDeployment.Spec.Template.Spec.Tolerations
 		log.Debugf("loaded tolerations from hive-operator deployment: %v", tolerations)
 		r.(*ReconcileHiveConfig).tolerations = tolerations
+		imagePullSecrets := operatorDeployment.Spec.Template.Spec.ImagePullSecrets
+		log.Debugf("loaded imagePullSecrets from hive-operator deployment: %v", imagePullSecrets)
```
2uasimojo: Does this `%v` expand to make the `LocalObjectReference.Name`z visible in the log message? [Later] [Yah](https://go.dev/play/p/Dur_WquSlA2), cool.

Also, nit, prefer structured logging (e.g. `WithField()`) vs embedding. (But you're following the precedent... which is probably mine :bashful:)

pkg/operator/util/apply.go
```diff
@@ -66,6 +67,21 @@ func WithNamespaceOverride(namespaceOverride string) RTOApplyOpt {
...
+// WithImagePullSecrets sets the image pull secrets of a service account.
+// It will not choke on non-service account objects, instead it will log a debug message
+// and skip over it.
+func WithImagePullSecrets(imagePullSecrets []corev1.LocalObjectReference) RTOApplyOpt {
+	return func(runtimeObj runtime.Object, hLog log.FieldLogger) error {
+		sa, ok := runtimeObj.(*corev1.ServiceAccount)
+		if !ok {
+			hLog.Debugf("runtime object is not a ServiceAccount, skipping")
+			return nil
+		}
```
2uasimojo: Eh, I understand why you did this, but this lenience will end up hiding errors if they crop up. I would prefer to see the `namespacedAssets` split up so we're only ever _trying_ to apply the image pull secrets to ServiceAccounts, and we can raise an error if `!ok` in this context.

pkg/install/generate.go
```diff
@@ -442,6 +444,7 @@ func GenerateInstallerJob(
 	podSpec := provision.Spec.PodSpec
 	podSpec.NodeSelector = nodeSelector
 	podSpec.Tolerations = tolerations
+	podSpec.ImagePullSecrets = append(podSpec.ImagePullSecrets, imagePullSecrets...)
```
2uasimojo: Wondering why I put these here instead of in [InstallerPodSpec()](https://github.com/openshift/hive/blob/853824843add16752214c1db8876c5b99bcf3979/pkg/install/generate.go#L154). It's probably because we can't be assured the ClusterProvision is fresh. This makes sense in that light.

2uasimojo (commented):
Looks like you need to regenerate bindata.

pkg/install/generate.go
dlom: Yes, I think so.  Both codepaths are hit eventually

dlom:
@2uasimojo this is ready for a round-2 review

pkg/controller/utils/podconfig.go
```diff
@@ -0,0 +1,27 @@
...
+// SharedPodConfig is initially created from the hive-operator pod and must be included in any Jobs or Pods
+// we create from downstream.
+type SharedPodConfig struct {
+	NodeSelector     map[string]string
+	Tolerations      []corev1.Toleration
+	ImagePullSecrets []corev1.LocalObjectReference
+}
+
+func ReadSharedConfigFromThisPod(cl client.Client) (*SharedPodConfig, error) {
+	thisPod, err := GetThisPod(cl)
```
2uasimojo: `GetThisPod()` can now be private, or even inlined in this func.

pkg/operator/hive/hive_controller.go
```diff
@@ -302,10 +303,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		r.(*ReconcileHiveConfig).hiveImagePullPolicy = pullPolicy
 		nodeSelector := operatorDeployment.Spec.Template.Spec.NodeSelector
 		log.Debugf("loaded nodeSelector from hive-operator deployment: %v", nodeSelector)
-		r.(*ReconcileHiveConfig).nodeSelector = nodeSelector
 		tolerations := operatorDeployment.Spec.Template.Spec.Tolerations
 		log.Debugf("loaded tolerations from hive-operator deployment: %v", tolerations)
-		r.(*ReconcileHiveConfig).tolerations = tolerations
+		imagePullSecrets := operatorDeployment.Spec.Template.Spec.ImagePullSecrets
```
2uasimojo: Does this work? I wouldn't have expected the Deployment to grow imagePullSecrets from an associated ServiceAccount -- I would have thought that would only happen at the Pod level.

We could probably get rid of this Deployment `Get()` and reuse `ReadSharedConfigFromThisPod()` instead. (We would need to add pod ns/name env [injection](https://github.com/openshift/hive/blob/66a6c9f9dc9184818547b9cc16855e8b7ee1c525/config/controllers/deployment.yaml#L50-L57).)

dlom:
@2uasimojo I've got a test build of this up and running on the team cluster.  A spoke is being provisioned, and the installer pod has inherited all of the proper pull secrets.  I'm worried that because the imageset runs in the cluster namespace, it won't have access to the same pull secrets (which I believe are all scoped by ACM into the hive namespace, that's why they're only referenced by name and not NS)

dlom:
Actually, those will all be the installer image I think, so those might need a different set of pull secrets?? There might still be another layer of elephants to go down

2uasimojo:
The release, installer, and CLI images are authed via the pull secret [referenced](https://github.com/openshift/hive/blob/8707a792ccd2d50dd752954ade8c0057bdc50501/vendor/github.com/openshift/hive/apis/hive/v1/clusterdeployment_types.go#L125) from the CD, potentially augmented with the contents of the global pull secret set up via [hiveconfig](https://github.com/openshift/hive/blob/a7b585d29dd24119825dff3c99a03082c4cf7c8c/vendor/github.com/openshift/hive/apis/hive/v1/hiveconfig_types.go#L42). That plumbing is already copacetic -- you don't need to fold this new thing into it.

2uasimojo:
Yeah, this looks nice and clean now, nice work.

/lgtm

PR #2726, opened 2025-08-02
HIVE-2898: Fix cleanup of leftover resources for VPC Endpoints and VPCEndpointServices

huangmingxia:
What does this PR do?

Due to incomplete cleanup of associated VPC Endpoints and VPCEndpointServices, deletion operations fail with resource in use errors, thereby blocking cluster deprovisioning.

### DIFF
--- a/pkg/controller/awsprivatelink/cleanup.go
+++ b/pkg/controller/awsprivatelink/cleanup.go
@@ -105,11 +105,11 @@ func (r *ReconcileAWSPrivateLink) cleanupPrivateLink(cd *hivev1.ClusterDeploymen
 		logger.WithError(err).Error("error cleaning up Hosted Zone")
 		return err
 	}
-	if err := r.cleanupVPCEndpoint(awsClient.hub, metadata, logger); err != nil {
+	if err := r.cleanupVPCEndpoints(awsClient.hub, metadata, logger); err != nil {
 		logger.WithError(err).Error("error cleaning up VPCEndpoint")
 		return err
 	}
-	if err := r.cleanupVPCEndpointService(awsClient.user, metadata, logger); err != nil {
+	if err := r.cleanupVPCEndpointServices(awsClient.user, metadata, logger); err != nil {
 		logger.WithError(err).Error("error cleaning up VPCEndpoint Service")
 		return err
 	}
@@ -201,7 +201,7 @@ func (r *ReconcileAWSPrivateLink) cleanupHostedZone(awsClient awsclient.Client,
 
 }
 
-func (r *ReconcileAWSPrivateLink) cleanupVPCEndpoint(awsClient awsclient.Client,
+func (r *ReconcileAWSPrivateLink) cleanupVPCEndpoints(awsClient awsclient.Client,
 	metadata *hivev1.ClusterMetadata,
 	logger log.FieldLogger) error {
 	idLog := logger.WithField("infraID", metadata.InfraID)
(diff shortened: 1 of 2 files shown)


2uasimojo:
This is looking good.
1. Do we need similar for HostedZone cleanup?
2. Can we please get some UT coverage added for these paths?

huangmingxia:
Thank you both for the review!

> Do we need similar for HostedZone cleanup?

No. For each apiDomain, there can be only one Private HostedZone per VPC in AWS Route53.  AWS enforces this constraint, so the current implementation that only handles one HostedZone is correct.

> Can we please get some UT coverage added for these paths?

Yes, I’m happy to add UT coverage for that.

huangmingxia:
Hi, I’ve added the UT. Could you please review it again when you have time? Thanks!
```
 % go tool cover -func=coverage.out | grep -E "(cleanupVPCEndpoint|Total)"
github.com/openshift/hive/pkg/controller/awsprivatelink/cleanup.go:204:                         cleanupVPCEndpoint                      100.0%
github.com/openshift/hive/pkg/controller/awsprivatelink/cleanup.go:239:                         cleanupVPCEndpointService               100.0%
```

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -216,14 +216,20 @@ func (r *ReconcileAWSPrivateLink) cleanupVPCEndpoint(awsClient awsclient.Client,
...
-	vpcEndpoint := resp.VpcEndpoints[0]
-	endpointLog := logger.WithField("vpcEndpointID", *vpcEndpoint.VpcEndpointId)
+	// Delete all VPC Endpoints found
+	var endpointIds []string
+	for _, vpcEndpoint := range resp.VpcEndpoints {
+		endpointIds = append(endpointIds, *vpcEndpoint.VpcEndpointId)
+	}
+
+	endpointLog := logger.WithField("vpcEndpointIDs", endpointIds)
+	endpointLog.Infof("found %d VPC Endpoints to delete", len(endpointIds))
```
2uasimojo: nit: prefer structured logging
```suggestion
	endpointLog.WithField("numEndpoints", len(endpointIds)).Info("found VPC Endpoints to delete")
```

pkg/controller/awsprivatelink/cleanup.go
```diff
@@ -245,14 +251,20 @@ func (r *ReconcileAWSPrivateLink) cleanupVPCEndpointService(awsClient awsclient.
...
-	service := resp.ServiceConfigurations[0]
-	serviceLog := logger.WithField("vpcEndpointServiceID", *service.ServiceId)
+	// Delete all VPC Endpoint Services found
+	var serviceIds []string
+	for _, service := range resp.ServiceConfigurations {
+		serviceIds = append(serviceIds, *service.ServiceId)
+	}
+
+	serviceLog := logger.WithField("vpcEndpointServiceIDs", serviceIds)
+	serviceLog.Infof("found %d VPC Endpoint Services to delete", len(serviceIds))
```
2uasimojo: as above

pkg/controller/awsprivatelink/cleanup.go
2uasimojo, line 1: nit: Given these two funcs are now acting on multiple resources, consider making their names plural.

`cleanupVPCEndpoints`
`cleanupVPCEndpointServices`

pkg/controller/awsprivatelink/cleanup_test.go
```diff
@@ -132,3 +138,231 @@ func TestCleanupRequired(t *testing.T) {
...
+					VpcEndpoints: nil,
+				}, nil)
+		},
+		expectedError: false,
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
```
2uasimojo: I think this is no longer necessary since go 1.14.

pkg/controller/awsprivatelink/cleanup_test.go
```diff
@@ -132,3 +138,231 @@ func TestCleanupRequired(t *testing.T) {
...
+					ServiceConfigurations: nil,
+				}, nil)
+		},
+		expectedError: false,
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
```
2uasimojo: "

2uasimojo (commented):
This looks clean.

/lgtm

A couple of nonblocking comments inline.

/hold

in case you want to fix them; but feel free to unhold otherwise.

pkg/controller/awsprivatelink/cleanup.go
huangmingxia: Thanks for reviewing. I’ve updated them.

pkg/controller/awsprivatelink/cleanup_test.go
huangmingxia: Many thanks. I’ve removed it.

pkg/controller/awsprivatelink/cleanup.go
huangmingxia: Thanks for reviewing these silly little issues. I’ll be more careful next time :)

2uasimojo: It's really not a big deal at all. (A "nit" is something really minor that probably nobody would ever notice.) Thanks for fixing.

PR #2729, opened 2025-08-05
HIVE-2302, HIVE-2644: Pass metadata.json through opaquely

2uasimojo:
Previously any time installer added a field to metadata.json, we would need to evaluate and possibly add a bespoke field and code path for it to make sure it was supplied to the destroyer at deprovision time.

With this change, we're offloading metadata.json verbatim (except in some cases we have to scrub/replace credentials fields -- see [HIVE-2804](https://issues.redhat.com//browse/HIVE-2804) / https://github.com/openshift/hive/pull/2612) to a new Secret in the ClusterDeployment's namespace, referenced from a new field: ClusterDeployment.Spec.ClusterMetadata.MetadataJSONSecretRef.

For legacy clusters -- those created before this change -- we attempt to retrofit the new Secret based on the legacy fields. This is best effort and may not always work.

This change then adds a new generic destroyer via the (existing) `hiveutil deprovision` command that consumes this metadata.json to deprovision the cluster.

This new behavior is the default, but we also include an escape hatch to run the platform-specific legacy destroyer by setting the following annotation on the ClusterDeployment:

`hive.openshift.io/legacy-deprovision: "true"`

### DIFF
--- a/contrib/pkg/deprovision/deprovision.go
+++ b/contrib/pkg/deprovision/deprovision.go
@@ -81,53 +81,7 @@ To run the generic destroyer, use the --metadata-json-secret-name parameter.`,
 				logger.Fatal("no platform configured in metadata.json")
 			}
 
-			// TODO: Make a registry or interface for this
-			var ConfigureCreds func(client.Client)
-			switch platform {
-			case aws.Name:
-				ConfigureCreds = awsutil.ConfigureCreds
-			case azure.Name:
-				ConfigureCreds = azureutil.ConfigureCreds
-			case gcp.Name:
-				ConfigureCreds = gcputil.ConfigureCreds
-			case ibmcloud.Name:
-				ConfigureCreds = ibmcloudutil.ConfigureCreds
-			case nutanix.Name:
-				// Snowflake! We need to inject the creds into the metadata.
-				// If env vars are unset, the destroyer will fail organically.
-				ConfigureCreds = func(c client.Client) {
-					nutanixutil.ConfigureCreds(c)
-					metadata.Nutanix.Username = os.Getenv(constants.NutanixUsernameEnvVar)
-					metadata.Nutanix.Password = os.Getenv(constants.NutanixPasswordEnvVar)
-				}
-			case openstack.Name:
-				ConfigureCreds = openstackutil.ConfigureCreds
-			case vsphere.Name:
-				// Snowflake! We need to (re)inject the creds into the metadata.
-				// (They were there originally, but we scrubbed them for security.)
-				// If env vars are unset, the destroyer will fail organically.
-				ConfigureCreds = func(c client.Client) {
-					vsphereutil.ConfigureCreds(c)
-					username, password := os.Getenv(constants.VSphereUsernameEnvVar), os.Getenv(constants.VSpherePasswordEnvVar)
-					// Accommodate both pre- and post-zonal formats
-					if metadata.VSphere.Username != "" {
-						metadata.VSphere.Username = username
-					}
-					if metadata.VSphere.Password != "" {
-						metadata.VSphere.Password = password
-					}
-					for i := range metadata.VSphere.VCenters {
-						if metadata.VSphere.VCenters[i].Username != "" {
-							metadata.VSphere.VCenters[i].Username = username
-						}
-						if metadata.VSphere.VCenters[i].Password != "" {
-							metadata.VSphere.VCenters[i].Password = password
-						}
-					}
-				}
-			}
-
-			ConfigureCreds(c)
+			ConfigureCreds[platform](c, metadata)
 
 			destroyerBuilder, ok := providers.Registry[platform]
 			if !ok {
@@ -161,4 +115,14 @@ To run the generic destroyer, use the --metadata-json-secret-name parameter.`,
 	cmd.AddCommand(NewDeprovisionvSphereCommand(logLevel))
 	cmd.AddCommand(NewDeprovisionNutanixCommand(logLevel))
 	return cmd
+}
+
+var ConfigureCreds = map[string]func(client.Client, *types.ClusterMetadata){
+	aws.Name:       awsutil.ConfigureCreds,
+	azure.Name:     azureutil.ConfigureCreds,
+	gcp.Name:       gcputil.ConfigureCreds,
+	ibmcloud.Name:  ibmcloudutil.ConfigureCreds,
+	nutanix.Name:   nutanixutil.ConfigureCreds,
+	openstack.Name: openstackutil.ConfigureCreds,
+	vsphere.Name:   vsphereutil.ConfigureCreds,
 }
--- a/contrib/pkg/deprovision/awstagdeprovision.go
+++ b/contrib/pkg/deprovision/awstagdeprovision.go
@@ -65,7 +65,7 @@ func completeAWSUninstaller(o *aws.ClusterUninstaller, logLevel string, args []s
 			"This is expected when in standalone mode. "+
 			"We expect to find your AWS credentials in one of the usual places.", err)
 	}
-	awsutils.ConfigureCreds(client)
+	awsutils.ConfigureCreds(client, nil)
 
 	return nil
 }
--- a/contrib/pkg/deprovision/azure.go
+++ b/contrib/pkg/deprovision/azure.go
@@ -74,7 +74,7 @@ func (opt *AzureOptions) completeAzureUninstaller(args []string) (providers.Dest
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get client")
 	}
-	azureutils.ConfigureCreds(client)
+	azureutils.ConfigureCreds(client, nil)
 
 	metadata := &types.ClusterMetadata{
 		InfraID: args[0],
--- a/contrib/pkg/deprovision/gcp.go
+++ b/contrib/pkg/deprovision/gcp.go
@@ -60,7 +60,7 @@ func (o *gcpOptions) Complete(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get client")
 	}
-	gcputils.ConfigureCreds(client)
+	gcputils.ConfigureCreds(client, nil)
 
 	return nil
 }
--- a/contrib/pkg/deprovision/ibmcloud.go
+++ b/contrib/pkg/deprovision/ibmcloud.go
@@ -69,7 +69,7 @@ func (o *ibmCloudDeprovisionOptions) Complete(cmd *cobra.Command, args []string)
 	if err != nil {
 		return errors.Wrap(err, "failed to get client")
 	}
-	ibmutils.ConfigureCreds(client)
+	ibmutils.ConfigureCreds(client, nil)
 
 	// Create IBMCloud Client
 	ibmCloudAPIKey := os.Getenv(constants.IBMCloudAPIKeyEnvVar)
--- a/contrib/pkg/deprovision/nutanix.go
+++ b/contrib/pkg/deprovision/nutanix.go
@@ -61,7 +61,7 @@ func (o *nutanixOptions) Complete(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get client")
 	}
-	nutanixutils.ConfigureCreds(client)
+	nutanixutils.ConfigureCreds(client, nil)
 
 	return nil
 }
--- a/contrib/pkg/deprovision/openstack.go
+++ b/contrib/pkg/deprovision/openstack.go
@@ -56,7 +56,7 @@ func (o *openStackOptions) Complete(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get client")
 	}
-	openstackutils.ConfigureCreds(client)
+	openstackutils.ConfigureCreds(client, nil)
 
 	return nil
 }
--- a/contrib/pkg/deprovision/vsphere.go
+++ b/contrib/pkg/deprovision/vsphere.go
@@ -60,7 +60,7 @@ func (o *vSphereOptions) Complete(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get client")
 	}
-	vsphereutils.ConfigureCreds(client)
+	vsphereutils.ConfigureCreds(client, nil)
 
 	return nil
 }
--- a/contrib/pkg/utils/aws/aws.go
+++ b/contrib/pkg/utils/aws/aws.go
@@ -7,7 +7,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
+	ini "gopkg.in/ini.v1"
+
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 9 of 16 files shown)


2uasimojo:
TODO: Deprovisioner side

pkg/controller/clusterclaim/clusterclaim_controller.go
```diff
@@ -556,13 +556,14 @@ func (r *ReconcileClusterClaim) applyHiveClaimOwnerRole(cd *hivev1.ClusterDeploy
...
 			},
-			// Allow read access to the kubeconfig and admin password secrets
+			// Allow read access to the kubeconfig, admin password, and metadata.json secrets
 			{
 				APIGroups: []string{corev1.GroupName},
 				Resources: []string{"secrets"},
 				ResourceNames: []string{
 					cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name, // HIVE-2485: n/a
 					cd.Spec.ClusterMetadata.AdminPasswordSecretRef.Name,
+					cd.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name,
```
2uasimojo: - Dur, copypasta
- We should check this for nils in case there's a timing problem with a legacy clusterpool cluster not yet having been retrofit with its metadataJSON secret ref.

2uasimojo:
/test e2e e2e-azure e2e-gcp e2e-vsphere e2e-openstack

:crossed_fingers:

2uasimojo:
- This is now rebased on #2746
- VSphere (hopefully) remedied: metadata.json can be in one of two different shapes depending on pre- or post-zonal. We should now be accounting for both.

/test e2e-vsphere

2uasimojo:
- `e2e` and `e2e-pool`: actual tests succeeded; infra timeout flake
- `e2e-openstack` looks like a flake where the openstack infra was ill
- `security` was HIVE-2937, since resolved via #2755...

...so I think I'll just rebase to pick up that last one anyway.

...but I might as well wait until #2754 lands.

2uasimojo:
Legacy granny switch validated via #2759 ✓

2uasimojo:
/test e2e-openstack

Pre-actual-test IPI setup flake

2uasimojo:
/test e2e-openstack

Same again

2uasimojo:
/test e2e-openstack

Looks like provision timed out?

contrib/pkg/deprovision/deprovision.go
```diff
@@ -1,26 +1,164 @@
...
+				logger.WithError(err).Fatal("failed to unmarshal metadata.json")
+			}
+
+			platform := metadata.Platform()
+			if platform == "" {
+				logger.Fatal("no platform configured in metadata.json")
+			}
+
+			// TODO: Make a registry or interface for this
+			var ConfigureCreds func(client.Client)
```
dlom: I think this is fine, it's readable and makes sense

contrib/pkg/deprovision/deprovision.go
```diff
@@ -1,26 +1,164 @@
...
+			case ibmcloud.Name:
+				ConfigureCreds = ibmcloudutil.ConfigureCreds
+			case nutanix.Name:
+				// Snowflake! We need to inject the creds into the metadata.
+				// If env vars are unset, the destroyer will fail organically.
+				ConfigureCreds = func(c client.Client) {
+					nutanixutil.ConfigureCreds(c)
+					metadata.Nutanix.Username = os.Getenv(constants.NutanixUsernameEnvVar)
+					metadata.Nutanix.Password = os.Getenv(constants.NutanixPasswordEnvVar)
+				}
```
dlom:
```suggestion
				// If env vars are unset, the destroyer will fail organically.
				ConfigureCreds = nutanixutil.ConfigureCreds
				metadata.Nutanix.Username = os.Getenv(constants.NutanixUsernameEnvVar)
				metadata.Nutanix.Password = os.Getenv(constants.NutanixPasswordEnvVar)
```

dlom: I think you can just write it like this.  The function is immediately invoked right below

contrib/pkg/deprovision/deprovision.go
```diff
@@ -1,26 +1,164 @@
...
+			case openstack.Name:
+				ConfigureCreds = openstackutil.ConfigureCreds
+			case vsphere.Name:
+				// Snowflake! We need to (re)inject the creds into the metadata.
+				// (They were there originally, but we scrubbed them for security.)
+				// If env vars are unset, the destroyer will fail organically.
+				ConfigureCreds = func(c client.Client) {
+					vsphereutil.ConfigureCreds(c)
```
dlom: Same here.  Nothing else inside the function is referencing this `c` variable, I think it can just be directly hoisted up into the switch case

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -971,6 +1018,201 @@ func (r *ReconcileClusterDeployment) reconcileInstallingClusterInstall(cd *hivev
...
+// clusterMetadata.MetadataJSONSecretRef to point to it. The caller is responsible for folding the
+// (potentially) modified clusterMetadata back into the cd and Update()ing it.
+// If forceUpdate is true, we will ensure the contents of the Secret's metadata.json key are equal
+// to metadataJSON. If forceUpdate is false, we will only update labels iff necessary.
+func (r *ReconcileClusterDeployment) ensureMetadataJSONSecret(
+	cd *hivev1.ClusterDeployment,
+	clusterMetadata *hivev1.ClusterMetadata,
+	metadataJSON []byte,
+	forceUpdate bool,
+	logger log.FieldLogger) error {
```
dlom:
```suggestion
	logger log.FieldLogger,
) error {
```

Cosmetic only suggestion

contrib/pkg/deprovision/deprovision.go
2uasimojo: That won't work without some refactoring, because ConfigureCreds is where those env vars get set, for both [nutanix](https://github.com/openshift/hive/blob/853824843add16752214c1db8876c5b99bcf3979/contrib/pkg/utils/nutanix/nutanix.go#L16-L21) and [vsphere](https://github.com/openshift/hive/blob/13ea4f477531cffcd351ef8b93786a3c94de1a84/contrib/pkg/utils/vsphere/vsphere.go#L16-L21).

Looking at it again, I think what makes sense is to pass the `metadata` into ConfigureCreds and set up the user/pass there. They are, after all, Creds that need to be Configured :)

My only objection to that is that I was trying to leave the legacy code path untouched as much as possible. Ultimately -- when we kill the legacy code path -- it'll be clean to do so; but in the meantime it would require me to either pass a dummy `metadata` is through or add a nil check condition in ConfigureCreds. The latter seems like the lesser evil. But I'm not sure it's worthwhile -- WDYT?

contrib/pkg/deprovision/deprovision.go
2uasimojo: Ack. But it would jell with how we're getting the correct platform-specific destroyer from the installer -- see LL132 & 137 below. I don't mind leaving it for now though.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: Ack. (I generally like golang's opinionated whitespace rules, but I feel like they really missed the mark on this one. An extra indent for 2nd+ lines of long conditionals or function defs would fix it.)

contrib/pkg/deprovision/deprovision.go
dlom: The most I would consider would be something like this:
```go
type ConfigureCredsFn func(client.Client)
...
var configureCreds ConfigureCredsFn
```

contrib/pkg/deprovision/deprovision.go
dlom: I think adding a metadata parameter and using a nil check to all of the ConfigureCreds implementations is a good compromise, especially if the line count is low (probably no more than +10 per file)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dlom: Yeah, the whole point of splitting all the parameters across different lines is to make future deltas very simple, and adding a unique trailer to the last item defeats the point

contrib/pkg/deprovision/deprovision.go
2uasimojo: What does this gain? I still have to define the funcs with the full param spec, right?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: I would say the main purpose is readability, especially when it's a long param list (or conditional).

WDYM "unique trailer"? The mandatory comma on the last line means that line doesn't need a delta if you add/remove params, but I'm not sure that's relevant here.

My gripe is that visually both accepted forms are ugly. They both align the 2nd+ param lines with the function (or conditional) body, which is Bad™. The one you suggest is marginally better because it puts a visual break between the two; but I find the outdenting of the last line jarring and visually confusing. Maybe it's just me.

I'm saying the Right™ way would be like:
```go
func (r *ReconcileClusterDeployment) ensureMetadataJSONSecret(
		cd *hivev1.ClusterDeployment,
		clusterMetadata *hivev1.ClusterMetadata,
		metadataJSON []byte,
		forceUpdate bool,
		logger log.FieldLogger) error {
	// Figure out the name of the Secret.
	// The default:
	secretName := fmt.Sprintf(metadataJSONSecretNameTemplate, cd.Name)
```
...or if you care about the comma thing:
```go
func (r *ReconcileClusterDeployment) ensureMetadataJSONSecret(
		cd *hivev1.ClusterDeployment,
		clusterMetadata *hivev1.ClusterMetadata,
		metadataJSON []byte,
		forceUpdate bool,
		logger log.FieldLogger,
		) error {
	// Figure out the name of the Secret.
	// The default:
	secretName := fmt.Sprintf(metadataJSONSecretNameTemplate, cd.Name)
```

But absent either of those being allowed, our choices are to put an extra newline as I've done, or put the last line outdented as you suggest. I think they're both equally icky, but if you prefer the latter, I'm okay with changing it.

contrib/pkg/deprovision/deprovision.go
2uasimojo: Done, via a separate commit in case we hate it :)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dlom: * Re: unique "trailer" (couldn't think of a better phrase): every other function parameter line ends with a comma, but the last line ends with `) error {`
* The choice here is binary I think: either split `) error {` out into a new line, or keep it on the end of the final line.  Either way, `gofmt` will do its thing and we'll end up (unconditionally) with one of these two results:

```go
func (r *ReconcileClusterDeployment) ensureMetadataJSONSecret(
	cd *hivev1.ClusterDeployment,
	clusterMetadata *hivev1.ClusterMetadata,
	metadataJSON []byte,
	forceUpdate bool,
	logger log.FieldLogger) error {
	// ...
}

func (r *ReconcileClusterDeployment) ensureMetadataJSONSecret2(
	cd *hivev1.ClusterDeployment,
	clusterMetadata *hivev1.ClusterMetadata,
	metadataJSON []byte,
	forceUpdate bool,
	logger log.FieldLogger,
) error {
	// ...
}
```

Personally, I think option 2 is easier to both read and edit.  `gofmt` is king, though I'm surprised it allows for two different "stable" formatted results

2uasimojo: Discussed offline. Concluded both options are icky, and also roughly evenly present across the repo. Leaving this alone simply for the sake of not restarting CI.

contrib/pkg/deprovision/deprovision.go
```diff
@@ -81,53 +81,7 @@ To run the generic destroyer, use the --metadata-json-secret-name parameter.`,
...
-						}
-						if metadata.VSphere.VCenters[i].Password != "" {
-							metadata.VSphere.VCenters[i].Password = password
-						}
-					}
-				}
-			}
-
-			ConfigureCreds(c)
+			ConfigureCreds[platform](c, metadata)
```
dlom, line 84: I love it

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -737,6 +780,11 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 					return reconcile.Result{}, err
 				}
 			}
+			if mdjsr := cd.Spec.ClusterMetadata.MetadataJSONSecretRef; mdjsr != nil && mdjsr.Name != "" {
+				if err := r.addOwnershipToSecret(cd, cdLog, mdjsr.Name); err != nil {
+					return reconcile.Result{}, err
```
huangmingxia, line 782: Hi @2uasimojo  I have a question regarding the cleanup of the metadata.json Secret. Please correct me if I missed anything, thank you!

Currently, the OwnerReference of the metadata.json Secret is only set when the cluster installation succeeds. If the installation fails and the Secret has already been created, it has no OwnerReference and will remain in the cluster when the ClusterDeployment is deleted - it won’t be removed automatically.

Regarding the leftover Secret, there is a potential issue:
If an invalid metadata.json Secret already exists before the upgrade, the legacy retrofit path calls ensureMetadataJSONSecret() with forceUpdate=false. In this case, if the Secret already exists, its contents will not be updated even if the metadata in the retrofit has changed. This can cause the cluster deletion to fail. (I think it works with the annotation.)

Would it be possible to set the OwnerReference when creating the Secret? This way, the Secret would be automatically cleaned up regardless of whether the installation succeeds.

As additional context:
I did some tests switching repeatedly between old and new hive versions on the same hub. 
When a ClusterDeployment installation failed on the new version, the metadata.json Secret was left behind without cleanup.
After reverting to the old version and creating a successful ClusterDeployment in the same namespace, upgrading hive again caused the existing secret to remain outdated (since forceUpdate=false), resulting in a mismatch between clusterdeprovision data and metadata.json, which led to cluster deletion failure.

2uasimojo: That's a great find @huangmingxia!

I'm really not worried about supporting "downgrades". (I think we could also simulate this behavior by creating a dummy Secret of the appropriate name before kicking off the installation with the new build. I'm also not worried about supporting that :) )

However, I agree we should be setting that OwnerReference immediately to avoid the scenario where the Secret is "orphaned" if the installation fails. PR incoming...

2uasimojo:
This rev still needs UT fixes. Just saving my work.

dlom:
/lgtm

@2uasimojo merge when you're ready

2uasimojo:
The `verify` fail smells like a testplatform issue. I'll try again on Monday.

2uasimojo:
> The `verify` fail smells like a testplatform issue. I'll try again on Monday.

Opened https://issues.redhat.com/browse/DPTP-4590

2uasimojo:
/hold cancel

QE looks good. There is one pending issue (start reading [here](https://issues.redhat.com/browse/HIVE-2302?focusedId=28373210&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28373210)) but it would need to be coordinated with the installer team, if we decide to address it at all.

Let's go!

PR #2731, opened 2025-08-11
HIVE-2391: vSphere zonal support

dlom:
xref: [HIVE-2391](https://issues.redhat.com//browse/HIVE-2391)

This PR is a re-do of #2541 

See also: https://github.com/openshift/installer/pull/10025


dlom:
xref: [HIVE-2396](https://issues.redhat.com//browse/HIVE-2396)

config/crds/hive.openshift.io_clusterdeployments.yaml
```diff
@@ -1144,26 +1146,455 @@ spec:
                         type: object
                         x-kubernetes-map-type: atomic
                       datacenter:
-                        description: Datacenter is the name of the datacenter to use
-                          in the vCenter.
+                        description: |-
+                          Datacenter is the name of the datacenter to use in the vCenter.
+                          Deprecated: Please use Platform.VSphere instead
```
jianping-shu: The Deprecated fields datacenter, defaultDatastore and vCenter are still defined as required. After hive upgrade, the users still need to configure them even configuring platform.vsphere.vSphere already, which means the transition not done completely.

contrib/pkg/createcluster/create.go
```diff
@@ -375,14 +369,8 @@ OpenShift Installer publishes all the services of the cluster like API server an
 	flags.StringVar(&opt.OpenStackIngressFloatingIP, "openstack-ingress-floating-ip", "", "Floating IP address to use for cluster's Ingress service")
 
 	// vSphere flags
-	flags.StringVar(&opt.VSphereVCenter, "vsphere-vcenter", "", "Domain name or IP address of the vCenter")
```
jianping-shu, line 346: The vsphere parameters are deleted but the vsphere sample for hiveutil create-cluster not updated accoringly, see the "case 3 - hiveutil" part of OCP-84265.

dlom: @jianping-shu I've updated the documentation.  Please let me know if you have any questions, or need a sample `platform-spec.json`

config/crds/hive.openshift.io_clusterdeployments.yaml
dlom: I've marked all of the deprecated fields as optional, please let me know if the issue persists

jianping-shu: I can see the update in hive.openshift.io_clusterdeployments.yaml now

contrib/pkg/createcluster/create.go
jianping-shu: The vsphere sample for hiveutil create-cluster is updated, fixed.

I've tried the following 2 cases on OCP-84265
1)--vsphere-platform-spec-json paramter is missing but the error could be improved
jianpingshu@jshu-mac bin % ./hiveutil create-cluster jshu1 --cloud=vsphere --vsphere-api-vip=192.168.1.2 --vsphere-ingress-vip=192.168.1.3 --vsphere-ca-certs=./vcenter-certificate -o yaml > jshu1.yaml
DEBU[0000] Couldn't find install logs provider environment variable. Skipping. 
FATA[0000] Error                                         command=create-cluster error="error decoding platform : unexpected end of JSON input"
2)I tried several json files but all got the following error, see this [code](https://github.com/openshift/hive/blob/a4611b9a1a8227de8aa7daab65085e559bd9a7dc/contrib/pkg/createcluster/create.go#L811) , looks like it needs "os.ReadFile" for the json file, can you help check?
jianpingshu@jshu-mac bin % ./hiveutil create-cluster jshu1 --cloud=vsphere --vsphere-api-vip=192.168.1.2 --vsphere-ingress-vip=192.168.1.3 --vsphere-ca-certs=./vcenter-certificate --vsphere-platform-spec-json=./platform2.json  -o yaml > jshu1.yaml
DEBU[0000] Couldn't find install logs provider environment variable. Skipping. 
FATA[0000] Error                                         command=create-cluster error="error decoding platform /Users/jianpingshu/program/test_case/HIVE/HIVE-2391/hiveutil/platform2.json: invalid character '/' looking for beginning of value"

jianping-shu:
@dlom I tried the new commit, install vsphere cd with old format i.e. without platform.vsphere.vSphere still failed due to the same error. Pls. check the comments and hive controlller log attached on https://issues.redhat.com/browse/HIVE-2917.

newtonheath:
@dlom can you pick up this again now private image work is in review - need to get this wrapped up

contrib/pkg/createcluster/create.go
dlom: @jianping-shu it's my mistake: The environment variable should contain the contents of the json, not a path to the json file itself.  I've updated the docs to match the correct intended usage

pkg/controller/clusterpool/clusterpool_controller.go
```diff
@@ -1311,6 +1311,9 @@ func (r *ReconcileClusterPool) createCloudBuilder(pool *hivev1.ClusterPool, logg
 		cloudBuilder.Cloud = platform.OpenStack.Cloud
 		return cloudBuilder, nil
 	case platform.VSphere != nil:
+		if platform.VSphere.VSphere == nil {
+			return nil, errors.New("VSphere CD with deprecated fields has not been updated by CD controller yet, requeueing...")
```
jianping-shu: 1. It should be CP/CP controller instead of CD/CD controller
2. I didn't see the code logic of converting platform.vsphere.attributes to platform.vsphere.vSphere in clusterpool_controller.go like in clusterpool_deployment.go

jianping-shu: Since ClusterPool support for vSphere is NOT supported officially yet(HIVE-2801 not closed and hive doc not updated). One choice is if we can make the judgement/statement that the ClusterPool only support the new format of platform.vsphere.vSphere, this part of code change can be avoided. In that case we need at least make the error in condition/log appropriate and clearer.

apis/hive/v1/vsphere/platform.go
```diff
@@ -18,18 +26,95 @@ type Platform struct {
...
 	Cluster string `json:"cluster,omitempty"`
 
 	// Network specifies the name of the network to be used by the cluster.
+	// Deprecated: Please use Platform.VSphere instead
+	// See also: Platform.ConvertDeprecatedFields
+	// +optional
 	Network string `json:"network,omitempty"`
+
+	// VSphere is the full spec of the vSphere platform.
+	VSphere *configv1.VSpherePlatformSpec `json:"vSphere,omitempty"`
```
2uasimojo: This is odd, as it makes for .spec.platform.vsphere.vsphere (and the deprecation comments above end up not being accurate).

I get why you did it, and I don't have a better suggestion at 5pm on a Friday night. Will sleep on it...

dlom: Yes, I'm not sure what you mean.  Let's discuss on Monday

pkg/controller/clusterpool/clusterpool_controller.go
dlom: I think ClusterPool support will have to be bundled into this PR, to prevent a mismatch in functionality between our CD/CP controllers.  I'm going to try to add a similar approach for the CP that I used on the CD

apis/hive/v1/vsphere/platform.go
2uasimojo: Realistically, we will "never" (in v1) remove the existing fields we're deprecating under spec.platform.vsphere. However, we should reach a point where those fields are ignored and consumers are using the new hotness exclusively. When that's the case, we really don't want elements under a path spelled spec.platform.vsphere.vSphere (especially with the difference in case, which I noticed later -- that's gonna drive some user batty).

Having slept on it, I've come up with two alternatives. Neither is beautiful, but IMHO they're _less_ not-beautiful.

1. Rename this element. Suggestions welcome. My initial thought is "infrastructure".
```suggestion
	Infrastructure *configv1.VSpherePlatformSpec `json:"infrastructure,omitempty"`
```

2. Inline the schema rather than naming it at all.
```suggestion
	configv1.VSpherePlatformSpec
```

I *think* this results in those fields ending up at the same level as the existing ones -- i.e. spec.platform.vsphere.vcenters, spec.platform.vsphere.failureDomains, etc. It doesn't look like any of the field names conflict with what we already have at that level, so it should work. Another selling point for this option is that it most closely matches how installer did it in the [install-config schema](https://github.com/openshift/hive/blob/17639510ebeacb4936b03c7359eb878f6931c984/vendor/github.com/openshift/installer/pkg/types/vsphere/platform.go#L53).

(If inlining doesn't work mechanically, we could include the fields explicitly, like
```suggestion
	// ...
	VCenters []configv1.VSpherePlatformVCenterSpec `json:"vcenters,omitempty"`

	// ...
	FailureDomains []configv1.VSpherePlatformFailureDomainSpec `json:"failureDomains,omitempty"`

	// etc.
```
...but I don't think this will be necessary.)

----

With either of these approaches, we could copy something else the installer did, which is to rename the **go** struct fields with a `Deprecated` prefix. As long as the json names don't change, this is not breaking and is allowable, and more strongly encourages programmatic consumers to transition.

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -635,6 +632,18 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateUpdate(admissionSpec
 	// Add the new data to the contextLogger
 	contextLogger.Data["oldObject.Name"] = oldObject.Name
 
+	// HIVE-2391
+	if oldObject.Spec.Platform.VSphere != nil && cd.Spec.Platform.VSphere != nil {
```
2uasimojo, line 673: This allows/enforces a model where you can transition from old => old+new. I think it would be better if we instead enforced old => new -- i.e. you're forced to scrub the deprecated fields when upconverting. I realize this is going to entail more code because we have to look at each field individually and make sure it's `""` (empty string). But IMHO the end result is a lot cleaner.

BTW, I did notice your comment about performance. We're really not concerned about adding more local comparison logic -- the resource consumption, wallclock time, etc is negligible against the overall operation: most of the burden comes from shoving objects over the network.

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -635,6 +632,18 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateUpdate(admissionSpec
...
+		// Moving from a non-zonal to a zonal shape is permitted.
+		// This check is faster than checking all the fields individually
+		if oldObject.Spec.Platform.VSphere.VSphere == nil && cd.Spec.Platform.VSphere.VSphere != nil {
+			contextLogger.Debug("Passed validation: HIVE-2391")
+			return &admissionv1beta1.AdmissionResponse{
+				Allowed: true,
```
2uasimojo, line 679: As written I think this short circuit would allow you to stealth-edit the other otherwise-immutable fields as part of upconverting. That is undesirable :)

Please add UT to confirm that this :arrow_up: can't happen.

pkg/controller/machinepool/awsactuator.go
```diff
@@ -392,14 +392,14 @@ func (a *AWSActuator) updateProviderConfig(machineSet *machineapi.MachineSet, in
 // spec matches the number of AZs, we will not check whether they are public or private. However, if there
 // are two subnets per AZ, we will filter out the public ones, leaving only the private ones. If the pool
 // specifies no subnets, an empty map is returned (this is a valid configuration, not an error).
-func (a *AWSActuator) getSubnetsByAvailabilityZone(pool *hivev1.MachinePool) (icaws.Subnets, error) {
```
dlom: Side effect of installer revendor

pkg/controller/machinepool/nutanixactuator_test.go
```diff
@@ -144,8 +144,14 @@ func TestGenerateMachineSets(t *testing.T) {
 			clusterDeployment: func() *hivev1.ClusterDeployment {
 				cd := testNutanixClusterDeployment()
 				cd.Spec.Platform.Nutanix.FailureDomains = []hivev1nutanix.FailureDomain{
-					{Name: "valid-domain-1"},
```
dlom, line 183: Side effect of installer revendor

go.mod
```diff
@@ -513,3 +527,7 @@ replace golang.org/x/net => golang.org/x/net v0.38.0
 
 // From installer: This is to force capi back for the older provider version
 replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.19.3
+
+replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v1.9.6
+
+replace github.com/openshift/installer => github.com/dlom/installer v0.0.0-20251022025850-76157f125d29
```
dlom: Currently pointing at my fork (https://github.com/openshift/installer/pull/10025) (actually it's this https://github.com/dlom/installer/tree/add-deepcopy-to-types-1.4.19-ec5)

jianping-shu:
jobs still failed. @dlom Pls. ping me when it is ready for testing, thanks

dlom:
/override ci/prow/security

to be fixed in another PR

dlom:
@2uasimojo this is ready for review (again)

/assign @2uasimojo

dlom:
@jianping-shu please re-verify this PR, it is ready again

jianping-shu:
Super! I'm on it.

jianping-shu:
@dlom I've done big part of testing (recorded on [HIVE-2917](https://issues.redhat.com//browse/HIVE-2917)), one minor finding so far:
In previous commit, the deprecated parameters are removed from hiveutil. But this is not the case in this commit.
Pls. check if it is expected.

jshu@jshu-mac bin % ./hiveutil create-cluster --help|grep vsphere
DEBU[0000] Couldn't find install logs provider environment variable. Skipping. 
platform spec for vsphere, encoded as JSON
create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=vsphere --vsphere-vcenter=vmware.devcluster.com --vsphere-datacenter=dc1 --vsphere-default-datastore=nvme-ds1 --vsphere-api-vip=192.168.1.2 --vsphere-ingress-vip=192.168.1.3 --vsphere-cluster=devel --vsphere-network="VM Network" --vsphere-ca-certs=/path/to/cert
create-cluster CLUSTER_DEPLOYMENT_NAME --cloud=vsphere --vsphere-api-vip=192.168.1.2 --vsphere-ingress-vip=192.168.1.3 --vsphere-ca-certs=/path/to/cert --vsphere-platform-spec-json="{<installer platform spec>}" [flags]
      --cloud string                                    Cloud provider: aws|azure|gcp|ibmcloud|nutanix|openstack|vsphere (default "aws")
      --vsphere-api-vip string                          Virtual IP address for the api endpoint
      --vsphere-ca-certs string                         Path to vSphere CA certificate, multiple CA paths can be : delimited
      --vsphere-cluster string                          Cluster virtual machines will be cloned into
      --vsphere-datacenter string                       Datacenter to use in the vCenter
      --vsphere-default-datastore string                Default datastore to use for provisioning volumes
      --vsphere-folder string                           Folder that will be used and/or created for virtual machines
      --vsphere-ingress-vip string                      Virtual IP address for ingress application routing
      --vsphere-network string                          Name of the network to be used by the cluster
      --vsphere-platform-spec-json string               Installer vsphere platform spec, encoded as JSON
      --vsphere-vcenter string                          Domain name or IP address of the vCenter

dlom:
@jianping-shu yes, this is expected.  I spoke with @2uasimojo and we decided to keep the older options in for backwards compatibility.  The new method (using `VSPHERE_INSTALLER_PLATFORM_SPEC_JSON`) is preferred, and the old flags have been re-routed to work with the new flow.

jianping-shu:
Added test cases on [HIVE-2391](https://issues.redhat.com//browse/HIVE-2391) for review.

apis/hive/v1/vsphere/machinepools.go
```diff
@@ -1,33 +1,27 @@
...
-type OSDisk struct {
-	// DiskSizeGB defines the size of disk in GB.
-	DiskSizeGB int32 `json:"diskSizeGB"`
+	// Topology is the vSphere topology that will be used for virtual machines.
+	// If it is not present, a default value will be used.
+	// +optional
+	Topology *vsphere.Topology `json:"topology,omitempty"`
 }
+
+type OSDisk = vsphere.OSDisk
```
2uasimojo: Is this necessary? The following compiles and passes UT -- does it break something else?

```diff
efried@efried-thinkpadp16vgen1:~/go/src/github.com/openshift/hive$ git diff
diff --git a/apis/hive/v1/vsphere/machinepools.go b/apis/hive/v1/vsphere/machinepools.go
index 37e27be7c..7f0aa7055 100644
--- a/apis/hive/v1/vsphere/machinepools.go
+++ b/apis/hive/v1/vsphere/machinepools.go
@@ -23,5 +23,3 @@ type MachinePool struct {
        // +optional
        Topology *vsphere.Topology `json:"topology,omitempty"`
 }
-
-type OSDisk = vsphere.OSDisk
diff --git a/pkg/controller/machinepool/vsphereactuator_test.go b/pkg/controller/machinepool/vsphereactuator_test.go
index 3c7d41bb8..b9518b76e 100644
--- a/pkg/controller/machinepool/vsphereactuator_test.go
+++ b/pkg/controller/machinepool/vsphereactuator_test.go
@@ -98,7 +98,7 @@ func testVSpherePool() *hivev1.MachinePool {
                                MemoryMiB:         32 * 1024,
                                NumCPUs:           4,
                                NumCoresPerSocket: 4,
-                               OSDisk: hivev1vsphere.OSDisk{
+                               OSDisk: vsphere.OSDisk{
                                        DiskSizeGB: 512,
                                },
                        },
diff --git a/pkg/validating-webhooks/hive/v1/machinepool_validating_admission_hook_test.go b/pkg/validating-webhooks/hive/v1/machinepool_validating_admission_hook_test.go
index fa0eeb11b..fe47c1550 100644
--- a/pkg/validating-webhooks/hive/v1/machinepool_validating_admission_hook_test.go
+++ b/pkg/validating-webhooks/hive/v1/machinepool_validating_admission_hook_test.go
@@ -704,7 +704,7 @@ func validAzureMachinePoolPlatform() *hivev1azure.MachinePool {
 func validvSphereMachinePoolPlatform() *hivev1vsphere.MachinePool {
        return &hivev1vsphere.MachinePool{
                MachinePool: installervsphere.MachinePool{
-                       OSDisk: hivev1vsphere.OSDisk{
+                       OSDisk: installervsphere.OSDisk{
                                DiskSizeGB: 1,
                        },
                },
diff --git a/vendor/github.com/openshift/hive/apis/hive/v1/vsphere/machinepools.go b/vendor/github.com/openshift/hive/apis/hive/v1/vsphere/machinepools.go
index 37e27be7c..7f0aa7055 100644
--- a/vendor/github.com/openshift/hive/apis/hive/v1/vsphere/machinepools.go
+++ b/vendor/github.com/openshift/hive/apis/hive/v1/vsphere/machinepools.go
@@ -23,5 +23,3 @@ type MachinePool struct {
        // +optional
        Topology *vsphere.Topology `json:"topology,omitempty"`
 }
-
-type OSDisk = vsphere.OSDisk
```

apis/hive/v1/vsphere/machinepools.go
```diff
@@ -1,33 +1,27 @@
...
+	DeprecatedTagIDs []string `json:"tagIDs,omitempty"`
 
-// OSDisk defines the disk for a virtual machine.
-type OSDisk struct {
-	// DiskSizeGB defines the size of disk in GB.
-	DiskSizeGB int32 `json:"diskSizeGB"`
+	// Topology is the vSphere topology that will be used for virtual machines.
+	// If it is not present, a default value will be used.
+	// +optional
+	Topology *vsphere.Topology `json:"topology,omitempty"`
```
2uasimojo, line 24: Previously we were getting the topology from the CD. I assumed this was:
- because it's the **cluster's** topology, the foundation on which you can place Machines, which you stuff into a subset of the available zones via `Zones`.
- so you don't have to repeat it for every MachinePool.

IIUC by including this field you're providing a way for a specific MachinePool to override the topology settings for a given FD when producing the mset for that FD. Are there legit, supported use cases for that? Is it only to provide backward compatibility for ResourcePool and TagIDs? If that's the case, we would probably need to make this a map, keyed by zone...

apis/hive/v1/vsphere/platform.go
```diff
@@ -18,18 +29,92 @@ type Platform struct {
...
+					ComputeCluster: setComputeClusterPath(p.DeprecatedCluster, p.DeprecatedDatacenter, logger),
+					Networks:       []string{p.DeprecatedNetwork},
+				},
+			},
+		},
+	}
+
+}
+
+// Copied (and slightly modified) from https://github.com/openshift/installer/blob/f7731922a0f17a8339a3e837f72898ac77643611/pkg/types/vsphere/conversion/installconfig.go#L75-L97
```
2uasimojo: I thought we were going to ask installer to export these so we don't have to dup them. Is there a reason we wouldn't be able to use them without modification?

(If we're going to do this, we don't have to hold up the PR for it, but let's at least get a jira paper trail going.)

contrib/pkg/deprovision/vsphere.go
```diff
@@ -91,13 +92,20 @@ func (o *vSphereOptions) Run() error {
 		return err
 	}
 
+	var vCenters []typesvsphere.VCenters
+	for _, vCenter := range strings.Split(o.vCenter, "::") {
+		vCenters = append(vCenters, typesvsphere.VCenters{
+			VCenter:  vCenter,
+			Username: o.username,
+			Password: o.password,
```
2uasimojo, line 100: This implies that we can only support multi vcenter if all vcenters are configured with the same creds, which is surely not a requirement for OCP in general.

Did we consider alternative designs here? Can we talk through it?

contrib/pkg/deprovision/vsphere.go
2uasimojo, line 52: We could use the more native `StringSliceVar`, which
- accepts multiple values delimited by commas, and/or
- accepts multiple instances of the option
- parses into the holder var automatically (though in this case we would still need to cycle through and populate the creds)

Since hiveutil isn't user-facing, we could even change the spelling of the option (`--vsphere-vcenters`) without worrying about deprecating etc.

apis/hive/v1/clusterdeprovision_types.go
```diff
@@ -132,8 +132,11 @@ type VSphereClusterDeprovision struct {
 	// CertificatesSecretRef refers to a secret that contains the vSphere CA certificates
 	// necessary for communicating with the VCenter.
 	CertificatesSecretRef corev1.LocalObjectReference `json:"certificatesSecretRef"`
-	// VCenter is the vSphere vCenter hostname.
-	VCenter string `json:"vCenter"`
+	// DeprecatedVCenter is the vSphere vCenter hostname.
+	// Deprecated: use VCenters instead
+	DeprecatedVCenter string `json:"vCenter"`
```
2uasimojo, line 137: I think it happens to be functionally irrelevant since this isn't a pointer, but adding `+optional` would at least make this field disappear from the `required` list in the schema, as well as being visually pro forma for a field we're deprecating.

Since this CRD is only generated by hive, I would expect any external consumers would be read-only. I wonder if we should describe how new code will behave to give consumers a hint as to how to change their code when they see this. Namely, it looks like we're going to only use the new thing for new instances, but not try to convert legacy instances in the field. Thus consumers would need to check both, like we do.

apis/hive/v1/vsphere/machinepools.go
jianping-shu: I was thinking the topology can be used to filter out all zones matched.
Pls. see scenario "Create infra6 MP" of Case 7 of OCP-84265.
When configures "hostGroup: sgao-host1-zone" in MP, I think it shall match up with FD sgao-host1-zone in theory, but the test result shows still 2 machinesets generated, 1 in sgao-host1-zone and 1 in sgao-host2-zone.
Seems it was used neither for matching nor for overriding.

apis/hive/v1/clusterdeprovision_types.go
jianping-shu: Does it means a test of hive upgrade is needed? I didn't do it yet.

contrib/pkg/deprovision/vsphere.go
jianping-shu: Pls forgive me if I was wrong.
Whether cases or CIs I found for zones, it is configured like with multi vcenters but 1 server. And the user/name is for server? So does OCP support multi servers for vsphere?

apis/hive/v1/vsphere/machinepools.go
2uasimojo: > the test result shows still 2 machinesets generated, 1 in sgao-host1-zone and 1 in sgao-host1-zone.

Assume this is a typo; one was `sgao-host2-zone`?

Looking at the [code](https://github.com/openshift/hive/blob/7ac33482e08dc9a0953fe58f89d4683e5bdaf798/vendor/github.com/openshift/installer/pkg/asset/machines/vsphere/machines.go#L394-L396), I think you would need to include FD.[ZoneType](https://github.com/openshift/hive/blob/38ef2afb15f1267c425e024b0816a36f0d882ab4/vendor/github.com/openshift/installer/pkg/types/vsphere/platform.go#L225): ["HostGroup"](https://github.com/openshift/hive/blob/38ef2afb15f1267c425e024b0816a36f0d882ab4/vendor/github.com/openshift/installer/pkg/types/vsphere/platform.go#L45) for the VMGroup field to get a non-default value... but AFAICS HostGroup is only used day 0; it's ignored in the MachineSet generation paths.

> I was thinking the topology can be used to filter out all zones matched.

Zones are always/only [matched](https://github.com/openshift/hive/blob/17639510ebeacb4936b03c7359eb878f6931c984/vendor/github.com/openshift/installer/pkg/asset/machines/vsphere/machinesets.go#L133-L157) based on `fd.name == mp.spec.platform.vsphere.zones[i]`.

----

We need to dig deeper here and figure out if there are any fields in `topology` that can legitimately override their counterparts in the fd. And even if they *can*, *should* they?

I think we've already seen that at least a subset of the fields can't be overridden; and the fact that they're required by the schema makes this awkward and annoying.

I'm thinking for the sake of simplicity, we should
- Remove the `topology` section from the MachinePool for this epic.
- Leave ResourcePool and TagIDs in place as overrides for _all_ machines in the pool (assuming overriding them  is effective).

If we get customers asking for topology things to be per-mset tunable later on, we can add `topology` back in... or do something else, if something else makes more sense (which it likely would, given the awkwardness of the schema requirements). Whereas if we commit to this shape now, it's really hard to walk it back if we didn't get it right.

My guess: YAGNI :)

contrib/pkg/deprovision/vsphere.go
2uasimojo: This is something we'll need to confirm with SPLAT/PCO. The metadata.json schema allows separate creds per vcenter, as shown by this chunk of code. To me that implies that it's _possible_ for different vcenters to have different creds.

apis/hive/v1/clusterdeprovision_types.go
2uasimojo: Oh, yeah, we should definitely have a test case where we
- start with an old hive build (prior to this change)
- start a deprovision, but don't let it finish (this could be tricky, but we could probably do it by borking the creds prior to deleting the CD)
- upgrade hive past this change
- (restore proper creds and) validate that the deprov completes successfully

Other more "normal" upgrade-y scenarios would also be useful. Install CD with old hive, upgrade, and validate that MachinePool (auto)scaling and deprov flows all work as expected. This would have some functional overlap with your existing scenarios that use the legacy install-config and MP shapes, but is probably worth a separate test suite.

apis/hive/v1/vsphere/machinepools.go
dlom: It's probably not necessary, I left it in just in case a downstream consumer was using that type

2uasimojo: Okay, yeah, we don't need to do that. If they're using the type, they can repoint their ref (to the installer's) when they revendor. It doesn't show up in our schema, so we don't need to worry about compat there.

apis/hive/v1/vsphere/machinepools.go
jianping-shu: > Assume this is a typo; one was sgao-host2-zone?
Yes, typo. Just corrected it.

apis/hive/v1/clusterdeprovision_types.go
jianping-shu: I'll add these 2 test cases.

contrib/pkg/deprovision/vsphere.go
jcpowermac: > Whether cases or CIs I found for zones, it is configured like with multi vcenters but 1 server. And the user/name is for server? So does OCP support multi servers for vsphere?

this is incorrect, there are multiple vCenters in CI. We can't assume they are the same username and password. IPI supports multiple vcenter topology.

apis/hive/v1/vsphere/platform.go
dlom: Yes, we are going to do it https://issues.redhat.com/browse/CORS-4333

2uasimojo:
/test e2e-vsphere

infra flake during IPI workflow

jianping-shu:
@dlom I didn't see the code changes regarding topology in MachinePool and creds for multi-vcenters. Pls. let me know when PR is ready for testing, thanks

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
2uasimojo: Fixed this up so you can do it either way (remove deprecated content or leave it and it'll be ignored).

Added UT for these paths as well.

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
2uasimojo: While this is true, it's not a problem, as the deprecated fields will be ignored after converting anyway.

UT is added to cover these upconversion special cases.

go.mod
2uasimojo: Revendor via #2796 brought in the relevant changes.

apis/hive/v1/vsphere/machinepools.go
2uasimojo: TL;DR:
- Remove Topology from MachinePool
- Revert deprecation of ResourcePool & TagIDs (these still override what's in the CD)

This is done.

contrib/pkg/deprovision/vsphere.go
2uasimojo: Did this.

Note: I changed `deprovision` but not `create-cluster`. I think I'm going to leave the vsphere `create-cluster` alone for now. I doubt it's ever used; it'll need a full rewrite; and we have [this](https://github.com/openshift-hive/hive-extras/tree/master/vsphere) which does the hard part (generating the right install-config shape).

(Later) Oh, I just noticed `VSPHERE_INSTALLER_PLATFORM_SPEC_JSON`. Yeah.

contrib/pkg/deprovision/vsphere.go
2uasimojo: Phew, this took some doing, but I've redesigned the creds format to match the relevant guts of the metadata.json schema. To wit, the key `vcenters` containing a yaml(/json)-encoded array of vcenter names and creds, like:

```yaml
apiVersion: v1
stringData:
  vcenters: |
    - vCenter: myvcenter1.devcluster.openshift.com
      username: vsphereuser1
      password: secretpassword1
    - vCenter: myvcenter2.devcluster.openshift.com
      username: vsphereuser2
      password: secretpassword2
kind: Secret
metadata:
  name: efried-vsphere-creds
  namespace: efried
type: Opaque
```

It's worth  mentioning that the use of these creds is kind of awkwardly wedged into the architecture that's common for all the other platforms.
- Creds for day 0 are embedded in the install-config. We don't do this for any other platform (except nutanix, new kid on the block, which looks like it modeled itself after vsphere :roll_eyes:).
- For deprovision, installer puts the creds into metadata.json (ditto nux) from which we scrub them to avoid leakage into the ClusterDeprovision/ClusterDeployment, and re-inject them before destroy. For other platforms (except nux, ditto) they're fed in separately. Presence of the metadata.json in ClusterDeprovision/ClusterDeployment is legacy and could be removed if we were sure we had upconverted all such clusters in the wild. (IDK if we'll ever have that confidence.) But we could conceivably do the scrubbing only for those instances, and leave the creds in the metadata.json Secret.
- Since we don't have hibernation for VSphere, and aren't super likely to (no substantial savings to be had, because on-prem?), there's no other place the creds are used.

TL;DR: If we were doing just VSphere, and doing it from scratch, we could probably skip the creds secret entirely. Let them percolate organically from the install-config to the metadata.json, both of which are already stored in same-namespace Secrets and thus equally secure. If they change, you gotta edit a Secret either way.

dlom:
Superseded by #2851

PR #2732, opened 2025-08-17
HIVE-2544: Support machinepool creation in Azure regions without availability zones

huangmingxia:
When users created a MachinePool in an Azure region that does not support availability zones,since the region does not provide availability zones, the MachinePool creation failed.

**Solution:**

- Based on Eric’s comments in https://github.com/openshift/hive/pull/2732#pullrequestreview-3129384057 
If no zones are specified, and the region has no zones, just pass through "no zones", and let the mset generator do what it does for a "no AZ region".

### DIFF
--- a/pkg/controller/machinepool/azureactuator.go
+++ b/pkg/controller/machinepool/azureactuator.go
@@ -156,10 +156,13 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
 			return nil, false, errors.Wrap(err, "compute pool not providing list of zones and failed to fetch list of zones")
 		}
 		if len(zones) == 0 {
-			return nil, false, fmt.Errorf("zero zones returned for region %s", cd.Spec.Platform.Azure.Region)
+			// No zones specified. Upstream will handle defaulting logic. Adding log here for visibility.
+			logger.WithField("region", cd.Spec.Platform.Azure.Region).Info("No availability zones detected for region. Using non-zoned deployment.")
+		} else {
+			computePool.Platform.Azure.Zones = zones
 		}
-		computePool.Platform.Azure.Zones = zones
 	}
+
 	// The imageID parameter is not used. The image is determined by the infraID.
 	const imageID = ""
 
--- a/pkg/controller/machinepool/azureactuator_test.go
+++ b/pkg/controller/machinepool/azureactuator_test.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/golang/mock/gomock"
-	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/pointer"
@@ -19,6 +18,7 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivev1azure "github.com/openshift/hive/apis/hive/v1/azure"
 	mockazure "github.com/openshift/hive/pkg/azureclient/mock"
+	testlogger "github.com/openshift/hive/pkg/test/logger"
 )
 
 type providerSpecValidator func(t *testing.T, providerSpec *machineapi.AzureMachineProviderSpec)
@@ -33,6 +33,7 @@ func TestAzureActuator(t *testing.T) {
 		expectedImage               *machineapi.Image
 		extraProviderSpecValidation providerSpecValidator
 		expectedErr                 bool
+		expectedLogs                []string
 	}{
 		// < 4.12
 		{
@@ -178,7 +179,10 @@ func TestAzureActuator(t *testing.T) {
 				mockGetVMCapabilities(client, "V1,V2")
 				mockListImagesByResourceGroup(client, []compute.Image{testAzureImage(compute.HyperVGenerationTypesV1)})
 			},
-			expectedErr: true,
+			expectedMachineSetReplicas: map[string]int64{
+				generateAzureMachineSetName(""): 3, // Non-zoned deployment
+			},
+			expectedLogs: []string{"No availability zones detected for region. Using non-zoned deployment."},
 		},
 		{
 			name:              "default V1 image exists and instance supports V1 images",
@@ -378,7 +382,10 @@ func TestAzureActuator(t *testing.T) {
 				mockGetVMCapabilities(client, "V1,V2")
 				mockListImagesByResourceGroup(client, []compute.Image{testAzureImage(compute.HyperVGenerationTypesV1)})
 			},
-			expectedErr: true,
+			expectedMachineSetReplicas: map[string]int64{
+				generateAzureMachineSetName(""): 3, // Non-zoned deployment
+			},
+			expectedLogs: []string{"No availability zones detected for region. Using non-zoned deployment."},
 		},
 		{
 			name:              "default V1 image exists and instance supports V1 images (4.12+)",
@@ -495,12 +502,13 @@ func TestAzureActuator(t *testing.T) {
 
 			aClient := mockazure.NewMockClient(mockCtrl)
 
-			// set up mock expectations
 			test.mockAzureClient(mockCtrl, aClient)
+
+			logger, hook := testlogger.NewLoggerWithHook()
 
 			actuator := &AzureActuator{
 				client: aClient,
-				logger: log.WithField("actuator", "azureactuator"),
+				logger: logger.WithField("actuator", "azureactuator"),
 			}
 
 			generatedMachineSets, _, err := actuator.GenerateMachineSets(test.clusterDeployment, test.pool, actuator.logger)
@@ -510,6 +518,10 @@ func TestAzureActuator(t *testing.T) {
 			} else {
 				assert.NoError(t, err, "unexpected error for test case")
 				validateAzureMachineSets(t, generatedMachineSets, test.expectedMachineSetReplicas, test.expectedImage, test.extraProviderSpecValidation)
+			}
+
+			for _, expectedLog := range test.expectedLogs {
+				testlogger.AssertHookContainsMessage(t, hook, expectedLog)
 			}
 		})
 	}


pkg/controller/machinepool/azureactuator.go
```diff
@@ -150,16 +150,27 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
...
+		} else {
+			computePool.Platform.Azure.Zones = zones
 		}
-		computePool.Platform.Azure.Zones = zones
+	} else if len(zones) == 0 {
+		computePool.Platform.Azure.Zones = []string{""}
+		logger.WithFields(log.Fields{
+			"region": cd.Spec.Platform.Azure.Region,
+			"zones":  computePool.Platform.Azure.Zones,
+		}).Warn("Region does not support availability zones. Deploying as non-zoned instead; user-specified zones are invalid and will be overridden.")
```
2uasimojo: We're generally not in the business of "fixing" a bad input / misconfiguration from the user. It seems like it would be a nice thing to do, but it's risky. Better to do exactly what is asked of us, even if that's an error, and force the user to fix their own config. In this case what that means is: if the user provides an explicit list of zones and the region doesn't support them, let's let the upstream (installer) code error out. The error will [bubble up](https://github.com/openshift/hive/blob/4012f2ff48203e73efcbd9e071562fb098d50041/pkg/controller/machinepool/machinepool_controller.go#L372-L380) into the `MachineSetsGenerated` status condition.

Note: Sometimes it can be a bit fuzzy. Interpreting "no zones specified" as "use all available zones" may seem like "fixing a bad input". We're considering it "a reasonable default" instead :)

pkg/controller/machinepool/azureactuator.go
```diff
@@ -150,16 +150,27 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
...
 		if len(zones) == 0 {
-			return nil, false, fmt.Errorf("zero zones returned for region %s", cd.Spec.Platform.Azure.Region)
+			computePool.Platform.Azure.Zones = []string{""}
+			logger.WithField("region", cd.Spec.Platform.Azure.Region).Info("Region does not support availability zones. Using non-zoned deployment.")
+		} else {
+			computePool.Platform.Azure.Zones = zones
 		}
-		computePool.Platform.Azure.Zones = zones
+	} else if len(zones) == 0 {
+		computePool.Platform.Azure.Zones = []string{""}
```
2uasimojo: We should be able to consolidate these two conditions.

Also, the upstream code already [defaults](https://github.com/openshift/hive/blob/17639510ebeacb4936b03c7359eb878f6931c984/vendor/github.com/openshift/installer/pkg/asset/machines/azure/machinesets.go#L27-L31) this value, so there's no need for us to do so.

pkg/controller/machinepool/azureactuator.go
```diff
@@ -150,16 +150,27 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
 		}
 	}
 
+	zones, err := a.getZones(cd.Spec.Platform.Azure.Region, pool.Spec.Platform.Azure.InstanceType)
```
2uasimojo: This looks like a pretty expensive call, so I would like to avoid doing it unless it's necessary. Having it done only if no Zones were specified isn't a *whole* lot better, but it is better :)

pkg/controller/machinepool/azureactuator.go
huangmingxia: Thank you for giving me a fresh perspective on how to approach these kinds of problems, it has been incredibly helpful for my way of thinking!!

> We're generally not in the business of "fixing" a bad input / misconfiguration from the user. It seems like it would be a nice thing to do, but it's risky.

I get what you mean.

> Better to do exactly what is asked of us, even if that's an error, and force the user to fix their own config. 

I agree with your point. This will provide very clear outcome, it’s good. I’ll make an update.
I noticed that the installer has this fallback behavior: in regions that don‘t support zones, if the user specifies zones in the install-config, the installer will deploy using non-zoned instead, without giving an explicit error. At the time, I was a bit puzzled why it didn’t report an error, but I suppose there might be a reason for handling it this way :)


> Note: Sometimes it can be a bit fuzzy. Interpreting "no zones specified" as "use all available zones" may seem like "fixing a bad input". We're considering it "a reasonable default" instead :)

I’m glad you told me this.

pkg/controller/machinepool/azureactuator.go
huangmingxia: Thank you, I’ll update it.

pkg/controller/machinepool/azureactuator.go
huangmingxia: Yeah, that’s true - I noticed it at the time, but my thinking wasn’t quite right then. I was trying to validate the zones first, so it seemed like I needed to call getZones every time. Now it’s much clearer, and I know how to handle it :)

pkg/controller/machinepool/azureactuator.go
2uasimojo: > I noticed that the installer has this fallback behavior: in regions that don‘t support zones, if the user specifies zones in the install-config, the installer will deploy using non-zoned instead, without giving an explicit error. 

Oh? That's interesting, and makes me want to think twice about the approach I suggested. Let's chat with the installer team and a) try to understand their rationale here, b) get their advice as to how they think we should handle this situation on the hive side.

huangmingxia: Yes, here is the installer code:
https://github.com/openshift/installer/blob/main/pkg/asset/machines/azure/machinesets.go#L27-L31 

In AWS and GCP, all regions are assumed to support zones by default. 
In Azure does the installer support deploying in regions without zones.

pkg/controller/machinepool/azureactuator.go
2uasimojo: It made sense with your initial design.

pkg/controller/machinepool/azureactuator.go
2uasimojo: Oh, that code is defaulting to zone-less if the user doesn't provide any explicitly. Was that the link you intended to paste? (It's the same as the one I mentioned [below](https://github.com/openshift/hive/pull/2732#discussion_r2283098646).) You said:

> in regions that don‘t support zones, if the user specifies zones in the install-config, the installer will deploy using non-zoned instead

Do you have a pointer to the code that does this? I.e. ignores mpool.Zones if the region has no zones? Or is it a behavior you observed when testing?

2uasimojo: [Thread](https://redhat-internal.slack.com/archives/CHFF1LCUR/p1755625267027199) with installer folk.

huangmingxia: > Was that the link you intended to paste? (It's the same as the one I mentioned https://github.com/openshift/hive/pull/2732#discussion_r2283098646.)

Yes, it's same :)

> Do you have a pointer to the code that does this? I.e. ignores mpool.Zones if the region has no zones? Or is it a behavior you observed when testing?

Before fixing this bug, I first reviewed the logic for handling regions without zones in the AWS, GCP, and Azure installers. I found that only the Azure installer supports non-zone deployments, therefore, the Hive side for AWS and GCP is fine.

I also performed a local test, using the following install-config file, in which the region does not have any zones assigned:
No zones in region:
```
mihuang@mihuang-mac hive % az vm list-skus -l usgovtexas --size Standard_D8s_v3 --zone --query '[].locationInfo[].zones'
[]
mihuang@mihuang-mac hive % az vm list-skus -l usgovtexas --size Standard_E4s_v3 --zone --query '[].locationInfo[].zones'
[]
```

```
    install-config.yaml: |
      apiVersion: v1
      baseDomain: qemag.azure.devcluster.openshift.com
      compute:
      - name: worker
        platform:
          azure:
            osDisk:
              diskSizeGB: 0
              diskType: ""
            type: ""
        replicas: 3
      controlPlane:
        name: master
        platform:
          azure:
            osDisk:
              diskSizeGB: 0
              diskType: ""
            type: ""
        replicas: 3
      metadata:
        creationTimestamp: null
        name: mihuanggov
      networking:
        clusterNetwork:
        - cidr: 10.128.0.0/14
          hostPrefix: 23
        machineNetwork:
        - cidr: 10.0.0.0/16
        serviceNetwork:
        - 172.30.0.0/16
      platform:
        azure:
          baseDomainResourceGroupName: os4-common
          cloudName: AzureUSGovernmentCloud
          outboundType: ""
          region: usgovtexas
          zones:
          - "1"
          - "2"
          - "3"
```

These are some logs after the cluster installation was completed. The machines show zones 0 and 1.  These are not the actual zones of the cluster's region. I suspect they are set by the Azure Machine Controller at runtime, based on the actual Azure VM information.

```
% oc get machines -A
NAMESPACE               NAME                                       PHASE     TYPE              REGION       ZONE   AGE
openshift-machine-api   mihuanggov-vblf8-master-0                  Running   Standard_D8s_v3   usgovtexas   1      134m
openshift-machine-api   mihuanggov-vblf8-master-1                  Running   Standard_D8s_v3   usgovtexas   0      134m
openshift-machine-api   mihuanggov-vblf8-master-2                  Running   Standard_D8s_v3   usgovtexas   0      134m
openshift-machine-api   mihuanggov-vblf8-worker-usgovtexas-6mlvh   Running   Standard_E4s_v3   usgovtexas   0      22m
openshift-machine-api   mihuanggov-vblf8-worker-usgovtexas-9btvz   Running   Standard_E4s_v3   usgovtexas   1      22m
openshift-machine-api   mihuanggov-vblf8-worker-usgovtexas-f62kd   Running   Standard_E4s_v3   usgovtexas   0      22m
```

```
% oc get machines -n openshift-machine-api mihuanggov-vblf8-master-0 -o yaml
apiVersion: machine.openshift.io/v1beta1
kind: Machine
metadata:
  annotations:
    machine.openshift.io/instance-state: Running
  creationTimestamp: "2025-08-15T15:51:03Z"
  finalizers:
  - machine.machine.openshift.io
  generation: 3
  labels:
    machine.openshift.io/cluster-api-cluster: mihuanggov-vblf8
    machine.openshift.io/cluster-api-machine-role: master
    machine.openshift.io/cluster-api-machine-type: master
    machine.openshift.io/instance-type: Standard_D8s_v3
    machine.openshift.io/region: usgovtexas
    machine.openshift.io/zone: "1"
  name: mihuanggov-vblf8-master-0
```

2uasimojo: Okay, this is pretty interesting.

In the installer log for this install-config.yaml, do you see a message about strict unmarshaling failing due to unknown fields? From what I can [see](https://github.com/openshift/hive/blob/01742053d4fc47b94cb68ab315d850f256b49ce7/vendor/github.com/openshift/installer/pkg/types/azure/platform.go#L31), there's no such thing as `platform.azure.zones`. So what I would expect is for the strict unmarshaling to fail, whereupon it will fall back to... um, "unstrict"? ...meaning that unknown fields, like `platform.azure.zones`, are ignored and dropped. So the effect would be as if you didn't include them.

In order to exercise the behavior we're looking for here, I think you would need to put the zones under one or more of:
- `platform.azure.defaultMachinePlatform.zones`
- `compute[].platform.azure.zones`
- `controlPlane.platform.azure.zones`

Can you try that and see if it results in an error?

----

If this region has no zones, I'm mystified by the `0`/`1` in the output table. Can you repeat your `az` query with a region that _does_ have zones so I can see what that output looks like?

huangmingxia: Thanks, I’ve updated the install-config, and the cluster is now being created.
```
    install-config.yaml: |
      apiVersion: v1
      baseDomain: qemag.azure.devcluster.openshift.com
      compute:
      - name: worker
        platform:
          azure:
            zones:
            - "1"
            - "2"
            - "3"
            osDisk:
              diskSizeGB: 0
              diskType: ""
            type: ""
        replicas: 3
      controlPlane:
        name: master
        platform:
          azure:
            zones:
            - "1"
            - "2"
            - "3"
            osDisk:
              diskSizeGB: 0
              diskType: ""
            type: ""
        replicas: 3
      metadata:
        creationTimestamp: null
        name: mihuanggov
      networking:
        clusterNetwork:
        - cidr: 10.128.0.0/14
          hostPrefix: 23
        machineNetwork:
        - cidr: 10.0.0.0/16
        serviceNetwork:
        - 172.30.0.0/16
      platform:
        azure:
          baseDomainResourceGroupName: os4-common
          cloudName: AzureUSGovernmentCloud
          outboundType: ""
          region: usgovtexas
          defaultMachinePlatform:
            zones:
            - "1"
            - "2"
            - "3"
```

> In the installer log for this install-config.yaml, do you see a message about strict unmarshaling failing due to unknown fields?

At that time, I didn’t check the provision pod logs.

huangmingxia: Region with zones:
```
$ az vm list-skus -l usgovvirginia --size Standard_E4s_v3 --zone --query '[].locationInfo[].zones'
[
  [
    "1",
    "3",
    "2"
  ]
]
```

huangmingxia: So sorry, after switching to the correct configuration, the cluster installation still failed. The error was:
```
time="2025-08-19T20:53:35Z" level=warning msg="Condition Ready has status: \"False\", reason: \"Failed\", message: \"virtualmachine failed to create or update. err: failed to create or update resource mihuanggov-xsr2k-rg/mihuanggov-xsr2k-bootstrap (service: virtualmachine): PUT https://management.usgovcloudapi.net/subscriptions/8fe0c1b4-8b05-4ef7-8129-7cf5680f27e7/resourceGroups/mihuanggov-xsr2k-rg/providers/Microsoft.Compute/virtualMachines/mihuanggov-xsr2k-bootstrap\\n--------------------------------------------------------------------------------\\nRESPONSE 400: 400 Bad Request\\nERROR CODE: LocationNotSupportAvailabilityZones\\n--------------------------------------------------------------------------------\\n{\\n  \\\"error\\\": {\\n    \\\"code\\\": \\\"LocationNotSupportAvailabilityZones\\\",\\n    \\\"message\\\": \\\"The resource 'Microsoft.Compute/virtualMachines/mihuanggov-xsr2k-bootstrap' does not support availability zones at location 'usgovtexas'.\\\"\\n  }\\n}\\n--------------------------------------------------------------------------------\\n\""
```

> In the installer log for this install-config.yaml, do you see a message about strict unmarshaling failing due to unknown fields? From what I can [see](https://github.com/openshift/hive/blob/01742053d4fc47b94cb68ab315d850f256b49ce7/vendor/github.com/openshift/installer/pkg/types/azure/platform.go#L31), there's no such thing as platform.azure.zones. So what I would expect is for the strict unmarshaling to fail, whereupon it will fall back to... um, "unstrict"? ...meaning that unknown fields, like platform.azure.zones, are ignored and dropped. So the effect would be as if you didn't include them.

The installer ignored the invalid  `platform.azure.zones` field, and proceeded with a non-zoned deployment. 

Thank you for clarifying my doubts, and I apologize as well.  I’ll update the code
> Better to do exactly what is asked of us, even if that's an error, and force the user to fix their own config.

2uasimojo: That's good news! Thanks for patiently working through it.

pkg/controller/machinepool/azureactuator.go
```diff
@@ -156,10 +156,12 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
 			return nil, false, errors.Wrap(err, "compute pool not providing list of zones and failed to fetch list of zones")
 		}
 		if len(zones) == 0 {
-			return nil, false, fmt.Errorf("zero zones returned for region %s", cd.Spec.Platform.Azure.Region)
+			computePool.Platform.Azure.Zones = []string{""}
+			logger.WithField("region", cd.Spec.Platform.Azure.Region).Info("Region does not support availability zones. Using non-zoned deployment.")
 		}
 		computePool.Platform.Azure.Zones = zones
```
2uasimojo: I think we need to wrap this in an `else`. Otherwise the branch above falls through and we'll override what we did on L159.

That said, remember that the upstream code will do this for us if `zones` is empty... so we're getting the right behavior anyway :)

So for the sake of cleaner code, I think we remove L159. And if you like, include a comment here mentioning the defaulting behavior.

2uasimojo (commented):
/lgtm

PR #2734, opened 2025-08-19
HIVE-2891: Add new HiveConfig field "HiveImagePullSecretRef"

dlom:

/assign @2uasimojo 
/cc @huangmingxia 

When deploying hive from a private image repository, in addition to supplying an `imagePullSecret` to the hive operator deployment, users will need to specify this same secret on the `HiveConfig`.  With this change, private image users will be able to provision clusters from any namespace, not just Hive's namespace

### DIFF
--- a/apis/hive/v1/hiveconfig_types.go
+++ b/apis/hive/v1/hiveconfig_types.go
@@ -41,6 +41,14 @@ type HiveConfigSpec struct {
 	// +optional
 	GlobalPullSecretRef *corev1.LocalObjectReference `json:"globalPullSecretRef,omitempty"`
 
+	// HiveImagePullSecretRef is used to specify a pull secret that can be used to pull Hive's own image.
+	// If hive has been deployed from a private registry, cluster installations will not succeed unless
+	// this reference is specified. This secret must live in the same namespace as the Hive operator's deployment.
+	// NOTE: This secret will be copied into Hive's TargetNamespace (if it is different than the namespace
+	// that the Hive operator is deployed in), as well as the namespace of every ClusterDeployment.
+	// +optional
+	HiveImagePullSecretRef *corev1.LocalObjectReference `json:"hiveImagePullSecretRef,omitempty"`
+
 	// Backup specifies configuration for backup integration.
 	// If absent, backup integration will be disabled.
 	// +optional
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -116,6 +116,9 @@ const (
 	// GlobalPullSecret is the environment variable for controllers to get the global pull secret
 	GlobalPullSecret = "GLOBAL_PULL_SECRET"
 
+	// HivePrivateImagePullSecret is the environment variable for controllers to get the pull secret for hive's own image
+	HivePrivateImagePullSecret = "HIVE_PRIVATE_IMAGE_PULL_SECRET"
+
 	// DefaultHiveNamespace is the default namespace where core hive components will run. It is used if the environment variable is not defined.
 	DefaultHiveNamespace = "hive"
 
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -774,6 +774,18 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	cdLog.Debug("checking for hive private image pull secret...")
+	switch requeue, err := r.ensurePrivateImagePullSecret(cd, cdLog); {
+	case err != nil:
+		cdLog.WithError(err).Log(controllerutils.LogLevel(err), "Error ensuring private image pull secret")
+		return reconcile.Result{}, err
+	case requeue:
+		cdLog.Debug("hive private image pull secret was out of date, requeueing...")
+		// The controller will not automatically requeue the cluster deployment
+		// since the controller is not watching for secrets. So, requeue manually.
+		return reconcile.Result{Requeue: true}, nil
+	}
+
 	// let's verify the release image before using it here.
 	if r.releaseImageVerifier != nil {
 		var releaseDigest string
@@ -2116,6 +2128,29 @@ func (r *ReconcileClusterDeployment) updatePullSecretInfo(pullSecret string, cd 
 		cdLog.WithField("secretName", mergedSecretName).Info("Created the merged pull secret object successfully")
 	}
 	return true, nil
+}
+
+func (r *ReconcileClusterDeployment) ensurePrivateImagePullSecret(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (bool, error) {
+	secretName := os.Getenv(constants.HivePrivateImagePullSecret)
+	if secretName == "" {
+		cdLog.Debug("Hive private image pull secret missing or empty, ignoring...")
+		return false, nil
+	}
+
+	srcNamespace := controllerutils.GetHiveNamespace()
+	destNamespace := cd.GetNamespace()
+
+	cdLog.
+		WithField("secretName", secretName).
+		WithField("sourceNamespace", srcNamespace).
+		WithField("targetNamespace", destNamespace).
+		WithField("targetCD", cd.Name).
+		Debug("Ensuring hive private image pull secret exists")
+
+	src := types.NamespacedName{Name: secretName, Namespace: srcNamespace}
+	dest := types.NamespacedName{Name: secretName, Namespace: destNamespace}
+
+	return controllerutils.CopySecret(r, src, dest, cd, r.scheme)
 }
 
 func configureTrustedCABundleConfigMap(cm *corev1.ConfigMap) bool {
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 	"text/template"
@@ -306,11 +307,11 @@ func EnsureRequeueAtLeastWithin(duration time.Duration, result reconcile.Result,
 	return reconcile.Result{RequeueAfter: duration, Requeue: true}, nil
 }
 
-// CopySecret copies the secret defined by src to dest.
-func CopySecret(c client.Client, src, dest types.NamespacedName, owner metav1.Object, scheme *runtime.Scheme) error {
+// CopySecret copies the secret defined by src to dest.  Returns true if an actual change was made.
+func CopySecret(c client.Client, src, dest types.NamespacedName, owner metav1.Object, scheme *runtime.Scheme) (bool, error) {
 	srcSecret := &corev1.Secret{}
 	if err := c.Get(context.Background(), src, srcSecret); err != nil {
-		return err
+		return false, err
 	}
 
 	destSecret := &corev1.Secret{}
@@ -327,23 +328,42 @@ func CopySecret(c client.Client, src, dest types.NamespacedName, owner metav1.Ob
 		}
 		if owner != nil {
 			if err := controllerutil.SetOwnerReference(owner, destSecret, scheme); err != nil {
-				return err
+				return false, err
 			}
 		}
 
-		return c.Create(context.Background(), destSecret)
-	}
-	if err != nil {
-		return err
+		return true, c.Create(context.Background(), destSecret)
+	} else if err != nil {
+		return false, err
 	}
 
+	// If the two secrets match...
 	if reflect.DeepEqual(destSecret.Data, srcSecret.Data) && reflect.DeepEqual(destSecret.StringData, srcSecret.StringData) {
-		return nil // no work as the dest and source data matches.
+		// and destination has no new owner...
+		if owner == nil {
+			// no work as the dest and source data matches, and no owner.
+			return false, nil
+		} else {
+			if slices.ContainsFunc(destSecret.OwnerReferences, func(ref metav1.OwnerReference) bool {
+				return ref.UID == owner.GetUID()
+			}) {
+				return false, nil // no work as the dest and source data matches and secret is owned
+			}
+		}
+
+		// We still need to register the new owner, down below:
+	} else {
+		destSecret.Data = srcSecret.DeepCopy().Data
+		destSecret.StringData = srcSecret.DeepCopy().StringData
 	}
 
-	destSecret.Data = srcSecret.DeepCopy().Data
-	destSecret.StringData = srcSecret.DeepCopy().StringData
-	return c.Update(context.Background(), destSecret)
+	// If we've made it this far, unconditionally set the owner ref
+	// (controllerutil handles deduplication etc)
+	if err := controllerutil.SetOwnerReference(owner, destSecret, scheme); err != nil {
+		return false, err
+	}
+
+	return true, c.Update(context.Background(), destSecret)
 }
 
 // BuildControllerLogger returns a logger for controllers with consistent fields.
--- a/pkg/imageset/generate.go
+++ b/pkg/imageset/generate.go
@@ -1,6 +1,7 @@
 package imageset
 
 import (
+	"os"
 	"time"
 
 	controllerutils "github.com/openshift/hive/pkg/controller/utils"
@@ -41,8 +42,6 @@ func GenerateImageSetJob(
 			MountPath: "/common",
 		},
 	}
-
-	imagePullSecrets := append(sharedPodConfig.ImagePullSecrets, corev1.LocalObjectReference{Name: constants.GetMergedPullSecretName(cd)})
 
 	podSpec := corev1.PodSpec{
 		NodeSelector:  sharedPodConfig.NodeSelector,
@@ -87,7 +86,13 @@ func GenerateImageSetJob(
 			},
 		},
 		ServiceAccountName: serviceAccountName,
-		ImagePullSecrets:   imagePullSecrets,
+		ImagePullSecrets: []corev1.LocalObjectReference{
+			{Name: constants.GetMergedPullSecretName(cd)},
+		},
+	}
+
+	if hiveImagePullSecret := os.Getenv(constants.HivePrivateImagePullSecret); hiveImagePullSecret != "" {
+		podSpec.ImagePullSecrets = append(podSpec.ImagePullSecrets, corev1.LocalObjectReference{Name: hiveImagePullSecret})
 	}
 
 	completions := int32(1)
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -3,6 +3,7 @@ package install
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
@@ -72,7 +73,8 @@ func CopyAWSServiceProviderSecret(client client.Client, destNamespace string, en
 
 	src := types.NamespacedName{Name: spSecretName, Namespace: hiveNS}
 	dest := types.NamespacedName{Name: destSecretName, Namespace: destNamespace}
-	return controllerutils.CopySecret(client, src, dest, owner, scheme)
+	_, err := controllerutils.CopySecret(client, src, dest, owner, scheme)
+	return err
 }
 
 // AWSAssumeRoleConfig creates or updates a secret with an AWS credentials file containing:
@@ -413,8 +415,15 @@ func InstallerPodSpec(
 		Containers:         containers,
 		Volumes:            volumes,
 		ServiceAccountName: serviceAccountName,
-		ImagePullSecrets:   []corev1.LocalObjectReference{{Name: constants.GetMergedPullSecretName(cd)}},
+		ImagePullSecrets: []corev1.LocalObjectReference{
+			{Name: constants.GetMergedPullSecretName(cd)},
+		},
 	}
+
+	if hiveImagePullSecret := os.Getenv(constants.HivePrivateImagePullSecret); hiveImagePullSecret != "" {
+		podSpec.ImagePullSecrets = append(podSpec.ImagePullSecrets, corev1.LocalObjectReference{Name: hiveImagePullSecret})
+	}
+
 	controllerutils.SetProxyEnvVars(podSpec, httpProxy, httpsProxy, noProxy)
 	return podSpec, nil
 }
@@ -442,7 +451,6 @@ func GenerateInstallerJob(
 	podSpec := provision.Spec.PodSpec
 	podSpec.NodeSelector = sharedPodConfig.NodeSelector
 	podSpec.Tolerations = sharedPodConfig.Tolerations
-	podSpec.ImagePullSecrets = append(podSpec.ImagePullSecrets, sharedPodConfig.ImagePullSecrets...)
 
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
@@ -493,7 +501,10 @@ func GenerateUninstallerJobForDeprovision(
 		ServiceAccountName: serviceAccountName,
 		NodeSelector:       sharedPodConfig.NodeSelector,
 		Tolerations:        sharedPodConfig.Tolerations,
-		ImagePullSecrets:   sharedPodConfig.ImagePullSecrets,
+	}
+
+	if hiveImagePullSecret := os.Getenv(constants.HivePrivateImagePullSecret); hiveImagePullSecret != "" {
+		podSpec.ImagePullSecrets = []corev1.LocalObjectReference{{Name: hiveImagePullSecret}}
 	}
 
 	completions := int32(1)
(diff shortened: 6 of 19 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2118,6 +2129,28 @@ func (r *ReconcileClusterDeployment) updatePullSecretInfo(pullSecret string, cd
...
+	cdLog.
+		WithField("secretName", secretName).
+		WithField("sourceNamespace", srcNamespace).
+		WithField("targetNamespace", destNamespace).
+		Debug("Ensuring hive private image pull secret exists")
+
+	src := types.NamespacedName{Name: secretName, Namespace: srcNamespace}
+	dest := types.NamespacedName{Name: secretName, Namespace: destNamespace}
+
+	// TODO: cross-NS ownership reference? is that even possible?
```
2uasimojo: It is not. However, we should parent it to the CD. There is precedent, and a handy helper function. See for example [how we do this](https://github.com/openshift/hive/blob/d09e43e7185bfc339838fc3ac0552adbcd4a83ab/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L2353) for the pull secret.

2uasimojo (commented):
I think this PR also needs to touch your [previous efforts](https://github.com/openshift/hive/pull/2725) to reference this secret from the hive-operator pod. We discussed how copying `imagePullSecrets` across namespaces will end us up with a reference to a dockercfg Secret that doesn't (and shouldn't) exist in the destination namespace. So I think we need to refactor all of those places to simply add the value of `hiveConfig.hiveImagePullSecretRef` to the `imagePullSecrets` of controllers, admission, imageset, provision, and deprovision.

We also need to address copying the Secret to the `targetNamespace` iff that namespace is different from where hive-operator is running.

Once all of that is done, I think we can revert some of your additions that look up the current pod -- they should no longer be needed. (However, I like the way you rolled up sharedPodConfig; I would like to keep that for tolerations/nodeSelector.)

dlom:
> we need to refactor all of those places to simply add the value of hiveConfig.hiveImagePullSecretRef to the imagePullSecrets of controllers, admission, imageset, provision, and deprovision

@2uasimojo after blindly forging ahead on this all day, I don't think this is possible either.  The `imagePullSecret` by definition only refers to local secrets in the same NS.  If the required secret is only in the hive NS, it really MUST be copied into the CD's namespace so that the imageset job can run in that NS

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dlom: Still haven't figured this one out.  Did you link the wrong thing, or am I blind?

2uasimojo: Bleh, sorry, [correct reference](https://github.com/openshift/hive/blob/d09e43e7185bfc339838fc3ac0552adbcd4a83ab/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L2305).

As we discussed, "controller reference" is not the same as "owner reference", and I think we want the former. The purpose is for the Secret to automatically get deleted/garbage-collected when the parent (the CD) goes away.

2uasimojo:
> > we need to refactor all of those places to simply add the value of hiveConfig.hiveImagePullSecretRef to the imagePullSecrets of controllers, admission, imageset, provision, and deprovision
> 
> @2uasimojo after blindly forging ahead on this all day, I don't think this is possible either. The `imagePullSecret` by definition only refers to local secrets in the same NS. If the required secret is only in the hive NS, it really MUST be copied into the CD's namespace so that the imageset job can run in that NS

Per our Meet™, clarified summary:
- #2725 copies `imagePullSecrets` (a list of names) from the current pod to the descendant pods. We don't want to do this, because it'll drag along the name of the generated dockercfg Secret. Rather, I think we ignore the current pod's `imagePullSecrets` and simply inject the value of your new hiveconfig field into the descendant pods' `imagePullSecrets`. This goes for both "stages": hive-operator => controllers/admission and controllers => imageset/prov/deprov.
- The hive-operator and the controllers/admission aren't necessarily running in the same namespace. HiveConfig.Spec.TargetNamespace defaults to `hive`, but
  - You might have deployed hive-operator somewhere other than `hive`; and/or
  - If TargetNamespace is explicitly set, we'll deploy controllers/admission there instead.

  So if this is the case, we need to make sure we're copying the Secret from hive-operator's ns to the TargetNamespace.
- Set controller reference on the copied Secrets as discussed [above](https://github.com/openshift/hive/pull/2734#discussion_r2292130866). (Uhh, I think for the controllers/admission copy, it's probably appropriate to parent to hive-controllers? I think the test would be: Edit HiveConfig.Spec.TargetNamespace, and make sure the Secret disappears from the old ns.)

dlom:
I've done an initial smoke test on this, with the operator, controllers, and CD all being in different non-default namespaces.  The operator uses a label on the controller-level secret to clean up if the controllers move, and the CDs each get an individually named secret copy that is owned by the CD

2uasimojo:
/test e2e

Actual test succeeded; we flaked afterward.

apis/hive/v1/hiveconfig_types.go
```diff
@@ -41,6 +41,14 @@ type HiveConfigSpec struct {
 	// +optional
 	GlobalPullSecretRef *corev1.LocalObjectReference `json:"globalPullSecretRef,omitempty"`
 
+	// HiveImagePullSecretRef is used to specify a pull secret that can be used to pull Hive's own image.
+	// If hive has been deployed from a private registry, cluster installations will not succeed unless
+	// this reference is specified. This secret must live in Hive's TargetNamespace.
```
2uasimojo: Is this true, or does it need to live in the namespace with hive-operator?

pkg/constants/constants.go
```diff
@@ -566,3 +570,13 @@ const (
...
+
+// GetImagePullSecretName returns name for image pull secret name per cluster deployment
+func GetImagePullSecretName(cd *hivev1.ClusterDeployment) string {
+	return apihelpers.GetResourceName(cd.Name, hiveImagePullSecretSuffix)
+}
+
+// GetImagePullSecretNameForDeprovision returns name for image pull secret name per cluster deprovision
+func GetImagePullSecretNameForDeprovision(cd *hivev1.ClusterDeprovision) string {
+	return apihelpers.GetResourceName(cd.Name, hiveImagePullSecretSuffix)
+}
```
2uasimojo: I think we can DRY these by making the param a meta.Object and using GetName() on it.

(Moot if you copy the Secret down just once with its original name, as [noted](https://github.com/openshift/hive/pull/2734#discussion_r2326063086).)

apis/hive/v1/hiveconfig_types.go
```diff
@@ -41,6 +41,14 @@ type HiveConfigSpec struct {
...
+	// NOTE: This secret will be copied into the namespace of every ClusterDeployment, overwriting any secret
+	// with the same name.
```
2uasimojo: I don't think this is true either, as it looks like you're naming those Secrets after the cd/deprovision.

Though thinking about it, I think I like this idea better. There's nothing stopping multiple CDs being created in the same namespace, and we would rather have just one copy of this pull secret per namespace.

pkg/install/generate.go
```diff
@@ -717,6 +719,7 @@ func completeAWSDeprovisionJob(req *hivev1.ClusterDeprovision, job *batchv1.Job)
 	job.Spec.Template.Spec.InitContainers = initContainers
 	job.Spec.Template.Spec.Containers = containers
 	job.Spec.Template.Spec.Volumes = volumes
+	job.Spec.Template.Spec.ImagePullSecrets = append(job.Spec.Template.Spec.ImagePullSecrets, corev1.LocalObjectReference{Name: constants.GetImagePullSecretNameForDeprovision(req)})
```
2uasimojo: Two things:
1. Why not do this in the calling func instead of here? I think all the other pieces are being done per-`complete*DeprovisionJob()` method because they can be different. (If you find otherwise, I would be receptive to a separate PR that pulls them out.) But this should always be the same, yah?
2. This is working because we happen to give the ClusterDeprovision the [same name](https://github.com/openshift/hive/blob/b69fcfef87fc9f72cb021b4bfa3f86e59a46e2ef/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L1918) as the ClusterDeployment. But it reads as if we're potentially referencing a _different_ Secret -- and one we haven't copied in. I can see that it would be painful to get the CD down to these funcs, so I'm okay leaving the logic as is, *but* I would like to see a comment that calls out that it's relying on those two things having the same name. [Later] But moot if we're using the original Secret name as noted above.

pkg/operator/hive/hive.go
```diff
@@ -42,6 +43,13 @@ const (
 	// hiveConfigHashAnnotation is annotation on hivedeployment that contains
 	// the hash of the contents of the hive-controllers-config configmap
 	hiveConfigHashAnnotation = "hive.openshift.io/hiveconfig-hash"
+
+	// hiveImagePullSecretLabel is the key (the value will always be "true") for the label we add to
+	// the secret (named in HiveConfig.Spec.HiveImagePullSecretRef) that is copied into the target
+	// hive namespace (named in HiveConfig.Spec.TargetNamespace). We use this to identify secrets in
+	// namespaces that were *previously* the configured TargetNamespace so we can clean
+	// them up.
```
2uasimojo: Why are we not just using owner references? I guess they would have to be on one of the Deployments/StatefulSets, but that should work, shouldn't it?

pkg/operator/hive/hive.go
```diff
@@ -353,7 +375,9 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 	// Apply shared pod config passed through from the operator deployment
 	hiveDeployment.Spec.Template.Spec.NodeSelector = r.sharedPodConfig.NodeSelector
 	hiveDeployment.Spec.Template.Spec.Tolerations = r.sharedPodConfig.Tolerations
-	hiveDeployment.Spec.Template.Spec.ImagePullSecrets = r.sharedPodConfig.ImagePullSecrets
+	if hiveConfigHasValidImagePullSecretReference(instance) {
+		hiveDeployment.Spec.Template.Spec.ImagePullSecrets = append(hiveDeployment.Spec.Template.Spec.ImagePullSecrets, *instance.Spec.HiveImagePullSecretRef)
+	}
```
2uasimojo: Just a thought, if you made the func return the ref instead of just checking for it, you could do something like:
```suggestion
	if ref := getImagePullSecretReference(instance); ref != nil {
		hiveDeployment.Spec.Template.Spec.ImagePullSecrets = append(hiveDeployment.Spec.Template.Spec.ImagePullSecrets, ref)
	}
```

pkg/operator/hive/hive.go
```diff
@@ -464,6 +492,76 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+
+	hiveImagePullSecretEnvVar := corev1.EnvVar{
+		Name:  constants.HivePrivateImagePullSecret,
+		Value: instance.Spec.HiveImagePullSecretRef.Name,
+	}
+	hiveContainer.Env = append(hiveContainer.Env, hiveImagePullSecretEnvVar)
+}
+
+func (r *ReconcileHiveConfig) dynamicCopyHiveImagePullSecret(hLog log.FieldLogger, instance *hivev1.HiveConfig) error {
```
2uasimojo: Okay, yeah, let's clean this up.

For the getting, you should be able to use the existing controller-runtime-esque wrapper [here](https://github.com/openshift/hive/blob/b69fcfef87fc9f72cb021b4bfa3f86e59a46e2ef/pkg/operator/hive/dynamicclient.go#L66). Example [Get()](https://github.com/openshift/hive/blob/b69fcfef87fc9f72cb021b4bfa3f86e59a46e2ef/pkg/operator/hive/hive.go#L589).

For the creating, you should be able to write a similar `Create()` wrapper in that file.

The end result should be that you can make this func feel a lot more like it would if `r` had/was a real c-r client.

pkg/operator/hive/operatorutils.go
```diff
@@ -88,3 +90,41 @@ func applyDeploymentConfig(hiveConfig *hivev1.HiveConfig, deploymentName hivev1.
 		container.Resources = *dc.Resources
 	}
 }
+
+func synchronizeUnstructuredSecrets(source, destination *unstructured.Unstructured) bool {
```
2uasimojo: ...and then all of this goes away.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +492,83 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+		destSecret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: destNS,
+				Labels: map[string]string{
+					hiveImagePullSecretLabel: "true",
+				},
+			},
+			Data:       srcSecret.DeepCopy().Data,
+			StringData: srcSecret.DeepCopy().StringData,
```
2uasimojo: Don't use [StringData](https://github.com/openshift/hive/blob/065cd2894eb3254bfd4a940a9da9691fe6b709aa/vendor/k8s.io/api/core/v1/types.go#L7557-L7563) here. Have a look at the docstring for why. (I think it ends up being a no-op because srcSecret.StringData will be nil, so harmless, but redundant.)

Also, I don't think we need to be making a DeepCopy() (or any kind of copy) of the Data. The only reason that would be necessary is if we were modifying one of them in a way that would affect the other. It's broadly a good practice for data safety if the objects are flying around different contexts where you don't necessarily have visibility/control over how they're being used, but in this case everything is tightly held within this func, so we can be assured the srcSecret is read-only and the destSecret is write-only.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +492,83 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+		lvalue == "true" &&
+		reflect.DeepEqual(destSecret.Data, srcSecret.Data) &&
+		reflect.DeepEqual(destSecret.StringData, srcSecret.StringData) {
+		hLog.Debugf("target secret %s in NS %s does exist, no update necessary", secretName, destNS)
+		return nil // no work as the dest and source data matches.
+	}
+
+	hLog.Debugf("target secret %s in NS %s does exist, updating it", secretName, destNS)
+	destSecret.Data = srcSecret.DeepCopy().Data
+	destSecret.StringData = srcSecret.DeepCopy().StringData
```
2uasimojo: As above.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +492,83 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+	destSecret.ObjectMeta.Labels[hiveImagePullSecretLabel] = "true"
+
+	_, err = h.CreateOrUpdateRuntimeObject(destSecret, r.scheme)
```
2uasimojo, line 558: Since this already has "create or update" semantics, we could shrink this func down some, removing L547 and some of the copying around it.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +492,83 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	if lvalue, present := destSecret.ObjectMeta.Labels[hiveImagePullSecretLabel]; present &&
+		lvalue == "true" &&
+		reflect.DeepEqual(destSecret.Data, srcSecret.Data) &&
+		reflect.DeepEqual(destSecret.StringData, srcSecret.StringData) {
```
2uasimojo: As above.

apis/hive/v1/hiveconfig_types.go
dlom: I made it per-CD because then k8s will GC the secret when the CD no longer exists.  If we're okay with leaving the dangling secret in a CD namespace, then we can get away with one per NS

pkg/operator/hive/hive.go
dlom: The Deployment (and others) might not exist yet (in the cluster) when the secret is getting copied, thus they cannot be referenced

pkg/operator/hive/hive.go
dlom: I really hate what this function is becoming if I try to roll up everything into a single `CreateOrUpdate` call. The 3 codepaths become entangled and it's hard to read and reason about what's going on.  `Create`ing and exiting early is the simplest option IMO

dlom: You can see what I mean

apis/hive/v1/hiveconfig_types.go
2uasimojo: Per discussion: Let's try OwnerReferences per CD for the (single) Secret and see if that works.

pkg/operator/hive/hive.go
2uasimojo: Per discussion: Let's parent it to the hive-controllers ServiceAccount.

But only if `targetNamespace` != hive-operator's namespace. Cause we don't want to delete a Secret we didn't create.

pkg/controller/utils/utils.go
```diff
@@ -327,23 +327,37 @@ func CopySecret(c client.Client, src, dest types.NamespacedName, owner metav1.Ob
...
+
+		// We still need to register the new owner, down below:
+	} else {
+		destSecret.Data = srcSecret.DeepCopy().Data
+		destSecret.StringData = srcSecret.DeepCopy().StringData
+	}
+
+	// If we've made it this far, unconditionally set the owner ref
+	// (controllerutil handles deduplication etc)
+	if err := controllerutil.SetOwnerReference(owner, destSecret, scheme); err != nil {
```
2uasimojo, line 362: :+1: 

Did we determine that:
- this correctly garbage collects the Secret when the last CD goes away
- when a not-last CD goes away, its owner reference is scrubbed from the Secret automatically

?

My concern with the second point is that the Secret CR could grow indefinitely e.g. in a usage model where I always put all my CDs in the same namespace.

pkg/imageset/generate.go
```diff
@@ -87,7 +86,10 @@ func GenerateImageSetJob(
 			},
 		},
 		ServiceAccountName: serviceAccountName,
-		ImagePullSecrets:   imagePullSecrets,
+		ImagePullSecrets: []corev1.LocalObjectReference{
+			{Name: constants.GetMergedPullSecretName(cd)},
+			{Name: os.Getenv(constants.HivePrivateImagePullSecret)},
```
2uasimojo: I think we want to skip adding this entry if it's blank. From what I've seen, it might still work, but I'm guessing we would get a warning in the pod events, which would be confusing/annoying for users.

pkg/install/generate.go
```diff
@@ -413,7 +415,10 @@ func InstallerPodSpec(
 		Containers:         containers,
 		Volumes:            volumes,
 		ServiceAccountName: serviceAccountName,
-		ImagePullSecrets:   []corev1.LocalObjectReference{{Name: constants.GetMergedPullSecretName(cd)}},
+		ImagePullSecrets: []corev1.LocalObjectReference{
+			{Name: constants.GetMergedPullSecretName(cd)},
+			{Name: os.Getenv(constants.HivePrivateImagePullSecret)},
```
2uasimojo: As above.

pkg/install/generate.go
```diff
@@ -493,7 +497,9 @@ func GenerateUninstallerJobForDeprovision(
 		ServiceAccountName: serviceAccountName,
 		NodeSelector:       sharedPodConfig.NodeSelector,
 		Tolerations:        sharedPodConfig.Tolerations,
-		ImagePullSecrets:   sharedPodConfig.ImagePullSecrets,
+		ImagePullSecrets: []corev1.LocalObjectReference{
+			{Name: os.Getenv(constants.HivePrivateImagePullSecret)},
```
2uasimojo: As above.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +473,88 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+		hLog.Debug("HiveImagePullSecret is not provided in HiveConfig, it will not be copied")
+		return nil
+	}
+
+	srcNS := r.hiveOperatorNamespace
+	destNS := GetHiveNamespace(instance)
+	secretName := ref.Name
+
+	if srcNS == destNS {
+		hLog.Debugf("hive operator and hive controllers live in the same NS '%s'", srcNS)
```
2uasimojo: nit: structured logging
```suggestion
		hLog.WithField("namespace", srcNS).Debug("hive operator and hive controllers live in the same NS", srcNS)
```

Similar throughout.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +473,88 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+				Name:      secretName,
+				Namespace: destNS,
+			},
+			Data: srcSecret.Data,
+			Type: srcSecret.Type,
+		}
+		secretFound = false
+	}
+
+	if err != nil {
```
2uasimojo: I think this needs to be folded into the previous conditional; else you'll splode on a 404 which I think is not the intent.
```suggestion
	} else if err != nil {
```

pkg/operator/hive/hive.go
```diff
@@ -464,6 +473,88 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+			destSecret.Data = srcSecret.Data
+			hLog.Debugf("target secret %s in NS %s is out of date, updating it", secretName, destNS)
+		}
+	} else {
+		hLog.Debugf("target secret %s in NS %s doesn't exist, creating it", secretName, destNS)
+	}
+
+	controllerSA := &corev1.ServiceAccount{}
+	if err := r.Get(context.TODO(), types.NamespacedName{Namespace: destNS, Name: "hive-controllers"}, controllerSA); err != nil {
+		return err
```
2uasimojo: A log message here would be appropriate, to let the reader know we're waiting for the SA to exist.

pkg/operator/hive/hive.go
```diff
@@ -464,6 +473,88 @@ func (r *ReconcileHiveConfig) includeGlobalPullSecret(hLog log.FieldLogger, inst
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	if secretFound {
+		if reflect.DeepEqual(destSecret.Data, srcSecret.Data) {
+			hLog.Debugf("target secret %s in NS %s does exist, no update necessary", secretName, destNS)
+			return nil // no work as the dest and source data matches.
```
2uasimojo: This skips the case where we haven't injected the owner reference.

pkg/controller/utils/utils.go
dlom: Yes and yes, k8s seems to be handling everything automatically

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -774,6 +774,18 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	cdLog.Debug("checking for hive private image pull secret...")
```
2uasimojo, line 777: This might be a little chatty as it'll show up on pretty much every reconcile. But we can rip it out later if that turns out to be an issue.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -2118,6 +2130,29 @@ func (r *ReconcileClusterDeployment) updatePullSecretInfo(pullSecret string, cd
 	return true, nil
 }
 
+func (r *ReconcileClusterDeployment) ensurePrivateImagePullSecret(cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) (bool, error) {
+	secretName := os.Getenv(constants.HivePrivateImagePullSecret)
+	if secretName == "" {
+		cdLog.Debug("Hive private image pull secret missing or empty, ignoring...")
```
2uasimojo, line 2136: Ditto chatty, given most deployments will probably hit this path, but ditto later.

huangmingxia:
Hi @dlom @2uasimojo I ran the tests based on the private image, and recorded the results in the test case [OCP-84803](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitems?query=OCP-84803). Could you please take a look when you have time? Thanks!
Other platform tests are still in progress.

dlom:
@huangmingxia The test case and the results look good to me.  We're still waiting for our scoped secret to be created ([APPSRE-12488](https://issues.redhat.com//browse/APPSRE-12488)), so we can't deploy this yet anyways

dlom:
The secret is now available, merging this so we can deploy today :shipit:

PR #2736, opened 2025-08-20
HIVE-2790: ClusterInstall can opt out of imageset job

2uasimojo:
ClusterInstall implementors can now request that hive skip the imageset job by adding the following annotation to their {Concrete}ClusterInstall *CRD* (not *CR*!):

`contracts.hive.openshift.io/clusterinstall.config: '{"skipImageSetJob": "true"}'`

This will cause ClusterDeployment.Status fields InstallerImage and CLIImage to get the value `"<NONE>"` for all ClusterDeployments with a ClusterInstallRef pointing to any Version of that {Concrete}ClusterInstall's Group+Kind.


hack/e2e-test.sh
```diff
@@ -117,6 +117,14 @@ go run "${SRC_ROOT}/contrib/cmd/hiveutil/main.go" create-cluster "${CLUSTER_NAME
   | jq '.items[0].spec.controlPlaneConfig.apiURLOverride = "bogus-url.example.com"' \
   | oc apply -f -
 
+echo "Creating fake clusterinstall artifacts"
+# Use the hive image as the release image. This is bogus, and would cause the imageset job to fail
+# if skipImageSetJob were unset.
+# Reference the pull secret generated by hiveutil above.
+sed "s,RELEASE_IMAGE,$HIVE_IMAGE,; s,PULL_SECRET,${CLUSTER_NAME}-pull-secret," \
+  ${0%/*}/hack/fakeclusterinstall.yaml \
```
2uasimojo:
```suggestion
  ${0%/*}/fakeclusterinstall.yaml \
```

apis/hivecontracts/v1alpha1/clusterinstall_types.go
```diff
@@ -39,6 +39,12 @@ type ClusterInstallSpec struct {
 	// This must be set as soon as all the information is available.
 	// +optional
 	ClusterMetadata *hivev1.ClusterMetadata `json:"clusterMetadata"`
+
+	// SkipImageSetJob tells hive not to resolve payload image references (installer, oc) from the
+	// release image specified via ImageSetRef. ClusterDeployment status fields InstallerImage and
+	// CLIImage will be set to "<NONE>".
+	// +optional
+	SkipImageSetJob bool `json:"skipImageSetJob"`
```
carbonin: I'm not sure about this being in the spec since this won't be up to the user to set. My understanding is it will depend entirely on the expectations of the cluster install controller, right?

Maybe an annotation is better here? What do you think @2uasimojo ?

2uasimojo: Hm, you're right @carbonin, this decision would presumably be linked to the specific installation process; it wouldn't be necessary to be able to control it per cluster. As such, _anything_ associated with the {Concrete}ClusterInstall CR (even an annotation) doesn't seem appropriate. It would be better to somehow configure it alongside the specific {Concrete}ClusterInstall CR**D**.

Today we happen to have AgentClusterInstall [hardcoded](https://github.com/openshift/hive/blob/7d9bf8a09eccdf70176cda14243e1d35c17d77d6/pkg/operator/hive/hiveadmission.go#L332-L339) in our list of known ClusterInstall contracts. We also provide a [mechanism](https://github.com/openshift/hive/blob/99abe10fe0c49f1ae08ffcdff00eb4198483bf05/pkg/operator/hive/configmap.go#L151-L160) to register new ones dynamically via an annotation on the **CRD**. Perhaps we could use another annotation on the CRD that indicates that all installs via that CRD should skip the imageset job? And I could also hardcode that flag in for AgentClusterInstall. WDYT?

hack/e2e-test.sh
```diff
@@ -223,6 +231,12 @@ case "${INSTALL_RESULT}" in
         ;;
 esac
 
+# Check the fakeclusterinstall. It should have had plenty of time to succeed by now.
+if [[ $(oc get cd fci -o jsonpath={.spec.installed}) != "tnue" ]]; then
```
2uasimojo: wow
```suggestion
if [[ $(oc get cd fci -o jsonpath={.spec.installed}) != "true" ]]; then
```

pkg/operator/hive/configmap.go
```diff
@@ -121,15 +123,33 @@ var metricsConfigConfigMapInfo = configMapInfo{
...
+// require discovery using CRDs
+var knownContracts = contracts.SupportedContractImplementationsList{{
+	Name: hivecontractsv1alpha1.ClusterInstallContractName,
+	Supported: []contracts.ContractImplementation{{
+		Group:   "extensions.hive.openshift.io",
+		Version: "v1beta1",
+		Kind:    "AgentClusterInstall",
+		Config:  map[string]string{string(hivecontractsv1alpha1.ClusterInstallConfigKeySkipImageSetJob): "true"},
+	}},
+}}
```
2uasimojo, line 147: ntr: moved from hiveadmission, as this now applies to hive-controllers as well.

pkg/operator/hive/hive_controller.go
```diff
@@ -565,7 +565,7 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	err = r.deployHive(hLog, h, instance, namespacesToClean, confighash, managedDomainsConfigHash, fpConfigHash, mcConfigHash)
+	err = r.deployHive(hLog, h, instance, namespacesToClean, confighash, managedDomainsConfigHash, fpConfigHash, mcConfigHash, scConfigHash)
```
2uasimojo, line 568: ntr: trigger redeploy of controllers if supported contracts config changes (e.g. a CRD gets updated to add the new annotation).

pkg/util/contracts/contracts.go
```diff
@@ -56,6 +58,19 @@ func (l SupportedContractImplementationsList) IsSupported(contract string, impl
 	return false
 }
 
+func (l SupportedContractImplementationsList) GetConfig(contract string, impl ContractImplementation) map[string]string {
```
2uasimojo, line 61: nts: TODO: Refactor everything in this file. Use maps instead of lists. Consolidate this function with the previous.

config/crds/hiveinternal.openshift.io_fakeclusterinstalls.yaml
```diff
@@ -3,6 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.16.3
+    contracts.hive.openshift.io/clusterinstall.config: '{"skipImageSetJob": "true"}'
```
2uasimojo, line 6: ntr: Example of what this annotation looks like on a CRD.

2uasimojo (commented):
@carbonin How's this? (I'll squash & re-message the commits later if you're good with the design.)

apis/hivecontracts/v1alpha1/clusterinstall_types.go
carbonin: What exactly is the output of the image set job? Do we know if it is possibly useful with any *ClusterInstall strategy?

carbonin: Found it in the patch. Yeah, I think this new approach is fine. I'll just need to follow up and actually add the annotation.

carbonin (approved):
Didn't do a code review, but I'm good with the CRD annotation approach for opting out of the job.

apis/hivecontracts/v1alpha1/clusterinstall_types.go
2uasimojo: > What exactly is the output of the image set job?

It simply sets the ClusterDeployment status fields InstallerImage and CLIImage to point to the URIs of openshift-install and oc, respectively, from the release payload. These are subsequently used by the provisioning flow to install the cluster and to capture logs on failure, respectively. (Why do we need to do this in a separate Job prior to spawning the provisioning pod? I couldn't say. It was like that when I got here.)

> Do we know if it is possibly useful with any *ClusterInstall strategy?

We do not, which is why I wanted to be careful to make this new behavior opt-in. I could definitely contrive scenarios where it's useful.

> I'll just need to follow up and actually add the annotation.

I think that would be goodness; but note that I did already [add it](https://github.com/openshift/hive/pull/2736/files#diff-a27ba32cf6fc6000eca3eee82e4106ae4d6fe404d8fbb4a46b67a701ee8c576dR145) to the existing hardcoded registration of AgentClusterInstall.

2uasimojo:
/hold cancel

@carbonin I think this is ready. Feel free to do a deeper review if you like. Otherwise Suhani plans to hit it early next week.

carbonin:
> @carbonin I think this is ready. Feel free to do a deeper review if you like. Otherwise Suhani plans to hit it early next week.

Don't hold off on my account. Looks good to me

pkg/util/contracts/contracts.go
```diff
@@ -56,6 +58,19 @@ func (l SupportedContractImplementationsList) IsSupported(contract string, impl
...
+	for _, c := range l {
+		if c.Name == contract {
+			for _, i := range c.Supported {
+				if i.Group == impl.Group && i.Kind == impl.Kind && i.Version == impl.Version {
```
suhanime, line 65: Nit: Considering that config is part of `ContractImplementation` struct, but the comparison of implementations do not compare configs - it would have been neat to have a 1 liner function that did ^^ comparison and returned true or false - since that's done twice in this file.

pkg/operator/hive/configmap.go
```diff
@@ -138,25 +158,36 @@ func (r *ReconcileHiveConfig) supportedContractsConfigMapInfo() configMapInfo {
...
 					val == "true" &&
 					allowedContracts.Has(label) { // lets store impls for this contract
-					contract := strings.TrimPrefix(label, "contracts.hive.openshift.io/")
-					curr := supported[contract]
-					curr = append(curr, impls...)
-					supported[contract] = curr
+
+					contract = strings.TrimPrefix(label, "contracts.hive.openshift.io/")
+					if cfgJson := crd.Annotations[label+".config"]; cfgJson != "" {
+						// TODO: Check & log error! But don't return the error, as that would allow
```
suhanime: Any particular reason you didn't log the error here? If I'm reading this correctly, logger is configured in the file, and if we don't catch and report the error here, I'm guessing it would fail at the addConfigVolume stage? Or would it fail silently?

pkg/util/contracts/contracts.go
2uasimojo: Agree. That is one of the things I intend to do when I address the self-comment [above](https://github.com/openshift/hive/pull/2736/files/a77925d5a165182e0db7b4366c6ff1a085671112#r2305413405).

pkg/operator/hive/configmap.go
2uasimojo: > Any particular reason you didn't log the error here?

Just because I don't have a logger in this context. It would be trivial to pass one in from the caller, though, as you note.

> if we don't catch and report the error here, I'm guessing it would fail at the addConfigVolume stage? Or would it fail silently?

I think it'll just proceed as if the config was empty/unspecified. And with no evidence anywhere, that's going to be really hard to debug :(

I think it's worth fixing this immediately. Thanks for catching it.

/hold

2uasimojo: Fixed. Holding for test, but feel free to re-review.

suhanime (commented):
/lgtm

pkg/operator/hive/configmap.go
2uasimojo:
```
time="2025-09-05T15:14:51Z" level=error msg="failed to unmarshal contract config annotation; ignoring" configJSON="\"skipImageSetJob\": \"true\"}" configKey=contracts.hive.openshift.io/clusterinstall.config controller=hive error="invalid character ':' after top-level value"
```

✓

PR #2737, opened 2025-08-21
Support vSphere TagIDs

lukaspj:
I was not able to properly generate the CRDs as the version of yq used is not support on Mac/Darwin ARM64 unfortunately, but this seems like a minor risk-free change anyways.

PR #2739, opened 2025-08-25
HIVE-2914: Makefile: Use yq v4

2uasimojo:
The version of yq previously used by our make rules was v2. This apparently has no releases built for darwin/arm64 (reported via #2737).

Request v4 from build-machinery-go; and update the syntax of our calls accordingly.

This results in some whitespace-only changes to our CRDs, but should otherwise be seamless.

Note that there's a yq call in e2e-pool-test.sh that we aren't touching as it runs in the CI env, which is still using v2.

Assisted-By: Cursor

I used cursor to try a couple of different approaches, including vendoring v2 and building it locally, but "we" couldn't get that to work. Ultimately cursor helped with the v2=>v4 syntax conversion, though I ended up keeping almost none of its exact changes.


dlom:
> Note that there's a yq call in e2e-pool-test.sh that we aren't touching as it runs in the CI env, which is still using v2.

Does that mean we lose the ability to run the e2e tests locally? I personally have never actually tried it, but I know it was theoretically possible

2uasimojo:
> Does that mean we lose the ability to run the e2e tests locally? I personally have never actually tried it, but I know it was theoretically possible

No, `ensure-yq` installs to a very specific path and sets the `$(YQ)` Makefile var to point to that binary. If you were running e2e-pool locally, it would still use whatever `yq` you have in your `$PATH`.

2uasimojo:
Skipping retest on top of #2740.

PR #2742, opened 2025-08-28
HIVE-2920: Allow skipping creation of release-config.yaml

suhanime:


2uasimojo:
/lgtm

None of our CI uses bundle-gen.py, so I think we're safe to fast-merge this.

PR #2695, opened 2025-06-04
HIVE-2849: Migrate AWS SDK to v2

2uasimojo:

PR #2744, opened 2025-09-03
HIVE-2923: AWS started spelling it `RequestID`

2uasimojo:
Possibly as a result of the AWS SDK v2 transition (HIVE-2949 / #2695), or possibly just an arbitrary API change, AWS started spelling the embedded key for request IDs in error messages `RequestID` (it was previously some case variant of `Request ID` -- with a space). This befuddled our regex that was scrubbing such request IDs out of messages we include in status conditions, so those conditions were being updated every reconcile, so their CRs were being requeued immediately, causing thrashing.

Fix the regex to add an optional space.

Also DRY a couple of other spots that were using similar regex-based scrubbing. These may or may not have been causing problems *yet*, but they surely would have at some point. Also tech debt reduction is goodness.

### DIFF
--- a/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
+++ b/pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
@@ -2464,27 +2464,3 @@ func Test_toSupportedSubnets(t *testing.T) {
 		}},
 	}}, inv)
 }
-
-func Test_filterErrorMessage(t *testing.T) {
-	tests := []struct {
-		err  error
-		want string
-	}{{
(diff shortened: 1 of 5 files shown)


pkg/controller/awsprivatelink/awsprivatelink_controller_test.go
```diff
@@ -2464,27 +2464,3 @@ func Test_toSupportedSubnets(t *testing.T) {
 		}},
 	}}, inv)
 }
-
-func Test_filterErrorMessage(t *testing.T) {
```
2uasimojo, line 2468: ntr: these tests moved to errorscrub_test.go

pkg/controller/utils/errorscrub.go
```diff
@@ -8,7 +8,7 @@ var (
 	newlineTabRE           = regexp.MustCompile(`\n\t`)
 	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
 	// aws
-	awsRequestIDRE    = regexp.MustCompile(`(, )*(?i)(request id: )(?:[-[:xdigit:]]+)`)
+	awsRequestIDRE    = regexp.MustCompile(`(, )*(?i)(request ?id: )(?:[-[:xdigit:]]+)`)
```
2uasimojo, line 11: ntr: the real fix is here, the addition of this `?`.

PR #2745, opened 2025-09-03
[WIP] HIVE-2859: Add `ManifestsSecretRef` to `ClusterPool` for manifest management

eliorerz:
Add support for user-provided manifests in `ClusterPool` to apply custom manifests to all clusters created from the pool, similar to `ClusterDeployment`.
This is done as part of the Nutanix provider support for supporting `ClusterPool` resource .

This PR introduces support for user-provided manifests in `ClusterPool`. This allows a common set of custom manifests to be applied to all clusters claimed from the pool, mirroring the existing functionality in `ClusterDeployment`. 

**Note**: This is a prerequisite for adding `ClusterPool` support for the Nutanix provider.


2uasimojo:
At a glance, this looks sane as far as it goes, and I think it's something we want to do.

However...

It assumes that the manifests will want to be the same for all clusters in the pool. This isn't necessarily a bad thing, but it seems pretty likely that at some point we'll want them to be customizable per cluster. As such, they would want to be added to ClusterDeploymentCustomization so they can be part of the Inventory.

But perhaps I'm getting ahead of myself, and we're not going to need that any time soon...

WDYT?

eliorerz:
@2uasimojo I had similar thoughts about this.
I think it largely depends on the long-term direction we want Hive to take.

The main reasons I chose the current approach are:

1. In our case, setting this configuration per cluster adds unnecessary complexity, and we don’t yet know whether we’ll ever need it per cluster.

2. Even if, in the future, users do want to set it per cluster, we can always extend the functionality to support that. From my perspective, both approaches could coexist without impacting the general behavior.

2uasimojo:
Right, so on that note, we recently [added](https://github.com/openshift/hive/pull/2499) the ability to _patch_ installer manifests.

And we did it by adding a field to the CDC.

And we made it so a [single CDC can be applied to all pool clusters](https://github.com/openshift/hive/blob/8dfac7acf50063ab5233cc1d81fd05841b393ed6/vendor/github.com/openshift/hive/apis/hive/v1/clusterpool_types.go#L110-L111) as well as [per cluster via Inventory](https://github.com/openshift/hive/blob/8dfac7acf50063ab5233cc1d81fd05841b393ed6/vendor/github.com/openshift/hive/apis/hive/v1/clusterpool_types.go#L99).

So fitting into that architecture, it seems like it would make the most sense to make the new field part of CDC, and then you would inherently get both the pool-wide and per-cluster behaviors at the same time.

Mechanically for the consumer, it's a little more involved, because you have to create an extra CR -- the CDC -- but you only have to create one of them, not a whole inventory.

PR #2748, opened 2025-09-11
HIVE-2915: Add non-resource regex in errorScrub for AWS AccessDenied errors

suhanime:
We have a regex that we use to scrub AWS errors for AccessDenied or NotAuthorized operations. That regex looks for both the User ARN and the resource ARN that gives the AccessDenied.

However, when firing a Describe for resources (like a loadbalancer), we might not receive the resource ARN in the error message, so errorScrub skipped the scrubing. This can result in thrashing - especially in STS AssumeRole clusters because the User ARN is returned with a unique identifier every time.

This commit adds a regex similar to the one we already have, just not expecting a resource towards the later half of the string.

A unit test that simulates the actual error is also included to prove the regex works and does not interfere with the regex we already have.

### DIFF
--- a/pkg/controller/utils/errorscrub.go
+++ b/pkg/controller/utils/errorscrub.go
@@ -8,9 +8,10 @@ var (
 	newlineTabRE           = regexp.MustCompile(`\n\t`)
 	certificateTimeErrorRE = regexp.MustCompile(`: current time \S+ is after \S+`)
 	// aws
-	awsRequestIDRE    = regexp.MustCompile(`(, )*(?i)(request ?id: )(?:[-[:xdigit:]]+)`)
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/errorscrub.go
```diff
@@ -24,6 +25,7 @@ func ErrorScrub(err error) string {
 	s := newlineTabRE.ReplaceAllString(err.Error(), ", ")
 	s = awsRequestIDRE.ReplaceAllString(s, "")
 	s = awsNotAuthorized.ReplaceAllString(s, `${1}XXX${2}XXX${3}XXX${4}`)
+	s = awsNotAuthorizedNoResource.ReplaceAllString(s, `${1}XXX${2}XXX${3}`)
```
2uasimojo: The order of these two lines is important, as the latter is a substring of the former. It may be worth adding a comment to that effect.

suhanime:
@2uasimojo Addressed your comments, ready for review

/unhold

2uasimojo:
Manually merging, as we need to get this into OSD prod. This has already passed CI at least once, and the rebases should be wholly orthogonal.

Issue #1889, opened 2022-10-05
Ability to Override httpCustomErrorCodePages in ClusterDeployment


John2020-cyber:
I have a Custom ConfigMap for HTTP error pages which I want to use on Spokes the cluster deployment creates the following syncset for ingresscontroller. 
```
apiVersion: hive.openshift.io/v1
kind: SyncSet
metadata:
  name: stakater-devtest-clusteringress
  namespace: stakater-devtest
  ownerReferences:
    - apiVersion: hive.openshift.io/v1
      blockOwnerDeletion: true
      controller: true
      kind: ClusterDeployment
      name: stakater-devtest
      uid: 8572daee-a49c-4ebd-bc55-61e110b0a0b9
  labels:
    hive.openshift.io/cluster-deployment-name: stakater-devtest
    hive.openshift.io/syncset-type: remoteingress
spec:
  clusterDeploymentRefs:
    - name: stakater-devtest
  resourceApplyMode: Sync
  resources:
    - apiVersion: operator.openshift.io/v1
      kind: IngressController
      metadata:
        name: default
        namespace: openshift-ingress-operator
      spec:
        clientTLS:
          clientCA:
            name: ''
          clientCertificatePolicy: ''
        defaultCertificate:
          name: stakater-devtest-serving-cert-secret-generated
        domain: apps.devtest.*******.cloud
        httpCompression: {}
        httpErrorCodePages:
          name: ''
        tuningOptions: {}
      status:
        availableReplicas: 0
        domain: ''
        selector: ''
  secretMappings:
    - sourceRef:
        name: serving-cert-secret-generated
        namespace: stakater-devtest
      targetRef:
        name: stakater-devtest-serving-cert-secret-generated
        namespace: openshift-ingress
```
If I apply the following Patch to the Spoke directly the field  httpErrorCodePages.name is overwritten by the SyncSet owned by ClusterDeployment:
```
kind: SelectorSyncSet
apiVersion: hive.openshift.io/v1
metadata:
  name: default-ingress-controller-patch
spec:
  resourceApplyMode: Sync
  clusterDeploymentSelector:
    matchLabels:
      provider: stakater
      patchRouter: "true"
  patches:
    - kind: IngressController
      apiVersion: operator.openshift.io/v1
      name: default
      namespace: openshift-ingress-operator
      patch: |-
        spec:
          replicas: 3
          nodePlacement:
            nodeSelector:
              matchLabels:
                node-role.kubernetes.io/infra: ""
[...]

PR #2749, opened 2025-09-11
Remove status and prune spec fields with default values of created IngressController resources

grandeit:
The `IngressController` resources created from a `ClusterDeployment` and subsequently land in the `-clusteringress SyncSet` have some undesired default values, that when the `IngressController` object is applied on the target cluster, could potentially overwrite already present settings.
To address this issue and only apply values that are settable via `ClusterDeployment` it is desirable to strip all the unwanted keys/values (which are created because of non-ommitable types in the IngressController API definition).
I have implemented this by creating an allowlist of `spec keys` and stripping all other keys from the `IngressController` resource.

Before this the generated `SyncSet/IngressController` resources looked like this:
```
kind: SyncSet
metadata:
  name: foo-clusteringress
  namespace: foo
spec:
  clusterDeploymentRefs:
    - name: foo
  resourceApplyMode: Sync
  resources:
    - apiVersion: operator.openshift.io/v1
      kind: IngressController
      metadata:
        name: default
        namespace: openshift-ingress-operator
      spec:
        clientTLS:
          clientCA:
            name: ''
          clientCertificatePolicy: ''
        defaultCertificate:
          name: foo-ingress-secret
        domain: apps.foo.bar.baz.com
        httpCompression: {}
        httpErrorCodePages:
          name: ''
        tuningOptions:
          reloadInterval: 0s
      status:
        availableReplicas: 0
        domain: ''
        selector: ''
```

After this proposed fix:
```
kind: SyncSet
metadata:
  name: foo-clusteringress
  namespace: foo
spec:
  clusterDeploymentRefs:
    - name: foo
  resourceApplyMode: Sync
  resources:
    - apiVersion: operator.openshift.io/v1
      kind: IngressController
      metadata:
        name: default
        namespace: openshift-ingress-operator
      spec:
        defaultCertificate:
          name: foo-ingress-secret
        domain: apps.foo.bar.baz.com
```

### DIFF
--- a/pkg/controller/remoteingress/remoteingress_controller.go
+++ b/pkg/controller/remoteingress/remoteingress_controller.go
@@ -16,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -252,7 +253,34 @@ func rawExtensionsFromClusterDeployment(rContext *reconcileContext) []runtime.Ra
 
 	for _, ingress := range rContext.clusterDeployment.Spec.Ingress {
 		ingressObj := createIngressController(rContext.clusterDeployment, ingress)
-		raw := runtime.RawExtension{Object: ingressObj}
+
+		// Remove status and prune spec to a strict allowlist of user-settable fields derived from ClusterDeployment.
+		// Keep httpErrorCodePages only when explicitly set in the ClusterDeployment.
+		// This prevents overwriting changes made by other controllers/users on the remote cluster with (empty) defaults.
+		var obj runtime.Object = ingressObj
+		unstructuredObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(ingressObj)
+		if err == nil {
+			delete(unstructuredObj, "status")
+			if spec, ok := unstructuredObj["spec"].(map[string]any); ok {
+				allowed := map[string]bool{
+					"domain":             true,
+					"routeSelector":      true,
+					"namespaceSelector":  true,
+					"defaultCertificate": true,
+				}
+				if ingress.HttpErrorCodePages != nil {
+					allowed["httpErrorCodePages"] = true
+				}
+				for k := range spec {
+					if !allowed[k] {
+						delete(spec, k)
+					}
+				}
+			}
+			obj = &unstructured.Unstructured{Object: unstructuredObj}
+		}
+
+		raw := runtime.RawExtension{Object: obj}
 		rawList = append(rawList, raw)
 	}
 
--- a/pkg/controller/remoteingress/remoteingress_controller_test.go
+++ b/pkg/controller/remoteingress/remoteingress_controller_test.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
@@ -714,6 +715,7 @@ type createdResourceInfo struct {
 	routeSelector      *metav1.LabelSelector
 	defaultCertificate string
 	httpErrorCodePages *configv1.ConfigMapNameReference
+	raw                map[string]any
 }
 type createdSyncSetInfo struct {
 	name           string
@@ -747,8 +749,7 @@ func (f *fakeKubeCLI) ApplyRuntimeObject(obj runtime.Object, scheme *runtime.Sch
 			created.resources = append(created.resources, cr)
 			continue
 		}
-		ic, ok := raw.Object.(*ingresscontroller.IngressController)
-		if ok {
+		if ic, ok := raw.Object.(*ingresscontroller.IngressController); ok {
 			cr := createdResourceInfo{
 				name:              ic.Name,
 				namespace:         ic.Namespace,
@@ -762,6 +763,32 @@ func (f *fakeKubeCLI) ApplyRuntimeObject(obj runtime.Object, scheme *runtime.Sch
 			}
 			if ic.Spec.DefaultCertificate != nil {
 				cr.defaultCertificate = ic.Spec.DefaultCertificate.Name
+			}
+			if uic, err := runtime.DefaultUnstructuredConverter.ToUnstructured(ic); err == nil {
+				cr.raw = uic
+			}
+			created.resources = append(created.resources, cr)
+			continue
+		}
+		// Handle unstructured IngressController objects since we stripped some fields
+		if u, ok := raw.Object.(*unstructured.Unstructured); ok && u.GetKind() == "IngressController" {
+			cr := createdResourceInfo{
+				name:      u.GetName(),
+				namespace: u.GetNamespace(),
+				kind:      u.GetKind(),
+				raw:       u.Object,
+			}
+			var typed ingresscontroller.IngressController
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &typed); err == nil {
+				cr.domain = typed.Spec.Domain
+				cr.namespaceSelector = typed.Spec.NamespaceSelector
+				cr.routeSelector = typed.Spec.RouteSelector
+				if !reflect.DeepEqual(typed.Spec.HttpErrorCodePages, configv1.ConfigMapNameReference{}) {
+					cr.httpErrorCodePages = &typed.Spec.HttpErrorCodePages
+				}
+				if typed.Spec.DefaultCertificate != nil {
+					cr.defaultCertificate = typed.Spec.DefaultCertificate.Name
+				}
 			}
 			created.resources = append(created.resources, cr)
 			continue
@@ -804,6 +831,23 @@ func validateSyncSet(t *testing.T, existingSyncSet createdSyncSetInfo, expectedS
 				assert.Equal(t, ic.routeSelector, resObj.routeSelector, "unexpected routeSelector on ingressController: %v", ic.name)
 				assert.Equal(t, ic.defaultCertificate, resObj.defaultCertificate, "unexpected DefaultCertificate on ingressController: %v", ic.name)
 				assert.Equal(t, ic.httpErrorCodePages, resObj.httpErrorCodePages, "unexpected HttpErrorCodePages field on ingressController: %v", ic.name)
+
+				// Make sure that only allowed keys are present in the IngressController spec
+				if resObj.raw != nil {
+					allowed := map[string]bool{
+						"domain":             true,
+						"namespaceSelector":  true,
+						"routeSelector":      true,
+						"defaultCertificate": true,
+						"httpErrorCodePages": true,
+					}
+					if spec, ok := resObj.raw["spec"].(map[string]any); ok {
+						for k := range spec {
+							assert.Truef(t, allowed[k], "unexpected spec key %q on ingressController: %v", k, ic.name)
+						}
+						assert.Equal(t, ic.httpErrorCodePages != nil, spec["httpErrorCodePages"] != nil, "mismatched field httpErrorCodePages on ingressController: %v", ic.name)
+					}
+				}
 			}
 		}
 		assert.True(t, found, "didn't find expected ingressController: %v", ic.name)


2uasimojo:
/ok-to-test

Hi @grandeit! Thanks for your contribution!

I think it makes sense where you're going with this, but I want some time to think it through, as I suspect (hope) there may be a better way to avoid marshaling nil/unset fields into the raw object entirely. That way we wouldn't have to hardcode what's allowable and "manually" strip the rest.

In the meantime, for the sake of our internal tracking and testing, we're going to want to link this to a jira card in the [HIVE project](https://issues.redhat.com/projects/HIVE). You should be able to create one once you've set up an account (free). If you have trouble with this, let me know.

grandeit:
Hey @2uasimojo

> I think it makes sense where you're going with this, but I want some time to think it through, as I suspect (hope) there may be a better way to avoid marshaling nil/unset fields into the raw object entirely. That way we wouldn't have to hardcode what's allowable and "manually" strip the rest.

Maybe to explain it a bit better:
The issue is not really about unset/nil values since those are correctly omitted but with fields that do not have an _empty_ (nil, etc.) value.
For example, `HttpErrorCodePages` is not defined as a pointer in the `IngressController` type and therefore won't be omitted even though it is marked as `omitempty`: https://github.com/openshift/api/blob/c71c7b22632545c09a40f4fb3f211d8f4176d1f4/operator/v1/types_ingress.go#L83
Another example would be `ClientTLS` which is described as optional yet is also not a pointer and additionally not even marked with `omitempty`.
https://github.com/openshift/api/blob/c71c7b22632545c09a40f4fb3f211d8f4176d1f4/operator/v1/types_ingress.go#L190

Imho, the _best_ fix to get rid of these values would be to change the types in the `IngressController` type definition but I'm unsure about that approach since it would probably take a very long time to get this shipped. Therefore, I opted for the _faster_ solution and implement this fix.

I chose the whitelist approach to be resilient against the introduction of new _broken_ fields in the `IngressController` type.

> 
> In the meantime, for the sake of our internal tracking and testing, we're going to want to link this to a jira card in the [HIVE project](https://issues.redhat.com/projects/HIVE). You should be able to create one once you've set up an account (free). If you have trouble with this, let me know.

I already have an account linked to my corp email, I think. Will have a look.

2uasimojo:
That makes sense. You are correct that changing the IngressController schema would require an act of the gods, a deprecation cycle, and a couple of years worth of releases.

The other concern I have is that, even with the offending fields removed in the syncset, we'll be `oc apply`ing the resource to the spoke cluster, which I believe will cause the target object to be overwritten, removing those fields anyway. Did you test running your change all the way through with a real spoke cluster?

2uasimojo:
Perhaps we need to back up a little bit and identify what your actual end goal is. Generally speaking, we tend to design things like this such that the management of $object is *either* on the hub *or* the spoke. Any time we try to mix and match, we run into problems like this one. Simply put, if you want to manage your IngressController from the hub ClusterDeployment, do that, but expect changes made directly on the spoke to be overwritten; but if you want to manage it from the spoke, do not use CD.Spec.Ingress at all.

Now, if your issue is that CD.Spec.Ingress does not allow you to specify your desired values for some subfield of IngressController, let's solve that by changing the schema of CD.Spec.Ingress instead. Because hive is not an OCP payload component, we have a lot more flexibility to introduce schema changes. And depending where/how you're running hive, you would have a much shorter wait between code merge and release ($days if OperatorHub; max $months if MCE/ACM).

grandeit:
> The other concern I have is that, even with the offending fields removed in the syncset, we'll be `oc apply`ing the resource to the spoke cluster, which I believe will cause the target object to be overwritten, removing those fields anyway. Did you test running your change all the way through with a real spoke cluster?

Yes, I tested this on a real cluster. On the initial creation of the IngressController, the empty fields are still present (even when stripped) but on subsequent applies they are not overwritten. Meaning, you can safely set, for example `httpCompression`, after the `IngressController` was created initially on the spoke cluster by hive.

> Perhaps we need to back up a little bit and identify what your actual end goal is. Generally speaking, we tend to design things like this such that the management of $object is either on the hub or the spoke. Any time we try to mix and match, we run into problems like this one. Simply put, if you want to manage your IngressController from the hub ClusterDeployment, do that, but expect changes made directly on the spoke to be overwritten; but if you want to manage it from the spoke, do not use CD.Spec.Ingress at all.

I understand this view, but consider the following:
An OPs team uses `RHACM` to manage a fleet of OpenShift clusters. Some of them have been installed via assisted installer and the `ClusterDeployment` was used to set the ingress certificate. Some clusters where just imported into `RHACM` and therefore don't have an installer created `ClusterDeployment`. The team has a `Policy` that sets TLS settings on _all_ `IngressControllers` to the company standard and they want to use this `Policy` regardless of cluster installation method.
If hive would just care about fields it is responsible for -> no problem.
If hive claims the _whole_ `IngressController` object -> OPs team now has to distinguish hive managed objects from non-hive managed objects and figure out a way to set fields like `httpCompression` or `ClientTLS` in hive managed `IngressControllers` which is currently not possible.

> Now, if your issue is that CD.Spec.Ingress does not allow you to specify your desired values for some subfield of IngressController, let's solve that by changing the schema of CD.Spec.Ingress instead. Because hive is not an OCP payload component, we have a lot more flexibility to introduce schema changes. And depending where/how you're running hive, you would have a much shorter wait between code merge and release ($days if OperatorHub; max $months if MCE/ACM).

To make this compatible with the view you expressed in the other paragraph, that hive should control the whole object, `ClusterDeployment` needs to expose _all_ fields in the `IngressController` otherwise it would be unusable.

grandeit:
I just found this issue #1889 which goes in the same direction. If that user wants to patch `httpCompression` he will also have troubles. Btw. his solution also violates the one SyncSet should control the whole object premise.

2uasimojo:
> `Policy`

You're free to use the ACM `Policy` for your hive-managed spokes' `IngressController`z -- just omit the `Ingress` section from the ClusterDeployment.

> ClusterDeployment needs to expose all fields in the IngressController otherwise it would be unusable.

"Unusable" might be a tad hyperbolic, as the Ingress feature has existed in its current form (with the one addition you pointed out) since the beginning of time.

> If that user wants to patch `httpCompression` he will also have troubles.

And in that case he can propose to add that field to ClusterDeployment.

> Btw. his solution also violates the one SyncSet should control the whole object premise.

You mean where he was trying to patch it in with a different syncset? Yeah, we ended up not doing that because it led to thrashing :)

There are specific circumstances and configurations where we could have multiple syncsets managing a single object, but those synscets have to be very carefully designed. Similarly, there *may* be specific cases where we could have parts of an object managed by syncsets and parts left to the spoke's admin, but again, that would require careful consideration; and we would want to try as hard as possible to find some other way first. And these statements really apply to generic syncsets; when hive is doing specialized management of spoke objects, we really want to retain tighter control unless there's no other option.

...and in this case, we have at least two other options:
1. Use ACM's `Policy` rather than CD.Spec.Ingress
2. Add the _specific_ fields you care about to the CD.Spec.Ingress API

But I'm afraid I have to take a hard line on the idea of trying to facilitate managing IngressController from both sides.

grandeit:
> > `Policy`
> 
> You're free to use the ACM `Policy` for your hive-managed spokes' `IngressController`z -- just omit the `Ingress` section from the ClusterDeployment.

But this is exactly what's broken? Even if I ommit the values in the `Ingress` section of the `ClusterDeployment` I can't overwrite `httpCompression` via `Policy` in the spokes' `IngressController` resources because the `SyncSet` will overwrite it with an empty object?

> > ClusterDeployment needs to expose all fields in the IngressController otherwise it would be unusable.
> 
> "Unusable" might be a tad hyperbolic, as the Ingress feature has existed in its current form (with the one addition you pointed out) since the beginning of time.

I said it would be unusable under the premise that only one controller should be allowed to write to a resource. Because in that case you can't even set the replica count which makes it pretty much unusable in it's current state imho.

> > If that user wants to patch `httpCompression` he will also have troubles.
> 
> And in that case he can propose to add that field to ClusterDeployment.
> 
> > Btw. his solution also violates the one SyncSet should control the whole object premise.
> 
> You mean where he was trying to patch it in with a different syncset? Yeah, we ended up not doing that because it led to thrashing :)

Only leads to thrashing because the generated `SyncSet` from the `ClusterDeployment` is writing values it has no business writing.

> 
> There are specific circumstances and configurations where we could have multiple syncsets managing a single object, but those synscets have to be very carefully designed. Similarly, there _may_ be specific cases where we could have parts of an object managed by syncsets and parts left to the spoke's admin, but again, that would require careful consideration; and we would want to try as hard as possible to find some other way first. And these statements really apply to generic syncsets; when hive is doing specialized management of spoke objects, we really want to retain tighter control unless there's no other option.
> 
> ...and in this case, we have at least two other options:
> 
> 1. Use ACM's `Policy` rather than CD.Spec.Ingress

CD.Spec.Ingress is a convenient way to set the ingress certificate at install-time when using the assisted installer.
Alternative is using `AgentClusterInstall`'s `spec.manifestsConfigMapRefs` to get the ingress certificate to the target cluster, I guess. This is probably what we will have to do since using CD.Spec.Ingress in the current state, i.e. overwriting values it shouldn't be, is just not feasible.

`Policy` is only available after installation.

> 2. Add the _specific_ fields you care about to the CD.Spec.Ingress API

Why even have this abstraction with the specific fields? Just put the whole `IngressController` object in the `ClusterDeployment`. This would also solve the "problem".

> 
> But I'm afraid I have to take a hard line on the idea of trying to facilitate managing IngressController from both sides.

Don't really understand the reasoning behind that stance since even Kubernetes implemented features like server-side apply to allow managing parts of a resource from different controllers.

2uasimojo:
> But this is exactly what's broken? Even if I ommit the values in the `Ingress` section of the `ClusterDeployment` I can't overwrite `httpCompression` via `Policy` in the spokes' `IngressController` resources because the `SyncSet` will overwrite it with an empty object?

Sorry, what I meant was, omit the *entire* CD.Spec.Ingress section. IOW don't ask hive to manage your ingress at all: manage it completely via some other tool, like ACM `Policy`, or straight gitops on the spoke, or even your own custom syncset that uses `patches` so you can be assured it's only touching the fields you want it to touch.

> Only leads to thrashing because the generated SyncSet from the ClusterDeployment is writing values it has no business writing.

Or from another perspective, because some other entity was trying to write to something it had no business trying to write because it was already managed exclusively by hive, per that same user's requested configuration :innocent: 

> CD.Spec.Ingress is a convenient way to set the ingress certificate at install-time when using the assisted installer.

I doubt Assisted Installer itself is paying attention to CD.Spec.Ingress, but I'll ask.

I think the convenience factor is more about
- the ability to define multiple IngressControllers via one CR -- analogous to how one MachinePool translates to multiple MachineSets
- automatic copying of cert secrets

> `Policy` is only available after installation.

Ditto CD.Spec.Ingress. I can promise you that hive is not applying those IngressControllers until after installation because we [ignore](https://github.com/openshift/hive/blob/e0ba5975f260a347edb4d14d38e9b71a9066f9d6/pkg/controller/clustersync/clustersync_controller.go#L316-L319) syncsets before that.

If you want day 0 control over IngressController beyond whatever is available via install-config.yaml (is that even a thing?) you could make use of additional manifests as you mentioned (though I wasn't aware Assisted had an analog to hive's Manifests{ConfigMap|Secret}Ref)

OR (new) [manifest patches](https://github.com/openshift/hive/blob/master/docs/enhancements/patch-install-manifests.md) -- but this only applies when using hive's provisioning. If you're really using Assisted Installer, those won't have any effect.

> Don't really understand the reasoning behind that stance since even Kubernetes implemented features like server-side apply to allow managing parts of a resource from different controllers.

Yup, it has been pointed out lately that hive hasn't incorporated SSA for syncsets. To be honest, we're not in a huge hurry to do that. And... would it even help in this case?

Hive is an abstraction on top of k8s (really on top of OCP). We're allowed to maintain a simplified architecture by making choices like "objects must be managed by the hub XOR the spoke".

2uasimojo:
BTW, regarding:

> Some clusters where just imported into `RHACM` and therefore don't have an installer created `ClusterDeployment`.

Hive has the concept of [adoption](https://github.com/openshift/hive/blob/master/docs/using-hive.md#cluster-adoption), which is distinct from RHACM "import". Essentially, you create a CD with particular fields filled in, give it a kubeconfig Secret, and hive will assume control over the day 2 operations it supports, as/if configured, including syncsets. And (I haven't tested this, but I _think_) that CD would be allowed to contain an `Ingress` section, which would cause hive to create the IngressController syncsets the same way it would for your other non-imported clusters.

I recognize this is the opposite of the direction we seem to be heading, but if you do decide you would rather beef up the Ingress schema with the fields you care about, this would provide you with a way to manage your IngressControllers uniformly across your whole fleet from hive rather than via `Policy`.

grandeit:
> Or from another perspective, because some other entity was trying to write to something it had no business trying to write because it was already managed exclusively by hive, per that same user's requested configuration 😇

One could argue that it was not per the user's requested configuration since the user never requested `httpErrorCodePages` to be overwritten with an empty value in the first place nor could he have changed it even if he wanted to.

On the topic of exclusively managing things by SyncSets: Isn't it rather inconsistent that the SyncSet generated for API certificates (one can set via `CD.spec.controlPlaneConfig.serviceCertificates`) uses patches and clearly doesn't claim ownership over the `APIServer` and `KubeAPIServer` resources?

2uasimojo:
> One could argue that it was not per the user's requested configuration since the user never requested `httpErrorCodePages` to be overwritten with an empty value in the first place nor could he have changed it even if he wanted to.

Yup, the all-or-nothing model is an acknowledged tradeoff.

Here's another way to think of it: In your case you (seem to) have one team responsible for both sides of the equation, at least for IngressController, so it would be easy for you to understand which fields are "owned" by the hub and which by the spoke. But hive frequently gets used for a model where the hub and spoke admins are on different teams, different departments, even different companies on different sides of the planet. In those cases it would be harder to make the spoke admin understand which specific fields they can touch and which they can't, than to tell them, "Oh, yeah, `IngressController`? You can't touch that, that's controlled from outside."

It's worth noting that the hypershift (hosted control planes) architecture came along ~5y after the inception of hive, and they had to think about a lot of these same issues, and they had the ability to take some lessons away from hive's experiences. In their case, I believe the divide is even stricter, because the spoke admin literally *can't* access the control plane pieces.

> On the topic of exclusively managing things by SyncSets: Isn't it rather inconsistent that the SyncSet generated for API certificates (one can set via `CD.spec.controlPlaneConfig.serviceCertificates`) uses patches and clearly doesn't claim ownership over the `APIServer` and `KubeAPIServer` resources?

If we had called the field `APIServer` perhaps.

> I doubt Assisted Installer itself is paying attention to CD.Spec.Ingress, but I'll ask.

I heard back from the Assisted maven who confirmed that their code is not referencing CD.Spec.Ingress.

----

So where are we on this? Would you like to move forward with a change that adds the specific controls you need from CD.Spec.Ingress, or will you move to exclusively using some other method like `Policy` or gitops?

grandeit:
> But hive frequently gets used for a model where the hub and spoke admins are on different teams, different departments, even different companies on different sides of the planet. In those cases it would be harder to make the spoke admin understand which specific fields they can touch and which they can't, than to tell them, "Oh, yeah, `IngressController`? You can't touch that, that's controlled from outside."

But how would the patch I propose make anything worse in this scenario?
The people can still pretend that the `IngressController` is _owned_ by hive, if they like to do that (which is what they have to do anyway because there is no enforcement, really) and people that are more "advanced" can just overwrite all the fields they want to overwrite. Both happy. :)

>So where are we on this? Would you like to move forward with a change that adds the specific controls you need from CD.Spec.Ingress, or will you move to exclusively using some other method like Policy or gitops?

If you don't like this patch, we will have to find some other solution outside of hive, I guess. Implementing extra fields in `CD.Spec.Ingress` seems like a waste of time imho. _If_ doing anything to `CD.Spec.Ingress`, it should be making it a list of `IngressController`, so you don't have to implement "passthrough" fields that serve no real purpose and will just bloat your test cases.

2uasimojo:
> But how would the patch I propose make anything worse in this scenario?
> ...
> just bloat your test cases.

Yeah, this. Either approach increases the test surface and maintenance burden, *but* we have long and deep precedent of adding API fields to expose knobs on the hub. That's in our wheelhouse in terms of engineering, testing, and maintenance. Whereas every attempt we've made to facilitate dual management of an object (usually via syncsets) has been a brittle snowflake, and has not resulted in a better UX for whoever was trying to do it.

> _If_ doing anything to `CD.Spec.Ingress`, it should be making it a list of `IngressController`

I'm not opposed to this idea philosophically, but it would be more work than value, a situation of, "perfect is the enemy of good." We're not going to rearchitect this 5+yo API that, for example, managed OpenShift uses for thousands of clusters. But if you want to improve it incrementally, we're still open to that.

2uasimojo:
For those following along at home, we have taken in https://issues.redhat.com/browse/HIVE-2952 and https://issues.redhat.com/browse/HIVE-2953 to remove the status section and add an explicit knob for `reloadInterval`, respectively.

Also I just found out that this

> managed OpenShift uses for thousands of clusters

isn't actually true. While those clusters do have `cd.spec.ingress` sections, the `remoteingress` controller is disabled in that env. I didn't dig further, but if ingress is being managed there, it's via some other mechanism.

That's not to say other hive consumers (e.g. in ACM land) aren't using it.

PR #2751, opened 2025-09-14
NO-JIRA: Add .ai_context directory for AI-assisted test generation

huangmingxia:
This PR adds a new directory `.ai_testgen` to the Hive repository to support AI-assisted test generation.

### DIFF
--- a/.ai_testgen/README.md
+++ b/.ai_testgen/README.md
@@ -1,0 +1,25 @@
+# AI-Assisted Test Generation
+
+This proposal aims to integrate an AI-assisted test generation system into the Hive upstream repository.  
+To achieve this, we suggest creating a dedicated directory `.ai_testgen` within the repository to store prompt files and related resources.  
+This ensures that contributors who already have the Hive repository can directly leverage these assets without the need to clone or maintain a separate repository.  
+
+## Two possible implementation options are described below:  
+
+**1. Clone external repository**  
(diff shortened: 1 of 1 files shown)


.ai_testgen/workflow_outputs/cco/OCPBUGS-60078/phases/test_requirements_output.yaml
2uasimojo, line 1: Looks like we have some CCO cross-pollution happening.

huangmingxia: Thank you for your review. I will remove the CCO test cases later. I just it might be helpful to show the CCO output now, so I included it.
By the way, I will continue to make adjustments to the agent files, as there are still many detailed issues to address :)

.ai_testgen/README.md
2uasimojo, line 1: This has the feel of something copied from a gdoc. IIUC this PR is executing on the things suggested here, so it feels kind of funny. Also it would be nice to avoid referring to specific users' forks in the main repo.

huangmingxia: Thanks for your suggestion! This file should be removed and replaced with USAGE.md. 
The README.md was added at the time while I was considering which approach to use, but in the end I felt the first approach was more appropriate.

PR #2752, opened 2025-09-15
HIVE-2936: HiveSLO - update env_name query

suhanime:
On HiveSLO grafana dashboard, a env_name variable is defined to fetch deployments for each shard. The query for the same is looking for specific values that can go out-of-sync when a new shard is created. This commit switches the query to use a regex that can filter out all the relevant values, thus making it a more dynamic, as long as the shards are named following the convention. The only shard that doesn't follow the naming convention is osd-stage-01, which is included explicitly.

Note that this variable isn't actually used for the dashboard as of now because the query is disabled. Regardless this change would ensure the query is same as that of Hive Metrics dashboard where the same variable is used as "Deployment Environment"

/assign @2uasimojo

### DIFF
--- a/hack/grafana-dashboards/grafana-dashboard-hive-slo.configmap.yaml
+++ b/hack/grafana-dashboards/grafana-dashboard-hive-slo.configmap.yaml
@@ -28,6 +28,7 @@ data:
               "uid": "P5108662370F9C14C"
             },
             "enable": false,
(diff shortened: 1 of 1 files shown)


2uasimojo (commented):
/lgtm

Maybe we can add this (directory?) to the exclusion list for running CI.

PR #2753, opened 2025-09-16
chore(KONFLUX-6210): fix and set name and cpe label for hive-mce-29

ralphbean:
For https://issues.redhat.com/browse/KONFLUX-6210, clair needs access to a name and cpe label that it can use to look up the image in VEX statements.

See also release-engineering/rhtap-ec-policy#149

Signed-off-by: Ralph Bean <rbean@redhat.com>
Assisted-by: Gemini

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,10 +77,11 @@ RUN if [ -e "/activation-key/org" ]; then ln -s /etc/rhsm-host /run/secrets/rhsm
 # TODO: should this be the operator?
 ENTRYPOINT ["/opt/services/manager"]
 
-LABEL name="hive"
+LABEL name="multicluster-engine/hive-rhel9"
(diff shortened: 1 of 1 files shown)


2uasimojo:
/assign @dlom 

I don't love having MCE-specific things in our otherwise-generic Dockerfile. Is there some other way to get that injected?

You can override `security` when ready to merge, Mark. This is a new, unrelated failure tracked by [HIVE-2937](https://issues.redhat.com/browse/HIVE-2937).

dlom:
@ralphbean We use the same Dockerfile to build all MCE images for all version 2.6-2.10.  Hardcoding an MCE 2.9 string in the Dockerfile isn't a viable solution IMO.  I think we can pass in these labels as environment variables, and configure the version one level up (in the konflux pipeline definitions)

suhanime:
Superseded by #2862

PR #2757, opened 2025-09-25
HIVE-2891: Copy secret for CD as early as possible

dlom:
/assign @2uasimojo 

When an existing cluster is deprovisioned, the CD controller jumps into a deprovision-specific code path before the secret has a chance to be copied.  We should do that ASAP instead

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -618,6 +618,17 @@ func (r *ReconcileClusterDeployment) reconcile(request reconcile.Request, cd *hi
 		return reconcile.Result{}, r.Update(context.TODO(), cd)
 	}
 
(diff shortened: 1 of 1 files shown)


2uasimojo:
I don't agree with "as early as possible". As a particular example, we shouldn't be taking any reasonably-avoidable action when the `reconcile-pause` annotation is [in play](https://github.com/openshift/hive/pull/2757/files#diff-5fc238b39e2a1fc17bd412fbac2d14eab4e0f205c91886b5671611bec4c665ccR391).

To resolve the problem that prompted this, we just need to make sure the copy happens before we need to stand up an imageset, provision, or deprovision pod. Moving the code block to right [above](https://github.com/openshift/hive/pull/2757/files#diff-5fc238b39e2a1fc17bd412fbac2d14eab4e0f205c91886b5671611bec4c665ccR632) where we enter `syncDeletedClusterDeployment()` should be good.

2uasimojo (commented):
/lgtm

PR #2877, opened 2026-03-23
HIVE-3127: Add benchmark harness for resource/remoteclient optimization

dlom:
xref: [HIVE-3127](https://redhat.atlassian.net/browse/HIVE-3127)

The clustersync controller and others that depend on pkg/resource (Helper) and pkg/remoteclient have been repeatedly flagged by downstream users for:
- Slow reconciliation times
- Memory leaks during large SyncSet applications
- Excessive API server load (unnecessary round trips and discovery calls)

These packages are legacy code with significant behavioral dependencies from downstream consumers. We cannot rewrite them, but targeted optimizations (50-100 LOC changes) in specific hot paths could significantly improve performance without altering behavior.

Each benchmark runs against a real kube-apiserver (via envtest) and reports custom metrics for round-trip count and byte transfer (in addition to standard allocation and wall-time measurements)

Some preliminary findings:
- CreateOrUpdate SteadyState allocates ~79MB/op for a no-op: 88x more than Apply for the same 2 roundtrips. Something isn't being cached properly.
- NewHelper construction costs 78MB and 205K allocs: the dominant cost in most benchmarks. More missed caching opportunities
- First Apply per package pays 2 extra discovery roundtrips (5 vs 3) due to disk-cache cold start. Caching...
- Hibernation takes 1 full second despite only 13 roundtrips and 1.3MB allocs. This is Wall-clock bound, likely connection/TLS overhead in the BuildKubeClient path.

Assisted-by: Claude Opus 4.6 (LLM wrote the original suite of benchmarks automatically, I have refined them extensively to meet my quality standards)

/assign @2uasimojo 

## Release Notes

* **New Features**
  * Added comprehensive benchmarking infrastructure for performance testing against real Kubernetes API servers
  * Introduced benchmark comparison tooling to measure performance changes against base commits
  * Added new make target and setup scripts for Kubernetes test environment initialization

* **Refactor**
  * Remote client builder now accepts optional configuration parameters for customization

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -384,7 +384,14 @@ install-tools:
 	go install $(GO_MOD_FLAGS) github.com/golang/mock/mockgen
 	go install $(GO_MOD_FLAGS) golang.org/x/lint/golint
 	go install $(GO_MOD_FLAGS) github.com/golangci/golangci-lint/cmd/golangci-lint
+	go install $(GO_MOD_FLAGS) sigs.k8s.io/controller-runtime/tools/setup-envtest
 
 .PHONY: coverage
 coverage:
 	hack/codecov.sh
+
+.PHONY: setup-envtest
+# Downloads envtest binaries and prints the required export KUBEBUILDER_ASSETS=... command.
+# Run the printed export command (or add it to your shell rc) before running benchmarks.
+setup-envtest:
+	hack/setup-envtest.sh
--- a/go.mod
+++ b/go.mod
@@ -74,6 +74,7 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
 	sigs.k8s.io/cluster-api-provider-azure v1.21.1-0.20250929163617-2c4eaa611a39
 	sigs.k8s.io/controller-runtime v0.22.3
+	sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240927101401-4381fa0aeee4
 	sigs.k8s.io/controller-tools v0.19.0
 	sigs.k8s.io/yaml v1.6.0
 )
--- a/hack/benchmark-comparison.sh
+++ b/hack/benchmark-comparison.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+# Compare benchmark performance before and after working changes.
+#
+# Uses git worktrees for safe comparison without touching your working tree.
+# Compares the current working tree against a base commit (default: HEAD~1).
+#
+# Usage:
+#   ./hack/benchmark-comparison.sh [base-commit]
+#
+# Examples:
+#   # Compare current working tree vs parent commit
+#   ./hack/benchmark-comparison.sh
+#
+#   # Compare current working tree vs main branch
+#   ./hack/benchmark-comparison.sh main
+#
+#   # Compare with custom benchmark settings
+#   BENCHTIME=10x COUNT=8 ./hack/benchmark-comparison.sh
+#
+#   # Benchmark specific packages
+#   BENCH_PKGS="./pkg/resource/" ./hack/benchmark-comparison.sh
+
+set -eo pipefail
+
+# Configurable benchmark parameters
+BENCHTIME=${BENCHTIME:-5x}
+COUNT=${COUNT:-6}
+BENCH_PATTERN=${BENCH_PATTERN:-'^Benchmark'}
+BENCH_PKGS=${BENCH_PKGS:-./test/benchmark/}
+BASE_COMMIT=${1:-HEAD~1}
+
+# Ensure we're in a git repository
+if ! git rev-parse --git-dir > /dev/null 2>&1; then
+    echo "Error: Not in a git repository"
+    exit 1
+fi
+
+# Ensure KUBEBUILDER_ASSETS is set
+if [ -z "$KUBEBUILDER_ASSETS" ]; then
+    echo "Error: KUBEBUILDER_ASSETS environment variable not set"
+    echo ""
+    echo "Run: make setup-envtest"
+    echo "Then: export KUBEBUILDER_ASSETS=\"\$(hack/setup-envtest.sh | tail -1 | cut -d'=' -f2 | tr -d '\"')\""
+    exit 1
+fi
+
+# Verify base commit exists
+if ! git rev-parse "$BASE_COMMIT" > /dev/null 2>&1; then
+    echo "Error: Base commit '$BASE_COMMIT' not found"
+    exit 1
+fi
+
+TEMP_DIR=$(mktemp -d)
+WORKTREE_DIR=$(mktemp -d -t hive-benchmark-worktree.XXXXXX)
+
+cleanup() {
+    if [ -d "$WORKTREE_DIR" ]; then
+        echo ""
+        echo "Cleaning up worktree..."
+        git worktree remove "$WORKTREE_DIR" --force 2>/dev/null || true
+        rm -rf "$WORKTREE_DIR" 2>/dev/null || true
+    fi
+}
+trap cleanup EXIT
+
+echo "==========================================================="
+echo "Benchmark Comparison"
+echo "==========================================================="
+echo "Base commit:     $BASE_COMMIT ($(git rev-parse --short "$BASE_COMMIT"))"
+echo "Working tree:    $(git rev-parse --short HEAD)$(git diff --quiet && git diff --cached --quiet || echo ' (dirty)')"
+echo "Benchmark time:  ${BENCHTIME}"
+echo "Run count:       ${COUNT}"
+echo "Pattern:         ${BENCH_PATTERN}"
+echo "Packages:        ${BENCH_PKGS}"
+echo "Results dir:     ${TEMP_DIR}"
+echo "==========================================================="
+echo ""
+
+# Run benchmarks in the current working tree
+echo "-> Running benchmarks in current working tree..."
+go test -bench="${BENCH_PATTERN}" -benchmem -benchtime="${BENCHTIME}" -count="${COUNT}" ${BENCH_PKGS} 2>&1 | tee "${TEMP_DIR}/new.txt"
+
+# Create worktree at base commit
+echo ""
+echo "-> Creating worktree at ${BASE_COMMIT}..."
+git worktree add --detach "$WORKTREE_DIR" "$BASE_COMMIT" --quiet
+
+# Run benchmarks at base commit (in worktree)
+echo "-> Running benchmarks at base commit..."
+(cd "$WORKTREE_DIR" && \
+    export KUBEBUILDER_ASSETS="$KUBEBUILDER_ASSETS" && \
+    go test -bench="${BENCH_PATTERN}" -benchmem -benchtime="${BENCHTIME}" -count="${COUNT}" ${BENCH_PKGS}) 2>&1 | tee "${TEMP_DIR}/old.txt"
+
+# Cleanup worktree early
+echo ""
+git worktree remove "$WORKTREE_DIR" --force
+rm -rf "$WORKTREE_DIR"
+
+# Compare results
+echo ""
+echo "==========================================================="
+echo "BENCHMARK COMPARISON"
+echo "==========================================================="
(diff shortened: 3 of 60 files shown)


pkg/remoteclient/remoteclient.go
```diff
@@ -185,11 +190,29 @@ func SetUnreachableCondition(cd *hivev1.ClusterDeployment, connectionError error
 	return
 }
 
+// BuilderOption configures optional behavior on a Builder.
+type BuilderOption func(*builderOptions)
```
dlom, line 194: NTR: This new `BuilderOption` (and associated machinery) is the only new code in production. I tried to make this as non-invasive as possible: all the callsites that construct new Builders are unchanged and this code no-ops in prod

hack/benchmark-comparison.sh
```diff
@@ -0,0 +1,127 @@
...
+fi
+
+# Ensure KUBEBUILDER_ASSETS is set
+if [ -z "$KUBEBUILDER_ASSETS" ]; then
+    echo "Error: KUBEBUILDER_ASSETS environment variable not set"
+    echo ""
+    echo "Run: make setup-envtest"
+    echo "Then: export KUBEBUILDER_ASSETS=\"\$(hack/setup-envtest.sh | tail -1 | cut -d'=' -f2 | tr -d '\"')\""
+    exit 1
+fi
```
2uasimojo, line 45: What would be really cool is if the benchmark could optionally be run against an actual cluster. I've been trying to track down where/how KUBEBUILDER_ASSETS is actually consumed to spoof the cluster via envtest, but have so far been unsuccessful.

pkg/remoteclient/kubeconfig.go
```diff
@@ -89,5 +95,12 @@ func (b *kubeconfigBuilder) UseSecondaryAPIURL() Builder {
 }
 
 func (b *kubeconfigBuilder) RESTConfig() (*rest.Config, error) {
-	return utils.RestConfigFromSecret(b.secret, false)
+	cfg, err := utils.RestConfigFromSecret(b.secret, false)
+	if err != nil {
+		return nil, err
+	}
+	if b.opts.transportWrapper != nil {
```
2uasimojo, line 102: This seems like it should be DRYable if, instead of func(*builderOptions), our BuilderOptions (possibly renamed to RESTConfigOptions) was func(*whatever_type_cfg_is). Unless that type is different in the different calls?

pkg/resource/helper_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	b.Run("ByType", func(b *testing.B) {
+		resources := []struct {
+			name string
+			obj  client.Object
+		}{
+			{"ConfigMap", benchutil.GenerateConfigMap("apply-cm", 0)},
+			{"Secret", benchutil.GenerateSecret("apply-secret", 0)},
+			{"Deployment", benchutil.GenerateDeployment("apply-deploy")},
+			{"ServiceAccount", benchutil.GenerateServiceAccount("apply-sa")},
```
2uasimojo, line 54: I wonder if it would be instructive to include some OpenShift-specific types?

pkg/resource/helper_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+// BenchmarkApplyRuntimeObject benchmarks ApplyRuntimeObject.
+func BenchmarkApplyRuntimeObject(b *testing.B) {
+	objects := []struct {
+		name string
+		obj  client.Object
+	}{
+		{"ConfigMap", benchutil.GenerateConfigMap("rt-cm", 0)},
+		{"Secret", benchutil.GenerateSecret("rt-secret", 0)},
+		{"ServiceAccount", benchutil.GenerateServiceAccount("rt-sa")},
```
2uasimojo, line 124: Curious why the list of resources isn't the same across the board.

In fact, seems like we could DRY the list via a wrapper that accepts the string prefix (here `rt`).

pkg/resource/helper_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	b.Run("BySize", func(b *testing.B) {
+		sizes := []struct {
+			name string
+			size int
+		}{
+			{"100B", 100},
+			{"1KB", 1024},
+			{"10KB", 10 * 1024},
+			{"100KB", 100 * 1024},
```
2uasimojo, line 172: similar DRY comment for sizes

test/benchmark/controller_sim_benchmark_test.go
```diff
@@ -0,0 +1,459 @@
...
+// BenchmarkControllerReconcileClusterSyncCreateOrUpdate is an A/B comparison with Apply.
+func BenchmarkControllerReconcileClusterSyncCreateOrUpdate(b *testing.B) {
+	b.Run("Small", func(b *testing.B) {
+		benchutil.ControllerHarness[*benchutil.BenchRemoteClient]{
+			NewObjects: func(_ *testing.B, ns string) []client.Object {
+				return []client.Object{
+					benchutil.CopyAndSetNamespace(benchutil.GenerateConfigMap("cou-cm", 0), ns),
+					benchutil.CopyAndSetNamespace(benchutil.GenerateSecret("cou-secret", 0), ns),
+					benchutil.CopyAndSetNamespace(benchutil.GenerateDeployment("cou-deploy"), ns),
+					benchutil.CopyAndSetNamespace(benchutil.GenerateServiceAccount("cou-sa"), ns),
```
2uasimojo, line 218: The aforementioned DRYing would manifest here as CopyAndSetNamespace accepting variadic args and returning a slice.

test/benchmark/README.md
```diff
@@ -0,0 +1,121 @@
+# Hive Resource/RemoteClient Benchmarks
+
+Performance benchmarks for validating safe optimizations to Hive's legacy
```
2uasimojo, line 3: The word 'safe' here seems like an overreach. IIUC this suite isn't doing anything to validate the safety of changes; it's just going to tell you how they perform against the original?

test/benchmark/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+The clustersync controller and others that depend on these packages have been
+repeatedly flagged by downstream users for slow reconciliation, memory leaks,
+and excessive API server load. These packages are legacy code with behavioral
+dependencies that prevent rewrites.
+
+These benchmarks enable targeted optimizations by:
+1. Identifying hot paths through round-trip and allocation measurement
+2. Proving performance improvements via A/B comparison
+3. Detecting behavioral regressions - unexpected metric changes indicate logic changes
```
2uasimojo, line 16: I guess this?

Makefile
```diff
@@ -384,7 +384,14 @@ install-tools:
 	go install $(GO_MOD_FLAGS) github.com/golang/mock/mockgen
 	go install $(GO_MOD_FLAGS) golang.org/x/lint/golint
 	go install $(GO_MOD_FLAGS) github.com/golangci/golangci-lint/cmd/golangci-lint
+	go install $(GO_MOD_FLAGS) sigs.k8s.io/controller-runtime/tools/setup-envtest
```
2uasimojo, line 387: benchstat?

test/benchmark/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Quick Start
+
+```bash
+# One-time setup
+make install-tools
+make setup-envtest
+export KUBEBUILDER_ASSETS="$(hack/setup-envtest.sh | tail -1 | cut -d'=' -f2 | tr -d '"')"
```
2uasimojo, line 24: This pattern is awkward. Only slightly better would be having `make setup-envtest` output the appropriate env var assignment... though that would need to be copy-pasted and is not as scriptable :(

test/benchmark/controller_sim_benchmark_test.go
2uasimojo, line 1: I can't really tell how closely this simulates a given controller. For example, do we have (tunable) configurations for the clustersync controller that exercise multiple resources per syncset as well as multiple syncsets? Those would perform differently if there's per-syncset setup that doesn't happen per-resource -- and that's the kind of insight that would be super informative to the consumer of this suite.

What would be super cool is if, instead of attempting to duplicate the controller logic, these could actually call the real controller's Reconcile() (in a loop: until it returns success; or errors a certain number of times; or after a certain amount of wallclock time) with the instrumented client. Obvs this would require a closer-to-real APIServer -- IIUC envtest doesn't have "real" object storage like fakeclient. And probably other really difficult things. But surely claude can do it!

test/benchmark/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Each benchmark reports:
+- Wall time - Total operation duration
+- Allocations - Memory allocations per operation (MB/op, allocs/op)
+- Round trips - HTTP calls to API server (custom metric)
+- Bytes transferred - Network traffic (custom metric)
+
+Key insight: Round-trip count changes indicate behavioral changes.
+If your optimization changes round-trip count, it's not just a performance
+improvement -- it's a logic change that needs careful review.
```
2uasimojo, line 44: Perhaps the comparison script could flag these visually somehow.

test/benchmark/README.md
```diff
@@ -0,0 +1,121 @@
...
+./hack/benchmark-comparison.sh main
+
+# Or specify custom settings
+BENCHTIME=10x COUNT=8 ./hack/benchmark-comparison.sh
+```
+
+## Architecture
+
+Benchmarks run against a real kube-apiserver + etcd (via controller-runtime's envtest),
+not mocks, giving accurate round-trip and byte measurements. Each benchmark uses
```
2uasimojo, line 70: Yeah, no. envtest is not a "real" KAS. I'm still not clear whether it has object storage, but at a minimum, it's *local*, so the round-trip times are certainly not "accurate". I suppose it might be set up with sleeps to simulate network latency/throughput? Anyway, suggest softening this language a bit.

As mentioned earlier, a goal would be the ability to run the suite against an actual cluster, so that we *could* get accurate RT measurements.

(Later) Oh, perhaps you meant a *count* of the *number* of round trips?

test/benchutil/benchenv.go
```diff
@@ -0,0 +1,97 @@
...
+	CD *hivev1.ClusterDeployment
+
+	// Namespace is a unique namespace for benchmark isolation.
+	Namespace string
+}
+
+// NewBenchEnv creates a fully initialized benchmark environment.
+func NewBenchEnv(b *testing.B) *BenchEnv {
+	b.Helper()
+	cfg := EnsureEnvTest(b)
```
2uasimojo, line 39: Yeah, yeah, so here, couldn't we optionally load up a real KUBECONFIG instead? And the rest would Just Work™?

test/benchutil/envtest.go
```diff
@@ -0,0 +1,155 @@
...
+		os.Exit(1)
+	}
+	root := strings.TrimSpace(string(out))
+	crdDir := filepath.Join(root, "config", "crds")
+	if _, err := os.Stat(crdDir); err != nil {
+		fmt.Fprintf(os.Stderr, "CRD directory not found at %s: %v\n", crdDir, err)
+		os.Exit(1)
+	}
+	return []string{crdDir}
+}
```
2uasimojo, line 108: This seems unnecessarily obfuscated. It's not really adding anything to just having the list of dirs hardcoded in a const at the top.

go.mod
```diff
@@ -74,6 +74,7 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
 	sigs.k8s.io/cluster-api-provider-azure v1.21.1-0.20250929163617-2c4eaa611a39
 	sigs.k8s.io/controller-runtime v0.22.3
+	sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240927101401-4381fa0aeee4
```
2uasimojo, line 77: Odd that this didn't produce a go.sum delta?

2uasimojo (commented):
This is really nice, Mark. Most of my comments are in the form of wishes :)

LMK when you're ready for another pass.

PR #2884, opened 2026-03-26
HIVE-3134: Axe the direct dependency on the Machine API operator

dlom:
xref: [HIVE-3134](https://redhat.atlassian.net/browse/HIVE-3134)

We're pulling in thousands of lines of code from the MAPI operator just to call two 10-line functions (and only one of those happens at runtime). What if we just didn't do that? It's still remains as an indirect dependency (for now...)

/assign @2uasimojo 

## Summary by CodeRabbit

* **Refactor**
  * Reworked vSphere provider-spec handling to use local decoding helpers and simplified actuator construction.

* **Chores**
  * Removed numerous vendored vSphere/cloud-provider and gcfg/warnings modules and pruned indirect dependencies to streamline the codebase.


pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -152,3 +152,20 @@ func setFolderPath(folder, datacenter string, logger log.FieldLogger) string {
...
+	if rawExtension == nil {
+		return &machineapi.VSphereMachineProviderSpec{}, nil
+	}
+
+	spec := new(machineapi.VSphereMachineProviderSpec)
+	if err := json.Unmarshal(rawExtension.Raw, &spec); err != nil {
+		return nil, fmt.Errorf("error unmarshalling providerSpec: %v", err)
+	}
+
+	// klog.V(5).Infof("Got provider spec from raw extension: %+v", spec)
```
2uasimojo: We should either nix this or replace it with our own logger call (which we could conceivably do in the caller's context to avoid having to pass in our logger).

2uasimojo (commented):
I would generally not be in favor of copy-pasting something that's available upstream, but these two funcs are trivial, and the decluttering of our vendor dir is worth it.

/lgtm

/hold The commented log line isn't crucial: giving you the option to address it here or later (or not at all).

2uasimojo:
/test e2e-vsphere

flake during setup

dlom:
> the decluttering of our vendor dir is worth it

Another impact from this (this is actually how I found this case): The MAPI operator also depends on govmomi, and if/when we wanted to upgrade to the latest (installer just did yesterday), we would have also had to wait for the MAPI operator to update _their_ dependency on it (a breaking change). Now, we are completely untied from them govmomi-wise

dlom:
@2uasimojo the flake might be related to this https://github.com/openshift/release/pull/76852

2uasimojo:
> @2uasimojo the flake might be related to this [openshift/release#76852](https://github.com/openshift/release/pull/76852)

As in, we have to do something to accommodate that change? Like making sure some new piece honors our request to get an extra two IPs? Are you chasing that down?

PR #2886, opened 2026-03-27
HIVE-3134: Axe sjson dependency

dlom:
xref: [HIVE-3134](https://redhat.atlassian.net/browse/HIVE-3134)

This one warrants a test to make sure we don't lost any functionality. The dep was also pulling in a 16 KB png into our tree. Goodbye!

/assign @2uasimojo 

## Summary by CodeRabbit

* **Refactor**
  * Improved label management system for manifest configurations with streamlined error handling.

* **Tests**
  * Added comprehensive test coverage for label configuration logic.

* **Chores**
  * Removed unused external dependency to reduce package size.


pkg/installmanager/installmanager.go
```diff
@@ -1059,6 +1055,28 @@ func patchLabelMachineSetManifest(manifestBytes []byte, poolsByType map[string]*
...
+	labels, _ := metadata["labels"].(map[string]interface{})
+	if labels == nil {
+		labels = map[string]interface{}{}
+		metadata["labels"] = labels
+	}
+	for k, v := range newLabels {
+		labels[k] = v
+	}
+	return json.Marshal(manifest)
+}
```
2uasimojo, line 1078: Hm. Is there a way to call an existing/already-imported upstream lib to decode into a metav1.Object on which we can [SetLabels()](https://github.com/openshift/hive/blob/715f33893b2cf13a04710935617124217e86e440/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/meta.go#L55)? What you've done here should be pretty straightforward and stable, but it feels... dirty.

2uasimojo (commented):
/test e2e

Not so sure about this one @dlom. I remember @lleshchi and I had to churn quite a bit to find a functional and elegant way to do this, and the hand-crafted json introspection you're doing is, no offense, only one of those things :P

I can be talked into it, but would like to explore other options first. See inline.

2uasimojo:
/test e2e

another infra flake

PR #2889, opened 2026-04-08
No-jira: Standard_D2s_v3: instance type doesn't meet requirements of 16 GB min Memory

miyadav:
PTAL when time permits small nit in readme , while doing Azure clusterdeployment , got below error - 

` Invalid value: \"Standard_D2s_v3\": instance type does not meet minimum resource requirements of 4 vCPUsAvailable, controlPlane.platform.azure.type: Invalid value: \"Standard_D2s_v3\": instance type does not meet minimum resource requirements of 16 GB Memory]`

## Summary by CodeRabbit

* **Documentation**
  * Updated Azure installation example to use larger VM sizing for compute and control plane nodes (example values changed from Standard_D2s_v3 to Standard_D4s_v3), ensuring example configs reflect increased resource sizing for installations.

### DIFF
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -528,7 +528,7 @@ For Azure, replace the contents of `compute.platform` and `controlPlane.platform
     azure:
       osDisk:
         diskSizeGB: 128
-      type: Standard_D2s_v3
+      type: Standard_D4s_v3
 ```
 
 and replace the contents of `platform` with:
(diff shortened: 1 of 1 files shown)


2uasimojo (commented):
/lgtm

This matches what we use in our [e2e](https://github.com/openshift/hive/blob/2ded71440cf0c33cc2c1ad0e6c8ad10ecef60aba/test/e2e/postinstall/machinesets/infra_test.go#L146) as well as [hiveutil create-cluster](https://github.com/openshift/hive/blob/0a2941a37e98e5eb42fccce9dfe3e8dd3bb5fa80/pkg/clusterresource/azure.go#L20) ✓

Nit: We could also update [this comment](https://github.com/openshift/hive/blob/fe033076489b3f4279717f1fac664207d3e58d13/apis/hive/v1/azure/machinepool.go#L11).

PR #2895, opened 2026-04-16
HIVE-3149: Azure: Prefilter SKU list query

2uasimojo:
Here we refactor our GetVirtualMachineSku() function to more closely match the modern version of its upstream (installer) counterpart:
- Use a more recent client package for the `List()` call to ensure the `location` filter is available.
- Add the `location` filter to limit the results by region on the server side. This should greatly reduce both the latency and the size (number of pages) of the response.
- Correct the error checking path for the scenario where no SKUs exist in the region. Previously this error was getting swallowed by the loop logic.
- Increase the timeout from 30s to 2m. This may or may not be necessary in light of the filtering improvement.

Note one deviation from the upstream logic: we've removed the section where we were ensuring the result is in the desired region, as this should already be accounted for by the location filter.

## Summary by CodeRabbit

* **Updates**
  * Switched to the newer Azure SKU client, now querying SKUs by region with server-side filtering and a longer timeout for SKU paging.
* **Bug Fixes**
  * More robust VM capability and zone discovery with better handling of missing SKUs, mismatched data, and paging errors.
* **Tests**
  * Expanded test coverage for zoneless deployments, zone-selection failures, pagination, and multiple SKU/error scenarios.

### DIFF
--- a/pkg/controller/machinepool/azureactuator.go
+++ b/pkg/controller/machinepool/azureactuator.go
@@ -201,7 +201,7 @@ func (a *AzureActuator) getZones(region string, instanceType string) ([]string, 
 
 	var res azureclient.ResourceSKUsPage
 	var err error
-	for res, err = a.client.ListResourceSKUs(ctx, ""); err == nil && res.NotDone(); err = res.NextWithContext(ctx) {
+	for res, err = a.client.ListResourceSKUs(ctx, region); err == nil && res.NotDone(); err = res.NextWithContext(ctx) {
 		for _, resSku := range res.Values() {
 			if strings.EqualFold(to.String(resSku.Name), instanceType) {
 				for _, locationInfo := range *resSku.LocationInfo {
@@ -268,7 +268,7 @@ func (a *AzureActuator) getVirtualMachineSku(ctx context.Context, name, region s
 
 	var page azureclient.ResourceSKUsPage
 	var err error
-	for page, err = a.client.ListResourceSKUs(ctx, fmt.Sprintf("location eq '%s'", region)); page.NotDone(); err = page.NextWithContext(ctx) {
+	for page, err = a.client.ListResourceSKUs(ctx, region); page.NotDone(); err = page.NextWithContext(ctx) {
 		if err != nil {
 			return nil, errors.Wrap(err, "error fetching SKU pages")
 		}
--- a/pkg/controller/machinepool/azureactuator.go
+++ b/pkg/controller/machinepool/azureactuator.go
@@ -2,9 +2,11 @@ package machinepool
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
+	azenc "github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/Azure/go-autorest/autorest/azure"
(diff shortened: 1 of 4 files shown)


2uasimojo:
/test e2e-gcp

unrelated, hijacking

pkg/controller/machinepool/azureactuator_test.go
```diff
@@ -32,17 +33,18 @@ func TestAzureActuator(t *testing.T) {
...
 		{
 			name:              "generate single machineset for single zone",
 			clusterDeployment: testAzureClusterDeployment(),
 			pool:              testAzurePool(),
 			mockAzureClient: func(mockCtrl *gomock.Controller, client *mockazure.MockClient) {
-				mockListResourceSKUs(mockCtrl, client, []string{"zone1"})
-				mockGetVMCapabilities(client, "V1,V2")
+				mockGetVMCapabilities(mockCtrl, client, "V1,V2")
 				mockListImagesByResourceGroup(client, []compute.Image{testAzureImage(compute.HyperVGenerationTypesV1)})
+				mockListResourceSKUs(mockCtrl, client, []string{"zone1"})
```
2uasimojo, line 47: ntr: Reordered mocks throughout to match invocation order. (Previously it didn't matter because all the mocks were independent and we didn't check the order; now it matters because two of them are going through the same Client interface func.)

pkg/controller/machinepool/azureactuator_test.go
```diff
@@ -364,11 +567,24 @@ func mockListResourceSKUs(mockCtrl *gomock.Controller, client *mockazure.MockCli
 	)
 }
 
-func mockGetVMCapabilities(client *mockazure.MockClient, hyperVGenerations string) {
-	capabilities := map[string]string{
-		"HyperVGenerations": hyperVGenerations,
-	}
-	client.EXPECT().GetVMCapabilities(gomock.Any(), gomock.Any(), gomock.Any()).Return(capabilities, nil)
+func mockGetVMCapabilities(mockCtrl *gomock.Controller, client *mockazure.MockClient, hyperVGenerations string) {
```
2uasimojo, line 642: ntr: The thrust of the refactor can be seen here: previously GetVMCapabilities was part of the Client interface and thus mocked directly; now that helper is not part of Client so we can mock the low-level ListResourceSKUs instead and have more control.

pkg/controller/machinepool/azureactuator.go
```diff
@@ -240,3 +242,44 @@ func (a *AzureActuator) gen2ImageExists(resourceGroupName string) (bool, error)
...
+			}
+			// Filter out resources that do not match the provided name
+			if strings.EqualFold(name, *sku.Name) {
+				return &sku, nil
+			}
+		}
+	}
+	// err is nil if we didn't find it (page.NotDone() == false above)
+	return nil, errors.Wrap(err, "error fetching SKU pages")
+}
```
2uasimojo, line 288: ntr: Moved from Client interface, and also fixed up some. May be worth pulling down the funcs before/after to compare.

2uasimojo:
@coderabbitai Please verify that I addressed [these comments](https://github.com/openshift/hive/pull/2895#pullrequestreview-4123060705) to your satisfaction.

2uasimojo:
/test e2e-azure

Timeout, but we allllmost got there. Hopefully between this and https://github.com/openshift/release/pull/76486 we're there.

PR #2904, opened 2026-05-08
HIVE-2946: Remove empty vSphere dataDisks from install-config

chdeshpa-hue:
## Summary
- remove empty vSphere `dataDisks` keys from install-config after install-manager mutates secrets, avoiding unknown-field warnings on older payloads
- keep explicitly configured `dataDisks` intact so newer payload behavior is preserved
- add focused unit tests for both removal and preservation scenarios

## Test plan
- [x] `go test ./pkg/installmanager -run 'Test_removeUnsupportedVSphereDataDisks|Test_pasteInPullSecret'`
- [ ] Run broader Hive CI jobs
- [x] Confirm warning is absent in ACM vSphere 4.18 flow

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/installmanager/installmanager.go
+++ b/pkg/installmanager/installmanager.go
@@ -2034,7 +2034,10 @@ func removeUnsupportedVSphereDataDisks(ic *installertypes.InstallConfig, icData 
 
 	modifiedBytes, err := yamlutils.ApplyPatches(icData, patches)
 	if err != nil {
-		return nil, errors.Wrap(err, "error removing unsupported vSphere dataDisks field from install-config")
+		// Best-effort cleanup to suppress warnings on older payloads.
+		// Aborting the install would be worse than the original warning.
+		log.WithError(err).Warn("failed to remove empty vSphere dataDisks from install-config; proceeding with original data")
+		return icData, nil
 	}
 	return modifiedBytes, nil
 }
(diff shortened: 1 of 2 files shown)


2uasimojo:
Thanks for working on this!

Since the symptom is just a log warning, and the hive workaround you're proposing is nontrivial, I'd rather hold out for the proper fix. That would be to add `omitempty` to the upstream [field](https://github.com/openshift/installer/blob/bfd1200dcac75793e21a9c685dcc206e3da9dfa9/pkg/types/vsphere/machinepool.go#L33) in the installer repo, and then we (hive) will pick it up when we revendor. You're welcome to propose the former.

@jcpowermac is there any reason *not* to include `omitempty` on the `DataDisks` field for vsphere machinepools? I see it's there for the corresponding field in e.g. [Azure](https://github.com/openshift/installer/blob/b1702662d4bb5dfd66dc784f5b4fcad7a4fe17b1/pkg/types/azure/machinepool.go#L85).

chdeshpa-hue:
@2uasimojo thanks for the reivew, once @jcpowermac answers to your query, I'll work one the installer fix if it is needed.

2uasimojo:
@chdeshpa-hue Discussed offline with Joe & Neil. They're on board with you proposing the installer change. We'll need to do this through an OCPBUGS card. Can you please open that and link it to the HIVE card? Let me know if you need any assistance.

chdeshpa-hue:
@2uasimojo Thank you for the help, here is the [PR](https://github.com/openshift/installer/pull/10551) for installer. I have created a bug and linked to hive card as well. I will now close this PR

PR #2913, opened 2026-05-27
HIVE-3163: update modules for various cves

2uasimojo:
## Summary by CodeRabbit

* **Security & Stability**
  * Hardened SSH/key handling and SK (security-key) verification; ED25519 key validation
  * Improved redirect handling with configurable sensitive-header scrubbing
  * HTML parsing and HTTP/2 connection/error handling improvements
  * Tightened crypto padding and platform syscall robustness

* **Chores**
  * Bumped vendored Go dependencies and third‑party libs (golang.org/x/net, x/crypto, x/sys, kiota-http-go)


hack/modcheck.go
```diff
@@ -21,51 +22,59 @@ func main() {
 		doFix = true
 	}
 	rootgmf := readGoMod(rootpath)
-	apisgmf := readGoMod(apispath)
+	rc := 0
+	rc += processGoMod(apispath, rootgmf)
+	rc += processGoMod(otepath, rootgmf)
+	os.Exit(rc)
```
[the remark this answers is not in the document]
2uasimojo: Yeah, there's also the issue that, if we ever add up to 256, we'll get a false "success". I'm not worried about these things right now.

hack/modcheck.go
```diff
@@ -21,51 +22,59 @@ func main() {
...
+}
+
+func processGoMod(subpath string, rootgmf *modfile.File) int {
+	defer fmt.Println("======================================")
+	subgmf := readGoMod(subpath)
 	needWrite := false
-	insync, err := processRequire(*apisgmf, mapRequire(rootgmf.Require), mapRequire(apisgmf.Require))
+	insync, err := processRequire(*subgmf, mapRequire(rootgmf.Require), mapRequire(subgmf.Require))
```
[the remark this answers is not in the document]
2uasimojo: It turns out that the nested bits of that struct are pointers and thus the updates still take hold even though the outermost var is passed by value. But it's not cool to rely on internal behavior like that, so I'm changing it to by-reference as suggested.

2uasimojo:
NB: when I run `modfix`, it changes some things in test/ote/go.mod; but when I run `go mod tidy` in that dir, it reverts them. At the moment, none of them are flagging CVEs (I'm pretty sure) so I'm going to leave them as tidied.

hack/modcheck.go
```diff
@@ -10,6 +10,7 @@ import (
 const (
 	rootpath = "go.mod"
 	apispath = "apis/go.mod"
+	otepath  = "test/ote/go.mod"
```
miyadav: @2uasimojo shall we drop this update  from this test ?

2uasimojo: Copying from [slack](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1779981221819139?thread_ts=1779916162.765429&cid=CE3ETN3J8) for posterity:

> We can, but it complicates things.
> The issue is that CVE scanners don't know that test/ote/* is not in the production code path, so if they see a known-vulnerable dep version in test/ote/go.mod, they'll flag hive as affected.
> We can close such issues as not-a-bug, but:
> - It takes extra effort for us to figure out that the vulnerable dep is isolated to test code
> - By flagging a card as not-a-bug, it prevents subsequent scans from flagging that same vuln (so the card isn't reopened/duplicated). But this introduces a hole whereby, if that vulnerable dep is later introduced in the production go.mod, it'll slip through the cracks.
> 
> What I'm discovering now is that the deps in test/ote are actually ahead of those in the root go.mod. So we could backtrace to figure out what's bumping them up (clearly some other dep that's not shared with the root); or we could bring the shared deps in the root go.mod up to the levels they're at in test/ote.
> I think the latter is worth a try, as I'm not seeing any major version changes. The ones that concern me the most would be the openshift ones (apis, client-go, library-go) as they may have cross-deps that try to drag in untenable changes in the root.
> But it should be pretty quick to find that out...

2uasimojo: I tried to unwind this for a while and couldn't get it done. The sticking point was openshift/origin, which is being consumed directly by test/ote. That guy pulls in a version of msgraph that's ahead of what installer expects, and has breaking changes.
Options:
- Downgrade openshift/origin. I fear OTE is taking advantage of modern features there, though.
- `replace` msgraph. Since there are breaking changes, we would need to vet that OTE still builds and runs. I'm not sure what that looks like yet, but I'm pretty sure we don't have at least the latter in presubmit CI.
- Wait for installer to upgrade msgraph (this has not happened in [master](https://github.com/openshift/installer/blob/ae2ae059f3b4b3ea1103593efb3f89e188ed938f/go.mod#L70) at the time of this writing) and revendor.

Any of those are more complex than we want to handle in this PR, so I've ripped out the modcheck and test/ote/go.mod commits for now.

PR #2916, opened 2026-06-01
Hive-3130: bundle-gen rewrite + port to shell

suhanime:
## Summary by CodeRabbit

* **Refactor**
  * Updated the Hive OperatorHub bundle generation and catalog update workflow to use shell-based automation.
  * Simplified version initialization to use a fixed major/minor prefix and compute versioning from the selected commit history.

* **Chores**
  * Removed legacy Python/GitHub automation components and trimmed related developer dependencies.
  * Refreshed Hive CSV template instructions to match the new bundle generation flow.

* **Bug Fixes**
  * Improved command-line help handling to correctly recognize `-h` and `--help` and validate arguments more consistently.


hack/bundle-gen.sh
```diff
@@ -0,0 +1,531 @@
...
+check_deps() {
+    local missing=()
+    for cmd in yq jq git gh; do
+        command -v "$cmd" &>/dev/null || missing+=("$cmd")
+    done
+    if [[ ${#missing[@]} -gt 0 ]]; then
+        echo "Error: missing required tool(s): ${missing[*]}" >&2
+        echo "  yq (v4): https://github.com/mikefarah/yq" >&2
+        echo "  gh:      https://cli.github.com/" >&2
+        exit 1
```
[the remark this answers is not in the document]
suhanime: curl's too minor to bother with - most linux users have it. We can ignore this one.

hack/bundle-gen.sh
```diff
@@ -0,0 +1,550 @@
...
+    echo "Cloning $HIVE_REPO to $HIVE_REPO_DIR"
+    git clone "$HIVE_REPO" "$HIVE_REPO_DIR"
+
+    cd "$HIVE_REPO_DIR"
+
+    # version2.sh sets globals: COMMIT, BRANCH_NAME, PREFIX, COMMIT_COUNT
+    # In OperatorHub mode the branch is always master; --dummy-bundle overrides
+    # this with an explicit branch (e.g. mce-2.1). Passing the branch explicitly
+    # avoids ambiguous auto-detection when the repo has many branches.
+    version_init "${DUMMY_BUNDLE:-master}" "${COMMIT_ISH:-}"
```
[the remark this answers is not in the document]
suhanime: This would break the local clone usage, which is usually how the users execute bundle-gen currently. While the lack of commit if dummy-bundle option is used (with a non-master branch) might result in error, it will still come up with a useful failure message.

hack/version2.sh
```diff
@@ -203,6 +199,13 @@ branch_from_commit() {
         fi
     done < <(git for-each-ref --format='%(refname)' refs/heads/ refs/remotes/ 2>/dev/null)
 
+    # Deduplicate by logical branch name: the same branch appearing under multiple
+    # remotes (e.g. refs/heads/master and refs/remotes/origin/master both resolve
+    # to "master") should be counted once.
+    mapfile -t here        < <(printf '%s\n' "${here[@]+"${here[@]}"}"              | sort -u | grep -v '^$')
+    mapfile -t ancestors   < <(printf '%s\n' "${ancestors[@]+"${ancestors[@]}"}"     | sort -u | grep -v '^$')
+    mapfile -t descendants < <(printf '%s\n' "${descendants[@]+"${descendants[@]}"}" | sort -u | grep -v '^$')
```
2uasimojo: I can't parse this. If all we're trying to do is deduplicate, can we just make `here`, `ancestors`, and `descendants` associative arrays?

@coderabbitai Can you break down this goofy syntax for me? I get the `sort` and `grep`. I get that `mapfile` is consuming a "file" with newlines and putting the contents into an array such that each line (minus the newline character) is one array entry. The bit in the middle is a mystery though. Talk me through from the first `<` to the last `"`.

2uasimojo: The "downstream code" is only consuming the contents if there's exactly one entry; so something like
```bash
    if [[ ${#here[@]} -eq 1 ]]; then
        log "debug" "Found unique branch ${here[0]} at commit $(shortsha)"
        echo "${!here[@]}"   # <== prints "all keys", but there's only one.
        return 0
    fi
```
...would work fine, no?

2uasimojo: Yeah, definitely. @suhanime wdyt?

hack/version2.sh
```diff
@@ -2,280 +2,36 @@
...
-        COMMIT=$(git rev-parse "$commit_ish_arg")
+        if ! COMMIT=$(git rev-parse --verify "$commit_ish_arg" 2>/dev/null); then
+            echo "Error: '$commit_ish_arg' could not be resolved. Ensure the commit exists and the repository is up to date." >&2
+            exit 1
+        fi
     else
-        # Assume HEAD. If we got branch_name but not commit_ish, and they're sitting somewhere unrelated,
-        # that's a user error. (To build on the branch commit, just specify the branch to commit_ish.)
+        # Assume HEAD
         COMMIT=$(git rev-parse HEAD)
```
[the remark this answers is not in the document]
suhanime: @coderabbitai will git rev-list -l work better instead?

hack/bundle-gen.sh
```diff
@@ -0,0 +1,522 @@
...
+
+Hive Bundle Generator and Publishing Script.
+
+This utility will:
+  1. Clone the repo specified by --hive-repo (default: $HIVE_REPO_DEFAULT).
+  2. Check out the commit-ish specified by --commit (default: tip of master).
+  3. Look for a hive image in --image-repo with the tag specified by
+     --image-tag-override (default: 10-char SHA of the checked-out commit).
+     - If not found, build and push that image.
+     - Skip this step with --skip-image-validation.
```
2uasimojo, line 50: This may want to mention that we also skip if the image repo isn't quay.

hack/bundle-gen.sh
```diff
@@ -0,0 +1,522 @@
...
+        command -v "$cmd" &>/dev/null || missing+=("$cmd")
+    done
+    if [[ ${#missing[@]} -gt 0 ]]; then
+        echo "Error: missing required tool(s): ${missing[*]}" >&2
+        echo "  yq (v4): https://github.com/mikefarah/yq" >&2
+        echo "  gh:      https://cli.github.com/" >&2
+        exit 1
+    fi
+
+    if ! yq --version 2>&1 | grep -q 'version v4'; then
```
2uasimojo, line 441: In my env, I have v3 installed as `yq` and v4 installed as `yq-v4.47.1`. It makes it convenient for me if `yq` in this script is written as `$YQ` (initialized as `${YQ:-yq}`) so I can override it with an env var. Otherwise I have to re`$PATH` or rename binaries, which is harder.

Of course, I may be the snowflake here. It may be time for me to make v4 my default. Feel free to push back if I'm being silly.

hack/bundle-gen.sh
```diff
@@ -0,0 +1,522 @@
...
+        --title "$pr_title" \
+        --body "$body"
+}
+
+# check_deps — fail fast with a clear message if required tools are missing
+check_deps() {
+    local required=(yq git)
+    # jq is only needed to parse the quay.io API response
+    if ! $SKIP_IMAGE_VALIDATION && [[ "${IMAGE_REPO%%/*}" == "quay.io" ]]; then
+        required+=(jq)
```
2uasimojo, line 423: jq also has incompatible version issues. Should we add a version check?

hack/bundle-gen.sh
```diff
@@ -0,0 +1,522 @@
...
+        required+=(gh)
+    fi
+
+    local missing=()
+    for cmd in "${required[@]}"; do
+        command -v "$cmd" &>/dev/null || missing+=("$cmd")
+    done
+    if [[ ${#missing[@]} -gt 0 ]]; then
+        echo "Error: missing required tool(s): ${missing[*]}" >&2
+        echo "  yq (v4): https://github.com/mikefarah/yq" >&2
```
2uasimojo, line 436: At the risk of overengineering, we could store the required tools and helpful strings as keys/values in an associative array, and print all/only the relevant pieces.

hack/version2.sh
2uasimojo, line 1: nit: version2.sh ignores args, which can be confusing (e.g. `version2.sh abc123` will always compute the version string for HEAD, not for `abc123`). Consider adding a check.

Another possibility is for version2.sh to accept the commit-ish and always do version_init. Then bundle-gen.sh could source it as `source version2.sh "$COMMIT"` and we could collapse/simplify some of the indirection in both scripts. It's probably not worth all that though...

2uasimojo (commented):
/lgtm

Nothing worth holding for. Feel free to submit a fup if you care to address any of these issues.

Thanks for tackling this!

hack/bundle-gen.sh
suhanime: From what I can tell, yq version 3 versus 4 are very different syntactically, making the check necessary. I don't think versions of jq are that dramatically different, and we are only using it for basic JSON parsing, so I think we are good here.

suhanime:
@2uasimojo addressed most of your comments via #2924

PR #2917, opened 2026-06-03
ROSAENG-6863: Replace certman-operator with direct cert-manager 

syed:
Add cert-manager Certificate CR management directly in Hive controllers, eliminating the dependency on the external certman-operator. Both the controlplanecerts and remoteingress controllers now create cert-manager Certificate CRs for CertificateBundles with Generate=true.

- Add pkg/controller/utils/certmanager.go with EnsureCertManagerCertificate helper that creates/updates cert-manager Certificate CRs with proper owner references and ACME DNS01 solver configuration
- Integrate into controlplanecerts and remoteingress reconcile loops
- Register cert-manager v1 types in the shared scheme
- Add RBAC permissions for cert-manager.io certificates
- Update vendored dependencies (cert-manager v1.20.2, gateway-api v1, and various cloud provider SDK bumps)

Assisted-by: Claude Code (claude-opus-4-6)


2uasimojo:
@coderabbitai I do not accept your refusal to review this PR. You should [know](https://github.com/openshift/hive/blob/master/.coderabbit.yaml#L26-L34) you are supposed to ignore the vendor/ directory except for a very small set of files, none of which feature in this PR. Please review the remaining files.

pkg/controller/remoteingress/remoteingress_controller.go
```diff
@@ -235,6 +241,30 @@ func (r *ReconcileRemoteClusterIngress) Reconcile(ctx context.Context, request r
 	return reconcile.Result{}, nil
 }
 
+// ensureCertManagerCertificates creates cert-manager Certificate CRs for any ingress
+// CertificateBundles with Generate=true.
+func (r *ReconcileRemoteClusterIngress) ensureCertManagerCertificates(ctx context.Context, cd *hivev1.ClusterDeployment, cdLog log.FieldLogger) error {
+	seen := map[string]bool{}
```
2uasimojo, line 247: nit: Consider using a [Set](https://github.com/openshift/hive/blob/153f94f1e7911714bde36e2232fca70b14a4808a/vendor/k8s.io/apimachinery/pkg/util/sets/set.go#L25).

pkg/controller/utils/certmanager.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// GetDomainsForCertBundle extracts the DNS names that should be included in a certificate
+// for the given CertificateBundle, based on the ClusterDeployment's control plane and
+// ingress configuration.
+func GetDomainsForCertBundle(bundle hivev1.CertificateBundleSpec, cd *hivev1.ClusterDeployment) []string {
+	var domains []string
+
+	if cd.Spec.ControlPlaneConfig.ServingCertificates.Default == bundle.Name {
+		controlPlaneDomain := fmt.Sprintf("api.%s.%s", cd.Spec.ClusterName, cd.Spec.BaseDomain)
```
2uasimojo, line 93: Can we count on this? What about APIURLOverride? Should we maybe be using [defaultControlPlaneDomain()](https://github.com/openshift/hive/blob/9ccaa4ce11cb94fde3a981a060dba216f5dbc5af/pkg/controller/controlplanecerts/controlplanecerts_controller.go#L431) instead?

pkg/controller/utils/certmanager.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	domains := GetDomainsForCertBundle(bundle, cd)
+	if len(domains) == 0 {
+		cdLog.WithField("bundle", bundle.Name).Debug("no domains found for certificate bundle, skipping cert-manager Certificate creation")
+		return nil
+	}
+
+	certName := fmt.Sprintf("%s-%s", cd.Name, bundle.Name)
+	certName = strings.ToLower(certName)
```
2uasimojo, line 52: nit: save a trivial amount of memory
```suggestion
	certName = strings.ToLower(fmt.Sprintf("%s-%s", cd.Name, bundle.Name))
```

pkg/controller/utils/certmanager.go
```diff
@@ -0,0 +1,151 @@
...
+			Kind:       "Certificate",
+			APIVersion: "cert-manager.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: certmanagerv1.CertificateSpec{
+			Subject: &certmanagerv1.X509Subject{
+				Organizations: []string{"Red Hat - Open Cluster Manager"},
```
2uasimojo, line 128: I'm not well-educated in this space, but this seems like it would not be appropriate for self-managed clusters?

pkg/controller/utils/certmanager.go
```diff
@@ -0,0 +1,151 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+const (
+	CertManagerIssuerName = "public-issuer"
+	CertManagerIssuerKind = "ClusterIssuer"
+	certDuration          = 90 * 24 * time.Hour
+	certRenewBefore       = 30 * 24 * time.Hour
```
2uasimojo, line 27: I need to understand these hardcoded values.

(Slight aside: any reason the first two are exported?)

2uasimojo (commented):
- Are there upgrade concerns with this change? That is, how would a hive consumer be doing cert management prior, and what would they need to do when upgrading hive to/past this commit to transition management to the native controllers?
- I'm concerned that there may be ROSA-specific pieces in this PR that would potentially break in non-ROSA hive envs.
- The `verify` fail can be remedied with `make update-gofmt`.
- The `security` fail can be ignored: those issues will be addressed elsewhere; we'll override the job when ready to merge.

syed:
Thank you for the review!

> * Are there upgrade concerns with this change? That is, how would a hive consumer be doing cert management prior, and what would they need to do when upgrading hive to/past this commit to transition management to the native controllers?

Prior to this change, cert management was done via [certman-operator](https://github.com/openshift/certman-operator). This is a custom ACME client which runs on the hive shards. We had an outage last week because Let's Encrypt started using a different root cert and there was a bug in certman-operator where we were not passing the full cert chain which caused TLS errors on the API endpoints. This is the primary motivation from us to move to cert-manager as it will remove the maintenance burden  of a bespoke operator. The way certman-operator works is it watches `ClusterDeployment` and looks for any `CertificateBundles` and if required generates them and stores them in a secret. This is how cert-manager works too. So instead of certman-operator, the cert will be generated and lifecycled by cert-manager. The secret which holds the cert and key has the same format for both operators so that doesn't need to change. So in theory the only migration step is to remove certman-operator after this change is deployed and configure cert-manager (this is already configured in all the hive shard clusters)

> * I'm concerned that there may be ROSA-specific pieces in this PR that would potentially break in non-ROSA hive envs. 

My apologies, I'm not familiar with other non-ROSA envs. One potential issue I see is DNS validation. `cert-manager` support most cloud providers so that is something that needs to be validated

2uasimojo:
> I'm not familiar with other non-ROSA envs.

So yeah, in addition to managed openshift (which btw is also more than just ROSA -- there are non-ROSA OSD spokes on AWS and GCP managed from those same shards) hive is also bundled with ACM, and available standalone on OperatorHub. It supports ~ half a dozen infrastructure providers, including on-prem as well as cloud models. And I'm afraid I don't know enough about cert management to understand all the ramifications of bringing it "in house" like this.

This work implies that, today, Certificate CRs are managed externally in hive hub environments. I assume there's more than one way that can be done -- e.g. on-prem vsphere under ACM _probably_ doesn't use openshift/certman-operator. And what happens when hive upgrades through this path on a hub where some other controller (e.g. gitops) expects to have ownership of those CRs? Do we end up thrashing them? If nothing else, I would expect owner references to be contended.

I'm starting to suspect that this thing may want to be conditioned on a HiveConfig and/or ClusterDeployment knob, such that the default behavior doesn't change on a hive upgrade. And if my concerns about hardcoded values (like the Organization and the renewal times) turn out to be valid, exposing those as configurables in the same place may be necessary/appropriate.

And wherever we go, I'm guessing we'll want doc changes to explain it.

I sure would like to bring in some more expertise in both certs in general and how they're managed/used with hive. I'm not sure who that would be.

@coderabbitai do you have thoughts on the above?

pkg/controller/utils/certmanager.go
syed: these are moved from certman-operator. I will remove them and use lib provided values

syed:
I also do not know enough about cert management across the whole org to give a definite answer. Let me talk to a few folks and get back

pkg/controller/utils/certmanager.go
dustman9000: We are still moving old certman-operator logic to hive? We not let cert-mgr determine these duration/renewal lifetimes?

syed:
Update from ARO Classic, they don't use the `CertificateBundleSpec` on their `ClusterDeployment` and manage certs outside of hive. One question I have is if someone sets `CertificateBundleSpec.Generate` on a spec on an on-prem, or a non-ROSA cluster, what does hive do?

2uasimojo:
> Update from ARO Classic, they don't use the `CertificateBundleSpec` on their `ClusterDeployment` and manage certs outside of hive. One question I have is if someone sets `CertificateBundleSpec.Generate` on a spec on an on-prem, or a non-ROSA cluster, what does hive do?

I can't see that hive does anything with that field today. @coderabbitai can you tell whether it has any effect? Or maybe if we offload the bundle contents to some other agent that may act on it?

syed:
Thanks for the info. As I understand `certman-operator` is the only entity that watches the `generate=true` flag and generates the certs. I haven't found any evidence of it being using in ARO or GCP. Now the question I want to pose is, would it be something that Hive would want to to natively since it exposes the "generate certs" feature in it's API to the end user?

2uasimojo:
> would it be something that Hive would want to to natively since it exposes the "generate certs" feature in it's API to the end user?

This sounds reasonable on the surface, but I don't really have the background to understand all the implications.

Would need to push an RFE through The Process™.

2uasimojo:
Pretty sure we decided to take a different approach, but I don't remember for sure. Reopen if still needed.

PR #2921, opened 2026-06-23
OCPBUGS-91668: vSphere machinepool: preserve old-style RHCOS template for upgraded clusters

chdeshpa-hue:
## Problem

After upgrading to MCE 2.9+ (ACM 2.14+ / OCP 4.19+), new Hive MachinePool objects for existing vSphere clusters generate MachineSets that reference an RHCOS template that does not exist in vCenter.

**Root cause:** [HIVE-2391](https://redhat.atlassian.net/browse/HIVE-2391) (PR #2851) stopped backfilling `Topology.Template` from existing cluster state. Installer template naming changed in Dec 2024 from `{infraID}-rhcos-{region}-{zone}` to `{infraID}-rhcos-{fdName}`.

## Fix

1. **Template backfill (actuator):** At construction, match each failure domain (empty `Topology.Template`) to remote MachineSets using workspace conjunction: Datacenter + Datastore + Server + VMGroup + ResourcePool (MCO PR #5745 pattern). Adopt the template from the first matching remote MachineSet.

2. **VMGroup source:** For host-group failure domains, read `VMGroup` from the spoke cluster's `Infrastructure/cluster` CR (`ZoneAffinity.HostGroup.VMGroup`) — not derived via `fmt.Sprintf` in Hive. Handles customer-created VM groups.

3. **In-memory only:** Templates applied on `DeepCopy()` of install config inside `GenerateMachineSets`; ClusterDeployment is never mutated (GCP actuator pattern).

4. **Webhook:** `Topology.Template` spoof allows unset→set on ClusterDeployment for manual remediation.

## Test Evidence

```bash
$ go test ./pkg/controller/machinepool/... -run 'TestVSphereActuator|TestBackfillVSphereTemplates' -count=1 -v
```

```
--- PASS: TestVSphereActuator (includes backfill_template_from_remote_MachineSet_end-to-end)
--- PASS: TestBackfillVSphereTemplates (includes host-group customer VMGroup case)
PASS
ok  	github.com/openshift/hive/pkg/controller/machinepool
```

## Files Changed

- `pkg/controller/machinepool/vsphereactuator.go` — vmGroupMap from spoke Infrastructure; template backfill
- `pkg/controller/machinepool/machinepool_controller.go` — thread Infrastructure into actuator builder
- `pkg/controller/machinepool/vsphereactuator_test.go` — e2e backfill + VMGroup tests
- `pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go` — Template spoof (unchanged this push)

Jira: [OCPBUGS-91668](https://redhat.atlassian.net/browse/OCPBUGS-91668)


2uasimojo:
Thanks for working on this, @chdeshpa-hue

Somehow your commit/PR got attached to an unrelated jira card. Would you please update both to reference https://redhat.atlassian.net/browse/OCPBUGS-91668 ?

I don't love the idea of essentially porting code from an ancient version of installer to generate a very specific template format. There are several potential problems with this, including that we still need an answer for the case where that format isn't valid. While I appreciate that this only triggers if we find that template on an existing MachineSet, I'm not convinced the additional code maintenance burden is worth it. For instance, what if the matching MachineSet is itself bogus for that reason? And sure, we can add yet more code to check on that... but you see what I'm getting at: we're throwing increasingly messy code at this issue.

In other actuators -- and in this one prior to the big zonal changes -- we try to copy the template from an existing master machine. However, that probably doesn't work here since the FD name is part of that template: we would need to find a master in the right FD, and I'm not sure that's always possible.

I'm thinking the least-worst approach here may be to make it easier for the user to set the Template in the ClusterDeployment. We have precedent for easing our webhooks to allow setting certain values if previously unset, but still disallow edits if already set. That, plus some documentation (hive repo docs; a KCS article; something else that can easily be found by customers/TAMs) describing the issue and the remedy, would be clean and crisp, if not the ultimate ideal UX.

WDYT?

chdeshpa-hue:
Thanks for the detailed feedback @2uasimojo — appreciate you taking the time.

### Acknowledging your concern

You're right that the current implementation ports installer-era naming logic (`{infraID}-rhcos-{region}-{zone}`) into Hive, and that's a coupling we shouldn't carry. If the installer changes naming conventions again, this becomes a maintenance point. I agree that's not the direction we want.

### The workaround we've been providing

We've already been advising customers on exactly the manual approach you're describing. In OCPBUGS-91668, the documented workaround is:

> *Manually set `spec.platform.vsphere.infrastructure.failureDomains[].topology.template` in the ClusterDeployment to your existing template name (e.g., `{clusterID}-rhcos-generated-region-generated-zone`) before creating the MachinePool.*

```yaml
spec:
  platform:
    vsphere:
      infrastructure:
        failureDomains:
          - topology:
              template: "{clusterID}-rhcos-generated-region-generated-zone"
            # ... rest of failure domain config
```

This is safe on all MCE versions and prevents the issue regardless of Hive version. We would also prepare a **KCS article** documenting this workaround and the upgrade impact so it's discoverable through the support knowledge base.

### Why manual workaround alone isn't sufficient

That said, with the growing adoption of RHACM/Hive for vSphere `ClusterDeployment` management, relying solely on a manual workaround has practical limitations:

1. **Discovery burden** — the customer must know to look for this *before* MachinePool creation fails. Nothing in the upgrade path signals that `Topology.Template` needs attention.
2. **Template name lookup** — the customer needs vCenter access or must run `oc get machineset -o json` on the managed cluster and parse the provider spec to find the correct template name.
3. **Per-failure-domain** — each failure domain needs the template set individually. Multi-FD deployments multiply the manual effort.
4. **Reactive, not preventive** — customers discover the problem only after machines fail to provision. By that point, they're already in a support case.

Two active SFDC cases are linked to this bug, which suggests the workaround documentation alone isn't catching the affected population in time.

### Revised code proposal: "Read, Don't Reconstruct"

Rather than constructing any template name format, the simplified approach **reads the template directly from existing remote MachineSet provider specs**:

```
for each failureDomain where Topology.Template == "":
    for each remoteMachineSet:
        template := extractTemplateFromProviderSpec(remoteMachineSet)
        if template != "":
            failureDomain.Topology.Template = template
            break
```

What this changes:
- **No installer code ported** — we never construct a template name, just read what's already running in the cluster
- **Format-agnostic** — works regardless of naming convention (old or new)
- **~15 lines** of backfill code instead of ~40
- **Same safety** — only activates when `Topology.Template` is empty AND remote MachineSets exist with a template set

### Proposed path forward (two small PRs)

| PR | Scope | Purpose |
|----|-------|---------|
| **This PR (simplified)** | Auto-backfill `Topology.Template` from remote MachineSet provider specs | Preventive fix — catches it silently using the same data source the customer would query manually |
| **Companion PR (new)** | Relax webhook to allow setting `Topology.Template` when previously unset | Per your suggestion — gives customers a manual escape hatch alongside the auto-fix |

Both together give us defense in depth: the code catches it automatically for new MachinePool creates, and the webhook relaxation lets customers (or support) manually override if needed. The KCS article would document both paths.

Happy to push the simplified diff and discuss further.

chdeshpa-hue:
**Follow-up — tested locally:**

While validating the manual workaround path, I found that `Topology.Template` is **immutable on existing ClusterDeployments**. Hive's admission webhook rejects the change:

```
Attempted to change ClusterDeployment.Spec which is immutable
Unsupported change:
Platform.VSphere.Infrastructure.FailureDomains.Topology.Template
```

This applies whether the value is a bare template name or a full vSphere inventory path.

This means for the affected population (existing clusters installed pre-Dec 2024, now on MCE 2.9+), there is currently **no manual path** to fix this — the ClusterDeployment already exists with an empty `Topology.Template`, and it cannot be patched.

The auto-backfill code fix is unaffected (it operates on a `DeepCopy()` during MachineSet generation, not a CD mutation), making it the only viable fix path for existing clusters today. The webhook relaxation companion PR moves from "nice-to-have" to "required for any manual workaround to work."

2uasimojo:
Okay, I still don't love this, but I'll concede it's a lesser evil than forcing consumers to notice, discover, and remedy by hand. To summarize:

- Cycle through remote msets looking for FD-specific Template values to populate into the in-memory install config structure we're passing to MachineSets(). I suppose we just take the first value we see for a given FD, and stop when all the FDs requested for the given pool are populated?
- Relax the webhook, so customers can preemptively remedy the issue by hand without having to go through the awkwardness of pausing reconciliation and deleting the VWHC.
- Write a KCS article describing this failure mode, what our code is _trying_ to do about it, how that code can fail (e.g. we don't find sufficient info in existing remote msets -- is it also possible for more than one template to exist in a given FD and we choose the wrong one??), and how to remedy it manually (by setting the values in the CD).

Thanks.

chdeshpa-hue:
Thanks @2uasimojo — confirmed on all three points. The force-push (`e647cb0`) addresses each:

**1. FD-specific template matching:**

Correct — we cycle through remote MachineSets, decode each provider spec, and match by datacenter (same datacenter as the failure domain) + infraID prefix (template name starts with `{infraID}-`). First match for a given FD wins, and we stop looking for that FD once populated. If all FDs in the pool get populated, we stop entirely.

**On "more than one template in a given FD":** In practice, a single FD has one RHCOS template per cluster (the installer creates exactly one per failure domain at install time). However, if somehow two different templates existed for the same FD (e.g., a leftover from a failed install attempt), we'd pick the first one found in the remote MachineSet list. Since all existing MachineSets in a healthy cluster reference the same template for a given FD, this isn't a realistic failure mode — but the KCS will document it as an edge case with the manual override as the remedy.

**2. Webhook relaxation:**

Included in this commit. `Topology.Template` can now be set if previously empty (unset → set is allowed; set → changed is still rejected). This gives customers/TAMs the manual path without needing to pause reconciliation.

**3. KCS article:**

In progress. Will cover: the failure mode (upgrade path that leaves `Topology.Template` empty), what the auto-backfill code does, when it can't help (no remote MachineSets exist, or template can't be determined), and the manual remedy via the now-relaxed webhook.

Also — Jira target version is being set to 5.0.0 so `/jira refresh` should clear the `invalid-bug` label once that's done.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -76,3 +88,77 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+		}
+		if spec.Workspace.Datacenter != fd.Topology.Datacenter {
+			continue
+		}
+		// Verify the template belongs to this cluster. Installer-generated templates are
+		// always prefixed with the cluster's infraID. This prevents picking up templates
+		// from other clusters sharing the same vCenter datacenter.
+		if !strings.HasPrefix(templateBaseName(spec.Template), infraID) {
+			continue
+		}
```
2uasimojo: I don't think it's possible to get a different infra ID's Template because the source of the data is a list of remote MachineSets already [retrieved](https://github.com/openshift/hive/blob/a1e2bddd5bd35371a2b4479c5eaeba0dba66510d/pkg/controller/machinepool/machinepool_controller.go#L514) from the target cluster. So I don't think we need this check (and coderabbit's comment below becomes n/a).

Furthermore, it is at least theoretically possible for the user to have built those MachineSets with a specified Template that doesn't have the infra ID in it at all, in which scenario this would have caused us not to grab it, resulting in the original failure mode.

@coderabbitai please confirm.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -76,3 +88,77 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+		if err != nil {
+			logger.WithError(err).WithField("machineSet", ms.Name).Warn("cannot decode VSphereMachineProviderSpec")
+			continue
+		}
+		if spec.Template == "" || spec.Workspace == nil {
+			continue
+		}
+		if spec.Workspace.Datacenter != fd.Topology.Datacenter {
+			continue
+		}
```
2uasimojo: Need to phone a friend here. @jcpowermac is Datacenter always sufficient to disambiguate two FDs? I thought we could have e.g. same Datacenter, different Datastores? Folders? Other?

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -892,6 +894,7 @@ func hasChangedImmutableField(oldObject, cd *hivev1.ClusterDeploymentSpec) (bool
 	opts := cmp.Options{
 		cmpopts.EquateEmpty(),
 		cmpopts.IgnoreFields(hivev1.ClusterDeploymentSpec{}, mutableFields...),
+		cmpopts.IgnoreFields(installervsphere.Topology{}, "Template"),
```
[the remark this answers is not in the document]
2uasimojo: Agree. We typically deal with this kind of situation by overriding the field in the in-memory CD with that of the proposed update if the conditions (in this case, that the field was originally blank) are satisfied.

Also, let's get unit tests for this change please.

pkg/controller/machinepool/vsphereactuator.go
jcpowermac: nope, vm-host zonal will always be the same datacenter. prior-4.22 zonal could be the same datacenter but isn't required. 

really it should be the fd, region,zone names, those shouldn't be duplicated

chdeshpa-hue:
**v2: Addressing review feedback**

**1. Removed infraID prefix check** (per @2uasimojo)
Remote MachineSets are already fetched from the target cluster — no cross-cluster contamination. Removed `strings.HasPrefix` check that would reject user-specified templates.

**2. Webhook: spoof pattern for Template** (per @2uasimojo)
Replaced `cmpopts.IgnoreFields(Topology{}, "Template")` with the established spoof pattern: if old Template was blank, copy new value into oldObject before immutability diff. Same convention as vSphere zonal conversion (line 704). Added 3 unit tests (blank-to-set allowed, set-to-different rejected, set-to-same allowed).

**3. Architecture: backfill as preprocessing** (per Eric's suggestion)
Moved template backfill out of the actuator struct. A standalone `backfillVSphereTemplates()` function enriches the in-memory install config in the controller before actuator creation. The actuator stays stateless about remote cluster state (aligned with GCP's scalar-extraction pattern and Nutanix/OpenStack actuators).

**4. MCO-style 5-field FD matching** (per @jcpowermac)
Replaced datacenter-only matching with the same 5-field conjunction MCO uses (PR #5745, OCPBUGS-77883): Datacenter + Datastore + Server + VMGroup + ResourcePool. This handles all three topologies:

| Topology | Key disambiguator |
|---|---|
| Non-zonal | Datacenter + Datastore (single FD) |
| Compute-cluster zonal | ResourcePool (unique per CC) |
| Host-group zonal | VMGroup = `{infraID}-{fdName}` |

The fd name does not appear in MachineSet metadata (name is `{infraID}-{pool}-{index}`, no fd labels). Workspace fields are the only stable mapping.

**5. Known limitation: pre-corrupted state**
If a prior Hive version already synced the wrong template to the remote MachineSet, the backfill reads that incorrect value. Remediation: set `Topology.Template` explicitly on the CD's failure domain — the webhook change (#2) now permits this for previously-blank fields.

@jcpowermac — does the MCO-style 5-field match satisfy the FD matching concern? This is the same approach MCO landed in PR #5745 after fixing the identical class of bug.

pkg/controller/machinepool/vsphereactuator_test.go
```diff
@@ -152,33 +403,16 @@ func testVSphereClusterDeployment() *hivev1.ClusterDeployment {
 	return cd
 }
 
-func testVSphereMachineSpec(machineType string) machineapi.MachineSpec {
-	rawVSphereProviderSpec, err := vsphereutil.RawExtensionFromProviderSpec(testVSphereProviderSpec())
```
2uasimojo, line 156: IIUC inlining this func is what got rid of ~7KLOC worth of vendored files. Cool 👍

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -76,3 +80,80 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+
+// findTemplateForFD searches remote MachineSets for one whose workspace fields match the
+// given failure domain using a 5-field conjunction (MCO PR #5745 pattern):
+// Datacenter + Datastore + Server + VMGroup + ResourcePool.
+// poolResourcePool overrides fd.Topology.ResourcePool when the MachinePool specifies one.
+func findTemplateForFD(fd *installervsphere.FailureDomain, remoteMachineSets []machineapi.MachineSet, infraID, poolResourcePool string, logger log.FieldLogger) string {
+	// Derive expected VMGroup: only host-group zonal FDs have a VMGroup set on their MachineSets.
+	vmGroup := ""
+	if fd.ZoneType == installervsphere.HostGroupFailureDomain {
+		vmGroup = fmt.Sprintf("%s-%s", infraID, fd.Name)
```
2uasimojo: Is this pattern reliable? The whole reason we're here is because such a pattern changed upstream.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -1393,6 +1393,15 @@ func (r *ReconcileMachinePool) createActuator(
 	case cd.Spec.Platform.OpenStack != nil:
 		return NewOpenStackActuator(masterMachine, r.Client, logger)
 	case cd.Spec.Platform.VSphere != nil:
+		if cd.Spec.Platform.VSphere.Infrastructure != nil && cd.Spec.ClusterMetadata != nil {
```
2uasimojo: I don't agree with positioning this in the actuator creation flow. It's especially problematic that we're modifying the `cd` here, as nothing guarantees something downstream won't `Update()` it, resulting in unintended side effects (though arguably "good" ones in some cases).

The move appears to be related to:

> **3. Architecture: backfill as preprocessing** (per Eric's suggestion)
> Moved template backfill out of the actuator struct. A standalone `backfillVSphereTemplates()` function enriches the in-memory install config in the controller before actuator creation. The actuator stays stateless about remote cluster state (aligned with GCP's scalar-extraction pattern and Nutanix/OpenStack actuators).

I don't think I made such a suggestion -- can you point to it? If I did, I must have worded it poorly :)

The part about "in-memory install config" is valid... but we don't have that yet here. But "before actuator creation" is something I wouldn't have suggested intentionally.

**TL;DR:** This logic should live in the vsphere actuator, modifying the [in-memory install-config](https://github.com/openshift/hive/pull/2921/changes#diff-60f269bfbf2159365622581af8bd650d9a0892550ff057819286af9e4264b251R55) via the [*copy* of the Infrastructure from the CD](https://github.com/openshift/hive/pull/2921/changes#diff-60f269bfbf2159365622581af8bd650d9a0892550ff057819286af9e4264b251R57).

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -76,3 +80,80 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+		if spec.Template == "" || spec.Workspace == nil {
+			continue
+		}
+
+		// 5-field match (mirrors MCO vsphere_helpers.go)
+		if spec.Workspace.Datacenter != fd.Topology.Datacenter ||
+			spec.Workspace.Datastore != fd.Topology.Datastore ||
+			spec.Workspace.Server != fd.Server ||
+			spec.Workspace.VMGroup != vmGroup ||
+			path.Clean(spec.Workspace.ResourcePool) != path.Clean(expectedRP) {
```
2uasimojo: @jcpowermac Please vet?

For reference: https://github.com/openshift/machine-config-operator/pull/5723/changes

In particular, I can't tell if the computation of vmGroup is valid.

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -708,6 +708,24 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateUpdate(admissionSpec
 		}
 	}
 
+	// Allow setting Topology.Template on failure domains where it was previously blank.
+	// This permits the machinepool controller's backfill to persist and allows customers
```
2uasimojo: See, this implies that we're deliberately/explicitly updating the CD after the backfill, which a) I don't see; and b) we shouldn't do.

pkg/validating-webhooks/hive/v1/clusterdeployment_validating_admission_hook.go
```diff
@@ -708,6 +708,24 @@ func (a *ClusterDeploymentValidatingAdmissionHook) validateUpdate(admissionSpec
...
+	// to set the template explicitly via CD edit.
+	if oldObject.Spec.Platform.VSphere != nil && oldObject.Spec.Platform.VSphere.Infrastructure != nil &&
+		cd.Spec.Platform.VSphere != nil && cd.Spec.Platform.VSphere.Infrastructure != nil {
```
2uasimojo: nit: Could fold this into the block above where we've already checked that old/new have `Spec.Platform.VSphere`.

chdeshpa-hue:
Thanks @2uasimojo — working through each point.

**1. Architecture: backfill positioning (machinepool_controller.go:1396)**

You're right, and it's actually worse than the mutation risk you flagged — `cd` comes from `r.Get()` (line 275), which in controller-runtime is a pointer into the informer cache. Mutating `cd.Spec.Platform.VSphere.Infrastructure.FailureDomains` in-place corrupts the shared cache for every controller reading that CD in the same process.

The root cause of the design mistake: we modeled `NewVSphereActuator` after the Nutanix and OpenStack actuators, which have minimal constructors (`masterMachine, client, logger`) because they don't need remote cluster data. But our template backfill *does* need remote cluster data, which left us with nowhere to put it — so it ended up in `createActuator` mutating `cd` directly.

The correct pattern is the **GCP actuator** (`gcpactuator.go:61-110`), which is the only actuator that faces the same problem: it extracts `imageID`, `network`, and `subnet` from `remoteMachineSets` at construction time and stores them on the struct. `cd` is never mutated.

Will refactor to match: `NewVSphereActuator` accepts `remoteMachineSets`/`infraID`/`poolResourcePool`, builds the FD→template map at construction, and applies it in `GenerateMachineSets` after the `DeepCopy()`. This also eliminates the need for the webhook change entirely (nothing persists to the CD), so that will be dropped.

**2. VMGroup computation (vsphereactuator.go:115, 140)**

Waiting for @jcpowermac's review on this.

**3. Attribution**

Apologies for the misattribution — that was my error in summarizing the review thread. Will correct in the next push description.

**4. Webhook nit (clusterdeployment_validating_admission_hook.go:715)**

Moot — the webhook change will be dropped entirely with the architecture fix above.

2uasimojo:
I don't think we should drop the webhook relaxation change. The backfill can still be invalid if the existing remote MachineSet we happen to choose is already invalid.

chdeshpa-hue:
**v3: Addressing review feedback (force-push `29791fd`)**

**1. Architecture: backfill moved into the actuator (machinepool_controller.go:1396)**

You're right — the prior version mutated `cd` in `createActuator`, which is an informer cache pointer. Fixed by following the **GCP actuator pattern**:

- `NewVSphereActuator` now accepts `remoteMachineSets`, `infraID`, and `failureDomains`
- Builds a read-only `templates map[string]string` at construction time via `buildTemplateMap()`
- Templates are applied inside `GenerateMachineSets` on the `DeepCopy()` — the original CD is never mutated
- `createActuator` only passes immutable data to the constructor; no `cd` mutation occurs

**2. Webhook: folded into existing vSphere block (line 715 nit)**

The `Topology.Template` spoof logic is now inside the existing `if oldObject.Spec.Platform.VSphere != nil && cd.Spec.Platform.VSphere != nil` block, not a separate top-level check.

**3. Webhook relaxation kept (per your July 6 comment)**

Confirmed — the spoof pattern stays. If the backfill reads an invalid template from a corrupted remote MachineSet, customers can still manually set the correct value via CD edit.

**4. ResourcePool override bug fix**

Found during code review: `findTemplateForFD` was using the MachinePool's `resourcePool` override for matching, but existing remote MachineSets always carry the FD's default ResourcePool. This meant a new MachinePool with a custom resource pool would fail to discover templates. Fixed: matching always uses the FD's own `Topology.ResourcePool`. The pool override only applies when generating new MachineSets (in `GenerateMachineSets`).

**Still pending:** @jcpowermac vetting of the vmGroup computation + 5-field match approach (from your July 1 request).

pkg/controller/machinepool/vsphereactuator.go
2uasimojo: Still this. @jcpowermac ?

2uasimojo (commented):
This is starting to look sane. I'm still concerned about the hardcoded vmgroup format string. Is there a more generic/future-proof way to handle this?

chdeshpa-hue:
Thanks @2uasimojo — on the future-proofing question specifically:

You're right that `fmt.Sprintf("%s-%s", infraID, fd.Name)` is Hive reconstructing an installer naming convention. I kept exact equality (not a looser match) because for HostGroup FDs the other workspace fields often collide, and VMGroup is the discriminator installer/MCO already use.

That said, I agree reconstructing the formula in Hive isn't ideal if the contract moves again.

A more generic option would be to **stop building the string** and instead **read** the expected VMGroup from the remote `Infrastructure` object we already fetch (`failureDomains[].zoneAffinity.hostGroup.vmGroup`), then exact-compare that to the MachineSet's `Workspace.VMGroup`. Same matching strength, no format knowledge in Hive — closer to how MCO treats Infra as source of truth.

Happy to take that follow-up if you prefer it over the reconstruct. Leaving the "is this pattern durable?" question for @jcpowermac as tagged.

2uasimojo:
We're still looking for @jcpowermac to weigh in on [this](https://github.com/openshift/hive/pull/2921#discussion_r3506673784), right?

chdeshpa-hue:
> We're still looking for @jcpowermac to weigh in on [this](https://github.com/openshift/hive/pull/2921#discussion_r3506673784), right?

Yes, vetting of the vmGroup computation + 5-field match approach (Pending).

pkg/controller/machinepool/vsphereactuator.go
jcpowermac: failure domain contains the vmgroup name why are we generating it?

https://github.com/openshift/api/blob/af5c920502e23802ac1aec0eeb47740520dcb3d6/config/v1/types_infrastructure.go#L1626

what if a customer manually created this group? 
we don't document day2 but doesn't mean someone couldn't do it

jcpowermac:
AI review...
```
Review: PR #2921 — vSphere machinepool template backfill

Stale PR body. It describes constructing the old-style name {infraID}-rhcos-{region}-{zone} and scanning for it — the actual code matches on the workspace conjunction and adopts whatever template the remote MachineSet uses (arguably better: survives user-renamed templates). The  "test evidence" block cites test names that don't exist at head. Rewrite to match.

Untested apply path: the a.templates[fd.Name] application in GenerateMachineSets (vsphereactuator.go:84-93) is never exercised end-to-end — the table test passes nil remote sets. Codecov confirms 10 uncovered lines in vsphereactuator.go. Add one case asserting the generated MachineSet's provider spec carries the backfilled template.

Duplicate test case: vsphereactuator_test.go:253 "MachinePool ResourcePool override does not affect template discovery" is byte-identical to the first case — no override actually exercised.

Webhook matching is index-based (clusterdeployment_validating_admission_hook.go:713-725). Acceptable: name-keyed would be more robust, but any reorder/append/remove of FDs still trips the full-spec immutability cmp on Name/Server etc., so the spoof can't create a false-allow.  

Verified: spoof is side-effect-free — oldObject is decoded fresh per request and later consumers (ingress L748, clusterMetadata L764+, poolRef L822) never read vSphere templates.
 ```

File: pkg/controller/machinepool/vsphereactuator.go [hunk 10]
```diff
@@ -47,20 +63,28 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+		failureDomain := &ic.VSphere.FailureDomains[i]
 		if pool.Spec.Platform.VSphere.ResourcePool != "" {
 			failureDomain.Topology.ResourcePool = pool.Spec.Platform.VSphere.ResourcePool
 		}
 		if len(pool.Spec.Platform.VSphere.TagIDs) > 0 {
 			failureDomain.Topology.TagIDs = pool.Spec.Platform.VSphere.TagIDs
 		}
+		if failureDomain.Topology.Template == "" {
+			if tmpl, ok := a.templates[failureDomain.Name]; ok {
+				failureDomain.Topology.Template = tmpl
```
jcpowermac, line 102: I don't particularly like this. Why would hive overwrite Infra? Or is this only in memory

pkg/controller/machinepool/vsphereactuator.go
chdeshpa-hue: Good catch. The reason we're generating it: hive stores failure domains as installer vendor types (`installervsphere.FailureDomain`), which have `ZoneType` but no `VMGroup` field. Only the configv1 type (`VSpherePlatformFailureDomainSpec.ZoneAffinity.HostGroup.VMGroup`) carries the actual name — and that lives on the spoke's `Infrastructure/cluster` CR, not in the hive CD.

You're right that this breaks for customer-created groups with non-standard names.

The machinepool controller already fetches `Infrastructure/cluster` from the spoke via `getInfrastructure()` (line 471). I'll switch to reading VMGroup from `infra.Spec.PlatformSpec.VSphere.FailureDomains[].ZoneAffinity.HostGroup.VMGroup` and passing a `map[string]string` (fd name → vmGroup) into the actuator instead of `infraID`. That eliminates the `fmt.Sprintf` entirely and handles any VMGroup naming.

pkg/controller/machinepool/vsphereactuator.go
chdeshpa-hue: In-memory only. Line 68 does `cd.Spec.Platform.VSphere.Infrastructure.DeepCopy()` into a local `ic` — all mutations (ResourcePool, TagIDs, Template) target that copy. The original `cd` is never written to. Same pattern as the GCP actuator.

I'll add explicit code comments at both sites to make this obvious.

chdeshpa-hue:
@2uasimojo @jcpowermac — thanks for the review. Here's what I'll do if the direction above works for you:

1. **Replace VMGroup generation with spoke Infrastructure CR read** — drop `fmt.Sprintf("%s-%s", infraID, fd.Name)`, build a `vmGroupMap` from the spoke's `ZoneAffinity.HostGroup.VMGroup` per failure domain, pass that into the actuator. Handles customer-created groups.
2. **Add DeepCopy code comments** — explicit annotations at the `DeepCopy()` call and the template application site so the in-memory-only intent is clear.
3. **Add end-to-end test for template backfill** — a `TestVSphereActuator` case that passes non-nil remote MachineSets through `NewVSphereActuator` → `GenerateMachineSets` and asserts the output MachineSet carries the backfilled template (currently 10 uncovered lines per Codecov).
4. **Fix duplicate test case** — `"MachinePool ResourcePool override does not affect template discovery"` is structurally identical to the non-zonal happy path (`buildTemplateMap` doesn't take a MachinePool). Will either repurpose to test something real or remove.
5. **Rewrite PR body** — update to describe the actual implementation (workspace conjunction matching + template adoption), not the old `{infraID}-rhcos-{region}-{zone}` approach.

Let me know if this works for you.

2uasimojo:
@chdeshpa-hue SGTM, rip it.

pkg/controller/machinepool/machinepool_controller.go
```diff
@@ -1393,7 +1397,12 @@ func (r *ReconcileMachinePool) createActuator(
 	case cd.Spec.Platform.OpenStack != nil:
 		return NewOpenStackActuator(masterMachine, r.Client, logger)
 	case cd.Spec.Platform.VSphere != nil:
-		return NewVSphereActuator(masterMachine, r.scheme, logger)
+		var fds []installervsphere.FailureDomain
+		if cd.Spec.Platform.VSphere.Infrastructure != nil {
+			fds = cd.Spec.Platform.VSphere.Infrastructure.FailureDomains
+		}
+		vmGroupMap := buildVSphereVMGroupMap(infrastructure)
+		return NewVSphereActuator(remoteMachineSets, vmGroupMap, fds, logger)
```
2uasimojo: nit: I would rather pass the more general objects and see these platform-specific calls inside `NewVSphereActuator()`. We try to limit platform-specific semantics to platform-named files -- it helps align our optional platform-specific e2e runs with PRs that actually affect them.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -76,3 +117,70 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+// buildTemplateMap creates a fd-name → template mapping by matching each failure domain
+// (where Template is empty) against remote MachineSets using workspace fields.
+func buildTemplateMap(failureDomains []installervsphere.FailureDomain, remoteMachineSets []machineapi.MachineSet, vmGroupMap map[string]string, logger log.FieldLogger) map[string]string {
+	templates := make(map[string]string)
+	for i := range failureDomains {
+		fd := &failureDomains[i]
+		if fd.Topology.Template != "" {
+			continue
+		}
+		template := findTemplateForFD(fd, remoteMachineSets, vmGroupMap, logger)
```
2uasimojo: This is potentially expensive, particularly if there are a lot of remote msets. So I'd like to enter this logic lazily: i.e. only if we need it to populate a Template. As currently written we're doing it up front every time we create the actuator.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -1,33 +1,64 @@
...
 	return &VSphereActuator{
-		logger: logger,
+		logger:    logger,
+		templates: templates,
 	}, nil
 }
 
+// buildVSphereVMGroupMap reads VMGroup names from the spoke cluster Infrastructure CR.
+// Installer failure domain types do not carry VMGroup; configv1 does via ZoneAffinity.
+func buildVSphereVMGroupMap(infrastructure *configv1.Infrastructure) map[string]string {
```
2uasimojo, line 52: again, I would rather build this lazily, on-demand.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -76,3 +117,70 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+		if fd.Topology.Template != "" {
+			continue
+		}
+		template := findTemplateForFD(fd, remoteMachineSets, vmGroupMap, logger)
+		if template != "" {
+			templates[fd.Name] = template
+			logger.WithFields(log.Fields{
+				"failureDomain": fd.Name,
+				"template":      template,
+			}).Info("found template backfill from remote MachineSet")
```
2uasimojo: I anticipate it being useful to know which remote mset we matched. (We could put this log call in `findTemplateForFD()` to save having to add that datum to the return.)

2uasimojo (commented):
Logically I think we're there. I'd just like to see the code flow/structure tightened up a bit. Thanks!

pkg/controller/machinepool/machinepool_controller.go
chdeshpa-hue: Agreed. vSphere `createActuator` is now a one-liner:

```go
return NewVSphereActuator(cd, remoteMachineSets, infrastructure, logger)
```

Removed `buildVSphereVMGroupMap` and failure-domain extraction from the controller; dropped the unused `installervsphere` import. Platform-specific prep now lives in `vsphereactuator.go`.

pkg/controller/machinepool/vsphereactuator.go
chdeshpa-hue: Removed upfront `buildTemplateMap()` from `NewVSphereActuator()`. The actuator stores `remoteMachineSets` and a per-FD cache; `templateForFailureDomain()` scans only when `GenerateMachineSets` hits an FD with empty `Topology.Template`. If every FD already has a template, no remote scan runs.

pkg/controller/machinepool/vsphereactuator.go
chdeshpa-hue: `buildVSphereVMGroupMap()` is no longer called at actuator construction. Added `vmGroupForFD()` which builds the map on first host-group FD lookup inside `findTemplateForFD()`.

pkg/controller/machinepool/vsphereactuator.go
chdeshpa-hue: Added an Info log in `findTemplateForFD()` on match:

```go
a.logger.WithFields(log.Fields{
    "failureDomain": fd.Name,
    "machineSet":    ms.Name,
    "template":      spec.Template,
}).Info("found template backfill from remote MachineSet")
```

Removed the duplicate log from the deleted `buildTemplateMap()`.

chdeshpa-hue:
Addressed the Aug 26 review feedback in 98d988247:

- Platform-specific prep moved into `vsphereactuator.go` (`createActuator` is now a one-liner)
- Remote MachineSet scan is lazy — only when an FD needs template backfill
- VMGroup map builds on first host-group FD lookup
- `findTemplateForFD()` logs the matched remote MachineSet name

Backfill still runs before MachinePool ResourcePool overrides so workspace matching uses the FD default pool. All unit tests pass.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -118,42 +128,33 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
-		if template != "" {
-			templates[fd.Name] = template
-			logger.WithFields(log.Fields{
-				"failureDomain": fd.Name,
-				"template":      template,
-			}).Info("found template backfill from remote MachineSet")
-		}
+func (a *VSphereActuator) templateForFailureDomain(fd *installervsphere.FailureDomain) string {
+	if fd.Topology.Template != "" {
+		return ""
```
2uasimojo, line 133: This feels a little weird. Based on the name/purpose of the func, this should be returning `fd.Topology.Template`. In theory that would make the caller behave differently... but as the caller is currently written, this condition isn't reachable anyway.

I'm tempted to say we should simply embed the code from this func into the calling context (and remove the unreachable condition). I get that that would require a rewrite of the tests, but I think that's okay: I don't mind if the tests operate at the broader scope of `GenerateMachineSets()` => check the resulting `providerSpec`s for the expected values.

pkg/controller/machinepool/vsphereactuator.go
```diff
@@ -118,42 +128,33 @@ func (a *VSphereActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool
...
+	return tmpl
 }
 
 // findTemplateForFD searches remote MachineSets for one whose workspace fields match the
 // given failure domain using a 5-field conjunction (MCO PR #5745 pattern):
 // Datacenter + Datastore + Server + VMGroup + ResourcePool.
 // Always matches against the FD's own ResourcePool since existing remote MachineSets
 // were created with the FD default, not a MachinePool-level override.
-func findTemplateForFD(fd *installervsphere.FailureDomain, remoteMachineSets []machineapi.MachineSet, vmGroupMap map[string]string, logger log.FieldLogger) string {
+func (a *VSphereActuator) findTemplateForFD(fd *installervsphere.FailureDomain) string {
```
2uasimojo, line 148: I was going to whine about this not caching... but as written we *should* be calling it no more than once for a given FD, so it's cool 👍

2uasimojo (commented):
One actionable comment, but not worth blocking the PR on. Feel free to do it in a followup.

Thanks for your patience with these reviews! This is some tricky code with historically lots of obscure and ugly bugs, so we have to be really careful.

/lgtm
/approve
/verified later,@chdeshpa-hue

2uasimojo:
gdi I never spell this right

/verified later @chdeshpa-hue

PR #2923, opened 2026-06-30
HIVE-3171: Network Policies: Phase 1

2uasimojo:
Add network policies for hive components. Unfortunately, most of these end up being allow-all due to limitations identifying the APIServer and external infrastructure providers. There are several layers in play here:
- hive-operator needs its netpol applied alongside its Deployment in order to start up and create the other components. For development, it's tied into `make deploy` via kustomize. For OperatorHub, it's included in the bundle.
- The controllers (hive-controllers, hive-clustersync, hive-machinepool) and hiveadmission get their netpols laid down by hive-operator. The manifests for these are in bindata.
- Netpols for spoke operations (imgset, provision, deprovision) are created by the clusterdeployment controller alongside their respective ClusterDeployments, parented thereto for automatic cleanup.

## Summary by CodeRabbit

* **New Features**
  * Added/expanded NetworkPolicy support across the hive operator, controllers, and admission components, including provisioning and deprovisioning workflows.
  * Added and applied new NetworkPolicy manifests for hive-operator, hive-controllers, and hiveadmission.
* **Bug Fixes**
  * Improved reconciliation and ownership handling so NetworkPolicy resources are created and managed reliably during install/uninstall flows.
  * Enhanced OpenShift component labeling for improved consistency across workloads.
* **Chores**
  * Updated RBAC and build/asset generation to include full NetworkPolicy lifecycle permissions.

### DIFF
--- a/config/operator/operator_role.yaml
+++ b/config/operator/operator_role.yaml
@@ -236,3 +236,15 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - networkpolicies
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -19,6 +19,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -526,6 +527,27 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
 				constants.SecretTypeLabel:            constants.SecretTypeKubeAdminCreds,
 			},
 			Controlled: false,
+		},
+		{
+			TypeToList: &networkingv1.NetworkPolicyList{},
+			LabelSelector: map[string]string{
+				constants.JobTypeLabel: constants.JobTypeImageSet,
+			},
+			Controlled: true,
+		},
+		{
+			TypeToList: &networkingv1.NetworkPolicyList{},
+			LabelSelector: map[string]string{
+				constants.JobTypeLabel: constants.JobTypeDeprovision,
+			},
+			Controlled: true,
+		},
+		{
+			TypeToList: &networkingv1.NetworkPolicyList{},
+			LabelSelector: map[string]string{
+				constants.JobTypeLabel: constants.JobTypeProvision,
+			},
+			Controlled: true,
 		},
 	}
 }
@@ -1444,6 +1466,23 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 	case apierrors.IsNotFound(err):
 		if areImagesResolved {
 			return nil, r.updateCondition(cd, hivev1.InstallImagesNotResolvedCondition, corev1.ConditionFalse, imagesResolvedReason, imagesResolvedMsg, cdLog)
+		}
+
+		// NOTE: For strict idempotence, we could requeue if the first return is true,
+		// but there's no real benefit and it's more efficient to keep rolling.
+		_, err := controllerutils.EnsureNetworkPolicy(
+			r,
+			r.scheme,
+			constants.JobTypeImageSet,
+			cd,
+			cdLog,
+			// No ingress
+			nil,
+			// Allow all egress (talk to apiserver)
+			[]networkingv1.NetworkPolicyEgressRule{{}},
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to ensure network policy for imageset job")
 		}
 
 		job := imageset.GenerateImageSetJob(cd, releaseImage, controllerutils.InstallServiceAccountName,
--- a/pkg/controller/clusterdeployment/clusterprovisions.go
+++ b/pkg/controller/clusterdeployment/clusterprovisions.go
@@ -12,6 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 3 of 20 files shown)


config/operator/operator_role.yaml
2uasimojo, line 1: TODO: Make sure this ends up in the OLM bundle.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -527,6 +528,14 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
...
 			Controlled: false,
 		},
+		{
+			// Parent all ImageSet NetworkPolicy CRs in the CD's namespace to the CD
+			TypeToList: &networkingv1.NetworkPolicyList{},
+			LabelSelector: map[string]string{
+				constants.JobTypeLabel: constants.JobTypeImageSet,
+			},
+			Controlled: false,
+		},
```
2uasimojo, line 537: Is this (and its cognates) redundant with the SetControllerReference [below](https://github.com/openshift/hive/pull/2923/changes#diff-731524f70bd751975723f23c4ba3f056d2891ceeacdfbec4bacba121331d8a8bR71)? And/or conflicting since that guy sets `Controlled: true` (I think)?

2uasimojo (commented):
In this incarnation, netpols created for imgset and provision (and deprovision, but this one doesn't matter as much because it'll get cleaned up as part of the delete process) stick around until the CD is deleted. Not sure if that's a problem. To remedy, we could do one of two things:
- Parent the netpols to the respective jobs instead of the CD. But this would be a chicken/egg: can't parent to an object that doesn't exist yet (because the owner ref uses UID, which is assigned at creation) but if we create the job first, it'll stall until the netpol is installed, resulting in unnecessary delay while it times out and retries.
- Add code to clean 'em up when their respective Jobs finish. Possible, but kinda ugly.

pkg/controller/utils/netpols.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	np := &v1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: cd.Namespace,
+			Labels:    labels,
+		},
+		Spec: v1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{MatchLabels: labels},
+			Ingress:     ingress,
```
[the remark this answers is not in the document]
2uasimojo: The generators share a labels map between the Job and the pod template, so the pod *does* get the label and the policy matches. Right?

pkg/controller/utils/netpols.go
```diff
@@ -0,0 +1,82 @@
...
+			Egress:      egress,
+			PolicyTypes: []v1.PolicyType{},
+		},
+	}
+	if ingress != nil {
+		np.Spec.PolicyTypes = append(np.Spec.PolicyTypes, v1.PolicyTypeIngress)
+	}
+	if egress != nil {
+		np.Spec.PolicyTypes = append(np.Spec.PolicyTypes, v1.PolicyTypeEgress)
```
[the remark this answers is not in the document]
2uasimojo: Okay, I think I addressed this by always including both PolicyTypes. Now the caller specifying `nil` or `[]` for `ingress` results in "deny all ingress", right?

config/operator/operator_role.yaml
2uasimojo: Update: Per [chai-bot](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1783536055351039), we don't do the OLM work yet. (I wouldn't mind confirmation from a human, though.)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: > Is this (and its cognates) redundant with the SetControllerReference [below](https://github.com/openshift/hive/pull/2923/changes#diff-731524f70bd751975723f23c4ba3f056d2891ceeacdfbec4bacba121331d8a8bR71)? And/or conflicting since that guy sets `Controlled: true` (I think)?

@coderabbitai Can you answer the above?

2uasimojo: @coderabbitai Should I even bother setting the owner reference in EnsureNetworkPolicy? Or is it sufficient to rely on the `generateOwnershipUniqueKeys` funcs?

pkg/controller/clusterprovision/clusterprovision_controller.go
```diff
@@ -559,6 +560,14 @@ func generateOwnershipUniqueKeys(owner hivev1.MetaRuntimeObject) []*controllerut
...
 			Controlled: false,
 		},
+		{
+			// Parent all provision NetworkPolicy CRs in the prov's namespace to the prov
+			TypeToList: &networkingv1.NetworkPolicyList{},
+			LabelSelector: map[string]string{
+				constants.JobTypeLabel: constants.JobTypeProvision,
+			},
+			Controlled: true,
+		},
```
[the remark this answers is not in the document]
2uasimojo: @coderabbitai The deprovision one also steals the owner reference from the CD, eh?

To keep all three parented to the CD, I think we could just move the prov/deprov ones in next to the imgset one. Would that work? Any downsides?

config/operator/operator_role.yaml
suhanime: <Adding for posterity>
Slack thread linked above confirmed that just adding the RBAC for netpols to the CSV is expected for now. The manifests would later need to be added to the /manifests directory in the bundle (Phase 2)
The entries from these rules array are used to set clusterPermissions via bundle-gen, so we are covered for now.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -1446,6 +1468,23 @@ func (r *ReconcileClusterDeployment) resolveInstallerImage(cd *hivev1.ClusterDep
 			return nil, r.updateCondition(cd, hivev1.InstallImagesNotResolvedCondition, corev1.ConditionFalse, imagesResolvedReason, imagesResolvedMsg, cdLog)
 		}
 
+		// NOTE: For strict idempotence, we could requeue if the first return is true,
+		// but there's no real benefit and it's more efficient to keep rolling.
```
suhanime, line 1472: Worried here because, if there is a netpol related error, controller can loop in backoff requeue indefinitely.

pkg/controller/clusterdeployment/clusterprovisions.go
```diff
@@ -139,6 +140,23 @@ func (r *ReconcileClusterDeployment) startNewProvision(
 		return reconcile.Result{}, err
 	}
 
+	// NOTE: For strict idempotence, we could requeue if the first return is true,
+	// but there's no real benefit and it's more efficient to keep rolling.
+	_, err = controllerutils.EnsureNetworkPolicy(
```
suhanime, line 145: Wouldn't it be a better idea to move this to another function that actually reconciles netpol instead of here in startNewProvision where we fire and forget? If somehow the netpol is deleted, the provision will never succeed.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
2uasimojo: That's the same pattern as for most other errors. What would be the alternative?

pkg/controller/clusterdeployment/clusterprovisions.go
2uasimojo: That's true in theory; but the same hole exists for the service account a few lines up, and I don't recall ever hearing of a problem with it.

I did make a deliberate architectural decision not to go "all the way" curating these netpols: a separate controller, finalizers, that kind of thing. If the user tries hard enough, they can certainly break these flows... but then that's on them, neh?

2uasimojo:
Thanks for the review @suhanime! Please lmk if you want to dig into these issues.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
suhanime: Originally I was thinking about bubbling it up via either CD conditions or blocking provision by setting conditions not met. But network policy certainly isn't important enough to block like that.

pkg/controller/clusterdeployment/clusterprovisions.go
suhanime: Yes, fair

PR #2924, opened 2026-07-01
HIVE-3130: Follow-up fixes from openshift/hive#2916 review

suhanime:
Address post-merge review comments from PR #2916:
- Add $YQ env var override (default: yq) so the yq binary name can be customized without PATH changes; update all yq invocations to use it
- Update usage text and jq comment in check_deps to mention that image validation is also skipped for non-quay image repos
- Accept an optional COMMIT-ISH arg in version2.sh main(), and error on unexpected extra args (previously silently ignored)

Assisted-by: Claude Sonnet 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Bundle generation now respects a configurable YAML tool path and validates the expected version, making the process more reliable across environments.
  * Version setup now handles invalid extra arguments more cleanly and supports passing a specific commit or revision when initializing versions.

### DIFF
--- a/hack/bundle-gen.sh
+++ b/hack/bundle-gen.sh
@@ -15,6 +15,7 @@ readonly HIVE_SUB_DIR="operators/hive-operator"
 readonly IMAGE_REPO_DEFAULT="quay.io/openshift-hive/hive"
 readonly COMMUNITY_OPERATORS_UPSTREAM="${COMMUNITY_OPERATORS_UPSTREAM:-git@github.com:redhat-openshift-ecosystem/community-operators-prod.git}"
 readonly CHANNEL_DEFAULT="alpha"
+readonly YQ="${YQ:-yq}"
 
 # Runtime state — set by parse_args
 VERBOSE=false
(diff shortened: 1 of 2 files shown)


hack/bundle-gen.sh
```diff
@@ -438,8 +441,10 @@ check_deps() {
         exit 1
     fi
 
-    if ! yq --version 2>&1 | grep -q 'version v4'; then
-        echo "Error: yq v4 is required (found: $(yq --version 2>&1))" >&2
+    local yq_version
+    yq_version=$("$YQ" --version 2>&1)
+    if ! grep -q 'version v4' <<<"$yq_version"; then
```
2uasimojo, line 446: femtonit: you can do this without forking a new process with something like `if [[ "$yq_version" != *"version v4"* ]]; then`

2uasimojo (commented):
/lgtm

Thanks for the fup!

2uasimojo:
/retest-required

Not sure what's going on there.

PR #2926, opened 2026-07-13
pkg/install: use a functional check for hiveutil

stevekuznetsov:
We would like to support copying the hiveutil image into containers that are not strictly RHEL. So, instead of hard-coding a check for `/etc/redhat-release` and assuming that things will work implicitly, we can just check the binaries directly and use the one that doesn't have linker errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installer compatibility by selecting the first executable `hiveutil` binary available.
  * Added a clear fatal error when no compatible binary can be executed.

### DIFF
--- a/pkg/install/generate.go
+++ b/pkg/install/generate.go
@@ -394,7 +394,7 @@ func InstallerPodSpec(
 			Command:         []string{"/bin/sh", "-c"},
 			// Large file copy here has shown to cause problems in clusters under load, safer to copy then rename to the file the install manager is waiting for
 			// so it doesn't try to run a partially copied binary.
-			Args:         []string{fmt.Sprintf("cp -v /bin/openshift-install /output/openshift-install.tmp && mv /output/openshift-install.tmp /output/openshift-install && major_version=$(sed -n 's/.*release \\([0-9]*\\).*/\\1/p' /etc/redhat-release) && ln -s /output/hiveutil.rhel${major_version} /output/hiveutil && /output/hiveutil install-manager --work-dir /output --log-level debug %s %s", cd.Namespace, provisionName)},
+			Args:         []string{fmt.Sprintf("cp -v /bin/openshift-install /output/openshift-install.tmp && mv /output/openshift-install.tmp /output/openshift-install && if /output/hiveutil.rhel9 version >/dev/null 2>&1; then ln -s /output/hiveutil.rhel9 /output/hiveutil; elif /output/hiveutil.rhel8 version >/dev/null 2>&1; then ln -s /output/hiveutil.rhel8 /output/hiveutil; else echo 'FATAL: neither hiveutil.rhel9 nor hiveutil.rhel8 can execute on this image' >&2; exit 1; fi && /output/hiveutil install-manager --work-dir /output --log-level debug %s %s", cd.Namespace, provisionName)},
 			VolumeMounts: volumeMounts,
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{


hack/test-hiveutil-compat.sh
```diff
@@ -0,0 +1,119 @@
...
+CURRENT_SCRIPT='
+set -e
+echo "=== Current logic (uses /etc/redhat-release) ==="
+major_version=$(sed -n "s/.*release \([0-9]*\).*/\1/p" /etc/redhat-release) || true
+if [ -z "${major_version}" ]; then
+    echo "FAIL: could not determine major version from /etc/redhat-release"
+    echo "      (file missing or unparseable)"
+    exit 1
+fi
+ln -sf /output/hiveutil.rhel${major_version} /output/hiveutil
```
2uasimojo: The absolute paths here and below only work if this script is run in a container (or in the root directory of not-a-container, which would probably be Bad™).

hack/test-hiveutil-compat.sh
```diff
@@ -0,0 +1,119 @@
...
+elif /output/hiveutil.rhel8 version >/dev/null 2>&1; then
+    echo "Selected: hiveutil.rhel8 (binary executes successfully)"
+    ln -sf /output/hiveutil.rhel8 /output/hiveutil
+else
+    echo "FAIL: neither hiveutil.rhel9 nor hiveutil.rhel8 can execute on this image"
+    echo "--- hiveutil.rhel9 output ---"
+    /output/hiveutil.rhel9 version 2>&1 || true
+    echo "--- hiveutil.rhel8 output ---"
+    /output/hiveutil.rhel8 version 2>&1 || true
+    exit 1
```
2uasimojo: If we get here, we never get to run the "fixed" part of this script.

Did we mean to make this section e.g. a function that could be selected based on args, so we could test either path independently?

pkg/install/generate.go
```diff
@@ -394,7 +394,7 @@ func InstallerPodSpec(
...
 			// so it doesn't try to run a partially copied binary.
-			Args:         []string{fmt.Sprintf("cp -v /bin/openshift-install /output/openshift-install.tmp && mv /output/openshift-install.tmp /output/openshift-install && major_version=$(sed -n 's/.*release \\([0-9]*\\).*/\\1/p' /etc/redhat-release) && ln -s /output/hiveutil.rhel${major_version} /output/hiveutil && /output/hiveutil install-manager --work-dir /output --log-level debug %s %s", cd.Namespace, provisionName)},
+			Args:         []string{fmt.Sprintf("cp -v /bin/openshift-install /output/openshift-install.tmp && mv /output/openshift-install.tmp /output/openshift-install && if /output/hiveutil.rhel9 version >/dev/null 2>&1; then ln -s /output/hiveutil.rhel9 /output/hiveutil; elif /output/hiveutil.rhel8 version >/dev/null 2>&1; then ln -s /output/hiveutil.rhel8 /output/hiveutil; else echo 'FATAL: neither hiveutil.rhel9 nor hiveutil.rhel8 can execute on this image' >&2; exit 1; fi && /output/hiveutil install-manager --work-dir /output --log-level debug %s %s", cd.Namespace, provisionName)},
```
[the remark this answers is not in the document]
2uasimojo: Environment tightly controlled; we can be confident /output/hiveutil doesn't exist beforehand. (If it does, that's a bug on whoever supplied that container.)

2uasimojo (commented):
Agree with the rabbit's comments on the test script; but that thing seems like maybe it's a throwaway that wasn't even intended to be checked in?

/retest-required
/override ci/prow/security
/lgtm

rogbas:
cc @2uasimojo

hack/test-hiveutil-compat.sh
stevekuznetsov: That's exactly what this does - runs the script in the container.

stevekuznetsov:
@2uasimojo happy to delete the script, wanted to add it to prove that the approach worked, since we could verify it upfront. Agreed that it doesn't need to be long-lived. If you can use it to verify that this works for base images you care about, that would be good. I checked UBI8, UBI9, Ubuntu and Azure Linux.

2uasimojo:
@stevekuznetsov Our e2e proves that it is backward compatible, which is enough for me. Feel free to delete the test script.

/retest-required

2uasimojo (commented):
/lgtm
/override ci/prow/security

PR #2927, opened 2026-07-20
OCPBUGS-86779: Fix TestAdmission flake caused by APIService routing delay

shahsahil264:
## Summary
Fix intermittent TestAdmission failure caused by kube-apiserver APIService routing not being fully configured even after the APIService reports readiness.

## Root Cause
The test waits for the hiveadmission deployment and APIService to be ready, but there is a gap where the kube-apiserver has not yet wired up request proxying to the aggregated endpoint. During this window, requests hit native RBAC and are rejected with `Forbidden` for `system:anonymous`.

Evidence: first subtest FAILS, but next two subtests PASS within milliseconds — confirming eventual consistency, not a setup failure.

## Fix
Added `waitForAdmissionRouting()` that sends a smoke probe request to the `dnszonevalidators` endpoint in a `wait.PollUntilContextTimeout` loop (1s interval, 2min timeout). Subtests only run once the probe succeeds, confirming routing is fully configured.

## Test Plan
- The fix itself is a test improvement — CI runs will verify the flake is resolved
- `go vet` and `go build` pass on the changed package

Jira: https://redhat.atlassian.net/browse/OCPBUGS-86779

## Summary by CodeRabbit

* **Tests**
  * Updated the post-deploy admission end-to-end flow with an additional routing readiness check to confirm the aggregated admission endpoint is reachable.
  * The test now repeatedly polls for readiness within a set timeout and starts the admission review cases only after routing is confirmed.
  * Transient “forbidden” responses during intermediate setup are treated as expected during polling, with clearer logging for unexpected errors.

### DIFF
--- a/test/e2e/postdeploy/admission/admission_test.go
+++ b/test/e2e/postdeploy/admission/admission_test.go
@@ -2,6 +2,7 @@ package admission
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -67,8 +68,8 @@ func waitForAdmissionRouting(t *testing.T, c dynamic.Interface, gvr schema.Group
 				"kind":    "DNSZone",
 			},
 			"resource": map[string]interface{}{
-				"group":   "hive.openshift.io",
-				"version": "v1",
+				"group":    "hive.openshift.io",
+				"version":  "v1",
 				"resource": "dnszones",
 			},
 			"operation": "CREATE",
@@ -95,10 +96,7 @@ func waitForAdmissionRouting(t *testing.T, c dynamic.Interface, gvr schema.Group
 				t.Logf("Admission routing not ready yet: %v", err)
 				return false, nil
 			}
-			// Any other error is unexpected; keep retrying to be safe,
-			// but log it so we can diagnose if the probe itself is broken.
-			t.Logf("Unexpected error during admission routing probe: %v", err)
-			return false, nil
+			return false, fmt.Errorf("unexpected error during admission routing probe: %w", err)
 		}
 		return true, nil
 	})
--- a/test/e2e/postdeploy/admission/admission_test.go
+++ b/test/e2e/postdeploy/admission/admission_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	admissionv1 "k8s.io/api/admission/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


2uasimojo:
`security` is a known fail. I’ll look at this in the morning. 

/assign @2uasimojo

test/e2e/postdeploy/admission/admission_test.go
```diff
@@ -45,12 +50,73 @@ func waitForAdmission(t *testing.T) bool {
...
+			}
+			return false, fmt.Errorf("unexpected error during admission routing probe: %w", err)
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("Timed out waiting for admission routing to be ready: %v", err)
+		return false
+	}
+	t.Log("Admission routing is ready")
```
2uasimojo, line 108: ✓ This string appears in the [test run](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hive/2927/pull-ci-openshift-hive-master-e2e/2079290248868663296/artifacts/e2e/test/build-log.txt).

We didn't hit the timing issue this time around, but that's okay.

test/e2e/postdeploy/admission/admission_test.go
```diff
@@ -45,12 +50,73 @@ func waitForAdmission(t *testing.T) bool {
...
+	return true
+}
+
 func TestAdmission(t *testing.T) {
 	c := common.MustGetDynamicClient()
 	gvr, _ := (&webhook.DNSZoneValidatingAdmissionHook{}).ValidatingResource()
 	if !waitForAdmission(t) {
 		return
 	}
+	if !waitForAdmissionRouting(t, c, gvr) {
```
2uasimojo: Let's fold this into `waitForAdmission()` like the others.

test/e2e/postdeploy/admission/admission_test.go
```diff
@@ -45,12 +50,73 @@ func waitForAdmission(t *testing.T) bool {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	err := wait.PollUntilContextTimeout(ctx, 1*time.Second, 2*time.Minute, true, func(ctx context.Context) (bool, error) {
+		_, err := c.Resource(gvr).Create(ctx, probe, metav1.CreateOptions{})
+		if err != nil {
+			// The Forbidden error for system:anonymous means kube-apiserver
+			// has not yet wired up routing to the aggregated API server.
+			if strings.Contains(err.Error(), "is forbidden") {
```
2uasimojo: This is okay, but perhaps you would prefer `IsForbidden()` from `k8s.io/apimachinery/pkg/api/errors`?

And if you're going to change this, perhaps a string match on `system:anonymous` would also be appropriate, just to narrow down to the specific issue we were facing.

```suggestion
			if errors.IsForbidden(err) && strings.Contains(err.Error(), "system:anonymous") {
```

2uasimojo (commented):
Thanks for tackling this!

2uasimojo (commented):
/lgtm
/approve
/override ci/prow/security

PR #2931, opened 2026-07-23
OCPBUGS-78995: Azure MachinePool: Accept OSImageID

2uasimojo:
Previously hive optionally accepted `OSImage`, a marketplace image spec, in an Azure MachinePool. If omitted, the installer's MachineSets() generator would leave the image specifier blank and the machine-api-operator would attempt to inject a reasonable default. But in several cases, MAO didn't do a good job of this, resulting in failure to boot the Machines.

In cases where marketplace images were appropriate, this could be worked around by simply filling in the `OSImage` field. However, in environments with no access to the marketplace, or where RHCOS images were not present there, and the managed/gallery image defaults didn't work, the user would be stuck with no recourse but to abandon the MachinePool and manage MachineSets directly.

Here we add a new field to the Azure MachinePool API: `OSImageID`, a string which expects the URN of an image in any of the (three: marketplace, managed, gallery) supported formats. This gives the user the ability to work around any defaulting failure.

## Summary by CodeRabbit

* **New Features**
  * Azure machine pools can select managed or gallery OS images with `osImageID`.
  * Added Azure dual-stack machine-network validation.
  * Added GCP sovereign-cloud OS image validation.
  * Added support for configuring a GCP workload identity token URL.
* **Bug Fixes**
  * Azure image selection now honors `osImageID` and handles resource paths directly.
  * Improved validation and defaults for AWS machine management settings.
* **Documentation**
  * Updated CRD and OpenAPI descriptions for Azure image selection.

### DIFF
--- a/apis/hive/v1/azure/machinepool.go
+++ b/apis/hive/v1/azure/machinepool.go
@@ -14,9 +14,17 @@ type MachinePool struct {
 	// OSDisk defines the storage for instance.
 	OSDisk `json:"osDisk"`
 
-	// OSImage defines the image to use for the OS.
+	// OSImage defines the marketplace image to use for the OS. Mutually exclusive with OSImageID;
+	// if both are specified, OSImage is ignored. If neither is specified, the managed cluster will
+	// inject a default.
 	// +optional
 	OSImage *OSImage `json:"osImage,omitempty"`
+
+	// OSImageID is the resource path of the managed or gallery image to be used on the VMs.
+	// Mutually exclusive with OSImage; if both are specified, OSImage is ignored. If neither is
+	// specified, the managed cluster will inject a default.
+	// +optional
+	OSImageID string `json:"osImageID,omitempty"`
 
 	// NetworkResourceGroupName specifies the network resource group that contains an existing VNet.
 	// Ignored unless VirtualNetwork is also specified.
--- a/pkg/controller/machinepool/azureactuator.go
+++ b/pkg/controller/machinepool/azureactuator.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	azenc "github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
 	machineapi "github.com/openshift/api/machine/v1beta1"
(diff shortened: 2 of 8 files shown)


pkg/controller/machinepool/azureactuator.go
```diff
@@ -127,39 +126,22 @@ func (a *AzureActuator) GenerateMachineSets(cd *hivev1.ClusterDeployment, pool *
...
-			// The HyperVGeneration is germane to the instance/disk type and affects the image used
-			// for the instance. "-gen-2" will be appended to the image name by installazure.MachineSets()
-			// when the HyperVGenerations capability (comma separated list of HyperVGenerations) includes "V2".
-			//
-			// capabilities := map[string]string{
-			//   "HyperVGenerations": "V1,V2",
-			// }
-			//
-			if _, ok := capabilities["HyperVGenerations"]; ok {
-				capabilities["HyperVGenerations"] = "V1"
```
2uasimojo, line 162: This was checking for *managed* (not gallery) images, and erroneously restricting the `capabilities` to `V1` even when no such images were found. However, in recent versions of the vendored installer code, that capability wasn't even being consulted. TL;DR: this chunk of code was a no-op.

With this PR, we *could have* added a bunch of logic to try to paper over the MAO bug -- look for gallery images, look for managed images, validate the generation of the discovered image(s) against the VM size -- but decided it would be too difficult to prove coverage of all edge cases.

vendor/github.com/openshift/installer/pkg/asset/machines/azure/azuremachines.go
2uasimojo, line 1: Note: This is here so we can validate the fix, but will cause `verify` to fail. Ultimately we'll pick up this change via a legit revendor.

apis/hive/v1/azure/machinepool.go
```diff
@@ -14,10 +14,18 @@ type MachinePool struct {
...
 	OSDisk `json:"osDisk"`
 
-	// OSImage defines the image to use for the OS.
+	// OSImage defines the marketplace image to use for the OS. Mutually exclusive with OSImageID;
+	// if both are specified, OSImage is ignored. If neither is specified, the managed cluster will
+	// inject a default.
 	// +optional
 	OSImage *OSImage `json:"osImage,omitempty"`
 
+	// OSImageID is the URN of the (managed, gallery, or marketplace) image to be used on the VMs.
```
patrickdillon: nit: I believe URNs only apply to marketplace images

2uasimojo: Eh, what's the right term for "string that identifies an image"?

2uasimojo: Replaced with "resource path" :shrug:

vendor/github.com/openshift/installer/pkg/asset/machines/azure/azuremachines.go
2uasimojo: > legit revendor

#2940

PR #2935, opened 2026-07-30
RFE-9669: Add spec.platform.aws.amiID to MachinePool

sadiquepp:
Allow specifying a custom AMI directly on a MachinePool's AWS platform spec, rather than only through the existing, explicitly-unsupported hive.openshift.io/image-id-override annotation.

- Add AMIID *string to apis/hive/v1/aws.MachinePoolPlatform.
- In pkg/controller/machinepool/awsactuator.go, resolve the AMI in this order: spec.platform.aws.amiID, then the hive.openshift.io/image-id-override annotation (unchanged), then the existing master-machine lookup fallback.
- Regenerate deepcopy, CRD manifests, and the app-sre SaaS template via make update.

Assisted-by: Claude (Anthropic)

## Summary by CodeRabbit

* **New Features**
  * AWS machine pools can now specify a custom worker boot image via the optional `spec.platform.aws.amiID`.
* **Bug Fixes**
  * AMI/boot image selection now uses a clearer precedence: `spec.platform.aws.amiID`, then the existing cluster annotation (when set), then the existing master-based AMI resolution.
  * Improved handling to prevent boot image ID pointer aliasing during updates.
* **Documentation**
  * Added a “Machine Pools” section explaining custom VM boot images, including an AWS `amiID` example.

### DIFF
--- a/apis/hive/v1/aws/machinepool.go
+++ b/apis/hive/v1/aws/machinepool.go
@@ -43,6 +43,11 @@ type MachinePoolPlatform struct {
 	// this field taking precedence when keys collide.
 	// +optional
 	UserTags map[string]string `json:"userTags,omitempty"`
+
+	// AMIID is the AMI to use for machines in this pool. When omitted, Hive falls back to its
+	// existing behavior of determining an appropriate boot image for the platform.
+	// +optional
+	AMIID *string `json:"amiID,omitempty"`
 }
 
 // SpotMarketOptions defines the options available to a user when configuring
--- a/docs/using-hive.md
+++ b/docs/using-hive.md
@@ -954,6 +954,26 @@ The `spec.replicas` and `spec.autoscaling` configurations cannot be configured s
 
 The `spec.autoscaling.maxReplicas` is an optional field. If it is not configured, then nodes will be auto-scaled without restriction based on resource utilization needs.
 
+#### Customizing VM boot image
+
+By default, we try to determine an appropriate boot image for VMs defined through MachinePools.
+For some platforms, a specific image can be supplied in the MachinePool spec.
+For example, for AWS, the AMI ID can be specified using `spec.platform.aws.amiID` field.
+
+```yaml
+apiVersion: hive.openshift.io/v1
+kind: MachinePool
+metadata:
+  name: mycluster-worker
+  namespace: mynamespace
+spec:
+  clusterDeploymentRef:
+    name: mycluster
+  name: worker
+  platform:
+    aws:
+      amiID: ami-01234567890123456
+```
 ##### Integration with Horizontal Pod Autoscalers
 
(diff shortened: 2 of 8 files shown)


apis/hive/v1/aws/machinepool.go
```diff
@@ -43,6 +43,12 @@ type MachinePoolPlatform struct {
 	// this field taking precedence when keys collide.
 	// +optional
 	UserTags map[string]string `json:"userTags,omitempty"`
+
+	// AMIID is the AMI to use for machines in this pool. When omitted, Hive falls back to its
+	// existing behavior of resolving the AMI from the cluster's master machine (or from the
+	// hive.openshift.io/image-id-override annotation, if set).
```
2uasimojo: I think we should probably not mention the annotation at all in the official API docstring. In fact, with or very shortly after this PR, we may want to consider deprecating it.

docs/using-hive.md
```diff
@@ -954,6 +954,26 @@ The `spec.replicas` and `spec.autoscaling` configurations cannot be configured s
 
 The `spec.autoscaling.maxReplicas` is an optional field. If it is not configured, then nodes will be auto-scaled without restriction based on resource utilization needs.
 
+#### Customizing AMI Images
+
+The AMI image to use for the worker nodes can be customized by specifying the `spec.platform.aws.amiID` field.
```
2uasimojo: > AMI image

RAS syndrome.

----

This part of the document tries to be platform-agnostic as much as possible, even though many of the snippets necessarily use a specific platform's config as an example. Can we reword this to be more generic? Like:

```
#### Customizing VM boot images

By default, we try to determine an appropriate boot image for VMs defined through MachinePools.
For some platforms, a specific image can be supplied in the MachinePool spec.
For example, for AWS, the AMI ID can be specified as follows:
```

I just looked and it turns out Azure is the only other platform where we allow OS image customization. So we _could_ include an Azure example as well. But in general we would prefer to let the API be authoritative rather than trying to keep this doc comprehensive.

pkg/controller/machinepool/awsactuator.go
```diff
@@ -62,8 +62,12 @@ func NewAWSActuator(
 		logger.WithError(err).Warn("failed to create AWS client")
 		return nil, err
 	}
-	amiID := pool.Annotations[hivev1.MachinePoolImageIDOverrideAnnotation]
-	if amiID != "" {
+	var amiID string
+	if pool.Spec.Platform.AWS != nil && pool.Spec.Platform.AWS.AMIID != nil && *pool.Spec.Platform.AWS.AMIID != "" {
+		amiID = *pool.Spec.Platform.AWS.AMIID
+		log.Infof("using AMI from spec.platform.aws.amiID: %s", amiID)
```
2uasimojo: I agree with the rabbit that we should use `logger` rather than `log` here.

Also, prefer structured logging (`logger.WithField("AMIID", amiID).Info("using AMI from spec.platform.aws.amiID")`)

I recognize you're following the neighboring code... but if you wouldn't mind fixing up L71 similarly, that would be great 👍

2uasimojo (commented):
This looks great, thanks! A few minor notes inline.

2uasimojo:
/lgtm
/approve

Thanks @sadiquepp!

/override ci/prow/security

See #2937

/hold to control merge order

2uasimojo:
I think this will rebase, retest, and then merge. We should no longer need to override `security` since #2937.

PR #2944, opened 2026-08-17
HIVE-2671: add netpols to e2e env

grokspawn:
verifies https://github.com/openshift/hive/pull/2923

Adds baseline deny-all networkpolicy objects to e2e so that the defensive policies from #2923 are required to maintain functionality. 

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for network isolation across operator, Hive, cluster, and cluster-pool namespaces.
  * Added validation that namespace migration paths continue to work when ingress and egress traffic are denied by default.
  * Added coverage for automatically applying isolation policies to newly created cluster-pool namespaces.
  * Improved test monitoring and cleanup to ensure isolation policies are applied consistently throughout test execution.

### DIFF
--- a/hack/e2e-common.sh
+++ b/hack/e2e-common.sh
@@ -129,13 +129,19 @@ function wait_for_namespace {
   exit 1
 }
 
-# apply_deny_all_netpol pre-creates namespace $1 (idempotently) and applies a
-# deny-all NetworkPolicy to it. This exercises Hive's non-OLM network policy
-# coverage: if any required netpol is absent or misconfigured, pods in that
-# namespace cannot communicate and the subsequent e2e tests will fail.
+# apply_deny_all_netpol applies a deny-all NetworkPolicy to namespace $1. By
+# default it also (idempotently) pre-creates the namespace; pass "false" as $2 to
+# skip that step -- e.g. from watch_clusterpool_namespaces, where the namespace
+# already exists and re-creating a just-reaped one would be wrong.
+# This exercises Hive's non-OLM network policy coverage: if any required netpol is
+# absent or misconfigured, pods in that namespace cannot communicate and the
+# subsequent e2e tests will fail.
 function apply_deny_all_netpol {
   local ns=$1
-  oc create namespace "${ns}" --dry-run=client -o yaml | oc apply -f -
+  local create_ns=${2:-true}
+  if [[ "$create_ns" == "true" ]]; then
+    oc create namespace "${ns}" --dry-run=client -o yaml | oc apply -f -
+  fi
   oc apply -f - <<EOF
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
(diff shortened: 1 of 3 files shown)


2uasimojo (commented):
This is great, thanks @grokspawn !

I think we may also want to deny-all the [CLUSTER_NAMESPACE](https://github.com/openshift/hive/blob/f0ad9a06c13178e230eb60beb2532ed5031b1df2/hack/e2e-common.sh#L71-L70).

And it would be nice if we could do the same for clusterpool namespaces, but those get created at runtime by the controller, so it would be more difficult (spawn a subthread with a namespace watch, kind of thing).

grokspawn:
> This is great, thanks @grokspawn !
> 
> I think we may also want to deny-all the [CLUSTER_NAMESPACE](https://github.com/openshift/hive/blob/f0ad9a06c13178e230eb60beb2532ed5031b1df2/hack/e2e-common.sh#L71-L70).
> 
> And it would be nice if we could do the same for clusterpool namespaces, but those get created at runtime by the controller, so it would be more difficult (spawn a subthread with a namespace watch, kind of thing).

Added both in https://github.com/openshift/hive/pull/2944/commits/f59f21daa482acdbf845ec094db22631d5114e08, with the slightly-more-complex background watcher as a separate commit so it could be considered separately.

hack/e2e-common.sh
```diff
@@ -129,6 +129,47 @@ function wait_for_namespace {
...
+  while true; do
+    while read -r ns; do
+      [[ -z "$ns" ]] && continue
+      if [[ " $seen " != *" $ns "* ]]; then
+        apply_deny_all_netpol "$ns"
+        seen="$seen $ns"
+      fi
+    done < <(oc get namespace -l 'hive.openshift.io/cluster-pool-name' \
+      -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null)
+    sleep 2
```
2uasimojo: - We can't get away from a race condition here, but we can narrow it way down if we use `-w` instead of `sleep`.
- Using `-o name --no-headers` is a bit cleaner/efficienter than the jsonpath loop.
- femtonit: I personally prefer the form where I see the loop's data before its logic.

In summary:
`oc get ns -l '...' -o name --no-headers -w | while IFS=/ read _ ns; do ...`

- We may not hit it in this test sequence, but in theory we could hot loop if a namespace gets deleted between when we retrieve it and when we idempotently create it in `apply_deny_all_netpol`. The fact that that guy always attempts to create the ns, even in this context, bugs me a little -- perhaps we could either split that func up or have it take a bool to determine whether it does that step?
- Again this is only theoretical, but the efficiency of the mondo-string-as-seen-cache drops as O(#namespaces). Using an associative array would not suffer that penalty.

(I _have_ been accused of overanalyzing non-production shell code in the past.)

hack/e2e-common.sh
```diff
@@ -129,6 +129,47 @@ function wait_for_namespace {
...
+# to each one it hasn't processed yet. Start it with & and capture the PID;
+# kill the PID in the EXIT trap.
+function watch_clusterpool_namespaces {
+  local seen=""
+  while true; do
+    while read -r ns; do
+      [[ -z "$ns" ]] && continue
+      if [[ " $seen " != *" $ns "* ]]; then
+        apply_deny_all_netpol "$ns"
+        seen="$seen $ns"
```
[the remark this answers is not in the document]
2uasimojo: Yeah, it wouldn't be hard for `apply_deny_all_netpol` to produce a success signal on which we could condition the addition of the ns to `seen`.

2uasimojo (commented):
I'm totally not gonna block on these comments. Feel free to address them or unhold.

/lgtm
/approve
/override ci/prow/security

2uasimojo:
/hold

The (intermittent) e2e failure will be addressed by #2945.

2uasimojo:
/hold cancel 

> The (intermittent) e2e failure will be addressed by #2945.

Landed

hack/e2e-common.sh
grokspawn: I prefer the `-w` approach, but I wasn't too fussed either way.  Adopted the `-w` and associative arrays.

grokspawn: The add'l logic for apply_deny_all_netpol to track seen introduces complexity but isn't totally unreadable.

hack/e2e-common.sh
```diff
@@ -129,6 +129,59 @@ function wait_for_namespace {
...
 
+# apply_deny_all_netpol applies a deny-all NetworkPolicy to namespace $1. By
+# default it also (idempotently) pre-creates the namespace; pass "false" as $2 to
+# skip that step -- e.g. from watch_clusterpool_namespaces, where the namespace
+# already exists and re-creating a just-reaped one would be wrong.
+# This exercises Hive's non-OLM network policy coverage: if any required netpol is
+# absent or misconfigured, pods in that namespace cannot communicate and the
+# subsequent e2e tests will fail.
+# The final `oc apply` (the NetworkPolicy) is the function's exit status, so
+# callers can condition on success.
```
2uasimojo: Both e2e suites are running under `set -e`. So while this comment is true, all of the invocations other than the one in `watch_clusterpool_namespaces` would actually kill the whole script if this `apply` fails.

That's probably acceptable/desirable though. No action needed.

hack/e2e-common.sh
```diff
@@ -129,6 +129,59 @@ function wait_for_namespace {
...
+    | while IFS=/ read -r _ ns; do
+        [[ -z "$ns" ]] && continue
+        [[ -n "${seen[$ns]}" ]] && continue
+        # The namespace already exists (we're reacting to it), so don't re-create it.
+        # Only record it as seen once the policy actually applies, so a transient
+        # failure leaves it eligible for a later event.
+        if apply_deny_all_netpol "$ns" false; then
+          seen[$ns]=1
+        else
+          echo "Failed to apply deny-all NetworkPolicy to namespace ${ns}; will retry on next event" >&2
```
2uasimojo: Eh, I guess this is in response to the [rabbit](https://github.com/openshift/hive/pull/2944#discussion_r3805418136). But now that I'm looking at it in situ, I don't like it: it enables a path where we happily allow a clusterpool CD to provision without the deny-all policy ever appearing.

IRL I would expect this branch to be reached virtually never, so it's not a huge deal; but if you respin, let's rip out the error check. Letting the whole test fail (under `set -e`) is both simpler and more correct.

hack/e2e-common.sh
```diff
@@ -129,6 +129,59 @@ function wait_for_namespace {
...
+        fi
+      done
+}
```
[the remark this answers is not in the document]
2uasimojo: Would running under `set -o pipefail` cause the whole script to fail if the `oc get` fails? Given how rarely we expect that to occur, that's probably the simplest and cleanest solution. (Again, not crucial to fix here and now.)

hack/e2e-common.sh
grokspawn: Concur (and yes).  Simplified.

2uasimojo:
/lgtm

Thanks @grokspawn!

PR #2945, opened 2026-08-18
HIVE-3171: operator: defer NetworkPolicy teardown until workload pods terminate

grokspawn:
## What

On `HiveConfig.spec.targetNamespace` change, the operator scrubs its resources from the former target namespace. Today each cleanup loop deletes its allow-all `NetworkPolicy` **before** (and without waiting for) the workload pods to terminate.

In environments where the cluster admin applies a baseline **deny-all** `NetworkPolicy` (a real production posture), removing Hive's allow-all policy while a controller/admission/sharded pod is still shutting down strands that pod without the egress it needs to release its leader lease and exit cleanly — leaving it in `Error` and lingering in the old namespace.

## How

Two-phase old-namespace teardown:

1. **Phase 1** (`deployHive`, `deployHiveAdmission`): delete workloads and their satellite objects, but **not** the `NetworkPolicies`. The repeated delete loop is factored into a shared helper `deleteAssetsFromOldNamespaces`.
2. **Phase 2** (`Reconcile`): once a former target namespace's workload pods have terminated (checked via the `hive.openshift.io/component` label), delete its `NetworkPolicies`; unlabel only fully-scrubbed namespaces and requeue any still draining.

The `hive-controllers` `NetworkPolicy` governs the sharded controllers as well, so `NetworkPolicy` removal is **centralized** — it can't correctly live in any single per-workload loop. `sharded_controllers.go` is therefore unchanged.

## Testing

Adds `pkg/operator/hive/hive_test.go` (deterministic, no cluster; fake kube clientset + gomock):
- `TestHivePodsGone` — label-existence and namespace selectivity of the pod-drain check.
- `TestScrubOldNamespaceNetworkPolicies_PodsPresent` — the guard: `NetworkPolicies` are **not** deleted while workload pods remain.
- `_PodsGone` / `_Mixed` — netpols deleted and namespace reported scrubbed only once drained.

The end-to-end symptom (a pod stranded by blocked egress) is only reproducible with a live CNI, which is why the netpol e2e coverage (#2944) surfaced it only intermittently; this unit test deterministically guards the operator's ordering contract.

## Related

Follow-up to the netpol e2e work in #2944 ([HIVE-2671](https://redhat.atlassian.net/browse/HIVE-2671)), kept as a separate branch/PR so the operator fix reviews independently of the e2e test changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup when Hive workloads move between namespaces.
  * Retains network access policies until workload pods have fully terminated, preventing disruptions during transitions.
  * Removes obsolete policies and namespace labels once cleanup is complete.
  * Automatically retries cleanup when workloads are still running and reports cleanup failures in status.
  * Applies required network access policies in the new target namespace during deployment.

### DIFF
--- a/pkg/operator/hive/apply.go
+++ b/pkg/operator/hive/apply.go
@@ -142,6 +142,24 @@ func applyRuntimeObject(h resource.Helper, rtoFactory toRuntimeObject, hLog log.
 	return h.ApplyRuntimeObject(requiredObj, scheme.GetScheme())
 }
 
+// deleteAssetsFromOldNamespaces deletes each of the named assets (by path) from
+// every namespace in namespacesToClean. It is a no-op for assets that are already
+// absent. This is the shared "phase 1" of old-target-namespace teardown: it
+// removes workloads and their satellite objects but intentionally NOT the
+// allow-all NetworkPolicies, which are removed later (once the workload pods have
+// terminated) by (*ReconcileHiveConfig).scrubOldNamespaceNetworkPolicies.
+func deleteAssetsFromOldNamespaces(h resource.Helper, hiveconfig *hivev1.HiveConfig, namespacesToClean []string, hLog log.FieldLogger, assetPaths ...string) error {
+	for _, ns := range namespacesToClean {
+		for _, assetPath := range assetPaths {
+			hLog.Infof("Deleting asset %s from old target namespace %s", assetPath, ns)
+			if err := deleteAssetByPathWithNSOverride(h, assetPath, ns, hiveconfig); err != nil {
+				return errors.Wrapf(err, "error deleting asset %s from old target namespace %s", assetPath, ns)
+			}
+		}
+	}
+	return nil
+}
+
 func deleteAssetByPathWithNSOverride(h resource.Helper, assetPath, namespaceOverride string, hiveconfig *hivev1.HiveConfig) error {
 	requiredObj, err := readRuntimeObject(assetPath)
 	if err != nil {
--- a/pkg/operator/hive/apply.go
+++ b/pkg/operator/hive/apply.go
@@ -142,12 +142,12 @@ func applyRuntimeObject(h resource.Helper, rtoFactory toRuntimeObject, hLog log.
 	return h.ApplyRuntimeObject(requiredObj, scheme.GetScheme())
 }
 
-// deleteAssetsFromOldNamespaces deletes each of the named assets (by path) from
-// every namespace in namespacesToClean. It is a no-op for assets that are already
-// absent. This is the shared "phase 1" of old-target-namespace teardown: it
+// deleteAssetsFromOldNamespaces idempotently deletes each of the named assets (by
+// path) from every namespace in namespacesToClean, no-opping for assets that are
+// already absent. This is the shared "phase 1" of old-target-namespace teardown: it
 // removes workloads and their satellite objects but intentionally NOT the
-// allow-all NetworkPolicies, which are removed later (once the workload pods have
-// terminated) by (*ReconcileHiveConfig).scrubOldNamespaceNetworkPolicies.
+// NetworkPolicies, which are removed later (once the workload pods have terminated)
+// by (*ReconcileHiveConfig).scrubOldNamespaceNetworkPolicies.
 func deleteAssetsFromOldNamespaces(h resource.Helper, hiveconfig *hivev1.HiveConfig, namespacesToClean []string, hLog log.FieldLogger, assetPaths ...string) error {
 	for _, ns := range namespacesToClean {
 		for _, assetPath := range assetPaths {
--- a/pkg/operator/hive/hive.go
+++ b/pkg/operator/hive/hive.go
@@ -57,19 +57,17 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 		"config/configmaps/install-log-regexes-configmap.yaml",
 		"config/rbac/hive_frontend_serviceaccount.yaml",
 		"config/controllers/hive_controllers_serviceaccount.yaml",
-		// NOTE: This policy applies to sharded controllers as well.
-		"config/netpol/hive-controllers.yaml",
 	}
-	// Delete the assets from previous target namespaces
-	assetsToClean := append(namespacedAssets, deploymentAsset)
+	// NOTE: This policy applies to sharded controllers as well.
+	netpolAsset := hiveControllersNetworkPolicyAsset
+	// Delete the (non-NetworkPolicy) assets from previous target namespaces. The
+	// allow-all NetworkPolicy is intentionally left in place here and removed later,
+	// once the workload pods have terminated, by scrubOldNamespaceNetworkPolicies --
+	// see that function for why the ordering matters.
+	if err := deleteAssetsFromOldNamespaces(h, instance, namespacesToClean, hLog, append(namespacedAssets, deploymentAsset)...); err != nil {
+		return err
+	}
 	for _, ns := range namespacesToClean {
-		for _, asset := range assetsToClean {
-			hLog.Infof("Deleting asset %s from old target namespace %s", asset, ns)
-			// DeleteAssetWithNSOverride already no-ops for IsNotFound
-			if err := deleteAssetByPathWithNSOverride(h, asset, ns, instance); err != nil {
-				return errors.Wrapf(err, "error deleting asset %s from old target namespace %s", asset, ns)
-			}
-		}
 		// The hive-controller binary creates a configmap and lease to handle leader election. Delete them.
 		// TODO: Dedup this with const cmd/manager/main.go:leaderElectionLockName
 		lockName := "hive-controllers-leader"
@@ -298,8 +296,10 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
 	hiveDeployment.Spec.Template.Annotations[hiveConfigHashAnnotation] = computeHash(
 		httpProxy+httpsProxy+noProxy, configHashes...)
 
-	// Load namespaced assets, decode them, set to our target namespace, and apply:
-	for _, assetPath := range namespacedAssets {
+	// Load namespaced assets, decode them, set to our target namespace, and apply.
+	// The NetworkPolicy is applied to the (new) target namespace alongside the rest;
+	// only its removal from *old* namespaces is deferred (see above).
+	for _, assetPath := range append(namespacedAssets, netpolAsset) {
 		if _, err := applyRuntimeObject(h, fromAssetPath(assetPath), hLog, withNamespaceOverride(hiveNSName), withGarbageCollection(instance)); err != nil {
 			hLog.WithError(err).Error("error applying object with namespace override")
 			return err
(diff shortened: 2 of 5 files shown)


pkg/operator/hive/hive.go
```diff
@@ -379,6 +379,77 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
...
+		hLog.WithField("namespace", namespace).WithField("pods", remaining).
+			Info("hive workload pods still terminating; deferring NetworkPolicy removal")
+		return false, nil
+	}
+	return true, nil
+}
+
+// scrubOldNamespaceNetworkPolicies is "phase 2" of old-target-namespace teardown.
+// For each former target namespace whose workload pods have fully terminated, it
+// deletes Hive's allow-all NetworkPolicies. It returns the subset of namespaces
```
2uasimojo: We should avoid saying "allow-all" here, as that characteristic may change in the future, and we'll likely forget to update the comment.

Similar throughout (I stopped marking them eventually).

pkg/operator/hive/hive.go
```diff
@@ -379,6 +379,77 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
...
+// that were fully scrubbed (pods gone AND NetworkPolicies deleted); any namespace
+// still hosting terminating pods is omitted so the caller can requeue and retry.
+//
+// The ordering -- delete workloads first (phase 1, in the deploy* funcs), wait for
+// their pods, then delete these NetworkPolicies -- matters in environments where
+// the admin applies a baseline deny-all NetworkPolicy. Removing Hive's allow-all
```
2uasimojo: "

pkg/operator/hive/hive.go
```diff
@@ -379,6 +379,77 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
...
+	if remaining := len(pods.Items); remaining > 0 {
+		hLog.WithField("namespace", namespace).WithField("pods", remaining).
+			Info("hive workload pods still terminating; deferring NetworkPolicy removal")
+		return false, nil
+	}
+	return true, nil
+}
+
+// scrubOldNamespaceNetworkPolicies is "phase 2" of old-target-namespace teardown.
+// For each former target namespace whose workload pods have fully terminated, it
```
2uasimojo, line 422: it *idempotently* deletes...

pkg/operator/hive/hive.go
```diff
@@ -379,6 +379,77 @@ func (r *ReconcileHiveConfig) deployHive(hLog log.FieldLogger, h resource.Helper
...
+// down in its target namespace. They must be removed from former target
+// namespaces only after the workload pods they protect have terminated.
+var oldNamespaceNetworkPolicyAssets = []string{
+	hiveControllersNetworkPolicyAsset,
+	hiveAdmissionNetworkPolicyAsset,
+}
+
+// hivePodsGone reports whether any Hive workload pods remain in the given
+// namespace. It uses the typed kube client rather than r.List because the
+// operator's dynamic clientFor has no case for Pods and would panic.
```
2uasimojo: It's trivial to add `"pods"` to [clientFor()](https://github.com/openshift/hive/blob/298a3d559283a575f613d1e2202b8f5effd4ed72/pkg/operator/hive/dynamicclient.go#L25).

(I don't have any particular objection to using the typed client, assuming it doesn't have caching problems, but if you decide to keep it, I don't think we need this comment.)

pkg/operator/hive/apply.go
```diff
@@ -142,6 +142,24 @@ func applyRuntimeObject(h resource.Helper, rtoFactory toRuntimeObject, hLog log.
...
+// every namespace in namespacesToClean. It is a no-op for assets that are already
+// absent. This is the shared "phase 1" of old-target-namespace teardown: it
+// removes workloads and their satellite objects but intentionally NOT the
+// allow-all NetworkPolicies, which are removed later (once the workload pods have
+// terminated) by (*ReconcileHiveConfig).scrubOldNamespaceNetworkPolicies.
+func deleteAssetsFromOldNamespaces(h resource.Helper, hiveconfig *hivev1.HiveConfig, namespacesToClean []string, hLog log.FieldLogger, assetPaths ...string) error {
+	for _, ns := range namespacesToClean {
+		for _, assetPath := range assetPaths {
+			hLog.Infof("Deleting asset %s from old target namespace %s", assetPath, ns)
+			if err := deleteAssetByPathWithNSOverride(h, assetPath, ns, hiveconfig); err != nil {
```
2uasimojo, line 155: Double checked: this is [idempotent](https://github.com/openshift/hive/blob/8b81f2d08e0fb3b8d3910d251eefd2fe7d15724e/pkg/resource/delete.go#L61-L62) ✓

pkg/operator/hive/hive_controller.go
```diff
@@ -610,11 +610,27 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	// If we get here, we've successfully scrubbed all *our* resources out of previous target
+	// Phase 2 of old-target-namespace teardown: now that the workloads have been
+	// deleted (phase 1, in the deploy* funcs above), remove Hive's allow-all
```
2uasimojo:
```suggestion
	// deleted (phase 1, in the deploy* funcs above), remove Hive's
```

pkg/operator/hive/hive_controller.go
```diff
@@ -625,6 +641,11 @@ func (r *ReconcileHiveConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	if len(fullyScrubbed) < len(namespacesToClean) {
+		hLog.Info("waiting for workload pods to terminate in former target namespaces before removing their NetworkPolicies; will retry")
```
2uasimojo: We could use set math here to include the remaining namespaces in the message. (I can't think of a reason we shouldn't use a set for `namespacesToClean` from the start.)

pkg/operator/hive/hive_test.go
```diff
@@ -0,0 +1,141 @@
...
+	h.EXPECT().Delete(gomock.Any(), gomock.Any(), "gone", "hiveadmission").Return(nil)
+
+	r := &ReconcileHiveConfig{
+		kubeClient: fakekubeclient.NewSimpleClientset(componentPod("busy", "hive-controllers-abc")),
+	}
+
+	scrubbed, err := r.scrubOldNamespaceNetworkPolicies(h, &hivev1.HiveConfig{}, []string{"gone", "busy"}, testLogger())
+	require.NoError(t, err)
+	assert.Equal(t, []string{"gone"}, scrubbed)
+}
```
2uasimojo, line 110: We need to demonstrate deletion of netpols. We may not need new test cases: it may suffice just to include one or both netpols in the clientsets and then `EXPECT()` them to be `Delete()`d or not, as appropriate.

pkg/operator/hive/hive_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+// unlabeledPod builds a Pod with no hive.openshift.io/component label (e.g. some
+// unrelated pod that happens to live in the namespace).
+func unlabeledPod(namespace, name string) *corev1.Pod {
+	return &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name},
+	}
+}
```
2uasimojo: nit: these could be collapsed into one func that takes a bool to add the component label or not.

pkg/operator/hive/hive_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func TestHivePodsGone(t *testing.T) {
```
2uasimojo: I appreciate the thoroughness, but if we can demonstrate coverage via the `scrub...` suites, it's probably overkill to have a whole suite for the `hivePodsGone` helper.

pkg/operator/hive/hive.go
grokspawn: It's trivial to add, but it means that the test has to create a fake dynamic client.  This attempt avoids it.  Let me know if you feel strongly enough about it.

pkg/operator/hive/hive_controller.go
grokspawn: Changing namespacesToClean to a set across the full usage would be a massive scope increase beside this fix.  It also could be performed separately.  Are you sure that's what you want?

pkg/operator/hive/hive_test.go
grokspawn: It's been my experience that bools in function params stack and lose meaning quickly (hence the strength of the options pattern).  Given the triviality of the functions I posited them as explicit, but there are in-between options.

pkg/operator/hive/hive_test.go
grokspawn: I'll retract this file, then.

grokspawn:
Thanks @2uasimojo — pushed 51b44f01f addressing the feedback:

- **"allow-all" wording** — removed throughout; comments now say "the NetworkPolicies Hive manages" so they don't drift if the policy changes.
- **`idempotently`** — noted in the `deleteAssetsFromOldNamespaces` doc comment.
- **`hive_controller.go` comment suggestion** — applied.
- **Remaining namespaces in the requeue log** — now computed via set math (`sets.New(...).Difference(...)`) and included in the message. I left `namespacesToClean` itself as `[]string` for now to keep the diff contained — happy to convert it to a `sets.Set[string]` end-to-end in a follow-up if you'd prefer.
- **`clientFor` / typed client** — kept the typed client and removed the justifying comment (keeps the unit test on a simple fake clientset). Glad to add a `pods` case to `clientFor()` and switch to `r.List` instead if you'd rather have the consistency.
- **Tests** — collapsed the two pod helpers into one; dropped the standalone `hivePodsGone` suite. Coverage is now via the `scrubOldNamespaceNetworkPolicies` table (which includes a label-selectivity case so unrelated pods don't block scrubbing). On demonstrating netpol deletion: the `PodsGone`/`unrelated pods`/`Mixed` cases assert `Delete()` is called for both netpols via the mock helper, and the `pods present` case relies on the strict mock to fail if any deletion happens — so both the "deleted" and "not deleted" directions are asserted.

pkg/operator/hive/hive_test.go
grokspawn: ... but the clanker unilaterally did it anyway.  sigh...

pkg/operator/hive/hive_controller.go
grokspawn: For now, just did local set math instead of changing the type through the callchains.

pkg/operator/hive/hive_test.go
2uasimojo: Ohh, I misread this initially, the `EXPECT().Delete()`s *are* the netpols -- I thought they were the pods :facepalm:

pkg/operator/hive/hive_test.go
```diff
@@ -0,0 +1,110 @@
...
+			ctrl := gomock.NewController(t)
+			h := mock.NewMockHelper(ctrl)
+			if tc.expectScrubbed {
+				// apiVersion/kind come from the decoded asset; match on namespace + name.
+				h.EXPECT().Delete(gomock.Any(), gomock.Any(), ns, "hive-controllers").Return(nil)
+				h.EXPECT().Delete(gomock.Any(), gomock.Any(), ns, "hiveadmission").Return(nil)
+			}
+			// When not expectScrubbed, the strict mock has no Delete expectations, so any
+			// NetworkPolicy deletion fails the test -- this is the assertion that guards
+			// against deleting a NetworkPolicy while workload pods still exist.
```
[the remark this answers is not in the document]
2uasimojo: This would have obviated my boneheaded comment [earlier](https://github.com/openshift/hive/pull/2945/changes/158d3435153b41adc356389ca60b977eefbeae89..51b44f01fef710af77979969b64668d9c66ee1b1#r3807239336) 😬

2uasimojo (commented):
/lgtm
/approve
/override ci/prow/security

pkg/operator/hive/hive_controller.go
2uasimojo: Yah, that's fine. I meant to mention that not all of my comments demanded action :P

pkg/operator/hive/hive.go
2uasimojo: No, not a big deal, I just didn't like the comment :)

